PR #5428, opened 2026-05-21
fix: recover main from #5421 wechat-sidecar fallout (orphan helpers + utoipa)

houko:
## What

Recovers main from two `cargo check --workspace --lib --tests` failures introduced by #5421 (WeChat in-process → sidecar):

1. **`crates/librefang-types/src/config/types.rs:5666, 5671`** — `default_channel_max_backoff_secs` and `default_channel_initial_backoff_2s` were `serde(default)` hooks for in-process channel exponential-backoff fields. With WeChat / Feishu / WeCom / Signal / Matrix all on sidecars, no consumer remains and `warnings = "deny"` (workspace lint) fires on the orphans. Delete. Record the deprecation in the comment block that already documents the earlier `default_channel_initial_backoff_secs` removal so the next channel migration knows what shape to revive when it needs the hooks back.

2. **`crates/librefang-api/src/openapi.rs:206-207`** — the utoipa OpenAPI bundle still referenced `routes::wechat_qr_start` and `routes::wechat_qr_status`, which #5421 deleted from `routes/channels.rs` (the mod comment at line 36 records the removal). Drop the references; equivalent endpoints now live on the WeChat sidecar HTTP surface, not on the API router.

## Verification

`cargo check --workspace --lib --tests` clean.

## Scope

Strictly the two main-red symptoms. Same root cause (#5421 fallout) so bundled per AGENTS.md tight-cluster rule.


houko:
Closing as a no-op: the main-recovery this PR was opened for already landed via another path, so the diff against current main is empty (additions=0). Nothing left to merge.

PR #5442, opened 2026-05-21
fix(api/auth): trusted-proxy gate on X-Forwarded-Proto for session cookies

### LINKED ISSUE #5441: [Medium] X-Forwarded-Proto honored without trusted-proxy check — Secure-cookie downgrade
# `X-Forwarded-Proto` is honored without a trusted-proxy check, allowing `Secure`-cookie downgrade

**Severity:** Medium
**Category:** HTTP API authentication & authorization
**Labels:** `security`, `auth`, `medium`

## Affected files
- `crates/librefang-api/src/server.rs:340-346` (`request_is_https`)
- `crates/librefang-api/src/server.rs:352-358` (`session_cookie_attrs`)
- `crates/librefang-api/src/server.rs:1061-1067` (`trusted_proxies` config exists but is not consumed)

## Description

`request_is_https` reads the request's `X-Forwarded-Proto` header unconditionally, and `session_cookie_attrs` uses that to decide whether to attach `Secure`. The `trusted_proxies` config already exists but is not consulted here.

- An attacker can set `X-Forwarded-Proto: https` against a plain-HTTP bind to force their own `Secure` flag (low impact).
- **The critical case**: when a TLS proxy fails to forward that header (a common naive nginx setup), real HTTPS sessions emit cookies **without `Secure`**; a single accidental HTTP redirect leaks them on the wire.

## Recommendation

Pick one:

1. Whenever any authentication is enabled, always attach `Secure`, and document "HTTP is dev-only"; or
2. Trust `X-Forwarded-Proto` only when the request source is in `state.trusted_proxies`, mirroring the existing pattern in `client_ip.rs`.

Add an integration test for "no trusted proxy configured + forged HTTP header" → does not attach `Secure`.

houko:
Closes #5441.

## Audit

`docs/issues/x-forwarded-proto-trusted-proxies.md` (Severity: **Medium**, Domain: HTTP auth).

`request_is_https` read `X-Forwarded-Proto` unconditionally; `session_cookie_attrs` used the result to decide whether to attach `Secure`. The `state.trusted_proxies` config — already wired into `client_ip::resolve_real_client_ip` — was not consulted here. Two failure modes:

* Attacker connecting directly to a plain-HTTP bind can forge `X-Forwarded-Proto: https` to attach a `Secure` flag to their own cookie (low-impact since they control the request anyway).
* **CRITICAL case**: a TLS-terminating proxy (common naive nginx setup) that **forgets to set the header** — real HTTPS sessions emit cookies WITHOUT `Secure`, and a single accidental HTTP redirect leaks them on the wire.

## Change

`crates/librefang-api/src/server.rs`:

* `request_is_https(headers, peer, trusted_proxies)` now fail-closes when peer is unknown OR not in `trusted_proxies`. Mirrors the trust model in `client_ip::resolve_real_client_ip`.
* `session_cookie_attrs(headers, peer, trusted_proxies)` threads the new parameters through.
* `dashboard_login` + `dashboard_logout` add an explicit `ConnectInfo<SocketAddr>` extractor and pass `state.trusted_proxies` to the cookie-attr builder.

## Tests

Five new cases in `server::xfp_trust_tests`:

* `untrusted_peer_with_xfp_https_header_is_not_treated_as_https` — the attacker-forges-header path. Pre-fix this returned true; now false.
* `trusted_peer_with_xfp_https_header_is_treated_as_https` — happy path (real TLS proxy in the configured allowlist).
* `trusted_proxy_with_no_header_is_not_treated_as_https` — pins the audit-flagged CRITICAL case so a TLS proxy that forgets to set the header leaves the cookie without `Secure`. Operators see a misconfig signal instead of silent downgrade.
* `missing_peer_collapses_to_not_https` — misordered router collapses to untrusted (fail-closed).
* `empty_trusted_proxies_collapses_to_not_https` — default deployment shape: no `trusted_proxies` configured means the header is never honoured, even from loopback. Operators that terminate TLS on localhost have to configure `trusted_proxies = ["127.0.0.1"]` explicitly.

`cargo test -p librefang-api --lib xfp_trust` → 5/5 pass. `cargo clippy -p librefang-api --all-targets -- -D warnings` clean.

## Prerequisite

Carries two prerequisite commits (cherry-picks of #5428 and #5437, which recover main from #5421 wechat-sidecar and #5433 teams-sidecar fallout respectively). Both dedupe on rebase once the prereqs land.


houko:
Superseded by #5581, which merged to main as 394a763b726e9ea67d333b1f15803861731f4732. Both PRs target the same audit item (`docs/issues/x-forwarded-proto-trusted-proxies.md`) and arrive at the same trust gate: `request_is_https` / `session_cookie_attrs` only honour `X-Forwarded-Proto` when the TCP peer is in `state.trusted_proxies`, with `dashboard_login` extracting `ConnectInfo<SocketAddr>` and passing the peer through.

A few differentials worth noting so you can confirm nothing of yours is left on the table:

- The merged version covers every test case here (untrusted-peer-ignored, trusted-peer-honoured, trusted-but-no-header → plain HTTP, empty-trusted-proxies fail-closed) and adds the **multi-value RFC 7239 leftmost-wins** case (`https,http` vs `http,https`) that this PR lacks.
- The merged version omits the `dashboard_logout` plumbing because `session_cookie_clear_attrs()` is now unconditional `Secure` (separate audit item `logout-no-secure-cookie`), so threading peer/trusted_proxies into logout isn't needed.
- `missing_peer_collapses_to_not_https` is structurally not applicable to the merged signature — main takes a non-Option `IpAddr` from `ConnectInfo<SocketAddr>` since axum guarantees `ConnectInfo` once `into_make_service_with_connect_info` is wired at the listener.
- The audit doc this PR deletes (`docs/issues/x-forwarded-proto-trusted-proxies.md`) was already removed on main alongside the #5581 merge.

Recommending closure in favour of the merged PR. Thanks for the work on this.

houko:
Closing as superseded by #5581 (merged 394a763b726e9ea67d333b1f15803861731f4732). The same trusted-proxy gate on `X-Forwarded-Proto` for the session-cookie `Secure` flag is live on main in `request_is_https` (fail-closed when `trusted_proxies` is empty), and #5581 is a superset — it also handles RFC 7239 multi-value leftmost-wins. Rebasing this branch would be a no-op re-implementation with a substantive conflict in server.rs.

Correction to my earlier comment: the audit doc `docs/issues/x-forwarded-proto-trusted-proxies.md` is in fact still present on main (#5581 touched only server.rs). I'll remove it in a separate trivial cleanup commit rather than carrying this PR.

PR #5459, opened 2026-05-21
feat(channels)!: migrate google_chat from in-process adapter to sidecar

houko:
## What

Migrates Google Chat from the in-process Rust adapter (`librefang-channels::google_chat`, 818 lines) to a stdlib-only Python sidecar (`librefang.sidecar.adapters.google_chat`). Behaviour parity preserved end-to-end. Net diff: +1550 / −2342 = **−792 lines**.

Follows the established `feat(channels)!:` migration pattern: matrix #5368 → feishu #5380 → email #5408 → dingtalk #5417 → wechat #5421 → teams #5433 → wecom (multiple) → whatsapp #5445 → **google_chat (this PR)**. `webhook` is now the only remaining in-process channel.

## Sidecar (Python, stdlib-only)

`sdk/python/librefang/sidecar/adapters/google_chat.py` (688 lines):

* In-module RS256 JWT signer — PKCS#8 PEM parser + ASN.1 DER reader + RSA modular-exponentiation via stdlib `int.pow` + PKCS#1 v1.5 SHA-256 padding. No `pyjwt` / `google-auth` / `cryptography` dependency.
* `BaseHTTPRequestHandler` over `HTTPServer` for inbound; 1 MiB body cap pre-allocation.
* `urllib.request.urlopen` outbound with 30 s timeout + 4096-char UTF-8-safe chunking.
* `_TokenCache` (`threading.Lock`-backed) with 5-minute refresh buffer.
* Pre-supplied `access_token` fallback path preserved from Rust adapter.

`sdk/python/tests/test_google_chat_adapter.py` — **27/27 passing** (env enforcement / PEM round-trip / SSRF allowlist / webhook parsing all 8 cases / UTF-8 chunking / outbound 401 cache-clear / JWT shape vs. 2048-bit test key).

### Three improvements over Rust

1. **401 invalidates cached token** — Rust kept stale tokens until TTL; sidecar re-runs JWT auth on the next send.
2. **1 MiB hard cap on webhook body** — Rust inherited axum's 2 MiB default; sidecar rejects oversized `Content-Length` before allocation.
3. **Start-time PEM validation** — Rust deferred RSA parse to first send; sidecar caches `(n, d)` in `__init__`, failing boot on invalid PEM.

## Rust deletion cascade

`crates/librefang-channels/src/google_chat.rs` (818 lines) deleted; cargo feature `channel-google-chat` + optional `rsa` dep removed; allowlist entry removed; `SIDECAR_CATALOG` entry added. Knock-on cleanups in `channel_bridge.rs` / `routes/channels.rs` / `routes/config.rs` / `channel_sender.rs` / `validation.rs` / `serde_helpers.rs` (orphan `deserialize_string_or_int_vec`). `GoogleChatConfig` + Default + `channels.google_chat` field removed from `librefang-types`.

OpenClaw migrator: legacy `[channels.google_chat]` now records a `SkippedItem` (matrix / feishu / teams shape). Four channel-table helpers (`map_dm_policy` / `map_group_policy` / `build_channel_table` / `allow_from_to_toml_array`) deleted — google_chat was their last consumer.

## Witness rotations

* `routes/channels.rs::missing_required_env_returns_412` + `_credentials_present_no_target_returns_200` retired — in-process witness pool of channels with `required: true` secret env vars is empty.
* `instance_helper_tests` witness rotates `google_chat` → `webhook` (4 `resolve_secret_env_overrides_*` tests adapt to `secret_env` / `WEBHOOK_SECRET`).
* CLI `init <channel>` wizard match collapses to fall-through (`maybe_write_channel_config` / `notify_daemon_restart` `#[allow(dead_code)]` for future use).

## Bundled main-red recovery

This PR incidentally fixes pre-existing main-red conditions blocking `cargo check --workspace --lib --tests` on plain `origin/main`:

* `default_local_probe_interval_secs` re-added (referenced at types.rs:3241 + 5754, defined nowhere — orphan from earlier sidecar cleanup).
* `default_channel_max_backoff_secs` + `default_channel_initial_backoff_2s` removed (dead code, last consumers migrated to sidecars).
* `routes::whatsapp_qr_*` + `routes::wechat_qr_*` utoipa references in `openapi.rs` removed (stale after WhatsApp / WeChat migrations).
* `openclaw.rs` test `cfg.channels.whatsapp.iter()` switched to `report.skipped` assertion.
* `librefang-types/config/mod.rs` test fixtures rotated `GoogleChatConfig::default()` → `WebhookConfig::default()`.

If a maintainer prefers, these can be split off into a separate prerequisite PR — let me know.

## Verification

* `cargo check --workspace --lib --tests` clean.
* `cargo clippy --workspace --all-targets -- -D warnings` clean.
* `cargo test -p librefang-channels --lib` → 418/418 pass.
* `pytest sdk/python/tests/test_google_chat_adapter.py` → 27/27 pass.
* `kernel_config_schema.golden.json` regenerated via the canonical `--ignored regenerate_golden` test.

## Out of scope

* Doc pages (`docs/src/app/configuration/channels/page.mdx`, `docs/src/app/zh/...`) still describe the in-process `[channels.google_chat]` shape — separate doc sweep following the established pattern. Dashboard `fld_google_chat` locale key (optional cleanup).
* Real end-to-end Google Workspace token-exchange smoke test — I don't have a service-account key. The RS256 signer is exercised against a generated 2048-bit test key (PEM parse, signature byte length, header/claims round-trip). Production smoke is operator-side.

## Operator action

Existing `[channels.google_chat]` block no longer recognised. Replace with `[[sidecar_channels]]`:

```toml
[[sidecar_channels]]
name = "google_chat"
command = "python3"
args = ["-m", "librefang.sidecar.adapters.google_chat"]
channel_type = "google_chat"

[sidecar_channels.env]
GOOGLE_CHAT_WEBHOOK_PORT = "8090"
# optional: GOOGLE_CHAT_SPACE_IDS, GOOGLE_CHAT_ACCOUNT_ID
```

Move the service-account JSON into `~/.librefang/secrets.env` as `GOOGLE_CHAT_SERVICE_ACCOUNT_JSON=<full JSON blob>` (one line; embed `\n` for the PEM newlines), and repoint the Bot's messaging endpoint in Google Cloud Console from `https://<host>/channels/google_chat/webhook` to `https://<host>:<GOOGLE_CHAT_WEBHOOK_PORT>/webhook`.


houko:
Self-review caught two real bugs in the first commit. Both fixed in cbe57f96 (pushed):

1. **`crates/librefang-channels/src/google_chat.rs` was not actually deleted** — the previous commit removed the `pub mod` declaration and the cargo feature but the 818-line adapter file stayed in the tree. The workspace compiled because nothing imported it, so it was a dead file riding along. The fixup commit stages the actual deletion.

2. **`on_send` reached for `cmd.user_id`** — the `Send` dataclass has `channel_id` + `text` + `content` + `thread_id` + `user`, no `user_id` field. Would have AttributeError'd on the first outbound message. Switched to the `cmd.channel_id or cmd.user.get("platform_id")` shape that teams.py / whatsapp.py use, plus an empty-text drop.

Added 5 new `on_send` tests against the real `Send` dataclass that would have caught the second bug. pytest now 32/32 (was 27/27). `cargo check --workspace --lib --tests` still clean.

Apologies for the noise. The original PR body's "verification" claim was honest at the time — both bugs only became visible by re-reading the diff against the SDK after pushing.

houko:
Rebased onto current main (4 commits ahead: #5454 / #5455 / #5456 / #5460). Force-pushed (with-lease) — `f88df563`.

**What changed on rebase** (vs. the prior force-push):

- **#5455 wins for webhook**: `webhook` is also a sidecar now. My PR no longer touches the `webhook` field (gone from `ChannelsConfig`); the SIDECAR_CATALOG keeps both `webhook` (from #5455) and `google_chat` (added here).
- **#4ab144c4 wins for the bundled main-red fixes**: my duplicate `default_local_probe_interval_secs` add / orphan-backoff removal / `whatsapp_qr_*` openapi cleanup / openclaw `_ch` rename are dropped — upstream did them.
- **`instance_helper_tests` 4-test suite retired**: the witness pool is empty (was rotated to google_chat in #4ab144c4 → no replacement after this PR). Original 412/200 pair already retired pre-rebase for the same reason.
- **`config/mod.rs::test_one_or_many_*` 4 tests re-pointed** at a local fixture struct `OoMTestRow` so OneOrMany regression coverage survives the empty-witness world. `test_channels_config_with_new_channels` + `test_all_new_channel_configs_serde` reduced to default-round-trip assertions (no specific channel to use).
- **`config_reload.rs` ReloadChannels test** rotates witness `google_chat` → `file_download_max_bytes` (the only non-`OneOrMany` field still on `ChannelsConfig`).
- **`routes/skills.rs` channel-instance helpers** switched their typed `Doc { channels: ChannelsConfig }` round-trip to a raw `toml::Table` since `ChannelsConfig` no longer carries any channel-named field.
- **`channel_sender.rs` + `messaging.rs`** macro+local-binding sites underscore-prefixed (`_channels` / `_agent_name`) and macros marked `#[allow(unused_macros)]` so `for_each_channel_field!` expanding to nothing doesn't trip `-D warnings`. Shape preserved so a future in-process channel re-enables both loops by appending one macro arm.
- **`channel_bridge.rs::channel_overrides`** same treatment; `start_channel_bridge_with_config` drops the now-empty `check_channel!` macro entirely.
- **`routes/config.rs` `ch!` block** collapsed to an empty `serde_json::Map`; macro shape kept as a comment block so the recipe is re-buildable.
- **`routes/channels.rs::FieldType::Number`** marked `#[allow(dead_code)]` (no in-process consumer left after the empty `ChannelMeta` fields).
- **`routes/channels.rs::webhook_route_suffix`** collapsed from `match channel_name { _ => None }` (clippy flagged the single-binding match) to a direct `let _ = channel_name; None`.
- **CHANGELOG entry rewritten**: "the last remaining in-process channel" → "after this PR there are **zero in-process channels**"; "rotates witness to webhook" → "retires the tests entirely"; pytest count 27 → 32 (the 5 `on_send` tests from the cbe57f96 fix); test-count parenthetical reflects the rebased reality.

**Verification (post-rebase)**:
- `cargo check --workspace --lib --tests` clean.
- `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `pytest sdk/python/tests/test_google_chat_adapter.py` → 32/32 pass.
- `kernel_config_schema.golden.json` regenerated against the post-rebase types.

Final PR shape: 2 commits (the feat + the cbe57f96 fix), 22 files, +1740 / −2899 = **−1159 net**.

houko:
Self-review (diff vs. teams.py / webhook.py) caught 5 operational shape bugs. Fixed in `2efc71d6`:

1. **`on_shutdown` blocked the asyncio loop** — `ThreadingTCPServer.shutdown()` waits for `serve_forever()` to exit. Extract `_shutdown_server()` that dispatches to a daemon thread + returns immediately (mirrors teams.py).
2. **`produce()` didn't catch `asyncio.CancelledError`** — framework cancellation never reached `on_shutdown`, leaking the listening socket + worker thread past SIGTERM. Wrap in `try / except asyncio.CancelledError: self._shutdown_server(); raise`.
3. **Single-threaded `HTTPServer`** — switched to `socketserver.ThreadingTCPServer` with `daemon_threads = True` + `allow_reuse_address = True` so concurrent webhook deliveries don't serialize and the socket re-binds cleanly after restart. Matches the `_ReusingServer` shape in teams.py / webhook.py.
4. **No `OSError` handler on bind** — port-in-use crashed with a bare stack trace. Catch + log cleanly via `log.error("google_chat webhook bind failed", ...)` so the framework restarts with backoff instead.
5. **No `GOOGLE_CHAT_BIND_HOST` env var** — operators behind a reverse proxy couldn't lock to `127.0.0.1` (the way `TEAMS_BIND_HOST` / `WEBHOOK_BIND_HOST` let them). Added with default `0.0.0.0`, schema field marked `advanced=true`, `.strip() or DEFAULT` fallback so `BIND_HOST=""` matches sibling behaviour.

6 new tests pin the new contract:
- `test_default_bind_host_is_0_0_0_0`
- `test_bind_host_env_overrides_default`
- `test_bind_host_empty_string_falls_back_to_default`
- `test_bind_host_field_in_schema_marked_advanced`
- `test_shutdown_server_returns_immediately_without_blocking` — uses a fake `_httpd.shutdown()` that `time.sleep(0.5)`s; the test fails if the caller blocks (pins the asyncio-safe invariant).
- `test_shutdown_server_is_idempotent`

pytest **38/38** pass (was 32). No Rust delta — `cargo check` / `clippy` unchanged.

Final PR shape: 3 commits, 22 files, +1922 / −2915 = **−993 net**.

houko:
Third review pass — found 2 more real bugs. Fixed in `93cfd610`:

1. **`message_id` was stuffed into `metadata` instead of passed as the top-level `protocol.message(message_id=...)` kwarg.** Net effect: kernel saw `ChannelMessage.platform_message_id = ""` for every inbound Google Chat message, so any later reaction / edit helper couldn't target the real message — `react("👍")` would silently target the empty string and fail at the API layer with a confusing "message not found" instead of reacting to the user's actual message. Also dropped the `channel_label` metadata key while I was there — invented name with zero kernel consumers; channel identity comes from the sidecar entry's `channel_type = "google_chat"` declaration.

2. **`loop.call_soon_threadsafe(emit, event)` was unnecessary indirection.** The framework's `emit` is `threading.Lock`-guarded stdout write (sidecar/runtime.py:344-350), so calling it directly from the worker thread is safe AND ~one-loop-tick faster than scheduling on the asyncio loop. Matches teams.py / webhook.py shape.

2 new tests:
- `test_parse_webhook_event_plain_text` asserts `params["message_id"] == "spaces/AAAA/messages/M1"` AND `"channel_label" not in params["metadata"]`.
- `test_parse_webhook_event_omits_empty_message_id` guards that an empty `message.name` doesn't surface as `message_id=""` (which the kernel would store and later reaction helpers would target).

pytest **39/39** (was 38). No Rust delta.

Final shape: 4 commits, +1965 / −2938 = **−973 net**.

PR #5469, opened 2026-05-21
fix(runtime/pii_filter): redact WhatsApp JIDs atomically (no partial-redact via phone regex)

f-liva:
## Summary

`librefang_runtime::pii_filter`'s `phone` built-in pattern
(`(?:\+\d{1,3}[\s\-]?)?\(?\d{2,4}\)?[\s.\-]?\d{3,4}[\s.\-]?\d{3,4}`)
greedily consumed the leading digits of a WhatsApp JID, leaving the
trailing digits and the `@<domain>` suffix visible in the output:

```
input:   "Sender JID: 393511083257@lid arrived"
before:  "Sender JID: [REDACTED]257@lid arrived"   <- info leak + parse-breaking
after:   "Sender JID: [REDACTED] arrived"
```

This affects every WhatsApp JID shape: `<digits>@lid`,
`<digits>@s.whatsapp.net`, `<digits>@c.us`, `<digits>@g.us`,
`<digits>@broadcast`, `<digits>@newsletter`.

## Why the fix is a new pattern (not a `phone` tweak)

`regex_lite` (the crate used by `pii_filter` for safety / size) has no
lookahead support, so a negative `(?!@(?:lid|s\.whatsapp\.net|…))`
anchor on the `phone` pattern isn't available. The fix instead adds a
`whatsapp_jid` pattern as the FIRST entry of `BUILTIN_PATTERNS` so the
full JID is consumed as a single unit before `phone` ever sees it.

```rust
(
    "whatsapp_jid",
    r"\b\d{5,20}@(?:lid|s\.whatsapp\.net|c\.us|g\.us|broadcast|newsletter)\b",
),
```

In `Pseudonymize` mode the JID gets a stable `[Whatsapp_jid-A]` mapping
via the existing `get_or_create_pseudonym` machinery (same shape as
`[Email-A]` / `[User-A]`).

## Tests

Six new tests in `pii_filter::tests` pin the contract:

- `test_redact_whatsapp_lid_jid` — full LID redact (no digits / no suffix survive)
- `test_redact_whatsapp_phone_jid` — full `@s.whatsapp.net` JID redact
- `test_redact_whatsapp_group_jid` — full `@g.us` JID redact
- `test_jid_redact_is_atomic_not_partial` — regression for the original
  bug: assert no ASCII digits and no `@<domain>` survive
- `test_pseudonymize_whatsapp_jid_is_stable` — same JID -> same pseudonym
  across calls
- `test_jid_does_not_match_plain_email_or_number` — negative-control:
  plain emails still hit the `email` pattern, bare phone numbers still
  hit the `phone` pattern

## Test plan

- [x] `cargo test -p librefang-runtime pii_filter --lib` -> 22/22 passed
- [ ] CI green


houko (changes requested):
Thanks for this — the partial-redact-via-`phone` leak is real and the "atomic full-JID match before `phone`" approach is the right shape (no lookahead in `regex_lite`, so a leading dedicated pattern is the correct workaround). The fix is well-targeted for the dotless domains where the bug actually bites, and the pseudonym wiring is verified correct (`capitalize_category("whatsapp_jid")` → `[Whatsapp_jid-A]`, and ordering-first holds in both `redact()` and `pseudonymize()`). Requesting changes for two reasons: a merge conflict, and a residual-leak gap of the same class this PR is fixing.

## 1. Merge state — rebase needed (blocking the merge button)

GitHub reports this PR as `CONFLICTING` / `mergeStateStatus=DIRTY` (`has-conflicts` label). A local trial merge onto current `origin/main` resolved cleanly with only a `CHANGELOG.md` auto-merge, so the conflict is trivial — `main` has gained new `### Fixed` entries in `[Unreleased]` since your base. Please rebase on `origin/main` so the `[Unreleased]` block re-stacks and GitHub recomputes mergeability. No code conflict in `pii_filter.rs`.

## 2. Residual-leak gap: device-suffixed JIDs are still partial-redacted

`crates/librefang-runtime/src/pii_filter.rs:88-89`

```rust
("whatsapp_jid", r"\b\d{5,20}@(?:lid|s\.whatsapp\.net|c\.us|g\.us|broadcast|newsletter)\b"),
```

WhatsApp multi-device addressing emits JIDs of the form `<phone>:<device>@<domain>`, e.g. `15551234567:5@s.whatsapp.net` and `15551234567:5@lid` (baileys / whatsmeow both produce these). The pattern requires `\d{5,20}` *immediately* before `@`, but in a device-suffixed JID the characters before `@` are `:5` (one digit after a colon), so `whatsapp_jid` does **not** match. `phone` then greedily matches the leading `15551234567`, producing:

```
input:  "15551234567:5@s.whatsapp.net"
output: "[REDACTED]:5@s.whatsapp.net"
```

That is exactly the partial-redact leak (`@s.whatsapp.net` suffix + device id survive) that this PR exists to eliminate — just for a shape the regex misses. Suggest accommodating the optional device suffix before the `@`:

```rust
r"\b\d{5,20}(?::\d{1,3})?@(?:lid|s\.whatsapp\.net|c\.us|g\.us|broadcast|newsletter)\b"
```

Please also extend `test_jid_redact_is_atomic_not_partial` (or add a sibling) with a device-suffixed input (`...:5@s.whatsapp.net` and `...:5@lid`) and assert no `@<domain>` and no residual digits survive — the current atomic-redaction test uses only the non-suffixed shape, so it passes while the leak path stays open.

## 3. Minor — over-redaction edge on dotless anchors (note, non-blocking)

The dotless domains (`lid`, `broadcast`, `newsletter`) are anchored only by trailing `\b`. A contrived address like `12345@lid.example.com` would be clipped to `[REDACTED].example.com` (the `\b` after `lid` is satisfied by the `.`). Very low likelihood and not worse than the status quo; just flagging since it surfaced while reading the pattern. No change required unless you want to anchor with `(?:\b|$)` semantics, which `regex_lite` makes awkward.

## CI failure classification

The only red check is **Quality (rustfmt)**, and it is a **pre-existing base failure, not yours**: every file in the fmt diff (`librefang-api/src/routes/skills.rs`, `server.rs`, `kernel/agent_execution.rs`, `pooled_driver.rs`, `kernel-metering`, `kernel-router`, `memory/*`, …) is whole-file drift on `main` and untouched by this PR. `pii_filter.rs` does **not** appear in the fmt diff — your own code is fmt-clean. All PR-relevant lanes are green (Unit, Ubuntu shards 1–4, Security, OpenAPI Drift, Live Integration Smoke). `main` needs a separate restore-CI fix; a plain rebase will not clear Quality, so don't chase it here.

## Summary

- Approach correct; fix is valid for the dotless shapes it targets.
- Blocking: (a) rebase to clear the conflict; (b) cover device-suffixed `:NN@` JIDs — current code leaks them via `phone`, the same bug class.
- rustfmt failure is inherited base drift, not attributable to this change.

f-liva:
Fixed + rebased.

**Pattern**: added optional `(?::\d+)?` group before `@` so multi-device JIDs (`<phone>:<device>@<domain>` — baileys / whatsmeow both emit these) are matched atomically alongside the six domain forms:

```
\b\d{5,20}(?::\d+)?@(?:lid|s\.whatsapp\.net|c\.us|g\.us|broadcast|newsletter)\b
```

So `15551234567:5@s.whatsapp.net` now redacts as a single unit instead of `[REDACTED]:5@s.whatsapp.net`.

**Regression test**: `test_jid_redact_atomic_for_device_suffixed_jid` exercises `:N@lid`, `:N@s.whatsapp.net`, `:N@c.us` variants and asserts no digit or `:` separator survives.

**Rebase**: rebased onto current `origin/main`. Only conflict was an additive CHANGELOG interleave (`[Unreleased]/Fixed` grew on `main` after my base) — resolved keeping both sides.

`cargo test -p librefang-runtime --lib pii_filter` 26/26 pass.

houko (approved):
Re-reviewed the current HEAD — resolved, approving. Verified the ordering is the load-bearing part: `whatsapp_jid` is the first entry in `BUILTIN_PATTERNS`, and both `redact()` (`pii_filter.rs:231-237`) and `pseudonymize()` (`pii_filter.rs:240-254`) replace over a running string in array order, so the full JID is consumed before `phone`/`email` can partial-match it. Dotless domains can't trip `email` (no `.`); dotted ones are won by the JID-first ordering. The device-suffix `(?::\d+)?` group and `test_jid_redact_atomic_for_device_suffixed_jid` close the multi-device shape. Pseudonym category wiring (`[Whatsapp_jid-A]`) checks out.

The `<digits>@lid.example.com` over-clip edge (dotless domains anchored only by trailing `\b`) is low-likelihood and no worse than status quo — fine to leave or follow up. Once any rebase is settled and CI is green, good to merge.

houko (changes requested):
The JID-as-a-unit approach is right and the six tests are solid — thanks. One same-class gap to close before merge:

`crates/librefang-runtime/src/pii_filter.rs` — the `whatsapp_jid` pattern `\b\d{5,20}@(?:lid|s\.whatsapp\.net|c\.us|g\.us|broadcast|newsletter)\b` requires digits *immediately* before `@`. A multi-device JID carries a device suffix — `<phone>:<device>@s.whatsapp.net` (e.g. `14155552671:5@s.whatsapp.net`) — so the pattern misses it and `phone` still partial-redacts to `[REDACTED]:5@s.whatsapp.net`. That is the exact leak (tail + suffix surviving) this PR is closing, just for the multi-device shape.

Please extend the pattern to accept an optional device suffix, e.g. `\b\d{5,20}(?::\d+)?@(?:lid|s\.whatsapp\.net|c\.us|g\.us|broadcast|newsletter)\b`, and add a regression test for the `:<device>@` form. The current `test_jid_redact_is_atomic_not_partial` only exercises the non-suffixed shape, so it stays green while the suffix path leaks.

houko (approved):
Re-reviewing at HEAD `c25f35f0` ("fix(pii_filter): redact device-suffixed WhatsApp JIDs", committed 2026-05-22 18:03 UTC). My previous CHANGES_REQUESTED asked to cover device-suffixed JIDs — that work is already in this HEAD and is verified, so this supersedes the stale request. Approving.

## Device-suffix gap is closed
`crates/librefang-runtime/src/pii_filter.rs:95`:
```
r"\b\d{5,20}(?::\d+)?@(?:lid|s\.whatsapp\.net|c\.us|g\.us|broadcast|newsletter)\b"
```
The optional `(?::\d+)?` group matches the `<phone>:<device>@<domain>` multi-device shape (baileys / whatsmeow). `test_jid_redact_atomic_for_device_suffixed_jid` (pii_filter.rs:507) pins `15551234567:5@s.whatsapp.net`, `15551234567:5@lid`, `393511083257:12@s.whatsapp.net`, `393511083257:1@c.us` and asserts no `@<domain>`, no JID digit, and no `:` device separator survives.

## Correctness verified
- Ordering is load-bearing and holds: `whatsapp_jid` is the FIRST entry of `BUILTIN_PATTERNS` (pii_filter.rs:94), pushed first in `PiiFilter::new` (pii_filter.rs:130), and both `redact()` (pii_filter.rs:231-237) and `pseudonymize()` (pii_filter.rs:240-254) iterate `self.patterns` in array order — so the full JID is consumed before `phone`/`email` can partial-match it.
- Pseudonym wiring: `capitalize_category("whatsapp_jid")` (pii_filter.rs:311) → `[Whatsapp_jid-A]`, asserted stable across calls by `test_pseudonymize_whatsapp_jid_is_stable`.
- No regression: `test_jid_does_not_match_plain_email_or_number` confirms bare email still routes through `email` and a bare phone through `phone`.
- Ran the scoped suite against the real `regex` crate (not `regex_lite`): `cargo test -p librefang-runtime pii_filter --lib` → 26 passed, 0 failed (8 new JID tests inclusive).

## Merge state
- `mergeable: MERGEABLE`; a local trial-merge onto current `origin/main` produced no conflict markers. `pii_filter.rs` and `CHANGELOG.md` are byte-identical between the PR's merge-base and current main, so the earlier `DIRTY`/CHANGELOG concern no longer applies. The `BLOCKED` mergeStateStatus was solely the standing CHANGES_REQUESTED, which this approval clears.
- CI: all required lanes green (Unit, Ubuntu shards 1–4, Security, OpenAPI Drift, Live Integration Smoke, Quality).

## Non-blocking notes (no change required)
- The comment at pii_filter.rs:90 still says `regex_lite`; main migrated `pii_filter` to the full `regex` crate (Cargo.toml:48). The no-lookahead constraint the comment relies on is still true for `regex`, so the reasoning stands — just a stale crate name.
- Dotless-domain over-clip edge (`12345@lid.example.com` → `[REDACTED].example.com`) is low-likelihood and no worse than status quo (an email at that position is matched by `email` anyway).

Issue #5476, opened 2026-05-21
bug(docs/config): [agents.<name>.proactive_memory] block in config.toml silently ignored — real path is agent.toml


nevgenov:
## Summary

Issue #4870 description proposed per-agent override via a `[agents.<name>.proactive_memory]` block in `config.toml`:

```toml
[proactive_memory]
enabled = true
auto_memorize = true

[agents.lifeos-daily-brief-data-collector.proactive_memory]
auto_memorize = false
```

In beta.12 (`2026.5.17-beta.12`), this syntax in `config.toml` is **silently ignored** — the block parses (TOML is well-formed) but never feeds into any `AgentManifest.proactive_memory`. The override has no effect.

The actual working path is to put a `[proactive_memory]` block in the agent's **own** `agent.toml`, which deserializes into `AgentManifest.proactive_memory: ProactiveMemoryOverrides`.

## Repro

```bash
# config.toml
[proactive_memory]
enabled = true
auto_memorize = false  # global default off
extraction_model = "vllm/claude-haiku-4"

[agents.my-agent.proactive_memory]
auto_memorize = true   # should override for my-agent only
```

Send a turn to my-agent. Expected: `auto_memorize` extraction LLM fires. Observed: no extraction call in the daemon log, only the unconditional `[Past exchange]` row from `remember_interaction_best_effort` (which is a separate path — see #5474).

Moving the block to `~/.librefang/workspaces/agents/my-agent/agent.toml` as:
```toml
[proactive_memory]
auto_memorize = true
```
…works as documented.

## Impact

- Operators following the #4870 issue body verbatim get silent no-op.
- Worse: there's no warning log, no validation error — config.toml accepts the block (it's a valid TOML table) and the kernel just doesn't read it. Took us several hours of empirical testing to realize the override wasn't taking effect.

## Proposed fix (one of)

1. **Honor the syntax** — add a parser path that reads `[agents.<name>.proactive_memory]` from `config.toml` and merges into the manifest on agent load. Matches the original #4870 proposal exactly.
2. **Reject the syntax** — emit a WARN at boot if any `[agents.*.proactive_memory]` table is found in `config.toml`, pointing to the correct location.
3. **Doc-only fix** — update CHANGELOG, the #4870 description (closed), and the `ProactiveMemoryOverrides` rustdoc to clarify the override lives in `agent.toml`, not `config.toml`.

[...]

PR #5351, opened 2026-05-20
fix(runtime): reconcile cascade-leak THEMATIC_HEADERS with post-#5053 prompt builder

f-liva:
## Drift surfaced by #5344

PR #5344 added a drift-pin test (`silent_response::tests::thematic_and_scaffold_headers_match_prompt_builder_output`) that snapshots which `## <header>` strings the current prompt builder emits. The snapshot exposed that the five legacy strings the cascade-leak detector was calibrated against — `## Sender`, `## Today`, `## Calendar`, `## Tasks`, `## Response Style` — are **none of them** emitted by today's `prompt_builder.rs`.

Consequence: the `1 structural + 1 thematic` branch of `is_cascade_leak` is effectively dead, and the detector silently degraded to a `2+ structural` rule. The canonical chat-template regurgitation shape from today's builder — `## Persona\n…\n## Memory` — has zero thematic hits and does NOT trip on its own. Exactly the class of cross-chat leak the detector exists to catch.

## Reconciliation

| Bucket | Old | New |
|---|---|---|
| `SCAFFOLD_ONLY_HEADERS` (structural-equivalent — never in a legit reply) | 2 entries | 23 entries: 21 from current builder + the legacy 2 retained defensively |
| `THEMATIC_HEADERS` ambiguous-subset (only counts when paired with structural) | 3 entries (`## Today` / `## Calendar` / `## Tasks`) | 1 current (`## Current Date`) + the legacy 3 retained defensively |

### Classification rationale

Every header `prompt_builder.rs` emits today, with my judgment on legit-reply plausibility:

| Header | Bucket | Why |
|---|---|---|
| `## Persona`, `## Memory`, `## Skills`, `## Workspace`, `## Identity`, `## User Context`, `## User Profile`, `## Long-Term Memory`, `## Peer Agents`, `## Output Channels`, `## Channel`, `## Connected Tool Servers (MCP)`, `## Tool Call Behavior`, `## Active Goals`, `## Heartbeat Checklist`, `## Safety`, `## Operational Guidelines`, `## First-Run Protocol`, `## Live Context`, `## Your Tools` | SCAFFOLD-ONLY | Pure prompt-frame: a genuine reply never narrates `## Persona\n…` or `## Memory\n…` back to the user. |
| `## Current Date` | AMBIGUOUS (`THEMATIC` only) | Real "what's today?" replies may render `## Current Date\n…`. Only counts when paired with structural. |
[...]

PR #5311, opened 2026-05-20
feat(runtime): notify owner on exhausted LLM rate-limit instead of silent drop

f-liva:
## Why

When the underlying LLM provider hits a kernel-wide rate-limit and retries are exhausted, the agent loop currently logs a single `WARN` and drops the turn. The user who triggered the turn — typically an owner chatting with their agent on WhatsApp / Telegram — sees **nothing**: no acknowledgement, no error banner, no estimate of when the agent will come back.

The original incident (2026-05-20, chat `!CbDmKayZoOd53YJsAOAJ`, msg `355199`): an owner sent an image to Ambrogio while the shared OAuth-Max account was inside its rolling 5-hour cap. The daemon logged

```
WARN ... Claude Code CLI streaming subprocess exited with error
       exit_code=1 stderr=
```

and the chat stayed silent for 45 minutes until the quota rolled over at 11:40 UTC.

## What

Three layers:

### 1. Driver — `crates/librefang-llm-drivers/src/drivers/claude_code.rs`

- Parse the Claude CLI's `rate_limit_event` stream-json lines (capturing `status`, `resetsAt`, `rateLimitType`).
- On non-success CLI exit with a captured `resets_at`, surface a structured `LlmError::RateLimited` whose `message` carries machine-readable `resets_at_unix=<ts>` and `rate_limit_type=<kind>` markers.
- Existing text-pattern fallback preserved and enriched with the same header shape for uniform downstream parsing.

### 2. Config — `crates/librefang-types/src/config/types.rs`, `crates/librefang-types/src/agent.rs`

- New `[system]` block with `timezone` (IANA name, defaults to UTC; unparseable values fall back to UTC with a warn-once log).
- New `[rate_limit_notify]` block both as `KernelConfig` field (deployment-wide default) and as `AgentManifest` field (per-agent override). Resolution: per-agent > kernel > hardcoded fallback.
- Both new structs wired into the matching `Default` impls (the canonical CLAUDE.md gotcha guard).

### 3. Runtime — `crates/librefang-runtime/src/rate_limit_notify.rs`

- The agent loop wraps both `call_with_retry` (`agent_loop/mod.rs`) and `stream_with_retry` (`agent_loop/run_streaming.rs`); on `Err` it calls `dispatch_via_kernel`, which:
  - detects the `[rate_limit_defer_ms]` marker the retry helper appends on exhaustion,
  - pulls active config + timezone via two new `KernelHandle` getters (`rate_limit_notify_config()`, `system_timezone()`),
[...]

Issue #3328, opened 2026-04-27
Skill workshop: capture and refine reusable skills from successful agent interactions


houko:
## Problem

Today `librefang-skills` only loads static, author-time-defined skills (TOML manifest + Python / JS / Prompt body). There is no path for an agent to **learn a new skill from a successful interaction** — when a user spends 20 turns teaching the agent a multi-step workflow, that workflow is lost the moment the conversation ends. The next time the user wants the same thing, they teach it again.

Hands give us a packaging primitive (multi-agent, role-keyed, persisted overrides), but a hand is author-time too. There is no in-loop capture / refinement / promotion path.

## Reference: openclaw

`extensions/skill-workshop` captures repeatable workflows as workspace skills with:

- `autoCapture`: detect user corrections (`"no, do it like X"`) and reusable workflow instructions (`"from now on, always Y"`) after agent turns; surface as a candidate skill draft.
- `approvalPolicy`: store learned skill changes as `pending` (default) or write them automatically. Pending suggestions show up in a workshop UI for the user to approve / edit / reject.
- `reviewMode`: `heuristic` (cheap pattern match), `threshold-LLM` (call a cheap model to confirm), `both` (heuristic gate → LLM review), `none` (manual capture only).
- `maxPending`: cap on pending suggestions per workspace to avoid drift.
- Safe writes: every promoted skill goes through the same SHA256 / prompt-injection scan as marketplace skills before being loaded.
- Prompt refresh: when a skill is promoted, the system prompt is rebuilt so the agent picks it up next turn — no restart.

## Proposed approach

This is a **needs-rfc** size feature. Outline:

1. New crate-internal module `librefang-skills::workshop`:
   - `CandidateSkill` struct (draft TOML + extracted body + provenance: agent ID, session ID, turn range).
   - `CaptureSource` enum: `UserCorrection`, `ExplicitInstruction`, `RepeatedToolPattern`.

2. Capture pipeline runs `after_turn` (paired with the `before_prompt_build` hook from #3326):
   - Heuristic pass: regex / keyword detection for "always", "from now on", "remember to", "the way to do X is".
   - Optional LLM pass via the auxiliary client (#3314) to confirm and shape the candidate.

3. Storage: `~/.librefang/skills/pending/<workspace>/<uuid>.toml` with a `status = "pending"` field. The existing skill loader skips pending skills.

4. Approval surface:
   - Dashboard page listing pending skills, diff against existing skills, approve / reject / edit.
[...]

PR #5480, opened 2026-05-21
feat(runtime): owner-notify triage gate (pre-turn aux-LLM classification)

f-liva:
## Summary

Adds an **opt-in, pre-turn aux-LLM gate** that decides whether an inbound message from a non-owner sender warrants notifying the owner before the agent replies. Designed for receptionist / butler-style agents that field requests from strangers on behalf of an owner.

## Motivation

Today, an agent that wraps `notify_owner` in its prompt has no structured way to decide *when* to call it on inbound messages from non-owner senders. Either it always fires (noisy — owner gets pinged for every \"hello\"), or it never fires (silent — owner misses real escalations). Each call to the primary LLM also pays the full primary-tier cost for what is fundamentally a cheap classification decision.

This PR adds a dedicated cheap aux slot — `AuxTask::OwnerNotifyTriage` — and folds its verdict into the primary turn via the same `combined_prefix` channel the sender prefix already uses, so the primary LLM sees `## Owner Notification Triage (auxiliary)` as a hint and decides how to act on it.

## Architecture

```
Inbound message (sender != owner)
        |
        v
[ owner_notify_gate::evaluate_stranger_request ]
        |
        |--- aux failure / disabled --> TriageVerdict::skip() ----+
        |                                                          |
        v                                                          |
TriageVerdict { notify_owner, category, guidance, origin }         |
        |                                                          |
        v                                                          |
[ render_guidance_block ] --(None on skip)-------------------------+
        |
        v (when Some)
Prepend to combined_prefix
        |
        v
Primary LLM turn (sees the block alongside sender prefix)
        |
        v
Primary decides: call notify_owner? reply autonomously?
                 emit context-sensitive acknowledgement to sender?
```

The aux model is prompted to **pre-write** the sender-facing acknowledgement in the sender's language, so the primary LLM can either echo it verbatim or adapt it — no template strings hard-coded in Rust.

## New surface

| Path | What |
|------|------|
| `librefang_types::config::AuxTask::OwnerNotifyTriage` | New variant; slug `owner_notify_triage`; routes through `[llm.auxiliary]` like every other side task |
| `librefang_types::agent::OwnerNotifyGateConfig { enabled, owner_user_ids }` | Per-agent block stamped onto `AgentManifest`; **default-OFF** — opt-in per agent |
| `librefang_runtime::owner_notify_gate` | New module: `TriageVerdict`, `TriageOrigin`, `evaluate_stranger_request`, `parse_verdict`, `render_guidance_block` |

### Operator opt-in

```toml
# agent.toml
[owner_notify_gate]
enabled = true
owner_user_ids = ["393xxxxxxxxx@s.whatsapp.net", "393xxxxxxxxx@lid"]

# config.toml
[llm.auxiliary]
owner_notify_triage = [\"anthropic:claude-haiku-4.5\"]
```

The gate inherits the agent's primary fallback chain when no aux entry is configured — feature works at primary-tier cost out of the box.

## Default-safe failure semantics

Every failure path collapses to `TriageVerdict::skip(...)`:

- empty inbound message
- aux LLM call error
- JSON parse failure
- no JSON-shaped object in response
- partial response (missing fields use safe defaults; missing `notify_owner` defaults to `false`)
- non-bool `notify_owner` field

`TriageVerdict::skip` yields `notify_owner = false` AND `render_guidance_block` returns `None` so **no block is injected**. When the gate is disabled or skipped, the pre-#5471 behaviour is preserved bit-for-bit.

## Tests landed

| Where | Count | Coverage |
|-------|-------|----------|
| `librefang_types::config::tests` | 5 | `AuxTask::OwnerNotifyTriage` slug stability, serde round-trip, `chain_for` lookup, TOML decode under `[llm.auxiliary]`, `BTreeMap` distinctness vs `SessionSummary` |
| `librefang_types::agent::tests` | 8 | `OwnerNotifyGateConfig::should_evaluate` — default-disabled, enabled-skips-owner, enabled-evaluates-non-owner, missing / empty sender, byte-exact case-sensitive match, TOML round-trip, empty-TOML default |
| `librefang_runtime::owner_notify_gate::tests` | 17 | `TriageVerdict::skip` shape; `parse_verdict` over clear-yes / clear-no / fenced-json / bare fence / prose-wrapped / missing-category / empty-category / missing-guidance / missing-`notify_owner` / invalid JSON / no-braces / empty / non-bool `notify_owner`; `evaluate_stranger_request` (async, with mock aux) over empty-message / notify-yes / notify-no / garbage response / missing display-name; `render_guidance_block` over skip / aux-yes / aux-no / empty-guidance-placeholder |

## Test plan

- [x] `cargo test -p librefang-types aux_task_owner_notify_triage --lib` → 5 / 5
- [x] `cargo test -p librefang-types owner_notify_gate --lib` → 8 / 8
- [x] `cargo test -p librefang-runtime owner_notify_gate --lib` → 17 / 17
- [x] `cargo test -p librefang-runtime agent_loop --lib` → 214 / 214 (no regression)
- [x] `cargo clippy -p librefang-types --lib --no-deps -- -D warnings` clean
- [x] `cargo clippy -p librefang-runtime --lib --no-deps -- -D warnings` clean
- [x] `cargo check --workspace --lib` clean
- [ ] CI green


houko (changes requested):
Compile-breaking syntax error in the new test module — this PR will not build, so the claimed `cargo test -p librefang-types ... 5/5` could not have run on the final state.

`crates/librefang-types/src/agent.rs:3628` — the test `owner_notify_gate_missing_fields_use_defaults` is missing its closing brace. The body ends at:

```rust
    fn owner_notify_gate_missing_fields_use_defaults() {
        let cfg: OwnerNotifyGateConfig = toml::from_str("").unwrap();
        assert!(!cfg.enabled);
        assert!(cfg.owner_user_ids.is_empty());
    /// Audit: session-mode-deserialize-fallback. Pin serde's   <-- no `}` before this
```

There is no `}` before the next `/// Audit: ...` doc comment / `#[test] fn session_mode_deserializes_...`. This unbalances the braces in the `tests` module and breaks `cargo test -p librefang-types` (and the CI unit-fast lane). Add the missing `}` after the last `assert!`.

The rest of the change reviews cleanly: `AuxClient::with_primary_only` / `resolve` / `AuxResolution.{driver,resolved}`, `MockLlmDriver::with_response`, and the `sender_user_id` binding in `agent_loop::run_agent_loop` all exist and are used correctly; the `combined_prefix` shadowing and default-safe `TriageVerdict::skip` semantics are sound. Just the one missing brace blocks the build — please fix and re-push.

houko (changes requested):
## Review — owner-notify triage gate

Thanks for the thorough write-up and the test matrix. The architecture (dedicated `AuxTask` slot, fold verdict into `combined_prefix`, fail to skip) is the right shape and the config wiring is correct. However the branch does not compile, so this is a hard request-changes, plus two design points to resolve.

### Blocking: the branch does not build (REAL failure, not base drift)

Every red check — Quality, Unit (lib+bin), all 4 Ubuntu shards, OpenAPI Drift, Security, Live Integration Smoke, Kernel Ignored/Restart — has the **same single root cause**:

```
error: this file contains an unclosed delimiter
  --> crates/librefang-types/src/agent.rs:3699:3
2043 | mod tests {            <- unclosed delimiter
3628 | fn owner_notify_gate_missing_fields_use_defaults() {   <- this delimiter might not be properly closed
```

At `crates/librefang-types/src/agent.rs:3631`, `owner_notify_gate_missing_fields_use_defaults()` is missing its closing `}`:

```rust
fn owner_notify_gate_missing_fields_use_defaults() {
    let cfg: OwnerNotifyGateConfig = toml::from_str("").unwrap();
    assert!(!cfg.enabled);
    assert!(cfg.owner_user_ids.is_empty());
// <-- missing `}` here; a `///` doc-comment for the NEXT test starts immediately
/// Audit: session-mode-deserialize-fallback. ...
#[test]
fn session_mode_deserializes_lowercase_persistent_and_new() {
```

`librefang-types` is a leaf dependency of the whole workspace, so this one missing brace fails the lib build and cascades into every downstream job. There is **no base-vs-real ambiguity here** — none of the known main-branch flakes (whole-file rustfmt drift, `comms_send_rejects_oversize`, `config_schema_overlay`, llvm-cov opentelemetry) appear in these logs; the reds are 100% PR-introduced. Add the missing `}` and re-push.

Note: the test plan claims `cargo check --workspace --lib` clean and `cargo test -p librefang-types ... 8/8` — that cannot be true against the pushed tree. Please re-run against the actual HEAD before re-requesting review.

### Design 1 — financial-DoS exposure when no cheap aux tier is configured

`evaluate_stranger_request` (`crates/librefang-runtime/src/owner_notify_gate.rs:138`) resolves through `AuxClient::resolve(AuxTask::OwnerNotifyTriage)`, and per `crates/librefang-runtime/src/aux_client.rs:130` the "absent = inherit" rule means that **when the operator has not configured `[llm.auxiliary] owner_notify_triage`, the gate runs on the primary driver**. The PR body confirms this: "feature works at primary-tier cost out of the box."

The difference from the other aux tasks (compression, session-title) is the **trigger**: this gate fires on *every inbound message from a non-owner sender* (`crates/librefang-runtime/src/agent_loop/mod.rs:567`), i.e. an attacker-controlled path on a public receptionist channel. An external party spamming the agent forces one primary-tier LLM call per message when no cheap tier is set — the exact financial-DoS vector the skill-workshop pattern guards against by returning `Indeterminate` when no cheap-tier credentials are present (see CLAUDE.md, skill-workshop section). Recommend mirroring that posture: when `OwnerNotifyTriage` resolves to the primary driver (empty `resolution.resolved`, `owner_notify_gate.rs:143-148`), skip the gate (`TriageVerdict::skip`) and log a one-time WARN, rather than silently billing primary-tier on an attacker-driven path. At minimum the docstring at `owner_notify_gate.rs:26-27` ("inherits the primary driver and still works, just at primary-tier cost") should be reframed as a warning, not a feature.

### Design 2 — justify the fail-safe direction explicitly

Failure paths collapse to `notify_owner = false` and `render_guidance_block` returns `None`. For preserving pre-#5471 behaviour this is reasonable. But the prompt's own YES-criteria include emergencies, scam/impersonation, and time-sensitive escalations (`owner_notify_gate.rs:54-58`). On aux failure those are silently *not* surfaced to the owner — the agent just replies autonomously. That is a defensible "no behaviour change, no noise" choice, but please state it as a conscious decision in the module docs (the docstring currently frames skip purely as a clean-prompt optimisation), so a future reader knows escalation-misses on aux outage are intentional.

### Minor / coordination

- Double-notify with sibling #5311: no functional collision — #5311 is *reactive* (owner ping on exhausted rate-limit, post/mid-turn) while this is *pre-turn proactive*. They notify on disjoint conditions. Worth a one-line mention in the PR body so a reader knows the two `notify_owner` paths were considered together.
- Determinism: good — `AuxiliaryConfig.tasks` is `BTreeMap<AuxTask, Vec<String>>` (`config/types.rs:1202`) and `render_guidance_block` is plain `format!` with no map iteration.
- Config hygiene: good — `OwnerNotifyGateConfig` has `#[serde(default)]` on field + struct, derives `Default`, and the `wizard.rs` `AgentManifest` literal is updated (`crates/librefang-kernel/src/wizard.rs:208`). This is an `agent.toml` block, not a `KernelConfig` field, so no `config_schema_overlay` descriptor is required — the #5311 overlay gap does not apply here.

Please fix the compile error, address Design 1, and document Design 2. Happy to re-review once CI is green.

f-liva:
Closed the brace on `owner_notify_gate_missing_fields_use_defaults` (`crates/librefang-types/src/agent.rs:3633`). The missing `}` was the single root cause of every red lane.

Verified locally on the new HEAD:
- `cargo check --workspace --lib` clean
- `cargo test -p librefang-types owner_notify_gate` → 8/8 pass

You were right about the test-plan claim — it was made on a tree that didn't actually compile, my mistake. Re-pushed and awaiting CI.

houko (changes requested):
Re-reviewed after your latest push. The architecture is the right shape — dedicated `AuxTask` slot, verdict folded into `combined_prefix`, fail-to-skip — and the config wiring, determinism, and call-site scoping all check out. The earlier compile-break (missing brace) is fixed. Two items block, one nit.

**1. (blocking) Financial-DoS: the gate bills the primary provider on an attacker-controlled path when no cheap aux tier is configured.** `evaluate_stranger_request` (`crates/librefang-runtime/src/owner_notify_gate.rs:260-269`) resolves `AuxTask::OwnerNotifyTriage` and calls `driver.complete()` without checking `resolution.used_primary`. When `[llm.auxiliary] owner_notify_triage` is unset, `AuxClient::resolve` returns the primary driver with `used_primary = true` (`crates/librefang-runtime/src/aux_client.rs:145-160`). Because the gate fires on every inbound message from a non-owner sender (`crates/librefang-runtime/src/agent_loop/mod.rs:553-600`) — an attacker-controllable path on a public receptionist agent — an external party can force one primary-tier LLM call per spammed message. This is the exact vector the skill-workshop pattern already guards: `crates/librefang-kernel/src/skill_workshop/mod.rs:454-463` returns `Indeterminate` when `resolution.used_primary` with the rationale "would burn user budget on every turn — skip." Please mirror it: on `used_primary`, return `TriageVerdict::skip(...)` and log a one-time WARN. The docstring at `owner_notify_gate.rs:26-27` ("inherits the primary driver and still works, just at primary-tier cost") should become a warning rather than a documented feature.

**2. Document the fail-safe direction.** Failure paths collapse to `notify_owner = false` / no guidance block. Given the YES-criteria include emergencies and scam/impersonation, an aux outage silently suppresses those escalations. That's a reasonable "no behaviour change" choice, but the module docs frame skip only as a prompt-cleanliness optimisation — please state the escalation-miss-on-outage trade-off as a conscious decision.

**3. (nit) Add a loop-level test.** The module tests are thorough, but nothing exercises the fold into `combined_prefix` at the `agent_loop/mod.rs:553-600` injection site. An integration test there would match the repo's "test at the injection site, not just the implementation site" guidance — and would have caught the primary-billing path above.

houko:
Status note (no new review — the standing change request still stands).

The only thing pushed since the last review (commit `de1e873`) is the `main` merge `04602b8`. The feature files are byte-for-byte unchanged across that range:

```
git diff de1e873 04602b8 -- \
  crates/librefang-runtime/src/owner_notify_gate.rs \
  crates/librefang-runtime/src/aux_client.rs \
  crates/librefang-runtime/src/agent_loop/mod.rs
# (empty)
```

So none of the three items from the previous review are addressed yet:

1. **(blocking) Financial-DoS on the attacker-controlled path.** `evaluate_stranger_request` (`crates/librefang-runtime/src/owner_notify_gate.rs:138-164`) still resolves `AuxTask::OwnerNotifyTriage` and calls `driver.complete()` without consulting `resolution.used_primary` — which `AuxResolution` already exposes (`crates/librefang-runtime/src/aux_client.rs:284`, set `true` at `:150/:160/:195` when the chain falls back to the primary). When `[llm.auxiliary] owner_notify_triage` is unset the gate bills the primary provider on every inbound non-owner message. Mirror the skill-workshop guard: on `used_primary`, return `TriageVerdict::skip(...)` + one-time WARN.
2. Module docs still frame the primary-tier fallback as a feature (`owner_notify_gate.rs:26-27`) and don't state the fail-safe (aux-outage → escalations silently suppressed) as a conscious decision.
3. No loop-level test at the `combined_prefix` injection site (`crates/librefang-runtime/src/agent_loop/mod.rs`).

On CI: the current reds are **base/toolchain drift dragged in by the merge**, not the feature.
- Quality (clippy) fails on `clippy::manual_pattern_char_comparison` in `crates/librefang-api/{routes/registry.rs,validation.rs,webchat.rs}` — pre-existing main code under a newer clippy (rust 1.95.0), none of it touched by this PR.
- Unit (lib+bin) fails on the single test `librefang-kernel config_reload::tests::test_external_auth_issuer_url_change_evicts_oauth_caches` — main-branch test, unrelated to this change.

The earlier compile break is genuinely fixed. The gate stays request-changes pending items 1–3; CI also needs a rebase on a green main once those land.

houko (changes requested):
Two issues need fixing before merge.

**1. Missing `used_primary` guard — financial-DoS on the critical turn path (blocker)**

`evaluate_stranger_request` calls `aux_client.resolve(AuxTask::OwnerNotifyTriage)` but never checks `resolution.used_primary`. When the operator has not configured `[llm.auxiliary] owner_notify_triage = [...]`, `AuxClient::resolve` returns the agent's primary fallback driver with `used_primary = true`. Every stranger-originated message then fires an extra primary-tier LLM call before the real turn — this is precisely the financial-DoS pattern `SkillWorkshopReview` was explicitly designed to prevent.

Compare `crates/librefang-kernel/src/skill_workshop/mod.rs:453-463`:
```rust
let resolution = aux.resolve(AuxTask::SkillWorkshopReview);
if resolution.used_primary {
    return ReviewDecision::Indeterminate { ... };
}
```

The fix is the same pattern: thread `AuxResolution` out of `evaluate_stranger_request` (or have it return `TriageVerdict::skip("no cheap-tier aux configured")` when `resolution.used_primary`) so the gate is a true no-op for operators who have not opted in at the infra level.

The PR description says "the gate inherits the agent's primary fallback chain when unconfigured — feature works at primary-tier cost out of the box", but that is the wrong default for a gate that fires on EVERY inbound stranger turn. The SkillWorkshop precedent is authoritative here.

**2. System prompt sent twice in `CompletionRequest` (bug)**

`owner_notify_gate.rs:154-161`:
```rust
messages: Arc::new(vec![
    Message::system(TRIAGE_SYSTEM_PROMPT),   // ← in messages array
    Message::user(user_payload),
]),
system: Some(TRIAGE_SYSTEM_PROMPT.to_string()),  // ← also in system field
```

Every other aux caller in this repo (compactor, context_compressor, proactive_memory, history_fold) uses `system: Some(...)` only and does NOT add `Message::system(...)` to the messages slice. Sending the system prompt in both slots doubles input tokens billed and may cause provider errors on APIs that reject system messages in the messages array when a top-level `system` field is also present.

**CI status note**: Quality/clippy fails are the pre-existing `boot.rs::map_identity` and `workspace_setup.rs::io_other_error` lints in kernel test code, not introduced by this PR. The "Test / Unit" job was canceled by a runner infrastructure timeout during apt-get, not a code failure — the 4 Ubuntu integration shards all passed.

houko (changes requested):
Reviewed at 847b0946. `cargo check --workspace --lib` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test -p librefang-runtime --lib owner_notify_gate` 23/23, `cargo test -p librefang-types owner_notify_gate` 8/8. The wiring, parser, fence/prose tolerance, and `should_evaluate` taxonomy are nicely engineered and well covered. Two blocking concerns plus rate-limit follow-up before merge.

## Blocking

### 1. Financial-DoS regression: gate bills the operator's primary chain when no aux is configured
`crates/librefang-runtime/src/owner_notify_gate.rs:260-269` resolves `AuxTask::OwnerNotifyTriage` and unconditionally runs the call, even when `resolution.used_primary == true`. The PR body explicitly endorses this ("feature works at primary-tier cost out of the box"). This contradicts the precedent landed for `AuxTask::SkillWorkshopReview` in #3328 — see `crates/librefang-kernel/src/skill_workshop/mod.rs:454-462`, which checks `resolution.used_primary` and returns `Indeterminate` precisely to avoid "burn[ing] user budget on every turn". CLAUDE.md calls out this regression by name: *"the workshop returns `Indeterminate` rather than billing the operator's primary provider, blocking a financial-DoS regression"*.

For this gate the impact is worse than skill workshop because it fires on **every inbound stranger message**, not on positive heuristic hits. An agent receiving routine traffic from non-owners pays primary-tier for every "hello" the moment an operator enables `[owner_notify_gate]` without first wiring `[llm.auxiliary] owner_notify_triage`.

Fix: in `evaluate_stranger_request`, if `resolution.used_primary` is true, return `TriageVerdict::skip("no aux chain configured for owner_notify_triage")` and emit a one-shot operator WARN with a pointer at the `[llm.auxiliary]` slug — matching skill workshop's posture. Documentation should also be updated to state that the gate is a no-op until a cheap-tier slot is wired.

### 2. No per-agent / per-sender rate limit on the triage path
`OwnerNotifyGateConfig` (`crates/librefang-types/src/agent.rs:717-748`) exposes only `enabled` + `owner_user_ids`. Every non-owner message synthesizes one aux call — there is no dedup window per `sender_user_id`, no per-agent cap, no per-owner spend ceiling. A chatty stranger or a spam burst translates 1:1 into aux calls. The trigger-dispatch concurrency caps documented in `docs/architecture/trigger-dispatch-concurrency.md` cap *primary* turns, not this aux side call.

Combined with finding 1, that's an unbounded amplifier. At minimum: add a short per-`(agent, sender_user_id)` verdict cache (e.g. 60-300s TTL) so repeat strangers don't re-triage on every message, and document the per-call cost shape in the operator docs.

## Should fix

### 3. `temperature = 0.0` + `..Default::default()` may regress against drivers that read both fields
`crates/librefang-runtime/src/owner_notify_gate.rs:280-284` sets `max_tokens = 512`, `temperature = 0.0`, **and** `system: Some(TRIAGE_SYSTEM_PROMPT.to_string())` while also stuffing the prompt as `Message::system(...)` in `messages[0]`. Some providers (notably the Anthropic driver) treat the dedicated `system` field as the canonical system prompt and ignore the leading `Message::system` — others concatenate. The current shape is non-deterministic across the provider matrix the kernel can route to. Pick one channel (either the dedicated `system` field OR `Message::system` in `messages`, not both) so the prompt the aux model sees is byte-identical regardless of which provider got picked. The deterministic-prompt-ordering rule in CLAUDE.md (#3298) also applies here — prompt-cache hit rates degrade if the same logical call shapes differently across runs.

### 4. PR description claims "default-off matching skill_workshop"; behaviour matches `enabled=false` but the primary-fallback semantics break the financial story
The PR positions the gate as "opt-in per agent, default-OFF" — that part is genuine (`OwnerNotifyGateConfig::default()` has `enabled: false`, verified by `owner_notify_gate_defaults_disabled_with_empty_owner_list`). But the *opt-in* path has the cost trap from finding 1: an operator enabling the gate per the docs without also wiring `[llm.auxiliary] owner_notify_triage` discovers the bill later. The docstring on `OwnerNotifyGateConfig` and the module-level docs on `owner_notify_gate.rs:135-149` should hard-warn (not just observe) that primary-tier inheritance is a DoS amplifier and operators must wire a cheap-tier slot before flipping `enabled = true`.

## Nits

- `crates/librefang-runtime/src/owner_notify_gate.rs:271` — `sender_display_name.unwrap_or("Unknown sender")` leaks the literal string `"Unknown sender"` into the aux prompt. If the sender truly is unknown, sending a deterministic placeholder is fine, but consider omitting the line entirely so the aux model isn't biased by a synthetic name.
- `crates/librefang-runtime/src/agent_loop/mod.rs:553-595` — the new gate block re-binds `combined_prefix` inside the `if`/`else`. The two `else` arms (no `should_evaluate`, no `aux_client`) just return `combined_prefix` unchanged — same as no-op. Consider an early-`if-let-chain` so the happy path stays shallow and the lint footprint shrinks.

## Rebase ask

`Mergeable: CONFLICTING` per `gh pr view`. After addressing the two blocking items, please rebase onto latest `origin/main` and force-push; I will re-review and merge with `--auto --squash` once the rebase is clean and CI is green.

f-liva:
Rebased onto upstream/main and addressed all review feedback from the 2026-05-25 review at 847b0946.

### Blockers resolved

**1. Financial-DoS guard (used_primary)**
`evaluate_stranger_request` now checks `resolution.used_primary` after `aux_client.resolve(AuxTask::OwnerNotifyTriage)`. When the resolution fell back to the primary driver (no `[llm.auxiliary] owner_notify_triage` configured), returns `TriageVerdict::skip("no aux chain configured for owner_notify_triage")` without calling the LLM. A one-shot `WARN` fires once per process pointing the operator at the `[llm.auxiliary]` slug. Mirrors the `SkillWorkshopReview` precedent in `skill_workshop/mod.rs:454-462`.

**2. Per-(agent, sender) verdict cache**
Added a process-wide `OnceLock<Mutex<VecDeque<CacheEntry>>>` with `DEFAULT_CACHE_TTL_SECS = 120` and `MAX_CACHE_ENTRIES = 256`. On cache hit within TTL, the cached verdict is returned with `TriageOrigin::CachedVerdict` (produces a guidance block, skips the aux call). Fresh `AuxModel` verdicts are inserted after a successful aux call. 4 cache-specific tests (hit, miss-by-agent, miss-by-sender, eviction cap).

### Should-fixes resolved

**3. Dual system prompt removed**
`CompletionRequest` now uses only the dedicated `system: Some(TRIAGE_SYSTEM_PROMPT)` field. `Message::system(TRIAGE_SYSTEM_PROMPT)` removed from `messages[0]`, matching every other aux caller in the repo (compactor, context_compressor, proactive_memory, history_fold).

**4. Doc update**
`OwnerNotifyGateConfig` docstring now has a `# Financial safety` section that hard-warns about the DoS amplifier and states operators must wire a cheap-tier slot before enabling. Module-level docs updated with three new sections: financial safety, escalation-miss-on-outage trade-off, and per-sender verdict caching.

### Nits resolved

- `sender_display_name`: when `None`, the "Sender display name:" line is omitted entirely from the aux prompt instead of injecting "Unknown sender".
- agent_loop call site: collapsed nested `if` / `if let` into a single `if let` tuple match to reduce nesting.

### Verification

- `cargo check --workspace --lib` clean
- `cargo clippy -p librefang-runtime --all-targets -- -D warnings` clean
- `cargo clippy -p librefang-types --all-targets -- -D warnings` clean
- `cargo test -p librefang-runtime --lib owner_notify_gate` 26/26
- `cargo test -p librefang-types --lib owner_notify_gate` 8/8
- `cargo test -p librefang-runtime --lib agent_loop` 236/236

Rebased onto current upstream/main; brace-fix commit absorbed into the main commit (clean 2-commit history: original feat + this fix).

houko (changes requested):
**Requesting changes.** The financial-DoS `used_primary` guard, the double-system-prompt fix, and the bounded TTL verdict cache all look correct on the current HEAD. But the `used_primary` fix silently gutted the end-to-end test coverage the PR body claims:

1. `crates/librefang-runtime/src/owner_notify_gate.rs:457-460` — the test helper `aux_with` builds `AuxClient::with_primary_only(driver)`, which uses `AuxiliaryConfig::empty()` so `resolve` always returns `used_primary = true` (`aux_client.rs:101-107`, `:160`). The new financial-DoS guard (`owner_notify_gate.rs:291`) then short-circuits with `skip("no aux chain configured")` *before* `driver.complete` on every test that uses `aux_with`. So the full happy path (`resolve → driver.complete → parse_verdict → insert_cached_verdict`) and the cache-hit short-circuit are never exercised through the public `evaluate_stranger_request` — only the standalone `parse_verdict` unit tests remain. Add a test seam so `evaluate_stranger_request` can run against a mock aux with `used_primary = false` (inject the resolved driver, or accept an `AuxResolution`), then re-add a notify-yes and notify-no case that reaches the parser and asserts cache insertion.
2. `crates/librefang-runtime/src/owner_notify_gate.rs:659-665` — `evaluate_returns_skip_when_aux_response_is_garbage` claims to test parse-failure handling, but because `aux_with` ⇒ `used_primary = true`, it returns via the financial-DoS guard and never reaches `parse_verdict`. It passes for the wrong reason. Assert on the `origin` reason (`DefaultedNoNotify { reason: "aux response not JSON" }`) and drive it through a non-primary aux so it fails if the parser isn't actually exercised.

Non-blocking: the new `## Owner Notification Triage (auxiliary)` header injected into `combined_prefix` is adjacent to #5351's cascade-leak header sets — not a conflict today, but worth sequencing awareness. (Mergeability: GitHub shows CONFLICTING but `git merge-tree` resolves cleanly — just needs a rebase to clear the flag.)

houko:
Thanks — the financial-DoS guard is correct and matches the `SkillWorkshopReview` precedent: `evaluate_stranger_request` skips with a one-shot WARN when `resolution.used_primary` (no cheap-tier aux configured), and skip-path verdicts aren't cached so an aux outage doesn't poison the cache. `parse_verdict` is robust, and `should_evaluate` gates on enabled + non-owner. Good test coverage (39 tests).

Blockers / nits:

1. **CONFLICTING** — append-hotspot conflicts (`AuxTask` enum, `AgentManifest` field, CHANGELOG, Cargo.lock, `lib.rs`, `wizard.rs`); no logic conflict, needs a rebase.
2. **`Test / Ubuntu (shard 4/4)` fails** (the other checks pass) — please surface the failing-test detail after rebase.
3. **Doc/impl mismatch:** the module doc says guidance is folded into the primary turn's *system prompt*, but the call site (`agent_loop/mod.rs:75-105`) prepends it to `effective_user_message` (user role). The user-role placement is the safer choice — just relabel the doc.
4. **Low-severity injection surface:** there's no length clamp on aux-produced `guidance`/`category` before it's injected into the prefix (bounded only by `max_tokens: 512`). A non-owner controls the inbound message the aux echoes into `guidance` — worth a clamp.

houko:
Thanks @f-liva — the design looks good; it's blocked in CI by one missing site.

`owner_notify_gate` isn't added to the exhaustive destructure in `crates/librefang-types/src/config/mod.rs` (~line 1639), so the #5476 guard fails with `error[E0027]` across all 6 lanes. Add the field to the non-override arm.

Two habits that catch this whole class pre-push:
1. When adding a struct field, grep every construction + destructure site (and the drift-guard tests) and update them in the same change.
2. Run the full unit lane locally rather than `-p`-scoped: `cargo nextest run --workspace -E 'kind(lib) | kind(bin)'`. The scoped run stays green while the cross-crate guard is red.

Minor: the body says "configurable TTL" but it's hardcoded `DEFAULT_CACHE_TTL_SECS` — either expose the knob or adjust the wording.

houko:
Heads-up: the latest merge from `main` (06-01) left the branch not compiling:

```
error[E0027]: pattern does not mention field `owner_notify_gate`
```

A `match` / destructuring pattern on this branch needs to add the new `owner_notify_gate` field (or a trailing `..`). Please re-push once it builds so the review can land on a green branch.

CI: https://github.com/librefang/librefang/actions/runs/26751044033

houko:
Three issues on the owner-notify triage path — two of them security-relevant.

**1. Triage verdict is cached without message content → stale/abusable verdict.** The cache key is `(agent_id, sender_user_id)` only:

```
crates/librefang-runtime/src/owner_notify_gate.rs:176   agent_id: String,
crates/librefang-runtime/src/owner_notify_gate.rs:177   sender_user_id: String,
crates/librefang-runtime/src/owner_notify_gate.rs:210   .find(|e| e.agent_id == agent_id && e.sender_user_id == sender_user_id)
crates/librefang-runtime/src/owner_notify_gate.rs:276   if let Some(cached) = lookup_cached_verdict(agent_id, sender_user_id, ttl) {
```

Within the 120s TTL the first verdict is reused for every later, substantively different message from the same sender. A stranger's `"ciao"` caches `notify_owner=false`; their next message `"my father collapsed, I need the owner's address now"` hits the cache before any aux request is built and reuses the stale no-notify verdict, suppressing a genuine escalation. It also cuts the attacker's way: prime the cache with a benign message, then send the real request and be guaranteed the no-notify guidance for the window.
*Fix:* fold a content discriminator into the key (e.g. hash of the trimmed message), or only short-circuit when the new message hashes equal to the cached one.

**2. Triage bypasses `injection_guard` on both ends.** Raw attacker text feeds the aux prompt unscanned (owner_notify_gate.rs:314), and the aux-derived guidance block is injected into the *primary* prompt unscanned:

```
crates/librefang-runtime/src/agent_loop/mod.rs:628   Some(prefix) => format!("{prefix}{block}\n\n"),
crates/librefang-runtime/src/agent_loop/mod.rs:637   let effective_user_message = match &combined_prefix {
crates/librefang-runtime/src/agent_loop/mod.rs:648   if let Some(warning) = crate::injection_guard::scan_message(user_message) {
crates/librefang-runtime/src/agent_loop/mod.rs:697       &effective_user_message,
```

`scan_message` only ever scans the raw `user_message`, never `combined_prefix`/`effective_user_message` — which is the string actually sent into `prepare_llm_messages` at line 697. Both inbound messages (`scan_message`) and tool results (`scan_tool_result` → `SkillVerifier`) normally cross this boundary; the triage block does not. A stranger can craft a message that steers the aux model to echo injected instructions into the "suggested acknowledgement", which then lands in an elevated, system-prompt-adjacent position with no warning prefix.
*Fix:* run `injection_guard::scan_message` on `effective_user_message` (the string sent at line 697) instead of only the raw `user_message`, so the triage block and combined prefix are covered.

**3. Streaming path skips the gate entirely.** `crates/librefang-runtime/src/agent_loop/run_streaming.rs` has zero references to the triage gate, so agents driven through the streaming path get no owner-notify triage at all — a behavioral divergence the PR body doesn't mention.

The default-OFF posture and financial-DoS guard are implemented correctly; issues 1 and 2 fire exactly on the documented opt-in happy path (gate enabled + cheap aux slot wired).

_Reviewed at 067de27c9._

houko (commented):

**Mid-sentence line wraps in `TRIAGE_SYSTEM_PROMPT` constant**

CLAUDE.md (Prose wrapping section): *"The only legitimate line break inside a paragraph is at a complete sentence boundary. This rule applies universally: … Source code doc-comments … and any multi-line free-form prose comment block."*

The `TRIAGE_SYSTEM_PROMPT` string literal in `crates/librefang-runtime/src/owner_notify_gate.rs` wraps every line at ~80 chars, breaking mid-sentence throughout. Example:

```rust
const TRIAGE_SYSTEM_PROMPT: &str = "You are a triage assistant for a personal-assistant agent. \
The agent's owner has set boundaries on when the agent may reply autonomously vs. when the \
owner must be looped in. You receive a single inbound message from a NON-OWNER sender and \
you decide whether the agent should notify the owner before answering.
```

Each sentence should be on one line; the `\` continuation characters and 80-char breaks should be removed so every complete thought runs to its terminal punctuation before a newline.

No other issues found; the gate logic, verdict caching, guidance-block rendering, config wiring, and test coverage all look correct.

---

houko:
Closing as stale.

This PR has gone roughly a month without author movement, is currently in a BLOCKED merge state, and isn't tied to any open tracked issue, so it's being closed to keep the review queue manageable.

This is not a rejection of the underlying idea.
If it's still relevant, please rebase onto current `main`, address any outstanding review feedback, and reopen this PR (or open a fresh one) — it'll be picked up again.

Thanks for the contribution.

PR #5538, opened 2026-05-22
fix(ci): restore main — rustfmt drift + 2 PR-only test failures

### LINKED ISSUE #5537: CI: restore main — rustfmt drift + 2 PR-only test failures
## Problem

The push-CI lane on `main` is green, but it **skips** the Quality (rustfmt) job and the integration test shards — those jobs are `pull_request`-triggered only. As a result, two classes of breakage have silently accumulated on `main` and now fail on **every open pull request**, blocking the queue.

This is the same pattern as the prior restore (#5307 / 7f58a89c).

## Failures gating every PR

1. **Workspace-wide rustfmt drift** — `cargo fmt --all --check` is dirty on current `origin/main` under CI's pinned rustfmt (1.9.0-stable). The Quality / "Check formatting" step (`cargo xtask fmt`) fails. ~29 files across 7 crates.

2. **`comms_send_rejects_oversize_message`** (`crates/librefang-api/tests/network_routes_integration.rs`) — the fixture builds a ~64 KiB payload, but `validation::MAX_MESSAGE_BYTES` is 256 KiB, so the handler returns 200 instead of 413 and the assertion fails. The fixture must track the real cap.

3. **`every_kernel_config_struct_field_is_exposed_via_overlay`** (`crates/librefang-api/tests/config_schema_overlay.rs`) — the `KernelConfig.external_auth_proxy` bool added in #5357 has no UI-overlay descriptor and no `EXCLUDED` entry, so the test flags it as invisible in the dashboard ConfigPage.

## Fix

A single restore PR:

- Run `cargo fmt --all` to clear the drift.
- Pin the oversize fixture to `validation::MAX_MESSAGE_BYTES + 1` so it tracks the cap.
- Surface `external_auth_proxy` on the general ConfigPage section (next to `require_auth_for_reads`, which it gates) and add the matching `EXCLUDED` entry.

houko:
Closes #5537

## Root cause

Push-CI on \`main\` is green but **skips** the Quality job (rustfmt, `cargo clippy --all-features`) and the integration shards — those lanes are `pull_request`-only. Drift and feature-gated compile breakage therefore accumulate on `main` undetected and gate every open PR's Quality job.

## What this PR does (rebased onto current main — scope reduced)

Two of the four issues from the original draft already landed on `main` independently, so this PR is now minimal:

1. **`ssh-backend` compile break** (`crates/librefang-runtime/Cargo.toml`, `Cargo.lock`) — dependabot #5271 bumped `russh-keys` 0.45 → 0.49 while `russh` stays pinned at 0.45 (last release on `rand 0.8`; later releases pull a forked ssh-key surface that clashes with the workspace `rand 0.10` dep — see the comment block in the Cargo.toml). `tool_exec_ssh.rs` is written against the 0.45 API, so `--features ssh-backend` fails to compile under 0.49. The break is invisible to default CI but fails Quality's `--all-features` clippy lane. Reverted `russh-keys` to 0.45; `Cargo.lock` drops the 0.49-only subtree. Verified: `cargo check -p librefang-runtime --features ssh-backend` passes.

2. **Workspace rustfmt drift** — `cargo fmt --all` over the tree; 17 files reformatted to satisfy CI's stable rustfmt (1.9.0-stable). Pure formatting, no logic changes.

## Already on main (dropped from this PR)

- `comms_send_rejects_oversize_message` fixture now pins `MAX_MESSAGE_BYTES + 1` on main (`network_routes_integration.rs:421`).
- `external_auth_proxy` overlay descriptor + EXCLUDED entry already present on main (`config_schema_overlay.rs:231`).

## Not in scope: prettier

The Quality job runs `cargo xtask fmt`, whose prettier step is gated on `has_command("prettier") || has_command("pnpm")`. The Quality runner installs neither, so CI prints **"prettier/pnpm not found — skipping web formatting"** and never checks prettier. Reformatting web/dashboard/docs would be a 200+-file diff that no CI gate verifies, so it is deliberately excluded.

## Verification

- `cargo check -p librefang-runtime --features ssh-backend` — passes (was the failing build).
- `cargo fmt --all --check` — clean (0 remaining diffs).
- Branch reset onto current `origin/main` and re-derived (not a stale rebase), so the diff is exactly the two concerns above.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1748,17 +1748,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "delegate"
-version = "0.13.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "780eb241654bf097afb00fc5f054a09b687dad862e485fdcf8399bb056565370"
-dependencies = [
- "proc-macro2",
- "quote",
- "syn 2.0.117",
-]
-
-[[package]]
 name = "deltae"
 version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4624,7 +4613,7 @@ dependencies = [
  "reqwest",
  "rusqlite",
  "russh",
- "russh-keys 0.49.2",
+ "russh-keys",
  "seccompiler",
  "serde",
  "serde_json",
@@ -5927,21 +5916,6 @@ checksum = "30d5b2194ed13191c1999ae0704b7839fb18384fa22e49b57eeaa97d79ce40da"
 dependencies = [
  "libc",
  "winapi",
-]
-
-[[package]]
-name = "pageant"
-version = "0.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "032d6201d2fb765158455ae0d5a510c016bb6da7232e5040e39e9c8db12b0afc"
-dependencies = [
- "bytes",
- "delegate",
- "futures",
- "rand 0.8.5",
- "thiserror 1.0.69",
- "tokio",
- "windows 0.58.0",
 ]
 
 [[package]]
@@ -7611,8 +7585,8 @@ dependencies = [
  "poly1305",
  "rand 0.8.5",
  "rand_core 0.6.4",
- "russh-cryptovec 0.7.3",
- "russh-keys 0.45.0",
+ "russh-cryptovec",
+ "russh-keys",
  "sha1",
  "sha2",
  "ssh-encoding",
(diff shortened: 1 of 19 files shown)


houko:
Added a third commit formatting `web/`, `crates/librefang-api/dashboard/`, and `docs/` with prettier (309 files).

Root cause: the CI "Quality / Check formatting" step runs `cargo xtask fmt`, which checks **both** `cargo fmt --all` (Rust) **and** `prettier --check` on web/dashboard/docs `src/**`. Main had pre-existing prettier drift (web 28, dashboard 220, docs 61) in addition to the rustfmt drift, so Quality was red on every PR even after the Rust fmt was clean. Ran `prettier --write` per dir; all three now pass `prettier --check`.

This PR now restores: rustfmt + prettier formatting + the two PR-only test failures (`comms_send_rejects_oversize`, `external_auth_proxy` overlay). No logic changes.

PR #5286, opened 2026-05-19
chore(memory): populate sessions.peer_id on save

f-liva:
## Summary

Migration v16 added a `peer_id` column to `sessions` (plus an
`idx_sessions_peer(agent_id, peer_id)` index) for per-user isolation
queries, but `save_session` never wrote the field. Every row inserted
since v16 has `peer_id IS NULL` — the index is dead weight and
`SELECT … WHERE peer_id = ?` returns nothing.

This wires the field through end-to-end:

- `Session.peer_id: Option<String>` in-memory.
- `save_session` INSERT + ON CONFLICT UPDATE write the column;
  `get_session` / `get_session_with_created_at` / `find_session_by_label`
  read it back.
- `send_message_full` and `execute_llm_agent` derive `peer_id` from
  `SenderContext.chat_id` for channel-scoped paths (the same value
  that already feeds `SessionId::for_sender_scope(agent, channel,
  chat_id)`); canonical / auto-dream / cron / fork / explicit-override
  sessions keep `None`.
- Existing pre-write rows get a one-shot backfill on first touch when
  the current turn supplies a peer — never trampling an already-set
  value.

Backfill of pre-existing rows is intentionally skipped — there is no
reliable way to reconstruct `peer_id` from the stored `messages` blob
without parsing each session's first inbound message, which is brittle.

## Test plan

- [x] `cargo test -p librefang-memory` — 243 passed, 1 ignored (3 suites)
- [x] `cargo test -p librefang-runtime` — 1818 passed, 6 ignored (14 suites)
- [x] `cargo clippy -p librefang-memory --all-targets -- -D warnings` clean
- [x] `cargo clippy -p librefang-kernel --lib -- -D warnings` clean
- [x] `cargo clippy -p librefang-runtime --all-targets -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] New regression test `test_save_session_persists_peer_id` reads
  back the raw SQLite column (not just the loaded `Session`) so a
  future bug that round-trips the in-memory field correctly while
  silently dropping the SQL bind still surfaces. Covers INSERT,
  UPSERT, and None-stays-None cases.

Note: `cargo check --workspace --lib` fails the api crate with a
pre-existing `opentelemetry_sdk` 0.31 vs 0.32 duplicate trait error
from a recent dependabot merge; the error reproduces on plain
`upstream/main` without these changes.

PR #5647, opened 2026-05-22
fix(runtime): propagate sender peer_id through remember_interaction_b…

### LINKED ISSUE #5474: bug(runtime): remember_interaction_best_effort drops sender peer_id — orphaned NULL-peer rows per turn in multi-user agents
## Summary

After every agent turn, `remember_interaction_best_effort` persists a `[Past exchange]` episodic memory for retrieve-continuity. The function does **not** accept `peer_id`, so every row lands with `peer_id = NULL` — even when the originating channel message carried a populated `SenderContext` (Telegram/Slack/Discord with known user identity).

For multi-tenant agents (one agent serving multiple human users through e.g. a Telegram group bot or shared workspace bot), this means:
- Every turn from any user → orphaned NULL-peer row stamped to `agent_id`.
- Retrieve is `(agent_id, peer_id)`-filtered for the current user, so `NULL ≠ "user-123"` → these rows are **invisible at recall**. No cross-user leakage of content, but…
- …no per-user session-bridge memory continuity either. The function's stated purpose (persist exchange so future turns can recall it) silently doesn't deliver in any multi-user channel.
- Storage bloat — one orphaned row per turn × N users × M agents accumulates indefinitely (eviction is per-agent cap, but the rows themselves are unreachable).

## Repro

Beta.12 (`2026.5.17-beta.12`), real Telegram traffic, single peer (peer_id=23244855):

```bash
# 1. Send message via Telegram to agent
# 2. Wait for reply
# 3. Query SQLite:
sqlite3 ~/.librefang/data/librefang.db "
SELECT scope, COALESCE(peer_id,'NULL') AS peer, substr(content,1,80)
FROM memories
WHERE agent_id='<agent-id>'
ORDER BY created_at DESC LIMIT 1;
"
```

Observed:
```
episodic|NULL|[Past exchange]
Them: <user message>
You: <agent reply>
```

`SenderContext.user_id` was populated end-to-end (kernel injected `sender_user_id` into manifest metadata, agent_loop read it as `ctx.sender_user_id = Some("23244855")`) — but the value is dropped at the `remember_interaction_best_effort` call site.

## Root cause

`crates/librefang-runtime/src/agent_loop/end_turn.rs:263`:

```rust
remember_interaction_best_effort(
    ctx.memory,
    ctx.embedding_driver,
    ctx.session.agent_id,        // ← passes only agent_id
    &interaction_text,
    ctx.streaming,
)
```

`crates/librefang-runtime/src/agent_loop/prompt.rs:66` signature has no `peer_id` parameter. Downstream `MemorySubstrate::remember_with_embedding_async` also doesn't accept it. The storage layer (`crates/librefang-memory/src/semantic.rs:131`) does accept `peer_id: Option<&str>` — proven by `auto_memorize`, which writes the same `agent_id, peer_id` pair correctly.

## Proposed fix

1. `remember_interaction_best_effort`: add `peer_id: Option<&str>` parameter (same shape as `auto_memorize`).
2. `MemorySubstrate::remember_with_embedding_async` (and the `_plain` variant): propagate `peer_id` to `SemanticMemoryStore::add`.
3. `end_turn.rs:263`: pass `ctx.sender_user_id` alongside `ctx.session.agent_id`.

Patch is ~10 lines across 3 files; storage layer already handles it.

## Severity

Medium for current users, high for anyone building a multi-tenant deployment on top of LibreFang (e.g. shared Telegram bot serving a team) — they get neither per-user continuity nor accurate storage accounting from the episodic memory path. No security impact (recall filter prevents bleed), but feature silently doesn't work as documented.

## Context

Found while building a multi-user Telegram CRM agent. Three related findings ([sibling issues to be filed]):
- `ProactiveMemoryOverrides` lacks `extraction_model` field — per-agent extractor swap unavailable.
- `[agents.<name>.proactive_memory]` config.toml block silently ignored (real path is `agent.toml [proactive_memory]`).

Chukwuebuka-2003:
## Type
- [x] Bug fix
## Summary
`remember_interaction_best_effort` dropped `sender_user_id` when persisting `[Past exchange]` episodic memories, causing every row to land with `peer_id=NULL`. For multi-tenant agents this rendered per-user recall continuity broken — rows were stored but invisible at query time since recall filters by `(agent_id, peer_id)`. Fixes #5474.
## Changes
- Added `peer_id: Option<&str>` to the `Memory` trait's `remember` method
- Threaded `peer_id` through `MemorySubstrate::store_with_chunking`, `remember_with_embedding`, `remember_with_embedding_async`, and `remember`
- Added `peer_id` parameter to `remember_interaction_best_effort`
- Pass `ctx.sender_user_id` at the call site in `end_turn.rs` (was already available in `FinalizeEndTurnContext`, just never used)
- Updated all downstream call sites with `None` for unchanged paths
- Storage layer (`SemanticStore::add`) already accepted `peer_id` — the parameter just wasn't plumbed through
## Attribution
## Testing
- [x] `cargo test --workspace` passes
## Security
- [x] No new unsafe code


houko (changes requested):
Thanks for the contribution, and welcome! Threading the sender `peer_id` through `remember_interaction` is the right fix and the call-site updates across `librefang-runtime` + `librefang-memory` look consistent. (Note: an earlier red CI on this branch was an unrelated main regression in `config_reload` that has since been fixed on main via #5652 — re-running CI on the updated branch should clear it.)

One thing needed before merge, per the repo testing rule in CLAUDE.md: this behaviour/signature change needs a test. Please add one that pins the propagation — e.g. drive the memory write with a known sender `peer_id` and assert the stored interaction carries it, plus a `None`/absent-peer_id case. A test at the injection site (where `remember_interaction` is invoked) is ideal so a future refactor that drops the argument is caught. Once that's in and CI is green, happy to merge.

houko (changes requested):
The core plumbing is correct — `ctx.sender_user_id` was already present in `FinalizeEndTurnContext` and unused at the `remember_interaction_best_effort` call site. The signature propagation through `Memory::remember`, `MemorySubstrate::remember_with_embedding` / `remember_with_embedding_async` / `store_with_chunking`, and down to `SemanticStore::remember_with_embedding_and_peer` is mechanically clean. All existing call sites correctly receive `None`. The shard-1 CI failure is a GitHub Actions credential timeout on checkout, not a code defect.

Two blocking issues:

**1. No regression test for the actual fix (CLAUDE.md §MANDATORY: Integration Testing)**

The PR adds no test that seeds an episodic memory with a non-`None` peer_id and then verifies the SQLite row carries the expected value. `substrate.rs` already has `test_substrate_remember_recall`; a parallel `test_remember_persists_peer_id` that calls `memory.remember(agent_id, "...", ..., Some("user-42"))` and then reads the raw column (as done in PR #5286's `test_save_session_persists_peer_id`) would catch any future regression silently. Without it, a refactor that keeps the in-memory type correct but drops the SQL bind will not be caught before production.

**2. Semantic mismatch with #5286 (`sender_user_id` ≠ `sessions.peer_id`)**

PR #5286 (`chore(memory): populate sessions.peer_id on save`) derives `sessions.peer_id` from `SenderContext.chat_id` — the platform conversation id — while this PR threads `sender_user_id` as `memories.peer_id`. At `messaging.rs:2226-2244` the kernel already distinguishes the two: `peer_id` is pulled from `sender_context.user_id`, but the session `peer_id` (#5286) is derived from `chat_id`. If the session-level isolation query joins on `(agent_id, peer_id)` using `chat_id` while memory recall filters on `sender_user_id`, the two scopes will disagree for group-chat scenarios where `chat_id != user_id`. The two PRs need to agree on a single canonical identity before either lands, or at minimum document the intentional divergence. This is not a theoretical concern: `messaging.rs:449` and `messaging.rs:2226` both call `peer_scoped_key("user_name", peer_id)` where `peer_id` is `sender_context.user_id`, consistent with this PR, but #5286 uses `chat_id` for sessions — that asymmetry will surface when per-user recall fails to match the session's `peer_id` filter.

Minor (non-blocking):
- The `#[allow(clippy::too_many_arguments)]` suppressor added to `remember_with_embedding` and `remember_with_embedding_async` is reasonable given the existing large signature, but the underlying argument count (now 7 in `MemorySubstrate::remember_with_embedding`) could be reduced with a small `EpisodicMemoryOptions` struct if a follow-on PR needs to add more parameters (e.g., `chat_scope`). Not required now.

houko (approved):
Both prior blockers are addressed:

1. Regression test added at the injection site (`crates/librefang-runtime/src/agent_loop/prompt.rs:919-1000`) — covers the `Some(peer_id)` / `None` paths, asserts peer-scoped recall hits the right row, cross-user query returns zero, and a NULL-peer row remains findable without filter. Dropping the `peer_id` arg breaks the test, so future regressions are caught at the injection layer.
2. Intentional divergence between `memories.peer_id` (per-user recall via `sender_user_id`) and `sessions.peer_id` (chat-scoped isolation via `chat_id`) documented at `crates/librefang-runtime/src/agent_loop/end_turn.rs:257-265`, with the rationale for not collapsing them (group-chat per-user recall would break).

Verified the regression test passes locally; CI is green across all required lanes. LGTM.

Issue #3054, opened 2026-04-24
feat(security): unified granular RBAC — user-centric permissions across all channels, tools, agents, and memory


DaBlitzStein:
## Motivation

LibreFang already has the building blocks for multi-user access control, but they exist as **isolated, incomplete subsystems** that don't compose into a coherent authorization story. As deployments grow from single-user setups to teams and communities, the lack of a unified RBAC model creates real pain:

- **Identity fragmentation**: A human who uses Telegram, Slack, and the web dashboard appears as three separate entities. There is no way to say "Alice across all channels has admin rights."
- **Coarse role granularity**: The current `UserRole` enum (`Viewer → User → Admin → Owner`) is a flat hierarchy with hard-coded action mappings. You cannot express "this user can chat with agent X but not agent Y" or "this user can use filesystem tools but not shell tools."
- **No per-user budgets**: `BudgetConfig` tracks global and per-provider spend but has no concept of per-user cost attribution or limits.
- **Channel-blind authorization**: `ChannelOverrides` controls DM/group policies per channel, and `ChannelToolRule` gates tools per channel, but neither is user-aware. You cannot say "user X is trusted in DMs but restricted in group chats."
- **Memory has no access scoping**: Any agent can read/write any memory namespace. There is no role-gated memory access or PII-aware field restrictions.
- **Audit gaps**: The Merkle hash chain `AuditLog` records agent-level actions (`ToolInvoke`, `ShellExec`, etc.) but does not track **who** (which human user) triggered the action, and there is no dashboard UI or API to query the audit trail.

This proposal unifies these subsystems into a coherent, granular, user-centric RBAC framework.

### What works
| Component | Location | Status |
|-----------|----------|--------|
| `UserRole` enum (Viewer/User/Admin/Owner) | `crates/librefang-kernel/src/auth.rs` | Functional but coarse |
| `AuthManager` with channel bindings | `crates/librefang-kernel/src/auth.rs` | Maps platform IDs to users |
| `Action` enum (8 actions) | `crates/librefang-kernel/src/auth.rs` | Hard-coded role→action mapping |
| `UserConfig` with `channel_bindings` | `crates/librefang-types/src/config/types.rs` | TOML-based user definitions |
| `Capability` enum (23 capabilities) | `crates/librefang-types/src/capability.rs` | Agent-level, not user-level |
| `ToolPolicy` (allow/deny rules) | `crates/librefang-types/src/tool_policy.rs` | Per-agent, not per-user |
[...]

PR #5663, opened 2026-05-23
feat(approval): cache per-session approvals so the same tool only prompts once

### LINKED ISSUE #5600: MCP Tool call approval is not persistant in the same session
### Description

with `auto_approve = false`, each time the agent is calling a tool provided by MCP requires approval.
on the web ui chat, it shows up a pennel asking approval. 
once clicked the approve, it MCP tool call went though.
however, in the same chat session, the request for approval popups very time the same MCP tool call (exact same tool name).
I would assume the same tool call in the same session only need to be approved once?

I also tried to add the tool name into the agent.toml configuration under the `capabilities.tools` and `exec_policy.allowed_commands`. but it does not seem to work

### Expected Behavior

MCP tool can be added / control with `capabilities.tools` and / or `exec_policy.allowed_commands`
in same session, the same tool only need to approved once / or has option to approve once for whole session?

### Steps to Reproduce

as description.
i used default `assistant` with JIRA MCP listed in the MCP Marketplace

### LibreFang Version

librefang 2026.5.17-beta.12

### Operating System

Linux (x86_64)

### Logs / Screenshots

_No response_

houko:
## Summary

Once a user approves `tool_name` inside a chat session, subsequent calls of the **same tool name in the same session** auto-approve. Matches what the issue reporter (and most IDE / MCP-client UX) expected: one approval per (session, tool) is enough.

In-memory only — daemon restart drops the cache, matching the existing `remembered: DashMap<(String, String), ApprovalDecision>` slot at `crates/librefang-kernel/src/approval.rs:69`.

## Wiring

| File | Change |
|---|---|
| `crates/librefang-kernel/src/approval.rs` | New `session_approvals: DashMap<(String, String), ()>` field + `remember_session_approval` / `has_session_approval` / `forget_session_approval` / `clear_session_approvals` helpers. `resolve()` populates the cache on `Approved` when the request carries a `session_id` and the policy flag is on. |
| `crates/librefang-kernel/src/kernel/handles/approval_gate.rs` | `submit_tool_approval` checks the cache after the hand-agent fast-path. Cache hit → `ToolApprovalSubmission::AutoApproved`, same shape as the hand-agent carve-out. |
| `crates/librefang-types/src/approval.rs` | New policy field `ApprovalPolicy.cache_approvals_per_session: bool` (default `true`). |

## Defensive guards

- Empty `session_id` or empty `tool_name` never enter the cache and never match — prevents a missing context from granting blanket approval to all future calls.
- Cache is per-session **and** per-tool. Different sessions still each see one prompt per distinct tool; different tools in the same session still each see a prompt.
- `Denied` outcomes never populate the cache — only `Approved`.

## Opt-out for compliance

Operators who need per-call approval (compliance environments) set `cache_approvals_per_session = false` in `[approval]` — the cache stays empty even on `Approved` and the old per-call semantics persist.

## Tests

7 new tests in `approval::tests`. All passing:

```
$ cargo test -p librefang-kernel --lib -- approval::tests::session approval::tests::resolve_approved approval::tests::resolve_denied
test approval::tests::session_approval_forget_clears_entry ... ok
test approval::tests::session_approval_round_trip ... ok
test approval::tests::session_approval_idempotent_and_safe_against_empty_keys ... ok
test approval::tests::session_approval_clear_all_for_session_drops_only_matching ... ok
test approval::tests::resolve_approved_skips_cache_when_policy_disables ... ok
test approval::tests::resolve_denied_does_not_populate_session_cache ... ok
test approval::tests::resolve_approved_populates_session_cache_when_policy_allows ... ok
test result: ok. 7 passed; 0 failed
```

Coverage spans: helper round-trip, empty-key defence, single-entry forget, scoped clear, end-to-end via `submit_request` + `resolve` for both `Approved` and `Denied` outcomes, and policy opt-out.

## What this does NOT do

- Does not persist the cache across daemon restarts. Same scope as the existing in-memory `remembered` slot. If durability is wanted, the `pending_approvals` SQLite table pattern is the precedent.
- Does not address the issue's second ask ("add tool name into `capabilities.tools` and `exec_policy.allowed_commands`"). That's a separate auto-allowlist feature — call sites are different, security implications are wider, and it deserves its own issue + PR.
- Does not auto-clear the cache on session reset. A `/new` or `/reboot` in a chat channel will not wipe per-session approvals today. Adding it where `kernel::session_lifecycle` resets sessions is one extra line (`approval_manager.clear_session_approvals(&session_id)`), worth a follow-up once we agree it's wanted.

Closes #5600.

### DIFF
--- a/crates/librefang-kernel/src/approval.rs
+++ b/crates/librefang-kernel/src/approval.rs
@@ -67,6 +67,15 @@ pub struct ApprovalManager {
     /// memory so user-level remembered decisions don't outlive an
     /// `acp` invocation in surprising ways.
     remembered: DashMap<(String, String), ApprovalDecision>,
+    /// Per-session approval cache (#5600): once a user approves a tool
+    /// inside a chat session, every subsequent call of that exact tool
+    /// name in the same session auto-approves without re-prompting.
+    /// Keyed on `(session_id, tool_name)`; populated by [`Self::resolve`]
+    /// on `Approved` outcomes and read in
+    /// `LibreFangKernel::submit_tool_approval`. Lives in memory only —
+    /// daemon restart drops the cache, matching the in-memory
+    /// `remembered` slot above.
+    session_approvals: DashMap<(String, String), ()>,
 }
 
 struct PendingRequest {
@@ -110,6 +119,7 @@ impl ApprovalManager {
             failure_rw_mutex: StdMutex::new(()),
             events_tx,
             remembered: DashMap::new(),
+            session_approvals: DashMap::new(),
         }
     }
 
@@ -135,6 +145,7 @@ impl ApprovalManager {
             failure_rw_mutex: StdMutex::new(()),
             events_tx,
             remembered: DashMap::new(),
+            session_approvals: DashMap::new(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


houko:
Found a security regression while reviewing this — pushed a fix.

**Issue**: the session cache check in `submit_tool_approval` runs after the hand-agent fast-path but does not look at `deferred.force_human`. When a Viewer/User chats with a hand-tagged agent the RBAC M3 carve-out (#3054) sets `force_human=true` on every call to surface approval to a human. With the cache as originally written:

- Call 1: `force_human=true` → hand-agent block skipped → cache miss → approval surfaced → user approves → cache populated.
- Call 2: `force_human=true` → hand-agent block skipped → **cache hit → AutoApproved**. RBAC M3 invariant broken from call 2 onward.

**Fix** (`1c307a9c`):
- `submit_tool_approval` (`approval_gate.rs:281`): added `&& !deferred.force_human` to the cache-lookup guard.
- `ApprovalManager::resolve` (`approval.rs:1012`): added `&& !from_force_human` to the cache-population guard, so even if a future caller forgets the read-side check, a force_human approval never taints the session cache.
- New test `resolve_approved_skips_cache_when_deferred_force_human` locks the write-side invariant.

All 8 session-cache tests pass locally:
```
cargo test -p librefang-kernel --lib -- approval::tests::session approval::tests::resolve_approved approval::tests::resolve_denied
test result: ok. 8 passed; 0 failed
```

Issue #3721, opened 2026-04-27
CLAUDE.md "Live Integration Testing" workflow is manual curl — automate as #[tokio::test]


houko:
## Severity
**Medium** for project quality. The repo already has the test infrastructure to enforce this — it just isn't used.

## Location
- `CLAUDE.md` — "MANDATORY: Live Integration Testing" section
- `crates/librefang-api/tests/api_integration_test.rs` — 4,173 lines, real axum + reqwest test harness already in place

## Finding
CLAUDE.md describes an 8-step "Live Integration Testing" workflow that contributors are expected to run **manually** after every endpoint, feature, or wiring change. The steps include:

1. `tasklist | grep -i librefang && taskkill //PID <pid> //F`
2. `cargo build --release -p librefang-cli`
3. `GROQ_API_KEY=... target/release/librefang.exe start &`
4. `curl http://127.0.0.1:4545/api/<endpoint>`
5. POST/PUT real payloads
6. Verify metering / cost side-effects via additional curl
7. `curl ... | grep -c newComponentName` to verify dashboard HTML
8. `taskkill //PID <pid> //F`

This is a manual checklist. It is **not** part of `cargo test`, **not** part of CI, and **not** enforced by any pre-merge gate. The default outcome is: contributors skip steps that are inconvenient, and regressions land.

The most common bug it would catch — "endpoint added in `routes.rs` but not registered in `server.rs`" — is the same class of bug CLAUDE.md repeatedly warns about. But there's no automation to actually catch it.

## Wasteful aspect
`api_integration_test.rs` already has a `TestServer` that spawns real axum + reqwest. Steps 4–6 of the manual workflow are **trivial** to express as `#[tokio::test]` cases against `TestServer`. The barrier is not capability; it's that the workflow was written as a checklist instead of as code.

## Step 7 is fragile regardless
`curl ... | grep -c newComponentName` to verify dashboard HTML breaks the moment Vite minifies a class name or React renames a component. This was never a robust check; the modern equivalent is Playwright or Vitest+JSDOM hitting the SPA.

## Suggested fix

**Phase 1 — automate the dead-route audit:**

Add to `api_integration_test.rs`:

```rust
#[tokio::test]
async fn dead_route_audit() {
    let server = TestServer::start().await;
    let routes = server.app.routes();  // axum 0.7+: Router::layer-listing
    for path in routes {
        // GET should not panic and should not return 404 if listed
        let r = server.client.get(path).send().await.unwrap();
[...]

PR #5679, opened 2026-05-23
feat(api): add source attribution to GET /api/tools response

### LINKED ISSUE #5676: feat(api): add source attribution (builtin/mcp_server) to GET /api/tools response
## Problem

\`GET /api/tools\` (\`crates/librefang-api/src/routes/tools_sessions.rs:68\`) returns a flat list of all tools. MCP tools carry \`\"source\": \"mcp\"\` but builtins have no \`source\` field, and MCP tools don't indicate WHICH server they come from. The dashboard cannot group tools by origin.

## Proposed fix

- Add \`\"source\": \"builtin\"\` to each builtin tool entry.
- Add \`\"mcp_server\": \"<server_name>\"\` to each MCP tool entry (resolvable via the tool name prefix \`mcp_{server}_{tool}\`).

DaBlitzStein:
Closes #5676. Builtins carry `"source":"builtin"`, MCP tools carry `"mcp_server":"<name>"` derived from the `mcp_{server}_{tool}` name prefix. Enables dashboard to group tools by origin.


houko (changes requested):
Thanks for picking up #5676. A few issues to address before this can merge.

## Blocking

### 1. Reimplements a known-buggy helper instead of calling the correct one

The MCP server-name extraction:

```rust
let mcp_server: Option<&str> = t
    .name
    .strip_prefix("mcp_")
    .and_then(|rest| rest.split_once('_'))
    .map(|(server, _tool)| server);
```

is the same naive split that `librefang_kernel::mcp::extract_mcp_server` performs, and `extract_mcp_server`'s own doc-comment explicitly warns against using it (`crates/librefang-runtime-mcp/src/lib.rs:2529-2543`):

> **Warning**: This heuristic splits on the first `_` after the `mcp_` prefix, so it only works for single-word server names (e.g. `"github"`). For server names that contain hyphens or underscores (e.g. `"my-server"` → `"mcp_my_server_tool"`), this returns only the first segment (`"my"`). Prefer [`resolve_mcp_server_from_known`] when the list of configured server names is available.

For an MCP server named `my-server` the tool is namespaced to `mcp_my_server_do_thing` (see `format_mcp_tool_name` + `normalize_name` at `lib.rs:2520-2578`), so this PR will report `"mcp_server": "my"`. The companion dashboard PR #5680 groups by `MCP:<server>`, so users will see unrelated servers merged under one group.

The list of configured servers is reachable from `AppState`, and the same crate already uses the robust pattern in `crates/librefang-api/src/routes/agents.rs:4326-4344`:

```rust
let configured_servers: Vec<String> = state
    .kernel
    .effective_mcp_servers_ref()
    .read()
    .map(|servers| servers.iter().map(|s| s.name.clone()).collect())
    .unwrap_or_default();
// ...
if let Some(server) = librefang_kernel::mcp::resolve_mcp_server_from_known(
    &tool.name,
    configured_servers.iter().map(String::as_str),
) { ... }
```

Please mirror that pattern here. If the resolver returns `None` (configured server list empty / tool from an unknown server), it's fine to fall back to omitting `mcp_server` rather than emitting a misleading prefix.

### 2. Missing integration test for the new wire shape

LibreFang's CLAUDE.md treats `#[tokio::test]` coverage in `crates/librefang-api/tests/` as mandatory for any route shape change:

> Add a `#[tokio::test]` against `TestServer` in the matching `tests/*.rs` file. ... This is the canonical replacement for the old curl checklist — it catches missing `server.rs` registrations, un-deserialized config fields, kernel↔API type drift, and empty/null payloads.

The existing `list_tools_returns_builtins_with_total` test (`crates/librefang-api/tests/system_tools_sessions_test.rs:91-131`) asserts `name` / `description` / `input_schema` but not `source` or `mcp_server`. Because the OpenAPI response is typed as `Vec<serde_json::Value>` the `OpenAPI Drift` check cannot catch a regression in the new fields either, so the integration test is the only safety net.

Please extend that test (or add a sibling) to assert:
- every builtin entry has `"source": "builtin"`,
- `total == tools.len()` still holds,
- a fixture MCP server with a hyphenated or underscored name (e.g. `my-server`) produces `"mcp_server": "my_server"` and `"source": "mcp"` — this would have caught issue 1 above.

## Non-blocking but in-scope

### 3. `get_tool` shape diverges from `list_tools`

Same file, lines 117-128: `GET /api/tools/{name}` for a builtin returns `{name, description, input_schema}` with no `source`. The MCP branch (lines 132-144) sets `source` but no `mcp_server`. Dashboard code that hits the detail endpoint for a tool it already grouped via list-tools will lose the attribution. Since the whole point of this PR is dashboard grouping, the detail endpoint should carry the same fields. Please align both endpoints in this PR rather than deferring — it's the same handler file and the same field shape.

## Nits

- Stripping the prefix in two passes (`strip_prefix("mcp_")` then `split_once('_')`) and converting `Option<&str>` to `String` only to push into a `serde_json::Value` is doing real work for what's now a one-liner once you switch to `resolve_mcp_server_from_known`.

houko (changes requested):
Source attribution direction is right and the wire shape matches what the sibling dashboard PR (#5680) expects to read, but a few items should land before merge.

## Blocking

- **Missing integration test for the new fields** (per CLAUDE.md / #3721 mandate). `crates/librefang-api/tests/system_tools_sessions_test.rs:92` (`list_tools_returns_builtins_with_total`) was not extended. Add assertions that:
  - every builtin entry carries `"source": "builtin"`,
  - MCP entries (when present in the test fixture) carry `"source": "mcp"` and a string `mcp_server`,
  - the field is additive (existing `name` / `description` / `input_schema` keys are unchanged).
  Without this, a future refactor of `list_tools` can silently drop `source` and #5680's grouping UI breaks with no test failure.

## Should fix

- **Reimplements an existing helper.** `crates/librefang-api/src/routes/tools_sessions.rs:87-91` duplicates `librefang_runtime_mcp::extract_mcp_server` (`crates/librefang-runtime-mcp/src/lib.rs:2538`). Call the helper instead — same logic, one source of truth, no risk of the two drifting.

- **Server-name truncation for hyphen/underscore names.** Both the inline parser and `extract_mcp_server` split on the first `_` after the `mcp_` prefix. Server `"my-server"` normalises to `"my_server"` (`normalize_name` at `crates/librefang-runtime-mcp/src/lib.rs:2576`), produces prefix `"mcp_my_server_"`, and the current code returns `"my"` — wrong group label in the dashboard. The robust helper for this is `resolve_mcp_server_from_known` (`crates/librefang-runtime-mcp/src/lib.rs:2556`), which longest-prefix matches against the actual configured server list. Lock `mcp_connections_ref` once, snapshot the names, and feed them in. The docstring on `extract_mcp_server` itself flags this limitation and points at `resolve_mcp_server_from_known` as the preferred API.

- **`GET /api/tools/{name}` is now inconsistent with `GET /api/tools`.** The single-tool endpoint at `crates/librefang-api/src/routes/tools_sessions.rs:117-128` (builtin branch) does not include `source: "builtin"`, while the MCP branch at line 140 has had `source: "mcp"` for a while and the list endpoint now adds `source: "builtin"`. A dashboard drilling from list → detail sees different shapes. Apply the same field there for symmetry (and ideally the same `mcp_server` derivation in the MCP branch).

## Nit

- `source` is a free-form JSON string. A small Rust enum with `#[serde(rename_all = \"lowercase\")]` would make the contract explicit and let utoipa surface the variants in the OpenAPI spec rather than `Vec<serde_json::Value>`. #5680 already pre-empts this by typing `source?: \"builtin\" | \"mcp\" | string` defensively on the TS side.

Verification run locally on top of this branch: `cargo check --workspace --lib` ok, `cargo clippy --workspace --all-targets -- -D warnings` ok, `cargo test -p librefang-api --test system_tools_sessions_test` 17/17 pass (but none exercise the new fields, see Blocking above).

houko (changes requested):
**Requesting changes.** The server-name resolver swap to `resolve_mcp_server_from_known` is done and a builtin-source test was added — thanks. Two items from the prior round are still open:

1. `crates/librefang-api/src/routes/tools_sessions.rs:102-130` (`get_tool`) — the detail endpoint `GET /api/tools/{name}` still returns `{name, description, input_schema}` for a builtin with no `"source"`, and the MCP branch (~:132-144) sets `"source":"mcp"` but no `mcp_server`. The list endpoint now emits both fields, so a dashboard drilling list→detail sees a different shape. Add `"source":"builtin"` to the builtin branch and the same `mcp_server` derivation to the MCP branch.
2. `crates/librefang-api/tests/system_tools_sessions_test.rs:128-141` — the added test only covers the builtin/no-MCP boot. The prior review asked for an MCP fixture with a hyphen/underscore server name asserting `"source":"mcp"` and the correctly-resolved `mcp_server` — the exact case the resolver swap fixes, and it has zero coverage. Add a test booting a fixture MCP server named e.g. `my-server` and assert the entry carries `"mcp_server":"my-server"` (note `resolve_mcp_server_from_known` returns the original, not normalized, name) and `"source":"mcp"`.

houko (approved):
Re-reviewed — both prior points addressed:
- The detail endpoint `GET /api/tools/{name}` now emits `"source":"builtin"` (`tools_sessions.rs:129`) and the MCP branch calls `librefang_kernel::mcp::resolve_mcp_server_from_known` to add `mcp_server` (`:137-156`). List/detail shape now matches.
- New tests cover the missing cases: `list_tools_mcp_hyphenated_server_carries_source_and_mcp_server` (seeds `my-server`, asserts hyphen-preserving resolution), `get_tool_mcp_hyphenated_server_carries_source_and_mcp_server`, and `get_tool_builtin_carries_source_field` (regression guard).

Approving.

Issue #5259, opened 2026-05-19
feat(dashboard) allowing users to select specific tools per each MCP server in Agent wizard


AIHunter83:
### Description

In some case the agent shouldn’t have access for all tools under one MCP server
For example
**1-MCP filesystem
2-MCP sql**

### Agent A
It should allowed only for 
Read file from MCP filesystem 
List tables and fetch data only from MCP SQL 

### Agent B
Has access for all in file system
Has Only access for 4 tools only for MCP SQL

### Agent C
Have access for all tolls for each selected MCP server as long as the tools left empty for example in MCP tools assigned per Agent

houko:
Triaged — the backend half is partly there, the wiring + UI is the actual unit of work.

**What already exists on `origin/main` at `2e2dd834`:**
- `AgentManifest.tool_allowlist: Vec<String>` — `crates/librefang-types/src/agent.rs:1148`. Glob-pattern matching is wired in the runtime tool-name gate.
- Per-MCP-server enable/disable lives in the manifest's `[[mcp_servers]]` block.

**What's missing for the per-tool-per-server UX the issue describes:**
1. **Per-server tool allowlist field** — the existing `tool_allowlist` is agent-global. A user who wants "Agent A can `read_file` on `mcp_filesystem` but cannot `read_file` on `mcp_sandbox`" can't express it. Either:
   - Extend `McpServerConfig` (in `crates/librefang-types/src/agent.rs`, search `pub struct McpServer`) with `pub tool_allowlist: Vec<String>` (and matching `tool_blocklist`), and gate at the MCP dispatch boundary; OR
   - Switch `tool_allowlist` from `Vec<String>` to `BTreeMap<server_name, Vec<String>>` with a `"*"` shorthand. Breaking change → migration + #[serde] adapter on the agent manifest deserialiser.
2. **MCP tool discovery in the wizard** — the wizard needs to enumerate the tools each MCP server exposes so the operator can pick. Today the dashboard knows about server names (from `mcp_marketplace` data), not per-server tool inventories. Either query the server at wizard time (slow + fail-prone for offline servers), or rely on a cached `tools/list` snapshot the kernel keeps per server.
3. **Wizard UI section** — under each server card in `agent-wizard` (search `dashboard/src/pages/agents` or the wizard component), render a multi-select of the server's tools. Wire to a query hook that loads cached tool inventories and a mutation hook that PATCHes the agent manifest.
[...]

PR #5680, opened 2026-05-23
feat(dashboard): Tools tab in agent detail with grouped view (closes #5677)

### LINKED ISSUE #5677: feat(dashboard): Tools tab in agent detail — grouped by source, with per-tool allow/block
## Problem

The per-agent tool editor is hidden behind a small \"Tools ✎\" badge in the capabilities section of the agent detail view. It opens as a drawer with a flat list of 170+ tool names (no grouping by source). Operators cannot easily find or manage specific tools.

## Proposed fix

Add a \"Tools\" tab alongside the existing Skills / MCP / Channels / Schedule tabs in the agent detail panel. The tab should:

1. Show tools grouped by origin: **Builtin** / **MCP: filesystem** / **MCP: git** / etc. (requires source attribution from the API — see related issue).
2. Each tool shows: name, source badge, current status (allowed/blocked/default).
3. Per-tool allow/block toggle writing to \`tool_allowlist\` / \`tool_blocklist\` via the existing \`updateAgentTools()\` API.
4. Save button following the same pattern as Skills/MCP tabs.

Depends on: API source attribution issue (for grouping data).

DaBlitzStein:
## Summary

Adds a **Tools** tab to the agent detail panel with group-level management:

- Groups tools by origin (Builtin + each MCP server)
- Toggle entire groups on/off, expand chevron for per-tool checkboxes
- Partial assignment shown with amber border and N/M count
- Preserves existing `tool_blocklist`/`tool_allowlist` on save
- Error toast on save failure
- `pnpm typecheck` zero errors

## Soft dependency

Works without #5679. When #5679 is merged, MCP server grouping improves for multi-word server names (e.g. `MCP: my-server` instead of `MCP: my`).

## Test plan

- [ ] Open agent detail → Tools tab appears between Skills and Schedule
- [ ] All mode shows groups as read-only cards with tool counts
- [ ] Customize → toggle groups on/off, expand for per-tool control
- [ ] Save preserves existing blocklist/allowlist
- [ ] Save failure shows error toast
- [ ] `pnpm typecheck` zero errors


houko (changes requested):
Nice scope — a single dashboard file, contained, and convention-compliant (data via `useTools` / `useAgentTools` hooks, no inline `fetch()`, no inline query-key arrays). Typecheck is clean against `feat/dashboard-tool-picker` HEAD. Two small things to clean up before this merges, plus one clarification on the dependency note in the PR body.

### 1. Dead `"mcp"` tab branch — `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx`

The `agentTab` discriminated union still carries an `"mcp"` member that is never reachable:

- L219: `useState<"conversation" | "memory" | "skills" | "mcp" | "tools" | "schedule" | "logs">`
- L1087: `case "mcp": return renderSkillsTab(agent);`
- L833–841: the tab bar only pushes `conversation / memory / skills / tools / schedule / logs` — there is no entry that ever sets `agentTab` to `"mcp"`.

Either drop `"mcp"` from the union and the `case` (preferred — it's purely dead code today), or actually wire an MCP-server-detail tab. Routing the dead case to `renderSkillsTab` is the worst of both worlds: it implies the two are interchangeable when they aren't, and a future reader will assume the case is load-bearing. Same reasoning belongs in this PR (the dead branch was introduced here), not deferred to a follow-up.

### 2. PR body overstates the dependency on #5679

The description and commit body both say _"Depends on #5679 for source attribution data."_ but #5679 is still open and the current backend (`crates/librefang-api/src/routes/tools_sessions.rs:79–88`) already emits `source: "mcp"` for every MCP tool today. The PR's grouping code at L1352–1360 in `AgentsPage.tsx` handles this correctly: it groups by `tool.source === "builtin" || (!tool.source && !tool.name.startsWith("mcp_"))` → Builtin, else MCP, with a `tool.name.replace(/^mcp_/, "").split("_")[0]` fallback for the server segment (which mirrors `extract_mcp_server` in `crates/librefang-runtime-mcp/src/lib.rs:2538`, same single-word-server caveat).

So the PR actually works **today**, against the merged backend, without #5679. Please either:

- drop the "Depends on #5679" line from the PR body / commit, or
- explicitly state that #5679 only upgrades the `MCP: <server>` grouping fidelity for multi-word server names (the visible behavior change once #5679 lands is "`MCP: my-server` instead of `MCP: my`"), so the dependency is soft, not hard.

The current wording invites a reviewer to block on #5679 unnecessarily.

### 3. Minor — copy mismatch with description

The PR description says badges are "active/blocked/filtered" but the code in `renderToolsTab` only renders `blocked` / `allowed` / `builtin` badges (L1417–1432). Not blocking, just worth syncing the description to what the UI actually shows so the next reader doesn't go hunting for the "filtered" state.

### What's fine and doesn't need changes

- Conventions: all data access goes through `useTools` (L156) and `useAgentTools` (L160 + L502), both gated on tab visibility so the network calls don't fire on unrelated tabs. Good.
- Empty/loading/data states are all handled (L1382–L1387 loading, L1387–L1391 empty, L1392+ data).
- Sort ordering for groups is deterministic (Builtin pinned first, then alphabetical) — matches the existing dashboard pattern.
- The footer summary block (L1462–L1474) has the right gating so it doesn't render a meaningless "0 declared · 0 allowed" when no config exists.

Once the dead `"mcp"` case is gone and the PR body claim about #5679 is reconciled, this is good to go.

houko (changes requested):
Verified locally: dashboard `pnpm typecheck` clean, `pnpm test --run` 736/736 pass, `pnpm build` succeeds (~266 kB AgentsPage chunk, no new heavy deps), `cargo check --workspace --lib` and `cargo clippy --workspace --all-targets -- -D warnings` clean. Data-layer rules followed — `useTools` / `useAgentTools` / `useUpdateAgentTools` are pre-existing hooks with proper `agentKeys.detail` + `agentKeys.tools` invalidation; no inline `fetch` / `api.*`. Gracefully degrades to the `mcp_` name-prefix split when `source` / `mcp_server` from #5679 are absent on current main. Two issues to address before merge.

## Should fix

- **Save silently wipes any existing `tool_blocklist` / `tool_allowlist`** — `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx:1430-1438`. The backend treats `Some([])` as "clear" (`crates/librefang-api/src/routes/agents.rs:4099-4105`). The PR body acknowledges this is intentional, but users who set up a blocklist via the existing triple-dropdown editor and then later open the new Tools tab will lose that config on first Save with no warning. `agentToolCfg?.tool_blocklist` is already loaded — either preserve it by sending `tool_blocklist: agentToolCfg.tool_blocklist ?? []`, or surface a one-time confirm ("This will replace your existing allowlist / blocklist with a pure `capabilities_tools` allowlist") when `tool_blocklist?.length || tool_allowlist?.length > 0`.

- **No error feedback on save failure** — `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx:1430-1446`. `handleSave` attaches only `onSuccess`. On 4xx / 5xx the mutation fails silently and the user sees the Save button re-enable with no indication. The existing tools editor at line 2507-2509 shows the canonical pattern: `addToast(toastErr(err, t("agents.tools_save_failed", ...)), "error")`.

## Nits

- The `useEffect` at `AgentsPage.tsx:507-519` re-seeds `toolsDraft` from `tabAgentToolsQuery.data` whenever data refetches. After Save, invalidation triggers a refetch → effect runs again with fresh server data → re-seeds. Functionally fine because `toolsDraft === null` post-save, but worth a one-line comment explaining the intended seed-once-per-visit semantics.

- `tool.mcp_server ?? tool.name.replace(/^mcp_/, "").split("_")[0] ?? "mcp"` at `AgentsPage.tsx:1374` — the second `??` is dead (`split` always returns a non-empty array, so `[0]` is `string`, never `undefined`). Harmless, drop for clarity.

- AgentsPage.tsx is now 2812 lines. Out of scope for this PR but worth tracking — `renderToolsTab` (~300 lines) is a candidate to extract into `src/components/AgentToolsTab.tsx` together with the existing tools editor modal.

houko:
Heads-up from the #5259 triage: when verifying whether #5259 (per-MCP-tool selection in agent wizard) is covered by this PR, an additional **kernel-side wiring gap** surfaced.

This PR saves per-tool checkbox state into `capabilities_tools`, but `crates/librefang-kernel/src/kernel/tools_and_skills.rs:188-232` (Step 3 — MCP tool assembly) explicitly bypasses the `capabilities.tools` filter for MCP-sourced tools. Only `tool_allowlist` (Step 4) actually narrows MCP exposure. Net effect: the per-tool MCP checkboxes in this PR will silently do nothing on the kernel side — the agent still receives every tool the MCP server publishes.

Two paths:
- **(a) Dashboard-only fix** (recommended for this PR): pivot the save target from `capabilities_tools` to `tool_allowlist` for MCP-sourced entries. Backend behaviour stays put. Closes #5259 transitively.
- **(b) Kernel behaviour change**: extend Step 3 to honour `capabilities.tools` for MCP tools as well. Cleaner long-term but breaks existing agents that rely on the current MCP-passthrough semantics; needs a migration note and probably a separate PR.

Could you address path (a) in the next revision? Otherwise this PR is a UX trap — the checkboxes look authoritative but have no effect.

Reference: full triage in https://github.com/librefang/librefang/issues/5259#issuecomment-4532296163

houko (changes requested):
Re-reviewed — one of the three prior items is fixed (the dead `"mcp"` agentTab routing), but the other two remain:

1. `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx:1428-1447` (`handleSave`) — still sends `tool_allowlist: []` and `tool_blocklist: []` unconditionally. A user who has previously configured a blocklist via the triple-dropdown editor will lose it silently the first time they touch the new Tools tab and save. Either preserve the existing `agentToolCfg?.tool_blocklist ?? []` / `tool_allowlist ?? []`, or surface a one-time confirmation when the existing values are non-empty.

2. `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx:1438-1446` (`handleSave`'s mutation call) — still attaches only `onSuccess`. The existing tools editor's pattern uses `addToast(toastErr(err, …), "error")` on failure; adopt the same so the user gets a signal when a 4xx/5xx happens instead of the Save button silently re-enabling.

Fixed since last round: the `"mcp"` member is gone from the `agentTab` union, the switch case is removed, and the tabs array no longer pushes it.

houko (approved):
Re-reviewed — both prior points addressed:
- `handleSave` no longer wipes existing allowlist/blocklist: `tool_allowlist: agentToolCfg?.tool_allowlist ?? []` and `tool_blocklist: agentToolCfg?.tool_blocklist ?? []` (`AgentsPage.tsx:1434-1435`).
- The mutation now attaches `onError: (e) => addToast(toastErr(e, ...), "error")` (`:1444-1449`) matching the existing tools-editor pattern, so save failures surface as toasts instead of the button silently re-enabling.

Approving.

Issue #3576, opened 2026-04-27
arch: inconsistent error contracts — String / anyhow / LibreFangError / LlmError mixed


houko:
## Summary
Error type differs at every layer boundary: tool runner uses `Result<String, String>`, agent loop uses `LibreFangError`, LLM layer uses `LlmError`, kernel has a 2-variant `KernelError` pass-through, and the wider codebase mixes `anyhow` + `thiserror` within the same modules.

## Evidence
- `crates/librefang-runtime/src/tool_runner.rs` — 10+ `Result<String, String>` sites at lines 2253, 2264, 2365, 2377, 2402, 2436, 2457, 2493, 2548
- `crates/librefang-runtime/src/agent_loop.rs` — `LibreFangError` with structured variants
- `crates/librefang-kernel/src/error.rs` — `KernelError` has only `LibreFang(#[from] LibreFangError)` and `BootFailed(String)` — barely earning its keep
- 246 `Box<dyn>` / `Arc<dyn>` sites, 31 `anyhow` references, 7 `thiserror` derives

## Why it matters
Error context is lost at every boundary crossing. Cannot pattern-match on "rate-limited LLM" vs "tool authorization failure" without string-sniffing. Retry/backoff in `agent_loop.rs:3920 handle_retryable_llm_error` already does exactly that. Operators get noisy logs with the same root cause appearing under different shapes.

## Suggested direction
One structured error type per layer boundary, enforced: `ToolError` for tool-runner outputs (replace `Result<String, String>`), keep `LlmError` for drivers, `LibreFangError` as the application-level enum, drop `KernelError` (2-variant pass-through). Use `anyhow` only inside binaries / `xtask`, never in library crates. Add `clippy::disallowed_types` to ban `anyhow::Result` in `librefang-kernel` / `librefang-runtime` / `librefang-api`.

---
_Reported by automated multi-agent code review (security / concurrency / architecture / performance / testing). Evidence is grep-verified file:line; please re-confirm before fixing._

houko:
Closing during triage. The observation is real — `String` / `anyhow::Error` / `LibreFangError` / `LlmError` are not consistently used across crates — but as a bare issue this rots: 16 days with zero engagement, no PR, no migration plan, no concrete first crate to unify.

[...]

PR #5720, opened 2026-05-25
fix(runtime): enforce ownership in schedule_delete; migrate schedule_* to ToolError (#3576)

houko:
## What

Second slice of the #3576 typed-error migration, plus an authorization fix found while doing it.

### 1. Migrate `schedule_*` tools to `ToolError` (#3576)

Follows the cron slice (#5258) exactly. `tool_schedule_create` / `tool_schedule_list` / `tool_schedule_delete` now return `Result<String, ToolError>` instead of `Result<String, String>`:

- `require_kernel` -> `require_kernel_typed`.
- Missing params -> `ToolError::MissingParameter`; bad schedule string -> `ToolError::InvalidParameter { name: "schedule", reason }`.
- Kernel errors -> `ToolError::upstream` (preserves the source chain instead of stringifying).
- Missing caller agent id -> `MissingParameter("agent_id")` via a local `caller_agent_id_missing` helper mirroring cron's (the MCP route passes `None` legitimately -> user-recoverable 400, not 500).

The internal `parse_schedule_to_cron` / `parse_time_to_hour` helpers keep their `Result<_, String>` shape (a pure sub-layer with no kernel contact) and are mapped to `ToolError` at the tool boundary. The dispatch arms narrow back to `Result<String, String>` with `.map_err(|e| e.to_string())`, the same bridge pattern as the cron slice, until every submodule has migrated.

### 2. Fix: `schedule_delete` skipped the ownership check (authz bypass)

`KernelHandle::cron_cancel(job_id)` removes a job by UUID with **no** ownership filtering (`kernel/handles/cron_control.rs:123`). `tool_cron_cancel` compensates by verifying the caller owns the job before cancelling (`cron.rs:86-100`). But `tool_schedule_delete` — which ends at the same `cron_cancel` — did no such check and did not even take `caller_agent_id`. So any agent with the `schedule_delete` tool could cancel **another agent's** job by learning its UUID, trivially bypassing the guard `cron_cancel` enforces.

Fixed by threading `caller_agent_id` into `tool_schedule_delete` and applying the same ownership check (`cron_list` -> membership -> `NotFound` collapsing unowned/missing, mirroring `cron_cancel`). `schedule_delete` now never reaches the kernel for a job the caller does not own.

## Tests

- `schedule.rs` unit tests (boundary, no kernel): create/list/delete without a kernel -> `Unavailable`; `caller_agent_id_missing` -> `MissingParameter`.
- `tests/tool_runner_forwarding_task_cron.rs` integration tests with the `CapturingKernel` stub: `schedule_delete` succeeds when owned, renders `NotFound` and never reaches the kernel when unowned (the authz regression test), and `MissingParameter` when `id` is absent.

## Verification (Docker; host has no native toolchain)

Ran in the repo's `Dockerfile.rust-dev` image with isolated `CARGO_HOME` / `CARGO_TARGET_DIR` volumes:

- `cargo test -p librefang-runtime --lib tool_runner::schedule::` -> 4 passed, 0 failed
- `cargo test -p librefang-runtime --test tool_runner_forwarding_task_cron` -> 15 passed, 0 failed (incl. the 3 new `schedule_delete` tests)
- `rustfmt --check` on the changed files -> clean
- `cargo clippy -p librefang-runtime --all-targets -- -D warnings` -> clean

### DIFF
--- a/crates/librefang-runtime/src/tool_runner/dispatch.rs
+++ b/crates/librefang-runtime/src/tool_runner/dispatch.rs
@@ -898,12 +898,20 @@ pub async fn execute_tool_raw(
         "task_status" => tool_task_status(input, *kernel).await,
         "event_publish" => tool_event_publish(input, *kernel).await,
 
-        // Scheduling tools (delegate to CronScheduler via kernel handle)
-        "schedule_create" => {
-            tool_schedule_create(input, *kernel, *caller_agent_id, *sender_id).await
-        }
-        "schedule_list" => tool_schedule_list(*kernel, *caller_agent_id).await,
-        "schedule_delete" => tool_schedule_delete(input, *kernel).await,
+        // Scheduling tools (delegate to CronScheduler via kernel handle).
+        // Second slice of the #3576 typed-error migration: the submodule now
+        // returns `Result<String, ToolError>`; the dispatch arms narrow it to
+        // `Result<String, String>` here at the boundary so the broader
+        // dispatch table stays uniform until every submodule has migrated.
+        "schedule_create" => tool_schedule_create(input, *kernel, *caller_agent_id, *sender_id)
+            .await
+            .map_err(|e| e.to_string()),
+        "schedule_list" => tool_schedule_list(*kernel, *caller_agent_id)
+            .await
+            .map_err(|e| e.to_string()),
+        "schedule_delete" => tool_schedule_delete(input, *kernel, *caller_agent_id)
+            .await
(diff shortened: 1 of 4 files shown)


houko:
Superseded by #5737, which consolidates all 15 #3576 tool_runner migration slices into a single PR (one commit per tool family) — splitting this into separate PRs created an unnecessary review burden. No content lost; #5737 contains this slice verbatim plus the others.

PR #5741, opened 2026-05-25
feat(api,dashboard): expose auto_evolve toggle in Skills tab

DaBlitzStein:
## Summary

Expose the existing `auto_evolve` manifest field in the API and dashboard:

- Add `auto_evolve` to `GET /api/agents/{id}` response
- Add `auto_evolve` to `PATCH /api/agents/{id}` handler (via `update_auto_evolve` on `AgentRegistry`)
- Add toggle switch in the Skills tab of agent detail panel — instant save via PATCH
- `pnpm typecheck` zero errors, `cargo check` clean

## Test plan

- [ ] Open agent detail → Skills tab → auto-evolve toggle visible
- [ ] Toggle off → toast "Auto-evolve disabled" → API returns `auto_evolve: false`
- [ ] Toggle on → toast "Auto-evolve enabled" → API returns `auto_evolve: true`
- [ ] `cargo check --workspace --lib` clean
- [ ] `pnpm typecheck` zero errors

Refs: #5675

### DIFF
--- a/crates/librefang-api/dashboard/src/api.ts
+++ b/crates/librefang-api/dashboard/src/api.ts
@@ -1247,6 +1247,7 @@ export interface AgentDetail {
   thinking?: { budget_tokens?: number; stream_thinking?: boolean };
   is_hand?: boolean;
   web_search_augmentation?: "off" | "auto" | "always";
+  auto_evolve?: boolean;
 }
 
 export async function getAgentDetail(agentId: string): Promise<AgentDetail> {
--- a/crates/librefang-api/dashboard/src/lib/mutations/agents.ts
+++ b/crates/librefang-api/dashboard/src/lib/mutations/agents.ts
@@ -150,6 +150,7 @@ export function usePatchAgent() {
         provider?: string;
         mcp_servers?: string[];
         schedule?: AgentSchedulePatch;
+        auto_evolve?: boolean;
       };
     }) => patchAgent(agentId, body),
     onSuccess: (_data, variables) => {
--- a/crates/librefang-api/dashboard/src/pages/AgentsPage.tsx
+++ b/crates/librefang-api/dashboard/src/pages/AgentsPage.tsx
@@ -1290,9 +1290,9 @@ export function AgentsPage() {
         {
           onSuccess: () => {
             addToast(
-              t("agents.detail.auto_evolve_saved", {
-                defaultValue: !autoEvolve ? "Auto-evolve enabled" : "Auto-evolve disabled",
-              }),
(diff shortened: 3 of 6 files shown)


houko (changes requested):
**Requesting changes.** Dashboard conventions are satisfied (the toggle uses `usePatchAgent`, which invalidates `agentKeys.detail` + `agentKeys.lists()` in `onSuccess`; no inline fetch; persists to SQLite + disk). Two issues:

1. `crates/librefang-api/src/routes/agents.rs:2775` and `:4625-4630` — `auto_evolve` is added to the `GET /api/agents/{id}` response and the `PATCH /api/agents/{id}` handler with no `#[tokio::test]`, and no test under `crates/librefang-api/tests/` references `auto_evolve`. Per the repo's route/wiring-change mandate, this needs a TestServer test. Add a `#[tokio::test]` that PATCHes `{"auto_evolve": false}` and asserts the subsequent GET returns `auto_evolve: false`.
2. `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx:1262-1275` — both the enabled and disabled toasts use the same i18n key `agents.detail.auto_evolve_saved` with different `defaultValue`s. `t()` resolves to the single translation for that key in any non-default locale, so a translated UI shows one fixed string for both directions. Use two distinct keys (`auto_evolve_enabled` / `auto_evolve_disabled`) selected by the new state.

houko (changes requested):
Re-reviewed `8e19af5c` — both prior points addressed (round-trip integration test `test_patch_agent_auto_evolve_persists` covers GET/PATCH; the two toasts now use distinct keys `auto_evolve_enabled` / `auto_evolve_disabled`). One new issue introduced by the function insertion:

- `crates/librefang-kernel/src/registry.rs:499-509` — `update_auto_evolve` was inserted INTO THE MIDDLE of `update_schedule`'s three-line doc comment. The first two lines of that doc block (`/// Update an agent's schedule mode (Reactive / Periodic / Proactive / Continuous). Mutates the manifest only — the kernel-level wrapper / LibreFangKernel::set_agent_schedule is what callers should use to`) now sit immediately above `pub fn update_auto_evolve(...)`, mis-attributing the docstring. The trailing two lines (`/// also stop the prior background loop and start the new one so the / runtime actually reflects the change without a daemon restart.`) now read as an orphaned fragment above `update_schedule`. Both `cargo doc` and IDE hovers will show garbled docs for both functions. Move `update_auto_evolve` either above or below the intact `update_schedule` doc block so each function's comment stays contiguous with its declaration.

PR #5756, opened 2026-05-26
fix(terminal): allow remote WebSocket + add ncurses-term for xterm-256color

GQAdonis:
## Summary

Two separate 403s were blocking the Terminal page on bossfang.prometheusags.ai:

- `/api/terminal/health` → 403 `remote_no_auth`: terminal rejects all remote connections when `allow_remote = false` (default)
- `/api/terminal/ws` → 403 `origin_mismatch`: `https://bossfang.prometheusags.ai` not in `allowed_origins`

After those were fixed, shell spawn failed because `xterm-256color` terminfo entry is not present in the `bookworm-slim` runtime image.

## Changes

**`k8s/base/configmap.yaml`** — add `[terminal]` section:
- `allow_remote = true` — permit remote/proxied connections
- `allow_unauthenticated_remote = true` — required since pod runs with `LIBREFANG_ALLOW_NO_AUTH=1` (no api_key)
- `allowed_origins = ["https://bossfang.prometheusags.ai"]` — restrict to dashboard origin only

**`Dockerfile`** — add `ncurses-term` to runtime apt-get install. The terminal handler pins `TERM=xterm-256color` (required by xterm.js); that terminfo entry is not in `bookworm-slim` without this package.

## Verification

- ConfigMap + rollout restart already applied to production cluster
- WebSocket connection error is resolved (no longer "Reconnecting...")
- Shell spawn fix requires Docker rebuild via CI (this PR triggers it)

## Test plan

- [ ] CI docker-publish builds and pushes new image
- [ ] Deploy new image, open Terminal page — shell prompt appears
- [ ] Type a command (`ls`, `echo hello`) — output renders correctly


GQAdonis:
Opened against wrong repo — BossFang-specific deployment config, not intended for upstream.

PR #5757, opened 2026-05-26
fix(runtime): tool_runner sandbox — RAII cleanup, TOCTOU removed, container_id redacted

leszek3737:
## Type
- [x] Bug fix

## Summary

Hardens `tool_runner/sandbox.rs`: cancel-safe RAII container cleanup, TOCTOU elimination, container_id leak fix, structured error types.

## Changes

- `SandboxGuard` RAII struct with `Drop` impl — cleanup runs even on cancel/panic via `block_in_place` + `block_on`
- Removed separate `is_docker_available()` TOCTOU check — `create_sandbox` fails directly
- Removed `container_id` from JSON response (no longer leaks to callers)
- Return type changed from `Result<String, String>` to `ToolResult` with typed `ToolError` variants
- Dispatch bridge added for backward compat

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Heads-up: #5737 ("migrate tool_runner tools to ToolError", #3576) just merged to `main`, which retyped `sandbox.rs` to `ToolResult`. The `ToolResult` re-typing and the `dispatch.rs` `.map_err(|e| e.to_string())` bridge here are now byte-identical to merged work — please drop those on rebase. Beyond the rebase, three concrete issues:

1. `crates/librefang-runtime/src/tool_runner/sandbox.rs` (disabled branch) — collapsing the disabled case to bare `ToolError::Unavailable("Docker sandbox")` drops the operator-actionable hint and **fails the existing `docker_exec_disabled_returns_actionable_message` test**, which asserts the message contains `docker.enabled=true`. Keep the actionable message via `upstream_msg(...)`.
2. `crates/librefang-runtime/src/tool_runner/sandbox.rs` (removed `is_docker_available()` pre-check) — deleting the pre-check loses the "Install Docker to use docker_exec" hint; a host without Docker now surfaces the raw spawn error. This is a diagnostics regression, not a TOCTOU win (the create call was never racing a meaningful gate). Restore the pre-check or fold its message into the create error.
3. `crates/librefang-runtime/src/tool_runner/sandbox.rs` (`SandboxGuard::drop`) — `block_in_place` + `Handle::current().block_on` is correct on the multi-thread daemon runtime, but the drop path has no test and `Handle::current()` panics if the guard is dropped during runtime shutdown / outside a reactor. Add a `#[tokio::test(flavor = "multi_thread")]` exercising the success path (guard constructed and dropped).

The `SandboxGuard` RAII and the `container_id` removal are the genuinely novel parts — keep those after rebasing onto the merged ToolError shape.

leszek3737:
done

houko (approved):
Re-reviewed the latest commit (`9be179f5` + merge `6b0b1a0e`) — all three prior points addressed:
- The disabled-branch actionable hint (`docker.enabled=true`) is preserved.
- The `is_docker_available()` pre-check stayed and the rationale comment is restored.
- `sandbox_guard_drop_does_not_panic` (`#[tokio::test(flavor = "multi_thread")]`) exercises the new `block_in_place` + `Handle::current().block_on` drop path with a graceful skip when Docker isn't on the CI host.

Approving.

PR #5758, opened 2026-05-26
fix(runtime): tool_runner workflow — artifact type check, deterministic sort, recursion limit

leszek3737:
## Type
- [x] Bug fix

## Summary

Fixes 4 bugs in `tool_runner/workflow.rs`: non-string artifact passthrough, nondeterministic sort, unlimited JSON recursion, and ambiguous cancel status.

## Changes

- `_artifact` value validated as string — non-string values (e.g. `{"_artifact":123}`) now return error
- Sort uses `.then_with(|| a.id.cmp(&b.id))` secondary key for determinism
- `resolve_workflow_input_artifacts` depth-limited to 128 (prevents stack overflow)
- `tool_workflow_cancel` classifies errors as not_found/already_terminal/error instead of always "cancelled"

## Attribution

## Testing

## Security
- [x] No new unsafe code

### DIFF
--- a/crates/librefang-runtime/src/tool_runner/workflow.rs
+++ b/crates/librefang-runtime/src/tool_runner/workflow.rs
@@ -42,46 +42,47 @@ pub(super) fn prepare_workflow_input(raw: Option<&serde_json::Value>) -> Result<
     }
 }
 
-/// Recursively walk `value` and rewrite every `{"_artifact": "sha256:..."}`
-/// object into the bare handle string. Anything else passes through
-/// unchanged. Malformed handles fail fast with a clear error message so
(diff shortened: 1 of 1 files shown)


houko (dismissed):
Re-reviewed — the original verdict from the round-1 review stands: the artifact non-string type check, `then_with(id)` deterministic secondary sort, 128-depth limit on artifact resolution, and the `workflow_cancel` not_found/already_terminal/error classification are all sound. The substring-matching against `cancel_workflow_run`'s upstream error strings is acceptable for a module still on `Result<String, String>`.

Approving as-is. **However, this is now CONFLICTING/DIRTY against current main** — please rebase onto `dd1de430` (or later) before it can land.

PR #5759, opened 2026-05-26
fix(runtime): tool_runner schedule — AM/PM parsing, minute precision, owner verification, cron validation

leszek3737:
## Type
- [x] Bug fix

## Summary

Fixes 5 bugs in `tool_runner/schedule.rs`: AM/PM time parsing, dropped minutes, hardcoded minute 0, missing owner verification on cancel, and invalid cron field acceptance.

## Changes

- `parse_time_to_hour_minute` handles `9:30am` / `6:45pm` patterns with colon before suffix
- Returns `(hour, minute)` tuple — minutes no longer discarded
- `daily at`, `weekdays at`, `weekends at` use parsed minutes instead of hardcoded 0
- `cron_cancel` verifies caller owns the job before deletion
- `validate_cron_fields` rejects out-of-range values (minute 0-59, hour 0-23, etc.)

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and already implements part of this PR — `tool_schedule_delete` now takes `caller_agent_id`, calls `cron_list` + `owns_job`, and returns `ToolError::NotFound`. Please drop the owner-verification / `caller_agent_id` dispatch wiring on rebase (it duplicates merged work). Two concrete validation bugs in the genuinely new part:

1. `crates/librefang-runtime/src/tool_runner/schedule.rs` (`validate_cron_field`) — the range/step parser falsely rejects valid expressions the underlying `cron` 0.16 engine accepts. `10-20/2` is split on `-` into `["10","20/2"]` and `"20/2".parse::<u32>()` fails; point-step `1/2` also fails. `cron`'s own tests accept both (`test_nom_valid_period_range`), and LibreFang passes the 5-field expression through verbatim — so this regresses the passthrough branch. Parse `start-end/step` properly or drop the range-step validation.
2. `crates/librefang-runtime/src/tool_runner/schedule.rs` (`validate_cron_fields`, day-of-week `0..=6`) — `cron` 0.16's `DaysOfWeek` accepts ordinal `7` for Sunday. Rejecting `* * * * 7` is a false negative against the engine. Use `0..=7` for dow.

Keep the AM/PM minute-precision parsing (new and useful); rework the cron validation onto the now-`ToolResult` signatures.

leszek3737:
zrobione

houko (approved):
Re-reviewed `dd133287` — both validator bugs fixed:
- `validate_cron_field` now strips the `/step` suffix first and parses base independently, so `10-20/2`, `1/2`, `*/5` all pass.
- `validate_cron_fields` accepts day-of-week ordinal `7`.

Approving. Optional nit (non-blocking): add a small test pinning `step=0` rejection (`0/0` → "step must be >= 1, got 0") so the contract stays explicit; and document in the helper that the passthrough only supports `[0-9*/,-]` (the `?` placeholder some `cron` engines accept is not handled by the natural-language fallback either, so this is consistent with existing behavior).

PR #5760, opened 2026-05-26
fix(runtime): tool_runner system — URL const, client reuse, error diagnostics

leszek3737:
## Type
- [x] Bug fix

## Summary

Hardens `tool_runner/system.rs`: extracts constants, reuses HTTP client, improves error reporting.

## Changes

- `IP_API_URL` and `IP_API_TIMEOUT` extracted to module-level constants
- `proxied_client()` singleton reuse instead of per-call builder
- Non-2xx responses include body preview in error message
- `proxied_client()` now sets `User-Agent` to `librefang/<CARGO_PKG_VERSION>` (was `LibreFang/0.1`)

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and already migrated `tool_location_get` to `ToolResult`/`ToolError`; this PR reverts it to `Result<String, String>` with `format!` errors. The const-extraction, client reuse, and body-in-error diagnostics are fine to re-apply after rebase, but the substantive new feature is non-functional:

1. `crates/librefang-runtime/src/tool_runner/system.rs` + `dispatch.rs` (`geo_ip_enabled`) — the privacy opt-in can never be disabled. There is no `geo_ip` / `geo_ip_enabled` field in `librefang-types` or kernel config, every construction site sets it to `true` (kernel `mod.rs`, `execute_tool` at `dispatch.rs:1585`, all test `make_ctx`), and `execute_tool`'s parameter list isn't extended to accept the flag — so the only production caller (`agent_loop/tool_call.rs:606`) can't set it `false`. Either add a backing config field and thread it through `execute_tool` from the agent loop, or drop the gate.
2. `crates/librefang-runtime/src/tool_runner/system.rs` (error message) — `"Geo-IP lookup is disabled. Enable it in config.toml: geo_ip_enabled = true"` points operators at a config key that does not exist; the instruction is false. Fix once a real config surface exists, or remove it.
3. `crates/librefang-runtime/src/tool_runner/system.rs` — the explicit `.header("User-Agent", "LibreFang/0.1")` is dropped. `proxied_client_builder` still injects a default UA so it's not a break, but it's an unmentioned behavior change — confirm intentional.

leszek3737:
done

houko (approved):
Re-reviewed `045f6c8d` — the non-functional `geo_ip_enabled` gate is removed cleanly: no references remain in `system.rs` or the dispatch ripple, and the misleading "Enable it in config.toml: geo_ip_enabled = true" error string is gone. The const/client-reuse/body-in-error parts are intact.

Approving. **Please update the PR title and body before merge** — both are now stale:
- Title still reads "URL const, client reuse, privacy opt-in, error diagnostics" — drop "privacy opt-in".
- Body still claims `geo_ip_enabled` parameter gates geo-IP lookup and `ToolExecContext` gained the field — neither is in the diff anymore.
- Also note the unmentioned wire-visible change: `proxied_client()` now sets UA to `librefang/<CARGO_PKG_VERSION>` (was `LibreFang/0.1`). Worth a one-liner in the description.

PR #5761, opened 2026-05-26
fix(runtime): tool_runner media — size limits, async fs, UUID filenames, ffmpeg deadlock, extension allowlist

leszek3737:
## Type
- [x] Bug fix

## Summary

Hardens `tool_runner/media.rs` with 11 fixes: size limits, format-before-read, async I/O, error propagation, unique filenames, safe integer conversion, ffmpeg race fix, and extension allowlist.

## Changes

- `read_with_size_limit()` checks `metadata().len()` before reading (MAX_IMAGE_BYTES/MAX_AUDIO_BYTES)
- Extension validated before file read (format-before-read pattern)
- `std::fs` replaced with `tokio::fs` throughout (no blocking in async)
- Silent errors (`let _ =`, `.is_ok()`) replaced with `?` propagation
- UUID-based filenames prevent `image_0.png` overwrite collisions
- Timestamp precision: `%Y%m%d_%H%M%S%.3f` + 8-char UUID suffix
- `u64 as u32` truncation replaced with `.filter(|&v| v <= u64::from(u32::MAX)).map(|v| v as u32)` — values exceeding u32 range are filtered out instead of silently truncated
- ffmpeg stdin/stdout race fixed with `tokio::join!`
- Unified MIME helpers + `ALLOWED_MEDIA_EXTS` allowlist

## Known follow-up

`ALLOWED_MEDIA_EXTS` includes `mp4`/`webm`/`pdf` but `audio_mime_from_ext` does not recognise them, so e.g. `foo.mp4` would pass `validate_ext` in `tool_media_transcribe` and then fail at `audio_mime_from_ext.ok_or(…)` with "Unsupported audio format". Options: drop `validate_ext` from `tool_media_transcribe` (the MIME lookup is the real gate) or split into separate `ALLOWED_IMAGE_EXTS` / `ALLOWED_AUDIO_EXTS`. Not blocking for this PR.

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and migrated all of `media.rs` to `ToolError`/`ToolResult` (plus `resolve_media_path`, `audio_mime_from_ext`), so every error site here must be re-expressed on rebase. The hardening (pre-read size limits, async fs, UUID filenames, ffmpeg deadlock fix via concurrent stdin-write + `wait_with_output`) is genuinely new and worth keeping. Three concrete issues:

1. `crates/librefang-runtime/src/tool_runner/media.rs` (`tool_speech_to_text`) — switching the extension fallback from `unwrap_or("mp3")` to `unwrap_or("")` + `validate_ext(&ext)?` regresses documented behavior. Main's comment is explicit that `speech_to_text` is permissive and the strict gate lives in `tool_media_transcribe`. After this change, extensionless audio or any ext not in `ALLOWED_MEDIA_EXTS` (`.opus`, `.aac`, `.amr`) that STT could previously transcribe now errors. Keep `speech_to_text` permissive; leave the strict gate in `tool_media_transcribe`.
2. `crates/librefang-runtime/src/tool_runner/media.rs` (`convert_to_ogg_opus`, stdin write) — the `tokio::join!` rewrite swallows the stdin write error with `let _ = stdin.write_all(...).await;`, where main propagated it. This contradicts the PR's own "replace silent errors with `?`" goal. Capture the write result inside the async block and surface it (the deadlock fix itself is correct — keep it).
3. `crates/librefang-runtime/src/tool_runner/media.rs` (`duration_secs`) — `u32::try_from(v.min(u32::MAX as u64)).ok()` always succeeds, so an absurd duration is silently clamped to ~4.29e9s rather than rejected. Prefer rejecting out-of-range durations. (Also: width/height `as u32` were left untruncated-fixed, inconsistent with the stated `u64 as u32` goal.)

leszek3737:
done

houko (approved):
Re-reviewed `061c4ee1` — all three prior items addressed:
- `tool_speech_to_text` is permissive again (no `validate_ext` on the STT path; the strict gate lives in `tool_media_transcribe` / `tool_media_describe` only).
- The `convert_to_ogg_opus` `tokio::join!` now captures the stdin-write result and propagates it via `?` instead of swallowing with `let _ = …`.
- Out-of-range `duration`/`width`/`height` no longer silently truncate (filter+map maps u64-overflow to `None`).

Approving. Two small notes (non-blocking):
- PR body still says "u32::try_from" — the diff uses `.filter(|&v| v <= u64::from(u32::MAX)).map(|v| v as u32)`. Same effect, but the description is inaccurate.
- `ALLOWED_MEDIA_EXTS` includes `mp4`/`webm`/`pdf` but `audio_mime_from_ext` does not, so e.g. `foo.mp4` would pass `validate_ext` in `tool_media_transcribe` and then fail at `audio_mime_from_ext.ok_or(...)` with "Unsupported audio format". Either drop `validate_ext` from `tool_media_transcribe` (the MIME lookup is the real gate) or split into separate `ALLOWED_IMAGE_EXTS` / `ALLOWED_AUDIO_EXTS`. Worth a follow-up but not blocking.

Issue #5699, opened 2026-05-24
feat(runtime-mcp): propagate sender peer/session context to MCP servers so they can authorize per-caller — currently call_tool is context-blind


nevgenov:
## Summary

`McpConnection::call_tool(&mut self, name, arguments)` in `crates/librefang-runtime-mcp/src/lib.rs:2061` has no parameter carrying the **calling user's identity** (the sender peer that drove the agent turn). MCP servers therefore only see whatever the agent decides to put into `arguments` — there is no kernel-attested signal of who is actually making the request.

For MCP servers that route per-user resources (e.g. an email/calendar adapter with one OAuth refresh token per peer, a multi-tenant CRM bridge, a per-user analytics tool), this means the only "who is calling" signal is what the agent self-reports in arguments. Currently, in a multi-user channel, a malicious or merely confused agent prompt can pass any `user_id` value and the MCP server will dutifully execute against it.

## Concrete scenario

A common multi-user pattern:
- One Telegram bot serving multiple users (`allowed_users = [...]`)
- One agent backing it, sessions are deterministic per `(channel, peer)`
- A per-user MCP adapter exposing `list_emails(user_id, ...)`, `send_email(user_id, ...)` etc., resolving `user_id` to a per-user OAuth refresh token, calling some downstream API

The agent, running in user A's session, currently controls the `user_id` argument freely. If a future prompt revision (or a prompt-injection from a malicious email body the user asked the agent to summarize) talks the model into `list_emails(user_id="<user-B>", ...)`, the adapter happily fetches user B's mailbox using user B's refresh token. **No kernel-enforced boundary**.

We've verified this is structural — `sender_user_id` is populated end-to-end inside `agent_loop` (used in `end_turn.rs` for memory hooks, see #5474 / PR #5647), but the value never reaches `call_tool` and therefore never reaches MCP servers. The only defense today is the agent's system prompt politely asking it to use the right `peer_id`, which is **a soft LLM guarantee, not a runtime check**.

## What we want

A way for MCP servers to receive — out-of-band from the `arguments` payload — the **kernel-attested** identity of the current caller:
- `sender_user_id` (the channel peer who drove this agent turn), and ideally
- `sender_chat_scope` / channel id (same shape used by #5227 chat_scope filter)

…so the MCP server can compare against any user-supplied `user_id` argument and refuse mismatches, or use the context directly (skipping the argument entirely).

[...]

PR #5763, opened 2026-05-26
fix(runtime): tool_runner shell — timeout clamp, streaming output, process group kill, Windows compat

leszek3737:
## Type
- [x] Bug fix

## Summary

Hardens `tool_runner/shell.rs` with 6 fixes: timeout clamping, Instant overflow protection, streaming output limits, process tree cleanup, kill_on_drop removal, Windows argv splitting.

## Changes

- User timeout clamped to policy max via `.min(policy_timeout)`
- `Instant::checked_add()` with 1h fallback prevents panic on overflow
- `wait_with_output()` replaced with streaming `stream_read_with_limit` — caps memory at `max_output_bytes`
- `process_group(0)` on Unix + `kill_process_tree()` for full process tree cleanup
- Removed misleading `kill_on_drop(true)` — explicit kill is now deterministic
- `windows_argv_split()` replaces `shlex::split` on Windows

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and migrated `tool_shell_exec` to `ToolResult = Result<String, ToolError>`. This PR is written against the old `Result<String, String>` contract, so it no longer compiles on `main`, and two concrete points:

1. `crates/librefang-runtime/src/tool_runner/shell.rs` — won't compile after rebase: the new `.ok_or("Missing 'command' parameter")?`, `return Err(format!(...))`, `.ok_or("Failed to capture stdout")?`, and `.map_err(|e| format!(...))` sites yield `String`/`&str`, which don't coerce into the `ToolError` return type. Re-express each as the matching variant (spawn/collect → `Upstream{message, source}`, parse → `InvalidParameter`, missing param → `MissingParameter`, interrupt/timeout → `upstream_msg`), and keep the migrated `#[cfg(test)] mod tests` block.
2. `crates/librefang-runtime/src/tool_runner/shell.rs` (after `kill_on_drop(true)` removed) — the two new fallible takes `child.stdout.take().ok_or(...)?` / `child.stderr.take().ok_or(...)?` run after spawn; if either `?` returns early, `child` is dropped without an explicit kill, leaking the just-spawned process. Keep `kill_on_drop(true)` as a backstop, or kill explicitly on the early-return paths.

The ideas (timeout clamp, streaming bounded output to avoid the pipe-buffer deadlock, process-tree kill, overflow guard) are sound and salvageable — `ExecPolicy.max_output_bytes` and `subprocess_sandbox::kill_process_tree` both exist. Minor: the truncation predicate changed `> max_output` → `>= max_output`, mislabeling output exactly at the cap as `[truncated]`.

leszek3737:
done

houko (dismissed):
Re-reviewed — fully migrated to `ToolError` with the spawn/parse/upstream/upstream_msg variants wired correctly, the migrated `#[cfg(test)] mod tests` is preserved, `kill_on_drop(true)` is kept as a backstop so the new `stdout.take()?` / `stderr.take()?` early-returns can't leak the spawned child, and `subprocess_sandbox::kill_process_tree` exists for the process-group path.

Approving. Two small follow-up suggestions (non-blocking):
- `windows_argv_split` (the naive quote-toggle splitter) doesn't handle escaped quotes (`\"`), backslash sequences, or empty quoted strings — a couple of round-trip tests covering `cmd "a b"` / `cmd \"escaped\"` would lock the contract.
- No new `#[tokio::test]` covers the streaming-drain / process-group-kill / timeout-clamp logic itself — the two preserved tests hit missing-param and unmatched-quotes only. Worth adding a thin integration test in a follow-up.

houko (changes requested):
The `windows_argv_split` rewrite in `81e012ab` looks correct — proper Windows CRT backslash/quote handling (`\"`, `\\`, the `2n` / `2n+1` backslash-before-quote rules), empty-quoted-string support (`""` → empty arg), `None` on unterminated quotes, clean `#[cfg(any(windows, test))]` gating, and 12 round-trip tests including the tricky multi-backslash cases. That resolves my earlier splitter follow-up — thanks.

**Blocking: the branch has picked up a bad merge.** HEAD (`ae65c493`) is **+952 / −10,487 across 106 files**, but this PR should only touch `crates/librefang-runtime/src/tool_runner/shell.rs` (+ its test file). The conflict resolutions in `c52cef5e` and `1c5f2980` ("chore: sync docs, SDK, and config with main after merge") discarded `main`'s newer work instead of integrating it, and the later main merge did not restore it. Concretely, vs `origin/main` the branch **deletes 35 files**, including:

- the entire `sdk/rust/librefang-sidecar-telegram/` crate (adapter.rs, dispatcher.rs, translator.rs, api/client.rs, format/*, schema.rs, …) and its docs (`docs/architecture/rust-{sidecar-sdk,telegram-sidecar}.md`, the `docs/src/app/**` pages) and `conformance/sidecar/corpus/**` — all merged into main via #5831,
- `scripts/run-xtask.sh`,

and reverts (not deletes) more merged work: `crates/librefang-runtime-media/src/media_understanding.rs` (1392 → 976 lines), the ToolError-era `tool_runner` files (`agent.rs`, `cron.rs`, `process.rs`, `sandbox.rs`, `dispatch.rs`, `hand.rs`, `web_legacy.rs`, `memory.rs`), `librefang-channels/src/bridge.rs`, `kernel/src/workflow.rs`, `xtask/src/dev.rs`, plus `CLAUDE.md` (drops the "Prose wrapping" section), `CHANGELOG.md`, `justfile`, `openapi.json`. Merging as-is would re-delete all of that from `main`.

GitHub also currently reports `CONFLICTING / DIRTY`, so it needs a re-merge regardless.

**Fix:** redo the merge so the diff against `main` is only the `shell.rs` change. Easiest path — cherry-pick the two real commits onto a fresh branch from current `origin/main`:

```
git fetch origin
git switch -c fix/tool-runner-shell-clean origin/main
git cherry-pick 7251f5f9 81e012ab   # resolve conflicts keeping only the shell.rs edits
# verify the diff is just shell.rs:
git diff origin/main --stat
```

then force-push that over the branch (or open a fresh PR). The `shell.rs` content itself is fine — only the merge metadata needs fixing.

Note: the tip merge `ae65c493` is one of mine — I have maintainer-edit on this branch, so if you'd prefer I can redo the clean rebase and force-push it for you. Just say the word.

houko:
Update: I started the clean rebase onto current `main` and hit a second, deeper issue, so I'm leaving this for you rather than force-pushing a guess.

Beyond the bad merge, `shell.rs` now has a genuine **semantic** conflict with `main`. While this PR was open, #5774 / #5818 independently reworked the same `run_shell` exec body: `main` now drains pipes via `child.wait_with_output()` inside a `select!` against a 100 ms interrupt tick + deadline (with `kill_on_drop(true)`), and registers/marks the child in `process_registry`. That overlaps heavily with this PR's redesign (spawned `stream_read_with_limit` tasks, manual `checked_add` deadline, process-group kill, `windows_argv_split`).

So a clean rebase isn't mechanical — it needs the two designs reconciled, e.g.:
- keep `main`'s `process_registry` register / `mark_finished` calls,
- decide pipe handling: `main`'s `wait_with_output` already fixes the >8KB pipe-buffer deadlock, so the spawned-reader approach is only needed if you specifically want the per-stream `max_output` cap *during* read (vs `main`'s after-the-fact truncation),
- the genuinely additive parts worth keeping are the `checked_add` timeout clamp (avoids overflow panic), the process-**group** kill (main's `kill_on_drop` only reaps the direct child, not grandchildren), and `windows_argv_split`.

Since this is your design call on security-sensitive subprocess handling, I didn't want to force-push my interpretation. Recommend: branch fresh from `origin/main`, port just the additive pieces above on top of main's current `run_shell`, and confirm the diff is `shell.rs`-only. Happy to review again once it's rebased.

houko:
Thanks — the core shell hardening is solid: the timeout clamp (`input_timeout.min(policy_timeout)`), concurrent pipe-draining via spawned `stream_read_with_limit` tasks (fixes the real pipe-buffer deadlock), and `process_group(0)` paired with `kill_process_tree`'s group-leader `kill(-pgid)` (kills orphans) are all correct, and the `windows_argv_split` CRT parser is well-tested.

Two things to address before merge:

**1. `process_registry` integration was removed (functional regression).**
`shell.rs:20-28` drops the `process_registry` + `session_id` params and deletes the `reg.register()` / `reg.mark_finished()` calls.
`crates/librefang-api/src/routes/tools_sessions.rs:302` passes `process_registry()` into dispatch specifically so shell-spawned PIDs surface via the `tools/sessions` API — after this change that API reports no shell processes.
If the removal is intentional, please call it out explicitly and justify it; otherwise restore the plumbing.

**2. Stale base — the merge reverts intervening `main` work.**
The branch's merge-base is far behind current `main`, so the net diff reverts changes across `agent.rs`/`media.rs`/`process.rs`/`hand.rs`/`task.rs`, plus the #5699 MCP `CallerContext` propagation and the skill-registry `allowed_skills` ACL gate (`dispatch.rs`).
CI reflects this (Quality + Ubuntu shard 2/4 + detect-secrets fail).
Please rebase onto current `main` and re-resolve — squash-merging through the conflicts would reintroduce removed code and drop the skill ACL check.

Minor: the truncation message lost the total-bytes count (`"exceeded {max_output} bytes"` vs the old `"{} total bytes"`) — small observability loss.

leszek3737:
done

houko (changes requested):
The merge is clean now — single-file diff, mergeable, CI green. The earlier blocker (deleted SDK crate / reverted merged work) is resolved. Thanks. One functional regression remains before this can land.

**Timeout clamp inverts the override contract** (`crates/librefang-runtime/src/tool_runner/shell.rs:33-35`):
```rust
let policy_timeout = exec_policy.map(|p| p.timeout_secs).unwrap_or(30);
let input_timeout = input["timeout_seconds"].as_u64().unwrap_or(policy_timeout);
let timeout_secs = input_timeout.min(policy_timeout);
```
`policy.timeout_secs` is the **default** (30s), not a maximum — `ExecPolicy` has no separate ceiling field (see `crates/librefang-types/src/config/types.rs:2014`, "Max execution timeout in seconds. Default: 30"). With `.min(policy_timeout)`, a caller-supplied `timeout_seconds` can now only *lower* the timeout, never raise it: `{"timeout_seconds": 300}` on the default policy silently runs with 30s. Pre-PR (`as_u64().unwrap_or(policy_timeout)`) honored upward overrides. Please either drop the `.min()` and keep the override, or add an explicit `max_timeout_secs` field and clamp against that instead of the default.

Two smaller points:

- The PR body says `wait_with_output()` was replaced by `stream_read_with_limit` and `kill_on_drop` was removed, but the head still calls `wait_with_output()` (`shell.rs:154`) and keeps `kill_on_drop(true)` (`shell.rs:127`). Keeping `kill_on_drop` is the right call (backstops the explicit-kill paths); the body just needs correcting — output is still fully buffered then truncated (`shell.rs:193-210`), so it does not cap peak memory.
- The clamp resolution is pure, platform-independent arithmetic. A short `#[test]` pinning `timeout_secs` resolution (override-up, override-down, default) would have caught the above and locks the contract per the crate's testing rule. Worth adding with the fix.

Process-group / Windows paths look correct: `process_group(0)` is `tokio`'s stable method (used identically in `process_manager.rs:106`), `#[cfg(unix)]`-gated so Windows routes to the taskkill path, and `kill_process_tree` is internally gated on `is_process_group_leader` so grace=0 is safe.

houko (dismissed):
The merge is clean now — single-file diff, mergeable, CI green. The earlier blocker (deleted SDK crate / reverted merged work) is resolved. Thanks. One functional regression remains before this can land.

**Timeout clamp inverts the override contract** (`crates/librefang-runtime/src/tool_runner/shell.rs:33-35`):
```rust
let policy_timeout = exec_policy.map(|p| p.timeout_secs).unwrap_or(30);
let input_timeout = input["timeout_seconds"].as_u64().unwrap_or(policy_timeout);
let timeout_secs = input_timeout.min(policy_timeout);
```
`policy.timeout_secs` is the **default** (30s), not a maximum — `ExecPolicy` has no separate ceiling field (see `crates/librefang-types/src/config/types.rs:2014`, "Max execution timeout in seconds. Default: 30"). With `.min(policy_timeout)`, a caller-supplied `timeout_seconds` can now only *lower* the timeout, never raise it: `{"timeout_seconds": 300}` on the default policy silently runs with 30s. Pre-PR (`as_u64().unwrap_or(policy_timeout)`) honored upward overrides. Please either drop the `.min()` and keep the override, or add an explicit `max_timeout_secs` field and clamp against that instead of the default.

Two smaller points:

- The PR body says `wait_with_output()` was replaced by `stream_read_with_limit` and `kill_on_drop` was removed, but the head still calls `wait_with_output()` (`shell.rs:154`) and keeps `kill_on_drop(true)` (`shell.rs:127`). Keeping `kill_on_drop` is the right call (backstops the explicit-kill paths); the body just needs correcting — output is still fully buffered then truncated (`shell.rs:193-210`), so it does not cap peak memory.
- The clamp resolution is pure, platform-independent arithmetic. A short `#[test]` pinning `timeout_secs` resolution (override-up, override-down, default) would have caught the above and locks the contract per the crate's testing rule. Worth adding with the fix.

Process-group / Windows paths look correct: `process_group(0)` is `tokio`'s stable method (used identically in `process_manager.rs:106`), `#[cfg(unix)]`-gated so Windows routes to the taskkill path, and `kill_process_tree` is internally gated on `is_process_group_leader` so grace=0 is safe.

leszek3737:
done

houko (dismissed):
The timeout-clamp blocker is fixed, and fixed the right way — thanks for adding an explicit ceiling rather than just dropping `.min()`.

`crates/librefang-runtime/src/tool_runner/shell.rs` now resolves against a real `max_timeout_secs`:

```rust
let default_timeout = exec_policy.map(|p| p.timeout_secs).unwrap_or(30);
let max_timeout = exec_policy.map(|p| p.max_timeout_secs).unwrap_or(300);
let requested_timeout = input["timeout_seconds"].as_u64().unwrap_or(default_timeout);
let timeout_secs = requested_timeout.min(max_timeout);
```

So `{"timeout_seconds": 300}` is now honored up to the ceiling instead of silently collapsing to 30, and an over-large request is capped at the ceiling rather than the default.

Config plumbing is complete — `crates/librefang-types/src/config/types.rs`: `#[serde(default = "default_max_timeout_secs")]` at `:2015`, `skip_serializing_if = "is_default_max_timeout"` at `:2016`, the `Default` impl entry `max_timeout_secs: 300` at `:2053`, and the `default_max_timeout_secs()` / `is_default_max_timeout()` helpers at `:2061-2067`. Deserializing an existing `config.toml` without the field is safe (it gets 300), and all 12 CI checks are green, which independently confirms the build and plumbing.

The new `timeout_override_is_clamped_to_max` test covers override-honored and override-capped and is platform-independent. Good.

Approving. Two non-blocking doc nits to fix before merge — neither affects behavior:

1. The `max_timeout_secs` doc-comment contradicts its own default. `crates/librefang-types/src/config/types.rs:2014` says "Default: 30", but `default_max_timeout_secs()` returns 300 (`:2062`) and the `Default` impl sets 300 (`:2053`). Looks like a copy-paste from the `timeout_secs` doc above it. Change the comment to "Default: 300" so the generated config reference is correct.

2. The PR body still says `wait_with_output()` was replaced by `stream_read_with_limit` and `kill_on_drop` was removed, but the head still calls `wait_with_output()` (`shell.rs:154`) and keeps `kill_on_drop(true)` (`shell.rs:127`) — both correct, keep them. Output is still fully buffered then truncated, so it does not cap peak memory. Please correct the body so the changelog isn't misleading; no code change needed.

houko (approved):
The timeout regression is fixed in `3ad8d3db`, and fixed the way I'd hoped — thanks.

`resolve_timeout()` (`crates/librefang-runtime/src/tool_runner/shell.rs:20-26`) drops the `.min(policy_timeout)` clamp and resolves the caller value over the policy default:

```rust
fn resolve_timeout(
    input: &serde_json::Value,
    exec_policy: Option<&librefang_types::config::ExecPolicy>,
) -> u64 {
    let policy_timeout = exec_policy.map(|p| p.timeout_secs).unwrap_or(30);
    input["timeout_seconds"].as_u64().unwrap_or(policy_timeout)
}
```

So `{"timeout_seconds": 300}` on the default 30s policy now runs with 300s again instead of silently collapsing to 30. Dropping the clamp is the right call given `ExecPolicy` has no separate ceiling field — `timeout_secs` is the default, not a max. Extracting a pure helper made it testable, and the five new `#[test]`s (`shell.rs:416-448`) pin the contract: default → 30, override-up → 300 (the regression guard I asked for), override-down → 5, no-policy → 30, no-policy-with-input → 120. All platform-independent.

The rest holds up at this head: the `checked_add` deadline overflow guard (`shell.rs:153-159`), `#[cfg(unix)] process_group(0)` + `kill_process_tree` on both the interrupt and timeout paths, the `windows_argv_split` CRT parser with its 12 round-trip tests, and `max_output` now sourced from `exec_policy.max_output_bytes`. CI is green.

Approving. One non-blocking cleanup before merge — the PR description now describes the opposite of the code on three points:

1. "User timeout clamped to policy max via `.min(policy_timeout)`" — the `.min()` was removed in this very fix; that line now describes the old bug.
2. "`wait_with_output()` replaced with streaming `stream_read_with_limit` — caps memory at `max_output_bytes`" — the head still calls `child.wait_with_output()` (`shell.rs:160`) and still buffers fully before truncating (`shell.rs:199-216`); there is no `stream_read_with_limit`. Output is truncated for display but peak memory is not capped.
3. "Removed misleading `kill_on_drop(true)`" — `kill_on_drop(true)` is still present (`shell.rs:133`) and correctly kept as the drop-path backstop.

Please update those three bullets to match the diff, then this is good to land. No code change needed.

houko (dismissed):
The timeout regression is fixed in `3ad8d3db`, and fixed the way I'd hoped — thanks.

`resolve_timeout()` (`crates/librefang-runtime/src/tool_runner/shell.rs:20-26`) drops the `.min(policy_timeout)` clamp and resolves the caller value over the policy default:

```rust
fn resolve_timeout(
    input: &serde_json::Value,
    exec_policy: Option<&librefang_types::config::ExecPolicy>,
) -> u64 {
    let policy_timeout = exec_policy.map(|p| p.timeout_secs).unwrap_or(30);
    input["timeout_seconds"].as_u64().unwrap_or(policy_timeout)
}
```

So `{"timeout_seconds": 300}` on the default 30s policy now runs with 300s again instead of silently collapsing to 30. Dropping the clamp is the right call given `ExecPolicy` has no separate ceiling field — `timeout_secs` is the default, not a max. Extracting a pure helper made it testable, and the five new `#[test]`s (`shell.rs:416-448`) pin the contract: default → 30, override-up → 300 (the regression guard I asked for), override-down → 5, no-policy → 30, no-policy-with-input → 120. All platform-independent.

The rest holds up at this head: the `checked_add` deadline overflow guard (`shell.rs:153-159`), `#[cfg(unix)] process_group(0)` + `kill_process_tree` on both the interrupt and timeout paths, the `windows_argv_split` CRT parser with its 12 round-trip tests, and `max_output` now sourced from `exec_policy.max_output_bytes`. CI is green.

Approving. One non-blocking cleanup before merge — the PR description now describes the opposite of the code on three points:

1. "User timeout clamped to policy max via `.min(policy_timeout)`" — the `.min()` was removed in this very fix; that line now describes the old bug.
2. "`wait_with_output()` replaced with streaming `stream_read_with_limit` — caps memory at `max_output_bytes`" — the head still calls `child.wait_with_output()` (`shell.rs:160`) and still buffers fully before truncating (`shell.rs:199-216`); there is no `stream_read_with_limit`. Output is truncated for display but peak memory is not capped.
3. "Removed misleading `kill_on_drop(true)`" — `kill_on_drop(true)` is still present (`shell.rs:133`) and correctly kept as the drop-path backstop.

Please update those three bullets to match the diff, then this is good to land. No code change needed.

PR #5764, opened 2026-05-26
fix(runtime): tool_runner web_legacy — SSRF protection, streaming body limit, unified UA, status check

leszek3737:
## Type
- [x] Bug fix

## Summary

Hardens `tool_runner/web_legacy.rs` with 7 fixes: SSRF protection, streaming body limits, unified User-Agent, HTTP status checking, search result capping.

## Changes

- `check_ssrf()` called before any HTTP request (reuses existing full SSRF check)
- `read_body_limited()` streams via `bytes_stream()` with 10MB cap — works without Content-Length
- `LEGACY_UA` constant with version from `CARGO_PKG_VERSION` — unified across fetch/search
- Non-2xx responses return error with status code + body preview
- `max_results` clamped to `MAX_SEARCH_RESULTS` (20)
- `read_body_limited()` used in both fetch and search paths
- `safe_truncate_str()` for char-boundary-safe string truncation

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and migrated `tool_web_fetch_legacy` / `tool_web_search_legacy` to `ToolResult`. This PR reverts them to `Result<String, String>`, so it no longer compiles on `main`, plus a behavior change:

1. `crates/librefang-runtime/src/tool_runner/web_legacy.rs` — won't compile after rebase: `.ok_or("Missing 'url' parameter")?`, `.map_err(|e| format!(...))`, `check_ssrf(url, &[])?`, and `read_body_limited(resp).await?` all yield `String` errors that don't coerce to `ToolError`. Make `read_body_limited` return `Result<String, ToolError>` (or convert at the call site), route SSRF/transport errors through `fetch_err`/`upstream_msg`/`InvalidParameter`, and keep the migrated `mod tests`.
2. `crates/librefang-runtime/src/tool_runner/web_legacy.rs` (`read_body_limited` non-2xx branch) — legacy fetch previously returned `Ok("HTTP {status}\n\n{body}")` for any status, so an agent fetching a 403/404/500 page still received the body. Turning every non-2xx into `Err("HTTP {status}: {preview}")` with only a 500-char preview drops the body on the common 4xx-with-useful-body case. Either keep returning the body as `Ok` for non-2xx, or, if surfacing status as an error is intended, document the behavior change.

The SSRF check and the streaming body cap are genuinely valuable (the old "no SSRF protection" comment is real); `check_ssrf` exists and `status` is captured before `resp` is consumed, so no use-after-move. Needs re-authoring on the `ToolError` shape plus a decision on the non-2xx behavior.

leszek3737:
done

houko (approved):
Re-reviewed — migrated to `ToolError` correctly (`MissingParameter`, `InvalidParameter` for SSRF reasons, `Upstream` via `fetch_err`, `upstream_msg` for the cap), and non-2xx now returns `Ok("HTTP {status}\n\n{body}")` again so agents fetching 403/404/500 still receive the body (the prior re-review concern).

Approving. One follow-up worth noting (non-blocking): `check_ssrf(url, &[])` resolves DNS once, then the actual fetch is done with a fresh `reqwest::Client::get(url).send()` that performs its own DNS lookup — so a hostname could resolve to a public IP for the SSRF check and to a private/metadata IP for the subsequent fetch (DNS rebinding window). The canonical `WebToolsContext` path in `web_fetch.rs:88-101` solves this by using `pinned_client(resolution)`; consider reusing it here in a follow-up, or document the legacy fetch as best-effort SSRF only.

PR #5766, opened 2026-05-26
fix(runtime): tool_runner canvas — XSS escape, whitelist parser, data: URI block, size limit

leszek3737:
## Type
- [x] Bug fix
- [x] Security

## Summary

Rewrites `sanitize_canvas_html` in `tool_runner/canvas.rs` from reject-on-danger to strip-and-allowlist strategy. Adds HTML escaping, tag whitelist, data: URI filtering, and size limits on all output types.

## Changes

- `escape_html()` escapes `<>&"'` in title before embedding
- Whitelist-based parser: only `ALLOWED_TAGS` (50 safe tags) pass through; `on*`/`style` attrs stripped
- `href`/`src` URL-validated — blocks `javascript:`, `vbscript:`; permits only safe `data:image/*`
- `create_dir_all` error propagated with `?` instead of `let _ =`
- Size limit checked on sanitized output AND final `full_html` document

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Two concrete logic bugs in the new `sanitize_canvas_html` (independent of the merge conflict — `sanitize_canvas_html` stays `Result<String, String>` post-#5737, so the signature is compatible):

1. `crates/librefang-runtime/src/tool_runner/canvas.rs` (the `&`-handling branch) — dead/always-false condition. `after` always begins at a `&` byte (we're inside the `bytes[pos] == b'&'` arm), so `after.starts_with('&')` is always true and `!after.starts_with("&")` is always false; the entity-preservation branch can never execute. Every `&` — including those in valid entities like `&amp;`/`&lt;` — is unconditionally rewritten to `&amp;`, so existing entities get double-escaped (`&amp;` → `&amp;amp;`) and render literally. Rewrite the condition to actually test "this `&` starts a valid entity" before preserving vs escaping, and add a `&amp;`/`&#39;` round-trip test.
2. `crates/librefang-runtime/src/tool_runner/canvas.rs` (text passthrough: `result.push(bytes[pos] as char)`) — UTF-8 corruption. Casting a single `u8` to `char` only works for ASCII; any multibyte character in the body (non-Latin text, em-dash, emoji) is split into bytes each reinterpreted as Latin-1, producing mojibake in the saved canvas. Walk `char_indices()` for text runs, or push the original `&str` slice between tags rather than byte-by-byte.
3. `crates/librefang-runtime/src/tool_runner/canvas.rs` (removed upfront `if html.len() > max_bytes`) — minor DoS regression: the old code rejected oversized input in O(1) before work; the new code fully parses and allocates before the post-sanitize check fires. Keep a cheap upfront `html.len()` guard in addition to the post-sanitize check.

(Existing tests pass only by luck — `test_sanitize_canvas_basic_html` is pure ASCII, and the reject-tests are caught by the early `dangerous_tags` check.) The sanitizer rewrite is meaningful hardening, but fix these first.

leszek3737:
done

houko (approved):
Re-reviewed `bb132fae` — all three prior bugs fixed with dedicated tests:
- The entity branch now detects well-formed named/numeric/hex entities and passes them through verbatim; new `test_sanitize_canvas_preserves_entities` / `_numeric_entities` / `_escapes_bare_ampersand` / `_escapes_malformed_entity` cover both arms (`tests/policy.rs:666-693`).
- The text passthrough walks `[start..pos]` byte slices anchored at ASCII delimiters and pushes `&html[start..pos]` as `&str`, so multibyte chars survive. `test_sanitize_canvas_utf8_passthrough` (`tests/policy.rs:696-700`) round-trips `"Zażółć gęślą jaźń — 日本語 🦀"`.
- Upfront `if html.len() > max_bytes` reject is kept before any allocation (`canvas.rs:205-211`), with `test_sanitize_canvas_upfront_size_limit` pinning the contract.

Approving. Small note (non-blocking): the dangerous-tag substring scan widened from `<script`/`<iframe` to include `<form`/`<input`/`<button`/`<meta`/`<link`/`<base` rejects any body containing those literal substrings (e.g. a `<basement>` mention). The whitelist parser would reject those tags anyway; consider trimming the upfront substring list to the genuinely script-executing ones and let the whitelist handle the rest.

PR #5767, opened 2026-05-26
fix(runtime): tool_runner knowledge — confidence clamp, input validation, result limits, property bounds

leszek3737:
## Type
- [x] Bug fix

## Summary

Hardens `tool_runner/knowledge.rs` with 7 fixes: confidence range check, empty input rejection, result limiting, property bounds, conditional lowercase, pre-allocated output.

## Changes

- Confidence clamped to 0.0–1.0 via `.clamp(0.0, 1.0)`
- Query strings trimmed and empty values rejected
- `DEFAULT_RESULT_LIMIT = 50`, configurable via `limit` param, capped at 2x default
- Properties bounded: max 50 keys, string values capped at 4096 chars
- `to_lowercase()` called once into binding — `Custom` variant preserves original case
- `String::with_capacity` + `write!` replaces `format!` + `push_str` in loop

## Attribution

## Testing

## Security
- [x] No new unsafe code

### DIFF
--- a/crates/librefang-runtime/src/tool_runner/knowledge.rs
+++ b/crates/librefang-runtime/src/tool_runner/knowledge.rs
@@ -11,9 +11,13 @@ use super::require_kernel_typed;
 use crate::kernel_handle::prelude::*;
 use std::sync::Arc;
 
+const MAX_PROPERTY_COUNT: usize = 50;
+const MAX_PROPERTY_VALUE_LEN: usize = 4096;
+
 fn parse_entity_type(s: &str) -> librefang_types::memory::EntityType {
     use librefang_types::memory::EntityType;
-    match s.to_lowercase().as_str() {
+    let lower = s.to_lowercase();
+    match lower.as_str() {
         "person" => EntityType::Person,
         "organization" | "org" => EntityType::Organization,
         "project" => EntityType::Project,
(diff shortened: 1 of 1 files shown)


houko (dismissed):
Re-reviewed — the original verdict stands: the confidence clamp (0.0..=1.0), empty-name/query rejection, requiring at least one of source/target/relation (DoS guard against unbounded traversal), the `_ => Custom(s.to_string())` case-preservation fix, and the client-side `.take(limit)` on results are all correct against `knowledge.rs` on current main. None of the mock `KernelHandle` impls exercises the empty-pattern path, so no test breaks.

Approving as-is. **CONFLICTING/DIRTY against current main** — please rebase onto `dd1de430` (or later) before it can land.

houko:
Reviewed — this is the cleanest of the tool_runner hardening set. `confidence.clamp(0.0, 1.0)`, the property cap (`take(50)` + per-value `take(4096)`), name trim/reject-empty, mandatory `source|target|relation` (blocks full-graph-scan DoS), `max_depth.min(10)` and `limit.min(100)` are all complete — I couldn't find a DoS surface. Preserving original case for `Custom(...)` in `parse_entity_type`/`parse_relation_type` is a nice improvement.

Just one LOW nit (optional, worth folding in while you rebase): the output header reports `Found {matches.len()} match(es)` (full count) but then prints only `.take(limit)` rows — slightly misleading.

Branch is CONFLICTING — needs a rebase onto current `main`, then it's good to go.

PR #5770, opened 2026-05-26
fix(runtime): tool_runner memory — truncation, pagination, key validation

leszek3737:
## Type
- [x] Bug fix

## Summary

Hardens `tool_runner/memory.rs` with output truncation, list pagination, and key validation.

## Changes

- `truncate_output()` caps recall output at 64 KiB with UTF-8-safe boundary and `... [truncated]` suffix
- `tool_memory_list` accepts `limit`/ `offset` params (default limit=100), shows total + next-offset hint
- `validate_key()` enforces max 256 chars, `[a-zA-Z0-9_.-]` charset — applied in store and recall
- Ripple: dispatch.rs passes input to memory_list, definitions.rs updated schema

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** One blocking data-compatibility bug, plus an in-series collision to sequence:

1. `crates/librefang-runtime/src/tool_runner/memory.rs` (`validate_key` applied in `tool_memory_recall`) — backward-incompatible read break: keys stored before this PR that contain characters outside `[a-zA-Z0-9_.-]` (notably `:`) become unrecallable. Colon-style namespacing is an established convention — this repo's own isolation test used `auto::fact` (this PR even changes that fixture to `auto.fact` at `tool_runner_memory_isolation.rs:439`, confirming `:` keys were valid). Applying `validate_key` on the **recall** path means an agent that previously stored `pref:lang` can no longer read it back; the data is silently stranded. Either (a) validate only on `store`, or (b) widen the charset to include `:`. Validating writes is fine; rejecting reads of already-stored keys is the regression.

2. Sequencing: this PR semantically collides with #5769 and #5771 in the same series — #5769 also edits the same `enforce_memory_acl(...)` call lines in `memory.rs` and the same `tool_runner_memory_isolation.rs` test, and #5771 ships a stale `memory_list` definition (empty `properties`) that would overwrite this PR's new pagination schema. Whichever lands second will conflict; please reconcile the `memory_list` schema and the `memory.rs` ACL-call lines across the three.

The 64 KiB UTF-8-safe truncation and the `limit`/`offset` pagination (stable since the backend sorts `ORDER BY key`) are fine.

leszek3737:
done

houko (changes requested):
Re-reviewed the new commits. Thanks — `validate_key` is now correctly off the recall path (store-only), so pre-existing `:`-namespaced keys are recallable again. But the rebase merge corrupted `definitions.rs`:

- `crates/librefang-runtime/src/tool_runner/definitions.rs:259-455` — a ~205-line block of old-style `ToolDefinition { name: "…".to_string(), … }` entries was re-introduced on top of main's current `tool_name::*`-based definitions, both inside the same memoized `builtin_tool_definitions()` `tools.extend(vec![…])`. As a result every agent now receives **13 duplicate tool definitions** (`web_search`, `shell_exec`, `notify_owner`, `agent_send`/`spawn`/`list`/`kill`, `memory_store`/`recall`/`list`, `wiki_get`/`search`/`write`), which breaks the no-duplicate / deterministic-prompt-ordering invariant. Worse, the two `memory_list` entries disagree: the literal one (`definitions.rs:395`) carries your new `limit`/`offset` pagination schema, while the canonical `tool_name::MEMORY_LIST` one (`definitions.rs:596`) still has the old empty-`properties` schema — so your headline pagination schema is shadowed by the stale duplicate.

Fix: drop the entire re-added literal block (259-455) and apply the pagination schema to the canonical `MEMORY_LIST` definition. After fixing, a quick guard is to assert `builtin_tool_definitions()` has no duplicate `name`s.

leszek3737:
done

houko (approved):
Re-reviewed — both prior blockers resolved:
- The duplicate ~205-line literal `ToolDefinition` block in `definitions.rs` is gone. `builtin_tool_definitions()` is back to 88 entries, one per tool name, and the new `tests::no_duplicate_tool_names` guard pins the invariant.
- The `memory_list` schema now lives at a single canonical entry (`definitions.rs:391`) with the new `limit` / `offset` pagination properties (no shadowed empty-schema duplicate).

Approving.

PR #5773, opened 2026-05-26
fix(runtime): tool_runner cron — sender_id override, TOCTOU reduction, HashSet lookup, empty job_id rejected

leszek3737:
## Type
- [x] Bug fix
- [x] Security

## Summary

Hardens `tool_runner/cron.rs`: prevents peer_id spoofing, reduces TOCTOU window, improves cancel performance, rejects empty job_id.

## Changes

- `tool_cron_create` always overrides `peer_id` with authenticated `sender_id` — caller cannot inject arbitrary peer_id
- Cancel validation reordered — `job_id` extraction and empty-check happen before kernel call
- Ownership check uses `HashSet<&str>` + `.contains()` instead of linear `.any()`
- Empty string rejected with `ToolError::InvalidParameter`
- New unit test covers empty job_id rejection path

## Attribution

## Testing
- [x] `cargo test -p librefang-runtime --lib` — 1796 passed, 0 failed

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** The TOCTOU-reorder, empty-`job_id` rejection, and `HashSet` lookup are all correct and merge cleanly. One issue: the headline security claim is not implemented.

1. `crates/librefang-runtime/src/tool_runner/cron.rs` (`tool_cron_create`) — the PR body claims it "always overrides peer_id with authenticated sender_id — caller cannot inject arbitrary peer_id," but the code does the opposite: when the caller supplies `peer_id` it is preserved (`if obj.contains_key("peer_id") { /* preserve explicit peer_id */ }`), and the `else` branch's `obj.remove("peer_id")` only runs when the key is already absent (a no-op). Net behavior is identical to the original (sender_id only fills a *missing* peer_id). Either implement the override — `obj.insert("peer_id", sender_id)` unconditionally when an authenticated sender_id exists — or drop the peer_id-spoofing claim from the PR body and keep this as a pure refactor.

leszek3737:
done

houko (approved):
Re-reviewed — the peer_id claim now matches the code: `tool_cron_create` unconditionally overrides with the authenticated sender_id (`obj.insert("peer_id", pid)` when a non-empty sender_id is present, strip otherwise), and the test was renamed to `test_cron_create_overrides_explicit_peer_id_with_sender` and asserts the override. The TOCTOU-reorder, empty-`job_id` rejection, and `HashSet` lookup are all good.

Approving.

PR #5774, opened 2026-05-26
fix(runtime): tool_runner dispatch — mutex split, fallback ACL, ACP args, spill wiring, snapshot ordering

leszek3737:
## Type
- [x] Bug fix
- [x] Refactor / Performance

## Summary

Fixes 8 issues in `tool_runner/dispatch.rs`: mutex contention, fallback skill ACL bypass, ACP args, integer truncation, approval ordering, spill/config wiring, process registry, snapshot timing.

## Changes

- MCP mutex split: server-name resolution (short lock, drop) from mutable call — reduces contention
- Fallback skill dispatch checks `allowed_skills` ACL before execution
- ACP `shell_exec` merges `input["args"]` into command string
- `as u32` truncation replaced with `u32::try_from(v).ok()`
- `__args_truncated` check moved before approval gate
- `spill_threshold_bytes`/`max_artifact_bytes` wired from config (was hardcoded 0)
- `process_registry` passed through to `tool_shell_exec`
- `maybe_snapshot()` moved before ACP branch — editor writes also checkpointed

## Attribution

## Testing

## Security
- [x] No new unsafe code

### DIFF
--- a/crates/librefang-api/src/routes/network.rs
+++ b/crates/librefang-api/src/routes/network.rs
@@ -1525,6 +1525,8 @@ pub async fn mcp_http(
             None, // session_id (MCP HTTP is not tied to a live session)
             None, // dangerous_command_checker (no session-scoped checker here)
             None, // available_tools (lazy-load pool not applicable to MCP bridge)
+            cfg.tool_results.spill_threshold_bytes,
+            cfg.tool_results.max_artifact_bytes,
         )
         .await;
 
--- a/crates/librefang-api/src/routes/tools_sessions.rs
+++ b/crates/librefang-api/src/routes/tools_sessions.rs
@@ -275,6 +275,8 @@ pub async fn invoke_tool(
         None, // session_id
         None, // dangerous_command_checker — session-scoped, not meaningful here
         None, // available_tools — lazy-load pool not applicable to REST bridge
+        cfg.tool_results.spill_threshold_bytes,
+        cfg.tool_results.max_artifact_bytes,
     )
     .await;
 
--- a/crates/librefang-runtime/src/agent_loop/tool_call.rs
+++ b/crates/librefang-runtime/src/agent_loop/tool_call.rs
@@ -596,6 +596,7 @@ pub(super) async fn execute_single_tool_call_inner(
     }
 
     let effective_exec_policy = ctx.manifest.exec_policy.as_ref();
+    let tr_cfg = ctx.opts.tool_results_config.clone().unwrap_or_default();
     let tool_timeout = ctx.kernel.as_ref().map_or(TOOL_TIMEOUT_SECS, |k| {
         k.tool_timeout_secs_for(&tool_call.name)
     });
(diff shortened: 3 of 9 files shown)


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and migrated `tool_shell_exec` to `ToolResult`, so `shell.rs` here needs a genuine rebase. The valuable pieces survive and are worth keeping: the fallback-skill `allowed_skills` ACL check (mirrors `skill.rs:297`, closes a real gap), the spill-threshold wiring, the `maybe_snapshot` move (the ACP write branch returns early so editor writes were never checkpointed), the `u32::try_from`, and the `__args_truncated` reorder. Three issues:

1. `crates/librefang-runtime/src/tool_runner/dispatch.rs:810` (ACP `shell_exec` branch) — merging `input["args"]` into the `sh -c` command string is unmotivated divergence. `shell_exec`'s schema (`definitions.rs:177`) declares only `command` and `timeout_seconds`; the local-exec path (`shell.rs`) never reads `input["args"]`. So the same call runs `command` only locally but `command + space-joined args` over ACP — and the join is unquoted (`push(' '); push_str(s)`), so an `args` entry with spaces/metacharacters is re-split by the shell. Drop the args merge from the ACP path.
2. `crates/librefang-runtime/src/tool_runner/shell.rs` — `_process_registry` is added to `tool_shell_exec`'s signature (and threaded through dispatch + all test call sites) but is unused (underscore-bound). The PR lists "process_registry passed through" as a fix, but it produces no behavior. Either wire it up (register the spawned child) or drop the parameter.
3. `crates/librefang-runtime/src/tool_runner/dispatch.rs:1238` (MCP mutex split) — the split only shortens the lock for the microsecond server-name resolution; the long-pole `conn.call_tool_with_caller(...).await` still runs while holding `mcp_conns.lock().await`, so MCP tool calls remain fully serialized. Not a correctness bug, but the "reduces contention" claim isn't delivered — adjust the description or hold the connection's own lock rather than the whole-vec mutex across the call.

leszek3737:
done

houko (approved):
Re-reviewed — the code is good now: the ACP `input["args"]`-into-`sh -c` merge is gone (the remaining `args` join only feeds the read-only-path classifier, which is pre-existing), and `process_registry` is genuinely wired (`register` / `mark_finished` in `shell.rs`). The fallback-skill ACL check, `maybe_snapshot` reorder, `u32::try_from`, and spill wiring are all correct.

Approving the code, with one PR-body cleanup request (not blocking): the description still lists the ACP args-merge as a change (no longer in the diff) and still claims the MCP mutex split "reduces contention" — but `conn.call_tool_with_caller(...).await` still runs inside the `mcp_conns` vec lock (`dispatch.rs:1353-1383`), so MCP tool calls remain serialized. Please either drop that claim or hold a per-connection lock instead of the whole-vec mutex across the await, and trim the body to match the diff.

PR #5775, opened 2026-05-26
fix(runtime): tool_runner agent — taint all inputs, narrow capabilities, deny None, network strict

leszek3737:
## Type
- [x] Bug fix
- [x] Security

## Summary

Hardens `tool_runner/agent.rs`: expands taint coverage, narrows capability mapping, denies unprivileged defaults, adds observability.

## Changes

- Taint checks on `agent_id`, `message`, `conversation_key` in send; `name`, `system_prompt` in spawn; `agent_id` in kill
- Removed `"*"` wildcard from all capability matchers — only exact tool names trigger implied caps
- `parent_allowed_tools=None` → empty vec instead of `Capability::ToolAll`
- `filter_map` with `tracing::warn!` on non-string items (was silent drop)
- `network` parameter: `.ok_or(...)` rejects non-boolean (was `.unwrap_or(false)`)
- Pre-allocated buffer with `String::with_capacity` + `write!` replaces `format!` in loop

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and fully migrated `agent.rs` to `ToolError`, so every hunk needs rewriting onto that shape on rebase. The useful idea is the expanded taint coverage (agent_id/message/conversation_key/name/system_prompt) — but it must be re-expressed as `ToolError::PermissionDenied`, and four capability/network regressions must be fixed first:

1. `crates/librefang-runtime/src/tool_runner/agent.rs:246` — changing the `parent_allowed_tools == None` default from `vec![Capability::ToolAll]` to `vec![]` is a regression, not hardening. `None` means an *unrestricted* parent (no allowlist), not an unprivileged one. With empty `parent_caps`, `validate_capability_inheritance` (capability.rs:171) rejects any child requesting any capability, so unrestricted parents can no longer spawn capable children. No test covers the `None` path, so this passes CI while breaking production spawns. Keep `ToolAll` for the `None` case.
2. `crates/librefang-runtime/src/tool_runner/agent.rs:170` — removing `|| t == "*"` from `has_net`/`has_shell`/`has_agent_spawn`/`has_agent_msg`/`has_memory` breaks wildcard parents. A parent with `allowed_tools = ["*"]` now yields only `ToolInvoke("*")`, which does NOT cover `NetConnect`/`ShellExec`/etc. (different enum variants — the reason this expansion fn exists). Preserve the `"*"` arms (or special-case `"*"` → the full implied resource-cap set).
3. `crates/librefang-runtime/src/tool_runner/agent.rs:220` — `input["network"].as_bool().ok_or("Missing or non-boolean 'network' parameter")?` makes `network` effectively required, but the agent_spawn schema (`definitions.rs:243`) lists only `name`/`system_prompt` as required and documents `network` as optional (the PR had to add `"network": false` to existing tests to keep them green). Distinguish missing (default false) from present-but-wrong-type.
4. `crates/librefang-runtime/src/tool_runner/agent.rs:171` — `|| t == "net_fetch"` in `has_net` is dead code: `net_fetch` is a WASM sandbox host-function name, not an `execute_tool`-level tool name in `allowed_tools` (those are `web_fetch`/`web_search`). Remove it.

leszek3737:
done

houko (changes requested):
Re-reviewed the new commits. Thanks — all four capability/network regressions are reverted (None → `ToolAll` restored, the `"*"` arms are back, `network` is optional again, `net_fetch` removed). One new blocker introduced by the taint expansion:

- `crates/librefang-runtime/src/tool_runner/agent.rs:65-69` (`tool_agent_send`) and `:315-319` (`tool_agent_kill`) — `check_taint_outbound_text(agent_id, …)` false-positives on UUID agent ids. `looks_like_opaque_token` (`taint.rs:126-152`) returns true for any ≥32-char, whitespace-free `[A-Za-z0-9-_./+=]` string containing a punct char — a canonical UUID like `550e8400-e29b-41d4-a716-446655440000` qualifies (the `-` is punct), so both tools return "Taint violation (agent_id)" before resolving the target. Since `agent_find` returns UUIDs and the schema documents `agent_id` as "the target agent's UUID or name", this breaks the primary, canonical addressing path for agent_send/agent_kill.

Fix: drop the taint check on the `agent_id` identifier field (keep it on the free-form content fields — `message`, `conversation_key`, `system_prompt`, `name`), or resolve the id to an agent first and validate the identifier shape instead of taint-scanning it.

leszek3737:
done

houko (approved):
Re-reviewed — the UUID taint false-positive is fixed: `check_taint_outbound_text(agent_id, ...)` is removed from both `tool_agent_send` and `tool_agent_kill`, with a clarifying comment in each ("agent_id is a UUID/name identifier, not free-form content"). The taint expansion is retained on the genuine content fields (message/conversation_key/name/system_prompt), and the four capability/network regressions from the prior round are still reverted.

Approving.

PR #5778, opened 2026-05-26
fix(runtime): tool_runner process — output cap, strict caller_id, arg logging, serde_json

leszek3737:
## Type
- [x] Bug fix

## Summary

Hardens `tool_runner/process.rs`: caps poll output, requires caller identity, adds arg logging, uses structured JSON.

## Changes

- `MAX_POLL_OUTPUT_BYTES` (256KB) caps stdout/stderr with `join_with_cap` and `truncated` flag
- `caller_agent_id` required — no more silent fallback to "default"
- Args parsing: explicit loop with `tracing::warn!` on non-string items (was silent drop)
- Write data passed as-is — no appended newline
- Manual JSON strings replaced with `serde_json::json!()`

## ⚠️ Breaking change

`tool_process_poll` now emits **`"stdout"` / `"stderr"` as a newline-joined `String`** instead of the previous JSON array (`Vec<String>`). No in-repo consumer parses these fields as arrays, so impact is limited to LLM prompt formatting. Any out-of-tree consumer relying on the old array shape will need to update.
## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and migrated `process.rs` to `ToolResult` (`ToolError::Unavailable("Process manager")` / `MissingParameter` / `upstream_msg`). This PR keeps `Result<String, String>`, so it no longer compiles on `main`. The 256KB output cap, strict caller id, and structured-arg logging are net-new and worth porting. Three points:

1. `crates/librefang-runtime/src/tool_runner/process.rs` (all five fns) — won't compile after rebase: `.ok_or("Process manager not available")` and bare `?` on `Result<_, String>`. Re-express onto the ToolError shape.
2. `crates/librefang-runtime/src/tool_runner/process.rs` (`tool_process_write`) — removing the always-appended `\n` is a tool-contract change, not a hardening fix; `main` appends it with a comment that REPLs expect line submission. Dropping it can break interactive REPL drivers. If intentional, justify it; otherwise keep the newline.
3. `crates/librefang-runtime/src/tool_runner/process.rs` (`join_with_cap`) — `buf.len() + line.len() + 1 > max_bytes` breaks on the first line that would exceed, so a single line longer than 256KB yields an *empty* stdout with `truncated=true` (drops all output) rather than a truncated prefix. Truncate within the over-long line (on a char boundary).

leszek3737:
done

houko (changes requested):
Re-reviewed the new commits. Thanks — the always-appended `\n` in `tool_process_write` is restored, and `join_with_cap` now truncates within an over-long line instead of dropping all output (with tests). One thing to confirm before this can merge:

- `crates/librefang-runtime/src/tool_runner/process.rs` (`tool_process_poll`) — the output shape changed: main serialized `"stdout"` / `"stderr"` as a JSON array (`Vec<String>`), this PR now emits a single newline-joined String (`{"stdout":["a","b"]}` → `{"stdout":"a\nb\n"}`). The 256KB cap makes joining reasonable, but any consumer parsing these as arrays will break. Please confirm no caller / dashboard / test relies on the array shape, and note the shape change in the PR body.

Minor (non-blocking): the `line.len() < remaining` guard flags a line whose length exactly equals `remaining` as `truncated=true` even though only the trailing `\n` is dropped.

leszek3737:
done

houko (approved):
Re-reviewed — the substantive fixes are in:
- `join_with_cap` no longer drops the whole output on a single over-long line (truncates within the line on a char boundary, with the `join_with_cap_exact_fit_not_truncated` and `_truncates_within_long_line` tests).
- The always-appended `\n` in `tool_process_write` is restored.

Approving with one PR-body note that should land before merge (not blocking the code review): `tool_process_poll` now emits `"stdout"` / `"stderr"` as a newline-joined String rather than the prior JSON array (`Vec<String>`). I verified no in-repo consumer parses these as arrays, so impact is limited to LLM prompts — but please call out the breaking output-shape change explicitly in the PR description so future readers (and any out-of-tree consumer) know.

PR #5781, opened 2026-05-26
fix(runtime): tool_runner skill — spawn_blocking, symlink TOCTOU, size pre-check, backpressure semaphore

leszek3737:
## Type
- [x] Bug fix
- [x] Refactor / Performance

## Summary

Hardens `tool_runner/skill.rs`: async-safe I/O, symlink TOCTOU fix, size pre-check, backpressure on usage tracking, DRY helpers, unified delete.

## Changes

- All `librefang_skills::evolution::*` ops wrapped in `tokio::task::spawn_blocking`
- `tool_skill_read_file`: open → canonicalize → check → read from handle (no TOCTOU)
- Metadata size check before read — rejects files over 32KB upfront
- `USAGE_SEMAPHORE` (4 permits) limits concurrent usage tracking tasks
- `resolve_skill()` helper + `serialize_result()` eliminates 5x duplicated fallback pattern
- Delete accepts `caller_agent_id`, uses `resolve_skill` for consistent lookup

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and migrated `skill.rs` to `ToolResult` (`ToolError::Unavailable`, `skill_not_found()` → `NotFound`, `PermissionDenied` for the allowlist/containment guards). This PR keeps `Result<String, String>`, so the whole file needs re-porting. The `spawn_blocking` offloading and the usage-tracking semaphore are net-new and worth keeping. Issues:

1. `crates/librefang-runtime/src/tool_runner/skill.rs` (all fns) — won't compile after rebase; preserve the `skill_not_found`/`PermissionDenied` mappings #5737 added.
2. `crates/librefang-runtime/src/tool_runner/skill.rs` (`resolve_skill`, cache-miss path) — uses `tokio::task::block_in_place(...)`, which **panics on a current-thread runtime**, and every `#[tokio::test]` in `tests/skills.rs` is the default current-thread flavor. Any test driving update/patch/delete/rollback/write_file/remove_file through a registry cache miss will panic (production is multi-thread, masking it). Use `spawn_blocking` for the disk load, or keep the load synchronous.
3. `crates/librefang-runtime/src/tool_runner/skill.rs` (`tool_skill_read_file`, the "symlink TOCTOU fix") — does not fully close the race. Flow: `File::open(&requested)` (follows symlinks) → size check on the fd → `requested.canonicalize()` (a *separate* fresh path resolution) → `canonical.starts_with(skill_root)` → read from the original fd. Because the containment check canonicalizes the path again rather than validating the identity of the already-open fd, an attacker who swaps `requested` from an out-of-dir symlink to an in-dir file *between* open and canonicalize makes the check pass while the fd still points outside. Derive the check from the fd (compare device/inode against a path inside `skill_root`), not a re-resolved path. The claim "open → canonicalize → check → read from handle (no TOCTOU)" overstates what it achieves.
4. `crates/librefang-runtime/src/tool_runner/skill.rs` + `tests/skills.rs:175` — the PR flips `skill_read_file` from truncate-with-`floor_char_boundary` (test asserts `contains("truncated")`) to *reject* files over 32KB (test changed to `expect_err`). A large reference/template file that previously read (truncated) now hard-fails. Keep truncation, or make rejection opt-in.
5. `crates/librefang-runtime/src/tool_runner/skill.rs` — the `caller_agent_id` added to `skill_evolve_delete` is only used for a `tracing::debug!` line (not passed to `delete_skill`, which takes no author). Reconsider whether the cross-file signature churn is worth a debug log.

PR #5783, opened 2026-05-26
fix(runtime): tool_runner fs — backslash rejection, canonicalize, TOCTOU fix, read limit, dir pagination, atomic write

leszek3737:
## Type
- [x] Bug fix
- [x] Security

## Summary

Hardens `tool_runner/fs.rs` and `apply_patch.rs` with 7 fixes: path validation, canonicalization, TOCTOU elimination, read limits, directory pagination, error context, atomic writes.

## Changes

- `reject_backslash_on_unix()` rejects backslash paths on non-Windows
- `check_absolute_path_inside_workspace` canonicalizes both sides with fallback
- `tool_file_read` opens file directly after resolve — no separate stat-then-open TOCTOU
- `max_bytes` parameter (default 10MB) enforced via metadata check before read
- `tool_file_list` accepts `limit`/ `offset` with summary footer
- All errors include `resolved.display()` in message
- `atomic_write()` in apply_patch — writes to tmp + rename, cleans up on failure

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and migrated `fs.rs` to `ToolResult`. This PR keeps `String` errors, so it no longer compiles on `main`. The security content (backslash rejection, read size limit, dir pagination, atomic write) is NOT in #5737 (it only swapped error channels) — still wanted after rebase. Issues:

1. `crates/librefang-runtime/src/tool_runner/fs.rs` (`tool_file_read`/`tool_file_write`/`tool_file_list`) — won't compile after rebase: `format!(...)` / `.ok_or("...")?` yield `String`. Convert: missing/invalid params → `ToolError::MissingParameter`/`InvalidParameter`, I/O failures → `ToolError::Upstream { .. }` (mirror the existing `map_err(|e| ToolError::Upstream{…})` already in the migrated `tool_file_read`).
2. `crates/librefang-runtime/src/tool_runner/fs.rs` — `Vec::with_capacity(file_size as usize)` where `file_size <= max_bytes` and `max_bytes` is fully caller-controlled (`input["max_bytes"].as_u64().unwrap_or(10MB)`). An LLM passing a huge `max_bytes` against a large file lets a single read pre-allocate that many bytes (OOM amplification). Cap `max_bytes` at a hard ceiling, or `with_capacity(file_size.min(cap))`.
3. Minor (not blocking): the "TOCTOU fix" framing is inaccurate — the old `read_to_string(&resolved)` performed no stat, so there was no stat-then-open race; and `resolved` is already canonicalized + workspace-jailed by `resolve_sandbox_path_ext`. The new `File::open` + `file.metadata()` is fine, just not closing a pre-existing TOCTOU. The backslash-on-Unix rejection in `check_absolute_path_inside_workspace` is genuinely useful (the `..`-component check uses `Path::components()`, which on Unix doesn't split on `\`) — keep it.

leszek3737:
done

houko (changes requested):
Re-reviewed the new commits. Thanks — migrated to `ToolError`, `max_bytes` is now capped (`HARD_MAX_READ_BYTES = 128MB`, rejecting oversize before allocating), and the backslash-on-Unix rejection is kept. Two atomic-write temp-path races in the new code:

- `crates/librefang-runtime/src/tool_runner/fs.rs:358` (`tool_file_write`) — the temp path is `resolved.with_extension(format!("{}.tmp", std::process::id()))`. The pid is constant within the daemon process, so two concurrent `file_write` calls to the same `resolved` path collide on the identical temp file — one write clobbers the other's temp before rename, silently losing/interleaving a write. The atomic-write was added for correctness, so this defeats its own purpose under concurrency.
- `crates/librefang-runtime/src/apply_patch.rs:31-32` (`atomic_write`) — same class: temp name `.librefang-tmp-<filename>` has no pid and no randomness, so concurrent patches to the same target file race on the same temp path.

Fix: add a per-call uniquifier (an atomic counter or a random suffix), not just the pid.

leszek3737:
done

houko (approved):
Re-reviewed — both atomic-write temp-path races are fixed:
- `tool_file_write` now uses `with_extension(format!("{}.tmp", rand::random::<u64>()))` so concurrent same-target writes don't collide on the identical temp file.
- `apply_patch.rs::atomic_write` is extracted into a dedicated helper with `.librefang-tmp-<filename>-<rand::random::<u64>()>` plus cleanup of the temp file on write/rename failure.

Approving. Small consistency note (non-blocking): `tool_file_write` returns the upstream `tokio::fs::write` error without removing the partially-written temp file on failure, while the new `apply_patch.rs::atomic_write` cleans up. Worth mirroring `apply_patch`'s cleanup pattern in `tool_file_write` in a follow-up.

PR #5784, opened 2026-05-26
fix(runtime): tool_runner hand — deterministic sort, empty id reject, config whitelist, output sanitization

leszek3737:
## Type
- [x] Bug fix

## Summary

Hardens `tool_runner/hand.rs`: deterministic list ordering, input validation, config key allowlisting, output sanitization.

## Changes

- `tool_hand_list` sorts by `id` for deterministic output
- Empty `hand_id` / `instance_id` rejected via `validate_required_string()`
- `ALLOWED_CONFIG_KEYS` whitelist — unknown keys dropped with `tracing::warn!`
- `sanitize_value()` truncates output strings to safe lengths (name 80, desc 200, instance_id 64)
- Schema validation combines all above checks

## Attribution

## Testing

## Security
- [x] No new unsafe code

### DIFF
--- a/crates/librefang-runtime/src/tool_runner/hand.rs
+++ b/crates/librefang-runtime/src/tool_runner/hand.rs
@@ -8,9 +8,58 @@ use super::require_kernel_typed;
 use crate::kernel_handle::prelude::*;
 use std::sync::Arc;
 
+const ALLOWED_CONFIG_KEYS: &[&str] = &[
+    "model",
+    "temperature",
+    "max_tokens",
+    "system_prompt",
+    "schedule",
+    "session_mode",
+    "auto_restart",
+    "budget_limit",
+];
+
+fn truncate_char_boundary(s: &str, max_len: usize) -> &str {
+    if s.len() <= max_len {
+        return s;
+    }
+    match s.char_indices().nth(max_len) {
+        Some((i, _)) => &s[..i],
+        None => s,
+    }
+}
+
+fn sanitize_value(val: &serde_json::Value, max_len: usize) -> String {
+    let s = match val {
+        serde_json::Value::String(s) => s.clone(),
+        other => other.to_string(),
+    };
+    if s.len() > max_len {
+        format!("{}…[truncated]", truncate_char_boundary(&s, max_len))
+    } else {
+        s
+    }
+}
+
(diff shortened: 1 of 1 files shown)


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and migrated `hand.rs` to `ToolResult`. This PR keeps `String` errors, so it no longer compiles on `main`. The sort/validation/sanitization are net-new and worth keeping. Issues:

1. `crates/librefang-runtime/src/tool_runner/hand.rs` (`sanitize_value`, the `&s[..max_len]` line) — panic bug. `s.len()`/`&s[..max_len]` are byte-based; if `max_len` lands mid-UTF-8-codepoint (a hand name/description with emoji/CJK/accents exceeding the limit — plausible metadata), the slice panics with "byte index is not a char boundary", crashing the tool call. Use char-aware truncation, e.g. `s.char_indices().nth(max_len).map_or(s.as_str(), |(i,_)| &s[..i])`.
2. `crates/librefang-runtime/src/tool_runner/hand.rs` (all four fns) — won't compile after rebase; convert via `ToolError::MissingParameter`/`InvalidParameter`/`Upstream`. `validate_required_string` should return `ToolResult<String>` (empty-string case → `InvalidParameter`).
3. `crates/librefang-runtime/src/tool_runner/hand.rs` (`ALLOWED_CONFIG_KEYS`) — the allowlist is hardcoded in the tool layer with no link to what `hand_activate` actually accepts; any valid config key not in the fixed list is silently dropped (only a `warn!`). If hands legitimately accept keys beyond these, that's a silent regression. Derive the allowlist from the hand manifest schema, or surface dropped keys to the caller rather than just logging.

leszek3737:
done

houko (approved):
Re-reviewed — all three points addressed:
- `sanitize_value` now truncates on a char boundary (`truncate_char_boundary` via `char_indices().nth()`), so the byte-slice panic on emoji/CJK metadata is gone.
- Migrated to `ToolError` (`validate_required_string` → `MissingParameter` / `InvalidParameter`).
- Dropped config keys are now surfaced to the caller in the tool output instead of only a server-side `warn!`.

PR #5785, opened 2026-05-26
fix(runtime): tool_runner goal — progress type fix, range validation

leszek3737:
## Type
- [x] Bug fix

## Summary

Fixes 2 bugs in `tool_runner/goal.rs`: wrong type for progress parsing, silent clamping of out-of-range values.

## Changes

- `as_u64()` → `as_f64()` — JSON float values now handled correctly
- `.min(100)` silent clamp replaced with explicit error when progress > 100
- Range check `!(0.0..=100.0).contains(&raw)` before cast — catches negative floats

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** GitHub flags this MERGEABLE/CLEAN, but that is a text-merge result — the merged file does **not** compile. The f64 parse + range validation are real fixes worth keeping; they just need to route through `ToolError`:

1. `crates/librefang-runtime/src/tool_runner/goal.rs` (`tool_goal_update`, the `progress` block) — #5737 (ToolError migration, #3576) migrated `tool_goal_update` to return `ToolResult` (`Result<String, ToolError>`). The text merge applies cleanly (the `progress` line was untouched by #5737), but the new code does `.ok_or("Parameter 'progress' must be a number".to_string())?` and `return Err(format!("...between 0 and 100..."))` — `String` does not `Into<ToolError>` (only `serde_json::Error` does), so both are type errors. I confirmed by merging this branch into `main` and inspecting the result. Emit `ToolError::InvalidParameter { name: "progress", reason: … }` for both the non-number and out-of-range cases.
2. `crates/librefang-runtime/src/tool_runner/goal.rs` — `input.get("progress").is_some()` treats an explicit JSON `null` as "present" → now errors, whereas the old `as_u64()` path treated `null` as absent. Likely fine, but confirm no caller sends `progress: null` to mean "leave unchanged".

Don't merge on the green flag here.

leszek3737:
done

houko (approved):
Re-reviewed — fixed: `tool_goal_update` now routes both the non-number and out-of-range cases through `ToolError::InvalidParameter`, so it compiles (the earlier "green flag but doesn't compile" state is resolved), and there are unit tests covering the range validation. The explicit-`null` behavior is a defensible non-issue — the schema documents omit-to-skip, not a `null` sentinel.

PR #5786, opened 2026-05-26
fix(runtime): tool_runner event — event_type validation, caller identity, reserved prefix guard

leszek3737:
## Type
- [x] Bug fix
- [x] Security

## Summary

Hardens `tool_runner/event.rs`: validates event_type, requires caller identity, blocks reserved prefixes.

## Changes

- `event_type` trimmed and empty/whitespace-only values rejected
- `caller_agent_id` required — anonymous agents cannot publish events
- Events with `system.` or `internal.` prefix blocked with permission-denied
- Trimmed value used in publish call and success message

## Attribution

## Testing

## Security
- [x] No new unsafe code


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and rewrote both `event.rs` and the `dispatch.rs` dispatch table, so this PR has hard text conflicts and no longer compiles. The validation/caller-auth is genuinely new and not superseded — keep it after rebase. Two points:

1. `crates/librefang-runtime/src/tool_runner/event.rs` + `dispatch.rs:899` — won't compile / hard conflict. Convert the new errors: empty `event_type` → `ToolError::InvalidParameter { name: "event_type", .. }`, reserved-prefix and missing-caller → `PermissionDenied(..)`.
2. `crates/librefang-runtime/src/tool_runner/event.rs` (the `caller_agent_id.ok_or(...)?` line) — hard-requiring `caller_agent_id` to be `Some` flips `event_publish` from the documented "pure passthrough" to caller-gated. The agent loop populates it for `agent_send`/`task_post`, so this is probably safe, but it is a contract change — confirm no legitimate system/cron-internal path reaches `event_publish` with `caller_agent_id = None`.

(Logic is otherwise correct — the reserved prefixes include the trailing dot, so `"systematic.x"` is not falsely blocked.)

leszek3737:
done

houko (approved):
Re-reviewed — both points addressed:
- Migrated to `ToolError` (empty `event_type` → `InvalidParameter`; reserved-prefix and missing-caller → `PermissionDenied`); the dispatch call site is updated; reserved prefixes carry the trailing dot so `"systematic.x"` isn't falsely blocked.
- Confirmed the caller-gating is safe: the live agent loop always passes `Some(caller_id)`, and the only `None` source is a degenerate deferred-resume record (empty/non-UUID `agent_id`), not a system/cron path.

PR #5787, opened 2026-05-26
fix(runtime): tool_runner a2a — session_id taint, SSRF diagnostics, zero-alloc agent check

leszek3737:
## Type
- [x] Bug fix
- [x] Security

## Summary

Hardens `tool_runner/a2a.rs`: adds taint check on session_id, improves SSRF error messages, removes unnecessary allocation.

## Changes

- `session_id` checked via `check_taint_outbound_text` before send — closes exfiltration gap
- SSRF errors include URL and reason from `check_ssrf` (was generic message)
- `Vec<String>` allocation replaced with `Iterator::any()` — short-circuit, no allocation

## Attribution

## Testing

## Security
- [x] No new unsafe code

### DIFF
--- a/crates/librefang-runtime/src/tool_runner/a2a.rs
+++ b/crates/librefang-runtime/src/tool_runner/a2a.rs
@@ -19,10 +19,10 @@ pub(super) async fn tool_a2a_discover(input: &serde_json::Value) -> ToolResult {
         .ok_or(ToolError::MissingParameter("url"))?;
 
     // SSRF protection: block private/metadata IPs
-    if crate::web_fetch::check_ssrf(url, &[]).is_err() {
+    if let Err(reason) = crate::web_fetch::check_ssrf(url, &[]) {
         return Err(ToolError::InvalidParameter {
             name: "url",
-            reason: "SSRF blocked: URL resolves to a private or metadata address".to_string(),
+            reason: format!("SSRF blocked for '{url}': {reason}"),
         });
     }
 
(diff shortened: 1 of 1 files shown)


houko (changes requested):
**Requesting changes.** Note: #5737 (ToolError migration, #3576) merged to `main` and migrated `a2a.rs` to `ToolResult` (wrapping the SSRF blocks in `ToolError::InvalidParameter`), so this PR has a hard conflict and no longer compiles. All three changes are still needed (the SSRF generic-message and the missing session_id taint check both persist on `main`). The session_id taint check is the most valuable part — keep it:

1. `crates/librefang-runtime/src/tool_runner/a2a.rs` (both SSRF blocks + the new session_id block) — won't compile / hard conflict. Convert: the session_id taint violation → `ToolError::PermissionDenied(violation)` (matching the existing message/url taint pattern), and the SSRF diagnostic → `ToolError::InvalidParameter { name: "url"/"agent_url", reason: format!("SSRF blocked for '{url}': {reason}") }`.
2. `crates/librefang-runtime/src/tool_runner/a2a.rs` (the session_id taint check — keep this) — it closes a real exfiltration gap: on `main`, `session_id = input["session_id"].as_str()` is passed straight to `client.send_task(&url, message, session_id)` with no taint scan, while `message` and `url` are scanned, so an LLM-controlled `session_id` carrying secrets reaches the external peer unchecked. Minor placement nit: the existing message taint check runs *before* the trust gate (deliberately, per the in-file comment); move the session_id check up next to it so a tainted attempt always reports the exfil reason rather than the trust-gate rejection.

houko (approved):
Re-reviewed — both points addressed:
- Migrated to `ToolError`: SSRF blocks → `InvalidParameter` (now including the underlying reason — a diagnostic improvement), taint violations → `PermissionDenied`.
- The `session_id` taint check is kept and placed above the trust gate, so a tainted attempt reports the exfil reason rather than the trust-gate rejection. This closes the real exfiltration gap (an LLM-controlled `session_id` previously reached the external peer unscanned).

PR #5800, opened 2026-05-27
fix(kernel): route auto_evolve creates through skill_workshop pending queue

### LINKED ISSUE #5798: fix(kernel): unify auto_evolve and skill_workshop into single pending-draft pipeline
## Problem

There are two independent skill auto-creation systems that confuse users and produce poor results separately:

1. **`auto_evolve` / `background_skill_review`** — LLM reviewer runs after each turn. Generates good semantic content BUT:
   - 30s timeout too short for local models (Ollama/qwen) — most reviews time out
   - Review prompt is too conservative — almost always returns `skip`
   - Creates skills directly without user approval

2. **`skill_workshop`** (PR #4741 / #3328) — Heuristic scanner detects repeated tool patterns, user corrections, explicit instructions. Has approval flow (pending drafts) BUT:
   - `RepeatedToolPattern` scanner generates useless drafts like `tool_sequence_file_read_x5` with just tool names listed
   - No semantic understanding of what the user was doing

From a user perspective, these should be **one system**: the agent learns from interactions and proposes skills for human review.

## Proposed fix

1. **Increase `background_skill_review` timeout** from 30s to 120s (local models need more time)
2. **Lower the review threshold** — add criteria like "3+ different tools in sequence" and "multi-step procedure that could be reused"
3. **Route `create` actions through `skill_workshop::storage::save_candidate`** instead of `evolution::create_skill` — so LLM-proposed skills land in the Pending tab for human approval instead of being created directly
4. Keep `update`/`patch` actions as direct operations (modifying skills the user already approved)

## Result

- One unified flow: agent turn → LLM review → pending draft with semantic content → human approval
- No more useless `tool_sequence_*` drafts — the LLM generates meaningful names, descriptions, and procedures
- Users see proposed skills in the Pending tab and decide what to keep

Refs: #3328

DaBlitzStein:
## Summary

- Route `background_skill_review` create actions through `skill_workshop::storage::save_candidate` instead of directly calling `evolution::create_skill` — LLM-proposed skills now land in the Pending tab for human approval
- Increase review timeout from 30s to 120s for local models (Ollama/qwen)
- Lower review threshold: add "3+ different tools in sequence" and "multi-step reusable procedure" criteria

## Test plan

- [ ] Agent with `auto_evolve = true` completes a multi-tool task → verify draft appears in Pending tab (not directly installed)
- [ ] `cargo test -p librefang-kernel --lib skill_workshop`
- [ ] `cargo clippy --workspace --all-targets -- -D warnings`

Closes #5798


houko (changes requested):
**Requesting changes.** The `"create"` routing is correct and preserves the security boundary (`save_candidate` runs the prompt-injection scan; `SecurityBlocked` → `ReviewError::Permanent`). Two issues:

1. `crates/librefang-kernel/src/kernel/tools_and_skills.rs:735-855` — only the `"create"` arm is routed to the pending queue. The `"update"` (~:772) and `"patch"` (~:840) arms still call `evolution::update_skill` / `patch_skill` followed by `do_reload()`, so the background reviewer can still silently rewrite or patch a live, already-installed skill without human approval. #5798's premise ("LLM-proposed skills land in Pending for human approval") is only half-enforced: an attacker who gets a malicious turn past review can mutate an existing trusted skill in place, bypassing the pending gate. Route update/patch through the pending queue too (capture them as candidates that supersede the target on approval), or document in the PR why in-place mutation of existing skills is intentionally left on the direct path. This is the same function/crate, so resolve it here rather than deferring.
2. `crates/librefang-kernel/src/kernel/tools_and_skills.rs` — the behavior change (create now queues instead of install + `do_reload`) has no regression guard; the listed `cargo test -p librefang-kernel --lib skill_workshop` exercises the pre-existing `storage` tests, not the new routing in `background_skill_review`. Add a test that drives the `"create"` branch (or factor the candidate-construction + `save_candidate` call into a small testable helper) and asserts a pending candidate is written and no live skill is installed.

(Consistency with #5678/#5741 checks out: `background_skill_review` is gated on `manifest.auto_evolve` (default true); the pending list endpoint + dashboard `PendingSkillsSection` read the pending dir directly, so drafts from the default-on `auto_evolve` path are visible/approvable even when the workshop feature is off — no orphaned drafts.)

houko (approved):
Re-reviewed the latest commit (docs on the update/patch arms + the extracted `build_reviewer_candidate` helper + four new unit tests). Both of my earlier concerns are addressed:

- **create → pending routing** now has a real regression guard: `create_action_saves_to_pending_queue_not_live_registry` asserts a pending file is written *and* no live skill dir is installed. The negative assertion is meaningful — `create_skill` would have done `skills_dir.join(name)` + `create_dir_all`, so this genuinely distinguishes pending-routing from direct install.
- **update/patch staying on the direct path** is now documented as a deliberate decision, and the doc is factually accurate: I verified `update_skill` (evolution.rs:1067) and `patch_skill` (evolution.rs:1195) both call `validate_prompt_content` → `SkillVerifier::scan_prompt_content` *before* touching disk, and both arms propagate `SecurityBlocked` → `ReviewError::Permanent`. So the prompt-injection boundary is still enforced on refinements of already-trusted skills.

Also confirmed the test API usage matches real signatures (`save_candidate` 4-arg, `list_pending` returns `Vec<CandidateSkill>` with `.name`, `WorkshopError::SecurityBlocked` exists), agent IDs are valid UUIDs (required by `validate_uuid_id`), and the injection payload sits in `prompt_context` which `save_candidate` scans.

No new correctness/security/architecture issues. LGTM.

PR #5802, opened 2026-05-27
fix(channels): skip Markdown→HTML conversion for sidecar adapters (#5795)

### LINKED ISSUE #5795: bug(channels): double formatting on non-streaming sidecar paths — Telegram and Matrix show raw HTML tags
## Problem

Non-streaming messages sent to Telegram and Matrix sidecars show literal HTML tags (e.g., visible \`<b>bold</b>\`) instead of rendered formatting.

## Root cause: double formatting

The Rust bridge and the Python sidecar both independently convert Markdown to HTML, causing double-escaping.

**Path:**
1. \`bridge.rs:send_response()\` calls \`format_for_channel(text, TelegramHtml)\` converting \`**bold**\` to \`<b>bold</b>\`
2. Result sent to sidecar as \`ChannelContent::Text("<b>bold</b>")\`
3. Python sidecar \`telegram.py:_send_text()\` calls \`_format_and_sanitize()\` which calls \`markdown_to_telegram_html()\` whose first step is \`_escape_html()\` converting \`<\` to \`&lt;\`
4. Telegram receives \`&lt;b&gt;bold&lt;/b&gt;\` with \`parse_mode=HTML\` and renders \`<b>bold</b>\` as literal text

The sidecar explicitly documents the expected contract at \`telegram.py:617\`: *"Daemon sends raw agent Markdown; the sidecar owns formatting now."* But \`send_response\` was not updated after the sidecar migration.

## Affected channels

- **Telegram** (\`markdown_to_telegram_html\` in both Rust formatter.rs and Python telegram.py)
- **Matrix** (\`markdown_to_matrix_html\` in Python matrix.py, plus Rust side formatting)

## Affected paths (non-streaming only)

Streaming works correctly because raw deltas bypass \`format_for_channel\`. Non-streaming paths that call \`send_response\` are broken:
- Bot commands (/help, /model, etc.)
- Error messages
- Auto-reply responses
- Multimodal responses (images, files, voice)
- Rate-limited responses
- Streaming fallback (when streaming fails)
- Access denied messages

## Proposed fix

In \`bridge.rs:send_response()\`, skip \`format_for_channel\` when the adapter is a sidecar that owns its own formatting. The sidecar contract says the daemon sends raw Markdown.

Two approaches:
- **A)** Add a trait method \`fn owns_formatting(&self) -> bool\` to \`ChannelAdapter\`. Sidecars return \`true\`, in-process adapters return \`false\`.
- **B)** Pass \`OutputFormat::Markdown\` (passthrough) for sidecar adapters instead of the channel-specific format.

houko:
## Root cause

`bridge.rs:send_response()` (line 2730) always called `format_for_channel(text, output_format)` before forwarding to the adapter. For Telegram (default format `TelegramHtml`) and Matrix, this converted `**bold**` → `<b>bold</b>`. The Python sidecar then applied its own `markdown_to_*_html` pass whose first step is `_escape_html`, converting `<` → `&lt;`. Telegram/Matrix received `&lt;b&gt;bold&lt;/b&gt;` with `parse_mode=HTML` and rendered literal tags.

The sidecar contract (`telegram.py:617`): "Daemon sends raw agent Markdown; the sidecar owns formatting now." This was never honoured by `send_response`.

Streaming was not affected because raw deltas bypass `format_for_channel` entirely.

## Affected non-streaming paths (all fixed)

- Bot commands (`/help`, `/model`, etc.)
- Error messages
- Auto-reply responses
- Multimodal responses (images, files, voice)
- Rate-limited responses
- Streaming fallback (when streaming fails)
- Access-denied messages

## Approach chosen

**Approach A**: add `fn owns_formatting(&self) -> bool` to `ChannelAdapter` trait (default `false`). `SidecarAdapter` overrides to `true`. `send_response` skips `format_for_channel` when the adapter owns formatting.

Chosen over Approach B (`OutputFormat::Markdown` for sidecars) because:
- Semantically correct — the method expresses the architectural contract, not just the format
- Catches all future sidecar adapters automatically via the override in `SidecarAdapter`
- Does not require threading the format decision through call sites that derive `output_format` from the channel type string

## Changes

- `crates/librefang-channels/src/types.rs`: add `owns_formatting()` to `ChannelAdapter` trait (default `false`) with doc comment explaining the sidecar contract
- `crates/librefang-channels/src/sidecar.rs`: override `owns_formatting()` → `true` in `SidecarAdapter`; extend `test_cap_gated_methods_default_without_ready` to assert `owns_formatting() == true`
- `crates/librefang-channels/src/bridge.rs`: `send_response` skips `format_for_channel` when `adapter.owns_formatting()` is `true`
- `crates/librefang-channels/tests/bridge_integration_test.rs`: add `SidecarMockAdapter` (owns_formatting = true) and two end-to-end regression tests:
  - `test_bridge_sidecar_adapter_receives_raw_markdown_not_html` — Telegram sidecar gets raw `**bold**`, not `<b>bold</b>`
  - `test_bridge_in_process_adapter_converts_markdown_to_channel_format` — Slack in-process adapter still gets converted mrkdwn

## Verification

`cargo check --workspace --lib` and `cargo test -p librefang-channels` were not runnable locally (no native toolchain; Docker volume out of disk space). The code was verified structurally:
- Hook: pre-commit passed (channel-policy, detect-secrets, CHANGELOG checks)
- CI will run `cargo nextest run --workspace -E 'kind(lib) | kind(bin)'` and the integration test shard

Fixes #5795

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -641,7 +641,7 @@
         "filename": "crates/librefang-api/src/routes/registry.rs",
         "hashed_secret": "4e5111dafe7d9739f9228906d90cf421d5ec5371",
         "is_verified": false,
-        "line_number": 439
+        "line_number": 453
       }
     ],
     "crates/librefang-api/src/routes/terminal.rs": [
@@ -1433,7 +1433,7 @@
         "filename": "crates/librefang-channels/src/sidecar.rs",
         "hashed_secret": "1a91d62f7ca67399625a4368a6ab5d4a3baa6073",
         "is_verified": false,
-        "line_number": 2680
+        "line_number": 2692
       }
     ],
     "crates/librefang-cli/src/main.rs": [
@@ -4038,5 +4038,5 @@
       }
     ]
   },
-  "generated_at": "2026-05-26T10:16:51Z"
+  "generated_at": "2026-05-27T23:25:26Z"
 }
--- a/crates/librefang-channels/src/bridge.rs
+++ b/crates/librefang-channels/src/bridge.rs
@@ -2727,6 +2727,11 @@ async fn resolve_prefix_chunk(
 }
 
 /// Send a response, applying output formatting and optional threading.
+///
+/// Formatting is skipped when `adapter.owns_formatting()` returns `true` (e.g.
+/// sidecar adapters such as Telegram and Matrix). Those adapters apply their own
+/// Markdown→HTML conversion internally; pre-converting here would cause
+/// double-escaping and visible raw HTML tags on the platform. See issue #5795.
 async fn send_response(
     adapter: &dyn ChannelAdapter,
     user: &ChannelUser,
@@ -2740,7 +2745,13 @@ async fn send_response(
         text_len = text.len(),
         "Sending response to channel"
     );
-    let formatted = formatter::format_for_channel(&text, output_format);
+    // Sidecar adapters own formatting — forward raw Markdown.
+    // In-process adapters receive the channel-converted form.
+    let formatted = if adapter.owns_formatting() {
+        text
+    } else {
+        formatter::format_for_channel(&text, output_format)
+    };
     let content = ChannelContent::Text(formatted);
 
     let result = if let Some(tid) = thread_id {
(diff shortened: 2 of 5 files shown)


houko:
Superseded by #5796 which landed earlier today (commit `dcd3b3f5`). Both PRs target #5795 with the same approach — adding `ChannelAdapter::owns_formatting()` and gating `format_for_channel` on it. #5796's version is now on main (`types.rs:888`, `sidecar.rs` override). Closing to avoid duplicate work.

PR #5334, opened 2026-05-20
fix(api): isolate attachment pre-inject per chat session — close cross-chat image leak

f-liva:
## Incident

Confirmed cross-chat data leak on a production WhatsApp deployment, 2026-05-20:

| When | Where | What |
|---|---|---|
| 08:43:13Z | WhatsApp DM `chat 121043` (private 1:1 with the agent's owner) | Owner sent an IMAGE (`image.png`, 47990 bytes — Amazon-order screenshot showing "metro laser, 27,98€") with caption "segna spesa in Shopping". |
| 08:43:14Z | Group session msgpack `7b6af157` | `[Image (image/jpeg) previously processed]` placeholder appeared at the same second as the DM inbound. |
| 09:44:04Z | WhatsApp group `chat 120957` "Non perdiamoci 💻" (public, with third parties) | Agent replied to another participant: *"Life coach, metro laser — il Signore misura ogni aspetto della sua vita con precisione millimetrica. 🎩📏 *(27,98€ ben spesi, Signore. Vuole che tenga d'occhio la spedizione?)*"* — the **specific** order details from the private DM, **verbatim**, in a public group. |

The placeholder text is produced by `MessageContent::strip_images()` (`crates/librefang-types/src/message.rs:261`) — it is emitted when a real `ContentBlock::Image` is compacted away after the LLM has already consumed it. Its presence in the *group* session at the exact second the DM inbound arrived is direct evidence the DM image was persisted into the wrong session.

## Root cause

`SessionWriter::inject_attachment_blocks` (kernel impl at `crates/librefang-kernel/src/kernel/handles/session_writer.rs:77`) took only `(agent_id, blocks)` and wrote into `entry.session_id` — the agent registry's single persistent session slot, NOT the per-chat session derived for the current turn.

For a chat agent with `session_mode = "persistent"` whose group chat stays warm, `entry.session_id` is that warm group session. Any DM-inbound image therefore landed in the group session instead of the DM session that the text part of the same request would derive via `SessionId::for_sender_scope(agent, "whatsapp", "<dm_jid>")`.

The attachment pre-inject step (in `crates/librefang-api/src/routes/agents.rs:1501` `inject_attachments_into_session`, called from both `send_message` and `send_message_stream`) runs BEFORE the kernel's `send_message_*` proper. PR #5288 (streaming sender-context resolver) fixed the analogous gap on the text path but **deliberately did not touch this attachment pre-inject step** — so even with #5288 deployed the attachment routing remained broken.

## The fix

[...]

PR #5816, opened 2026-05-28
fix(runtime): unbreak coverage build — thread session_id into two SessionWriter test stubs

houko:
## Problem

`main` is red on the **Workspace coverage (llvm-cov)** job:

```
error[E0050]: method `inject_attachment_blocks` has 3 parameters
but the declaration in trait `inject_attachment_blocks` has 4
  --> crates/librefang-runtime/tests/proactive_memory_extraction_model_override.rs:213:9
```

#5334 added a `session_id: SessionId` parameter to
`SessionWriter::inject_attachment_blocks` (between `agent_id` and `blocks`)
and updated the production impl plus most test kernels, but missed two
integration-test stub impls. The crate **lib** compiles, so a
`cargo check --lib` does not catch it — only the `--tests` targets (and the
coverage job, which builds all targets) fail.

## Fix

Add the missing `session_id` parameter to the two stubs so they match the
trait and the other 17 implementors:

- `crates/librefang-runtime/tests/tool_runner_memory_isolation.rs`
- `crates/librefang-runtime/tests/proactive_memory_extraction_model_override.rs`

Both are no-op stubs; the parameter is bound as `_` / `_session_id`.

## Verification

A repo-wide scan of all `inject_attachment_blocks` implementors confirms these
two were the only remaining 3-parameter stubs; all others (production
`session_writer.rs`, the other test kernels, `host_functions.rs`,
`tool_runner/tests/*`) already carry the 4-parameter signature.

Local Docker verification was blocked by a full Docker VM disk, so this relies
on CI — the same **Workspace coverage (llvm-cov)** job that flagged the
breakage, plus the Ubuntu integration shards, gate this PR.

### DIFF
--- a/crates/librefang-runtime/tests/proactive_memory_extraction_model_override.rs
+++ b/crates/librefang-runtime/tests/proactive_memory_extraction_model_override.rs
@@ -212,6 +212,7 @@ impl librefang_kernel_handle::SessionWriter for OverrideKernel {
     fn inject_attachment_blocks(
         &self,
         _: librefang_types::agent::AgentId,
+        _: librefang_types::agent::SessionId,
         _: Vec<librefang_types::message::ContentBlock>,
     ) {
     }
--- a/crates/librefang-runtime/tests/tool_runner_memory_isolation.rs
+++ b/crates/librefang-runtime/tests/tool_runner_memory_isolation.rs
@@ -247,6 +247,7 @@ impl librefang_kernel_handle::SessionWriter for IsolationKernel {
     fn inject_attachment_blocks(
         &self,
         _agent_id: librefang_types::agent::AgentId,
+        _session_id: librefang_types::agent::SessionId,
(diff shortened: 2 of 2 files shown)


houko:
Folded into #5808.

The two `main`-red breakages are interdependent and neither PR's CI can go
green alone:

- **This PR** fixes the `inject_attachment_blocks` 3-vs-4 parameter compile
  error (E0050) from #5334's missed stubs — but on its own it still inherits
  the `Instant::now() - Duration` panic that fails the Test jobs.
- **#5808** fixes that `Instant` panic — but on its own it still hits *this*
  compile error, so its `Test / Unit`, `Test / Ubuntu` and `Quality` jobs all
  fail with the same E0050.

To get a single green PR that unblocks `main` in one merge, I cherry-picked this
commit (a7117294) onto `#5808` (`fix-main-red-ci-5726`). #5808 now carries both
fixes. Closing here in favor of #5808.

PR #5843, opened 2026-05-28
fix(dashboard): apply search filter to FangHub skills grid

DaBlitzStein:
## Summary

- `fanghubItems` was filtered by `selectedCategory` but not by `search`, so typing in the search bar had no effect on the FangHub tab
- Add the same name/description text filter that `buildRemoteItems` already applies to clawhub/skillhub results
- Add `search` to the `useMemo` dependency array

## Verification

- `pnpm typecheck` — zero errors
- One file changed: `SkillsPage.tsx`

## Test plan

- [ ] Open Skills → Browse (FangHub tab)
- [ ] Type in search bar → FangHub skills filter by name/description
- [ ] Clear search → all skills reappear
- [ ] Switch to clawhub tab → search still works (pre-existing)


houko (approved):
Reviewed — LGTM. This fixes a real bug: `fanghubItems` filtered the FangHub registry by `selectedCategory` only, so the Browse search bar had no effect on the FangHub tab while it worked on every other hub.

Confirmed:
- The new filter predicate (`SkillsPage.tsx:1672-1677`) is identical to the one `buildRemoteItems` already applies to the other hubs (lines 1659-1664), so behavior is now consistent across all hubs.
- Adding `search` to the `useMemo` dependency array was exactly the missing piece — without it the memo would serve a stale filtered list.
- No category/search double-filter issue: the search handler clears `selectedCategory` and the category handler clears `search`, so they're mutually exclusive at runtime.
- Type-safe (`FangHubSkill.name`/`description` are both `string`), no new `fetch`/`api.*` (goes through the existing `fanghubQuery`), nothing reaches an LLM prompt so no ordering concern.

Optional follow-up (non-blocking): a `SkillsPage` render test would be nice, but there's no existing component harness for this 2000+ line page and the mirrored `buildRemoteItems` predicate is itself untested — fair to leave for a separate effort.

PR #5852, opened 2026-05-29
refactor(subprocess): unify the three sidecar bridges onto a shared transport crate

houko:
## What

Unifies LibreFang's three hand-rolled "persistent JSON-over-stdio subprocess" bridges onto one shared low-level crate, so the transport layer is fixed and maintained once instead of three times.
Adds a **fourth consumer on top of the new transport** that demonstrates its value: an out-of-process memory extractor (see "A fourth consumer" below).

Before this PR the three bridges — channels `SidecarAdapter`, `plugin_runtime::HookProcessPool`, and `context_engine::SidecarContextEngine` — each re-implemented spawn + reply reading + lifecycle, and had drifted apart: inconsistent reply-line caps (none / 4 MiB-post-read / 16 MiB), `HookProcessPool` accepted `timeout_secs` but never enforced it, and the channels reader was unbounded.

## The shared crate: `crates/librefang-subprocess`

Depends on no `librefang-*` crate, so it sits below both `librefang-channels` and `librefang-runtime`.

- `read_capped_line` — bounded line read (a child streaming without a newline can't OOM the daemon). Bound is `AsyncBufRead + Unpin` so the per-byte read loop is always served from a `BufReader`'s in-memory buffer.
- `write_line_timeout` — a write+flush that's timeout-bounded (a child that stops reading its stdin can't wedge a caller).
- `SubprocessTransport` — the id-matched request/reply layer (background reader, `request(json) -> Result<Value, TransportError>`, `subprocess_transport_exited` metric + WARN, `kill_on_drop`) for the consumer whose protocol is id-matched.
- `SupervisedTransport` — a self-respawning wrapper around `SubprocessTransport` (lazy spawn, cooldown-rate-limited respawn) so a crashed sidecar recovers on a later call rather than staying dead until daemon restart.

The crate offers the low-level primitives separately because the three migrated consumers' protocols genuinely differ (id-matched request/reply vs positional request/reply vs event-stream) — only the line handling is truly common.

## The three migrations

1. **`context_engine::SidecarContextEngine`** → the id-matched `SubprocessTransport`. Sheds ~220 lines of duplicated transport; 5 engine tests pass through it. Also fixed a latent bug: a dead child now surfaces as `TransportError::Dead`, distinct from a `{"error": …}` reply (`Remote`).
2. **`plugin_runtime::HookProcessPool`** (positional protocol — replies carry no id, so it can't use `request`) → `read_capped_line` + `write_line_timeout`. Gains the per-call **timeout it never enforced** and an enforced-during-read 4 MiB cap. The configured `timeout` is now the *shared* wall-clock budget across write + read rather than 2× as it would be with split timeouts. Pooling, env_clear + seccomp, respawn, RSS, and state-file lock are unchanged. 20 plugin-runtime tests pass.
3. **channels `SidecarAdapter`** (event-stream + supervision + conformance corpus) → its inbound reader now uses `read_capped_line` via a thin `read_event_line` wrapper that preserves the exact `next_line()` shape its `select!` matches; an over-cap line (64 MiB) tears the connection down and the supervisor respawns. Its event protocol and supervision/respawn loop are untouched. 472 lib + 27 bridge-integration + 4 conformance-corpus + 3 doc-tests pass.

## A fourth consumer: out-of-process memory extractor

The new transport is reusable enough that one more consumer fits cleanly on top of it. The `MemoryExtractor` trait already existed and is the right seam — the store holds `Arc<dyn MemoryExtractor>` — so this adds `SidecarMemoryExtractor` (a `MemoryExtractor` impl backed by `SupervisedTransport`) without touching the storage or dedup paths.

**New operator-visible config surface:** `[proactive_memory.extractor_sidecar]` with `command`, `args`, `request_timeout_secs`. Precedence rules (now documented on the config struct itself):
- When present and `command` is non-empty, the sidecar handles every `auto_memorize` call AND **shadows the built-in LLM extractor + any per-agent `extraction_model` override**. The kernel emits a `WARN` at boot in that case.
- An empty `command` is treated as a misconfiguration (logged + ignored, falls through to the built-in path), so a half-edited TOML block doesn't silently spawn `""` every turn.

**What stays in Rust:** the store, the dedup `decide_action` decision, and the prompt-context formatting. The sidecar's output can only influence what's *proposed* to remember, never retrieval ranking or storage layout.

**Down-extractor degradation:** a missing / crashed sidecar surfaces as "nothing memorized this turn" (best-effort, post-turn path), and `SupervisedTransport` auto-respawns on the next call.

Ships a dependency-free Python reference (`docs/examples/memory_extractor_sidecar.py`) and 2 module-level tests (`extracts_via_sidecar`, `missing_sidecar_memorizes_nothing`) plus 2 init-path tests (`empty_sidecar_command_is_ignored_and_falls_through_to_llm`, `configured_sidecar_command_takes_precedence_over_llm`). The corresponding CHANGELOG entry is already in `[Unreleased]`.

## Verification (dev container)

- `cargo check` / `cargo clippy --no-deps -- -D warnings` / `cargo fmt --check` — clean across `librefang-subprocess`, `librefang-runtime`, `librefang-channels`, `librefang-types`.
- `cargo test -p librefang-subprocess` — 8 (incl. the new `supervised_cooldown_window_short_circuits_then_releases` that actually exercises the cooldown gating, not just "broken stays broken").
- `cargo test -p librefang-runtime --lib context_engine::sidecar` — 5; `… plugin_runtime` — 20; `… proactive_memory_sidecar` — 2.
- `cargo test -p librefang-channels` — 472 + 27 + 4 + 3 (incl. the conformance corpus and the real-subprocess bridge integration path).

Full `cargo check --workspace` left to CI (the operator's running daemon holds the shared target volume locally). The shared crate is additive; each migration is behaviour-preserving except the deliberate robustness improvements noted above.

## Net effect

A transport-layer fix now lands once. The series also closed real gaps: `HookProcessPool` gained timeout enforcement + an in-read cap (with a single shared deadline across write + read), channels gained a bounded reader, the context engine gained a clean Dead-vs-Remote distinction and an operator-visible exit metric, and the memory extractor gained an out-of-process option for operators who want to run extraction with their own LLM key / a cheap local model / embeddings without affecting storage or dedup.


houko:
Update: added a 4th commit (`5b1ba76f`) building on the shared crate — `SupervisedTransport`, a lazy self-respawning wrapper. `SidecarContextEngine` now auto-respawns a crashed sidecar (cooldown-rate-limited) instead of degrading to the built-in engine until daemon restart, removing the documented v1 limitation. +2 crate tests (respawn-after-exit, cooldown-blocks-storm); all prior tests still green. Branch now: extract+ContextEngine → HookProcessPool → channels → auto-respawn.

houko:
Update: 5th commit (`dfe1cda2`) — memory extraction can now run out-of-process via `SidecarMemoryExtractor` (implements the existing `MemoryExtractor` trait over `SupervisedTransport`). Selected by `[proactive_memory.extractor_sidecar]`; sidecar returns simple items and the daemon assigns id/created_at/source; store + dedup stay in Rust. A down sidecar degrades to no-op and auto-respawns. +Python reference + 2 tests (62 proactive_memory tests pass). Branch now: extract+ContextEngine → HookProcessPool → channels → auto-respawn → memory extractor.

PR #5937, opened 2026-05-31
feat(channels): ship librefang-sidecar-telegram binary in release tarballs

### LINKED ISSUE #5936: Ship librefang-sidecar-telegram binary in release assets
## Context

PR #5831 added the Rust Telegram sidecar adapter at `sdk/rust/librefang-sidecar-telegram/` with feature parity to the Python adapter. Great work!

However, in the beta.15 release assets (`librefang-aarch64-apple-darwin.tar.gz` etc.), only the main `librefang` binary is included — no `librefang-sidecar-telegram` binary, and `librefang` itself has no `sidecar` subcommand.

This means to actually use the Rust adapter today, users have to:
1. Install a Rust toolchain locally
2. Clone the repo
3. `cargo build --release -p librefang-sidecar-telegram`
4. Point `[[sidecar_channels]] command` at the resulting binary

## Ask

Could the release pipeline build and ship `librefang-sidecar-telegram` (and presumably future Rust SDK adapters) alongside the main `librefang` binary in the platform tarballs?

Example expected layout after `librefang update`:
```
~/.librefang/bin/
  librefang
  librefang-sidecar-telegram   # ← currently missing
```

That would make the migration from in-process `[[channels.telegram]]` (removed in beta.13) to `[[sidecar_channels]]` a one-step config change for operators who don't want a Python runtime in the stack.

## Why this matters for us

We're on a multi-bot setup (6 Telegram bots, 7 agents on Mac Mini) and considering the beta.12 → beta.15 upgrade primarily for #5814 (custom-URL TTS) and #5800 (auto_evolve ↔ skill_workshop routing). The Telegram migration is blocking because:
- We don't want a Python sidecar runtime as a new dependency
- Compiling Rust on the production host adds maintenance surface

Happy to test pre-release builds if helpful.

Thanks!

houko:
Closes #5936.

The Rust Telegram sidecar (#5831) shipped only as source, so using it meant installing a Rust toolchain, cloning the repo, building `librefang-sidecar-telegram`, and pointing a `[[sidecar_channels]]` block at the resulting path.
This bundles the binary inside the existing release tarballs so `librefang update` lands it at `~/.librefang/bin/` and the daemon auto-resolves it — making the migration off the Python sidecar a one-step config change.

## Part 1 — Release build (bundle inside existing tarballs)

Both workflows change in lockstep (`release-cli.yml` is the manual mirror of `release.yml`'s CLI jobs).
For each channel-capable **native** target the in-scope jobs are `cli_mac` (x86_64 + aarch64), `cli_linux` (x86_64-gnu + aarch64-gnu[cross]), `cli_musl` (x86_64 + aarch64, both cross), `cli_windows` (x86_64 + aarch64).
Per job:

- A new "Build Telegram sidecar" step builds the sidecar for the same target with `--manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml` (using `cross` where that job already uses cross — i.e. all of `cli_musl`, and the aarch64 leg of `cli_linux`).
- The Package step copies the sidecar binary (from its separate-workspace output `sdk/rust/librefang-sidecar-telegram/target/<target>/release/`) next to `librefang` and adds it to the **same** archive.
  macOS ad-hoc codesigns it like `librefang`; `cli_musl` verifies it is statically linked; Windows adds `librefang-sidecar-telegram.exe` to the `Compress-Archive` path list.

No new release assets are introduced — the binary ships inside the existing per-target `librefang-<target>.tar.gz` / `.zip`, so `EXPECTED_PLATFORMS=12` and the cosign `SHA256SUMS` manifest are unchanged.

**Deliberately skipped:** the `cli_android` job and all `mini` variants (Android and `mini` carry no channels).

## Part 2 — Install scripts (extract the bundled binary)

- `install.sh` (POSIX sh): after installing `librefang`, if `librefang-sidecar-telegram` is present in the extracted tarball, `chmod +x` it (and on macOS strip quarantine + ad-hoc codesign). Absent ⇒ skipped silently (backward compatible with older tarballs).
- `install.ps1`: mirror — if `librefang-sidecar-telegram.exe` is present, copy it to `$InstallDir`; absent ⇒ skipped.

## Part 3 — Daemon auto-resolution

`resolve_sidecar_command(command, home_dir)` in `crates/librefang-channels/src/sidecar.rs`, called once in `SidecarAdapter::new` (where `home_dir` is already available, so the resolved command flows into both the adapter field and the spawn `SpawnCtx`):

- Eligible only when `command` is empty or the bare stem `librefang-sidecar-telegram` (no path component).
- Resolution order, first hit wins: the daemon's own executable directory (`current_exe().parent()`), then `<home_dir>/bin/`, then fall through to the original command (PATH lookup — historical behavior).
- An absolute / relative path, or any other program (`python3`, `uv`, `./librefang-sidecar-telegram`, …), passes through unchanged so explicit operator intent always wins.

Three focused unit tests cover: resolves to `<home>/bin` when the bundled binary is present (bare stem and empty command); falls through unchanged when no bundled binary exists; leaves explicit commands untouched even when a bundled binary is on disk.

## Part 4 — Docs + CHANGELOG

- `docs/architecture/rust-telegram-sidecar.md` and `docs/src/app/configuration/channels/page.mdx` now state the binary ships in release tarballs, lands in `~/.librefang/bin/` via `librefang update`, and is auto-resolved — manual build no longer required. Added an "Auto-resolution" section documenting the search order.
- `CHANGELOG.md` entry added under the existing `[Unreleased]` `### Added`.

## Verification

Run via the repo's `Dockerfile.rust-dev` image (no native cargo on the host; Linux container, scoped commands only):

- `cargo test -p librefang-channels --lib` → **482 passed, 0 failed** (includes the 3 new `resolve_sidecar_command_*` tests, confirmed passing individually).
- `cargo build --release --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml` → compiles clean, binary produced (~8 MB).
- `cargo clippy -p librefang-channels --all-targets -- -D warnings` → clean.
- `cargo fmt -p librefang-channels -- --check` → clean.
- `sh -n web/public/install.sh` → syntax OK (shellcheck not available on the host).
- `release.yml` / `release-cli.yml` validated as YAML.

## Not locally verifiable — needs CI on a real tag

The **release pipeline itself cannot be exercised locally** (no tag, no GitHub Actions runners, and `cargo build` for the kernel is forbidden in-session).
The workflow edits — the cross-compiled sidecar build steps, the `cross --manifest-path` invocation for the separate workspace, the macOS codesign / musl static-link / Windows zip packaging, and the assertion that asset names and the `EXPECTED_PLATFORMS=12` / cosign manifest are unchanged — must be validated by CI on a real tag (or a `release-cli` manual dispatch).
The sidecar build was proven to compile for the container's Linux target only; per-target cross builds (aarch64, musl, Windows, macOS) are CI's responsibility.

## Update — PR-time CI guard (post-review)

Added a PR-time guard so the sidecar's compilability and the cross-compile risk are no longer deferred to CI-on-tag.

- `ci.yml`: new `changes` output `sidecar_rust` (set when `sdk/rust/librefang-sidecar{,-telegram}/` or the release workflows change), feeding a new `rust-telegram-sidecar` job modeled on the existing `skills-wasm-sdk` job.
  The job runs `cargo fmt --check` + `clippy -D warnings` + `test` + `build --release` on the sidecar workspace, then `cargo install cross` + `cross build --release --target aarch64-unknown-linux-musl --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml`.
  That cross step **really runs cross on this PR**, proving up front whether cross can mount and resolve the out-of-workspace `../librefang-sidecar` path dependency — the exact risk the original PR could only flag as "needs CI-on-tag validation".
  `rust-cache` lists both `sdk/rust/librefang-sidecar-telegram` and the sibling `sdk/rust/librefang-sidecar`.
- `release.yml` + `release-cli.yml`: the four native sidecar-building jobs (mac / linux / musl / windows) now cache the sidecar's separate `target/` by listing `.` and `sdk/rust/librefang-sidecar-telegram` under `rust-cache` `workspaces:` (mini / android untouched).

If the cross step fails on this PR's CI, that is the real answer we couldn't get locally — it would mean the sidecar workspace needs a `Cross.toml` / mount tweak so cross can see `../librefang-sidecar` (a follow-up).

Local verification (Docker dev image) of the native `rust-telegram-sidecar` step: `cargo fmt --check` clean, `clippy --all-targets -- -D warnings` clean, `cargo test` 41 passed, `cargo build --release` OK.
All three workflow YAMLs pass `yaml.safe_load` and `actionlint`.
The cross-musl step is intentionally CI-only.


houko:
## ⚠️ Highest-priority CI-on-tag risk: cross + sibling path dependency

The release-pipeline changes cannot be exercised by this PR's CI (release.yml only triggers on tag push). The single thing to validate before/at the next release:

`librefang-sidecar-telegram` is its own cargo workspace at `sdk/rust/librefang-sidecar-telegram/`, and it path-depends on a **sibling** separate workspace: `librefang-sidecar = { path = "../librefang-sidecar" }` (i.e. `sdk/rust/librefang-sidecar/`, which is **outside** the telegram workspace root).

For the **cross** targets (`aarch64-unknown-linux-gnu`, `x86_64/aarch64-unknown-linux-musl`), `cross build --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml` runs in a container that mounts the workspace root. Whether cross also mounts the sibling `../librefang-sidecar` path dep is version-dependent; if it doesn't, the cross builds fail with an unresolved path dependency. The repo-root `Cross.toml` does **not** apply here (cross resolves config relative to the built workspace, not the repo root).

Native targets (mac x86_64/aarch64, linux-gnu x86_64, windows x86_64/aarch64) are not affected — the sidecar was confirmed to compile on the host Linux target.

### De-risk without cutting a release
Trigger `release-cli.yml` via **workflow_dispatch on an existing tag** (its stated purpose) — it mirrors these exact build steps and will surface a cross failure on the sibling path dep without publishing a new release.

### Fallback if cross fails on the path dep
Give `sdk/rust/librefang-sidecar-telegram/` its own `Cross.toml` that mounts the parent (`sdk/rust/`), or set `CROSS_CONTAINER_OPTS` to mount the sibling, or vendor `librefang-sidecar` into the telegram workspace for the cross builds.

PR #5967, opened 2026-06-01
fix(kernel): cron day-of-week follows POSIX convention (0 and 7 = Sunday)

### LINKED ISSUE #5966: Cron day-of-week uses crate 1-7 mapping, not POSIX 0-7 — Sunday jobs rejected, weekday ranges shift
## Summary

LibreFang cron schedules interpret the day-of-week field with the `cron` crate's 1-7 convention rather than the POSIX/Vixie convention every operator expects (`0` and `7` both = Sunday, Mon=1 … Sat=6). The result is two operator-visible bugs:

1. **Sunday-only schedules are rejected.** An expression like `0 16 * * 0` (every Sunday 16:00) produces "no future fire time" and the job never persists/fires, because the `cron` crate (v0.16.0) defines `inclusive_min = 1` for the DOW unit and rejects `0`.
2. **Numeric weekday ranges silently shift by one day.** `0 11 * * 1-5` (intended Mon–Fri) actually fires Sun–Thu, skipping Friday, because `1` maps to Monday in POSIX but to Sunday in the crate.

## Root cause

LibreFang converts the user's 5/6-field cron expression to the 7-field form the `cron` crate requires at a single site (`compute_next_run_after_opt` in `crates/librefang-kernel/src/cron.rs`). That conversion copies the day-of-week field verbatim and performs no remap, so the POSIX value is handed straight to a crate that numbers the days differently.

`cron-0.16.0/src/time_unit/days_of_week.rs`: `inclusive_min() = 1`, `inclusive_max() = 7`, phf map `"sun" => 1 … "sat" => 7`.

## Impact

Any agent with a Sunday cron, or a numeric weekday range/list, is mis-scheduled. This was hit in production (a personal-automation agent lost all of its Sunday jobs and had weekday jobs firing a day early).

## Proposed fix

Remap the day-of-week field from POSIX (0-7) to the crate's (1-7) convention at the single conversion site, before the expression reaches the crate: `0`/`7` → Sunday, and shift `1-6` → `2-7`, expanding ranges/lists/steps. Named days (`sun`, `mon`, …) already match the crate and pass through. Malformed fields pass through verbatim so the crate still surfaces the validation error.

I have a fix + regression tests ready and will open a PR linking this issue.

DaBlitzStein:
## Summary

Cron day-of-week was passed to the `cron` crate (which numbers weekdays 1-7 with Sunday=1) without any remap from the POSIX/Vixie convention (0-7 with `0`==`7`==Sunday).
Two user-visible bugs resulted:

- Sunday-only schedules such as `0 16 * * 0` were rejected as unschedulable (the crate refuses `0`).
- Numeric weekday ranges such as `1-5` (Mon-Fri) silently shifted by one day, firing Sun-Thu.

The fix remaps the day-of-week field at the single 5/6→7-field conversion site (`remap_posix_dow_to_cron`) before the expression reaches the crate, mapping POSIX `0..7` onto the crate's `1..7` (`posix % 7 + 1`).
Wildcards, ranges, lists and step values are all handled; malformed input is passed through verbatim so the crate still produces its normal parse error.

Closes #5966.

## Changes

- `crates/librefang-kernel/src/cron.rs`: add `remap_posix_dow_to_cron`, call it on the dow field in the 5-field and 6-field conversion paths.
- `CHANGELOG.md`: entry under `[Unreleased]` → Fixed.

## Verification

- `cargo test -p librefang-kernel --lib` — the 8 affected cron tests pass:
  `remap_posix_dow_single_days_match_crate_convention`, `remap_posix_dow_passes_through_wildcards`,
  `remap_posix_dow_ranges_and_lists`, `remap_posix_dow_steps`, `remap_posix_dow_returns_malformed_verbatim`,
  `sunday_only_cron_is_schedulable`, `weekday_range_fires_friday_and_skips_sunday`, `test_compute_next_run_cron_with_dow`.
- `cargo fmt --check` clean (enforced by pre-commit hook).


houko (approved):
Reviewed — LGTM, approving.

Verified the premise and the fix end-to-end:
- The `cron` 0.16 crate numbers day-of-week `1-7` (Sun=1 … Sat=7) and rejects `0` (confirmed in its `time_unit/days_of_week.rs`: `inclusive_min()=1`, `inclusive_max()=7`). So the old `format!("0 {trimmed} *")` path genuinely broke `0 16 * * 0` (parse error → "no future fire time") and shifted `1-5` to the crate's Sun-Thu. Both bugs are real.
- `to_cron = posix % 7 + 1` is correct (0→1, 6→7, 7→1). Traced the range walk for `5-7`→{1,6,7}, `6-1`→{2,7,1}, `*/2`→{1,3,5,7}, `1-5`→{2,3,4,5,6} — all correct, including wrap-through-Sunday.
- `*` / `?` pass through (the crate accepts `?` as `All` in the dow field); named days pass through verbatim, so no double-remap against the crate's own name map; malformed input falls through to the crate's validator.
- `cron.rs:1014` is the only `parse::<cron::Schedule>()` site in the workspace, so the "single conversion site" claim holds. The separate `validate_cron_expr` (`librefang-types/src/scheduler.rs:940`) only does char-class + field-count validation and never parses with the crate, so it won't reject Sunday-`0` at job-creation time — no second gate to update.
- Good regression coverage: `sunday_only_cron_is_schedulable` and `weekday_range_fires_friday_and_skips_sunday`.

One non-blocking nit: a bare-number-with-step like `5/2` is treated as the single day `5` (span 0), whereas Vixie cron interprets `N/step` as `N-max/step`. Obscure and untested either way; a one-line doc note on the remap helper clarifying that `N/step` means just `N` would be enough if you want to be precise. No change required to merge.

PR #5987, opened 2026-06-02
fix(kernel): trigger evaluator self-deadlocks when per-event budget is exhausted (#5977)

### LINKED ISSUE #5977: kernel(triggers): evaluator self-deadlocks when per-event trigger budget is exhausted
## Summary

`TriggerEngine::evaluate_with_resolver` (`crates/librefang-kernel/src/triggers.rs`) self-deadlocks the moment the per-event trigger budget (`max_triggers_per_event`) is exhausted. Any agent that has more triggers matching a single event than the budget allows will hang the kernel's event-evaluation thread indefinitely — the storm-prevention path itself wedges the kernel.

## Root cause

The evaluation loop holds a `DashMap` shard **write** guard for the trigger it is currently visiting:

```rust
for id in ids {
    let Some(mut entry) = self.triggers.get_mut(&id) else { continue; }; // write-locks shard N
    ...
    if matches.len() >= self.max_triggers_per_event {
        warn!(
            ...
            total_registered = self.triggers.len(), // <-- read-locks EVERY shard, incl. N
            ...
        );
        break;
    }
    ...
}
```

`dashmap` 6.2.1 implements `len()` as `self.shards.iter().map(|s| s.read().len()).sum()` — it acquires a read lock on every shard. When it reaches the shard already write-locked by the live `entry: RefMut`, `read()` blocks forever (same-thread write-then-read on a non-reentrant `RwLock`). Deterministic deadlock; both the eval thread and the test harness thread park in `futex_do_wait` at 0% CPU.

## Reproduction

`kernel::triggers::tests::test_per_event_trigger_budget` (added in the same commit, #1656) registers 5 triggers matching `All` with a budget of 3. On the 4th match it enters the `warn!` branch and hangs. Observed wall-clock hang of 8h+ at 0% CPU under `cargo test -p librefang-kernel`.

## Fix

Snapshot the registered-trigger count **before** the loop takes any shard write-lock — the sorted `ids` vec is already that snapshot:

```rust
let total_registered = ids.len(); // lock-free, taken before the loop
for id in ids { ... warn!(..., total_registered, ...); ... }
```

PR incoming.

## Affected

`crates/librefang-kernel/src/triggers.rs`, present since #1656 (2026-03-26). Severity: kernel-level liveness — event evaluation hangs for any agent whose matching-trigger count exceeds `max_triggers_per_event`.

DaBlitzStein:
## Summary
Fixes #5977.

`AgentRegistry::evaluate_event` called `self.triggers.len()` inside the budget-exhausted `warn!` branch of its match loop, while a `self.triggers.get_mut(&id)` `RefMut` still held that shard's write-lock.
`DashMap::len()` read-locks every shard, so on the first high-fan-out event that hit `max_triggers_per_event` the evaluator self-deadlocked on a single thread.

The fix snapshots the registered-trigger count with a lock-free `ids.len()` before the match loop and uses that in the log line, so no shard read-lock is taken while a write-lock is held.

## Changes
- `crates/librefang-kernel/src/triggers.rs`: hoist `total_registered = ids.len()` before the loop; use it in the budget-exhausted branch instead of `self.triggers.len()`.

## Verification
- `cargo check -p librefang-kernel --all-targets` clean.
- Logic verified against current `main`: the `self.triggers.len()`-under-write-lock pattern is still present upstream.


houko (dismissed):
Reviewed the fix and the deadlock mechanism. Approving — the root cause is correct and the fix is sound.

What I verified:
- **Lock lifecycle**: in `evaluate_with_resolver` (`crates/librefang-kernel/src/triggers.rs:811`), `entry: RefMut` is bound for the whole loop body and `entry.value_mut()` is still borrowed at the `warn!` (lines 888-894), so the shard write-lock is provably held when the budget-exhausted branch runs.
- **Deadlock confirmed against the actual dep**: dashmap 6.2.1 (`Cargo.lock:1853`) implements `_len()` as `self.shards.iter().map(|s| s.read().len()).sum()` — it read-locks every shard, including the one `get_mut` already write-locked, on a non-reentrant `RwLock`. So the pre-fix `self.triggers.len()` self-deadlocked. The fix replaces it with `ids.len()` (a lock-free `Vec::len`) hoisted before the loop.
- **No new lock/race**: the change only removes a lock acquisition; no semaphores/permits/channels in this path, so nothing to leak.
- **Test**: `test_per_event_trigger_budget` registers 5 matching triggers with budget 3; the 4th match enters the `warn!` branch under the live `RefMut`, genuinely exercising the path that hung on `main`.

One non-blocking nit: the CHANGELOG entry and PR summary name the method `AgentRegistry::evaluate_event`, but the actual symbol is `TriggerEngine::evaluate_with_resolver` (`triggers.rs:785`) — there's no `AgentRegistry`/`evaluate_event` in that file (the linked issue gets it right). Worth correcting in the CHANGELOG since it ships, but not a merge blocker.

Issue #4807, opened 2026-05-08
feat(kernel): budget-aware fallback chain — skip exhausted providers, try next in chain


DaBlitzStein:
## Summary

When a provider's budget is exhausted, the pre-dispatch gate (#4800/#4801) rejects the entire call before the `FallbackDriver` gets a chance to try alternative providers. A Moonshot-exhausted agent with an ollama fallback configured gets blocked entirely instead of falling back.

## Proposed design

Inject a `ProviderBudgetChecker` trait (defined in `librefang-llm-driver`, the trait crate) into `FallbackChain`. Before each provider attempt, check its budget. If exhausted, skip to next provider. Stamp `actual_provider` on `CompletionResponse` so usage is recorded against the provider that actually served the request.

Key components:
1. `ProviderBudgetChecker` trait in `librefang-llm-driver` (avoids circular dep)
2. `FallbackChain` checks budget per-provider before calling driver
3. `KernelBudgetChecker` implements the trait using `MeteringEngine`
4. `actual_provider: Option<String>` on `CompletionResponse` → `AgentLoopResult` → `UsageRecord`
5. Remove the pre-dispatch provider gate from `messaging.rs` (the chain handles it internally now)

## Related

- #4800, #4801 (pre-dispatch gate — current fix, but blocks fallback)
- #4757 (closed — attempted `BudgetGatedDriver` but wiring was incomplete)
- #3736 (TOCTOU race — fixed with reserve/settle)

PR #5988, opened 2026-06-02
fix(metering): gate provider hourly token budget pre-call so exhaustion flags the fallback chain (#5980)

### LINKED ISSUE #5980: metering: provider hourly token budget is enforced post-call only; exhaustion-flag path (check_provider_budget) is never called
## Summary

A configured per-provider hourly token budget (`[budget.providers.<id>] max_tokens_per_hour`) never prevents a call from being made. Enforcement happens **only after** the provider HTTP response returns, and even when the post-call check returns `QuotaExceeded` the usage row is **recorded anyway** — so the rolling-hour counter keeps climbing far past the cap. The `MeteringEngine::check_provider_budget` method, which was intended (per #4807) to flag `ProviderExhaustionStore` for the fallback chain, is **never called** in the kernel/runtime hot path.

Observed live: a Moonshot agent with `max_tokens_per_hour = 1_200_000` sitting at **4,802,612** tokens in the rolling window, every call logging:

```
Resource quota exceeded: Provider 'moonshot' exceeded hourly token budget: 4802612 + 17987 / 1200000
```

The cap is effectively a no-op label.

- `crates/librefang-kernel-metering/src/lib.rs:626` — `check_provider_budget` defined (and calls `flag_provider_budget_exhausted`), but a repo-wide grep finds **zero** kernel/runtime call sites (only metering unit tests + an API comment).
- `crates/librefang-kernel/src/kernel/messaging.rs:424-437` — the pre-dispatch provider-budget gate was deliberately removed (cites #4807), in favour of `ProviderExhaustionStore` flagged by `check_provider_budget` + consumed by the fallback chain — but since `check_provider_budget` is never invoked, the store is never flagged.
- `crates/librefang-kernel/src/kernel/agent_execution.rs:1280-1303` (and the ephemeral path `messaging.rs:717-728`) — the only enforcement is `check_all_and_record` **after** the LLM call; on `QuotaExceeded` it logs and then `let _ = self.metering.engine.record(&usage_record)` records the row anyway.
- `crates/librefang-kernel-metering/src/usage.rs:692-710` — the token check uses a correct 1-hour sliding window (`timestamp > datetime('now','-1 hour')`), so this is **not** a missing-reset bug; the window is fine. The problem is purely that nothing gates the call before spend, and usage is recorded even on breach.

Config: `ProviderBudget` (`crates/librefang-types/src/config/types.rs:4612-4621`), `max_tokens_per_hour: u64` default `0` (unlimited), scope per-provider/global (`BudgetConfig.providers: BTreeMap<String, ProviderBudget>`).

## Proposed fix

Restore #4807's intended design rather than re-adding a hard pre-call error:
1. Call `check_provider_budget` pre-dispatch so it flags `ProviderExhaustionStore`; the fallback chain then routes to another provider (or surfaces a clean budget message) instead of blindly spending.
2. On a post-call `QuotaExceeded` from `check_all_and_record`, do **not** unconditionally `record()` — or, if accurate accounting of already-spent tokens is desired, record but ensure the pre-call gate prevents the next call so the counter cannot run 4× past the cap.

## Tests

- With `max_tokens_per_hour` set and the rolling window already at/over cap, the next dispatch is gated (flag set / fallback taken) before the provider HTTP send.
- A breaching call does not grow the counter unboundedly across consecutive ticks.

Discovered while diagnosing a runaway autonomous-tick agent (~18K tokens/tick every 2 min). Separate from the memory-loop issues #5978 / #5979.

DaBlitzStein:
## Summary

Restores the #4807 pre-dispatch per-provider budget design **and makes the flag load-bearing**, closing the no-op that @houko flagged in CHANGES_REQUESTED.

The original commit flagged an over-cap provider in the shared `ProviderExhaustionStore` but the flag was never consulted in the reported single-Moonshot-agent scenario, so the agent kept running ~4x past its hourly cap.

## Changes

**Commit 1 (`ec6c7b4`)** — the pre-dispatch gate: `flag_provider_budget_if_exhausted` runs `MeteringEngine::check_provider_budget` before dispatch so a breach flags the provider in the shared store. Called from the three dispatch sites (`agent_execution.rs`, `messaging.rs` ephemeral + streaming).

**Commit 2 (`4b1d83f`)** — wires the flag to the two paths that ignored it:

- **Multi-provider agent path** (`llm_drivers.rs` `resolve_driver`): the agent-fallback chain now carries each slot's provider name and builds the `FallbackDriver` via `with_models_and_providers(...)` with the **same** `MeteringEngine::exhaustion_store()` handle the gate flags (mirrors `boot.rs:698-714`). Previously it used `with_models(chain)` — no store, empty provider names — so `is_slot_exhausted` short-circuited to `false` and re-dispatched the exhausted provider. When the store is unwired the provider-less builder is kept (no regression).
- **Single-provider agent path**: there is no slot to skip to, so `flag_provider_budget_if_exhausted` now returns whether the provider is over cap, and the three dispatch sites hard-return `QuotaExceeded` when `resolve_effective_fallbacks(...)` is empty. A new pure helper `provider_exhausted_blocks_call(exhausted, manifest, cfg)` encapsulates that decision. When a fallback chain exists the call sites do **not** hard-return — they defer to the store-aware `FallbackDriver`, preserving the #4807 philosophy.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — 0 warnings.
- New `librefang-kernel` tests `kernel::tests::provider_budget_gate_5980` (3 passed):
  - `single_provider_over_cap_is_hard_blocked` — over-cap single-provider agent is blocked, store reports `BudgetExceeded`, and the gate stays blocking on re-check (counter never drains via a leaked dispatch).
  - `multi_provider_over_cap_defers_to_fallback_skip` — over-cap multi-provider agent is NOT hard-blocked; the `FallbackDriver` built over the same store handle pre-skips the exhausted `moonshot` slot and serves from healthy `groq` (`actual_provider == Some("groq")`).
  - `no_budget_configured_gate_is_noop` — no provider budget configured ⇒ gate returns false, store never flagged, never blocks.

Rebased onto current `origin/main` to clear the stale base.

Closes #5980


houko (dismissed):
Thanks for tackling this — the diagnosis in #5980 is solid and a pre-dispatch gate is the right shape. But as written the fix flags a store that the per-agent execution driver never reads, so it's a no-op for the reported scenario. Requesting changes.

**The flag is written, but nothing on the agent path consumes it**

`flag_provider_budget_if_exhausted` → `check_provider_budget` → `store.mark_exhausted(provider, BudgetExceeded, …)`. That store is only honored by `FallbackDriver`/`FallbackChain`, and only when (a) a store is attached via `with_exhaustion_store` and (b) the slot has a non-empty `provider_name`. The driver the agent loop actually runs comes from `resolve_driver`:

- With fallbacks configured, `crates/librefang-kernel/src/kernel/llm_drivers.rs:326-328` builds `FallbackDriver::with_models(chain)` — that constructor sets `exhaustion_store: None` (`fallback.rs:120`) and leaves every `provider_name` empty. Both `is_slot_exhausted` and `record_exhaustion` early-return on missing store / empty name (`fallback.rs:269-292`), so the flag is ignored.
- With no fallbacks, `llm_drivers.rs:331` returns the bare `primary` driver, which has no store awareness at all — exactly #5980's case (a single Moonshot agent).

The new comment ("`resolve_driver` reads the same store and skips the exhausted slot") describes the intended behavior, but the diff doesn't change `resolve_driver` to attach the store or set provider names, so it isn't true yet. To close the loop, `resolve_driver` needs to use `with_models_and_providers` (or `set_provider_names`) so slots carry provider names, **and** attach the shared `ProviderExhaustionStore` via `with_exhaustion_store` — including a single-slot wrapper for the no-fallback case so a solo over-cap provider surfaces a clean `AllProvidersExhausted` instead of overspending.

**Issue's proposed-fix item #2 not addressed**

`crates/librefang-kernel/src/kernel/agent_execution.rs:1303` still does `let _ = self.metering.engine.record(&usage_record)` unconditionally after a post-call `QuotaExceeded`, so the rolling counter keeps climbing (the 4.8M-over-1.2M symptom). Either drop the unconditional record on breach, or rely on the pre-call gate — but the gate has to be wired into a driver that reads it first.

**Test coverage**

No test added. The metering unit tests only prove the flag is *set*, not that the execution-path driver *honors* it — which is the whole bug. Per CLAUDE.md / #3721 this needs a test showing: provider at/over cap → next dispatch is skipped/failed-over before the HTTP send, and the counter doesn't grow unbounded across consecutive ticks.

Happy to re-review once `resolve_driver` carries the store + provider names through.

houko:
The pre-dispatch budget gate looks like a no-op on the normal configured-fallback path. The metering side flags exhaustion correctly, but the `FallbackDriver` that actually serves the request on that path is built without the store handle (and with empty per-slot provider names), so it can never read the flag and never skips the exhausted slot.

**Where the flag is set vs. where the driver is built**

```
crates/librefang-kernel/src/kernel/agent_execution.rs:954
    self.flag_provider_budget_if_exhausted(&manifest.model.provider);
crates/librefang-kernel/src/kernel/agent_execution.rs:957
    let driver = self.resolve_driver(&manifest)?;

crates/librefang-kernel/src/kernel/llm_drivers.rs:326
    return Ok(Arc::new(librefang_runtime::drivers::fallback::FallbackDriver::with_models(chain),));
```

`with_models` leaves the driver unwired, and `is_slot_exhausted` short-circuits to `false` on both guards:

```
crates/librefang-llm-drivers/src/drivers/fallback.rs:120   exhaustion_store: None,
crates/librefang-llm-drivers/src/drivers/fallback.rs:114   provider_name: String::new(),
crates/librefang-llm-drivers/src/drivers/fallback.rs:285   let Some(store) = &self.exhaustion_store else { return false; };
crates/librefang-llm-drivers/src/drivers/fallback.rs:288   if entry.provider_name.is_empty() { return false; }
```

The boot-time `default_driver` (boot.rs:705-714) *is* wired correctly, but `resolve_driver` only returns it in the narrow `claude_code + cli_profile_dirs` branch (llm_drivers.rs:151) or on a driver-creation-failure fallback (line 249) — never on the normal configured-fallback path this PR targets.

**Scenario:** operator sets `[budget.providers.groq] max_cost_per_hour_usd` with a `groq -> openai` fallback chain. groq usage crosses the cap. Next turn: `flag_provider_budget_if_exhausted("groq")` marks groq exhausted in the metering engine's store; `resolve_driver` returns a *fresh* `FallbackDriver::with_models(chain)` whose `exhaustion_store == None` and whose slots all have empty `provider_name`; on dispatch `is_slot_exhausted` returns `false` for groq on both guards, so `health_order` tries groq first and the over-cap provider is dispatched again — the exact overspend #5980 set out to fix. The store the gate flags is a different object from the (absent) store this driver reads.

**Suggested fix:** in `resolve_driver`, build the chain as `(driver, model, provider)` triples and construct via `FallbackDriver::with_models_and_providers(...).with_exhaustion_store(<shared ProviderExhaustionStore handle>)`, mirroring boot.rs:705-714 (the store is `Arc`-backed, so cloning preserves shared state). Both the store handle **and** non-empty per-slot provider names are required, or `is_slot_exhausted` short-circuits. The existing `check_provider_budget_flags_exhaustion_store` test only checks the metering engine's store, not the driver returned by `resolve_driver`, so it can't catch this — worth an integration test that asserts end-to-end skip behavior through the resolved driver.

_Reviewed at 04f70f029._

DaBlitzStein:
Rebased onto current `origin/main` to clear the stale base. On test coverage: the substantive budget-exhaustion logic exercised by this change lives in `MeteringEngine::check_provider_budget`, which already has unit coverage in `librefang-kernel-metering` (`test_check_provider_budget_under_limit`, `_over_limit`, `_zero_limit_skipped`, `_separate_providers_isolated`, `_tokens_per_hour`). The new `flag_provider_budget_if_exhausted` is a thin forwarding wrapper to that tested function (empty-provider guard + lookup + side-effecting call), and the call-site is pre-call wiring in `execute_llm_agent`/`messaging.rs`. Happy to add a kernel-level wrapper test if you'd prefer explicit coverage there.

houko:
Thanks for picking this up — the direction is right (restore the #4807 design and call `check_provider_budget` pre-call so it can flag exhaustion). But as written the gate is a **no-op in the single-provider case**, which is exactly the scenario #5980 reports (one Moonshot agent running ~4× over the hourly cap), and there are no tests. Two concrete gaps:

**1. The flag is written to a store nothing reads when there's no fallback chain.**
`flag_provider_budget_if_exhausted` → `check_provider_budget` → `flag_provider_budget_exhausted` writes to `MeteringEngine.exhaustion`. For that flag to actually skip a call, a `FallbackDriver` must read the same store via `is_slot_exhausted`. But boot only builds a store-aware `FallbackDriver` when the chain has >1 entry (`boot.rs:701-714`); the single-driver branch (`boot.rs:716-717`) returns a bare driver with **no exhaustion store and no slot to skip to**. So for a single-provider agent the flag lands in a store no driver consults, and there is nowhere to fall back even if it did — the call still goes out and the counter keeps climbing.

**2. Agent-level `fallback_models` builds a `FallbackDriver` without the store.**
`resolve_driver` constructs `FallbackDriver::with_models(chain)` (`llm_drivers.rs:326-329`) without `.with_exhaustion_store(...)`, so `is_slot_exhausted` short-circuits to `false` (`fallback.rs:285-287`); the slots also have empty `provider_name`, which the same function treats as not-exhausted. The flag is ignored on this path too.

The only path where the flag currently does anything is: default provider + a global `fallback_providers` chain of ≥2, going through boot's `len > 1` branch — which does not include the reported single-provider case.

**Suggested fix:** to stop the bleed even with no fallback, the flag-only approach isn't sufficient. When the gate hits and the resolved driver isn't a store-aware `FallbackDriver` with a healthy slot, the current call should hard-return `QuotaExceeded` pre-call rather than relying on a fallback slot that may not exist. The minimal form is: gate hit → return `Err(QuotaExceeded)`; treat "let the fallback chain take over on multi-provider" as a separate enhancement on top of that. (If you want to keep the flag-only path for multi-provider, also pass `.with_exhaustion_store(...)` + `with_models_and_providers(...)` in the `resolve_driver` `with_models` branch.)

**Tests (required, per the repo's integration-testing policy and #5980's acceptance items):** a kernel→driver e2e test, not just the metering-layer unit coverage, asserting (a) a single-provider agent over the hourly cap has its next dispatch blocked and the counter stops; (b) a multi-provider agent over the cap falls back to a healthy provider; (c) no budget configured → behaviour unchanged.

DaBlitzStein:
Thanks @houko — addressed both defects from your review. Pushed `4b1d83f` (rebased onto current `origin/main`).

**1. Multi-provider path now store-aware.** `resolve_driver` (`crates/librefang-kernel/src/kernel/llm_drivers.rs`) builds the agent-fallback chain as `(driver, model, provider)` triples via `with_models_and_providers(...)` and attaches the shared handle from `self.metering.engine.exhaustion_store()` — the same store the gate flags — mirroring `boot.rs:698-714`. So a budget-exhausted slot is now `record_skip`-ed on dispatch instead of re-invoked. When the accessor returns `None` the provider-less builder is kept, so the no-store case is unchanged.

**2. Single-provider path now hard-returns.** As you noted, the flag-only approach can't help when there is no slot to skip to. `flag_provider_budget_if_exhausted` now returns `bool` (over-cap?), and the three dispatch sites (`agent_execution.rs:~974`, `messaging.rs:~581` ephemeral, `messaging.rs:~2280` streaming) hard-return `QuotaExceeded` when `resolve_effective_fallbacks(...)` is empty, via the new pure helper `provider_exhausted_blocks_call`. When a fallback chain exists they defer to the now-store-aware `FallbackDriver` rather than hard-blocking, keeping the #4807 behaviour you wanted to protect.

**3. Kernel→driver e2e tests** (`kernel::tests::provider_budget_gate_5980`), covering your three required cases: single-provider over-cap blocked + counter stops; multi-provider over-cap skips the exhausted slot through the resolved driver and serves from the healthy one; no budget configured ⇒ unchanged. `cargo clippy --workspace --all-targets -- -D warnings` is clean.

One thing I want to flag for a possible follow-up (pre-existing, not introduced here): the gate keys on the raw `manifest.model.provider`, so an agent configured with `provider = "default"` (resolved to a concrete provider only inside `resolve_driver`) isn't matched against `[budget.providers.<id>]`. The reported case uses an explicit provider so it's covered; happy to thread the resolved provider through the gate in a separate change if you think it's worth it.

houko:
Re-reviewed the two follow-up commits — this resolves my earlier CHANGES_REQUESTED. The flag is now actually consumed on the agent path, both directions:

- **Multi-provider agents:** `resolve_driver` builds the chain via `with_models_and_providers(...)` (carrying each slot's provider name) and attaches the *same* handle from `MeteringEngine::exhaustion_store()` via `with_exhaustion_store(store)` — so the slot the gate flagged is now non-empty and `is_slot_exhausted` actually skips it, instead of the old `with_models` path that left the store unset and short-circuited to `false`. That was the exact no-op I called out.
- **Single-provider agents** (the reported single-Moonshot case, no slot to fail over to): `provider_exhausted_blocks_call` returns `exhausted && fallback-chain-is-empty`, and the three dispatch sites hard-return `QuotaExceeded` instead of dispatching past the cap.

The split is correct: block only when exhausted *and* there's no viable fallback; otherwise let the now-store-aware `FallbackDriver` do the skipping. The `provider_budget_gate_5980` tests cover both. The `match exhaustion_store() { None => fb }` arm also degrades safely when metering is off (the gate writes to the same store, so they stay consistent).

LGTM from my side — clearing my changes-requested. Nice fix.

houko (commented):
Two issues found during daily review pass:

**1. Misplaced doc comment and `#[allow]` attribute in `agent_execution.rs`**

The new functions are inserted after the existing context lines `/// Execute the default LLM-based agent loop.` and `#[allow(clippy::too_many_arguments)]`, so in the resulting file those two items are syntactically attached to `flag_provider_budget_if_exhausted` (not `execute_llm_agent`):

```rust
/// Execute the default LLM-based agent loop.          ← wrong: was execute_llm_agent's doc
#[allow(clippy::too_many_arguments)]                   ← wrong: was execute_llm_agent's allow
/// Pre-dispatch per-provider budget gate (#5980…)     ← flag_provider_budget_if_exhausted's doc
pub(crate) fn flag_provider_budget_if_exhausted(…) { … }

pub(crate) fn provider_exhausted_blocks_call(…) { … }

#[allow(clippy::too_many_arguments)]   ← correctly precedes execute_llm_agent
#[instrument(…)]
pub(super) fn execute_llm_agent(…) { … }   ← no doc comment
```

`execute_llm_agent` loses its `/// Execute the default LLM-based agent loop.` doc. `flag_provider_budget_if_exhausted` gains it (merged with the correct doc as a two-paragraph doc comment) plus an unnecessary `#[allow]`.

Fix: move the two new methods to before the `/// Execute the default LLM-based agent loop.` line (i.e. earlier in the `impl` block), or extract them to a dedicated `impl` block so the existing doc + attribute stay cleanly above `execute_llm_agent`.

**2. Duplicate CHANGELOG entry for `#5980`**

The diff adds two bullets that both reference `(#5980)`: a detailed one near the top of the changes in the hunk and a short one-liner further down. Remove the shorter duplicate.

---

CHANGELOG.md
houko, line 1: There are two `[Unreleased]` entries for the same fix (#5980):

1. The detailed one near the top of the hunk — `**metering(budget): the pre-dispatch per-provider budget gate now actually enforces the hourly cap…**` — describes the full root cause, two fixes, and the three regression test names.
2. A shorter duplicate further down — `**metering(budget): enforce per-provider hourly token/cost budgets pre-dispatch…**` — covers the same change in one line.

The shorter duplicate should be removed; the detailed entry is the correct one to keep. Having two entries for the same issue will also trip the `pre-commit` duplicate-`[Unreleased]` guard.

---

houko (commented):
One CHANGELOG issue to fix before merge: duplicate entry for #5980 (see inline comment on `CHANGELOG.md`).

---

houko (dismissed):
The metering logic looks good and the earlier review concerns are addressed — one small thing still blocks merge.

**CHANGELOG.md: duplicate `[Unreleased]` entry for #5980.** There are two bullets for this change:
- the detailed one (`the pre-dispatch per-provider budget gate now actually enforces…`) — keep this one
- a redundant one-liner further down (`enforce per-provider hourly token/cost budgets pre-dispatch…`) — please remove this duplicate

This is the item from the prior review that's still outstanding.

**Minor, while you're in there:** `execute_llm_agent` lost its `/// Execute the default LLM-based agent loop.` doc comment (it's present on `main`). Please restore it.

Once the duplicate CHANGELOG bullet is removed, this is good to merge.

Issue #5844, opened 2026-05-28
fix(kernel,dashboard): auto_evolve created skills must go through approval and auto-assign to creator agent


DaBlitzStein:
## Problem

When an agent with `auto_evolve = true` creates or updates a skill via `background_skill_review`:

1. **No approval gate for creates**: the skill is installed directly without appearing in the Pending queue for human review (unless `auto_evolve_mode = "controlled"` is configured, which is not the default on existing agents)
2. **No approval gate for updates/patches**: when the reviewer proposes an update or version bump to an existing skill, it applies directly — no diff review, no approve/reject
3. **Not added to creator's allowlist**: agents with `skills_mode = "allowlist"` create skills they can never use — the new skill is globally installed but not in the agent's `capabilities.skills` list
4. **Not visible in agent's Skills tab**: the agent detail Skills tab only shows assigned skills, with no section for "available but not assigned" skills

## Expected behavior

1. Auto_evolve creates → skill appears in Pending tab for approval
2. Auto_evolve updates/patches → skill version change appears in Pending with a diff view (old vs new) for approval
3. On approval, the skill is automatically added to the creating agent's allowlist (for new creates)
4. The agent's Skills tab shows a section "Available" with skills that exist globally but aren't assigned to this agent, with a + button to assign

## Current behavior

- Creates silently install globally
- Updates/patches silently apply in-place
- Agent's allowlist unchanged after create
- Agent cannot use its own creation
- No approval, no visibility, no diff review for version changes

### Kernel
- `approve_candidate` in `skill_workshop/storage.rs`: after promoting to active, call `registry.add_skill_to_agent(agent_id, skill_name)` to update the creator agent's allowlist
- Store `agent_id` in the candidate provenance (already done)
- For updates/patches in `controlled` mode: create a candidate with the diff (old_content vs new_content) so the reviewer can see what changed before approving

### Dashboard
- Pending tab: show diff view for update/patch candidates (old version vs proposed version)
- Agent Skills tab: add "Available" section below "Installed" showing unassigned global skills with assign/unassign buttons

Refs #5819, #5798

houko:
Triage / coverage cross-ref:

- **Part 1 (no approval gate for creates)** — partially addressed by #5828 (#5819): it adds an opt-in `auto_evolve_mode = "controlled"` that routes background-review *creates* through the pending queue. Note #5828 currently defaults to `free`, so the default behavior (direct install) is unchanged — it provides the mechanism, not a safer default.
[...]

PR #5990, opened 2026-06-02
feat(skills): auto-assign approved workshop skill to its creator agent

### LINKED ISSUE #5989: Approving a pending workshop skill does not assign it to the creating agent
## Summary

When an operator approves a pending skill-workshop candidate via `POST /api/skills/pending/{id}/approve` (or the CLI / dashboard equivalent), the candidate is promoted into the skills registry but is **not** added to any agent's skill allowlist.

The agent that actually taught the behaviour — and whose `agent_id` is recorded on the candidate — therefore cannot use the skill it produced until an operator manually edits that agent's `agent.toml` and adds the skill name to `skills = [...]`.

## Expected behaviour

Approving a candidate should append the promoted skill to the **creating agent's** allowlist, so the workshop loop (capture → review → approve → use) closes without a manual config edit. The approval should remain successful even if the agent can no longer be resolved (e.g. it was deleted between capture and approval) — auto-assign is a best-effort convenience, not a hard precondition.

## Current behaviour

`approve_pending_candidate` calls `reload_skills()` (so the skill is globally loadable) but never touches the agent registry. The candidate's `agent_id` field is available at the approval site but unused for assignment.

## Notes

- The creating agent id is already persisted on `CandidateSkill.agent_id`.
- A no-op when the skill is already present on the allowlist keeps re-approval idempotent.
- Assignment should clear `skills_disabled` so the newly added skill is actually live for that agent.

DaBlitzStein:
Closes #5989.

## Summary

Approving a pending skill-workshop candidate promoted it into the skills registry but never added it to any agent's allowlist, so the agent that taught the behaviour could not use the skill it produced until an operator hand-edited its `agent.toml`.
This closes the capture → review → approve → use loop.

## Changes

- `approve_pending_candidate` (`routes/skills/skill.rs`) reads the candidate's `agent_id` before promotion and, after `reload_skills()`, appends the new skill to that agent's allowlist.
The approval response gains an `auto_assigned_to` field echoing the creator agent id.
- New `KernelApi::add_skill_to_agent(agent_id, skill_name)` resolves the agent by name then UUID and delegates to the registry.
- New `AgentRegistry::add_skill(id, skill_name)` — append-only, no-op if the skill is already present, clears `skills_disabled`, notifies listeners.

## Behaviour notes

- A missing or empty creator id is a silent no-op.
- An add failure (e.g. the agent was deleted between capture and approval) is logged at WARN and does **not** fail the approval — auto-assign is best-effort.
- Re-approval is idempotent: `add_skill` no-ops when the skill is already on the allowlist.

## Verification

- Added `pending_approve_auto_assigns_skill_to_creator_agent` to `crates/librefang-api/tests/skill_workshop_pending_routes_test.rs`: registers an agent with an empty allowlist, stages a candidate carrying that agent's id, approves it over HTTP, and asserts both the `auto_assigned_to` response field and the registry side effect (skill appended, `skills_disabled` cleared).
- `cargo clippy -p librefang-api -p librefang-kernel --all-targets -- -D warnings` + `cargo fmt --check`: run in the consolidated verification pass.


houko (changes requested):
Traced capture → pending → approve → assign. The security model holds up well; the blocker is durability.

**Request changes — auto-assignment is in-memory only and is lost on restart, which defeats the PR's goal**

`KernelApi::add_skill_to_agent` (`crates/librefang-kernel/src/kernel_api.rs:1241`) calls only `AgentRegistry::add_skill`, which mutates the in-memory `DashMap` and fires `notify_changed()`. It never persists. Every other allowlist mutation in the kernel does two durable writes:
- `set_agent_skills` (`kernel/agent_state.rs:413-465`): `memory.substrate.save_agent(&entry)` (SQLite — the authoritative copy per the docstring at `agent_state.rs:34`, and the source boot reloads from at `kernel/boot.rs:1691`), plus rollback on persist failure.
- `set_agent_channels` (`agent_state.rs:532-559`): same, plus `persist_manifest_to_disk(agent_id)` for the `agent.toml` mirror.

This PR does neither. So after the next daemon restart, the agent is reloaded from SQLite with the original skill-less allowlist and is right back to needing a hand-edited `agent.toml` — the exact state the summary says this eliminates. Suggest routing the assign through `set_agent_skills` (or replicating its snapshot → `save_agent` → rollback-on-failure → `persist_manifest_to_disk` sequence) so it's durable.

The new test (`skill_workshop_pending_routes_test.rs:424`) only asserts the in-memory side effect, so it stays green while the user-visible behavior is broken. Please add coverage that the assignment survives a reload (assert the SQLite row / `save_agent`, mirroring the existing `set_agent_skills` tests).

**Verified correct (not issues):**
- Provenance is trustworthy: `candidate.agent_id` is kernel-set from a parsed `AgentId` (`skill_workshop/mod.rs:115`, `:283`), UUID-validated at every storage entry point — not user-controlled.
- Security boundary preserved: auto-assign runs only on the `Ok(result)` arm, *after* `approve_candidate` → `evolution::create_skill` re-runs `validate_name` + the Critical prompt-injection scan; it adds only the validated `result.skill_name` and modifies no skill content. No path lets an unscanned/unapproved skill reach an agent, and it targets only the capturing agent (no cross-agent escalation).
- No stale tool-cache: the preceding `reload_skills()` bumps `skill_generation`, invalidating the per-agent cache.
- Edge cases handled: empty/missing creator id → no-op; deleted agent → `AgentNotFound` logged, approval still succeeds; re-approval idempotent.

**Minor (worth folding in):**
- The candidate is read from disk twice on approve — `load_candidate` (`routes/skills/skill.rs:357`) and again inside `approve_candidate` (`storage.rs:557`). Consider returning the resolved `agent_id` from `approve_candidate` and dropping the extra read.
- `add_skill_to_agent` tries `find_by_name` before the UUID parse; since the stored id is always a UUID string, UUID-first resolution would be strictly safer.

houko:
Closing as superseded by #6003.

#6003 folds this auto-assign-approved-skill-to-creator behavior (#5844 part 3) into the larger skill-evolution-mode PR, and addresses the durability gap raised in review here — the assignment was in-memory only (`registry.add_skill`) and was lost on restart, since it never persisted via `save_agent` / the manifest mirror.

Thanks for this — closing in favor of the consolidated PR. If anything here wasn't carried over, please flag it on #6003. Not deleting the branch.

Issue #5979, opened 2026-06-02
runtime(loop_guard): a blocked tool call is a hard error that aborts the whole turn and panics the agent


DaBlitzStein:
## Summary

When the loop guard blocks a tool call (the agent issued the same `(tool_name, params)` enough times to hit `block_threshold`), the block is surfaced to the agent loop as a **hard error** (`is_error: true`, status `Error`). Because the loop-guard block message is not classified as a soft error, it triggers the "skipping remaining tool calls" path, **aborts the entire tool batch for that turn**, and counts toward the consecutive-all-failed exit. Three such iterations end the streaming loop and **record an agent panic**.

This means a blocked *duplicate read* (e.g. `memory_recall`) does not merely get suppressed — it crashes the whole turn. Observed live: 14 panics recorded against one agent in ~1 hour, each preceded by

```
Tool call blocked by loop guard tool=memory_recall
Tool execution failed — skipping remaining tool calls
Tool failures in 3 consecutive iterations — exiting streaming loop  consecutive_all_failed=3
```

## Root cause (source, `2026.5.31-beta.16`)

- `crates/librefang-runtime/src/agent_loop/tool_call.rs:485-500` — the `LoopGuardVerdict::Block` arm returns a `ToolResult` with `is_error: true`, `status: ToolExecutionStatus::Error`.
- `crates/librefang-runtime/src/agent_loop/message.rs:176-182` — `is_soft_error_content()` only matches path-traversal / sandbox-escape / "arguments were truncated" / missing-parameter strings. The loop-guard messages ("Blocked: tool '…' called N times with identical parameters" / "… returning identical results repeatedly") match none of these.
- `crates/librefang-runtime/src/agent_loop/mod.rs:1454-1466` — because the block is `is_error && !is_soft_error`, the batch is aborted ("skipping remaining tool calls") and the iteration is a hard failure; the consecutive-all-failed counter then reaches `MAX_CONSECUTIVE_ALL_FAILED` and the loop exits + panics.

The duplicate-detection key itself is correct — `SHA-256(tool_name | serde_json(params))` (`loop_guard.rs:537-544`), threshold 5 within one loop instance — so this fires only on genuinely identical repeats. The bug is the *handling*: a guard whose entire purpose is to gently stop a repeating call should not be a fatal hard error.

## Proposed fix

Treat a loop-guard block as a **soft** outcome:
[...]

Issue #2381, opened 2026-04-13
agent hand multiple call skill in pararel


diarmp:
### Description

when my agent multiple call skill in pararel i got error :
"error": {
    "message": "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages... — raw: API error (400): {
  "error": {
    "message": "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_4tnNgg5QAC70PdKMnRhQQedp, call_ytjfYZjOh1XQ61j3doAx2jXc, call_Lztf6DdtCAoK54U7s9qSqd5k, call_JXcYOA7Lmr9EPJ8wPiXJ03Wk, call_aTquIy01LA6wmduwpjnyz8Ou",
    "type": "invalid_request_error",
    "param": "messages",
    "code": null
  }

### Expected Behavior

agent can multiple call skill in pararel

### Steps to Reproduce

1. run librefang in docker
2. create hand with task call skill multiple
3. check log and you got this error

### LibreFang Version

latest

### Operating System

Other

### Logs / Screenshots

<img width="1444" height="304" alt="Image" src="https://github.com/user-attachments/assets/54cfd635-9448-4406-a6ed-f90301d8558b" />

<img width="1676" height="537" alt="Image" src="https://github.com/user-attachments/assets/c066a40b-ae4e-4ff0-9f20-6c177b1d5c51" />

PR #5991, opened 2026-06-02
fix(runtime): history fold preserves tool-result content on omit AND parse failure (#5978)

### LINKED ISSUE #5978: runtime(history_fold): omitted-summary stub causes the agent to re-issue the same tool call (memory_recall loop)
## Summary

When the history-fold summariser model omits one or more `tool_use_id`s from its JSON response, the affected tool-result blocks are overwritten with the static stub `[history-fold] [summarisation unavailable]`. The agent reads that placeholder as *"this tool's result is not available yet"* and **re-issues the identical tool call** on the next turn. For read tools that are called often — notably `memory_recall` — this produces an endless re-issue loop, which then trips the loop guard (see companion issue) and the agent effectively loses access to its own memory.

Observed live on a Moonshot-backed agent: every turn logged

```
history_fold: model omitted summaries for some stale tool_use_ids — those blocks fall back to the static stub  missing_ids=["memory_recall:22","memory_recall:23",…,"memory_recall:27"]
```

immediately followed by repeated `memory_recall` calls and `Tool call blocked by loop guard tool=memory_recall`.

`crates/librefang-runtime/src/history_fold.rs`:

- `:86` — `const FALLBACK_SUMMARY: &str = "[summarisation unavailable]";`
- `:277-288` — omitted ids are detected, a `warn!` is emitted, and the function returns `(map, None)` (no per-id fallback text).
- `:323-328` — apply loop:
  ```rust
  let summary = summaries_by_id
      .get(tool_use_id)
      .cloned()
      .or_else(|| bulk_fallback.clone())     // None on the happy path
      .unwrap_or_else(|| FALLBACK_SUMMARY.to_string());
  let new_content = format!("{FOLD_PREFIX} {summary}");
  ```

So an omitted id yields `[history-fold] [summarisation unavailable]` in place of the real prior result. Two things make this a re-issue attractor:
1. The original tool-result content is **discarded** rather than preserved, so the model has no record of what the call returned.
2. The wording "summarisation unavailable" reads as an unresolved/failed result, not a concluded tool call.

Why summaries get omitted: the fold prompt (`:528-544`) asks for one `{id, summary}` per stale block; models under token pressure silently return fewer entries (and id-normalisation mismatches — trailing whitespace / quoting — can also drop a real id into the wrong key, `:272-276`).

## Proposed fix

1. **Preserve original content for omitted ids** instead of dropping to `FALLBACK_SUMMARY`: when an id has no summary, fold to a preview-truncated copy of the *original* tool result (`[history-fold] [result preserved — summarisation omitted] <preview>`), so the model still sees what the tool returned and does not retry.
2. **Strengthen the fold prompt** to mandate one entry per id ("include one entry per id even if empty/error").
3. Change the genuine-failure stub wording to explicitly close the call (e.g. append "do not retry this tool call").

## Tests

- `omitted_ids_preserve_original_content` — response omits one id; assert the block keeps a substring of the original content and is NOT `summarisation unavailable`.
- `omitted_ids_do_not_carry_fallback_summary_string`.

Companion: the loop-guard hard-error/panic amplifier is filed separately. Repro environment: agent with long histories that trigger folding + a frequently-called read tool.

DaBlitzStein:
Closes #5978.

## Problem

The history-fold apply loop lost real tool-result content in two ways, and both drove an endless `memory_recall` re-issue loop on a live Moonshot/Kimi deployment (session 81620c15):

1. **Omitted ids.** When the aux summariser returned a valid batch but simply *omitted* one id, the apply loop substituted the `[summarisation unavailable]` fallback string for that result — discarding content that was still on hand.
2. **Total parse failure.** When the aux model returned a body that did not parse as the expected `[{id,summary}]` JSON, the loop dumped that raw, unparseable text verbatim over **every** stale tool result as a "bulk summary".

Either way, a recalled `memory_recall` result lost its content on fold, the agent saw garbage/"unavailable" where its answer used to be, and re-issued `memory_recall` — forever (until the loop guard blocked it; see the companion #5979).

## Relationship to #6009

#6009 (merged) fixes the *specific live trigger* of case 2: thinking-by-default fold models prepend a `<think>…</think>` block, which made the JSON parse fail. With `<think>` stripped, the kimi deployment no longer hits the parse failure.

This PR is complementary, not redundant: it makes the apply loop safe for **any** parse failure (truly malformed JSON, a different non-JSON preamble, a truncated body), so a future fold-model quirk can no longer dump raw text over every stale result and restart the loop. The omitted-id case (#1) is independent of #6009 entirely.

## Fix

Both paths now retain each stale block's **preview-truncated original content**:

- The omitted-id case keeps the original result instead of the `[summarisation unavailable]` stub.
- The Parse-failure arm is routed through that same content-preserving branch (empty map + no bulk fallback), so each block keeps its own truncated original instead of the raw non-JSON body. The now-dead `BatchSummariseFailure::Parse.raw` field is dropped.

Genuine network failures (Call/Empty) still use the static stub — there is no per-block original to recover there.

## Scope

`crates/librefang-runtime/src/history_fold.rs` only. The loop-guard hardening (`tool_call.rs`) is the companion PR #5992 (#5979). Rebased on current `main` (includes #6009); no textual overlap — #6009 changes the parser (`parse_labeled_summaries`), this changes the apply loop.

## Verification

Unit tests in `history_fold.rs`, all passing:
- `omitted_ids_preserve_original_content` / `omitted_ids_do_not_carry_fallback_summary_string` — omitted-id preservation.
- `parse_failure_preserves_original_per_block_content` / `parse_failure_does_not_apply_raw_response_as_bulk_summary` — parse-failure preservation (replaces the test that codified the old bulk-summary behavior).

`cargo test -p librefang-runtime --lib` (history_fold suite), `cargo clippy -p librefang-runtime --all-targets -- -D warnings`, `cargo fmt --check`: clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -651,8 +651,9 @@ _308 PRs from 7 contributors since v2026.5.17-beta.12._
   Regression tests: `router::tests::sidecar_default_does_not_collide_across_bots` (two sidecars register under distinct `telegram:<name>` keys, no last-writer-wins collision) and `sidecar::tests::test_sidecar_stamps_account_id_from_adapter_name` (a real sidecar subprocess stamps `account_id` from the config name, not the `ready`-event account, and preserves an adapter-supplied one).
 
 - **runtime(history-fold): preserve omitted tool-result content instead of substituting an "unavailable" stub** (#5978) (@DaBlitzStein).
-  The fold's apply loop unconditionally replaced every omitted tool result with the `[summarisation unavailable]` fallback string, so a recalled memory tool result lost its content and the agent re-issued `memory_recall` forever, an endless loop that drained tokens.
-  Omitted results now keep their preview-truncated original content, breaking the loop while still bounding the folded size.
+  The fold's apply loop discarded the real tool result in two cases — an id the model silently omitted from an otherwise-valid batch, and a response that could not be parsed as `[{id,summary}]` JSON at all (the latter was dumped verbatim over every stale block as a "bulk summary").
+  Either way a recalled memory tool result lost its content, the agent read it as "no answer yet", and re-issued `memory_recall` forever — an endless loop that drained tokens (verified live: a Moonshot/Kimi response that failed the JSON parse triggered exactly this).
+  Both cases now keep each block's preview-truncated original content, breaking the loop while still bounding the folded size; the raw unparseable response is no longer applied as a bulk summary.
 
 - **kernel(cron): day-of-week now follows the POSIX convention (`0` and `7` both mean Sunday)** instead of the `cron` crate's 1-7 mapping (#5966) (@DaBlitzStein).
   Sunday-only schedules like `0 16 * * 0` were previously rejected as unschedulable, and numeric weekday ranges such as `1-5` silently shifted by one day (firing Sun-Thu instead of Mon-Fri).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -650,6 +650,10 @@ _308 PRs from 7 contributors since v2026.5.17-beta.12._
   Registration key and resolution key now both derive from `SidecarChannelConfig::name`, so they line up.
   Regression tests: `router::tests::sidecar_default_does_not_collide_across_bots` (two sidecars register under distinct `telegram:<name>` keys, no last-writer-wins collision) and `sidecar::tests::test_sidecar_stamps_account_id_from_adapter_name` (a real sidecar subprocess stamps `account_id` from the config name, not the `ready`-event account, and preserves an adapter-supplied one).
 
+- **runtime(history-fold): preserve omitted tool-result content instead of substituting an "unavailable" stub** (#5978) (@DaBlitzStein).
+  The fold's apply loop unconditionally replaced every omitted tool result with the `[summarisation unavailable]` fallback string, so a recalled memory tool result lost its content and the agent re-issued `memory_recall` forever, an endless loop that drained tokens.
+  Omitted results now keep their preview-truncated original content, breaking the loop while still bounding the folded size.
+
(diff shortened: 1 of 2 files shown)


houko (dismissed):
Reviewed the full diff against `crates/librefang-runtime/src/history_fold.rs` and issue #5978. The fix is correct, well-scoped, and the pairing invariant is intact. Approving.

**Bug fix is correct**
- Root cause confirmed at the old apply loop (`history_fold.rs:323-328`): `.or_else(|| bulk_fallback.clone()).unwrap_or_else(|| FALLBACK_SUMMARY)`. On the Ok path `bulk_fallback` is `None`, so an id the summariser omitted fell straight to `[history-fold] [summarisation unavailable]`, discarding the real prior result — the exact `memory_recall` re-issue attractor in #5978.
- The new three-way branch resolves this: (1) per-id model summary, (2) `bulk_fallback`, (3) neither → preview-truncated copy of the original `content`. Branch 3 is reachable only when the batched call *succeeded* but the id was omitted — precisely the broken case.

**Provider contract (#2381) preserved**
- The rewrite stays in place inside the same `ContentBlock::ToolResult` arm; `tool_use_id`/`tool_name`/`is_error`/`status` untouched, no block dropped. Every `tool_use` keeps its matching `tool_result`.

**No bloat / re-fold regression**
- Content bounded by `FOLD_PREVIEW_CHARS` (500), so the trim still does its job; all branches keep the `[history-fold]` prefix so `is_already_folded` continues to recognize them.

**Tests are thorough**
- `omitted_ids_preserve_original_content` and `omitted_ids_do_not_carry_fallback_summary_string` both assert the original content survives AND the static stub is absent; the staleness setup is consistent with the existing `batched_call_invokes_llm_once_per_pass` assertion.

Minor (non-blocking): when an omitted block's original `content` is empty, branch 3 yields the prefix with a trailing space and no payload — harmless, and the strengthened fold prompt makes it unlikely in practice.

PR #5992, opened 2026-06-02
fix(runtime): loop-guard block is soft, and a persistent block stall degrades to a real reply (#5979)

### LINKED ISSUE #5979: runtime(loop_guard): a blocked tool call is a hard error that aborts the whole turn and panics the agent

DaBlitzStein:
Closes #5979.

## Problem

When the loop guard blocked a repeated `(tool, params)` call it returned a result with status `Error` — a hard error that aborted the rest of the tool batch and counted toward `MAX_CONSECUTIVE_ALL_FAILED`. Three consecutive blocks (e.g. an agent re-issuing an identical `memory_recall`) exited the streaming loop and recorded an agent panic instead of letting the model adjust.

Making the block soft is necessary but **not sufficient**: once the block is a soft `Skipped`, a model that keeps re-issuing the blocked call no longer panics — it spins to `max_iterations`, and the channel bridge sanitizes the resulting `MaxIterationsExceeded` into user-visible **silence**. No panic, but still no reply (observed live on a Moonshot/Kimi deployment).

## Fix (two parts, total dependency)

1. **Block is a soft outcome.** The block result is now `Skipped` (a soft status), so it no longer aborts the batch or trips the consecutive-all-failed exit. The block message still steers the model; the genuinely fatal runaway stays caught by the circuit breaker.

2. **Persistent block stall degrades to a real reply.** Both agent loops (streaming `run_streaming` and non-streaming `run_agent_loop`) now detect a *block-only* iteration — every tool result a soft loop-guard block, no success, no hard error, no assistant prose — via `ToolResultOutcomeSummary::is_block_only`, backed by a new `soft_error_count` tally. After `block_stall_degrade_after` consecutive such iterations (new `AutonomousConfig.block_stall_degrade_after`, default `2`, `None`/`0` disables) the next completion is issued with an empty tools vec. The OpenAI-compatible driver resolves `tool_choice` to `None` on empty tools, so the model is compelled to answer in prose. The forced turn finalizes through the normal end-turn path, preserving the tool_use/tool_result pairing invariant.

Part 2 has a total dependency on Part 1 (without the soft `Skipped`, a block is a hard `Error` and never reaches the block-only path), so they ship together.

## Changes

- `crates/librefang-types/src/agent.rs` — new `AutonomousConfig.block_stall_degrade_after: Option<u32>` (`#[serde(default)]`, default `Some(2)`), with `DEFAULT_BLOCK_STALL_DEGRADE_AFTER` const re-used by the runtime fallback.
- `crates/librefang-runtime/src/agent_loop/tool_call.rs` — `ToolResultOutcomeSummary` gains `soft_error_count`; `from_blocks` tallies soft `is_error:true` results (previously uncounted); new `is_block_only()` helper.
- `crates/librefang-runtime/src/agent_loop/run_streaming.rs` and `mod.rs` — block-stall detector + one-shot tools-stripped completion, mirrored in both loops.

## Verification

- New unit tests `tool_call::loop_guard_block_tests::{soft_block_counts_toward_soft_error_total, block_only_iteration_is_detected, a_success_alongside_a_block_is_not_block_only, a_hard_error_alongside_a_block_is_not_block_only, no_results_is_not_block_only, consecutive_block_only_reaches_degrade_threshold_then_resets_on_progress}` — all pass.
- `cargo check --workspace --lib`: clean.
- `cargo clippy -p librefang-runtime --all-targets -- -D warnings`: clean.
- `cargo fmt --check`: clean.


houko (dismissed):
Reviewed. The fix is correct, minimal, and well-targeted. Approving.

What I verified:
- **Single source of truth, both paths fixed.** The change lands in the unified `precheck_loop_guard` (`tool_call.rs:666-690`), consumed by both the serial path (`execute_single_tool_call_inner`, `tool_call.rs:713`) and the parallel/streaming group path (`tool_call.rs:506`). One typed `Error → Skipped` flip covers serial and streaming; `Skipped` is soft per `ToolExecutionStatus::is_soft_error()` (`librefang-types/src/tool.rs:53-55`).
- **Both downstream gates key off `is_soft_error()`.** Batch abort (`mod.rs:1470-1472`, streaming `run_streaming.rs:1254-1256`) and the consecutive-failure tally (`ToolResultOutcomeSummary::from_blocks`, `tool_call.rs:142`) both classify the block as soft, so it no longer aborts the batch nor increments `consecutive_all_failed`.
- **The runaway is still bounded.** `CircuitBreak` is unchanged (`tool_call.rs:664` → `Err(Internal)`), and `LoopGuard::check` enforces `global_circuit_breaker` (default 30 total calls) at the top of every check (`loop_guard.rs:176-183`) regardless of soft/hard status. A model that keeps re-issuing a blocked call still hits that absolute cap — the guard's purpose is preserved; only the premature 3-iteration kill is removed.
- **Provider contract intact (#2381).** The blocked result is staged via `process_executed` with its `tool_use_id` and `is_error: true`, and `pad_missing_results()` runs unconditionally — no dangling `tool_use`.
- **`Skipped` over `Denied`/`ModifyAndRetry` is the right pick** — the latter inject false guidance text; `Skipped` injects none and lets the block's own content steer the model.
- No production code branches on `status == Error` directly; all consumers use `is_error()`/`is_soft_error()`. CHANGELOG attribution correct.

One non-blocking nit: the issue's second requested test — a mixed batch (one blocked call + other pending calls) not aborting the rest — isn't asserted end-to-end; it's covered only transitively by the shared `is_soft_error()` check. Optional to add a batch-level case; I wouldn't gate the merge on it.

houko:
This is the right shape — typed soft status, not substring matching. The direction looks good to me.

Confirmations from reading it against current `main`:

- The single `Error` → `Skipped` change on the `LoopGuardVerdict::Block` arm (`tool_call.rs`) is enough for both kill paths, since both gate on `is_soft_error()`:
  - batch-abort: `agent_loop/mod.rs` (`is_error && !(status.is_soft_error() || …)`)
  - hard-failure tally: `ToolResultOutcomeSummary::from_blocks` in `tool_call.rs` (`!status.is_soft_error() && …`)
  `Skipped.is_soft_error() == true` (`librefang-types/src/tool.rs`) short-circuits both, so one typed source of truth fixes both — matching the PR description.
- `Skipped` over `Denied`/`ModifyAndRetry` is the correct call: those two inject guidance text in `append_tool_result_guidance_blocks` that would be false for a loop-guard block; `Skipped` injects none and the block's own content carries the steer. The `CircuitBreak` arm still aborts via `LibreFangError::Internal`, so the genuinely-fatal runaway is untouched.
- Same typed approach just landed in #5984 (c9e082c4) for memory/wiki ACL denials (`Denied`, also soft). The two are complementary — different producers, same `is_soft_error()` boundary, zero file overlap (that PR only touched `tool_runner/*`) — and notably this one does **not** lean on `is_soft_error_content` substring matching, which is the spoofable path we want to keep out.
- The three `loop_guard_block_tests`, including the pre-fix `Error`-reaches-3 regression guard, are exactly the right coverage.

One ask before merge: the branch is `BEHIND` `main` — please rebase onto current main (#5984 and #5967 landed after you branched). The code hunk still applies cleanly since the Block arm on main is unchanged; the only likely touchpoint is the CHANGELOG `Fixed` block.

houko:
The soft block-stall degrade strips tools from the *request*, but `recover_text_tool_calls` runs against the full, unchanged `available_tools` — so on the degrade turn a model that keeps emitting `<function=...>` prose has it re-parsed back into a `ToolUse`, re-executed, re-blocked into `Skipped`, and the loop never reaches the prose finalization the PR intends.

**The strip and the recovery use different tool lists**

```
crates/librefang-runtime/src/agent_loop/mod.rs:1085   tools: if force_tools_stripped {
crates/librefang-runtime/src/agent_loop/mod.rs:1086       std::sync::Arc::new(Vec::new())
crates/librefang-runtime/src/agent_loop/mod.rs:1170   let recovered = recover_text_tool_calls(&response.text(), available_tools);
crates/librefang-runtime/src/agent_loop/mod.rs:1176   response.tool_calls = recovered;
crates/librefang-runtime/src/agent_loop/mod.rs:1177   response.stop_reason = StopReason::ToolUse;
```

Same pattern in the streaming loop:

```
crates/librefang-runtime/src/agent_loop/run_streaming.rs:723   tools: if force_tools_stripped {
crates/librefang-runtime/src/agent_loop/run_streaming.rs:885   let recovered = recover_text_tool_calls(&response.text(), available_tools);
crates/librefang-runtime/src/agent_loop/run_streaming.rs:892   response.stop_reason = StopReason::ToolUse;
```

`available_tools` is a function parameter that is never reassigned; `force_tools_stripped` only gates `request.tools`, and it is reset to `false` (mod.rs:1113) *before* recovery runs.

**Scenario** (threshold=2, OpenAI-compatible model like Kimi that emits tool calls as text — the case the PR body itself cites):
1. Iter A: model emits `memory_recall(...)` → loop-guard Blocked → `Skipped` (tool_call.rs:708); `consecutive_block_only=1`.
2. Iter B: same → `consecutive_block_only=2` → degrade fires, `force_tools_stripped=true`.
3. Iter C: request goes out with `tools=[]`; `force_tools_stripped` reset to `false`; model emits the same `<function=memory_recall>{...}</function>` prose. `recover_text_tool_calls(text, available_tools)` — `available_tools` still has `memory_recall` — returns the synthetic call; lines 1176-1177 set `stop_reason=ToolUse`. The "forced prose" turn is converted back into a tool-use turn → Blocked → `Skipped`.
4. Iters D, E, … oscillate strip→recover→block→strip until `max_iterations`, where the channel bridge sanitizes `MaxIterationsExceeded` into the same user-visible silence #5979 set out to fix.

The unit test `consecutive_block_only_reaches_degrade_threshold_then_resets_on_progress` is pure arithmetic — it hard-codes that the stripped turn yields `EndTurn` and never builds a request or calls `recover_text_tool_calls`, so it cannot see the strip→recover→ToolUse transition.

**Suggested fix:** pass the request's *effective* tool list to `recover_text_tool_calls` (empty when stripped) so recovery returns nothing on a degrade turn and the `EndTurn`/prose path finalizes. Capture a per-iteration snapshot at request-build time (`let stripped_this_turn = force_tools_stripped;` / the same `tools` value), since the live flag is reset before recovery. Also add a real one-shot degrade latch so a model that produces empty prose on the stripped turn still breaks the loop with a fallback reply instead of re-arming. A `#[tokio::test]` driving the real loop with a mock driver returning `<function=...>` text on the stripped completion would lock this in.

_Reviewed at 3132a0b51._

Issue #5819, opened 2026-05-28
feat(skills): configurable skill evolution mode (free/controlled) + registry PR proposal


DaBlitzStein:
## Problem

Auto-evolve (background_skill_review) currently:
- Creates go to pending queue (#5800) but updates/patches apply directly
- No configurable approval mode — operator has no choice between auto-apply and review-first
- No mechanism to propose evolved skills back to the public registry
- Review prompt is too conservative — rarely proposes improvements to existing skills

### 1. Evolution mode per agent

```toml
[skill_workshop]
evolution_mode = "controlled"  # "free" | "controlled"
```

- **`free`**: creates, updates, and patches apply directly (current behavior for update/patch). Agent evolves autonomously.
- **`controlled`**: ALL mutations (create, update, patch) go through the pending queue for human approval. Nothing changes without explicit approve.

### 2. Version tracking in pending drafts

When auto_evolve proposes an update/patch to an existing skill, the pending draft should show:
- Current version of the skill
- Proposed changes (diff view)
- Changelog entry explaining why the change was proposed
- One-click approve that bumps the version

### 3. "Propose to Registry" button

Dashboard Pending/Installed skill detail gets a "Propose to Registry" action that:
- Opens a PR against `librefang/librefang-registry` (or configured registry repo)
- Includes the evolved skill TOML + prompt_context
- Auto-generates PR description with: what changed, why (from the LLM review), version diff
- Links back to the agent/session that triggered the evolution

### 4. Less conservative review prompt

The background_skill_review prompt should more aggressively propose improvements to existing skills when:
- The agent used a skill but had to work around a limitation
- The agent discovered a better approach than what the skill prescribes
- Error handling or edge cases were discovered during execution

## Current state

- Timeout increased 30s→120s (#5800) ✅
- Creates routed through pending (#5800) ✅  
- Version history exists in evolution.rs ✅
- approval_policy exists in skill_workshop config ✅

## What's needed

- [ ] `evolution_mode` config field (free/controlled) per agent
- [ ] Route update/patch through pending when mode=controlled
- [ ] Diff view in pending drafts for updates to existing skills
- [ ] "Propose to Registry" dashboard button + gh PR automation
- [ ] Review prompt improvements for existing skill refinement

PR #6003, opened 2026-06-03
feat(kernel): skill evolution_mode + gate auto_evolve updates through approval (#5844, #5819)

houko:
PR A of the combined plan for #5844 + #5819 (skill `auto_evolve` approval gating + evolution mode).
Kernel + types only — the dashboard diff / "Available" UI (PR B) and the "Propose to Registry" action (PR C) land separately.

Refs #5844, #5819 (not "Closes" — the issues are fully resolved only when B and C land).

### 1. `evolution_mode` config field (`free` / `controlled`)
`SkillWorkshopConfig` gains `evolution_mode: EvolutionMode` (`crates/librefang-types/src/agent.rs`), a typed `Copy` enum matching the sibling `ApprovalPolicy` / `ReviewMode` derive set + `#[serde(rename_all = "lowercase")]`, default `Free`.
Per-agent only — `agent.toml` / `HAND.toml [agents.<name>]`, never `config.toml` (#5476).
Field is `#[serde(default)]` and added to `SkillWorkshopConfig::default()`.

- **`free`** (default): today's behavior. A reviewer `create` queues for human approval (#5800); an `update` / `patch` to an already-approved skill applies directly.
- **`controlled`**: every mutation — create, update, patch — routes to the pending queue.

### 2. Route update/patch through pending in `controlled` mode
In `background_skill_review` (`crates/librefang-kernel/src/kernel/tools_and_skills.rs`) the `update` and `patch` arms now resolve `evolution_mode` from the agent manifest (`resolve_evolution_mode`) and, when `Controlled`, build an update candidate and route it through the SAME `save_candidate` path creates use, instead of calling `evolution::update_skill` / `patch_skill` directly.
For a `patch`, the full rewritten body is materialized first via `librefang_skills::evolution::fuzzy_find_and_replace`, so an update draft always carries a complete `prompt_context.md`.

**Security win (#5844):** the controlled update/patch now crosses `SkillVerifier::scan_prompt_content` inside `save_candidate` — a malicious LLM-proposed update can no longer bypass the injection scan by riding the direct path.

### 3. Pending-draft model extension for updates
`CandidateSkill` (`crates/librefang-kernel/src/skill_workshop/candidate.rs`) gains:
- `kind: CandidateKind` (`create` / `update`), `#[serde(default)]` → defaults to `Create`
- `target_skill_id` / `current_version` / `proposed_version` (`Option<String>`, `#[serde(default)]`)

All new fields are `#[serde(default)]`, so existing on-disk drafts (`~/.librefang/skills/pending/<agent>/<uuid>.toml`) still deserialize.
The diff is computed at display time (PR B) — this only stores the data.

### 4. Auto-assign to creator on approval (#5844)
New `LibreFangKernel::approve_pending_skill` (on the `KernelApi` trait) wraps `storage::approve_candidate` (which re-runs the injection scan via `evolution::create_skill`) + a registry mutation the storage layer cannot do on its own (no `AgentRegistry` handle).
On a CREATE approval it appends the new skill to the creating agent's `manifest.skills` via `AgentRegistry::update_skills`.
Idempotent (no double-add); an empty all-skills allowlist is left untouched so the agent's access is never narrowed; updates need no re-assignment.
`POST /api/skills/pending/{id}/approve` now calls this kernel method (the phantom-vs-collision `AlreadyInstalled` disambiguation branch is unchanged).

## Tests (scoped `cargo test -p librefang-kernel` / `-p librefang-api`; CI is the gate)

Kernel (`crates/librefang-kernel/src/kernel/tests.rs`):
- `resolve_evolution_mode_reads_per_agent_manifest` — per-agent resolution + `Free` fallback for unknown agent
- `controlled_update_queues_pending_and_leaves_skill_untouched` — controlled update lands in pending as `kind=Update` with target/version metadata; installed skill body + version unchanged
- `free_update_applies_directly_and_bumps_version` — default free path still applies directly (0.1.0 → 0.1.1), no pending draft
- `controlled_update_crosses_injection_scan` — a malicious update body is blocked by the same scan as creates; nothing reaches pending
- `approve_create_auto_assigns_to_creator_allowlist` — approve appends to allowlist + idempotent double-add
- `approve_create_leaves_all_skills_agent_unpinned` — empty allowlist stays empty

Candidate model (`candidate.rs`):
- `legacy_candidate_without_update_fields_defaults_to_create` — old drafts deserialize, `kind` defaults to create
- `update_candidate_round_trips_with_kind_and_target` — update draft TOML round-trip

API integration (`crates/librefang-api/tests/skill_workshop_pending_routes_test.rs`):
- `approve_create_auto_assigns_skill_to_creator_allowlist` — through `kernel.approve_pending_skill` (the path the route uses)

## Verification
No native Rust toolchain on this host (and the repo dev-image Docker is unavailable), so CI is the verification gate per `CLAUDE.md`.
Code was checked by hand for type/borrow correctness and rustfmt-shape; all literal `SkillWorkshopConfig {` / `CandidateSkill {` construction sites (incl. the two exhaustive API-test literals) were updated for the new fields.

## Known boundary (not a regression, called out for the reviewer)
The CLI `librefang skill pending approve` operates on the local `skills_root` filesystem and does not talk to the daemon, so it still calls `storage::approve_candidate` directly and cannot auto-assign to the live registry (the registry is in-process to the daemon).
Auto-assign happens on the daemon/API approve path, which is where #5844 needs it.
Giving the CLI a daemon HTTP client is a separate domain and out of scope for this PR.

---
Refs #5844
Refs #5819


houko (commented):
CHANGELOG continuation-line indentation fixed in a follow-up commit (lines 626-630 were not indented, so they rendered as standalone paragraphs rather than continuation text of the list item).

Two issues that need addressing before merge:

**1. Wrong HTTP status (409) when the Update target skill is missing**

`approve_pending_candidate` has a generic catch-all:
```rust
Err(WorkshopError::Skill(e)) => (StatusCode::CONFLICT, format!("promotion rejected: {e}")),
```

Before this PR, `approve_candidate` only took the `Create` path, where `WorkshopError::Skill(SkillError::AlreadyInstalled)` is the expected conflict.
After this PR, the `Update` approval path can emit `WorkshopError::Skill(SkillError::NotFound(...))` when the target skill was deleted between candidate capture and approval.
That gives the API consumer a `409 "promotion rejected: Skill 'foo' not found"` — the wrong status code and a misleading message.

This should be a `422 Unprocessable Entity` (or a new `WorkshopError` variant) with a message like "target skill 'foo' no longer exists — reject this candidate and re-capture once the skill is reinstalled."

A targeted arm above the catch-all is sufficient:
```rust
Err(WorkshopError::Skill(SkillError::NotFound(name))) => (
    StatusCode::UNPROCESSABLE_ENTITY,
    format!("target skill '{name}' no longer exists; reject this candidate and re-capture"),
),
Err(WorkshopError::Skill(e)) => (StatusCode::CONFLICT, format!("promotion rejected: {e}")),
```

**2. Missing HTTP-level integration test for the route change (CLAUDE.md #3721)**

`POST /api/skills/pending/{id}/approve` was changed from calling `storage::approve_candidate` directly to calling `state.kernel.approve_pending_skill`.
The new test `approve_create_auto_assigns_skill_to_creator_allowlist` calls `kernel.approve_pending_skill` directly on the kernel struct — it does not exercise the HTTP route via `TestServer` + `reqwest`.

CLAUDE.md (refs #3721) requires: "Add a `#[tokio::test]` against `TestServer`... hit the endpoint with `reqwest`, assert status and response shape."

A test that seeds a Create candidate, `POST`s to `/api/skills/pending/{id}/approve` via `reqwest`, and asserts the 200 + allowlist side-effect would close this gap.
An HTTP-level test for the Update approval path (seed a `CandidateKind::Update` candidate, install the target skill, approve via HTTP, assert the skill body changed) would also cover the new routing in `storage.rs`.


---

houko:
Both points addressed in 6cf49e5f9.

**1. 422 for a missing Update target** — `approve_pending_candidate` now has a targeted arm for `WorkshopError::Skill(SkillError::NotFound)` returning `422` with `kind: "target_skill_missing"`, placed above the generic `Skill(e)` → 409 catch-all (`crates/librefang-api/src/routes/skills/skill.rs`).
One deviation from the suggested snippet: `SkillError::NotFound`'s payload is a human-readable detail string, not a bare skill name (`load_installed_skill_from_disk` builds `"Skill '{name}' directory does not exist"` — `crates/librefang-skills/src/evolution.rs:1836`), so the response embeds it as a detail rather than re-quoting it as a name.
The pending file is left in place (the kernel errors before the cleanup step), and the test asserts that.

**2. HTTP-level integration tests** — added in `skill_workshop_pending_routes_test.rs`:
- `pending_approve_update_rewrites_target_skill_body` — seeds the target via a real create-approve over HTTP, stages a `CandidateKind::Update` candidate, POSTs `/api/skills/pending/{id}/approve`, asserts 200 + `prompt_context.md` rewritten + patch version bump to 0.1.1 + pending file dropped.
- `pending_approve_update_missing_target_returns_422` — stages an Update candidate whose target was never installed, asserts 422 + `kind` + pending file preserved.

For the Create path specifically: `pending_approve_assigns_skill_to_creating_agent` (same file, pre-existing) already POSTs the approve route over the TestServer harness and asserts the 200 + allowlist append, and it now exercises the new `kernel.approve_pending_skill` routing since the route was switched — so the HTTP-level Create coverage requested is held by that test rather than a new duplicate.

Verified locally: `cargo test -p librefang-api --test skill_workshop_pending_routes_test` → 24 passed, 0 failed.

PR #6011, opened 2026-06-04
fix(runtime): tolerate <think> preamble in history_fold summary parsing (#6009)

### LINKED ISSUE #6009: history_fold: JSON parse fails when aux model produces <think> preamble (deepseek-v4-flash default thinking)
## Symptom

`history_fold` consistently degrades to bulk-summary fallback when `deepseek-v4-flash` (or any model where thinking is on by default) is configured as the fold aux model. Every fold pass emits:

```
WARN history_fold: JSON parse failed — applying raw response as bulk summary
     count=N error=JSON parse failed: expected value at line 1 column 1
```

The bulk-summary fallback corrupts the fold entry in session context — the agent reads a raw thinking block as if it were a structured summary and develops incorrect behavioral patterns in subsequent turns.

## Root cause (read from current source)

`summarise_batch` in `crates/librefang-runtime/src/history_fold.rs` builds a `CompletionRequest` with `thinking: None` (line ~596), then passes the raw response text to `parse_labeled_summaries`:

```rust
// history_fold.rs ~L632
let raw = response.text();
parse_labeled_summaries(&raw).map_err(|error| BatchSummariseFailure::Parse { raw, error })
```

`parse_labeled_summaries` calls `strip_code_fence` (which handles `` ```json `` / `` ``` `` wrappers) but has **no handling for `<think>…</think>` preambles**:

```rust
// history_fold.rs L642-645
fn parse_labeled_summaries(text: &str) -> Result<BTreeMap<String, String>, String> {
    let body = strip_code_fence(text.trim()).unwrap_or_else(|| text.trim().to_string());
    let value: serde_json::Value =
        serde_json::from_str(&body).map_err(|e| format!("JSON parse failed: {e}"))?; // ← fails here
```

When deepseek-v4-flash is the aux model, `response.text()` returns:

```
<think>
…(reasoning block, hundreds of tokens)…
</think>
[{"id":"toolu_xyz","summary":"…"},…]
```

`strip_code_fence` does not match this prefix, so `serde_json::from_str` receives `<think>` at column 1 and returns `expected value at line 1 column 1`.

## Why `thinking: None` does not help

The fold `CompletionRequest` (line ~596) explicitly sets `thinking: None`, meaning LibreFang does **not** request extended thinking. However, for deepseek-v4-flash the provider config notes *"Thinking mode is on by default"* — the API enables thinking regardless of whether the caller requests it. The driver has no `thinking: disabled` signal to send to suppress it. `parse_labeled_summaries` is therefore the correct place for the defensive strip, just as `strip_code_fence` already handles another model-output quirk (markdown fences).

## Proposed fix

**Minimal — strip `<think>` block in `parse_labeled_summaries` before JSON parse:**

```rust
fn parse_labeled_summaries(text: &str) -> Result<BTreeMap<String, String>, String> {
    let trimmed = strip_think_block(text.trim());
    let body = strip_code_fence(trimmed.trim())
        .unwrap_or_else(|| trimmed.trim().to_string());
    let value: serde_json::Value =
        serde_json::from_str(&body).map_err(|e| format!("JSON parse failed: {e}"))?;
    // …rest unchanged
}

/// Strip a leading `<think>…</think>` block if present (DeepSeek v4-flash and
/// similar models that emit reasoning unconditionally). Returns a `&str` slice
/// starting after `</think>` so no allocation is needed when the tag is absent.
fn strip_think_block(s: &str) -> &str {
    if let Some(end) = s.find("</think>") {
        s[end + "</think>".len()..].trim_start()
    } else {
        s
    }
}
```

This mirrors the existing `strip_code_fence` approach and is consistent with the comment at line ~640: *"Tolerates a single markdown code-fence wrapper around the JSON array because some providers (notably reasoning-tier models) still emit fenced output even when told not to."* `<think>` preambles are the same category of provider quirk.

**Alternative (deeper fix):** propagate a `disable_thinking` flag into the aux `CompletionRequest` when the resolved model has `supports_thinking = true`, so the driver can send the appropriate API-level signal. This would address the root at the request level rather than the parse level, but requires changes across more crates.

## Setup (for reference)

- v2026.x-beta.x  
- Aux model: `deepseek:deepseek-v4-flash` (or any model where provider emits `<think>` unconditionally)  
- `[llm.auxiliary] fold = ["deepseek:deepseek-v4-flash"]`  
- **Workaround:** switch fold to a non-thinking model (e.g. `deepseek:deepseek-chat`). Works around the symptom but loses the cost/quality profile of v4-flash for fold.

Happy to test a patch.

Refs: #4842 (reasoning_echo_policy), #5203 (token budget / EOF truncation), #4866 (batched fold persistence)

houko:
## Root cause

`history_fold` consistently degraded to the bulk-summary fallback when a
thinking-by-default model (e.g. `deepseek-v4-flash`) was the fold aux model.

`summarise_batch` builds its `CompletionRequest` with `thinking: None`
(`history_fold.rs` ~L596), but for these models the provider enables extended
thinking server-side regardless — there is no `thinking: disabled` signal to
send. The response text therefore arrives as:

```
<think>
…(reasoning block)…
</think>
[{"id":"toolu_xyz","summary":"…"}]
```

`parse_labeled_summaries` ran `strip_code_fence(text.trim())` then
`serde_json::from_str`. `strip_code_fence` only matches markdown fences, so
`<think>` reached the JSON parser at column 1 and `from_str` failed with
`expected value at line 1 column 1`. `summarise_batch` then fell back to the
raw-bulk-summary path, which corrupts the fold entry: the agent reads a raw
thinking block as if it were a structured per-tool summary.

## Fix

A minimal, parser-side defensive strip — the same shape as the existing
`strip_code_fence` model. New helper `strip_think_preamble` removes a single
*leading* `<think>…</think>` block (case-insensitive tag, possibly multi-line,
non-greedy on the first `</think>`) plus surrounding whitespace, then parsing
proceeds as before.

Order in `parse_labeled_summaries`: **strip think → strip code fence → parse**,
so it is robust to:
- no think block (no-op, plain JSON still parses),
- `<think>…</think>` directly followed by JSON,
- `<think>…</think>` followed by a ` ```json ` fence,
- whitespace / newlines between any of the above.

An unterminated `<think>` (truncated response) is left unchanged so the JSON
parse surfaces the failure as before rather than silently swallowing the whole
response.

The request construction (`thinking: None`, `response_format`, token budget)
and the fallback semantics are **unchanged** — only the parser is made
tolerant.

## Tests

Added to the existing `tests` module in `history_fold.rs`, following the
`strip_code_fence_*` patterns:

- `strip_think_preamble_removes_leading_block`
- `strip_think_preamble_is_case_insensitive`
- `strip_think_preamble_noop_without_block`
- `strip_think_preamble_leaves_unterminated_block`
- `parse_labeled_summaries_tolerates_think_preamble`
- `parse_labeled_summaries_tolerates_think_preamble_then_json_fence`
- `parse_labeled_summaries_still_parses_plain_json`

## Verification

- No new dependency (plain string ops, matching `strip_code_fence`).
- Not compiled/tested locally (no native toolchain in this environment); CI is
  the gate for `cargo check --workspace --lib`, clippy, and the runtime unit
  tests.

Closes #6009

### DIFF
--- a/crates/librefang-runtime/src/history_fold.rs
+++ b/crates/librefang-runtime/src/history_fold.rs
@@ -1732,15 +1732,17 @@ mod tests {
         let input = "<think>\nThe tool listed files. I'll summarise each one.\n</think>\n\
                      [{\"id\":\"toolu_1\",\"summary\":\"Listed files.\"}]";
         let out = parse_labeled_summaries(input).expect("should parse past <think> preamble");
-        assert_eq!(out.get("toolu_1").map(String::as_str), Some("Listed files."));
+        assert_eq!(
+            out.get("toolu_1").map(String::as_str),
+            Some("Listed files.")
+        );
     }
 
     #[test]
     fn parse_labeled_summaries_tolerates_think_preamble_then_json_fence() {
         let input = "<think>reasoning across\nmultiple lines</think>\n\
                      ```json\n[{\"id\":\"toolu_2\",\"summary\":\"Read config.\"}]\n```";
-        let out = parse_labeled_summaries(input)
-            .expect("should strip think block then json fence");
+        let out = parse_labeled_summaries(input).expect("should strip think block then json fence");
         assert_eq!(out.get("toolu_2").map(String::as_str), Some("Read config."));
     }
 
--- a/crates/librefang-runtime/src/history_fold.rs
+++ b/crates/librefang-runtime/src/history_fold.rs
@@ -640,12 +640,8 @@ async fn summarise_batch(
 /// because some providers (notably reasoning-tier models) still emit
 /// fenced output even when told not to.
 fn parse_labeled_summaries(text: &str) -> Result<BTreeMap<String, String>, String> {
-    // Thinking-by-default models (e.g. deepseek-v4-flash) prepend a
-    // `<think>…</think>` reasoning block even when the fold request sets
-    // `thinking: None` — the provider enables it server-side and we have
-    // no signal to suppress it (#6009).  Strip that preamble before the
-    // fence/JSON parse so the trailing JSON array is the parsable body
-    // instead of `<think>` arriving at column 1.
+    // Thinking-by-default models (e.g. deepseek-v4-flash) prepend a `<think>…</think>` reasoning block even when the fold request sets `thinking: None` — the provider enables it server-side and we have no signal to suppress it (#6009).
+    // Strip that preamble before the fence/JSON parse so the trailing JSON array is the parsable body instead of `<think>` arriving at column 1.
     let stripped = strip_think_preamble(text.trim());
     let body = strip_code_fence(stripped.trim()).unwrap_or_else(|| stripped.trim().to_string());
     let value: serde_json::Value =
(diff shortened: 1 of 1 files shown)


houko (commented):
Prose-wrapping fixed in a follow-up commit (one sentence per line per CLAUDE.md).

Three smaller observations for the author:

**1. `None` branch returns `s` instead of `trimmed` (minor inconsistency)**

In `strip_think_preamble`, the `Some` arm returns content sliced from `trimmed` (leading whitespace stripped), while the `None` arm returns `s` (the original with any leading whitespace intact).
The caller passes `text.trim()` into `strip_think_preamble`, so there is no functional bug here.
But the divergence is surprising — a reader expects the two arms to be symmetric.
Consider returning `trimmed.to_string()` in the `None` arm, or adding a comment explaining the intentional asymmetry.

**2. No test for mixed-case closing tag `</THINK>`**

The doc-comment says the match is case-insensitive, and `lower.find("</think>")` does handle `</THINK>` correctly (because `lower` lowercases both tags before searching).
There is a test for `<THINK>reasoning</THINK>` (mixed-case open), but no test where specifically the *closing* tag is uppercase (`<think>…</THINK>`).
A test would be cheap insurance and complete the case-insensitivity story promised by the doc-comment.

**3. `<think>` with attributes will silently fall through**

`lower.starts_with("<think>")` does not match `<think type="cot">` or any tag with attributes.
In practice no known provider adds attributes to this tag, so this is not a live bug.
A brief comment in the function acknowledging the limitation (or a looser check like `lower.starts_with("<think") && lower[6..].trim_start().starts_with('>')`) would make future-proofing visible.


---

Issue #4842, opened 2026-05-09
[Bug] deepseek-v4-flash lacks multi-turn reasoning_content handling — fails after tool calls


nevgenov:
### Summary

`deepseek-v4-flash` defaults to **thinking mode enabled**. When the agent performs tool calls, DeepSeek returns `reasoning_content` that MUST be preserved in subsequent requests. LibreFang's OpenAI driver does NOT preserve it for V4 Flash — only for `deepseek-reasoner`/R1 and Kimi.

**Result**: agent works for simple messages, but as soon as a tool is called, subsequent requests fail with 400/500 errors. Agent becomes unusable until session reset.

### Reproduce

1. Configure an agent with model `deepseek-v4-flash`
2. Send simple messages → ✅ works
3. Send a message that triggers a tool call (e.g., `memory_search`, `ask_agent`, `shell_exec`)
4. All subsequent messages → ❌ 500 / "Sorry, something went wrong"

Switching to `deepseek-chat` (V3, no thinking mode) or `deepseek-reasoner` (R1, handled) works fine.

### Root cause

In `crates/librefang-llm-drivers/src/drivers/openai.rs`:

**1. `is_deepseek_reasoner()` — only matches R1, not V4 Flash:**
```rust
fn is_deepseek_reasoner(&self, model: &str) -> bool {
    let m = model.to_lowercase();
    m.contains("deepseek-reasoner") || m.contains("deepseek-r1")
    // deepseek-v4-flash is NOT matched
}
```

**2. `reasoning_content` is set to `None` for V4 Flash:**
```rust
reasoning_content: if is_deepseek_r {
    None                                    // R1: strips it (correct)
} else if has_tool_calls
    && self.kimi_needs_reasoning_content(...) {
    Some(String::new())                     // Kimi: empty string (correct)
} else {
    None                                    // V4 Flash: ignored → BUG
},
```

**3. `thinking` mode is never disabled for V4 Flash:**
```rust
thinking: if self.kimi_needs_reasoning_content(&request.model) {
    Some(serde_json::json!({"type": "disabled"})) // only Kimi
} else {
    None                                          // V4 Flash: thinking ON → BUG
},
```

Per [DeepSeek API docs](https://api-docs.deepseek.com/guides/thinking_mode#tool-calls):
> "for turns that do perform tool calls, the `reasoning_content` must be fully passed back to the API in all subsequent requests. If your code does not correctly pass back `reasoning_content`, the API will return a 400 error."

### Expected behavior

[...]

Issue #4745, opened 2026-05-07
Feature request: Manually editable model capabilities


FrantaNautilus:
### Description

Current implementation of model capabilities relies on the provider declaring the capabilities in one of the format that are supported (`capabilities` field and model name). However, the `capabilities` field is not a universal stamdard among providers. Covering all possible edge cases is complicated, therefore it would be good to have means to configure the capabilities per model in the config file and/or in the UI.
The models are already present in the config files in form of model field of agents (and fallback models), so this system of models written to config files would not change the state management of the LibreFang.

### Alternatives Considered

Only possible alternative is to use AI gateway to wrap the provider. However, `capabilities` are not universally available even among AI gateway programs.

### Additional Context

Closely connected with other issue: [https://github.com/librefang/librefang/issues/4702](https://github.com/librefang/librefang/issues/4702).

PR #6013, opened 2026-06-04
fix(runtime,kernel): redact images for text-only models via catalog supports_vision (#6010)

### LINKED ISSUE #6010: Bug: OpenAI driver serializes image blocks as image_url without checking model vision capability (text-only models return HTTP 400)
## Symptom

A Telegram bot (`bot-A`) routes to an agent (`agent-A`) backed by a **text-only** model (DeepSeek `deepseek-v4`, `supports_vision = false`). When a user sends a photo, the bot replies:

```
Agent error: LLM driver error: Request failed:
{"error":{"message":"Failed to deserialize the JSON body into the target type:
messages[30]: unknown variant `image_url`, expected `text` at line 1 column 147537"}}
```

Column ~147537 is the inlined base64 of the downscaled image — i.e. the driver embedded the image as a `data:` URL into the chat-completions request sent to a model whose API only accepts text content parts.

This reproduces for **any** sidecar/channel bot whose default agent runs on a text-only OpenAI-compatible model. Vision-capable models are unaffected.

## Root cause (read from current `main`)

`crates/librefang-llm-drivers/src/drivers/openai.rs`, `build_request()` (lines ~900-924) serializes both image block variants into `image_url` **unconditionally** — there is no `supports_vision` gate:

```rust
// crates/librefang-llm-drivers/src/drivers/openai.rs:900
ContentBlock::Image { media_type, data } => {
    parts.push(OaiContentPart::ImageUrl {
        image_url: OaiImageUrl {
            url: format!("data:{media_type};base64,{data}"),
        },
    });
}
ContentBlock::ImageFile { media_type, path } => {
    match tokio::task::block_in_place(|| std::fs::read(path)) {
        Ok(bytes) => {
            let data = base64::engine::general_purpose::STANDARD.encode(&bytes);
            parts.push(OaiContentPart::ImageUrl {
                image_url: OaiImageUrl {
                    url: format!("data:{media_type};base64,{data}"),
                },
            });
        }
        Err(e) => { warn!(path = %path, error = %e, "ImageFile missing, skipping"); }
    }
}
```

The **only** image-aware preprocessing is Moonshot-gated:

```rust
// crates/librefang-llm-drivers/src/drivers/openai.rs:1131
if self.is_moonshot() {
    self.preprocess_moonshot_files(&mut request).await?;
}
```

So for a non-Moonshot, non-vision model (DeepSeek, Mistral, etc.) the image block reaches the API verbatim as `image_url`, and the provider rejects the request with HTTP 400 `unknown variant image_url, expected text`.

`CompletionRequest` (`crates/librefang-llm-driver/src/lib.rs:245`) carries only `model: String` — no capability flag — so `build_request` has no per-request vision signal today; the driver relies on name heuristics (`is_moonshot()`, `is_deepseek_*`).

## Why the inbound `[media]` description does not save it

The channel layer already downloads the photo to disk and calls the inbound image-description path, which **prepends** a textual description to the message. But it does **not strip** the `ContentBlock::Image`/`ImageFile` block — so even when a description is present, the original image block is still serialized to `image_url` and the 400 still fires.

## Proposed fix

Generalize the existing Moonshot precedent (`preprocess_moonshot_files`, `openai.rs:241`, which already replaces eligible image/file blocks with a text marker) to **all non-vision targets**: when the target model is not vision-capable, replace `ContentBlock::Image`/`ContentBlock::ImageFile` with a text marker (e.g. the inbound `[media]` description, or a `[image omitted: <media_type>]` placeholder) instead of emitting `OaiContentPart::ImageUrl`.

Capability metadata for exactly this routing already exists from #4702 (LlmDriver capability metadata + capability-aware FallbackChain routing for vision/image_file) — plumbing a `supports_vision` signal into `build_request` (or gating the image arms behind it) would close the gap.

**Minimal:** in the `ContentBlock::Image`/`ImageFile` arms of `build_request`, when the model is not vision-capable, push a `Text` marker instead of `ImageUrl`.

## Setup (for reference)

- v2026.5.31-beta.16
- Pattern: channel/sidecar bot whose default agent uses a text-only OpenAI-compatible model (`supports_vision = false`); user sends an inbound photo.
- Inbound image description is enabled (`[media] image_description = true`), but the image block is still serialized → 400.

Happy to test a patch.

Refs: #4702 (capability metadata / fallback routing precedent)

houko:
## Problem

A channel/sidecar bot whose default agent runs a **text-only** model (e.g.
DeepSeek `deepseek-v4`, `supports_vision = false`) breaks the moment a user
sends a photo. The OpenAI-compatible API rejects image content parts with
HTTP 400:

```
unknown variant `image_url`, expected `text`
```

The channel layer downloads the image and prepends a textual `[media]`
description, but it does **not** strip the `ContentBlock::Image` / `ImageFile`
block, so the image still reaches the driver, is inlined as a `data:` URL, and
400s the whole turn. This affects every entry path with a text-only default
agent (WebUI, channels, sidecars, triggers).

## Approach (authoritative, catalog-driven)

The previous revision of this PR used a driver-local name deny-list in
`openai.rs`. That was rejected by the maintainer as non-authoritative. This
revision moves the gate to the model catalog and redacts upstream of the
driver, so it covers **all** entry paths and **all** OpenAI-compatible drivers,
respects user capability overrides, and never hard-codes provider names.

LibreFang's authoritative vision flag is the model catalog's
`effective_capabilities().supports_vision`, which already honours user
overrides (#4745). This is surfaced to request-build time via the same
`CatalogQuery` trait that #4842 used to thread `reasoning_echo_policy`.

## Changes

1. **`CatalogQuery::supports_vision_for(model) -> bool`** — new trait method
   mirroring `reasoning_echo_policy_for`, with a default body returning `true`
   (fail open) so non-overriding mocks/stubs keep sending images unchanged.
   `crates/librefang-kernel-handle/src/catalog_query.rs`

2. **Kernel override** — resolves the flag from the catalog, honouring user
   overrides via `effective_capabilities`, fails open on a catalog miss
   (`unwrap_or(true)`), using the same catalog accessor as
   `lookup_reasoning_echo_policy` and the same call shape as `ws.rs`.
   `crates/librefang-kernel/src/kernel/handles/catalog_query.rs`

3. **Agent-loop gate** — at the request-build site in `run_agent_loop`, look up
   `supports_vision` via the kernel handle (same `kernel.as_ref().map(|k| …)`
   pattern as the adjacent `reasoning_echo_policy` lookup). When
   `!supports_vision`, redact image blocks to a text placeholder via a small
   pure helper `redact_images_for_text_only(messages, model)` **before**
   building the request. The vision path is byte-identical to before.
   `crates/librefang-runtime/src/agent_loop/mod.rs`

4. **`openai.rs` reverted** to `main` — the driver again serializes images
   unconditionally; the gate now lives upstream.

### Why this is correct

- **Authoritative**: capability comes from the catalog
  (`effective_capabilities().supports_vision`), not a name heuristic.
- **Respects user overrides (#4745)**: `effective_capabilities` applies the
  per-model override, so an operator who force-enables vision on a
  mis-declared model keeps sending images.
- **Covers all paths**: redaction happens in the shared agent loop, so every
  inbound surface (WebUI, channels, sidecars, triggers) is protected, not just
  the OpenAI driver / a known family list.
- **Fails open**: no kernel handle wired, or a catalog miss → images are sent
  unchanged, so vision and unknown models are never degraded.

## Tests

- `crates/librefang-kernel-handle/src/tests.rs` —
  `catalog_query_supports_vision_defaults_to_true`: a stub that doesn't
  override the method returns `true` (mirrors the existing
  `catalog_query_default_returns_none` test).
- `crates/librefang-runtime/src/agent_loop/tests/utilities.rs` —
  `redact_images_for_text_only_replaces_image_and_imagefile_blocks`: both
  `Image` and `ImageFile` blocks become text placeholders that mention the
  model; text/other blocks and block counts are preserved.
  `redact_images_for_text_only_is_noop_without_images`: messages without image
  blocks pass through structurally unchanged.

## Verification

Local compilation/tests were not possible in this environment (no native
`cargo`/`rustc`, no reliable Docker). The change follows the existing
`reasoning_echo_policy_for` precedent and `ws.rs` catalog-call shape, and is
left for CI to compile and run as the gate.

Closes #6010

### DIFF
--- a/crates/librefang-kernel-handle/src/catalog_query.rs
+++ b/crates/librefang-kernel-handle/src/catalog_query.rs
@@ -24,6 +24,22 @@ pub trait CatalogQuery: Send + Sync {
         librefang_types::model_catalog::ReasoningEchoPolicy::None
     }
 
+    /// Whether the given model supports vision (image) input, resolved from
+    /// the model catalog's effective capabilities (#6010). Consulted at
+    /// request-build time to decide whether image content blocks may be sent
+    /// to the model or must be redacted to a text placeholder first —
+    /// text-only OpenAI-compatible models reject image content parts with
+    /// HTTP 400 (`unknown variant image_url, expected text`).
+    ///
+    /// Default impl returns `true` (fail open) so non-overriding mocks and
+    /// stubs keep sending images unchanged. The real kernel impl applies user
+    /// capability overrides (#4745) via `effective_capabilities` and also
+    /// fails open on a catalog miss, so vision-capable models are never
+    /// degraded.
+    fn supports_vision_for(&self, _model: &str) -> bool {
+        true
+    }
+
     /// Resolve the effective proactive-memory `extraction_model` for the
     /// agent identified by `agent_id` (#5475). Looks at the agent's
     /// manifest `[proactive_memory] extraction_model` and falls back to
--- a/crates/librefang-kernel-handle/src/tests.rs
+++ b/crates/librefang-kernel-handle/src/tests.rs
@@ -233,3 +233,15 @@ fn catalog_query_default_returns_none() {
         ReasoningEchoPolicy::None
     );
 }
+
+#[test]
+fn catalog_query_supports_vision_defaults_to_true() {
+    // #6010: a stub/mock that doesn't override `supports_vision_for` must
+    // report `true` (fail open), so test fixtures and any handle without a
+    // catalog wired keep sending image content blocks unchanged. Only the
+    // real kernel impl, which consults the model catalog, can return `false`.
+    let stub = StubKernel;
+    assert!(stub.supports_vision_for("deepseek-v4"));
+    assert!(stub.supports_vision_for("gpt-4o"));
+    assert!(stub.supports_vision_for("anything-else"));
+}
--- a/crates/librefang-kernel/src/kernel/handles/catalog_query.rs
+++ b/crates/librefang-kernel/src/kernel/handles/catalog_query.rs
@@ -26,11 +26,28 @@ impl LibreFangKernel {
             .map(|entry| entry.reasoning_echo_policy)
             .unwrap_or_default()
     }
+
+    /// Inherent mirror of [`kernel_handle::CatalogQuery::supports_vision_for`]
+    /// (#6010). Resolves the model's effective vision capability from the
+    /// catalog, honouring user capability overrides (#4745) via
+    /// `effective_capabilities`. Fails open (`true`) on a catalog miss so an
+    /// unknown / user-defined model is never silently stripped of image input.
+    pub(crate) fn lookup_supports_vision(&self, model: &str) -> bool {
+        let catalog = self.model_catalog_ref().load();
+        catalog
+            .find_model(model)
+            .map(|m| catalog.effective_capabilities(m).supports_vision)
+            .unwrap_or(true)
+    }
 }
 
 impl kernel_handle::CatalogQuery for LibreFangKernel {
     fn reasoning_echo_policy_for(&self, model: &str) -> ReasoningEchoPolicy {
         self.lookup_reasoning_echo_policy(model)
+    }
+
+    fn supports_vision_for(&self, model: &str) -> bool {
+        self.lookup_supports_vision(model)
     }
 
     /// Resolve the per-agent `extraction_model` for proactive memory
--- a/crates/librefang-runtime/src/agent_loop/mod.rs
+++ b/crates/librefang-runtime/src/agent_loop/mod.rs
@@ -295,6 +295,39 @@ fn build_sender_prefix(manifest: &AgentManifest, sender_user_id: Option<&str>) -
     Some(format!("[{}]: ", sanitize_sender_label(raw)))
 }
 
+/// Replace every `Image` / `ImageFile` content block in `messages` with a
+/// short text placeholder, leaving all other blocks and messages untouched
+/// (#6010).
+///
+/// Called only when the target model has no vision support. Text-only
+/// OpenAI-compatible models reject image content parts with HTTP 400
+/// (`unknown variant image_url, expected text`), which previously broke any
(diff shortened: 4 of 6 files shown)


houko (commented):
**Streaming path was unguarded — pushed a fix.**

`run_agent_loop_streaming` built `CompletionRequest` from `messages.clone()` directly, skipping the `supports_vision_for` check added to the non-streaming path.
Any production deployment using the WebUI or channel bots (which run the streaming loop) would still receive HTTP 400 on image content with a text-only model.

Fix applied: `redact_images_for_text_only` is now `pub(super)` and the streaming loop mirrors the non-streaming vision gate before constructing `CompletionRequest`.

---

**Advisory: `lookup_supports_vision` inlines what `effective_capabilities_for` already provides**

In `crates/librefang-kernel/src/kernel/handles/catalog_query.rs`, the implementation:
```rust
catalog.find_model(model).map(|m| catalog.effective_capabilities(m).supports_vision).unwrap_or(true)
```
duplicates the two-step logic that `catalog.effective_capabilities_for(model)` already encapsulates (that method exists at `model_catalog.rs:735`).
Using it would reduce this to a one-liner:
```rust
catalog.effective_capabilities_for(model).map(|c| c.supports_vision).unwrap_or(true)
```
Not a bug, but inconsistency with the existing API surface means a future change to `effective_capabilities_for`'s logic would need to be mirrored here manually.

---

**Advisory: `supports_vision_for` is re-evaluated inside the iteration loop**

`api_model` is constant across loop iterations, but `supports_vision_for` does a fresh `ArcSwap::load()` on each call.
A catalog hot-reload mid-turn could flip vision support between iterations: early iterations send images, later ones redact them, and the session history becomes inconsistent.
The `reasoning_echo_policy` lookup has the same pattern, so this is not unique to this PR, but it is worth documenting the "snapshot per iteration" semantics or hoisting both lookups before the loop for consistency.


---

PR #6014, opened 2026-06-04
fix(skills): assign approved workshop skill to the creating agent (#5989)

### LINKED ISSUE #5989: Approving a pending workshop skill does not assign it to the creating agent

houko:
## Summary

Closes #5989.

Approving a pending skill-workshop candidate (`POST /api/skills/pending/{id}/approve`, and by extension the CLI / dashboard, which share this route) promoted the skill into the registry via `reload_skills()` but never touched any agent's allowlist. The teaching agent — whose id is recorded on `CandidateSkill.agent_id` — therefore could not use the skill it produced until an operator manually edited `agent.toml` and added the name to `skills = [...]`.

This PR closes the workshop loop (capture → review → approve → use) by auto-assigning the promoted skill to its creating agent on approval.

## Root cause

`approve_pending_candidate` in `crates/librefang-api/src/routes/skills/skill.rs` called `state.kernel.reload_skills()` (global loadability) but had no follow-up that read `CandidateSkill.agent_id` and appended the skill to that agent's `manifest.skills`. The `agent_id` was available on the candidate but unused for assignment.

## Fix

In `approve_pending_candidate`:

- Read `CandidateSkill.agent_id` via `storage::load_candidate` **before** calling `approve_candidate` (which deletes the pending TOML on success, after which `agent_id` is unrecoverable). Best-effort: a read failure yields `None` and does not block the approve.
- After promotion + `reload_skills()`, call a new best-effort helper `assign_skill_to_creator`:
  - `None` / unparseable id / agent no longer in the registry → log a `WARN` and return; the approve still returns 200 (auto-assign is a convenience, not a precondition).
  - skill already on the allowlist AND `skills_disabled` is already false → no-op (no redundant DB write).
  - otherwise append the skill (idempotent — no duplicate if already present) and route through the existing `kernel.set_agent_skills`, which clears `skills_disabled` so the skill is live, persists the manifest (`save_agent` upsert), and invalidates the agent's cached tool list.
- The phantom-pending recovery branch (`already_promoted`) also re-runs the assign so re-approval converges to the same end state (e.g. the agent didn't exist on the first pass but was recreated later).

Reuses the existing `kernel.set_agent_skills` path rather than inventing a new registry method, so manifest persistence, skill-name validation (the promoted skill is registered by the preceding `reload_skills()`), rollback, and cache invalidation all stay consistent with the `PUT /api/agents/{id}/skills` route.

## Tests

Added three `#[tokio::test]` cases to `crates/librefang-api/tests/skill_workshop_pending_routes_test.rs` (same `TestAppState` + `tower::oneshot` harness as the existing approve tests):

- `pending_approve_assigns_skill_to_creating_agent` — register a real agent, approve a candidate carrying that `agent_id`, assert the promoted skill is appended to the agent's allowlist and `skills_disabled` is cleared.
- `pending_approve_assignment_is_idempotent` — re-approve the same skill (phantom-pending path) and assert the skill appears exactly once on the allowlist.
- `pending_approve_succeeds_when_creating_agent_gone` — approve a candidate whose `agent_id` was never registered; assert the approve still returns 200 and the skill is still promoted.

## Scope

This is the targeted slice of the broader #5844 (approval + assignment). It does NOT implement #5844's diff-review flow for skill *updates* — only the assign-on-approve behaviour #5989 asks for.

## Verification

No local compile/test: this environment has no native `cargo`/`rustfmt` and no usable Docker, so I could not build or run the suite locally. CI is the gate — the new integration tests run in the `Test / Ubuntu` lanes and the fmt/clippy checks run in CI.

### DIFF
--- a/crates/librefang-api/src/routes/skills/skill.rs
+++ b/crates/librefang-api/src/routes/skills/skill.rs
@@ -354,6 +354,15 @@ pub async fn approve_pending_candidate(
     axum::extract::Path(id): axum::extract::Path<String>,
 ) -> impl IntoResponse {
     let skills_root = state.kernel.home_dir().join("skills");
+    // Capture the creating agent id BEFORE promotion: `approve_candidate`
+    // deletes the pending TOML on success, after which `agent_id` is no
+    // longer recoverable. Best-effort — a read failure here must not block
+    // the approve (auto-assign is a convenience, not a precondition).
+    let creator_agent_id =
+        match librefang_kernel::skill_workshop::storage::load_candidate(&skills_root, &id) {
+            Ok(c) => Some(c.agent_id),
+            Err(_) => None,
+        };
     match librefang_kernel::skill_workshop::storage::approve_candidate(
         &skills_root,
         &skills_root,
@@ -364,6 +373,12 @@ pub async fn approve_pending_candidate(
             // `skills_root`; refresh the in-memory registry so the next
             // turn's prompt build sees the new skill.
             state.kernel.reload_skills();
+            // Auto-assign the promoted skill to the agent that produced it
+            // so the workshop loop (capture → review → approve → use)
+            // closes without a manual `agent.toml` edit (#5989). Best-effort:
+            // logs and continues if the agent was deleted between capture
+            // and approval, so the approve still returns 200.
+            assign_skill_to_creator(&state, creator_agent_id.as_deref(), &result.skill_name);
             (
                 StatusCode::OK,
                 Json(serde_json::json!({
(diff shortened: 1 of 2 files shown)


houko (commented):
Two items worth addressing before merge:

**1. Undocumented ordering constraint: `assign_skill_to_creator` must follow `reload_skills()`**

`set_agent_skills` inside `assign_skill_to_creator` validates that the skill name exists in the in-memory registry.
A newly-promoted skill is not in the registry until `reload_skills()` runs.
If a future refactor swaps the two calls, every fresh approval would silently skip the auto-assignment with a `WARN` log that looks like an operator misconfiguration rather than a code-ordering mistake.

Suggested comment at the call site:
```rust
// reload_skills() MUST precede assign_skill_to_creator: set_agent_skills
// validates that the skill name is already in the registry.
state.kernel.reload_skills();
assign_skill_to_creator(&state, creator_agent_id.as_deref(), &result.skill_name);
```

**2. Phantom-recovery path ignores `agent_id` from the second `load_candidate`**

In the `AlreadyInstalled` branch, the code calls `load_candidate` a second time (to compare bodies) and stores the result in `candidate`.
`assign_skill_to_creator` is then called with `creator_agent_id.as_deref()` — the value captured *before* `approve_candidate`.
If the pre-call capture returned `None` (transient I/O failure) but the second `load_candidate` succeeded and its `candidate.agent_id` is `Some`, the auto-assignment is silently skipped with a spurious `WARN`.

The window is narrow (two reads of the same path with no mutation between them), but the fix is a one-liner:
```rust
let effective_agent_id = creator_agent_id
    .as_deref()
    .or_else(|| candidate.as_ref().map(|c| c.agent_id.as_str()));
assign_skill_to_creator(&state, effective_agent_id, &skill_name);
```


---

PR #6023, opened 2026-06-06
fix(channels): resolve channel_send mirror owner via bindings, not just default_agent

### LINKED ISSUE #4824: feat(channels): mirror outbound channel_send into the inbound-routing session for cross-agent context
## Summary

When an agent calls the native `channel_send` tool, the outbound message currently has no path back into the session that will receive the user's reply. If Agent A sends a message into a channel and the user replies, the reply lands in the session of whichever agent owns the inbound channel binding (Agent B), not in Agent A's session. Agent B has no record of what Agent A pushed, so it cannot respond with context.

## Current behavior

- `tool_channel_send` (`crates/librefang-runtime/src/tool_runner.rs:4745`) executes the send and returns a `ToolResult` containing only an acknowledgement string (e.g. `"Message sent to {recipient} via {channel}"`). The message body itself is not persisted anywhere in any agent's session history.
- Inbound channel messages are routed by `SessionId::for_channel(agent_id, "channel:chat_id")` (`crates/librefang-kernel/src/kernel/messaging.rs:1764`), where `agent_id` is the agent bound to the channel — not the agent that called `channel_send`.
- Per-adapter self-echo filters (Messenger, Feishu, etc.) drop reflections of the bot's own send by matching the bot's platform user ID, so the outbound never re-enters as an inbound message.

Net effect: in any deployment with more than one agent (e.g. one cron/automation agent plus one chat-handling agent on the same channel), outbound sends from non-channel-owner agents become invisible to the agent that will eventually process the user's reply.

## Proposed change

When `channel_send` is invoked, in addition to recording the tool result on the calling agent, write a synthetic turn into the inbound-routing session — i.e. `SessionId::for_channel(channel_owner_agent_id, "channel:chat_id")` — containing the actual outbound message body and an attribution to the originating agent.

Sketch:

1. In `tool_channel_send`, after a successful send, resolve the channel-owning agent for `(channel, chat_id)` using the existing inbound routing logic.
2. If that agent differs from the caller (or always, for consistency), append a message to the channel-owning agent's session for that chat. A `system` role with structured content seems safest — e.g. `{ from_agent_id, channel, chat_id, body, sent_at }` — to avoid colliding with strict alternating-role drivers.
3. No new response is triggered on the receiving session; the mirrored turn is context only. The channel-owner agent reacts on the next genuine inbound user message, which now has full context.

## Why this approach

- Preserves the current "one session per (channel-owner, channel, chat_id)" invariant — no addressing scheme changes for callers.
- Surgical: localized to `tool_channel_send` and a small kernel helper to resolve the channel-binding owner.
- Backwards-compatible: callers that don't care still get the existing tool result on their own session; channel-owner agents simply gain context they didn't have before.
- Composable with future enhancements (e.g. an optional `on_behalf_of_session` parameter for explicit routing, or correlation tokens for platform-native reply threading) without conflicting with them.

## Alternatives considered

- **Caller specifies target session** (`on_behalf_of_session: Uuid`): more explicit but requires every caller — including cron-scheduled agents with no session reference — to know the routing key.
- **Per-(channel, chat_id) shared transcript**: cleanest model conceptually (one timeline, multiple agent speakers), but a much larger refactor of session loading and history trimming.
- **Correlation tokens + platform reply primitives**: useful when a chat hosts multiple parallel threads, but only meaningful on platforms with reply-to support (Telegram, Slack threads, Discord); does not help on flat channels.

The proposal above is intended as the smallest change that closes the gap; the alternatives can be layered on later if needed.

## Open questions

1. Should the mirrored turn be written unconditionally, or only when `caller_agent_id != channel_owner_agent_id`? Unconditional is simpler and lets the calling agent see its own outbound as context too if it ever processes inbound on the same channel.
2. Role choice: `system`, `assistant`, or a new content-block kind? `system` avoids alternating-role driver issues but may be surfaced oddly in dashboards.
3. Should the mirror also fire when `channel_send` fails partway (delivered to platform but ack lost)? Probably yes — the user may still see and reply to the message.
4. Does the existing per-adapter self-echo filter need to be aware of mirrored turns to avoid double-counting if a platform later does reflect the outbound back as inbound?

### LINKED ISSUE #6022: bug(channels): channel_send mirror ignores chat_id binding, lands in default_agent session not bound agent (regresses #4824)
## Summary

`channel_send`'s outbound mirror (added in #4824 / PR #4932) resolves the channel-owning agent with `resolve_channel_owner(channel, _chat_id)`, which **ignores `chat_id` entirely** and resolves purely by the channel's `default_agent`.
When a per-conversation binding routes a specific `(channel, chat_id)` to a non-default agent, the mirror is written into the **default agent's** session, not the bound agent's.
The bound agent — the one that will actually receive and process the user's reply — never gains the outbound context, which is the exact gap #4824 set out to close.

## Affected code

`crates/librefang-kernel/src/kernel/handles/channel_sender.rs` (current `main`):

```rust
fn resolve_channel_owner(&self, channel: &str, _chat_id: &str) -> Option<AgentId> {
    let cfg = self.config.load_full();
    let agent_name = sidecar_default_agent(&cfg.sidecar_channels, channel)?; // chat_id ignored
    self.agents.registry.find_by_name(agent_name).map(|e| e.id)
}
```

`crates/librefang-runtime/src/tool_runner/channel.rs::mirror_channel_send_to_session`:

```rust
let owner_id = kh.resolve_channel_owner(channel, recipient);
// ...
let session_id = SessionId::for_sender_scope(owner, channel, Some(recipient));
```

The `_chat_id` parameter is threaded all the way down but discarded at resolution time. `[[bindings]]` `peer_id` routes (highest specificity on the inbound path) are never consulted.

## Why this is a regression of #4824

The inbound router and the mirror disagree on which agent owns a bound conversation:

- **Inbound** (`peer_id` binding): a message in conversation `C` on channel `K` is routed to bound agent **B**, landing on `SessionId::for_sender_scope(B, K, Some(C))`.
- **Outbound mirror**: `channel_send` into the same `(K, C)` resolves the owner as the channel `default_agent` **D** (≠ B) and writes the mirror to `SessionId::for_sender_scope(D, K, Some(C))`.

So the outbound context lands on D while the user's reply lands on B. B answers the reply with no record of what was sent — the precise failure mode #4824 describes, still present for any binding-routed conversation.

## Reproduction (channel-agnostic)

Any sidecar channel `K` works; nothing Matrix-specific.

1. Configure channel `K` with `default_agent = "D"`.
2. Add a per-conversation binding routing conversation `C` on `K` to a different agent `B`:
   ```toml
   [[bindings]]
   agent = "B"
   match_rule = { channel = "K", peer_id = "C" }
   ```
3. Give some agent `S` the `channel_send` tool and have it call:
   `channel_send(channel="K", recipient="C", message="probe")`.
4. Inspect sessions for conversation `(K, C)`:
   - `SessionId::for_sender_scope(D, K, Some(C))` — **gets the mirror turn**.
   - `SessionId::for_sender_scope(B, K, Some(C))` — does **not**, despite B being the agent the binding routes inbound replies to.

Observed: the mirror lands in D's session; B's session is untouched. Expected: the mirror lands in B's session (the agent that owns the bound conversation and will process the reply).

## Suggested fix

`resolve_channel_owner` should mirror the inbound router's precedence: resolve a `(channel, chat_id)` binding **first**, fall back to `sidecar_default_agent(channel)` only when no binding matches. That keeps the mirror's destination consistent with where inbound traffic for the same conversation is routed.

Since the binding resolver already exists on the inbound path, the change is localized to `resolve_channel_owner` (consult bindings before the `default_agent` scan) plus a regression test asserting a bound `(channel, chat_id)` resolves to the bound agent, not the channel default.

## Environment

- Version `2026.5.31-beta.16`; confirmed code unchanged on `main` (`648356c92`).

neo-wanderer:
Closes #6022

## Summary

`channel_send`'s outbound mirror (PR #4932, closes #4824) resolves the channel-owning agent via `resolve_channel_owner(channel, _chat_id)`, which ignored `chat_id` and resolved purely by the channel's `default_agent` (`sidecar_default_agent`).
When a `[[bindings]]` `peer_id` rule routes a specific `(channel, chat_id)` conversation to a non-default agent, the inbound reply is delivered to the bound agent B but the outbound mirror is written to the default agent D's session.
The bound agent never sees its own outbound context — the exact gap #4824 set out to close.

The fix makes owner resolution binding-first: the kernel now iterates its (specificity-sorted) bindings and resolves the first whose `match_rule` matches `(channel, chat_id)`, falling back to the channel `default_agent` only when no binding matches.
To kill the duplicate-matcher drift that caused this bug class, the binding match logic is hoisted into a pure `BindingMatchRule::matches` method in `librefang-types` and is now the single source of truth shared by both the inbound channel `MessageRouter` and the outbound mirror's owner resolution.

## Changes

- **`crates/librefang-types/src/config/types.rs`**: add `BindingMatchRule::matches(channel, account_id, peer_id, guild_id, roles)`, a pure matcher reproducing the exact semantics of the former private matcher in the channels router (all specified fields ANDed; unset fields are wildcards; `roles` matches when the rule's role list is empty or intersects the provided roles).
- **`crates/librefang-channels/src/router.rs`**: refactor the private `binding_matches` to delegate to `BindingMatchRule::matches`; signature and call sites (`resolve_binding`) unchanged, zero behaviour change.
- **`crates/librefang-kernel/src/kernel/handles/channel_sender.rs`**: rewrite `resolve_channel_owner` to be binding-first — match the kernel bindings via `BindingMatchRule::matches(channel, None, chat_id, None, &[])` (the inbound-for-outbound semantics already used by `bound_recipients_for_agent`), resolve the bound agent name → id, and fall back to the existing `sidecar_default_agent` path when no binding matches; rename `_chat_id` → `chat_id`; the binding lock is held only briefly and dropped before the fallback.
- **`crates/librefang-kernel/src/kernel/handles/channel_sender.rs`** (logging): add `tracing::debug!` routing-decision logs at each branch (binding match, binding-matched-but-agent-unregistered, default_agent fallback, unresolved) with structured `channel` / `chat_id` / agent-identifier fields only — no message bodies or PII — so an operator can see why the mirror landed where it did on the per-send hot path.
- **Tests**:
  - `librefang-types`: seven unit tests for `BindingMatchRule::matches` covering channel-only wildcard, all-unset, peer_id match/mismatch, account_id constraint, guild_id constraint, roles intersection, and the AND-of-all-specified-fields case.
  - `librefang-kernel`: `resolve_channel_owner_prefers_binding_then_default_agent` boots a real kernel, registers two agents, adds a `peer_id` binding, and asserts the mirror owner resolves to the bound agent for a matching `(channel, peer_id)`, falls back to the channel `default_agent` for an unbound chat, and resolves to `None` when neither a binding nor a `default_agent` exists.

## Verification

- `cargo check --workspace --lib` — passes.
- `cargo clippy -p librefang-types -p librefang-channels -p librefang-kernel --all-targets -- -D warnings` — clean, zero warnings.
- `cargo test -p librefang-types` — passes; new `binding_matches_*` tests green (7/7).
- `cargo test -p librefang-channels` — passes; all `router::tests::test_binding_*` behaviour-preservation tests green (484 total).
- `cargo test -p librefang-kernel` — passes; `resolve_channel_owner_prefers_binding_then_default_agent` green.

## Out-of-scope follow-up

The kernel-side fix replicates only the **bindings + default_agent** tiers of owner resolution.
The inbound `MessageRouter` also has `direct_route` and `user_default` tiers that the kernel's `ChannelSender` impl cannot currently reach, so full 4-tier parity for the mirror owner would require wiring a `MessageRouter` reference into the mirror path — a larger refactor deferred to a separate tracking issue, to be filed and linked after this PR lands.


houko:
The direction here is right and the extraction of `binding_matches` into a single `BindingMatchRule::matches` (eliminating the dual matcher) is clean and behavior-equivalent — I verified that field-by-field and the inbound path is unchanged. But `resolve_channel_owner` relies on a premise that doesn't hold on boot-from-config, and it can **re-open the cross-agent leak this PR is meant to close** (#6022 / #4824).

**The kernel mesh binding store is not specificity-sorted on boot.**
`resolve_channel_owner` takes the **first** match (`bindings.iter().find(...)`), commenting that the Vec is kept specificity-sorted. But:
- `self.mesh.bindings` is a `Mutex<Vec<AgentBinding>>` stored as-is by `MeshSubsystem::new` (`subsystems/mesh.rs:73`), populated from `config.bindings` in **config file order** (`boot.rs:1263`, `1523`).
- Only the runtime `add_binding` path sorts (`bindings_and_handle.rs:196`). A plain config-boot store is unsorted.
- The inbound router uses a **separate** store that *is* sorted (`librefang-channels/src/router.rs:152` `load_bindings`), so inbound is most-specific-first but the outbound mirror reads the unsorted kernel store.

So when two bindings both match the same `(channel, chat_id)` — e.g. a broad `channel`-only binding (agent D) plus a specific `peer_id` binding (agent B) — and the broad one is written first in config, the mirror resolves to **D** while the inbound reply still routes to **B**. Outbound context lands on the wrong agent: the same failure mode as #6022, with the trigger changed from "default_agent" to "binding declaration order".

The tests don't catch this because they inject via `kernel.add_binding(...)` (which sorts) and use a single binding — they never exercise the unsorted config-boot path or a multi-match contest.

**Suggested fix (option A, minimal):** don't depend on store order — select the highest-specificity match explicitly:

```rust
bindings
    .iter()
    .filter(|b| b.match_rule.matches(channel, None, chat_id, None, &[]))
    .max_by_key(|b| b.match_rule.specificity())
    .map(|b| b.agent.clone())
```

(Alternatively, option B: sort the Vec inside `MeshSubsystem::new` so the kernel store shares the inbound router's invariant — broader blast radius, so please add a store-ordering assertion test if you go that way.)

**Add two regression tests** for the currently-uncovered paths:
1. **Boot-from-config**: start from `KernelConfig { bindings: vec![...], .. }` directly (not `add_binding`) and assert the mirror owner is correct — this is the unsorted path.
2. **Multi-match contest**: put two matching bindings on the same `(channel, chat_id)` — a broad `channel`-only one and a specific `peer_id` one — with the **broad one first** in config, and assert the owner resolves to the specific `peer_id` agent. With the current `find`-based code this test fails; that's the regression it should pin.

The fallback-to-`default_agent` path, the `BindingMatchRule::matches` extraction, and the out-of-scope follow-up you flagged (direct_route / user_default layers) all look fine — just the binding-layer ordering needs to not rely on the unsorted store.

neo-wanderer:
Good catch — the premise is real and verified: `MeshSubsystem::new` (`subsystems/mesh.rs:73`) stores `config.bindings` as-is in file order, only `add_binding` (`bindings_and_handle.rs:197`) sorts, and the inbound router keeps a separately-sorted copy (`router.rs::load_bindings`). So `resolve_channel_owner`'s `find`-first read of the unsorted kernel store could mirror outbound context to a broad binding declared earlier in config while the inbound reply routed to a more-specific one — the same split as #6022, retriggered by declaration order. The comment claiming the Vec is "kept specificity-sorted on load" was wrong for the config-boot path.

Fixed in b1273082f (option A — select max specificity explicitly, no dependence on store order):

```rust
let mut best: Option<&AgentBinding> = None;
for b in bindings.iter().filter(|b| b.match_rule.matches(channel, None, chat_id, None, &[])) {
    if best.is_none_or(|cur| b.match_rule.specificity() > cur.match_rule.specificity()) {
        best = Some(b);
    }
}
best.map(|b| b.agent.clone())
```

One refinement over the `max_by_key` snippet: `max_by_key` returns the *last* element on a specificity tie, whereas the inbound router does a stable descending sort and takes the *first* match. To keep the two paths from drifting on equal-specificity bindings, the fold replaces only on strictly-greater specificity, so the first equal-specificity match wins — matching inbound exactly.

Added both regression tests you asked for (`crates/librefang-kernel/src/kernel/tests.rs`):
1. `resolve_channel_owner_uses_unsorted_config_boot_bindings` — boots from `KernelConfig { bindings: vec![..] }` directly (not `add_binding`), exercising the unsorted store.
2. `resolve_channel_owner_prefers_specific_binding_over_broad_in_config_order` — broad `channel`-only binding declared **first**, specific `peer_id` binding second; asserts the specific agent wins. Fails against the old `find`-based code.

`cargo check -p librefang-kernel --lib`, `clippy -p librefang-kernel --lib -- -D warnings`, and the three `resolve_channel_owner_*` tests all pass locally.

Issue #6017, opened 2026-06-05
bug(cli): cron enable/disable POST a PUT-only route — silent no-op, prints success


neo-wanderer:
## Summary

`librefang cron enable <ID>` / `cron disable <ID>` print a success message
(`✔ Cron job <ID> enabled.`) but never change the job's state. The CLI sends an
HTTP request the server does not route, receives a non-JSON error response,
fails to detect it, and reports success anyway. The `enabled` flag is unchanged
in both `cron list` and `GET /api/cron/jobs`.

## Environment

- librefang `main` @ `fd46c5f9e` (2026-06-03)
- Daemon running in a Docker container; CLI invoked via
  `librefang cron enable <id>` against that same daemon.

## Repro

```text
$ librefang cron list
| 6d3adb85-... | <agent> | */30 * * * * | no  | Poll GitHub activity. |

$ librefang cron enable 6d3adb85-...
  ✔ Cron job 6d3adb85-... enabled.

$ librefang cron list
| 6d3adb85-... | <agent> | */30 * * * * | no  | Poll GitHub activity.   # still "no"
# GET /api/cron/jobs also reports "enabled": false
```

Enabling the same job via the documented API endpoint works and persists:

```text
$ curl -X PUT -H "X-API-Key: $KEY" .../api/cron/jobs/6d3adb85-.../enable -d '{"enabled":true}'
{"id":"6d3adb85-...","enabled":true}
# cron list now shows "yes" and survives restart
```

## Root cause — verb + shape mismatch

The server registers a single **PUT** route that reads `enabled` from the
request **body**:

- `crates/librefang-api/src/routes/workflows/mod.rs:166`
  ```rust
  .route("/cron/jobs/{id}/enable", axum::routing::put(toggle_cron_job))
  ```
- `crates/librefang-api/src/routes/workflows/cron.rs:163` — `toggle_cron_job`
  reads `body["enabled"].as_bool().unwrap_or(true)`. There is **no** `/disable`
  route; enable and disable both go through this one PUT, distinguished by the
  body bool.

The CLI instead sends a **POST** with **no body**, signalling intent via a path
suffix the server ignores:

- `crates/librefang-cli/src/commands/automation.rs:485` (`cmd_cron_toggle`)
  ```rust
  let endpoint = if enable { "enable" } else { "disable" };
  client.post(format!("{base}/api/cron/jobs/{id}/{endpoint}")).send()
  ```

[...]

Issue #123, opened 2026-03-13
[Feature] Add Feishu WebSocket Long Connection Support


houko:
## Description

Add Feishu WebSocket long connection support for real-time messaging.

## OpenFang Reference

- Original Issue: https://github.com/RightNow-AI/openfang/issues/297

PR #6024, opened 2026-06-07
fix(cli): daemon_json surfaces error-less 4xx instead of silent success (#6019)

### LINKED ISSUE #6019: bug(cli): daemon_json swallows 4xx responses — only 5xx is surfaced
## Summary

The shared CLI helper `daemon_json` (`crates/librefang-cli/src/commands/common.rs`) only treats `status.is_server_error()` (5xx) as a failure.
Any **4xx** response (e.g. a 405 from an HTTP verb mismatch, a 404, or a 400) passes straight through.
If that 4xx body is not JSON, or is JSON but lacks an `error` key, `r.json::<serde_json::Value>().unwrap_or_default()` yields `Null`, and the calling command — which typically checks `body.get("error").is_some()` — reads no error and reports **success**.

## Where

`crates/librefang-cli/src/commands/common.rs`, in `daemon_json`:

```rust
let status = r.status();
let body = r.json::<serde_json::Value>().unwrap_or_default();
if status.is_server_error() {   // <-- 5xx only; 4xx slips through
    ui::error_with_fix(/* ... */);
}
body
```

## Impact

Any command whose request hits a 4xx with a non-JSON or `error`-less body silently reports success.
This is exactly what masked #6017: `librefang cron enable/disable` POSTed to a PUT-only route, got a 405 with a non-JSON body, and `daemon_json` waved it through so the CLI printed ✔ while the job was never toggled.
#6017 fixes the cron verb mismatch; this issue tracks the helper-level gap that hid it.

## Safe-fix consideration

The fix is **not** a blanket hard-error on all 4xx.
Some commands intentionally parse a 4xx JSON error body that carries a useful `error` key and surface their own command-specific message (e.g. validation failures).
A safe widening should only treat a 4xx as a fatal daemon error when the body has **no** usable `error` key — when the body *does* carry `error`, return it so the caller's existing handler runs unchanged.
That preserves the structured-error UX while closing the silent-success hole for non-JSON / `error`-less 4xx bodies.

Discovered via #6017.

houko:
## Root cause

`daemon_json` (`crates/librefang-cli/src/commands/common.rs`) only treated
`status.is_server_error()` (5xx) as a failure. Any **4xx** (405/404/400)
slipped straight through. When the 4xx body is non-JSON or JSON without an
`error` key, `r.json::<serde_json::Value>().unwrap_or_default()` yields
`Null`; the caller — which checks `body.get("error").is_some()` — sees no
error and prints **success**.

This is the helper-level gap that masked #6017: `librefang cron enable/disable`
POSTed to a PUT-only route, got a **405** with a non-JSON body, and
`daemon_json` waved it through so the CLI printed `✔` while the job was never
toggled.

## The fix

The widening is deliberately **not** a blanket hard-error on all 4xx (per the
issue's "Safe-fix consideration"). Some commands intentionally parse a 4xx
JSON error body that carries a usable `error` key and surface their own
command-specific message (validation failures). Behaviour by status:

- **5xx** → surfaced as a daemon error (unchanged).
- **4xx WITH a usable `error`** (non-empty `error` string in a JSON object)
  → body returned untouched, caller's existing `body.get("error")` handler
  runs (unchanged).
- **4xx WITHOUT a usable error** (non-JSON `Null` body, JSON object lacking
  `error`, or empty / non-string `error`) → surfaced via the same
  `ui::error_with_fix(error-daemon-returned, …)` the 5xx branch already used.
  No new UI API or i18n key invented.
- **2xx/3xx** → never surfaced here (unchanged).

The classification is extracted into two pure helpers —
`body_carries_usable_error` and `should_surface_status_error` — so the
4xx/5xx logic is unit-testable without binding sockets. `daemon_json`'s
return type and its success / error-carrying behaviour are preserved.

`crates/librefang-cli/src/commands/common.rs`:
- new `body_carries_usable_error` / `should_surface_status_error` helpers
- `daemon_json` condition changed from `status.is_server_error()` to
  `should_surface_status_error(status, &body)`

## Test

New `#[cfg(test)] mod tests` in `common.rs`:
- `usable_error_requires_non_empty_error_string` — `error` must be a
  non-empty string; empty/whitespace, non-string, missing-key, and Null
  bodies are all "not usable".
- `client_error_without_usable_body_is_surfaced` — 405 + Null body (the
  #6017 shape), 404 with an `error`-less object, 400 with an empty `error`
  are all surfaced as failures (no more silent success).
- `client_error_with_usable_body_is_left_for_caller` — 400/409 carrying a
  real `error` string are returned for the caller to handle (UX preserved).
- `server_error_is_always_surfaced` / `success_status_is_never_surfaced` —
  pin the 5xx and 2xx ends so the widening can't drift.

## Verification

No local compile/test: this environment has no native `cargo`/`rustc` and no
reliable Docker, so verification is delegated to CI (`cargo check`, clippy,
and the `Test / Unit (lib+bin)` lane that runs these crate-scoped unit
tests). The change is additive (two pure helpers + one condition swap) and
reuses the existing 5xx error surface and i18n keys, so the blast radius is
the `daemon_json` failure-classification path only.

Closes #6019. Closes the helper-level gap that masked #6017 (#6018 fixes the
cron verb mismatch itself).


houko (commented):
Two comment-style issues against CLAUDE.md conventions. The logic and tests are correct — these are purely comment/doc hygiene items.

**1. Multi-paragraph doc comments on the two new helpers (`common.rs` lines 198–220)**

CLAUDE.md: "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."

Both `body_carries_usable_error` and `should_surface_status_error` carry 7-8 line multi-paragraph `///` blocks.
Suggested replacements (one line each):

```rust
/// Returns true if `body` is a JSON object with a non-empty string `error` field.
pub(crate) fn body_carries_usable_error(body: &serde_json::Value) -> bool {
```

```rust
/// Returns true when `daemon_json` should surface `status` as a fatal error rather than returning the body.
pub(crate) fn should_surface_status_error(
```

**2. Inline test comments that reference issue numbers or just describe WHAT (`common.rs` lines 857–907)**

CLAUDE.md: "Don't explain WHAT the code does, since well-named identifiers already do that. Don't reference the current task, fix, or callers … issue #123 … since those belong in the PR description and rot as the codebase evolves."

Comments to remove:
- `// Empty / whitespace-only error strings are not usable messages.` (line 857) — WHAT
- `// Non-string \`error\` (object/array/number/null) is not a usable message.` (line 860) — WHAT
- `// Object without an \`error\` key, and non-object bodies.` (line 863) — WHAT
- `// 5xx is a daemon error regardless of body shape (historical behaviour).` (line 872) — WHAT; test name says it already
- `// #6019 / #6017: a 405 (PUT-only route POSTed) with a non-JSON body / deserializes to Null; previously this slipped through as success.` (lines 886–887) — issue-number reference + WHAT
- `// 404/400 with an \`error\`-less JSON object are equally surfaced.` (line 893) — WHAT
- `// A 4xx that carries a real \`error\` key is returned for the caller's / own command-specific handler — not hard-errored here.` (lines 906–907) — WHAT

Comment worth keeping: `// Null is what \`unwrap_or_default()\` yields for a non-JSON body.` — explains a non-obvious runtime invariant (WHY this specific test value, not WHAT the assertion does).

---

PR #6039, opened 2026-06-09
fix(api): accept `version` field in ClawHubInstallRequest (#6038)

### LINKED ISSUE #6038: ClawHub skill install always fails with 422: dashboard sends `version` but ClawHubInstallRequest denies unknown fields
## Summary

Every ClawHub skill install from the dashboard fails with **HTTP 422 Unprocessable Entity**. The install never reaches the handler — the request body is rejected at the `Json<ClawHubInstallRequest>` extractor.

## Reproduction

Click "Install" on any ClawHub skill in the dashboard (Skills page). The request fails silently. Reproduced directly against the API:

```
$ curl -s -o /dev/null -w "%{http_code}\n" -X POST http://127.0.0.1:4545/api/clawhub/install \
    -H "Authorization: Bearer <token>" -H "Content-Type: application/json" \
    -d '{"slug":"github-helper","version":"latest","hand":null}'
422
```

Response body:

```
Failed to deserialize the JSON body into the target type: version: unknown field `version`, expected `slug` or `hand`
```

It fails even with a valid `slug`, so it is not a missing-field problem — it is the extra `version` field being rejected.

## Root cause

The dashboard always sends a `version` field (`crates/librefang-api/dashboard/src/api.ts`, `clawhubInstall`):

```ts
post("/api/clawhub/install", { slug, version: version || "latest", hand }, ...)
```

But the backend request type denies unknown fields and has no `version` field (`crates/librefang-api/src/types.rs`):

```rust
#[derive(Debug, Deserialize, utoipa::ToSchema)]
#[serde(deny_unknown_fields)]
pub struct ClawHubInstallRequest {
    pub slug: String,
    #[serde(default)]
    pub hand: Option<String>,
}
```

`deny_unknown_fields` + the dashboard's `version` => the axum `Json` extractor returns 422 before the handler runs. This affects **all** ClawHub installs via the dashboard, on current `main`.

## Impact

ClawHub skill installation from the dashboard is completely broken. The China-mirror handler (`clawhub_cn_install`) uses the same request type and is affected identically.

## Proposed fix

Accept the field the dashboard already sends: add `version: Option<String>` with `#[serde(default)]` to `ClawHubInstallRequest`. The installer currently resolves to the latest published version (`client.install(slug, dir)` takes no version), so the field is accepted to keep the request well-formed rather than driving version selection yet. PR incoming.

DaBlitzStein:
## Summary

Closes #6038.

Installing any skill from the dashboard's ClawHub / SkillHub browser fails with `422 Unprocessable Entity`.
The dashboard's `clawhubInstall` (`dashboard/src/api.ts`) posts `{ slug, version, hand }`, but `ClawHubInstallRequest` was declared `#[serde(deny_unknown_fields)]` with only `slug` and `hand` — no `version` — so the axum `Json` extractor rejected every request before the handler ran ("unknown field `version`").

The one shared `ClawHubInstallRequest` type backs **three** install handlers, so the single missing field broke all of them:
- `POST /api/clawhub/install` (`routes/skills/clawhub.rs::clawhub_install`)
- `POST /api/clawhub-cn/install` (`routes/skills/clawhub.rs::clawhub_cn_install` — China mirror)
- `POST /api/skillhub/install` (`routes/skills/skillhub.rs::skillhub_install`)

## Change

- Add `version: Option<String>` with `#[serde(default)]` to `ClawHubInstallRequest` (`crates/librefang-api/src/types.rs`).
  The installer still resolves to the latest published version — the field is accepted to keep the request well-formed under `deny_unknown_fields`, not (yet) to drive version selection. `slug`-only bodies keep working.
- Regenerated `openapi.json` (the type is a `utoipa::ToSchema`); the pre-commit hook auto-synced `xtask/baselines/openapi.sha256`.

## Verification

- `cargo check -p librefang-api` — clean.
- `cargo test -p librefang-api --lib clawhub_install_request` — 2/2 pass:
  - `clawhub_install_request_accepts_dashboard_body_with_version` — the exact dashboard body `{"slug","version","hand"}` now deserializes.
  - `clawhub_install_request_slug_only_still_works` — backward-compatible slug-only body.
- `cargo xtask codegen --openapi` — regenerated spec; `ClawHubInstallRequest.properties` now lists `slug`, `version`, `hand`.

## Out-of-scope follow-ups

None. The handler ignoring `version` (always installing latest) is intentional and documented on the field; wiring real version selection is a separate feature, not a regression.


houko:
Re-reviewed after the latest push — this is a correct, well-scoped fix for the 422 and the two regression tests (`accepts_dashboard_body_with_version` / `slug_only_still_works`) plus the regenerated `openapi.json` cover both shapes. Good.

One thing worth a tracked follow-up (not blocking this PR): the field is accepted but silently ignored — the installer always resolves to latest, so a client that requests a specific `version` gets latest with no signal. Today the dashboard only ever sends `"latest"`, so there's no user-visible impact, but once a non-`latest` value can be sent this becomes a silent wrong-result. Suggest either wiring `version` through to the resolver in the follow-up, or, until then, a `warn!`/`debug!` when `version` is `Some(v)` and `v != "latest"` so it's observable rather than silently dropped. The doc-comment already flags the intent ("rather than driving version selection yet"), so a one-line issue to honor it later would close the loop.

houko (commented):

**Doc-comment mid-sentence line wrap in `crates/librefang-api/src/types.rs`**

The doc comment for the `version` field wraps mid-sentence at ~80 chars:

```rust
/// Requested version (e.g., "latest"). The dashboard always sends this;
/// the installer currently resolves to the latest published version, so
/// the field is accepted to keep the request well-formed under
/// `deny_unknown_fields` rather than driving version selection yet.
```

CLAUDE.md (Prose wrapping section): *"The only legitimate line break inside a paragraph is at a complete sentence boundary. This rule applies universally: … Source code doc-comments."*

Fix — one sentence per line:

```rust
/// Requested version (e.g., "latest"). The dashboard always sends this; the installer currently resolves to the latest published version, so the field is accepted to keep the request well-formed under `deny_unknown_fields` rather than driving version selection yet.
```

After updating the Rust source, regenerate `openapi.json` (pre-commit hook / `cargo xtask codegen openapi`) to update the description string and `xtask/baselines/openapi.sha256`.

No other issues found; the serde fix and both regression tests look correct.

---

PR #6042, opened 2026-06-09
fix(dashboard/agents): stage Skills-tab edits behind a Save button

### LINKED ISSUE #6041: Dashboard: agent Skills tab persists immediately — no Save button like the Tools tab
## Summary

On an agent's detail panel, the **Skills** tab and the **Tools** tab behave inconsistently:

- **Tools tab** stages changes locally and only persists when you click **Save** (disabled until dirty). Switching tabs discards the edit.
- **Skills tab** fires a `PUT /api/agents/{id}/skills` **immediately** on every add / remove / Customize / Reset-to-all. There is no Save button and no way to back out — a misclick is instantly written to `agent.toml`.

## Impact

No "change your mind" safety net on skill assignment. An accidental remove or a "Reset to all" misclick is persisted with no confirmation/undo, unlike every other staged-edit surface in the dashboard.

## Expected

The Skills tab should mirror the Tools tab: stage add/remove/customize/reset into local draft state, show a **Save** button that's disabled until there are pending changes, and only issue the PUT on Save. Leaving the tab discards the draft.

## Pointers

- `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx` — `renderSkillsTab` (immediate `setAgentSkillsMutation.mutate(...)` in the add/remove handlers) vs `renderToolsTab` (`toolsDraft` + `handleSave`).

DaBlitzStein:
## Summary

Closes #6041.

The agent **Skills** tab persisted every change immediately (`PUT /api/agents/{id}/skills` on each add / remove / Customize / Reset-to-all) with no Save button and no way to back out — inconsistent with the sibling **Tools** tab, which stages a draft and only writes on **Save**. A misclick (or "Reset to all") was instantly written to `agent.toml`.

This makes Skills mirror Tools exactly.

## Change

All in `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx`:

- Added a `skillsDraft: string[] | null` state next to the existing `toolsDraft` (`null` = pristine, otherwise the staged allowlist).
- Added a draft-sync `useEffect` mirroring the Tools one: seeds the draft from the persisted allowlist when an allowlist-mode agent's Skills tab opens; resets to `null` when leaving the tab (so the draft is discarded).
- `renderSkillsTab`: add / remove / Customize / Reset-to-all now only mutate `skillsDraft` — no network call. The view derives `isDirty`/`usesAllSkills` from the draft, so pending edits are visible before saving.
- Added a bottom **Save** button (disabled until `isDirty`) that issues the single `setAgentSkillsMutation` PUT, then clears the draft on success. An empty allowlist still saves as all-mode (unchanged write semantics).
- **Auto-evolve** stays an immediate toggle — it's a separate boolean, not part of the skill allowlist.

No changes to `src/lib/queries`, `src/lib/mutations`, or `src/api.ts`; the mutation hook (`useSetAgentSkills`) and its invalidation are reused as-is.

## Verification

- `pnpm typecheck` — clean.
- `pnpm lint src/pages/AgentsPage.tsx` — 0 errors. The one new `react-hooks/exhaustive-deps` warning on the skills `useEffect` is the same intentional pattern as the existing `toolsDraft` effect (the draft is deliberately excluded to avoid re-seeding); the lint policy allows warnings.
- `pnpm test --run` — skills + key-factory suites green (`agents-skills.test.tsx`, `queries/agents-skills.test.tsx`, `keys.test.ts`: 42 passed). `pnpm build` succeeds.
- Pre-existing, unrelated: `src/pages/ModelsPage.test.tsx` fails in isolation on `main` too (zustand `localStorage.setItem` throwing under jsdom) — untouched by this PR (diff is one file). Flagged separately, not bundled here.

## Notes

Manual browser click-through was not performed in this environment; behavior is asserted via typecheck/lint/build and by structural parity with the already-shipped Tools-tab Save flow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -801,6 +801,10 @@ _308 PRs from 7 contributors since v2026.5.17-beta.12._
   Under load a `GET /run` landing inside that write window lost the `try_lock` and surfaced a live run as not running — a dashboard `/run` poll would flicker "stopped", and the `goal_run_start_then_stop_with_agent` integration test flaked on it.
   The loop now updates the in-memory state under the lock, releases it, and persists the cloned snapshot outside the lock; `state` is written only by the single loop task, so the snapshot stays consistent.
   This shrinks the lock hold to a few synchronous field writes, so `try_lock` no longer realistically contends.
+- **fix(dashboard/agents): stage agent Skills-tab edits behind a Save button instead of persisting on every click** (#6041) (@DaBlitzStein).
+  The Skills tab fired `PUT /api/agents/{id}/skills` immediately on every add / remove / Customize / Reset-to-all, with no Save button and no way to back out — unlike the Tools tab, which stages a draft and only writes on Save.
+  Skills now mirrors Tools: edits stage into a local `skillsDraft`, a Save button (disabled until dirty) issues the single PUT, and switching tabs discards the draft.
+
 - **channels: stop silently swallowing an upgraded operator's channel config, and explain why the WeChat/etc. configure form is empty** (@houko).
   After the channel → sidecar migration (#5317–#5459) the in-process `[channels.<vendor>]` config blocks were removed from `ChannelsConfig`, so an operator upgrading from a pre-migration build lost every configured channel on first boot: the old block deserialised into nothing, the dashboard channels page (which only renders `configured` rows) showed the WeChat card vanishing, and the only signal was a generic "Unknown config field (ignored)" log that never mentioned sidecars.
   Re-adding the channel then failed just as quietly — the configure form is schema-driven off `python3 -m librefang.sidecar.adapters.<name> --describe`, which fails when the Python sidecar SDK is not installed (and a pre-migration WeChat ran in-process Rust, so an upgrader never had it), leaving the Add-picker drawer blank with no inputs and no explanation.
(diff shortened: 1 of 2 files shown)


houko:
Re-reviewed the latest revision — traced the full state machine and it's a faithful, correct mirror of the Tools-tab draft pattern:

- `skillsDraft` seeds from the pinned set only in allowlist mode and stays `null` in all-mode, so the all-view renders until the operator customizes (`usesAllSkills = persisted.length === 0 && !isDirty`).
- `isDirty` correctly compares the sorted draft against `persisted` by length + membership, so Save is disabled on a no-op and the edge cases hold: `customizeFromAll` (draft = all available, persisted `[]`) → dirty; `resetToAll` while in allowlist mode (draft `[]`, persisted non-empty) → dirty; `resetToAll` while already all-mode → not dirty.
- `handleSaveSkills` fires the single PUT with `assigned` and clears the draft on success; switching tabs resets the draft (the discard path).

No issues found. One note for consistency, not blocking: the seed `useEffect` reads `skillsDraft` in its `skillsDraft === null` guard but omits it from the deps `[agentTab, tabAgentSkillsQuery.data]` — that's the same intentional deviation the Tools effect makes (you don't want to re-seed on every draft edit), so it's fine and consistent, just worth a `// eslint-disable-next-line react-hooks/exhaustive-deps` if the lint rule ever turns strict.

Issue #4983, opened 2026-05-12
feat(kernel,runtime): async task tracker — non-blocking workflow/delegation results via kernel event injection


DaBlitzStein:
## Summary

Agents need a way to launch async operations (workflows, agent_send delegations, external calls) and receive results without blocking their conversation loop or requiring a separate orchestrator agent burning tokens on continuous polling.

## Motivation: real-world failure on production Rodela instance

Agent `ltdata` (Telegram assistant) launched a workflow via `workflow_run`. The workflow took several minutes. Result:

1. `workflow_run` blocked ltdata's agent loop for the entire duration
2. The user sent follow-up messages during this time — ltdata was unresponsive
3. The workflow eventually timed out
4. ltdata received no run_id, no partial status, no error context — just a timeout
5. The agent was effectively "bricked" for the duration, unable to serve any user

The alternative `workflow_start` (async) returns a run_id immediately, but ltdata has no mechanism to check back on it. The run_id is returned to the LLM as a tool result, but by the next user message the agent has moved on and forgotten about it.

The current workaround is the `orchestrator` agent pattern: a separate agent on `continuous` mode (every 120s) that polls for pending work. This costs ~$2/day in LLM calls even when idle, and introduces coordination complexity between the orchestrator and the actual agents.

## Problem statement

There are three async patterns in LibreFang, all with the same gap:

| Pattern | Launch | Result delivery | Problem |
|---------|--------|-----------------|---------|
| `workflow_run` | Blocking tool call | Returns when done | Agent frozen, user ignored, timeout kills it |
| `workflow_start` | Returns run_id | Nothing — fire and forget | Agent forgets, user never gets result |
| `agent_send` | Sends to peer agent | Peer response in sender's session | Blocking — sender waits for full response |

None of these allow: "launch → continue serving user → get notified when done → act on result."

### Architecture

```
Agent launches async op → kernel registers (agent_id, task_type, task_id)
  → agent continues normally (responds to user, handles new messages)
  → async op completes → kernel injects result into agent's session as synthetic message
  → agent processes result on next turn (or wakes up if idle)
```

### Kernel task registry

```rust
struct PendingTask {
    agent_id: AgentId,
    session_id: SessionId,
    task_type: TaskType,        // Workflow, AgentSend, Webhook, etc.
    task_id: String,            // run_id, message_id, etc.
    launched_at: Instant,
    timeout: Option<Duration>,
    callback_channel: Option<String>,  // channel to notify user through
}

[...]

Issue #5960, opened 2026-06-01
Feature request: add creator_match / claimed_by_match filters to TaskClaimed and TaskCompleted trigger patterns


nevgenov:
## Summary

`TaskPosted` supports `assignee_match` to narrow trigger fires to tasks assigned to a specific agent. `TaskClaimed` and `TaskCompleted` are bare unit variants with **no filtering at all**, making it impossible to scope notification triggers to tasks posted by a specific orchestrator agent.

## Root cause (verified in current source)

In `crates/librefang-kernel/src/triggers.rs`, the three task patterns are:

```rust
TaskPosted {
    #[serde(default)]
    assignee_match: Option<String>,   // ← supports filtering
},
TaskClaimed,    // ← no fields, fires for every claim system-wide
TaskCompleted,  // ← no fields, fires for every completion system-wide
```

The matching logic for `TaskClaimed`/`TaskCompleted` is:

```rust
TriggerPattern::TaskClaimed => matches!(
    event.payload,
    EventPayload::System(SystemEvent::TaskClaimed { .. })
),
TriggerPattern::TaskCompleted => matches!(
    event.payload,
    EventPayload::System(SystemEvent::TaskCompleted { .. })
),
```

No filtering is applied. Meanwhile `SystemEvent::TaskPosted` already carries a `created_by: Option<String>` field (visible in tests), so the data is available — it just isn't exposed in the pattern.

## Impact

In a multi-agent setup where one orchestrator agent delegates tasks to workers, the orchestrator typically listens to `task_claimed`/`task_completed` to notify the end user. Without a `creator_match` filter, the orchestrator fires on **every** claim/completion in the system — including tasks posted by the UI, by other agents, or by automation scripts — producing spurious notifications.

Concrete scenario: orchestrator agent posts a task → worker claims it → `task_claimed` fires on orchestrator ✅. But if any other source posts a task, the orchestrator's trigger fires too ❌.

## Proposed fix

Add optional filter fields symmetrical to `assignee_match` on `TaskPosted`:

```rust
/// Match when a task is claimed from the Task Board.
TaskClaimed {
    /// Optional: only fire if the task was originally posted by this agent.
    /// Accepts agent UUID, display name, or `"self"`.
    #[serde(default)]
    creator_match: Option<String>,
},

/// Match when a task is completed on the Task Board.
TaskCompleted {
    /// Optional: only fire if the task was originally posted by this agent.
    /// Accepts agent UUID, display name, or `"self"`.
[...]

PR #6044, opened 2026-06-09
feat(runtime,kernel): non-blocking agent_send via the async-task tracker

### LINKED ISSUE #6043: feat(runtime,kernel): non-blocking agent_send — extend the #4983 async-task tracker to inter-agent delegation
## Problem

`agent_send` is **synchronous-only**: `tool_agent_send` (`crates/librefang-runtime/src/tool_runner/agent.rs:94-111`) awaits `kh.send_to_agent_as(...)`, which runs the *target* agent's full loop and returns its final response as the tool result. That await is wrapped by the tool-execution timeout (`tool_timeout_secs`, default 120s), so **any delegation that takes longer than the timeout always fails** — and even within the budget it blocks the caller's loop, so a Telegram-facing agent goes unresponsive while it waits.

### Real failure (production proteo)

`deannatroi` delegated a research task to agent `Viajecinio` via `agent_send`. Viajecinio's nested loop did legitimate multi-step web research (DuckDuckGo + page fetches + several deepseek-v4 turns) that ran past 120s, so deannatroi's `agent_send` was killed:

```
WARN run_agent_loop: Tool execution timed out after 120s  tool=agent_send  agent.name=deannatroi
```

From deannatroi's side the delegation simply "failed", even though Viajecinio was working fine. This is the exact class of failure that #4983 (async task tracker) was created to solve — but #4983 was only wired to **workflows**, never to `agent_send`.

## What already exists (the template)

#4983 / its PR shipped a complete async-completion path that **`workflow_start` already uses**:

- `KernelHandle::start_workflow_async_tracked(workflow_id, input, caller_agent_id, caller_session_id)` (`crates/librefang-kernel-handle/src/workflow_runner.rs:283`) — registers the run on the async-task tracker, spawns it detached, returns a `run_id` immediately.
- On completion the kernel calls `complete_async_task(...)` (`crates/librefang-kernel/src/kernel/task_registry.rs:257`), which injects a `TaskCompletionEvent` back into the originating `(agent, session)` — **mid-turn** (`AgentLoopSignal::TaskCompleted`, #956 channel) if the loop is live, else **wake-idle** (spawns a fresh turn with a `[System] [ASYNC_RESULT]` message), with the trigger-lane + per-agent concurrency semaphores applied.

`agent_send` has no equivalent — there is no async variant and no `register_async_task` wiring.

## Proposal

Extend the **existing** #4983 tracker to inter-agent delegation (no new transport needed — reuse `register_async_task` / `complete_async_task`):

1. Add a non-blocking mode to `agent_send` — an `async: bool` param (default `false`, fully backward-compatible) on the existing tool, OR a sibling `agent_start` tool mirroring the `workflow_run` / `workflow_start` split. (Leaning toward the `async` param to avoid a confusing new tool name next to `agent_spawn`.)
2. Add a `KernelHandle` method analogous to `start_workflow_async_tracked` (e.g. `send_to_agent_async_tracked(agent_id, message, caller_agent_id, caller_session_id, conversation_key)`) that registers a `TaskKind` for the delegation, spawns the callee loop detached, and on completion calls `complete_async_task` with the callee's response.
3. Return a `task_id` immediately so the caller's LLM frees its turn; the reply arrives later via the existing mid-turn / wake-idle injection.
4. Depth guard (`AGENT_CALL_DEPTH`, `agent.rs:85-92`) and taint checks stay; the async spawn must propagate the depth scope.

## Acceptance

- `agent_send` with `async=true` returns a `task_id` without blocking; the target's reply is injected back into the caller's session on completion (mid-turn and wake-idle both covered).
- Blocking `agent_send` (default) is byte-for-byte unchanged.
- A long (> `tool_timeout_secs`) delegation no longer hits the tool timeout in async mode.
- Integration test mirroring `async_task_tracker_runtime_test.rs` but for an `agent_send` delegation.

DaBlitzStein:
## Summary

Closes #6043.

`agent_send` was **synchronous-only**: `tool_agent_send` awaited the target agent's full reply as the tool result, so any delegation longer than `tool_timeout_secs` (default 120s) was killed mid-flight, and the caller's loop stayed blocked the whole time. On production this showed up as a Telegram-facing agent going unresponsive while a research sub-agent ran, then failing with `Tool execution timed out after 120s tool=agent_send`.

This extends the **existing** async-task tracker (#4983 — previously wired only to `workflow_start`) to inter-agent delegation. No new transport: it reuses `register_async_task` / `complete_async_task` and the already-present `TaskKind::Delegation`.

## Change

- **New trait method** `KernelHandle::send_to_agent_async_tracked` (`crates/librefang-kernel-handle/src/agent_control.rs`) with a default impl that falls back to blocking `send_to_agent_as` for handles without a tracker (mocks/tests).
- **Kernel impl** (`crates/librefang-kernel/src/kernel/handles/agent_control.rs`) mirrors `start_workflow_async_tracked`: resolves target + caller, registers a `TaskKind::Delegation { agent_id, prompt_hash }`, spawns the callee loop detached via `self_handle`, and on completion calls `complete_async_task` with the reply — which the tracker injects back into the caller's session (mid-turn `AgentLoopSignal::TaskCompleted`, else wake-idle). Falls back to a blocking send when no caller session is parseable (so the result is never silently dropped).
- **Tool** (`tool_runner/agent.rs`): `agent_send` gains an `"async": boolean` param (default false). When true it routes to the tracked path and returns `{ "task_id", "status": "delegated", "note" }` immediately. The depth guard still runs first; the synchronous `AGENT_CALL_DEPTH` scope is intentionally not carried across the detached spawn. `caller_session_id` is now threaded from `dispatch.rs`.
- **Tool definition** updated to describe the two modes (blocking vs `async`).

Blocking `agent_send` (the default) is byte-for-byte unchanged.

## Verification

- `cargo clippy -p librefang-kernel-handle -p librefang-kernel -p librefang-runtime --all-targets -- -D warnings` — clean.
- `cargo test -p librefang-runtime --lib tool_runner::tests` — 237 passed (incl. 2 new: `agent_send_async_routes_to_tracked_path_and_returns_task_id`, `agent_send_async_without_caller_is_invalid_parameter`).
- `cargo test -p librefang-kernel --test async_task_tracker_test` — 15 passed (existing `TaskKind::Delegation` registration / dedup / completion-injection coverage already exercises the delivery side this PR drives).

## Notes / follow-ups

- This is the imperative async primitive. A complementary **reactive** path (a first-class `agent_message_received` trigger so a recipient owns its reaction, generalising `task_posted`) was discussed as a separate, larger design — not in scope here.
- `prompt_hash` uses `DefaultHasher` (opaque, caller's-choice per the field docs); swap to a stable hash if cross-process dedup of delegations is ever needed.


houko (commented):
**CHANGELOG.md: `kernel(triggers)` entry (#5960) was accidentally clobbered**

The diff replaced the first sentence of the existing `kernel(triggers): TaskClaimed and TaskCompleted triggers gain an optional creator_match filter` bullet (#5960, @nevgenov) with the new `feat(runtime,kernel)` entry, but the *second* sentence of the `kernel(triggers)` entry was left in place and is now dangling as a continuation of the async `agent_send` description.

This line belongs to #5960, not to `agent_send`:

> Accepts an agent UUID, display name, or `"self"`; absent (`#[serde(default)]`) it preserves the legacy fire-for-all behaviour and existing string-form triggers still parse. The original poster is threaded onto the `TaskClaimed` / `TaskCompleted` events from the task record.

The CHANGELOG should read (two separate bullets):

```markdown
- **feat(runtime,kernel): non-blocking `agent_send` — `"async": true` delegates without blocking the caller** (#6043) (@DaBlitzStein).
  `agent_send` was synchronous-only: it awaited the target agent's full reply as the tool result, so any delegation longer than `tool_timeout_secs` (default 120s) was killed mid-flight and the caller's loop stayed blocked the whole time (a Telegram-facing agent went unresponsive while it waited).
  Passing `"async": true` now registers the delegation on the existing async-task tracker (#4983, previously wired only to `workflow_start`) via the new `KernelHandle::send_to_agent_async_tracked`, spawns the callee loop detached, and returns a `task_id` immediately. The callee's response is injected back into the caller's session when it finishes — mid-turn if the loop is live, else wake-idle — reusing `TaskKind::Delegation` and `complete_async_task`.
  Blocking `agent_send` (the default) is unchanged; the depth guard still rejects too-deep chains before they can fire async.
- **kernel(triggers): `TaskClaimed` and `TaskCompleted` triggers gain an optional `creator_match` filter** (#5960) (@nevgenov), symmetrical to `TaskPosted`'s `assignee_match`, so an orchestrator can scope claim/completion notifications to tasks it originally posted instead of firing on every claim/completion system-wide.
  Accepts an agent UUID, display name, or `"self"`; absent (`#[serde(default)]`) it preserves the legacy fire-for-all behaviour and existing string-form triggers still parse. The original poster is threaded onto the `TaskClaimed` / `TaskCompleted` events from the task record.
```

Please restore the `kernel(triggers)` bullet and remove its dangling sentence from the `feat(runtime,kernel)` entry.

---

houko (commented):
**CHANGELOG.md: new entry inserted mid-bullet, corrupting the `#5960` entry**

The `#6043` entry is inserted between the first sentence and the continuation sentence of the `TaskClaimed`/`TaskCompleted` trigger filter entry (#5960, @nevgenov).

In the resulting file the `#5960` entry reads as a single orphaned sentence, and the `Accepts an agent UUID, display name, or \`"self"\`; absent (\`#[serde(default)]\`) it preserves the legacy fire-for-all behaviour and existing string-form triggers still parse. The original poster is threaded onto the \`TaskClaimed\` / \`TaskCompleted\` events from the task record.` continuation — which clearly describes the `creator_match` trigger filter, not `agent_send` — ends up visually attached to the new `#6043` entry.

The fix is to move the `#6043` bullet to a position *after* the full `#5960` entry (i.e., after the `Accepts an agent UUID…` line), so the TaskClaimed entry stays intact:

```diff
 - **kernel(triggers): `TaskClaimed` and `TaskCompleted` triggers …** (#5960) (@nevgenov)…
   Accepts an agent UUID, display name, or `"self"`; absent (`#[serde(default)]`) it preserves the legacy fire-for-all behaviour and existing string-form triggers still parse. The original poster is threaded onto the `TaskClaimed` / `TaskCompleted` events from the task record.
+- **feat(runtime,kernel): non-blocking `agent_send` …** (#6043) (@DaBlitzStein).
+  …
```

---

CHANGELOG.md
houko, line 1: The new `feat(runtime,kernel)` bullet was inserted in the middle of the existing `kernel(triggers): TaskClaimed and TaskCompleted` entry, splitting it across two separate bullets.

In the diff, the hunk looks like this:

```
 - **kernel(triggers): `TaskClaimed` and `TaskCompleted` … system-wide.
+- **feat(runtime,kernel): non-blocking `agent_send` …
+  …
   Accepts an agent UUID, display name, or `"self"`; …
```

The indented `Accepts an agent UUID, display name, or "self"; absent…` sentence is the **second paragraph of the `kernel(triggers)` bullet**, not a continuation of the new `feat(runtime,kernel)` entry. As written it now reads as if that paragraph belongs to the async `agent_send` change, which is incorrect — it describes the `creator_match` filter for TaskClaimed/TaskCompleted.

Please move the new `feat(runtime,kernel)` block to **after** `Accepts an agent UUID, display name, or "self"; …` so the TaskClaimed/TaskCompleted entry is kept intact.

---

houko (commented):
One CHANGELOG formatting issue to fix before merge (see inline comment on `CHANGELOG.md`)."

---

houko (commented):
CI is failing with 9 compile errors — all in `crates/librefang-runtime/src/tool_runner/tests/mod.rs`. Two root causes:

**1. `DispatchCapture::send_to_agent_async_tracked` is missing the `chat_id` parameter (E0050)**

The trait signature in `agent_control.rs` was extended to include `chat_id: Option<&str>` as the last parameter, but the test mock at line 654 only has 5 parameters (excluding `&self`). Fix:

```diff
     async fn send_to_agent_async_tracked(
         &self,
         _agent_id: &str,
         _message: &str,
         _caller_agent_id: &str,
         caller_session_id: Option<&str>,
         _conversation_key: Option<&str>,
+        _chat_id: Option<&str>,
     ) -> Result<String, librefang_kernel_handle::KernelOpError> {
```

**2. Eight test call sites pass 4 arguments to `tool_agent_send` but the function now takes 5 (E0061)**

`tool_agent_send` gained `chat_id: Option<&str>` as a fifth parameter, but the test updates only added `caller_session_id`. Each call needs one more trailing `None`:

```
line 856:  tool_agent_send(&input, Some(&kernel), None, None)           → …, None, None, None)
line 872:  tool_agent_send(&input, Some(&kernel), Some("parent-agent"), None)   → …, None, None)
line 893:  same
line 902:  same
line 929:  same
line 964:  tool_agent_send(&input, Some(&kernel), Some("parent-agent"), Some(session)) → …, Some(session), None)
line 1001: tool_agent_send(&input, Some(&kernel), None, None)           → …, None, None, None)
line 1034: tool_agent_send(&input, Some(&kernel), Some("parent-agent"), None)   → …, None, None)
```

No logic changes required — just the trailing `None` (tests have no channel context, so `None` is correct). `cargo check -p librefang-runtime` will be green after these edits.

---

houko (changes requested):
Requesting changes — the branch does not compile (CI red on Build / Linux aarch64, Test / Unit, Test / Ubuntu shard 1/4, and Quality).

Root cause from the build logs (run 27844448423), and these are internal to this PR — `send_to_agent_async_tracked` is newly introduced here and the PR is only 1 commit behind `main`, so it's not a rebase/drift issue:

- `error[E0050]: method 'send_to_agent_async_tracked' has 6 parameters but the declaration in trait 'send_to_agent_async_tracked' has 7` — the trait declaration and its impl disagree on arity; the signature was changed in one place but not the other.
- Multiple `error[E0061]: this function takes 5 arguments but 4 arguments were supplied` — call sites weren't updated to the new signature.

Please reconcile the trait/impl signature and update every call site so it compiles, then push — happy to re-review once CI is green. The non-blocking-`agent_send` direction itself looks right; this is purely a half-applied signature change.

houko:
Pushed a small test-only fix (f1e1bcaf4) to get the build green.

The production side (trait, real impl, the `tool_agent_send` callsite) was correct, but two test sites in `crates/librefang-runtime/src/tool_runner/tests/mod.rs` hadn't been updated for the new `chat_id` parameter, which is what the red CI was:

- the mock impl of `send_to_agent_async_tracked` was missing the 6th param (E0050), and
- the 8 `tool_agent_send(...)` callsites passed 4 args instead of 5 (E0061).

Added the missing `chat_id` param to the mock and `None` as the chat_id argument at each callsite — no production change. CI should be green now; once it is, this is good to go from my side.

houko (approved):
The build-breaking test gaps are fixed (f1e1bcaf4): the mock impl and the 8 `tool_agent_send` callsites now match the new `chat_id` signature. Approving once CI is green.

PR #6046, opened 2026-06-09
fix(kernel): persist agent skill & MCP-server assignments to agent.toml

### LINKED ISSUE #6045: bug(kernel): per-agent skill & MCP-server assignments don't survive restart (set_agent_skills/set_agent_mcp_servers skip persist_manifest_to_disk)
## Summary

Assigning a skill (or MCP server) to an agent via the dashboard/API persists only to the SQLite blob, **not** to `agent.toml`. Because boot reconciliation treats the on-disk manifest as authoritative, the assignment is silently **wiped on the next daemon restart**.

## Reproduce

1. Install a skill so it's in the registry.
2. Assign it to an agent (dashboard Skills tab → Save, or `PUT /api/agents/{id}/skills`). Response is `200`; `GET /api/agents/{id}/skills` shows it as `assigned`.
3. Restart the daemon.
4. The skill is gone from the agent. `agent.toml` never contained it (its mtime never changed).

## Root cause

`set_agent_skills` (`crates/librefang-kernel/src/kernel/agent_state.rs`) updates the in-memory registry and calls `self.memory.substrate.save_agent(&entry)` (DB), but **never calls `self.persist_manifest_to_disk(agent_id)`**. `set_agent_mcp_servers` has the same gap.

The sibling setters all DO persist to disk: `set_agent_model`, `set_agent_channels`, `set_agent_schedule`, `set_agent_tool_filters` (tools). Skills and MCP are the only two that don't — they are structurally identical otherwise, so this is an oversight, not a design choice.

On boot, `boot.rs` loads each agent from the substrate, reads its `agent.toml`, and when the projection differs **overwrites the DB with the disk manifest** (log: `Agent TOML on disk differs from DB, updating`). So any field written only to the DB (a skill assigned via this path) loses to the unchanged `agent.toml`.

## Fix

Add `self.persist_manifest_to_disk(agent_id);` to `set_agent_skills` and `set_agent_mcp_servers` after the successful `save_agent`, mirroring the sibling setters. (PR incoming.)

## Note

This regressed once before via a fork-local patch that was never upstreamed and got dropped on a rebase; putting the call in the kernel setter (not the API route handler, which has since been split into `routes/agents/` modules by #5975) keeps it robust against API refactors.

DaBlitzStein:
## Summary

Closes #6045.

Assigning a skill or MCP server to an agent via the dashboard/API (`PUT /api/agents/{id}/skills` / `.../mcp_servers`) persisted only to the SQLite blob, **not** to `agent.toml`. Because boot reconciliation treats the on-disk manifest as authoritative, the assignment returned `200`, showed as `assigned` in the API, then was **silently wiped on the next daemon restart**.

## Root cause

`set_agent_skills` and `set_agent_mcp_servers` (`crates/librefang-kernel/src/kernel/agent_state.rs`) update the in-memory registry and call `save_agent` (DB) but **never call `persist_manifest_to_disk`**. The sibling setters all do: `set_agent_model`, `set_agent_channels`, `set_agent_schedule`, `set_agent_tool_filters` (tools). Skills and MCP were the only two missing it — structurally identical otherwise, so this was an oversight.

On boot, `boot.rs` loads each agent from the substrate, reads its `agent.toml`, and when the projection differs **overwrites the DB with the disk manifest** (`"Agent TOML on disk differs from DB, updating"`). Anything written only to the DB is therefore lost.

## Change

- Add `self.persist_manifest_to_disk(agent_id);` to `set_agent_skills` and `set_agent_mcp_servers`, after the successful `save_agent`, mirroring the sibling setters.
- Placed in the **kernel setter** (not the API route handler) so it survives the `routes/agents.rs` → `routes/agents/` module split (#5975) and any future API refactor.

## Verification

- New regression test `test_set_agent_skills_persists_allowlist_to_agent_toml` (kernel lib): boots a kernel, installs a skill, spawns an agent with a known `source_toml_path`, calls `set_agent_skills`, and asserts the on-disk `agent.toml` contains the assigned skill. Passes (`1 passed`).


houko (commented):
**Missing regression test for `set_agent_mcp_servers`**

The PR adds `self.persist_manifest_to_disk(agent_id)` to both `set_agent_skills` and `set_agent_mcp_servers`, but the regression test (`test_set_agent_skills_persists_allowlist_to_agent_toml`) only covers `set_agent_skills`. There is no parallel test for `set_agent_mcp_servers`.

The same silent-wipe-on-restart bug existed in both setters, and the fix is structurally identical. Per project convention (CLAUDE.md / #3721), a fix that changes two setters should be backed by a test at each injection site, not just one.

Please add `test_set_agent_mcp_servers_persists_allowlist_to_agent_toml` mirroring the existing test: boot kernel, spawn agent with a known `source_toml_path`, call `set_agent_mcp_servers([<server-name>])`, then read the on-disk `agent.toml` and assert it contains the assigned server name.

---

houko (commented):
Two CLAUDE.md style violations to fix before merge.

---

**1. `CHANGELOG.md` — four sentences packed onto one continuation line**

CLAUDE.md is explicit: *"CHANGELOG.md bullets — one sentence per line within a bullet."*

The new `fix(kernel)` entry's continuation line contains four sentences on a single line.
Split into one sentence per line:

```markdown
- **fix(kernel): persist per-agent skill & MCP-server assignments to `agent.toml` so they survive a restart** (#6045) (@DaBlitzStein).
  `set_agent_skills` and `set_agent_mcp_servers` wrote only the SQLite blob and never called `persist_manifest_to_disk`, unlike the sibling setters (`set_agent_model` / `set_agent_channels` / `set_agent_schedule` / `set_agent_tool_filters`).
  Since boot reconciliation re-syncs each agent from its on-disk `agent.toml` and overwrites DB-only fields, a skill/MCP assigned via the dashboard returned `200`, showed as assigned in the API, then was silently wiped on the next daemon restart.
  Both setters now persist to disk.
  Regression test: `test_set_agent_skills_persists_allowlist_to_agent_toml`.
```

---

**2. `crates/librefang-kernel/src/kernel/agent_state.rs` — two multi-line `//` comment blocks**

CLAUDE.md: *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."*

The first block (8 lines before `set_agent_skills`'s `persist_manifest_to_disk` call) and the second (4 lines before the MCP one) both violate this.
The WHY is non-obvious and worth a comment, but it should fit in one line each.

Suggested replacements:

```rust
// Without this, boot reconciliation silently overwrites DB-only fields from the on-disk agent.toml on restart.
self.persist_manifest_to_disk(agent_id);
```

```rust
// Same persistence requirement as set_agent_skills: boot reconciliation overwrites DB-only fields from agent.toml.
self.persist_manifest_to_disk(agent_id);
```

The detailed mechanism (boot.rs log message, sibling list) belongs in the PR body — where it already appears — rather than in an inline comment that will rot as the codebase evolves.

---

houko (commented):

**Missing regression test for `set_agent_mcp_servers`**

The PR adds `persist_manifest_to_disk` to both `set_agent_skills` (lines ~460–475 of `agent_state.rs`) and `set_agent_mcp_servers` (lines ~534–540), and correctly adds `test_set_agent_skills_persists_allowlist_to_agent_toml` for the skills path. But there is no parallel test for the MCP-servers path — a future regression in that setter could go undetected.

Please add `test_set_agent_mcp_servers_persists_servers_to_agent_toml` alongside the existing test, using the same pattern (spawn agent, call `set_agent_mcp_servers`, `read_to_string` the agent.toml, assert the server name appears). The two paths got identical fixes; they should have identical coverage.

No other issues found.

houko (dismissed):
Requesting changes to consolidate the three unresolved review points so this doesn't merge with known gaps. All are small and the fix logic itself is correct — just needs the coverage and style cleanup before merge.

1. **Missing regression test for `set_agent_mcp_servers`** (raised 06-10 and 06-12, still open). The PR fixes both `set_agent_skills` and `set_agent_mcp_servers`, but only the skills path has a test. Add `test_set_agent_mcp_servers_persists_allowlist_to_agent_toml` mirroring the existing one: boot kernel, spawn agent with a known `source_toml_path`, call `set_agent_mcp_servers([<server>])`, read the on-disk `agent.toml`, assert the server name is present. Two identical fixes should have identical coverage.

2. **CHANGELOG continuation packs four sentences on one line** (raised 06-11). CLAUDE.md: "CHANGELOG.md bullets — one sentence per line within a bullet." Split the new `fix(kernel)` entry's continuation into one sentence per line.

3. **Two multi-line `//` comment blocks in `agent_state.rs`** (raised 06-11). The 8-line block before `set_agent_skills`'s `persist_manifest_to_disk` call and the 4-line block before the MCP one violate the kernel crate's one-line-per-comment convention. Keep the WHY but compress each to a single line; the detailed mechanism (boot.rs log, sibling list) already lives in the PR body.

None of these are logic blockers — the persist fix is correct and the breadth is complete (skills + mcp were the only two setters missing the call). Re-request review once they're addressed.

PR #6051, opened 2026-06-09
fix(types): self-referential $ref in a tool schema overflows the stack (DoS from untrusted MCP/skill schemas)

houko:
## Problem

`normalize_schema_for_provider` runs for **every non-Anthropic provider** and inlines `$ref`/`$defs` via `resolve_refs` → `inline_refs`. The inner `inline_refs` cloned the referenced definition and immediately recursed into the freshly-inlined copy, with **no cycle or depth guard**:

```rust
if let Some(def) = defs.get(name) {
    *val = def.clone();
    inline_refs(val, defs);   // recurse into a copy that still contains the same $ref
    return;
}
```

A self-referential `$defs` makes this recurse forever and overflow the stack, aborting the daemon:

```json
{ "type": "object",
  "properties": { "child": { "$ref": "#/$defs/Node" } },
  "$defs": { "Node": { "type": "object",
      "properties": { "child": { "$ref": "#/$defs/Node" } } } } }
```

Mutually-recursive defs (`A → B → A`) trigger the same. Tool schemas come from **untrusted MCP servers and marketplace skills** (the function itself notes "Some MCP servers / skill definitions serialize schemas as strings"), so the input is attacker-influenced — a malicious or buggy third-party MCP server / skill can crash the daemon with one tool definition.

The sibling `properties`/`items` recursion in `normalize_schema_for_strict_validators` was likewise **depth-unbounded**, so a maliciously deep (acyclic) schema could overflow too.

## Fix

- **Cycle guard:** thread an active-`$ref`-name set through `inline_refs`. When a ref re-enters one already being expanded on the current path, stop and leave a permissive `{"type":"object"}` placeholder instead of recursing. This breaks `Node → Node` and `A → B → A` after one expansion.
- **Depth backstop:** add a shared `MAX_SCHEMA_DEPTH = 128` to both `inline_refs` and `normalize_schema_for_strict_validators` (threaded as a `depth` counter), so a deep acyclic schema truncates to an empty object schema rather than overflowing. Real tool schemas nest only a handful of levels.

## Verification

Three regression tests — each overflowed the stack before the fix:

```
cargo test -p librefang-types --lib tool::
test result: ok. 41 passed; 0 failed
```

- `self_referential_ref_does_not_overflow`
- `mutually_recursive_refs_do_not_overflow`
- `deeply_nested_schema_is_truncated_not_overflowed`

`cargo clippy -p librefang-types --lib` is clean.

## How it was found

Multi-agent audit of the workspace; the cyclic-`$ref` path was traced through `normalize_schema_for_provider` and confirmed before fixing.

### DIFF
--- a/crates/librefang-types/src/tool.rs
+++ b/crates/librefang-types/src/tool.rs
@@ -271,6 +271,12 @@ pub struct DecisionTrace {
 // Real tool schemas nest only a handful of levels; this bound is far above any legitimate use.
 const MAX_SCHEMA_DEPTH: usize = 128;
 
+// Total `$ref`-expansion budget shared across one schema normalization.
+// The path-scoped cycle set and the depth cap bound *chain length*, not *total work*: an acyclic schema of ~40 chained `$defs` where each definition holds two `$ref`s to the next (A1 → {2×A2}, A2 → {2×A3}, …) stays under `MAX_SCHEMA_DEPTH` yet forces ~2^40 definition clones — a kilobyte-sized tool definition from a malicious MCP server could still hang/OOM the daemon.
+// Once the budget is spent, further `$ref`s collapse to the same `{"type":"object"}` placeholder used for cycles.
+// Real tool schemas expand at most a few dozen refs, so 4096 is comfortably above any legitimate use while keeping worst-case work small (each expansion clones one definition from the input).
+const MAX_SCHEMA_EXPANSIONS: usize = 4096;
+
 /// Normalize a JSON Schema for cross-provider compatibility.
 ///
 /// Several providers (Gemini, Groq, OpenAI strict mode, …) ship strict JSON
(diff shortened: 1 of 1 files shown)


houko (commented):
**Daily code-review pass — one fix pushed (b2644a1)**

The security fix itself is correct: the cycle guard (`active` push/pop around `$ref` expansion) and the shared depth backstop are both sound, and the three regression tests directly cover the described overflow paths.

One structural comment issue was fixed in b2644a1:

**Doc-comment mis-attachment (`MAX_SCHEMA_DEPTH`)**
The new `///` lines for `MAX_SCHEMA_DEPTH` were appended directly to the `normalize_schema_for_provider` `///` block with no break between them. In Rust a continuous `///` run attaches to the *next* item — which here was `const MAX_SCHEMA_DEPTH`, not the public function — so `normalize_schema_for_provider` was left without a doc comment entirely. The fix changes the constant to `//` comments (it's private, no doc needed) and moves them above the function's doc block.

The same commit also re-flows the three newly-added comment blocks to one-sentence-per-line per the repo's prose-wrapping rule (CLAUDE.md: "The only legitimate line break inside a paragraph is at a complete sentence boundary … one sentence = one line").

---

PR #6098, opened 2026-06-12
fix(llm-drivers): treat deepseek-v4-pro as thinking-with-tools so reasoning_content is echoed

### LINKED ISSUE #6097: deepseek-v4-pro hits 400 "reasoning_content must be passed back" — not recognised as thinking-with-tools
## Summary

`deepseek-v4-pro` agents fail every multi-turn tool-call conversation with:

```
400 {"error":{"message":"The reasoning_content in the thinking mode must be passed back to the API.","type":"invalid_request_error"}}
```

## Root cause

`is_deepseek_v4_thinking_with_tools` (`crates/librefang-llm-drivers/src/drivers/openai.rs`) only matches `deepseek-v4-flash`. `deepseek-v4-pro` was **deliberately excluded** with the comment *"V4 Pro is reported as working out-of-the-box (#4842 workaround section) — must not be lumped in with V4 Flash"* and an explicit `assert!(!...)` test.

Production proves that assumption wrong (or DeepSeek changed the contract): `deepseek-v4-pro` has the **same** echo requirement as V4 Flash — once a tool-call thinking turn is in the history, the next request must echo its `reasoning_content`, or DeepSeek returns 400. Because V4 Pro isn't matched, it never gets `ReasoningEchoPolicy::Echo`, so the content is dropped and every subsequent turn 400s.

## Impact

A delegated agent running `deepseek-v4-pro` dies on its first tool-call follow-up turn: `agent_send` (blocking and async) and shared-queue tasks to it never complete. A sibling agent on the same model only avoided it by keeping a long enough history window that it never trimmed off the thinking turn.

## Fix

Match `deepseek-v4-pro` in `is_deepseek_v4_thinking_with_tools` (alongside Flash) so the `Echo` policy applies. PR incoming.

DaBlitzStein:
## Summary

Fixes #6097. `deepseek-v4-pro` agents 400 on every multi-turn tool-call conversation with *"The reasoning_content in the thinking mode must be passed back to the API."*

`is_deepseek_v4_thinking_with_tools` only matched `deepseek-v4-flash`; V4 Pro was excluded on the #4842 assumption that it "works out-of-the-box". Production disproves that — V4 Pro has the same echo requirement, so without `ReasoningEchoPolicy::Echo` its `reasoning_content` is dropped and DeepSeek rejects the next request. A delegated agent on `deepseek-v4-pro` therefore died on its first tool-call follow-up turn (so `agent_send` / shared-queue tasks to it never executed).

## Change
- `is_deepseek_v4_thinking_with_tools` now matches `deepseek-v4-pro` as well as `deepseek-v4-flash` (substring match keeps date-stamped variants working).
- Doc comment updated to record why the original exclusion was wrong.
- Moved `deepseek-v4-pro` from the `does_not_match_others` test to the `matches` test (plus `DeepSeek-V4-Pro` and a date-stamped variant).

## Verification
- `cargo test -p librefang-llm-drivers --lib is_deepseek_v4` → 2 passed, 0 failed.
- Scope kept minimal (single function + its tests + CHANGELOG); no catalog/history-fold changes bundled.


DaBlitzStein:
Moving to **draft** — honesty note on the evidence.

This change is a defensible contract-alignment (`deepseek-v4-pro` genuinely *is* a thinking-with-tools model: its assistant turns carry `thinking` + `tool_use` blocks), and the original exclusion rested on the now-questionable "#4842 works out-of-the-box" assumption.

**But I can't currently confirm it fixes a problem that manifests in normal use.** The incident that prompted it was an agent on `deepseek-v4-pro` 400ing with *"reasoning_content must be passed back"*. On closer inspection a **sibling agent on the same model** runs 100+ `thinking + tool_use` assistant turns **without** the echo policy and never hits the 400 — so the missing echo policy can't be the whole story. The real differentiator between the two agents is `max_history` (aggressive trimming on the failing one), not the echo policy.

Holding as draft until I can confirm whether `deepseek-v4-pro` actually requires the echo (and isn't getting it some other way for the working agent). Did not want to put an unconfirmed fix in the review queue.

DaBlitzStein:
**Confirmed live — promoting out of draft.**

Deployed this fix to a host running a `deepseek-v4-pro` agent that was 400ing on every tool-call continuation with *"reasoning_content must be passed back"*. After the deploy the agent runs its multi-turn tool-calling task cleanly — no 400, no agent-loop failure, `reasoning_content` round-tripped on each turn:

```
usage prompt_tokens=23238 completion_tokens=100   ✓
Captured reasoning_content from response len=1417  → completion_tokens=696  ✓
completion_tokens=3185                              ✓
Captured reasoning_content len=4852                 ✓
```

So `deepseek-v4-pro` does require the same `reasoning_content` echo as V4 Flash; treating it as thinking-with-tools fixes it. (The sibling agent on the same model that appeared to work without the echo turned out to be a red herring for the fix — both echo it now and both run fine.)

houko:
Reviewed — the change is correct, minimal, and well-tested. Two non-blocking notes.

**1. Fix targets the fallback, not the catalog — confirm the policy actually reaches the wire.**
`fallback_reasoning_echo_policy` only fires when the request carries no explicit policy (`ReasoningEchoPolicy::None` — catalog miss or pre-registry build), per its own doc comment.
If a community-catalog entry for `deepseek-v4-pro` ever sets a non-`None` `reasoning_echo_policy`, the request carries that value and this fallback never runs, making the change a no-op for catalog-resolved requests.
The added test verifies `is_deepseek_v4_thinking_with_tools("deepseek-v4-pro") == true` at the function level, but not end-to-end that `Echo` lands on the actual outbound request.
Low risk in practice: there is no in-tree catalog entry for the model (builtins sync from the external registry into `home_dir/providers/`, default `None`), and the production 400 was observed on exactly this fallback path — so the fix targets the right layer.
Flagging only so a follow-up that adds a catalog entry doesn't silently shadow this.

**2. Breadth check — no parallel site needs the same change.**
The other model-aware thinking path, `parse_labeled_summaries` / `strip_think_preamble` in `crates/librefang-runtime/src/history_fold.rs`, is model-agnostic (it strips any leading `<think>…</think>` block; `deepseek-v4-flash` is only named as an example, not gated on). So `is_deepseek_v4_thinking_with_tools` is the only model-gated point, and this PR covers it.

**Nit:** the CHANGELOG bullet carries only `(@DaBlitzStein)` — sibling entries in the same `### Fixed` block include the issue/PR number too (e.g. `(#6083)`). Consider appending `(#6097)` for consistency.

houko (approved):
Approved. Correct, minimal, well-tested fix; the two notes on my earlier comment are non-blocking.

PR #6101, opened 2026-06-14
fix(channels): preserve caller-supplied channel name case in channel_send (#6078)

### LINKED ISSUE #6078: channel_send lowercases caller-supplied channel name → lookup fails for capitalized sidecars
## Summary

`channel_send` tool unconditionally lowercases the `channel` parameter before adapter lookup. If a `[[sidecar_channels]]` entry has any uppercase letter in `name` (e.g. `bot-A`), `channel_send(channel="bot-A", ...)` fails with **"channel not found"** because the registered adapter key preserves the original case while the lookup key is folded to `bot-a`.

Plain text replies through the agent loop (incoming → outgoing same channel via sidecar) are unaffected — only the explicit `channel_send` tool path fails. This is most visible for **voice / file attachments** because they have no other delivery route.

## Reproducer

1. Add a sidecar with a name containing uppercase:
```toml
[[sidecar_channels]]
name = "bot-A"
channel_type = "telegram"
...
```
2. From an agent with `channel_send` capability, call:
```
channel_send(channel="bot-A", recipient="<chat_id>", file_path="/path/to/audio.ogg")
```
3. Tool returns `Channel Send failed` — adapter "bot-A" not found.
4. Rename to lowercase (`name = "bot-a"`) and the same call succeeds.

## Root cause (verified by inspection at v2026.6.10-beta.17)

`crates/librefang-runtime/src/tool_runner/channel.rs:140`:
```rust
let channel = input["channel"]
    .as_str()
    .ok_or("Missing 'channel' parameter")?
    .trim()
    .to_lowercase();   // ← folds caller-supplied name
```

The lowercased value is then passed to `kh.send_channel_media(&channel, ...)` / `kh.send_channel_text(...)` which look up the adapter by exact key. Adapter registration (`channel_bridge.rs`, post-#5996) keeps `sidecar_config.name.clone()` as the account_id with its original case → lookup mismatch.

## Suggested fix

Either:
- **A.** Drop the `.to_lowercase()` on caller input — caller-supplied case should match registration. (Simpler; behavior-preserving for already-lowercase setups.)
- **B.** Lowercase **both sides** consistently — fold at registration AND at lookup. Requires touching the registration path so existing in-the-wild capital-cased sidecars don't break silently.

(A) is minimal and matches the intent visible in #5996 (preserve sidecar name as the routing key).

## Impact

- Multi-bot Telegram setups that use any capital in sidecar `name` cannot send voice/file via `channel_send`.
- Workaround: rename sidecar `name` to lowercase. Trade-off: dashboards / config remain harder to read.

## Environment

- LibreFang `2026.6.10-beta.17`
- macOS (aarch64)
- Sidecar: `librefang-sidecar-telegram` (Rust, bundled)
- Discovered while verifying #5955 fix in beta.17 — text routing isolation works as expected, voice path surfaced the residual fold.

## Related

- #5955 / PR #5996 — `account_id` propagation that made this surface visible (qualified `channel: telegram:<name>` keys now preserve case at registration)
- Workaround on operator side has been used since 2026-05-31; filing now that the fix path is clear.

houko:
## Summary

`channel_send` force-lowercased the `channel` argument before the kernel's case-sensitive `send_channel_*` lookup, while channel adapters register under their config name with original case. Any sidecar whose name carries uppercase (e.g. `bot-A`) became unreachable — a latent regression since the case-preserving registration in #5996.

## Change

Pass the channel name through verbatim (still trimmed) in `crates/librefang-runtime/src/tool_runner/channel.rs`.

## Behaviour note

Channel matching is now case-sensitive (was effectively lowercase-only). Operators who registered an uppercase sidecar and worked around the bug by calling `channel_send` with a lowercased name must use the exact registered case. Already-matching setups are unaffected.

## Test plan

- `cargo build -p librefang-runtime`
- `cargo clippy -p librefang-runtime -- -D warnings`

Closes #6078

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -785,6 +785,8 @@ _308 PRs from 7 contributors since v2026.5.17-beta.12._
 
 ### Fixed
 
+- **channels: `channel_send` no longer lowercases the caller-supplied channel name, so capitalized sidecars are reachable again** (@houko). The `channel` tool argument was force-lowercased before the kernel's case-sensitive `send_channel_*` lookup, while channel adapters register under their config name with original case — so an agent calling `channel_send(channel="bot-A", …)` looked up `bot-a` and failed with a not-found error for any sidecar whose name carried uppercase (a latent regression since the case-preserving registration in #5996). The name is now passed through verbatim (still trimmed). Closes #6078.
+
 - **channels: the dashboard configure form is no longer empty for sidecar adapters when `librefang-sdk` is not pip-installed** (@houko).
   The Add-a-channel form is schema-driven off `python3 -m librefang.sidecar.adapters.<name> --describe`, but the boot-time probe (`routes/sidecar_describe.rs::describe_sidecar`) spawned the interpreter without the binary-embedded SDK on `PYTHONPATH` — unlike the live channel-spawn path, which has injected the embedded copy since the SDK was bundled.
   So on a fresh host with only `python3` (no `pip install librefang-sdk`), `--describe` failed with `ModuleNotFoundError`, and every adapter without a hand-maintained `static_fields` fallback (telegram, ntfy, gotify, mastodon, …) rendered a blank configure drawer — even though the adapter source ships embedded in the daemon binary.
--- a/crates/librefang-runtime/src/tool_runner/channel.rs
+++ b/crates/librefang-runtime/src/tool_runner/channel.rs
@@ -133,11 +133,7 @@ pub(super) async fn tool_channel_send(
 ) -> Result<String, String> {
     let kh = require_kernel(kernel)?;
 
-    // Preserve the caller-supplied channel name verbatim (#6078). Channel
-    // adapters register under their config name with original case
-    // (`channel_bridge`), and the kernel's `send_channel_*` lookups are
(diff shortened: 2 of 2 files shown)


houko (commented):
Two things to address before merge:

**1. Missing regression test**

The behavior of `tool_channel_send` changed (channel names are no longer lowercased), but there is no `#[tokio::test]` covering this path.
Per project rules (CLAUDE.md "MANDATORY: Integration Testing"), any behavior change in a tool runner should be accompanied by a test that would catch a revert.
A minimal unit test in `crates/librefang-runtime/src/tool_runner/channel.rs` (or a dedicated `tests/` file in that crate) should:
- call `tool_channel_send` with `channel = "Bot-A"` and assert the kernel lookup uses `"Bot-A"`, not `"bot-a"`.

**2. Test plan lists `cargo build`**

`cargo build -p librefang-runtime` is explicitly blocked by project convention (CLAUDE.md: "Do NOT run `cargo build` or `cargo run` locally").
The correct local verification is:
```
cargo check --workspace --lib
cargo clippy --workspace --all-targets -- -D warnings
cargo test -p librefang-runtime
```

(I've already pushed one mechanical fix to this branch — the 5-line comment block was collapsed to a single line per the "one short line max" comment rule.)

---

Issue #5120, opened 2026-05-16
memory: memory_store accepts attacker-controlled peer:<id>: key prefix


houko:
## Problem
`MemoryAccess::memory_store(key, value, agent_id, peer_id)` accepts an arbitrary `key` string with no validation. `tool_memory_store` passes the LLM-supplied key through unchanged. An agent with no `peer_id` context can store at `key = "peer:victim:user_name"`, landing it in the shared namespace under that literal string. `memory_list(Some("victim"))` then strips `peer:victim:` and surfaces the planted entry as if `victim` had written it.

## Evidence
- `crates/librefang-kernel/src/kernel/handles/memory_access.rs:12-72` — `memory_store` accepts `key: &str` unchanged.
- `crates/librefang-runtime/src/tool_runner.rs:4757-4768` — `tool_memory_store` passes `key` through with only JSON-shape validation (string, non-empty).
- Combined with the `memory_list` prefix-strip at `memory_access.rs:99-113`, the planted row appears as a legitimate peer entry.

## Impact
Cross-tenant data pollution. A compromised or misbehaving agent can poison any other peer's namespace. This survives the per-agent isolation added in #5071 because the attack uses the shared (`agent_id = None`) namespace where peer-scoped keys live.

## Suggested fix
1. Reject `key` starting with `peer:` at the `memory_store` boundary in `memory_access.rs:12-72`. The `peer:` prefix is the substrate's own namespace marker — agents must not write it directly.
2. Move peer scoping out of the key string entirely (separate column) so an LLM-supplied key string can never collide with the internal namespace marker.

Issue #5119, opened 2026-05-16
memory: peer_scoped_key is non-injective, causes cross-peer key collisions


houko:
## Problem
`peer_scoped_key(key, peer_id) = format!("peer:{pid}:{key}")` is not injective: the colon delimiter is not escaped in either `pid` or `key`, and `memory_list` recovers peer keys by `strip_prefix("peer:{pid}:")`. A peer id containing `:` collides with a different (peer_id, key) pair, leaking the latter's keys into the former's listing.

## Evidence
- `crates/librefang-kernel/src/kernel/manifest_helpers.rs:517-522` — `peer_scoped_key` construction.
- `crates/librefang-kernel/src/kernel/handles/memory_access.rs:101-104` — `memory_list` filter:

```rust
let prefix = format!("peer:{pid}:");
all_keys.into_iter()
    .filter_map(|k| k.strip_prefix(&prefix).map(|s| s.to_string()))
```

The test at `kernel/tests.rs:1746-1748` pins this fragile format.

Concrete collision: peer `"u"` writing key `"456:prefs.color"` and peer `"u:456"` writing key `"prefs.color"` both serialize to `peer:u:456:prefs.color`. `memory_list(Some("u"))` returns the stripped tail `456:prefs.color` for both rows.

## Impact
Cross-peer key leakage in `memory_list`, ambiguous identity for any peer_id that contains `:` (chat IDs from some channels do — e.g. `user:42`, Slack `T1:U2`). The two writers' values overwrite each other on the same key, and the operator has no warning that one peer is reading another peer's data.

## Suggested fix
1. Percent-encode `:` (and any reserved delimiter) in `pid` and `key` at write time; decode on read.
2. Or store peer/key as separate columns (`agent_id`, `peer_id`, `key`) instead of synthesizing a composite primary key.
3. Reject `:` in peer_id and key at the `memory_store`/`memory_list` boundary.

PR #6102, opened 2026-06-14
fix(memory): percent-encode colons in peer_id so Matrix peers can use Memory (#6100)

### LINKED ISSUE #6100: Memory current is incompatible with Matrix
### Description

I've set up Librefang in combination with my Matrix server as a channel. I am now witnessing Memory fail consistently. It appears that this is due to the `peer_id`, by default contains a colon. It would be awesome to be able to use Memory in combination with Matrix.

Huge thanks for this awesome software!

### Expected Behavior

Memory works when using Matrix as a channel.

### Steps to Reproduce

1. Combine Librefang with Matrix channel integration
2. Witness memory functionality fail

### LibreFang Version

2026.6.11-beta.18

### Operating System

Linux (aarch64/ARM64)

### Logs / Screenshots

Image

houko:
## Summary

#5119 / #5120 made the per-peer key framing `peer:{pid}:{key}` injective by rejecting any `peer_id` containing `:`. Matrix user IDs are natively `@user:matrix.org`, so every Matrix user was locked out of `memory_store` / `memory_recall` / `memory_list` with an `InvalidInput` error.

## Change

Instead of rejecting, percent-encode the colon before it enters the key (`escape_peer_id`: `%` → `%25`, `:` → `%3A`). The framing stays injective: peer `T1` (prefix `peer:T1:`) can no longer strip the escaped key `peer:T1%3AU2:…` of peer `T1:U2`, so the cross-peer isolation boundary from #5119 is preserved — now enforced by escaping rather than rejection. Colon-free peer_ids encode to themselves, so existing rows are byte-identical and need no migration. Empty `peer_id` and `peer:`-prefixed keys are still rejected.

Files:
- `manifest_helpers.rs`: new `escape_peer_id`; `peer_scoped_key` escapes instead of rejecting colons.
- `memory_access.rs`: `reject_bad_peer_id` drops the colon check; `memory_list` builds the recovery prefix from the escaped pid.
- `agent_execution.rs` / `messaging.rs`: stale "rejects colon-bearing" comments updated.

## Tests

- Updated `test_peer_scoped_key` (unit): colon peer_ids now assert escaped output, plus Matrix and `%`-escaping cases and a `T1` / `T1:U2` disjoint-prefix assertion.
- Rewrote the #5119 / #5120 security integration test (`kernel_handle_contract_memory.rs`): colon peer_ids are accepted and round-trip; a colon-prefix peer cannot read another peer's rows; legacy raw-colon plants stay invisible to the escaped peer.

## Test plan

- `cargo test -p librefang-kernel --lib test_peer_scoped_key`
- `cargo test -p librefang-kernel --test kernel_handle_contract_memory`
- `cargo clippy -p librefang-kernel -- -D warnings`

Closes #6100


houko (commented):
The CHANGELOG diff replaced the `channels: the dashboard configure form is no longer empty…` entry header with the new memory entry header, leaving its five continuation lines orphaned under the memory bullet (the `The Add-a-channel form…` block through `First-party adapter --describe is dependency-free…` on lines 789–793 of the PR-branch file). Pushed a one-line fix that restores the channels entry as a separate bullet so both fixes are correctly attributed.

---

PR #6103, opened 2026-06-14
feat(hands): consume a Codeberg-hosted skill registry via registry.registry_host (#6095)

### LINKED ISSUE #6095: Hands: 开发团队 支持 codeberg
### Description

Image

https://codeberg.org/

### Alternatives Considered

_No response_

### Additional Context

_No response_

houko:
## Summary

The registry sync path hardcoded GitHub's tarball/clone URLs, so Hands could only pull the skill registry from GitHub. This adds an optional `registry.registry_host` (full base URL, e.g. `https://codeberg.org`) so a Codeberg/Forgejo-hosted registry can be consumed.

## Change

- `RegistryConfig.registry_host: Option<String>` (`#[serde(default)]`, default `None` = GitHub). `Default` impl + config doc block updated.
- `registry_sync.rs`: a `registry_urls(Option<&str>)` helper derives the archive URL, git-clone URL, and tarball top-level prefix. `None` reproduces the exact pre-#6095 GitHub URLs (byte-identical, no migration). GitHub and Forgejo/Codeberg differ in the archive path (`/archive/refs/heads/main.tar.gz` vs `/archive/main.tar.gz`) and the tarball prefix (`librefang-registry-main/` vs `librefang-registry/`); handled without a forge trait/enum.
- Threaded `registry_host` through `sync_registry` / `refresh_registry_checkout` / `git_clone_fallback` / `download_and_extract` / `catalog_sync::sync_catalog_to` and all call sites.

## Scope / deferred

Registry **sync** only. Opening PRs to a Codeberg registry (the Forgejo API differs from GitHub's) and cross-platform login OAuth remain GitHub-only — both need a forge-API abstraction decision and are out of scope here.

## Design note

Any non-`None` `registry_host` is treated as the Forgejo scheme. This is correct for Codeberg (this issue's only target); a future GitHub-compatible host would need finer detection.

## Tests

- `registry_urls` unit tests: `None` yields the exact GitHub URLs/prefix; `Some("https://codeberg.org")` yields the Forgejo archive URL + `librefang-registry/` prefix; a trailing slash on the host is trimmed.

## Test plan

- `cargo build -p librefang-types -p librefang-runtime -p librefang-kernel -p librefang-api -p librefang-cli`
- `cargo test -p librefang-runtime registry_urls`
- `cargo clippy --workspace -- -D warnings`

Closes #6095


houko (commented):
**`docs/operations/config-reload.md` / `build_reload_plan` not updated for `registry_host`**

`crates/librefang-kernel/src/config_reload.rs` contains `build_reload_plan`, which classifies every `KernelConfig` field as hot-reload / restart-required / read-live. `docs/operations/config-reload.md` is the ops-facing reference table derived from it, and CLAUDE.md says a test guards against drift between the two.

This PR adds `registry_host` to `RegistryConfig` (part of `KernelConfig`) but doesn't touch either file. If `build_reload_plan` doesn't classify the new field the drift-guard test will fail in CI; if it does implicitly handle it (e.g. via a catch-all "read-live" branch) the docs table is stale.

Please check `build_reload_plan` and, if needed, add a row for `registry.registry_host` to `docs/operations/config-reload.md`. The field reads fresh from `cfg.registry.registry_host.as_deref()` on each sync call, so it is most likely "read-live" — but the table should say so explicitly.

---

Issue #6021, opened 2026-06-06
mcp server configuration in database


GQAdonis:
### Description

One of the difficulties I have had is with the fact that adding MCP servers at runtime requires writing to the configuration file, which means that on Kubernetes, the file cannot be in a ConfigMap--it must be a file on an attached drive. I would like an option (and would be willing to build it) where configuration first loaded into the configuration database and used from there, where the configuration can be handled in the database, which is easier to manage.

### Alternatives Considered

I have tried creating a volume and copying the file in an init container to that volume, but that had issues that could be solved, but I would consider keep all configuration in the database after initial load a better option.

houko:
The storage + boot-merge slice (#6106) has merged.

The remaining runtime wiring (API write-path, CLI, hot-reload, and config.toml -> DB import) is now tracked in #6113.
Closing this issue in favor of that follow-up.

PR #6106, opened 2026-06-14
feat(mcp): SQLite-backed MCP server config storage + boot merge (#6021)

houko:
## Summary

Bounded storage slice of #6021 — SQLite-backed MCP server configuration, so `config.toml` can be a read-only ConfigMap on Kubernetes and MCP servers can be managed in the database.

The codebase already has a mature SQLite substrate (a migration ladder + per-table stores sharing one r2d2 pool), so this extends that pattern rather than introducing new infra.

## Change

- **Migration v43** (`migration.rs`): `SCHEMA_VERSION` 42 → 43; `migrate_v43` creates `mcp_server_configs(name TEXT PK, entry_json TEXT NOT NULL, created_at, updated_at)` + an audit row. The full `McpServerConfigEntry` is stored as one JSON blob so the table tracks the struct without a per-field schema migration.
- **`McpConfigStore`** (new `librefang-memory/src/mcp_config_store.rs`): `upsert` / `get` / `load_all` / `delete` / `count` over the shared pool. `upsert` uses `ON CONFLICT(name) DO UPDATE` so `created_at` survives and `updated_at` refreshes; `load_all` is ordered by name for a deterministic boot merge. 6 unit tests.
- **Boot merge** (`kernel/boot.rs`): loads DB MCP configs and overlays them over the file-backed `[[mcp_servers]]` by name (DB overrides file; DB-only names appended; **empty table is a byte-for-byte no-op** so file-only setups are unchanged). On store error it logs `warn!` and continues with the file config — the DB is an augmentation, not a hard boot dependency.

## Scope / deferred (follow-ups)

- Rewire the API write-path (`POST /api/mcp/servers`) and the CLI (`librefang mcp add/rm`) onto `McpConfigStore` — needs a DB-vs-file write toggle (e.g. `[mcp] storage = "db"`) so K8s deployments opt into DB writes.
- Surface DB entries in `config_reload` / `reload_mcp_servers` for hot-reload (boot-time load only for now).
- A one-time `config.toml` → DB import path.

## Test plan

- `cargo test -p librefang-memory mcp_config_store`
- `cargo test -p librefang-memory migration`
- `cargo build -p librefang-kernel && cargo clippy -p librefang-memory -p librefang-kernel -- -D warnings`

Refs #6021 (storage + boot-merge slice; API/CLI wiring is a follow-up).


houko (commented):
Pushed one style commit (`5939a56`): CHANGELOG bullet split to one sentence per line, and multi-paragraph comment blocks in `boot.rs`, `migration.rs`, and `mcp_config_store.rs` compressed to one line each per project rules.
One finding left as a review comment: the boot-merge loop in `boot.rs` has no direct test (see inline note).

---

Issue #3745, opened 2026-04-27
Error system stringly-typed: LlmError::failover_reason matches lowercase substrings; LibreFangError flattens to String


houko:
**Location**
- `crates/librefang-llm-driver/src/lib.rs:71-205` (`failover_reason()`)
- `crates/librefang-types/src/error.rs:38, 52, 68, 72`

**Problem**
1. `failover_reason()` takes `LlmError::Api { status, message: String }`, lowercases, and substring-matches `"rate limit"`, `"credit"`, `"balance"`, `"context"`, `"model not found"` to decide retry vs failover. Provider rewords or i18n's the error → classification becomes `Unknown` overnight.
2. `LibreFangError` flattens reqwest / rusqlite / wasmtime errors to `String`:
   ```rust
   Memory(String)
   LlmDriver(String)
   Network(String)
   Serialization(String)
   ```
   Only `Io(#[from] std::io::Error)` keeps source. `e.source()` chain is dead at the top level — retry logic can't `match` on the underlying type, only re-parse the string.

**Impact**
Failover policy is a substring lookup that any provider tweak breaks. Higher-level retry/circuit-break logic can't make typed decisions.

**Fix**
- `LlmError::Api` gains `code: Option<ProviderErrorCode>` (rate_limit_exceeded / insufficient_quota / context_length_exceeded). Each driver populates it once when parsing the provider's JSON. `failover_reason()` reads code first, falls back to substring.
- `LibreFangError::{Memory, LlmDriver, Network}` change to `#[from]` typed errors; preserve source chain.

---
*Filed via automated multi-perspective audit (security / concurrency / architecture / error-handling / frontend / performance subagents). Curated against #3359–#3409 to avoid duplicates.*

PR #6107, opened 2026-06-14
refactor(runtime): migrate web_search.rs to ToolError (#3576 slice)

houko:
## Summary

One slice of the ongoing structured error-contracts migration (#3576). Migrates `crates/librefang-runtime/src/web_search.rs` from stringly-typed `Result<String, String>` to `Result<String, ToolError>`, reusing existing `ToolError` variants.

Per `docs/architecture/error-contracts.md`, the migration intentionally changes the LLM-visible error strings to the variant `Display` form — affected tests are re-pinned to the exact new strings (strengthened from `.contains()` to exact `.to_string()` asserts).

## Variant mapping

| Failure | Variant |
|---|---|
| absent API key / unconfigured SearXNG URL | `Unavailable` |
| `pageno == 0`, invalid category | `InvalidParameter { name, reason }` |
| reqwest `.send()`/`.json()`/`.text()` failures | `upstream(e)` — preserves the `reqwest::Error` source chain (#3745) |
| non-success status, no results, all providers failed | `upstream_msg(...)` |

Brave key-rotation logic preserved (matches on `e.to_string()`). Dispatch boundary still narrows `ToolError -> String` via `Display`. Stays in-crate (`librefang-runtime`); no cross-crate error-shape changes; does not add the `clippy::disallowed_types` gate.

## Test plan

- `cargo clippy -p librefang-runtime --all-targets -- -D warnings` (verified: clean)
- `cargo test -p librefang-runtime --lib web_search::` (verified: 24 passed)
- `cargo test -p librefang-runtime --lib tool_runner::error::` (verified: 14 passed)

Refs #3576 (one slice; does not close the issue).


crates/librefang-runtime/src/web_search.rs
```diff
@@ -164,8 +165,8 @@ impl WebSearchEngine {
         debug!("Auto: falling back to DuckDuckGo");
         match self.search_duckduckgo(query, max_results).await {
             Ok(result) if !result.trim().is_empty() => Ok(result),
-            Ok(_) => Err("All search providers failed; DuckDuckGo returned empty results (likely captcha-blocked). Configure a Brave, Tavily, Jina, Perplexity, or SearXNG provider for reliable search.".to_string()),
-            Err(e) => Err(format!("All search providers exhausted. Last error (DuckDuckGo): {e}")),
+            Ok(_) => Err(ToolError::upstream_msg("All search providers failed; DuckDuckGo returned empty results (likely captcha-blocked). Configure a Brave, Tavily, Jina, Perplexity, or SearXNG provider for reliable search.")),
+            Err(e) => Err(ToolError::upstream_msg(format!("All search providers exhausted. Last error (DuckDuckGo): {e}"))),
```
houko, line 169: The `Err(e) =>` arm wraps an existing `ToolError` in a new `upstream_msg(format!("… {e}"))`, converting `e` to its `Display` string and losing whatever source chain `e` carries (e.g. the `reqwest::Error` that `search_duckduckgo` may have returned).

The PR description says `reqwest` failures are routed via `ToolError::upstream` to preserve the source chain per #3745, but this specific arm re-wraps rather than propagates.

If the "all providers exhausted" context string matters, one option is to keep `upstream_msg` here since the message is the deliverable and the chain is already summarised in `{e}` — but that should be a conscious choice, not a side-effect. If the source chain matters at the dispatch boundary (e.g. for structured logging of the root `reqwest::Error`), the arm could propagate `e` directly and let `search_auto`'s callers see the DuckDuckGo error as-is.

---

houko (commented):
Pushed one style commit (`e20745c`) with three mechanical fixes:

- **CHANGELOG.md**: the new entry had four sentences on a single line; split to one sentence per line per the prose-wrapping rule (new prose is in scope).
- **dispatch.rs**: removed the `// #3576: search() returns Result<String, ToolError>; narrow here.` comment added by this PR — task-reference comments belong in the PR body, not in code.
- **web_search.rs tests**: removed three `// #3576:` comment blocks (two were multi-line) above the re-pinned assertions — same reason.

One ambiguous finding left as an inline comment: the `search_auto` DuckDuckGo fallback arm re-wraps `e: ToolError` in `upstream_msg(format!("… {e}"))`, which may lose the `reqwest::Error` source chain the PR says is preserved.

---

PR #6110, opened 2026-06-14
fix(docs): pin brace-expansion override to 2.0.2 to unbreak the Cloudflare docs build

houko:
## Type

- [x] CI / Tooling

## Summary

The `librefang-deploy` Cloudflare Workers build (which builds the `docs/`
Next.js site via the `@opennextjs/cloudflare` adapter) has been failing on
`main` and therefore on every PR built on top of it.

Root cause is the pnpm override added for the brace-expansion ReDoS advisory
(CVE-2025-5889):

```
"brace-expansion@>=2.0.0 <2.0.3": ">=2.0.3"
```

The target `>=2.0.3` is open-ended, so pnpm resolves it to the newest release
(`5.0.6`). brace-expansion 3.x+ is ESM-only and exposes no `default` export,
while `minimatch@8.0.7` — pulled in transitively by the OpenNext adapter —
does `import expand from 'brace-expansion'`. The build dies with:

```
SyntaxError: The requested module 'brace-expansion' does not provide an export named 'default'
```

## Fix

Pin the override to the patched CJS release `2.0.2`. It still satisfies the
advisory (the fix landed in 2.0.2) and keeps the CommonJS default-export
interop that `minimatch@8.0.7` relies on. Diff is two lines (package.json +
the lockfile `overrides` section); no other dependencies change.

## Verification

Reproduced and verified with `pnpm@10.11.1` (the version pinned in
`docs/package.json` `packageManager`):

| override value | resolved brace-expansion | `import minimatch` |
| --- | --- | --- |
| `>=2.0.3` (before) | `5.0.6` | SyntaxError: no `default` export |
| `2.0.2` (after) | `2.0.2` | OK |

## Security

- [x] No new unsafe code
- [x] Keeps the CVE-2025-5889 remediation (brace-expansion >= 2.0.2)


docs/package.json
```diff
@@ -71,7 +71,7 @@
     "overrides": {
       "picomatch@<2.3.2": ">=2.3.2",
       "picomatch@>=4.0.0 <4.0.4": ">=4.0.4",
-      "brace-expansion@>=2.0.0 <2.0.3": ">=2.0.3",
+      "brace-expansion@>=2.0.0 <2.0.3": "2.0.2",
```
houko, line 74: The override selector key still reads `brace-expansion@>=2.0.0 <2.0.3`, which classifies `2.0.2` as part of the vulnerable range (it satisfies `>=2.0.0 <2.0.3`). Setting the target to `2.0.2` while the key declares that version vulnerable is a contradiction that needs clarification.

Two possible truths:

1. **`2.0.3` was never published** — `>=2.0.3` always resolved to the ESM-only 3.x/5.x line, and `2.0.2` is the actual CVE-2025-5889-patched release. If so, this fix is correct, but the selector key should also be tightened to `<2.0.2` (or removed) so the override accurately reflects the vulnerable boundary. Leaving `<2.0.3` in the key while targeting `2.0.2` means packages that already resolved to the safe `2.0.2` will be caught by the override and "forced" to `2.0.2` — a no-op, but it misleads future readers about what version is actually unsafe.

2. **`2.0.3` exists and is the patched release** — In that case, pinning to `2.0.2` re-introduces CVE-2025-5889 and the target should be `2.0.3` (not `>=2.0.3`, to stay on CJS).

Could you confirm which scenario applies (e.g., link to the advisory or `npm view brace-expansion versions`) and, if it's scenario 1, also update the selector key to `<2.0.2`? That will make the override internally consistent and correct for the next person who reads it.

---

houko (commented):
One question on the version boundary before this merges — see inline.

---

Issue #5671, opened 2026-05-23
RFC: rethink channel inbound routing — HITL vs AITL topology, channel-instance binding, per-conversation /agent override


DaBlitzStein:
> This is a design RFC, not an implementation PR. Filing it to open the debate before any code lands. The full spec lives below; a short version is at the top. Feedback welcome especially on whether the architectural premise is correct before I send atomic PRs.

## TL;DR

The current `resolve_or_fallback()` inbound dispatch chain (`crates/librefang-channels/src/bridge.rs:2321-2398`) applies the same priority sequence to every inbound message regardless of topology (DM vs group vs anything else). Combined with `ThreadOwnershipRegistry`'s 5-minute TTL and `AgentManifest.channels = []` accept-all semantics (#5294 layers b/c), this causes production-visible bugs where a DM rotates between unrelated agents whenever the human goes quiet for longer than the TTL.

Proposed shift:

- **One channel instance hosts exactly one agent.** `[[sidecar_channels]]` already supports the instance concept via `name`; we make `agent` mandatory on it.
- **Two explicit topologies:** **HITL** (Human-In-The-Loop, DM-style) and **AITL** (Agent-In-The-Loop, group/channel space with other participants). Engagement semantics differ between the two; the agent-selection question never arises because each instance is 1:1.
- **Slash command `/agent <name>` operates at conversation scope**, not instance scope. Per-conversation override persisted in DB; instance default seeded from config. Two-level lookup at dispatch.
- **Multi-agent coordination stays inside LibreFang** (via `agent_send`); it does not happen \"across\" a single external channel.
- The fallback chain (`bindings → direct_routes → user_defaults → channel_defaults → \"assistant\" → list_agents().first()`) is **deleted** from the inbound path. Dispatch becomes a 2-level table lookup.

## Why the current code rotates agents on a DM

Reproduced in production journal:

\`\`\`
turn 1 @ 11:58:44   agent.name=researcher        session.id=7b992bc5...
turn 2 @ 16:35:44   agent.name=legal-assistant   session.id=7b6d3bac...
\`\`\`

Same Telegram bot, same human peer, single DM conversation. The peer sent two consecutive messages with ~4.5 hours between them. The bot ended up running each turn on a **completely different agent**, with **different sessions** (no shared history). The second agent introduced itself as a new identity (\"¡Hola! Soy legal-assistant…\") because it genuinely had no record of the prior conversation.

Sequence of decisions inside the daemon:

[...]

Issue #5323, opened 2026-05-20
feat(channels): per-conversation agent routing — extend ThreadOwnership + alias triggering for multi-agent groups


DaBlitzStein:
## Context

Builds on top of merged work in this area:

- **#3414** — `ThreadOwnershipRegistry` (channel + thread, TTL 5min, claim/refresh/re-claim).
- **#2619** — channel default agent's `metadata.routing.aliases` auto-merged into `group_trigger_patterns`.
- **#2480** — group addressee detection (`is_vocative_trigger`, `is_addressed_to_other_participant`, `SenderContext.group_participants`, flag `LIBREFANG_GROUP_ADDRESSEE_GUARD`).
- **#2522** — Telegram session scoped per `chat_id` (`SessionId::for_channel(agent, "<channel>:<chat>")`).
- **#3321** — `route_assistant_by_metadata()` per-agent attention scoring (aliases × pesos 6/2/1 + semantic bonus).

And work that was **attempted but never landed**:

- **#3419 / #3420 (both CLOSED)** — added `account_id: Option<String>` to `ThreadKey` for multi-tenant + lazy `sweep_expired` task. Verified in current `crates/librefang-channels/src/thread_ownership.rs:32`: `ThreadKey` still does not have `account_id`.
- **#2294 (CLOSED)** — bundled SQLite roster + group_members tool + reply precheck + alias triggering. Closed (no atomic split).

## The gap (multi-agent groups)

When a Telegram (or any) group has more than one LibreFang agent joined, the current code can't route an `@agentB` mention (or an alias declared by agentB) to agentB if agentB is not the channel's `default_agent`. Concrete failure modes verified in code, not speculation:

1. **Aliases wiring is single-agent.** `crates/librefang-api/src/channel_bridge.rs:1934-1969` only loads the **default agent**'s `metadata.routing.aliases`/`weak_aliases` into `group_trigger_patterns`. Aliases of non-default agents joined to the same channel are never merged.
2. **No `@agentname → agent_id` mapping.** `crates/librefang-channels/src/telegram.rs:2940-2943, 2983-3014` only sets `was_mentioned=true` for the bot's own username. Mentioning a specific agent by name is invisible to the bridge.
[...]

Issue #5294, opened 2026-05-19
bug(channels): inbound routing falls to non-deterministic "first available" agent when SidecarChannelConfig has no default + empty channels allowlist accepts all (related #4926)


DaBlitzStein:
## Summary

After migrating Telegram (and other channels) to sidecar via #5241, the `SidecarChannelConfig` struct has **no `default_agent` field** — only `name`, `command`, `args`, `env`, `channel_type`, restart/overflow knobs. The old `[channels.telegram]` config had `default_agent = "X"` which used to seed the `AgentRouter.channel_defaults` at boot. With sidecar, that path is gone.

Result: when an inbound message arrives on a sidecar channel, `AgentRouter::resolve_with_context` (`crates/librefang-channels/src/router.rs:202`) returns `None` because:
- No binding for `(channel, peer_id)` set
- No `channel_defaults["telegram"]`
- No `default_agent`

Then `resolve_or_fallback` (`crates/librefang-channels/src/bridge.rs:2780`) tries `find_agent_by_name("assistant")` → typically `None` → falls to **"first available agent"** — which is whichever agent the registry happens to iterate first (non-deterministic, ID/spawn-order-dependent).

## Repro

1. Existing setup: agent `fandangorodelo` with `channels = ["telegram"]` in its manifest, configured Telegram via `[[sidecar_channels]]`.
2. Create a second agent (any purpose, e.g. `Profesor`) without restricting its `channels` list (defaults to `channels = []`).
3. Send a Telegram message to the bot.
4. **Inbound message routes to `Profesor` instead of `fandangorodelo`** because (a) router has no `channel_defaults["telegram"]`, (b) router falls to first-available, (c) `Profesor` was created later but happens to win the iteration order.

Confirmed empirically in our deploy:

```
DEBUG run_agent_loop: Memory decision: ADD new: Has generated file at `output/horario_definitivo.md`...
  agent.name=Profesor  agent.id=600770a9-bfbc-5e93-b83d-c9e7729ca55c  session.id=9bdaf4de-...
```

The user wrote to `fandangorodelo` on Telegram; the loop ran on `Profesor`. File ended up in `~/.librefang/workspaces/agents/Profesor/output/`.

## Root cause — 3 layers

**Layer A**: `SidecarChannelConfig` (`crates/librefang-types/src/config/types.rs:2155`) is missing `default_agent: Option<String>`. The in-process `TelegramConfig`/`DiscordConfig`/etc. all have it.

[...]

PR #3419, opened 2026-04-27
fix(channels): multi-tenant ThreadKey + sweep_expired wiring (#3414 follow-up)

houko:
## Why

Follow-up review feedback for #3414 (thread ownership). Two real bugs and one documentation gap surfaced after merge:

1. **Multi-tenant key collision (security).** `ThreadKey` was `(channel, thread)` only. Two Slack workspaces (account A vs B) producing the same `thread_ts` collided in the registry — account A's claim shadowed B's distinct conversation. In a multi-workspace install this lets one workspace's thread-ownership decisions leak into another's routing.

2. **Unbounded registry growth (quality / DoS).** `sweep_expired` was implemented but had no caller. `decide` evicted only on access, so a daemon in a Slack workspace with churning `thread_ts` accumulated stale entries until restart. Registry grew without bound until the process died.

3. **REST `push_message` bypass clarification.** `push_message` (cron / workflow / `agent_send` / REST `/push`) intentionally does not consult the registry — caller has already chosen the sender, no routing decision to gate. Added a doc-comment so future readers don't mistake the asymmetry for a bug.

## What

- `ThreadKey { channel, account_id: Option<String>, thread }` — constructor takes `(channel, account_id, thread)`. Empty `account_id` normalizes to `None` so single-tenant traffic keeps a stable key.
- `BridgeManager::ensure_thread_ownership_sweep_started` — armed lazily on the first `start_adapter` call (so the sync constructor stays runtime-free for unit tests). Runs a 60-second `tokio::time::interval` tied to `shutdown_rx`.
- Doc-comment on `push_message` explaining the bypass and pointing to a future opt-in `respect_thread_ownership` parameter if user-visible pain shows up.

## Tests

Three new tests; all pass:

- `thread_ownership::tests::distinct_accounts_do_not_collide_on_same_thread_id`
- `thread_ownership::tests::empty_account_id_normalized_to_none`
- `bridge::tests::multi_tenant_account_id_flows_into_thread_key`

`cargo clippy --workspace --all-targets -- -D warnings` clean. `cargo test -p librefang-channels --lib --no-fail-fast` 843 passed.

## Out of scope

- Mention-spoofing — by design ("@-mention re-claims" matches the issue spec).
- Restart drops all claims — single-process MVP.
- Real-dispatch integration test for the gate's order-of-operations vs RBAC / journal / auto-reply — captured in helper's doc-comment for follow-up.

## Branch hygiene

[...]

PR #6127, opened 2026-06-16
feat(channels,kernel): per-conversation agent routing for multi-agent groups (#5323)

houko:
## Summary

Closes the multi-agent-group routing gap in #5323: when more than one agent serves a channel, a group message that names a specific non-default agent now reaches that agent instead of always falling to the channel default, and the conversation stays sticky to it.

Built on top of #5671 PR-A (`SidecarChannelConfig.agent` + `available_agents`, merged in #6105), which adopted Model A. This is the AITL routing layer that PR-A's CHANGELOG entry anticipated ("`resolve_or_fallback`, dispatch … arrive in later PRs"). All five steps from the issue land in one PR.

## What changed (per issue step)

**Step 1 — `ThreadKey` reshape (`thread_ownership.rs`).** The conversation-ownership claim key grows three optional slices, all defaulting to `None` so the historical `(channel, thread)` key is reproduced byte-for-byte:
- `account_id` — two bot accounts on one channel-type no longer collide (the unlanded #3419/#3420 fix).
- `chat_id` — two chats reusing a forum-topic id stay distinct.
- `peer_id` — per-sender stickiness, so two users in one thread can talk to two different agents without contaminating each other.

`decide_with_ttl` lets the bridge pass a per-channel TTL while every channel shares one registry.

**Step 2 + 4 — multi-agent alias/mention routing (`bridge.rs`, `channel_bridge.rs`).** A new resolution step runs ahead of the binding/default chain for group messages:
- Explicit `@`-mention surfaced by the adapter in `metadata["mention_names"]`, resolved against agent names/handles (channel-eligibility enforced via the existing `manifest.channels` allowlist).
- A non-default agent's `channel_overrides.group_trigger_patterns` alias matching the message text, scored by a new deterministic per-agent attention scorer `librefang_channels::bridge::best_alias_match` (reuses the compiled-regex cache; ties → `None`).

The dispatch path also consults the scorer before the previously non-deterministic "first available" fallback, closing layer (c) of #5294. The kernel-side `route_assistant_by_metadata_for_channel` enumerates channel-eligible agents and delegates to the cached scorer (no new `regex` dependency added to `librefang-api`).

**Step 1b — claim-aware dispatch + `include_dms`.** A topic-less group now claims by `chat_id` instead of bypassing the registry (gap #5). A live claim makes a follow-up sticky to the same agent without a fresh mention; an explicit address re-claims for the new agent, preserving #3334's refresh/re-claim TTL semantics. The two duplicated gates (text + multimodal paths) are unified into one `conversation_ownership_allows` helper.

**Step 5 — config knobs (`ChannelOverrides`).** `conversation_ownership_ttl_seconds` (default `600`) and `conversation_ownership_include_dms` (default `false`).

**Step 3 — sidecar mention surfacing (Python).** Telegram (UTF-16-correct entity parsing), Discord, Slack, and Matrix now attach `mention_names` and a per-group `sender_user_id` so the bridge can route to a named agent and scope claims per peer. Telegram previously set no mention metadata at all.

## Design notes

- The issue proposed a `metadata.routing.handles` JSON convention; aliases were already refactored to typed `channel_overrides.group_trigger_patterns`, so this PR routes against agent **names** + those typed aliases rather than reintroducing the untyped bag. The "`ambrogio, ayúdame`" example from the issue works purely through the alias scorer on message text — no platform `@`-mention needed.
- Routing respects the per-agent `manifest.channels` allowlist (the mechanism `resolve_or_fallback` already enforced). The channel-side `available_agents` whitelist from #5671 remains schema-only ("not yet consulted") and its enforcement is the forthcoming `/agent` PR — out of scope here.

## Verification

- `cargo clippy -p librefang-channels -p librefang-api -p librefang-types --lib -- -D warnings` — clean.
- `cargo test -p librefang-channels --lib` — 493 passed, incl. new `best_alias_match_*`, `build_thread_key_*`, `resolve_addressed_agent_routes_explicit_mention`, `same_thread_distinct_account_id_does_not_collide`, `same_thread_distinct_peer_does_not_collide`, `per_channel_ttl_overrides_default`.
- `cargo test -p librefang-types --lib config::` — 157 passed, incl. `test_conversation_ownership_knobs_roundtrip`.
- Python sidecar suites (telegram/discord/slack/matrix) — 288 passed, incl. new `test_inbound_mentions_and_sender_user_id` (covers the UTF-16 offset case).

Live LLM verification is not required — no LLM call path changed.

## Test-plan coverage (from the issue)

| Scenario | Covered by |
| --- | --- |
| `@agentB` in multi-agent group routes to B | `resolve_addressed_agent_routes_explicit_mention`, `best_alias_match_picks_unique_winner` |
| Follow-up from same peer (no mention) stays with B | sticky-holder branch in `resolve_or_fallback` |
| Another peer in same thread not contaminated | `same_thread_distinct_peer_does_not_collide` + per-peer `ThreadKey` |
| TTL expires → claim drops → re-evaluate | `per_channel_ttl_overrides_default` |
| Alias declared by non-default agentB routes to B | `best_alias_match_picks_unique_winner` |
| Topic-less group respects claim via chat_id | `build_thread_key_falls_back_to_chat_id_without_topic` |
| Same thread_id under different account_id no collision | `same_thread_distinct_account_id_does_not_collide` |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -822,6 +822,14 @@ In-crate only; no cross-crate error-shape changes.
 
 ### Added
 
+- **channels(routing): per-conversation agent routing for multi-agent groups** (#5323) (@houko) — the AITL routing layer on top of #5671 PR-A's `agent` / `available_agents` schema.
+  When more than one agent serves a channel, a group message that names a specific non-default agent now reaches that agent instead of the channel default.
+  Two addressing paths: an explicit `@`-mention the adapter surfaces in `metadata["mention_names"]` (resolved against agent names/handles), and a non-default agent's declared `channel_overrides.group_trigger_patterns` alias matching the text — scored by a new deterministic per-agent attention scorer (`librefang_channels::bridge::best_alias_match`, reusing the compiled-regex cache) that the channel dispatch path consults before the previously non-deterministic "first available" fallback (closes layer (c) of #5294).
+  `ThreadKey` (the conversation-ownership claim key) grows three optional slices — `account_id` (multi-tenant: two bot accounts on one channel-type no longer collide, the unlanded #3419/#3420 fix), `chat_id` (two chats reusing a forum-topic id), and `peer_id` (per-sender stickiness so two users in one thread can talk to two different agents without contaminating each other) — all defaulting to `None`, reproducing the historical `(channel, thread)` key byte-for-byte.
+  A topic-less group now claims by chat id instead of bypassing the registry, and a live claim makes a follow-up sticky to the same agent without a fresh mention; an explicit address re-claims for the new agent, preserving the #3334 TTL semantics.
+  New per-channel `[channel_overrides]` knobs: `conversation_ownership_ttl_seconds` (default `600`) and `conversation_ownership_include_dms` (default `false`).
+  The Telegram / Discord / Slack / Matrix sidecar adapters now surface `mention_names` and a per-group `sender_user_id` so the bridge can route and scope per peer.
+  Additive and backward-compatible: single-agent channels and existing configs are unchanged.
 - **hands/registry: consume a Codeberg-hosted skill registry via `registry.registry_host`** (#6095) (@houko) — the registry sync path hardcoded GitHub's tarball/clone URLs.
   A new optional `registry_host` (full base URL, e.g. `https://codeberg.org`) derives the archive URL, git-clone URL, and tarball top-level prefix from that host; unset (default `None`) reproduces the exact GitHub URLs, so existing setups stay byte-identical and need no migration.
   GitHub and Forgejo/Codeberg differ in more than the host (archive path `/archive/refs/heads/main.tar.gz` vs `/archive/main.tar.gz`; prefix `librefang-registry-main/` vs `librefang-registry/`), handled in a small `registry_urls()` helper without a forge trait/enum.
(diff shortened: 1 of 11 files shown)


houko (commented):
**Redundant scorer invocation in `resolve_or_fallback` (step 4) for group messages**

In `crates/librefang-channels/src/bridge.rs`, `resolve_or_fallback` calls `route_assistant_by_metadata_for_channel` twice on the same message text when `message.is_group` is true and the routing chain returns `None`:

- **Step 2** — inside `resolve_addressed_agent`, which is only called for groups. If no explicit `@`-mention resolves, it calls `route_assistant_by_metadata_for_channel(ct, text)`.
- **Step 4** — unconditionally after the routing chain, calling `route_assistant_by_metadata_for_channel(ct, text)` again.

Both calls use identical arguments (`ct`, `text`) against the same candidate set (channel-eligible agents whose `manifest.channels` allows the channel). `best_alias_match` is deterministic and pure, so the second result is always identical to the first. The step-4 invocation is unreachable-in-effect for group messages.

For non-group messages (DMs) step 2 is skipped, so step 4 does run — but the routing chain almost always resolves a DM to the bound agent before reaching step 4.

**Options:**
1. Guard step 4 with `if !message.is_group { … }` to skip the redundant call for groups.
2. Remove step 4 entirely and document that groups rely on step 2; DMs rely on the binding chain which should always resolve before reaching the "first available" tail.
3. Leave as-is if this is intentional defence-in-depth (the extra call is cheap since the regex cache is warm).

Just flagging — the current behaviour is correct, this is purely an efficiency / clarity question. Happy to defer to your judgement on which option fits the intended design.

---

Issue #5672, opened 2026-05-23
bug(channels): channel-side /commands ignore account_id — multi-bot deployments route commands to wrong agent (and /agent leaks user-defaults across bots)


nevgenov:
# bug(channels): channel-side /commands ignore `account_id` — multi-bot deployments route commands to the wrong agent (and /agent leaks user-defaults across bots)

## Summary

In a multi-bot Telegram (or other multi-instance channel) deployment, every channel-side slash command (`/model`, `/new`, `/reboot`, `/compact`, `/stop`, `/usage`, `/think`, `/btw`) is dispatched to the **first-registered** channel agent regardless of which bot received the message. Separately, `/agent` writes a per-`peer_id` override (`user_defaults`) that leaks across every bot the same user has access to.

The regular non-command message dispatcher at `bridge.rs:2725` calls `resolve_with_context(...)` and routes correctly — the regression is isolated to the `/command` arms.

## Repro

Tested e2e on `v2026.5.17-beta.12` (commit `953fc93`) against three Telegram bots configured with distinct `default_agent` mappings:

```toml
[[channels.telegram]]
account_id = "bot-a"                    # first in config order
bot_token_env = "BOT_A_TOKEN"
default_agent = "agent-A"               # provider = anthropic, model = claude-sonnet-4-6

[[channels.telegram]]
account_id = "bot-b"
bot_token_env = "BOT_B_TOKEN"
default_agent = "agent-B"               # provider = deepseek,  model = deepseek-v4-flash

[[channels.telegram]]
account_id = "bot-c"
bot_token_env = "BOT_C_TOKEN"
default_agent = "agent-C"               # provider = deepseek,  model = deepseek-chat
```

All three agents exist, run, and are correctly bound in `daemon.log`:
```
INFO telegram default agent: agent-A (...) [channel: telegram:bot-a]
INFO telegram default agent: agent-B (...) [channel: telegram:bot-b]
INFO telegram default agent: agent-C (...) [channel: telegram:bot-c]
```

### Bug A — `/model` and other channel commands collapse to system default

After a **clean daemon restart** (in-memory `user_defaults` empty, verified via `kill -9` + launchd respawn):

| # | Bot user typed `/model` in | Reply | Expected per `default_agent` |
|---|---------------------------|-------|------------------------------|
| 1 | `bot-b` | `Current model: claude-sonnet-4-6 (provider: anthropic)` | `deepseek-v4-flash / deepseek` |
[...]

PR #4961, opened 2026-05-12
feat(types,kernel,api): per-agent channel allowlist

DaBlitzStein:
## Summary

Fixes #4926

Adds a `channels` field to `AgentManifest` that restricts which
configured channels an agent can use. Empty list means all channels
(backward-compatible default).

- `AgentManifest.channels: Vec<String>` with `serde(default)` + `vec_lenient`
- `AgentRegistry::update_channels()` for in-memory mutation
- `LibreFangKernel::set_agent_channels()` with rollback on DB persist failure
- `KernelApi` trait method + impl delegation
- `GET /api/agents/{id}/channels` returns assigned, available, and mode
- `PUT /api/agents/{id}/channels` updates the allowlist and persists to disk
- `configured_channel_names()` helper in channels routes
- OpenAPI spec registration
- Wizard default: `channels: vec![]`

## Test plan
- [x] `cargo check -p librefang-api --lib` — compiles clean
- [x] `cargo clippy -p librefang-api -p librefang-kernel -p librefang-types --lib -- -D warnings` — zero warnings
- [ ] `cargo test -p librefang-api` — integration tests pass
- [ ] GET returns `{"assigned":[],"available":["telegram",...],"mode":"all"}` for fresh agent
- [ ] PUT with `{"channels":["telegram"]}` persists and survives restart

PR #6131, opened 2026-06-16
feat(channels): deterministic inbound dispatch — channel-instance binding lookup (#5671 Model A)

houko:
## What

Implements the **Model A** dispatch rewrite from the #5671 RFC: channel inbound routing becomes a deterministic two-level table lookup, removing the production DM-rotation bug. This is PR-B/C/D of the RFC rollout; PR-A (the `SidecarChannelConfig.agent` / `available_agents` schema fields) already landed in #6105.

## The bug being fixed

`resolve_or_fallback` applied the same topology-blind priority chain to every inbound message. With no binding, no `"assistant"` agent, and an expired thread-ownership claim (5-minute TTL), agent selection fell to `list_agents().first()` — non-deterministic by `HashMap` iteration order. Two turns of the same DM, sent hours apart, resolved to **different agents** with **different sessions** (no shared history). The RFC has the production journal showing a Telegram DM rotating `researcher` → `legal-assistant` across two messages.

A DM has one participant talking to one agent — there is nothing for a resolver to decide. The fix makes selection a table lookup instead of a chain.

## Changes (three atomic commits)

- **`feat(memory)`** — migration v44 adds `channel_instance_defaults` (one row per `[[sidecar_channels]]` instance) and `conversation_bindings` (per-`(instance, conversation)` override). New `ChannelBindingStore` over both; `resolve(instance, conversation_id)` does the two-level lookup (override → instance default → `None`). Stores the agent **name**, not the per-spawn `AgentId` uuid, since config and the registry resolve by stable name. Wired into `MemorySubstrate`.
- **`feat(kernel)`** — at boot, seed `channel_instance_defaults` from each `[[sidecar_channels]] agent`. Re-seeds (upsert) every boot since config owns the default; per-conversation overrides are untouched.
- **`feat(channels)`** — `resolve_or_fallback` consults the lookup first, keyed by `instance = metadata["account_id"]` and `conversation_id = sender.platform_id` (uniform: chat id for groups, peer id for DMs). A configured binding wins outright. New `ChannelBridgeHandle::resolve_bound_agent` (default `None`, kernel adapter overrides with the DB lookup).

## Scope decisions

- **Non-breaking transition.** The legacy chain is kept as a fallback for instances with **no** binding configured, now emitting a deprecation `WARN` at the non-deterministic `list_agents().first()` tail. Configured instances become deterministic; unconfigured ones keep today's behaviour. This is the migration story raised in the RFC review rather than a hard delete.
- **Allowlist + router cache deliberately bypassed on the bound path.** Model A explicitly deletes `AgentManifest.channels` as a dispatch filter, and the deterministic DB lookup replaces the router's `set_user_default` cache (which only existed to paper over the non-deterministic chain). The operator's explicit `[[sidecar_channels]] agent` binding is authoritative.
- **Engagement untouched.** The HITL/AITL gate (`should_process_group_message`) and the send path are unchanged — engagement still runs after selection, so groups stay silent-by-default and DMs always engage.
- **Out of scope (later PRs):** the `/agent` slash command (the only writer of `conversation_bindings` — the table lands empty but read-active), the REST mirror, and the dashboard UI.

## Collision to resolve

`#4961` is a DRAFT adding `AgentManifest.channels` as a dispatch filter — that is Model B, which this PR's direction deprecates. It should be redirected before it lands, per the RFC thread.

## Verification

- `cargo test -p librefang-memory` — 7 new `ChannelBindingStore` tests (per-instance scoping / #5672 leak guard, override-wins-over-default, fall-through) + 31 migration tests pass.
- `cargo test -p librefang-channels` — `bound_agent_wins_over_legacy_fallback` and `no_binding_falls_through_to_legacy_chain` pass.
- `cargo check -p librefang-kernel -p librefang-api --lib` — clean.
- `cargo clippy -p librefang-memory -p librefang-channels -p librefang-kernel -p librefang-api --all-targets -- -D warnings` — zero warnings.
- `cargo fmt` applied.

Refs #5671. Supersedes the inbound-routing portion of #5323; eliminates the #5294 layers b/c non-determinism by construction.


houko (commented):
**`crates/librefang-channels/src/bridge.rs` — binding key for group messages (`resolve_or_fallback`)**

The dispatch comment and the PR body both state:

> `conversation_id` is the chat id for groups and the peer id for DMs — uniformly `message.sender.platform_id`

But in a Telegram group, `sender.platform_id` is the *sender's user ID*, not the *group chat ID*.
Using it as the binding key would mean two different users in the same group resolve as two different conversations, and a binding seeded for the group-chat ID would never match any inbound message.
For DMs the choice is correct (`sender.platform_id` == peer ID == the conversation).

Could you confirm one of the following:
1. The Telegram (and other) adapters intentionally populate `sender.platform_id` with the *chat ID* for group messages (i.e. not the sender's user ID), making the field uniform across DMs and groups?
2. Or the group case should use a different field — e.g. something from `message.metadata`, or a forthcoming `chat_id` field — and the current code is a latent bug that only goes unnoticed because configured instances with `[[sidecar_channels]] agent` set are always DM-only in practice?

The two new tests cover only `is_group: false`, so neither CI nor the test suite catches the group path today.
If groups simply can't reach the bound path (because the engagement gate keeps them silent before `resolve_or_fallback` is called), a short comment at that call site would make the intent explicit and prevent a future reader from "fixing" the field to the group chat ID inadvertently.

---

PR #6136, opened 2026-06-16
fix(plugin): respect `LIBREFANG_HOME` when resolving plugin directory

HuaGu-Dragon:
## Type

- [ ] Agent template (TOML)
- [ ] Skill (Python/JS/Prompt)
- [ ] Channel adapter
- [ ] LLM provider
- [ ] Built-in tool
- [x] Bug fix
- [ ] Feature (Rust)
- [ ] Documentation / Translation
- [ ] Refactor / Performance
- [ ] CI / Tooling
- [ ] Other

## Summary

respect `LIBREFANG_HOME` when resolving plugin directory

like

https://github.com/librefang/librefang/blob/dceae897a7a14d6b092999e07d25bd6aa9ab392a/crates/librefang-kernel/src/config.rs#L504-L512

## Testing

- [x] `cargo test --workspace` passes
- [ ] Live integration tested (if applicable)

## Security

- [x] No new unsafe code

### DIFF
--- a/crates/librefang-runtime/src/context_engine/scriptable/mod.rs
+++ b/crates/librefang-runtime/src/context_engine/scriptable/mod.rs
@@ -1765,10 +1765,7 @@ impl ScriptableContextEngine {
 
 /// Default plugin directory: `~/.librefang/plugins/`.
 pub fn plugins_dir() -> std::path::PathBuf {
-    dirs::home_dir()
-        .unwrap_or_else(|| std::path::PathBuf::from("."))
-        .join(".librefang")
-        .join("plugins")
+    crate::plugin_manager::librefang_home().join("plugins")
 }
 
 /// Load a plugin manifest from `~/.librefang/plugins/<name>/plugin.toml`.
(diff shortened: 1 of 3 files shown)


houko (approved):
LGTM. Correct fix: `plugins_dir()` previously ignored `LIBREFANG_HOME` and diverged from `config::librefang_home()`. Nice that it also fixes the second copy in `context_engine/scriptable/mod.rs` (which used a `.` fallback and ignored the env var too), consolidating both onto one helper.

Two non-blocking notes:

- The new test mutates the global `LIBREFANG_HOME`, which this crate otherwise avoids on purpose (see the comment in `artifact_store.rs::default_artifact_storage_dir_ends_with_data_artifacts`). The `ENV_LOCK` only serializes the two new tests against each other, not against other tests in the same process that read `LIBREFANG_HOME` live (`model_catalog::resolve_home_dir`, `artifact_store::default_artifact_storage_dir`). CI is fine since nextest isolates per process; a local `cargo test -p librefang-runtime` could flake. Acceptable here since the test genuinely needs to set the var to verify it's honored.
- Minor behavior change worth a line in the description: the no-`HOME` fallback shifts from `/tmp/librefang` to `/tmp/.librefang` and drops the `warn!`, aligning with the canonical `librefang_home()` in the kernel. Intentional, just undocumented.

PR #6141, opened 2026-06-17
fix(security): close channel media RBAC bypass and audit findings

houko:
## Summary

A multi-agent scan across the workspace (per-crate + cross-cutting security/correctness lenses) surfaced a set of defects; each was adversarially re-verified against source before being fixed. This PR lands the seven confirmed findings as one tight security/correctness cluster.

### 1. Channel media RBAC bypass (high) — `crates/librefang-channels/src/bridge.rs`
The multimodal (image/voice/video) dispatch path authorized `message.sender.platform_id`, which for **group** messages is the group/chat JID, not the individual sender. With RBAC enabled, an unauthorized group member could bypass the allowlist by sending media instead of text — exactly the voice-note vector flagged in the cross-chat-leak work (2d5a4458). The text path already keys on `sender_user_id(message)`; the media path now does too. Behavior-preserving for DMs (where `sender_user_id` falls back to `platform_id`).

### 2. Channel media skips policy/rate-limit gating (medium) — same file
The debounced media branch reached `dispatch_with_blocks` directly, skipping the `should_process_group_message` / `dm_policy` / per-channel + per-user rate-limit gating the text path enforces. Media could bypass a `mention_only`/`ignore` group or a DM `dm_policy=ignore`, and per-user rate limits never throttled media. A `media_dispatch_allowed` helper now runs the identical gating, placed at the single ungated entry point (`flush_debounced`) so the five already-gated `dispatch_with_blocks` call sites inside `dispatch_message` do not double-count rate-limit tokens.

### 3. Prompt-injection scanner zero-width/bidi bypass (medium) — `librefang-skills/src/verify.rs`, `librefang-runtime/src/prompt_builder.rs`, `librefang-kernel/src/kernel/prompt_context.rs`
The Aho-Corasick scan ran over raw text, so a zero-width or bidi code point spliced into a literal (`ignore<ZWSP>previous instructions`) evaded every Critical pattern, while invisible chars were only a non-blocking Warning and were never stripped before reaching the LLM prompt. Now: the scanner also scans an invisible-stripped variant (in-word insertion) and a separator-restored variant (zero-width-for-space), deduplicated across passes; `sanitize_for_prompt` drops invisible/format code points; and the previously-unsanitized `collect_prompt_context` content slot strips them too. The benign-i18n Warning behavior (U+200E/U+200F, bidi isolates in RTL descriptions) is left intact rather than hard-blocked.

### 4. ClawHub install path traversal (medium) — `crates/librefang-api/src/routes/skills/clawhub.rs`
`clawhub_install` and `clawhub_cn_install` joined `req.hand` into `workspaces/hands/<hand>` with no validation, unlike the sibling `install_skill`. A `../`-bearing id escaped the hands directory. Both now run `validate_skill_identifier(hand_id, "hand")` before the join.

### 5. Gemini streaming over-charges cached tokens (medium) — `crates/librefang-llm-drivers/src/drivers/gemini.rs`
The streaming usage path set `input`/`output` tokens but dropped `cache_read_input_tokens`, so cached Gemini prompt tokens were billed at the full input rate. It now captures `cached_content_token_count`, mirroring the two non-streaming paths.

### 6. MCP stdio server dies on one bad frame (low) — `crates/librefang-cli/src/mcp.rs`
A malformed JSON body collapsed to EOF and an oversized frame returned `Err`, killing the whole session on a single bad message. `read_message` now returns a `Frame` enum distinguishing true EOF from recoverable protocol errors; the run loop replies with JSON-RPC `-32700` and continues.

## Verification

- `cargo check` and `cargo clippy --all-targets -- -D warnings` are green across all seven touched crates (`librefang-channels`, `-llm-drivers`, `-skills`, `-runtime`, `-kernel`, `-api`, `-cli`).
- New regression tests added per fix:
  - channels: group-media authorizes the individual sender and the group/DM gate predicate holds.
  - skills: both bypass shapes (separator substitution and in-word insertion, incl. U+202E) now flag Critical.
  - prompt_builder: `sanitize_for_prompt` drops U+200B / U+202E.
  - gemini: streamed chunk with `cachedContentTokenCount` populates `cache_read_input_tokens`.
  - mcp: malformed-then-valid frame sequence recovers (no session kill); oversized body returns a recoverable error, not `Err`.
- Affected crates' unit suites pass (e.g. `librefang-skills` 211/211) and the gemini integration test passes.

## Notes

- Two scanner findings were rejected during adversarial verification and are intentionally not included.
- `reply_precheck` (an LLM-cost optimization gated on `group_policy=all`) is deliberately not replicated into the media gate — it is a silence optimization, not a security gate, and adding it would introduce a billed LLM call on the media path.

### DIFF
--- a/crates/librefang-api/src/routes/skills/clawhub.rs
+++ b/crates/librefang-api/src/routes/skills/clawhub.rs
@@ -310,6 +310,18 @@ pub async fn clawhub_install(
     Json(req): Json<crate::types::ClawHubInstallRequest>,
 ) -> impl IntoResponse {
     let home = state.kernel.home_dir();
+    // Reject path-traversal payloads in `hand` before it reaches any
+    // `Path::join` below — mirrors the guard `install_skill` applies to
+    // the same field. Without this, `{"hand":"../../something"}` escapes
+    // `~/.librefang/workspaces/hands/`. (audit: clawhub-install-path-traversal)
+    if let Some(ref hand_id) = req.hand {
+        if let Err(reason) = validate_skill_identifier(hand_id, "hand") {
+            return (
+                StatusCode::BAD_REQUEST,
+                Json(serde_json::json!({"error": reason})),
+            );
+        }
+    }
     let skills_dir = if let Some(ref hand_id) = req.hand {
         let hand_dir = home.join("workspaces").join("hands").join(hand_id);
         if !hand_dir.exists() {
@@ -689,6 +701,18 @@ pub async fn clawhub_cn_install(
     Json(req): Json<crate::types::ClawHubInstallRequest>,
 ) -> impl IntoResponse {
     let home = state.kernel.home_dir();
+    // Reject path-traversal payloads in `hand` before it reaches any
+    // `Path::join` below — mirrors the guard `install_skill` applies to
+    // the same field. Without this, `{"hand":"../../something"}` escapes
+    // `~/.librefang/workspaces/hands/`. (audit: clawhub-install-path-traversal)
+    if let Some(ref hand_id) = req.hand {
+        if let Err(reason) = validate_skill_identifier(hand_id, "hand") {
+            return (
+                StatusCode::BAD_REQUEST,
+                Json(serde_json::json!({"error": reason})),
+            );
+        }
+    }
     let skills_dir = if let Some(ref hand_id) = req.hand {
         let hand_dir = home.join("workspaces").join("hands").join(hand_id);
         if !hand_dir.exists() {
--- a/crates/librefang-channels/src/bridge.rs
+++ b/crates/librefang-channels/src/bridge.rs
@@ -1151,6 +1151,32 @@ fn flush_debounced(
                 None
             };
 
+            // --- Gate the debounced media path (parity with dispatch_message) ---
+            // The text branch below routes through `dispatch_message`, which
+            // enforces group_policy / dm_policy / rate-limits before
+            // dispatching. The media branch reaches `dispatch_with_blocks`
+            // directly, so without this it would skip those gates entirely —
+            // letting an attacker bypass a `mention_only`/`ignore` group (or a
+            // DM `dm_policy=ignore`) and per-user throttling by sending media
+            // instead of text. Run the identical gating here, on the only
+            // entry point that lacks it. (The five `dispatch_with_blocks`
+            // call sites inside `dispatch_message` itself are already gated,
+            // so the gate intentionally lives here, not inside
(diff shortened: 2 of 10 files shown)


houko (commented):
The security fixes are correct and the test coverage is solid — no logic issues found.

One project-wide style rule is violated throughout: **CLAUDE.md mandates "never write multi-line comment blocks — one short line max."** Every changed file in this PR introduces multi-line `//` or `///` blocks. Each should be collapsed to a single explanatory line (keeping only the WHY) or removed if the identifier already makes it obvious.

Specific locations:

**`crates/librefang-api/src/routes/skills/clawhub.rs`** — two identical 8-line `//` blocks (one in `clawhub_install`, one in `clawhub_cn_install`) should each become one line, e.g. `// Validate hand_id before Path::join to prevent path traversal, mirroring install_skill.`

**`crates/librefang-channels/src/bridge.rs`**
- The 14-line `// --- Gate the debounced media path ---` block before the `media_dispatch_allowed` call in `flush_debounced` → one line.
- The ~30-line `///` doc-comment on `media_dispatch_allowed` → one line explaining the purpose at the call site is enough; the function name itself is descriptive.

**`crates/librefang-cli/src/mcp.rs`**
- The 7-line doc-comment on `read_message` → one line.
- The 3-line `///` comment on `Frame::ProtocolError` → one line.
- The 2-line `// A single malformed or oversized frame…` comment in the `run_mcp_server` loop → one line.

**`crates/librefang-kernel/src/kernel/prompt_context.rs`** — the 17-line comment block before the invisible-chars `.filter()` → one line (the const name `INVISIBLE_PROMPT_CHARS` already communicates what it is; a single WHY sentence suffices).

**`crates/librefang-runtime/src/prompt_builder.rs`**
- The 4-line `///` comment before `INVISIBLE_PROMPT_CHARS` const → one line.
- The 3-line `// Drop invisible / format code points…` block inside `sanitize_for_prompt` → one line.

**`crates/librefang-llm-drivers/src/drivers/gemini.rs`** — the 2-line `// Cached prompt tokens (subset of input_tokens) for / cache-read pricing.` comment → one line.

**`crates/librefang-skills/src/verify.rs`** — several multi-line blocks explaining the three-variant scan strategy (the `// Invisible/format code points…` block, the `// Scan one text variant…` block) → collapse each to one line at the head of its section.

The prose-wrapping rule (one sentence per line, no column limit) applies only at sentence boundaries inside prose paragraphs — it does not override the "one short line max" rule for code comment blocks.

---

PR #6142, opened 2026-06-17
feat(dashboard/hands): GitHub/Codeberg registry source selector

### LINKED ISSUE #6137: 开发团队支持选择 GitHub  Codeberg
开发团队这里默认是 GitHub， UI 增加选择 GitHub 和 Codeberg 的功能

houko:
## Summary

Closes #6137.

The Hands ("开发团队") catalogue is pulled from a registry that defaulted to GitHub, with no way to switch to Codeberg from the dashboard — the backend `registry.registry_host` config (landed in #6095 / #6103 for the Codeberg-hosted registry) could only be changed by hand-editing `config.toml`. This surfaces it as a UI control, and hardens the backend host handling it relies on.

## Change

A compact source selector in the Hands page header, wired entirely through the existing config data-layer hooks:

- **Read** — `useFullConfig()` → `registry.registry_host`. Unset or an explicit `github.com` shows **GitHub** (the default); `codeberg.org` (any case / trailing slash) shows **Codeberg**; anything else shows **Custom (host)**.
- **Write** — `useSetConfigValue({ path: "registry.registry_host", value })`. GitHub clears it to `null`; Codeberg sets the Codeberg base URL. The mutation already invalidates the config cache, so the control reflects the persisted value.
- Disabled while the config is loading, refetching (post-write window — prevents the controlled `<select>` snapping back to the stale value), errored (prevents a stray click overwriting the real config with `null`), or a write is in flight.
- A failed live reload (`reload_error`) toasts an error, not a misleading success; `restart_required` toasts the restart variant.

## Backend root-cause fix

`registry_sync::registry_urls` treated any `Some(host)` as a Forgejo/Codeberg forge, so an explicit `registry_host = "https://github.com"` derived the Forgejo archive scheme (`/archive/{branch}.tar.gz`, prefix `{repo}/`) — which does not match GitHub's tarball layout (`/archive/refs/heads/{branch}.tar.gz`, prefix `{repo}-{branch}/`), so the sync silently failed. It now trims trailing slashes, drops an empty host, and folds an explicit `github.com` (any case) back to the GitHub branch. This is what lets the UI honestly treat `github.com` as GitHub instead of masking it as a broken value.

## Conventions

Follows the dashboard rules: API access only through `src/lib/queries` / `src/lib/mutations` hooks, theme tokens (no hex), `lucide-react` icon, accessible name via `<label htmlFor>`, TypeScript strict (no `any`), loading/empty/error states handled.

## Files

- `crates/librefang-runtime/src/registry_sync.rs` — github.com host folding + test.
- `crates/librefang-api/dashboard/src/pages/HandsPage.tsx` — `RegistrySourceSelector`.
- `crates/librefang-api/dashboard/src/pages/HandsPage.test.tsx` — config-hook mocks + selector tests.
- `crates/librefang-api/dashboard/src/locales/{en,zh,uk}.json` — `hands.registry_source*` strings.

## Verification

- Backend: `cargo test -p librefang-runtime registry_urls` (4/4, incl. the explicit-github-host case) and `cargo clippy -p librefang-runtime -D warnings` pass.
- Dashboard: `pnpm typecheck`, `eslint` (0 errors), `i18n-parity` (all locales), and `vitest run src/pages/HandsPage.test.tsx` (20/20, incl. default GitHub, switch-to-Codeberg writes the host, switch-back clears to null, explicit github.com reads as GitHub, disabled on error / during refetch, reload_error → error toast, restart_required toast) pass.

## Out of scope

The issue screenshot also shows the "开发团队" hand's own GitHub-specific settings (`gh` CLI, GitHub token). Making that hand operate against Codeberg is a change to the hand's `HAND.toml`, which lives in the external (Codeberg-hosted) registry, not this repo — so it is not addressable here.


houko:
Pushed one follow-up commit (`6052ec3`) with three mechanical fixes against the project comment/accessibility rules:

**1. Removed the 6-line JSDoc on `RegistrySourceSelector`** — CLAUDE.md forbids multi-line comment blocks (one short line max) and prohibits referencing issue numbers inside source comments (`#6095/#6103`, `#6137`) since those rot as the codebase evolves and belong in the PR description, which is exactly where they already appear.

**2. Collapsed the two-line inline comment** (`// A pre-existing custom forge URL is shown read-only...`) to one line — same rule.

**3. Removed the redundant `aria-label` on the `<select>`** — the element is already correctly labelled via `<label htmlFor="hands-registry-source">`, which is the preferred HTML pattern. When both are present, `aria-label` silently overrides the programmatic label association, making the `<label>` element invisible to assistive technology. The tests continue to pass unchanged because `@testing-library` resolves accessible names from the `<label>` association as well.

No logic or test changes — the three new selector tests still cover the same cases.

---

Issue #6179, opened 2026-06-17
sec(install): install.sh SHA256 verification has two silent skip paths — no GPG/cosign


BunnyMoth:
### Description

The install script (`web/public/install.sh`) attempts SHA256 verification after downloading the binary but has two silent failure paths that allow installation to proceed without any verification:

1. If `sha256sum` or `shasum` is absent from `PATH` (line 266) — prints a yellow warning and continues installing
2. If the `.sha256` file is missing from the GitHub release (line 247) — the curl fetch fails silently and installation proceeds unverified

No GPG or cosign signature verification exists anywhere in the script. The `scripts/workers/install-sh.ts:17` worker issues a 302 redirect directly to the GitHub asset URL with no intermediary signing step.

Identified via independent security audit of main branch, June 2026 (commit `6196f133`).

### Expected Behavior

SHA256 verification should be a hard requirement:
- If `sha256sum` is absent, the install should fail with a clear actionable message rather than proceeding unverified
- If the `.sha256` release asset cannot be fetched, the install should fail rather than proceeding unverified
- Ideally, cosign or GPG signature verification should be added as a non-optional step

### Steps to Reproduce

1. Review web/public/install.sh lines 247-266
2. Line 266: `echo "No sha256sum/shasum found, skipping checksum verification."` — proceeds to install
3. Line 247: if the curl fetch of `$CHECKSUM_URL` fails, verification block is skipped entirely
4. Run the install script on a minimal Alpine container where sha256sum is absent — binary installs without verification

### LibreFang Version

v2026.6.10-beta.17 (commit 6196f133)

### Operating System

Windows

houko:
Resolved by #6191 (merged), which enforces SHA256 verification in install.sh and removes the silent skip paths.

PR #6191, opened 2026-06-18
fix(sec): enforce SHA256 verification in install.sh, remove silent skip paths

BunnyMoth:
Two silent skip paths in the SHA256 verification block are now hard failures:
- Missing .sha256 file: prints the URL and exits 1
- Missing hash tool (sha256sum/shasum): prints "install coreutils" and exits 1

Checksum mismatch was already a hard failure and is unchanged.
Syntax verified with bash -n.


houko (commented):
The CHANGELOG entry is a single line with no description and no issue number, while every other entry in this section follows the convention `**title** (#NNNN) (@author). Sentence one. Sentence two.` The `(@mrchn)` attribution is present (good), but the entry needs at least: the issue number it fixes (if one exists), and 1–2 sentences explaining the security problem that was fixed (the old code silently skipped SHA256 verification when no hash tool was found; that was the actual vulnerability). Please expand the entry to match the project convention before merge.

---

houko (approved):
LGTM — correctly closes the integrity-check bypass.

The script previously proceeded silently in two cases: the `.sha256` file missing from the release, and no `sha256sum`/`shasum` in `PATH`. Both now `exit 1` with a clear message instead of installing an unverified binary, and the actual hash-comparison path is unchanged. This is the right security/convenience tradeoff for an installer and matches what #6179 asked for. CI green. Approving.

Issue #6204, opened 2026-06-18
Hands 设置bug


Jengro777:
### Description

https://github.com/user-attachments/assets/b728fdbb-3b7c-4b51-bfbe-ce8c6c8cf3b2

### Expected Behavior

--

### Steps to Reproduce

看视频

### LibreFang Version

v2026.6.17-beta.20 

### Operating System

Linux (x86_64)

PR #6213, opened 2026-06-18
fix(hands): merge Hand settings on save instead of replacing the whole config

houko:
## Summary

Addresses #6204 (Hands settings bug). The issue is a screen-recording with no text, so I traced the Hands settings surfaces for the strongest concrete, code-grounded bug.

## Root cause

The Hand settings editor builds its PUT payload only from `draft` — the per-session diff containing just the keys whose input fired `onChange` (`HandsPage.tsx`). `update_hand_settings` forwards that partial map straight to `HandRegistry::update_config`, which does a **wholesale replace** (`entry.config = config`, `registry.rs:1355`, pinned by the `update_config_replaces_and_refreshes` unit test). So for a Hand with more than one setting, saving one field overwrote the instance config with just that key and reverted every other previously-saved setting to its default. Invisible for single-setting Hands; data-loss for multi-setting Hands.

## Fix

- `crates/librefang-api/src/routes/skills/hands.rs` — `update_hand_settings` now reads the active instance's current config and **merges** the incoming keys over it, making the endpoint a true partial update. The registry's `update_config` replace contract is left unchanged (it has one other caller and a unit test pinning it).
- `crates/librefang-api/dashboard/src/pages/HandsPage.tsx` — the editor seeds its payload from the saved `current_values` before overlaying the draft, as defense in depth.

## Confidence / scope note

**Medium confidence** this is the bug the (text-less) video depicts — it's the strongest concrete settings-save defect in this area. If the video instead shows the agent-config tab (system prompt / tools), that path persists and refetches correctly; the only residual there is the state-resync `useEffect` potentially wiping an unsaved draft on a window-focus refetch — flagging for whoever has seen the video.

No backend integration test was added: the hands test harness installs no fixture Hand, so the settings happy-path is unreachable without new fixture infrastructure (out of scope for this fix). Recommend a follow-up that adds a fixture Hand with ≥2 settings and asserts PUT{a}→PUT{b}→GET keeps both.

## Verification

> ⚠️ Local verification was not possible (host disk full, OrbStack down → no `cargo`; disk too tight to risk `pnpm`). Both changes are static-reviewed: backend — `HandInstance.config` and the incoming body are both `HashMap<String, Value>`, so read→`extend`→`update_config(merged)` type-checks; frontend — `current_values?: Record<string, unknown>` spreads cleanly into the payload. **CI is the verifier** — `cargo check --workspace --lib`, `cargo test -p librefang-api`, `cargo test -p librefang-hands` (replace contract unchanged), and `pnpm typecheck`.


crates/librefang-api/src/routes/skills/hands.rs
houko, line 1: Per CLAUDE.md §"What you MUST do for any route / wiring change": every route change needs a `#[tokio::test]` against `TestServer` before merge.

The PR body explains the blocker honestly — no fixture Hand exists in the test harness, so the settings happy-path is unreachable. That's a real constraint, not an excuse to skip, so it needs to be resolved in this PR rather than deferred to a follow-up.

Minimal unblocking path (no new fixture infrastructure required): add a test that POSTs to `PUT /api/hands/{id}/settings` against `start_test_server()` with a non-existent `hand_id` and asserts the `404` + "Activate the hand first" error shape. That covers the route registration, auth middleware pass-through, and `None` branch without needing a live Hand. For the merge path, a second test can install a minimal stub Hand (or mock `HandRegistry`) and assert the `extend` semantics — but the `404` test alone satisfies the "route wired and gated" bar that reviewers check.

---

houko (commented):
Style fixes pushed directly (1 commit); one blocking item flagged inline.

---

houko (commented):
CLAUDE.md requires a `#[tokio::test]` against `TestServer` for any route behaviour change (`update_hand_settings` now merges instead of replacing). The PR body acknowledges the gap and proposes a follow-up, but CLAUDE.md's rule is: defer only if the fix would touch a *different crate or domain* — the test would live in `crates/librefang-api/tests/`, the same domain as this route. "New fixture infrastructure" for a `Hand` instance is worth adding here rather than deferring; the suggested test shape from the PR body is exactly right: install a fixture Hand with ≥ 2 settings, `PUT {a}`, `PUT {b}`, `GET`, assert both keys are present. Please add that before merge.

---

PR #6221, opened 2026-06-18
fix(runtime): allow close_range, getresuid/gid, and pipe splicing syscalls in seccomp sandbox

pavver:
## Summary

This PR resolves sandbox crashes (`SIGSYS — disallowed syscall, seccomp triggered`) occurring on rolling-release/newer Linux environments (such as Arch Linux) when running hook scripts under the default locked-down seccomp sandbox.

#### [MODIFY] [plugin_runtime.rs](file:///home/pavver/librefang-ukrainian-cli-i18n/crates/librefang-runtime/src/plugin_runtime.rs)
Added the following system calls to the seccomp allowlist (`apply_seccomp_allowlist`):
- `close_range` — used by Rust `std::process::Command` (and newer glibc) to close file descriptors during process startup.
- `getresuid` & `getresgid` — used by `bash` (which `/bin/sh` links to on Arch Linux) during startup for real/effective/saved ID checks.
- `vmsplice`, `splice` & `tee` — used by optimized coreutils (like `yes` in the `overflow_kills_child_without_waiting_for_timeout` test) for fast pipe data transfers.
- `statfs` & `fstatfs` — queried by glibc/bash to inspect filesystem metadata (such as read-only capabilities or mount types).
- `umask` — used to read or modify the file mode creation mask for processes executing scripts.

## Attribution

## Testing

1. **Clippy Validation**:
   ```bash
   CARGO_TARGET_DIR=/home/pavver/openfang/target-worktree cargo clippy --workspace --all-targets -- -D warnings
   ```
   *Result*: 0 warnings / errors.
2. **Cargo Tests**:
   ```bash
   CARGO_TARGET_DIR=/home/pavver/openfang/target-worktree cargo test -p librefang-runtime --lib -- locked_down_default_hook_completes_round_trip native_runtime_round_trip native_runtime_timeout_is_enforced overflow_kills_child_without_waiting_for_timeout test_scriptable_hook_accepts_full_runtime_path
   ```
   *Result*: `ok. 5 passed; 0 failed` (All 5 previously failing sandbox tests now pass successfully).


pavver:
# Detailed Diagnostics: Seccomp Sandbox Disallowed Syscalls on Modern Linux

This document explains the technical details of the seccomp sandbox failures observed on modern Linux distributions (such as Arch Linux running kernel `7.0.x` and glibc `2.38+`) and how this commit resolves them.

---

## 1. System Environment
- **Operating System:** Arch Linux (`x86_64`)
- **Kernel Version:** `7.0.11-arch1-1 #1 SMP PREEMPT_DYNAMIC`
- **Toolchain:** Rust (latest std library spawn implementation)
- **Interpreter Linkage:** `/bin/sh` symlinked to `bash` (standard on Arch Linux, unlike Ubuntu/Debian which uses `dash`).

---

## 2. Diagnosed Issues & Syscall Failures

During execution of `librefang-runtime` sandbox hook tests, several processes were killed with `SIGSYS — disallowed syscall, seccomp triggered (signal 31)`. Detailed analysis via process tracing revealed the following system call failures:

### A. The `close_range` System Call
* **Symptom:** Tests like `native_runtime_timeout_is_enforced` failed with `SIGSYS`.
* **Root Cause:** In modern Rust toolchains (and newer glibc), the standard library process-spawning machinery (`std::process::Command`) optimizes file descriptor cleanup in the child process. After `fork` but before `execve`, instead of iterating and calling `close()` on every file descriptor individually, it invokes the **`close_range`** system call (added in Linux 5.9).
* **Impact:** Since `close_range` was missing from the seccomp allowlist, the child process was killed by the kernel during spawn setup before `execve` could occur.

### B. The `getresuid` & `getresgid` System Calls
* **Symptom:** Plain script execution tests (`native_runtime_round_trip`, `locked_down_default_hook_completes_round_trip`) failed with `SIGSYS`.
* **Root Cause:** Unlike lightweight shells like `dash`, `bash` performs more thorough initialization checks. During startup on Arch Linux, `bash` issues `getresuid` and `getresgid` queries to retrieve real, effective, and saved user/group IDs.
* **Impact:** The omission of these system calls resulted in immediate termination of the shell process on startup.

### C. The `vmsplice` & Splicing System Calls (`splice`, `tee`)
* **Symptom:** `overflow_kills_child_without_waiting_for_timeout` timed out after 30 seconds instead of throwing `InvalidOutput`.
* **Root Cause:** This test pipes the output of coreutils `yes` to head. On modern Linux, GNU coreutils `yes` is optimized to use the zero-copy **`vmsplice`** system call when piping output.
* **Impact:** Because `vmsplice` was blocked, `yes` was terminated instantly by the kernel, yielding only a few bytes of output. As a result, the stdout byte cap (1 MiB) was never reached, causing the test runner to hang waiting for more stdout, eventually timing out.
* **Resolution:** Whitelisting `vmsplice` (along with safe splicing calls `splice` and `tee`) allows `yes` to perform its optimized zero-copy writes, triggering the cap check and terminating as intended.

### D. The `statfs`, `fstatfs`, & `umask` System Calls
* **Symptom:** Part of filesystem-related sandbox test checks and script preparation.
* **Root Cause:** When `bash` starts or when programs attempt to write/inspect directories (e.g., during file redirects or setup of temp directories), the shell and glibc library query filesystem statistics (`statfs` and `fstatfs`) to check the filesystem type and its capabilities (e.g., whether it is read-only, has specific features). Similarly, the `umask` system call is used to read or modify the file creation mask for new files created by the script.
* **Impact:** Disallowing these system calls causes errors or terminations during script execution or setup.

---

## 3. Whitelisted Syscalls Safety Audit
All added syscalls are safe to allow in an unprivileged agent sandbox:
- `close_range`: Only closes file descriptors owned by the process. Safe.
- `getresuid` / `getresgid`: Read-only queries for user and group IDs. Safe.
- `vmsplice` / `splice` / `tee`: Zero-copy pipe splicing operations. Safe (no filesystem or network breakout capabilities).
- `statfs` / `fstatfs`: Query file system metadata and properties. Safe (does not allow reading file contents or modifying data).
- `umask`: Queries or changes file mode creation mask for the current process. Safe.

houko (approved):
LGTM — reasonable seccomp-allowlist compatibility fix.

The additions are the syscalls modern glibc/musl reach for that the previous allowlist was killing: `close_range` (fd cleanup), `getresuid`/`getresgid` (credential reads), `statfs`/`fstatfs`/`umask`, and the `splice`/`vmsplice`/`tee` pipe family. These are fd-scoped or read-only and don't grant the sandbox capabilities beyond what read/write on already-held fds already allows; `statfs`/`umask`/`getres*` are informational.

The `splice`/`vmsplice`/`tee` trio is the broadest addition — fine to allow since it only moves data between fds the process already holds, but a one-line comment grouping/justifying those three would help future auditors of this allowlist (non-blocking).

Single-file change, CI green. Approving.

PR #6229, opened 2026-06-19
feat(runtime): add agent label to librefang_tool_call_total (#6226)

### LINKED ISSUE #6226: librefang_tool_call_total has no agent dimension — tool failures can't be attributed per-agent
## Summary

`librefang_tool_call_total` (added in #3495) is labeled only by `tool` and `outcome` — there is no agent dimension, so tool failures cannot be attributed to the agent that produced them.

## Current state

`record_tool_call_metric` emits the counter with exactly two attributes:

```rust
// crates/librefang-runtime/src/agent_loop/tool_call.rs:33
pub(super) fn record_tool_call_metric(tool_name: &str, is_error: bool) {
    let outcome = if is_error { "failure" } else { "success" };
    metrics::counter!(
        "librefang_tool_call_total",
        "tool" => sanitize_tool_label(tool_name),
        "outcome" => outcome,
    ).increment(1);
}
```

This is great for a global tool success/failure rate, but with N agents sharing the same tool (e.g. `shell_exec`), a single misbehaving agent is indistinguishable from healthy ones at the metric layer.

## Motivating scenario

A cron-driven automation agent whose `shell_exec` calls get rejected by the exec allowlist / metacharacter gate contributes to `tool="shell_exec", outcome="failure"` exactly like every other agent's `shell_exec` traffic. An operator watching the global counter sees "shell_exec failing" but has no way to build a per-agent alert ("agent X's tool-failure rate spiked") without dropping to raw session transcripts. The trace spans carry `agent.id`, but the metric does not, so dashboards/alerts can't slice by agent.

## Precedent

`librefang_cron_fires_total` already carries an `agent` label (`crates/librefang-kernel/src/cron.rs`), so per-agent metric cardinality is already considered acceptable in the codebase.

## Proposal

Add an `agent` attribute to `librefang_tool_call_total`. The agent id/name is in scope at the call sites in `tool_call.rs` (the loop has `session.agent_id`); thread it into `record_tool_call_metric`. Cardinality stays bounded (agents × tools), and both labels are already sanitized/capped.

## Acceptance

- `librefang_tool_call_total{agent, tool, outcome}`.
- Existing queries keep working (added label, no removal).
- A per-agent tool-failure-rate panel/alert becomes expressible.

houko:
## Summary

`librefang_tool_call_total` (added in #3495) was labeled only by `tool` and `outcome`, so tool failures could not be attributed per-agent.
This adds an `agent` dimension so the counter becomes `librefang_tool_call_total{agent, tool, outcome}`, mirroring the existing per-agent `librefang_cron_fires_total{agent}` precedent already accepted in the codebase.

## Changes

- `crates/librefang-runtime/src/agent_loop/tool_call.rs`
  - `record_tool_call_metric` now takes an `agent_id: &str` first argument and emits an `agent` label.
  - The agent id is sanitized and length-capped via the shared `sanitize_tool_label` helper — identical handling to the `tool` label — so a hallucinated or hostile caller id cannot blow up metric cardinality. The helper doc-comment is generalized to cover both dimensions.
  - All call sites updated to thread the agent id from `session.agent_id` (`ctx.agent_id_str`):
    - `execute_single_tool_call` (Ok + Err paths).
    - `execute_tool_group` Phase 1 serial pre-check (`Blocked` + `CircuitBreak`).
    - `execute_tool_group` Phase 2 concurrent collector (Ok + Err) — the agent id is captured once from the shared `&ctx` borrow so the concurrent loop needs no re-borrow.
- `crates/librefang-runtime/src/agent_loop/tests/utilities.rs`
  - Updated the two existing `record_tool_call_metric` tests for the new signature.
  - Added `test_record_tool_call_metric_carries_agent_label` (regression for #6226): asserts the counter emits `agent`/`tool`/`outcome` labels, and that an over-long id containing a control char is control-char-stripped and capped at 64 chars.
- `CHANGELOG.md`: entry under `## [Unreleased]` → `### Added`.

## Verification

Run via the repo's Docker dev image (`librefang-rust-dev:latest`), scoped to `librefang-runtime`, per CLAUDE.md (no native cargo on host).

- `cargo check -p librefang-runtime --lib` — Finished, no errors.
- `cargo clippy -p librefang-runtime --all-targets -- -D warnings` — Finished, zero warnings.
- `cargo test -p librefang-runtime --lib record_tool_call_metric` — 3 passed; 0 failed:
  - `test_record_tool_call_metric_success_outcome`
  - `test_record_tool_call_metric_failure_outcome`
  - `test_record_tool_call_metric_carries_agent_label`
- `cargo fmt -p librefang-runtime` applied.

Deferred work: none.

Closes #6226


houko:
Adversarial review follow-up: the `agent` label was emitted but the metric's `describe_counter!` help string in `crates/librefang-telemetry/src/metrics.rs` still read "labeled by tool name and outcome".
Operators reading /metrics wouldn't know the agent dimension exists, and it left the metric inconsistent with the cron description that cites it as the cardinality precedent.
Fixed the help string to "labeled by agent, tool name, and outcome (#3495, #6226)".
Verified in the dev container: `cargo check` + `cargo clippy --all-targets -- -D warnings` clean on librefang-runtime/librefang-telemetry, and the new `test_record_tool_call_metric_carries_agent_label` regression test plus the two existing outcome tests pass.

PR #6233, opened 2026-06-19
feat(runtime): tool-call telemetry fidelity — failure-type breakdown, per-tool latency histogram, span error status

### LINKED ISSUE #6228: Tool-call telemetry fidelity: failure-type breakdown, per-tool latency histogram, span error status
## Summary

Tool-call telemetry currently collapses every failure to a single `outcome="failure"` bool, exports no per-tool latency, and never marks the tool span as errored. Three related fidelity gaps, grouped because they all live in the same tool-dispatch path.

### 1. Failure-type breakdown

`record_tool_call_metric(tool_name, is_error)` (`crates/librefang-runtime/src/agent_loop/tool_call.rs:33`) reduces all failure modes to one bit. A model misusing a tool (call blocked by the exec allowlist / metacharacter gate), an approval denial, a tool timeout, and a tool that genuinely crashed are indistinguishable in the metric.

The distinction already exists in the type system — `ToolExecutionStatus` (`librefang-types`) carries `Denied` / `ModifyAndRetry` / etc., and `tool_call.rs` already branches on these in `append_tool_result_guidance_blocks`. Thread it into the metric.

Proposal: add a `failure_type` attribute (or equivalent), e.g. `{ hard_error, blocked, approval_denied, timeout, circuit_break }`, populated from the existing status. Lets operators alert on "tool crashed" while ignoring "model fat-fingered a blocked command."

### 2. Per-tool latency histogram

Tool execution duration is already captured in the decision trace (`execution_ms`) but never exported as a metric. There is no `p50/p95/p99` per tool.

Proposal: export `librefang_tool_execution_seconds{tool}` (histogram) from the same dispatch path that records the existing counter.

### 3. Span error status on `execute_single_tool_call`

The tool span (`execute_single_tool_call`, ~`tool_call.rs:458`, fields `tool.name` / `tool.id`) never records `is_error` and never sets the span status to error. Result: a trace error-filter (`hasError = true`) never matches a failed tool, and the service-level `errorRate` reads `0` even when tools are failing.

Proposal: record `tool.outcome` on the span and set span status to error on a hard tool failure, so trace-level error filtering works.

## Acceptance

- `librefang_tool_call_total` (or a sibling) distinguishes failure types.
- `librefang_tool_execution_seconds{tool}` histogram exists.
- `execute_single_tool_call` spans carry outcome and are marked errored on hard failure.

## Notes

Complements the per-agent `tool_call_total` label and the loop-exit-reason counter filed separately. Cardinality stays bounded (`failure_type` is a small fixed enum; `tool` is already sanitized/capped).

houko:
Closes #6228.

Three related fidelity gaps in the tool-dispatch path (`crates/librefang-runtime/src/agent_loop/tool_call.rs`) are closed.

### 1. Failure-type breakdown on `librefang_tool_call_total`
`record_tool_call_metric` previously collapsed every failure to a single `outcome=failure` bool, so a blocked call, an approval denial, a tool timeout, and a genuine crash were indistinguishable.
A new bounded `failure_type` label is derived from the existing `ToolExecutionStatus` (the same enum `append_tool_result_guidance_blocks` already branches on).
The variant set is a small fixed enum, so cardinality stays bounded; `tool` is already sanitized and capped.

### 2. Per-tool latency histogram
`librefang_tool_execution_seconds{tool}` (histogram) now exports the dispatch duration that was already measured for the decision trace (`execution_ms`), emitted in seconds (`ms / 1000.0`).
The duration is carried on `ExecutedToolCall.execution_ms` so every metric call site (serial wrapper, parallel group collector) can emit it without re-borrowing the trace.
Short-circuit paths that never ran the timed body (fork-allowlist rejection, incognito drop, hook block, loop-guard block, circuit break) pass `None` and skip the histogram, so no bogus `0s` sample skews the distribution.

### 3. Span error status on `execute_single_tool_call`
The span now declares `tool.outcome` and `otel.status_code` and records them after execution.
`tool.outcome` carries the bounded failure-type label on every call; `otel.status_code = "error"` (the field `tracing-opentelemetry` maps to the OTel span status) is set **only** on a genuine hard failure / circuit break / timeout — not on a model-fat-fingered blocked or denied call — so a `hasError=true` trace filter and service-level errorRate finally match a failed tool instead of reading 0.

## ToolExecutionStatus → failure_type mapping

The real variants (read from `librefang-types/src/tool.rs`: `Completed`, `Error`, `WaitingApproval`, `Denied`, `Expired`, `ModifyAndRetry`, `Skipped`) map as:

| condition | failure_type | span status=error? |
|---|---|---|
| `is_error == false` (any status, incl. `Completed` / `WaitingApproval`) | `none` | no |
| `Skipped` | `blocked` | no |
| `Denied` | `approval_denied` | no |
| `ModifyAndRetry` | `approval_denied` | no |
| `Expired` | `timeout` | yes (genuine execution failure) |
| `Error` | `hard_error` | yes |
| `is_error == true`, no status (circuit-break `Err` path) | `circuit_break` | yes |
| defensive: `is_error == true` with `Completed` / `WaitingApproval` | `hard_error` | yes |

`Denied` and `ModifyAndRetry` are folded into one `approval_denied` bucket because both are approval-policy soft outcomes (the model's call was rejected / asked to be revised), and both keep the span status OK so they don't inflate the service errorRate.

## Merge note re #6229

This PR and #6229 (`feat/6226-tool-call-agent-label`) both touch `record_tool_call_metric`'s signature.
#6229 adds an `agent` label (`record_tool_call_metric(tool_name, agent_id, is_error)`); this PR adds `status` and `execution_ms` (`record_tool_call_metric(tool_name, is_error, status, execution_ms)`).
Whichever merges second will hit a trivial conflict on the function signature and its call sites — the resolution is to keep BOTH additions: `record_tool_call_metric(tool_name, agent_id, is_error, status, execution_ms)`, with the `agent` label on the counter alongside `failure_type`, and the histogram unchanged.
This PR is based on `origin/main`, which does not yet carry #6229's change, so it implements its change cleanly on top of main and does not add the agent label.

## Verification

All run in the sanctioned Linux dev container (`librefang-rust-dev`), scoped to `librefang-runtime`:

- `cargo check -p librefang-runtime --lib` — pass
- `cargo clippy -p librefang-runtime --all-targets -- -D warnings` — pass (boxed the `Member::Blocked` variant to satisfy `large_enum_variant` after `ExecutedToolCall` grew a field)
- `cargo fmt -p librefang-runtime --check` — pass
- `cargo test -p librefang-runtime --lib -- span_outcome failure_type record_tool_call_metric` — 11 passed, 0 failed

New tests (all in `crates/librefang-runtime/src/agent_loop/tests/utilities.rs`):

- `test_failure_type_label_mapping` — every `ToolExecutionStatus` (+ no-status) maps to the documented label
- `test_record_tool_call_metric_emits_failure_type` — the counter carries the right `failure_type` for each failing status
- `test_record_tool_call_metric_success_failure_type_is_none` — success path carries `failure_type=none`
- `test_record_tool_call_metric_emits_latency_histogram` — `librefang_tool_execution_seconds{tool}` recorded; 1500 ms → 1.5 s
- `test_record_tool_call_metric_skips_histogram_without_duration` — no histogram when duration is absent
- `test_span_outcome_hard_error_sets_error_status` / `test_span_outcome_circuit_break_sets_error_status` — span carries `tool.outcome` and `otel.status_code=error`
- `test_span_outcome_blocked_does_not_set_error_status` — blocked/denied records outcome but NOT span error status
- `test_span_outcome_success_is_not_errored` — success records `tool.outcome=none`, no span error status

The two pre-existing metric tests (`test_record_tool_call_metric_failure_outcome`, `_success_outcome`) were updated to the new signature and still pass.

## Out-of-scope follow-ups

None.


houko:
Adversarial review pass — the telemetry logic is correct.
Verified: `failure_type_label` is exhaustive over all 7 `ToolExecutionStatus` variants; `WaitingApproval` is `is_error=false` so it short-circuits to `none` and is never mislabeled as a failure; the `Some(Completed|WaitingApproval) => hard_error` arm is only reachable on the defensive `is_error=true` path.
The histogram is correctly gated behind `if let Some(ms)` so short-circuit paths emit no bogus 0s sample, and all five `ExecutedToolCall` construction sites carry `execution_ms`; the timed path (incl. `Expired` timeout) flows through the real `trace_start.elapsed()` measurement.
The boxed `Member::Blocked` change updated both the construction (`Box::new`) and deref (`*executed`) sites with no `.unwrap()` introduced.
One real defect fixed (3a41779c4, on top of the upstream 3d44d375b inline-comment fix): the span-field comment, the `record_tool_span_outcome` doc comment, and the CHANGELOG bullet all described `otel.status_code=error` as set only on hard-error / circuit-break and omitted `timeout`, contradicting the code (which sets it for timeout too — a timed-out body is a genuine execution failure). Brought all three prose sites in line with the code and added `test_span_outcome_timeout_sets_error_status` to lock the timeout arm.
Verified in the dev container: `cargo check -p librefang-runtime --lib` clean, `cargo clippy -p librefang-runtime --all-targets -- -D warnings` clean, `cargo fmt --check` clean, targeted tests 12 passed / 0 failed.

PR #6239, opened 2026-06-19
feat(runtime): CoT-pruning + developer-loop aggregation (#6173)

### LINKED ISSUE #6173: [Proposal] Advanced Context Engineering: CoT-Pruning & Developer Loop Aggregation
### Description

Hello maintainers! 

We would like to propose a set of enhancements for LibreFang's context management system. As LLMs with larger context windows (such as 1M+ tokens) and extensive reasoning capabilities (e.g., DeepSeek-R1, o1/o3-mini) become standard, managing the context window dynamically becomes critical for both latency and cost. 

Currently, LibreFang uses `gateway_compression.rs`, `context_compressor.rs`, and `history_fold.rs`. While these are highly effective, they suffer from two issues in long-running developer chat sessions:
1. **CoT Bloat:** Stale reasoning steps from reasoning-centric models fill up the context with thousands of unnecessary `<think>` tokens.
2. **Micro-Iteration Noise:** Continuous developer loops (edit, run test, fail, edit, run test, pass) pollute the history with massive terminal outputs and intermediate errors that are no longer relevant to the current code state.

To address these, we propose implementing **CoT-Pruning** and **Developer Loop Aggregation** in the runtime.

---

### Problem
Reasoning models generate extensive internal reasoning traces before responding. While critical in the active turn, these past thoughts (e.g., the assistant's internal deliberations from 5 turns ago) waste significant context space.

### Proposed Solution
Automatically strip reasoning traces (both `<think>...</think>` tags in text and raw `reasoning_content` API metadata) from older assistant messages.

### Proposed Configuration
Add configuration settings under the `[compaction]` section of `KernelConfig` and `AgentManifest`:
```toml
[compaction]
# Strip reasoning content from assistant messages older than X turns. Set to 0 to disable.
strip_reasoning_after_turns = 3
```

### Technical Design (Rust Runtime)
- **Location:** Integrated into `crates/librefang-runtime/src/session_repair.rs` or `crates/librefang-runtime/src/agent_loop/prompt.rs` during the message preparation phase (`prepare_llm_messages`).
- **Algorithm:** 
  1. Traverse the message list backward.
  2. Count assistant turns.
  3. For assistant messages older than `strip_reasoning_after_turns`, prune the `<think>` block from the text content and clear the `reasoning_content` field from the provider metadata.
  4. Call `session.mark_messages_mutated()` and save the session.

---

### Problem
When the agent executes autonomous loops (e.g., repeating `file_write` and `run_tests` until a compilation error is fixed), it populates the history with redundant `ToolUse` and `ToolResult` pairs. This intermediate noise is irrelevant for future turns and degrades performance.

### Proposed Solution
Aggregate consecutive runs of developer tools into a single consolidated turn, stripping out intermediate terminal logs and intermediate file writes while preserving only the final success/failure outcome and final diffs.

### Proposed Configuration
```toml
[compaction]
# Aggregate consecutive autonomous code/test loops
aggregate_developer_loops = true
# Maximum consecutive steps before triggering aggregation
max_loop_steps_before_aggregate = 5
```

### Technical Design (Rust Runtime)
- **Location:** Implement as a new filter module or integrate into `session_repair.rs`.
- **Algorithm:**
  1. Identify contiguous runs of tool execution blocks containing developer actions (e.g., `run_tests`, `run_command`, `file_write`, `replace_file_content`, `grep_search`).
  2. If the run length exceeds threshold limits:
     - Retain the initial intent/call and the final outcome.
     - Replace intermediate runs with a synthesized user message indicating that compaction occurred:
       > `[DEVELOPER LOOP AGGREGATED] Agent executed 4 intermediate compilation and fix attempts on files X, Y. Intermediate compile errors were resolved. The final execution state is shown below.`
     - Prune intermediate failing logs.

---

## Questions for Maintainers & Design Decisions

Before starting work, we would love to hear your feedback on the overall design and discuss the following open questions:

1. **Configurable Developer Tool List:** What tools should be classified as "developer tools" for loop aggregation? Should the list be hardcoded in the runtime (e.g., `run_tests`, `file_write`, `run_command`, `replace_file_content`, `grep_search`), or should we expose it as a customizable config array (e.g., `developer_tools = ["run_tests", "file_write"]`)?
2. **CoT Deletion vs. Truncation:** For CoT-Pruning, should we delete the reasoning block entirely, or truncate it (e.g., preserving only the first 100 characters for debugging purposes)?
3. **Prompt Cache Impact:** Since modifying past messages dynamically invalidates the provider prompt cache (specifically for Anthropic's cache control), we should run pruning only at turn boundaries or pre-compaction. What is your preferred boundary for applying these prunes to minimize cache misses?

We look forward to hearing your thoughts!

### Alternatives Considered

_No response_

### Additional Context

_No response_

houko:
## Summary

Implements the two context-engineering features proposed in #6173 by @pavver: **CoT-pruning** and **developer-loop aggregation**.
Both are opt-in, default-off passes that run inside the existing compaction pass and reshape only the kept (verbatim) messages.

## Features

**CoT-pruning** — `[compaction] strip_reasoning_after_turns` (default `0`, disabled).
Strips `<think>...</think>` spans from text and drops `ContentBlock::Thinking` blocks from assistant turns older than the configured number of assistant turns.

**Developer-loop aggregation** — `[compaction] aggregate_developer_loops` (default `false`) + `max_loop_steps_before_aggregate` (default `5`).
Collapses a run of consecutive developer-tool steps into the first step, a deterministic placeholder, and the last step.

Both knobs are wired through `CompactionTomlConfig` (global `config.toml`), the per-agent `CompactionOverrides` (`agent.toml`, refs #5476), and the runtime `CompactionConfig`.

## How the design constraints are satisfied

- **Cache boundary.** Both passes run only in `compactor::compact_messages`, never in `prepare_llm_messages` / the per-turn hot path, so they do not invalidate the provider prompt cache more often than compaction already does.
- **`ReasoningEchoPolicy`.** CoT-pruning is a no-op for `Echo` models (DeepSeek V4 Flash, #4842) — those providers require the reasoning echoed back on `tool_calls` turns or the API returns 400. The policy is already a parameter of `compact_messages`.
- **Storage fidelity.** Only the compacted working session is reshaped; the append-only message journal retains the originals. Compaction already rewrites the working session, so this adds no new lossiness to the store of record.
- **Tool-pairing safety.** Loop aggregation only ever removes whole tool-use/result pairs, so no `tool_use_id` is orphaned. Non-loop or short runs are left verbatim.
- **Tool taxonomy, not a name list.** Developer tools are identified via the existing `classify_tool` → `ToolApprovalClass::{Mutating, ExecCapable}`, not a hardcoded list (answers the proposal's Q1).
- **Determinism (#3298).** The aggregation placeholder uses sorted/de-duped tool names and no timestamp.

## Changes

- `crates/librefang-types/src/config/types.rs` — three new `CompactionTomlConfig` fields + defaults.
- `crates/librefang-types/src/agent.rs` — matching `CompactionOverrides` fields + `resolve` / `is_empty`.
- `crates/librefang-runtime/src/compactor.rs` — `prune_stale_reasoning`, `aggregate_developer_loops`, helpers, integration into `compact_messages`, and unit tests.
- `crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json` — regenerated for the three new fields.
- `CHANGELOG.md` — `[Unreleased]` entries.

## Tests

Unit tests in `compactor.rs`: reasoning stripped past the threshold / kept within it, `Echo`-policy no-op, zero-disabled no-op, `<think>` strip (case-insensitive / multiple / unclosed), long-run aggregation with tool-pairing preserved, short-run and disabled no-ops, and read-only tools not aggregated.

## Verification

Local cargo/Docker verification was **not** run on the authoring host: it has no native Rust toolchain, and the sanctioned Docker dev image compiles into a volume that exhausted the constrained internal disk.
Per the repo's no-local-toolchain fallback, CI is the verification gate here — `cargo fmt`, clippy `-D warnings`, the `kernel_config_schema` golden test, and `librefang-runtime` / `librefang-types` unit tests all run there.
The golden fixture was hand-updated to match schemars output; if CI's golden test reports a diff, I'll reconcile it.

Closes #6173.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -841,6 +841,14 @@ _308 PRs from 7 contributors since v2026.5.17-beta.12._
   Reasons: `completed`, `max_iterations`, `repeated_tool_failures`, `circuit_break`, `content_filtered`, `error`.
   The counter increments exactly once per termination from a thin wrapper around the streaming and non-streaming loops, so no branch fall-through can double-count.
   Alert with `rate(librefang_agent_loop_exits_total{reason!="completed"}) > 0` per agent.
+- **feat(runtime): CoT-pruning for stale reasoning traces during compaction** (#6173) (@houko).
+  Strips `<think>...</think>` spans and `Thinking` content blocks from assistant turns older than `[compaction] strip_reasoning_after_turns` (default 0, disabled), reclaiming context that stale reasoning would otherwise occupy.
+  Runs only at the compaction boundary, never on the per-turn hot path, so it does not invalidate the provider prompt cache more often than compaction already does.
+  Skipped entirely for models whose `ReasoningEchoPolicy` is `Echo` (DeepSeek V4 Flash), which require the reasoning echoed back on tool_calls turns.
+  Implements part of the proposal in #6173 by @pavver.
+- **feat(runtime): developer-loop aggregation during compaction** (#6173) (@houko).
+  Collapses long runs of consecutive developer-tool steps — classified by the existing `ToolClass` taxonomy (`Mutating` / `ExecCapable`), not a hardcoded name list — into the first step, a deterministic placeholder, and the last step, opt-in via `[compaction] aggregate_developer_loops` (default off) with `max_loop_steps_before_aggregate` (default 5).
+  Whole tool-use/result pairs are removed together so no `tool_use_id` is orphaned.
 
 ### Breaking Changes
 
--- a/crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json
+++ b/crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json
@@ -895,6 +895,11 @@
     "CompactionTomlConfig": {
       "description": "Session compaction configuration (exposed in `[compaction]` TOML section).\n\nControls when and how the LLM-based history compaction runs. Internal algorithmic ratios (base_chunk_ratio, safety_margin, etc.) are kept as private constants inside the runtime compactor and are not exposed here.",
       "properties": {
+        "aggregate_developer_loops": {
+          "default": false,
+          "description": "Aggregate consecutive developer-tool loops (edit/run/test) during compaction (default: false).",
+          "type": "boolean"
+        },
         "keep_recent": {
           "default": 10,
           "description": "Number of recent messages to preserve verbatim (default: 10).",
@@ -909,6 +914,13 @@
           "minimum": 0.0,
           "type": "integer"
         },
+        "max_loop_steps_before_aggregate": {
+          "default": 5,
+          "description": "Minimum consecutive developer-tool steps before loop aggregation triggers (default: 5).",
+          "format": "uint32",
+          "minimum": 0.0,
+          "type": "integer"
+        },
         "max_retries": {
           "default": 3,
           "description": "Maximum retries for LLM summarization (default: 3).",
@@ -920,6 +932,13 @@
           "default": 1024,
           "description": "Maximum tokens for summary output (default: 1024).",
           "format": "uint",
+          "minimum": 0.0,
+          "type": "integer"
+        },
+        "strip_reasoning_after_turns": {
+          "default": 0,
+          "description": "Strip reasoning from assistant messages older than this many turns during compaction; 0 disables (default: 0).",
+          "format": "uint32",
           "minimum": 0.0,
           "type": "integer"
         },
@@ -5988,10 +6007,13 @@
         }
       ],
       "default": {
+        "aggregate_developer_loops": false,
         "keep_recent": 10,
         "max_chunk_chars": 80000,
+        "max_loop_steps_before_aggregate": 5,
         "max_retries": 3,
         "max_summary_tokens": 1024,
+        "strip_reasoning_after_turns": 0,
         "threshold_messages": 30,
         "token_threshold_ratio": 0.7
       },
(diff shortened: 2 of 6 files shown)


houko (commented):
**Consecutive `Role::User` messages in aggregated output** (`developer_loop_placeholder`, `compactor.rs`)

The placeholder inserted between two elided steps has `role: Role::User`.
In the aggregated sequence the surrounding layout is:

```
[tool_use  — Assistant]   ← first step
[tool_result — User]      ← first step result
[placeholder — User]      ← ⚠ consecutive User turn
[tool_use  — Assistant]   ← last step
[tool_result — User]
```

`tool_result` and `placeholder` are both `Role::User`, producing two consecutive user-role messages.
Most provider APIs (including Anthropic's) accept consecutive user turns in practice — they are coalesced before forwarding — but the Anthropic Messages API *rejects* two consecutive user messages when they both contain only `text` blocks if there is no preceding `assistant` turn (see the "alternating roles" validation in their API docs).
Since compaction is meant to produce a well-formed session, worth checking whether the output of `compact_messages` is ever fed raw into a provider call without the normal role-coalescing pass.

If the session already runs through a role-merge step before the final API call this is fine; if not, consider making the placeholder `Role::Assistant` (a brief summary note) or a `ContentBlock::ToolResult`-style synthetic user block that pairs with a synthetic tool-use so the sequence stays alternating.


---

PR #6255, opened 2026-06-20
chore(runtime): warn on cleanup failures in shell and web_fetch_to_file

maoxin1234:
## Summary

Replaces two silent `let _ =` ignores with `tracing::warn` logs so cleanup failures are observable:

1. **`tool_runner/shell.rs`**: when `shell_exec` is interrupted or times out, it now logs a warning if `kill_process_tree` fails. Previously the failure was swallowed, making it hard to notice orphaned processes.

2. **`web_fetch_to_file.rs`**: when the atomic write or rename fails, it now logs a warning if removing the temporary partial-download file fails. Previously the temp file leak was silent.

## Verification

- `cargo fmt --all`
- `cargo test -p librefang-runtime web_fetch_to_file` (4 passed)
- `cargo test -p librefang-runtime shell` (108 passed)


houko (approved):
LGTM. Reviewed: replaces three silent `let _ =` cleanup ignores with `tracing::warn` structured logs (orphaned process trees, leaked partial-download temp files become observable); control flow is preserved exactly and it follows the repo logging conventions. Real CI — Quality/clippy, all 4 Ubuntu test shards, Unit, Build, Security, Live Integration Smoke — is green.

The red `detect-secrets scan` check is a stale-base artifact: this branch predates #6262, which retired detect-secrets in favor of gitleaks, so the legacy baseline gate fails here. It is not a real secret (the change is logging-only). Merging onto current main — which uses gitleaks — is clean.

houko (approved):
LGTM — reviewed clean (structured-log cleanup, control flow preserved); real Rust CI green; branch updated to current main.

PR #6256, opened 2026-06-20
refactor(api): make WebhookStore async-safe with tokio RwLock (closes #2)

maoxin1234:
## Problem

`WebhookStore` guarded its in-memory state with `std::sync::RwLock` and exposed synchronous `list` / `get` / `create` / `update` / `delete` methods that are called directly from async axum handlers.
A blocking lock acquired on a runtime worker thread can stall the executor under contention, and the synchronous `std::fs` persistence on the write path compounds it by blocking the thread during disk I/O.

Fixes #2.

## Change

- `webhook_store.rs`: replace `std::sync::RwLock` with `tokio::sync::RwLock`; `list` / `get` / `create` / `update` / `delete` are now `async`.
- `persist` runs the atomic file write on `tokio::task::spawn_blocking` so disk I/O never blocks the async runtime.
- `load` stays synchronous on purpose — it is called once at daemon boot / test setup, off the async request path — with a doc-comment explaining why.
- `routes/webhooks.rs`: handlers `.await` the store calls. In `get` / `create` / `update` / `delete`, the `!Send` `ErrorTranslator` was being held across the new await point (breaking the axum `Handler` bound); resolve the translated messages up front and drop the translator before the await, matching the existing pattern in `test_webhook` (per CLAUDE.md).
- Webhook-store unit tests converted to `#[tokio::test]` and `.await` the async methods.
- `tests/api_deny_unknown_fields_test.rs`: `.await` the `list()` call.

## Verification

- `cargo fmt -p librefang-api`
- `cargo check -p librefang-api --all-targets` — clean
- `cargo test -p librefang-api --lib webhook_store` — 38 passed
- `cargo test -p librefang-api --test api_deny_unknown_fields_test` — 2 passed


houko (approved):
LGTM. Reviewed: the std→tokio `RwLock` migration is correct and complete — all `WebhookStore` methods are now `async`, every one of the 6 call sites was updated, the `!Send` `ErrorTranslator` is resolved-and-dropped before each `.await`, disk I/O moved to `spawn_blocking`, and there's no deadlock or lock-ordering hazard. Dropping the `unwrap_or_else(|e| e.into_inner())` poison recovery is correct since `tokio::sync::RwLock` doesn't poison. Real CI — Quality/clippy, all 4 Ubuntu test shards, Unit, Build, Security, Live Integration Smoke — is green.

The red `detect-secrets scan` check is a stale-base artifact: this branch predates #6262, which retired detect-secrets in favor of gitleaks, so the legacy baseline gate fails here. It is not a real secret. Clean to merge onto current (gitleaks) main.

houko (approved):
LGTM — reviewed clean (tokio RwLock migration, all call sites updated, no deadlock/!Send-across-await); real Rust CI green; branch updated to current main.

Issue #4614, opened 2026-05-04
feat: Persist canonical UUID per agent name to prevent silent history loss on respawn


nevgenov:
## Problem

Agent UUIDs can change for various reasons — internal errors, edge cases in the lifecycle management, or other circumstances that are not always obvious to the user. When this happens and the agent is re-spawned under a new UUID, all memories, sessions, and conversation history stored under the old UUID become **orphaned**: the data still exists in the database but is no longer reachable by the agent or the user.

This is **silent data loss**. The user has no indication that history was lost — the agent simply "forgets" everything, with no warning or recovery path.

## Proposed Solution

Introduce a **canonical UUID registry** that maps `agent_name → uuid`, persisted independently of agent lifecycle.

### Core behavior

1. **On first spawn** — register `agent_name → canonical_uuid` in the registry
2. **On respawn** — look up registry first; if a canonical UUID exists for this name → restore it instead of generating a new one
3. **On explicit delete** — show a confirmation prompt before removing the registry entry

### Confirmation on delete (CLI / API / Telegram)

```
⚠️  Deleting agent "nika" will permanently remove its canonical UUID
    and all associated memories and sessions.
    This action cannot be undone.

Confirm? [y/N]
```

Only on explicit confirmation → purge the entry from the registry.

### Optional: migration utility

```bash
librefang agent merge-history <name> --from <old-uuid>
```

Reassigns orphaned sessions/memories from a known old UUID to the current canonical one.

## Expected Behavior

| Action | Current | Expected |
|--------|---------|----------|
| Agent respawns (any reason) | New UUID generated, history lost | Same UUID restored from registry |
| Explicit `agent delete` + confirmation | Silently removes | Registry entry removed, clean slate |
| Explicit `agent delete` without confirmation | Silently removes | Aborted with data-loss warning |
| Explicit UUID reset | Not possible | `--reset-uuid` flag, with confirmation |

## Why not just rely on config files?

Config files can be regenerated or overwritten. The registry should be a **separate, append-only source of truth** — similar to how SSH `known_hosts` or a CA cert store works. Even if the agent config is reset, the identity mapping survives.

## Notes

The implementation could be as simple as a `agent_identities.toml` (or a dedicated DB table) with entries like:

```toml
[agents.nika]
canonical_uuid = "660bef7c-04d5-4480-8af2-0ce029981a14"
created_at = "2026-04-01T10:00:00Z"
```

[...]

PR #6258, opened 2026-06-20
fix(channels): rank explicit per-peer bindings above the sidecar instance default

### LINKED ISSUE #6257: bug(channels): per-room binding shadowed by sidecar default_agent (instance default outranks peer binding)
### Summary

When a sidecar channel is configured with a `default_agent` (e.g. a Matrix or Telegram adapter), that value is seeded as the channel's **instance default** (`channel_instance_defaults`). Any explicit per-conversation `[[bindings]]` whose `match_rule` pins a `peer_id` (a specific room / DM) is then **never honored for inbound traffic** — every inbound message routes to the instance `default_agent`, regardless of the room it arrived in.

The per-room binding is effectively dead: it matches the peer but is never consulted.

### Repro (minimal)

Config:

```toml
[[sidecar_channels]]
name = "matrix"
channel_type = "matrix"
default_agent = "agent-A"      # seeds the channel-wide instance default

[[bindings]]
agent = "agent-B"              # a different agent, bound to one specific room
match_rule = { channel = "matrix", peer_id = "!room:example.org" }
```

1. Send an inbound message into `!room:example.org`.
2. **Expected:** routed to `agent-B` (the explicit per-room binding).
3. **Actual:** routed to `agent-A` (the instance default).

### Root cause

Routing precedence in `resolve_or_fallback` (`crates/librefang-channels/src/bridge.rs`):

```
3. /agent conversation override
4. sticky holder (#5323)
5. instance default  (resolve_instance_default)   <-- returns agent-A here
6. [[bindings]] router chain (resolve_with_context) <-- never reached
```

The instance default (branch 5, `boot.rs` seeds it from `sidecar_channels.agent`) is consulted **before** the `[[bindings]]` chain (branch 6), which the code comments label a "legacy fallback". But a binding that pins a specific `peer_id` is *more* specific than a channel-wide default and should win.

There is also a latent secondary bug that surfaces once precedence is fixed: the router's name→id cache used for binding resolution is seeded at channel-bridge startup from `kernel.agent_registry().list_arcs()` only — i.e. **spawned** agents. An agent that is defined but spawns lazily (first cron fire), or one reconciled to Suspended after an unclean shutdown, is absent from that registry at bridge-startup, so even when its binding is reached, the name→id lookup misses and it falls through to the default. The cache should also be seeded from the canonical agent-identity registry (every defined agent), independent of spawn state.

### Fix

A PR follows from `neo-wanderer:fix/inbound-binding-cache-unspawned-agents`:

1. `AgentRouter::resolve_specific_binding` — evaluates only bindings that pin a `peer_id`; consulted in `resolve_or_fallback` between the sticky holder and the instance default. Channel-only bindings stay in the lower-precedence chain, preserving the existing instance-default-beats-channel-default ordering.
2. Seed the router name cache from the canonical agent-identity registry in addition to the spawned set, so defined-but-unspawned bound agents resolve.

Unit tests cover the precedence and the specific-binding selection; existing sticky/override precedence tests are unchanged.

neo-wanderer:
Fixes #6257.

## Problem

A sidecar channel configured with `default_agent` seeds a channel-wide **instance default**. An explicit per-conversation `[[bindings]]` pinned to a `peer_id` (a specific room / DM) was then never honored for inbound traffic — every inbound message routed to the instance default, regardless of the room. The per-room binding matched the peer but was never consulted.

Cause: in `resolve_or_fallback` (`crates/librefang-channels/src/bridge.rs`) the instance default (branch 5) is consulted before the `[[bindings]]` router chain (branch 6). A binding pinning a specific `peer_id` is more specific than a channel-wide default and must outrank it.

Secondary latent bug (surfaces once precedence is fixed): the router's name→id cache is seeded at channel-bridge startup from `agent_registry().list_arcs()` (spawned agents only). A defined-but-unspawned agent — one that spawns lazily on its first cron fire, or was reconciled to Suspended after an unclean shutdown — is absent, so even when its binding is reached the name lookup misses and routing falls to the default.

## Changes

- **`AgentRouter::resolve_specific_binding`** (`router.rs`): evaluates only bindings whose `match_rule` pins a `peer_id`. Consulted in `resolve_or_fallback` between the sticky-holder check and the instance default. Channel-only bindings stay in the lower-precedence `resolve_with_context` chain, under the instance default — preserving the existing #5671 ordering.
- **`binding_context`** (`bridge.rs`): extracted so the per-peer check and the legacy chain evaluate the identical context.
- **Router name-cache seeding** (`channel_bridge.rs`): seed from the canonical agent-identity registry (`agent_identities().list()` — every defined agent, spawn-independent, refs #4614) in addition to the live spawned set; the spawned set is applied last so the live id wins for hand-derived `<hand>:<role>` agents not in the identity registry. Extracted as `seed_router_agent_names`.
- Updated the `resolve_or_fallback` precedence doc-comment (now 7 tiers).

## Verification

- `cargo check -p librefang-channels -p librefang-api --lib` — clean.
- `cargo clippy -p librefang-channels -p librefang-api --lib --all-targets -- -D warnings` — clean.
- New unit tests, all green:
  - `router::tests::resolve_specific_binding_only_matches_peer_id_rules` — only peer-pinned rules resolve; channel-only excluded.
  - `bridge::tests::peer_binding_outranks_instance_default` — end-to-end precedence.
  - `channel_bridge::tests::seed_router_includes_defined_but_unspawned_agents` — defined-but-unspawned bound agent resolves instead of falling to default.
  - `channel_bridge::tests::seed_router_spawned_overlays_identity` — spawned id overrides canonical for a shared name.
  - Existing `sticky_holder_outranks_instance_default` / `conversation_override_outranks_sticky_holder` unchanged.

### DIFF
--- a/crates/librefang-api/src/channel_bridge.rs
+++ b/crates/librefang-api/src/channel_bridge.rs
@@ -2419,6 +2419,28 @@ fn apply_channel_proxy<A>(
 // EmailCredentials + resolve_email_credentials removed alongside the
 // in-process adapter. See SIDECAR_CATALOG in routes/channels.rs.
 
+/// Seed the router's name->id cache used for binding resolution.
+///
+/// `identities` should enumerate *every defined* agent (the canonical identity
+/// registry), so a binding to an agent that has not been spawned yet still
+/// resolves; `spawned` overlays the live runtime set (which additionally
+/// carries hand-derived `<hand>:<role>` agents that are absent from the
+/// identity registry). Later inserts win, so passing `spawned` last lets the
+/// live id override the canonical one in the (refs #4614: identical) event they
+/// ever diverge.
+fn seed_router_agent_names(
+    router: &AgentRouter,
+    identities: impl IntoIterator<Item = (String, AgentId)>,
+    spawned: impl IntoIterator<Item = (String, AgentId)>,
+) {
+    for (name, id) in identities {
+        router.register_agent(name, id);
+    }
+    for (name, id) in spawned {
+        router.register_agent(name, id);
+    }
+}
+
 /// Start the channel bridge for all configured channels based on kernel config.
 ///
 /// Returns `Some(BridgeManager)` if any channels were configured and started,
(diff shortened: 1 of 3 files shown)


houko (approved):
LGTM. Reviewed the diff: the fix inserts an explicit per-peer binding tier between the sticky-holder check and the channel-wide instance default, iterating `self.bindings` (stored sorted by specificity descending, so the resolution is deterministic), and correctly falls through to the instance default when the matched agent isn't channel-eligible rather than stranding the message. The `binding_context` extraction is a faithful refactor, and there's an end-to-end `#[tokio::test]` (`peer_binding_outranks_instance_default`) driving the real `resolve_or_fallback` plus unit tests. CI is fully green.

Issue #6117, opened 2026-06-15
bug(security/channels): channel_send can dispatch to a different recipient on the same channel (cross-chat message leak)


houko:
### Summary

An agent's `channel_send` tool can dispatch a message to a *different* recipient than the current turn's inbound peer on the **same** channel, with no validation — a cross-chat message leak.

Observed live (2026-05-19): a `claude-code` driver agent received a WhatsApp message from its owner and dispatched two audio voice notes to a *stranger's* WhatsApp chat (a JID from a prior turn's context). Both messages landed in the wrong chat. Likely trigger: model hallucination / sticky context emitting `channel_send` with the wrong `recipient`.

### Root cause

When a driver spawns a CLI subprocess and re-exposes LibreFang's tool surface through the `/mcp` HTTP bridge, the bridge has no notion of the current turn's peer: `network.rs::mcp_http` called `execute_tool` with `sender_id: None` and `channel: None` hardcoded.
So `tool_channel_send` has no peer scope to validate against, and a mis-targeted `recipient` on the same channel is dispatched unchecked.

### Impact

Security / privacy: an agent can be induced (by hallucination or prompt manipulation) to send content intended for one contact to a different contact on the same channel. This is a confidentiality leak, not just a UX bug.

### Repro (shape)

1. Agent on a WhatsApp channel via the `claude-code` driver (CLI subprocess + `/mcp` bridge).
2. Owner sends an inbound message; in the same or a later turn the model emits `channel_send` with a `recipient` JID belonging to a different chat.
3. The message is delivered to the wrong recipient — no guard rejects the cross-chat dispatch.

### Prior fix attempt

PR #5282 implemented an end-to-end fix: propagate the turn's peer scope (`sender_user_id` / `sender_channel`) through `CompletionRequest` → the `claude-code` driver's per-invocation `mcp_config.json` headers → `/mcp` → `execute_tool`, and add a guard in `tool_channel_send` that refuses a same-channel dispatch to a mismatched `recipient` (cross-channel dispatch stays allowed; legitimate cross-chat escalation routes through `notify_owner`).
That PR was closed unmerged during a stale-PR sweep, so the leak is most likely still present on `main`.
Re-verify the current behaviour before re-implementing, and confirm whether other drivers that bridge through `/mcp` share the missing-peer-scope gap.

### Suggested labels

bug, area/security, area/channels, area/runtime

houko:
[...]

PR #6261, opened 2026-06-21
fix(runtime): channel_send without recipient replies to the group, not the speaker

neo-wanderer:
## Problem

A `channel_send` call **without** an explicit `recipient` is meant to reply to the current conversation. In a group/room it instead targeted the **speaker** rather than the **room**.

`tool_channel_send` (`crates/librefang-runtime/src/tool_runner/channel.rs`) auto-filled the recipient from `sender_id` (the individual who spoke), while the #6117 cross-chat dispatch guard a few lines below computed its `expected_chat` from `sender_chat_id` first (`sender_chat_id.filter(non-empty).or(sender_id)`). The two fallbacks **disagreed**.

In a group the speaker's id is not a conversation, so the send routed to it as if it were one. Concretely on Matrix a no-recipient file send resolved to the speaker MXID `@user:hs` instead of the room `!room:hs`; the homeserver rejected the `PUT m.room.message` with:

```
403 M_FORBIDDEN: User @bot:hs not in room @user:hs
```

The sidecar `Send` command is fire-and-forget (no `Response.error` correlation — the known long-term gap noted in `sdk/python/librefang/sidecar/runtime.py`), so that 403 surfaced **only on the sidecar's stderr**. The tool returned `File '…' sent to … via <channel>` and the agent reported success while nothing was delivered.

The explicit-recipient path was unaffected — the guard correctly rejects a mismatched explicit recipient; only the implicit (auto-fill) path resolved a target the guard never validated.

## Fix

Resolve the send target and the guard's `expected_chat` through one shared helper, `resolve_send_target(explicit, sender_chat_id, sender_id)`:

- explicit recipient wins;
- else the **conversation id** (`sender_chat_id` — room / group / chat);
- else `sender_id` (DM, where the chat id coincides with the sender or none is stamped).

Because the auto-fill now returns exactly what the guard treats as canonical, a no-recipient send can no longer resolve a chat the guard would reject. DM replies are unchanged (they fall through to `sender_id`). Out-of-band callers (cron / trigger, no identity stamped) still surface the existing "Missing 'recipient'" error.

This is purely the recipient-resolution defect. The separate fire-and-forget "phantom success" gap (a failed sidecar send still reports success) needs the `Response.error` protocol change already tracked in `runtime.py` and is **not** in scope here.

## Changes

- `resolve_send_target` helper unifies the send fallback with the guard's `expected_chat`.
- Updated the stale comment that claimed the auto-filled recipient "targets the current chat by construction" (untrue for groups before this fix).

## Verification

- `cargo check -p librefang-runtime --lib` — clean.
- `cargo clippy -p librefang-runtime --lib` — no warnings.
- `cargo fmt -p librefang-runtime -- --check` — clean.
- `cargo test -p librefang-runtime --lib channel::tests` — 5/5 pass:
  - `auto_fill_replies_to_group_not_speaker` (the regression: room over speaker),
  - `auto_fill_falls_back_to_sender_for_dm`,
  - `explicit_recipient_wins`,
  - `auto_fill_matches_guard_expected_chat` (send target ≡ guard expectation),
  - `no_identity_resolves_none`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -842,6 +842,10 @@ _308 PRs from 7 contributors since v2026.5.17-beta.12._
 
 ### Fixed
 
+- **fix(runtime): `channel_send` without a `recipient` now replies to the group, not the speaker** (#6261) (@neo-wanderer).
+  In a group conversation a no-recipient `channel_send` auto-filled the recipient from `sender_id` (the individual who spoke) instead of `sender_chat_id` (the room / group).
+  The send then targeted the speaker's user id as if it were a conversation — e.g. a Matrix file send routed to `@user:hs` rather than the room `!room:hs`, which the homeserver rejected with `403 not in room` (visible only on the sidecar's stderr, so the tool still reported success).
+  The auto-fill now resolves the same canonical target the #6117 cross-chat guard validates against — the conversation id first, falling back to `sender_id` only for DMs — via a shared `resolve_send_target` helper, so the send target and the guard's `expected_chat` can no longer disagree.
 - **fix(channels): the streaming final answer now fires a push notification** (#6248) (@houko) — reported and diagnosed by @neo-wanderer.
   On streaming channels (Telegram, Matrix) the agent's final answer was delivered as an *edit* of the `"…"` streaming placeholder, and edits do not generate a client push notification — so the push fired on the empty placeholder and the actual answer landed silently (unless it overflowed the per-message length limit, whose tail chunk is a fresh, notifying message).
   Both adapters now finalize the answer as a *fresh* message — Telegram sends the answer and deletes the placeholder; Matrix sends a new `m.room.message` and redacts the placeholder — so the notification fires on the answer regardless of length.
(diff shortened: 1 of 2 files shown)


houko:
Rebased onto `main` to clear the `.secrets.baseline` conflict. #6262 replaced detect-secrets with gitleaks and deleted that baseline file, so the conflict resolves by dropping it — your code and CHANGELOG entry are otherwise untouched. `gitleaks scan` passes on the rebased tree. Please `git fetch` + reset your local branch since it was force-updated.

houko (approved):
LGTM (rebased onto current main to clear the `.secrets.baseline` conflict; reviewed independently). Correct fix: a no-recipient `channel_send` in a group was auto-filling the recipient from `sender_id` (the speaker) instead of `sender_chat_id` (the room), so sends routed to the speaker's user id and silently failed (e.g. Matrix 403 not-in-room). The PR unifies the auto-fill and the #6117 guard's `expected_chat` through one `resolve_send_target` helper so the two fallbacks can't disagree, with 5 new unit tests. CI is fully green.

PR #6287, opened 2026-06-23
feat: pluggable context-rewrite modules — per-agent engine + host-run request_llm_summary (closes #6264)

### LINKED ISSUE #6264: Proposal: Pluggable Context Rewrite Modules
### Description

Hi.

I would like to propose discussing a pluggable module system for agent context management.

Short version:

> A context module should own the context rewrite strategy.
> It should be able to modify context through deterministic rules.
> It should be able to ask the runtime to run an LLM summary.
> It should be able to combine both approaches.
> Existing hardcoded context passes could gradually become the built-in implementation of this module API.

## Why

Context handling is one of the most important parts of the runtime.

LibreFang already has several hardcoded mechanisms that are effectively context rewrite policies:

- tool result spill / truncation;
- per-turn tool-result budget enforcement;
- context guard compaction for older tool results;
- history fold for stale tool results;
- LLM-based context compression;
- developer-tool loop aggregation;
- reasoning / thinking pruning;
- image and media stripping;
- safe history trimming;
- tool_use / tool_result pairing protection.

These mechanisms are useful, but they are currently spread across the runtime and embedded directly into the agent-loop pipeline.

I think it would be useful to gradually move this into a dedicated context policy / context module layer.

## Motivation

Different agent types likely need different context strategies.

For example:

- a web-search agent needs to clean HTML/search noise, deduplicate sources, and preserve URLs, dates, titles, and source names;
- a coding agent needs to preserve commands, exit codes, test/compiler errors, file paths, recent changes, and unfinished task state;
- a workflow agent may need to pin final outputs and aggressively compress intermediate steps;
- a chat agent may be able to compress old small-talk turns more aggressively.

A universal compression/trimming strategy is unlikely to work equally well for all agent types.

## Proposed Shape

This should not be only an "LLM summarization plugin".

The broader shape is:

> The context module manages the context rewrite strategy.

It should be able to:

- edit incoming user/system messages before they are written into context;
- edit tool results before they are added to history;
- compress or rewrite old tool results without an LLM;
- aggregate sequences of developer-tool calls;
- remove or collapse noisy/repeated output;
- prune reasoning / thinking blocks;
- decide which messages should stay verbatim;
- decide when context should be compacted;
- receive an explicit runtime request to compact now;
- ask the runtime to perform LLM summarization;
- combine deterministic rewrite and LLM summarization.

## Two Compaction Directions

It is important to support both directions.

### 1. Module-driven compaction

The module can inspect context pressure and decide what to do:

- do nothing yet;
- run a cheap deterministic rewrite;
- request an LLM summary;
- run a mixed strategy.

Example:

```text
Context is above 50%.
First collapse stale shell/test outputs.
Then summarize the cleaned middle section.
Keep the last 8 messages verbatim.
```

### 2. Runtime-requested compaction

The runtime should also be able to ask the module to compact context.

Possible reasons:

- context window pressure;
- manual `/compact`;
- cron/session compaction;
- provider overflow recovery;
- before an expensive LLM call;
- background maintenance;
- model switch to a smaller context window.

So the module API should support not only `should_compact`, but also an explicit request:

```text
runtime -> module:
  "Please compact this context now.
   Reason: token_pressure/manual/cron/provider_overflow.
   Budget: X tokens.
   Context window: Y.
   Preserve safety invariants."
```

The module could respond with:

- `NoChange`;
- a deterministic `ContextPatch`;
- a full replacement `Vec<Message>`;
- `RequestLlmSummary`;
- a mixed plan.

## LLM Boundary

I do not think an external module should receive `Arc<dyn LlmDriver>` directly.

A serializable contract seems safer.

The module could return an LLM summary request:

```json
{
  "action": "request_llm_summary",
  "messages": [],
  "prompt": "Summarize for a coding agent. Preserve file paths, commands, errors, and current task state.",
  "max_tokens": 1200
}
```

The runtime remains the owner of:

- the LLM driver;
- credentials;
- provider routing;
- fallback providers;
- budget accounting;
- telemetry;
- rate limits;
- safety checks.

After the LLM call, the runtime could either apply the summary plan itself or return the summary to the module for a final rewrite.

## Possible API Sketch

Conceptually:

```text
ContextSnapshot -> ContextDecision
```

`ContextSnapshot` could contain:

- agent id / session id;
- agent metadata;
- messages;
- tools metadata;
- context window;
- estimated tokens;
- invocation reason;
- current phase.

Possible invocation reasons:

```text
incoming_message
tool_result
pre_llm_call
runtime_requested_compaction
manual_compact
cron_compaction
provider_overflow
after_turn
```

Possible decisions:

```text
NoChange
PatchMessages(Vec<ContextPatch>)
ReplaceMessages(Vec<Message>)
RequestLlmSummary(CompactionPlan)
MixedPlan(Vec<ContextAction>)
```

Possible actions:

```text
truncate_tool_result
spill_tool_result
fold_tool_results
aggregate_tool_loop
prune_reasoning
strip_media
summarize_messages
preserve_recent
pin_message
drop_message
```

The runtime should validate the result before sending it to a provider:

- tool_use / tool_result pairing must remain valid;
- history must start with a valid role for strict providers;
- pinned messages must be respected;
- session repair should still run as a final safety net;
- unsafe module output should fall back to built-in behavior.

## Debugging Benefit

This would also help with debugging context loss.

Right now, when an agent loses context, it is hard to know why:

- was it safe trim?
- history fold?
- context guard?
- artifact spill?
- LLM compression?
- developer-loop aggregation?
- memory recall?
- prompt assembly?

A module-based system could emit trace output like:

```text
tool_result abc shortened from 80k chars to 3k
messages 12..24 collapsed as stale shell loop
reasoning blocks removed from 6 old assistant turns
LLM summary requested for cleaned middle section
kept last 8 messages verbatim
```

That would make context behavior much easier to inspect, test, and tune.

## Hot Reload

Another benefit is development ergonomics.

If context modules can be hot-reloaded, it becomes much easier to debug and iterate on context strategies:

- change module rules;
- replay the same session;
- compare output messages;
- inspect trace;
- adjust thresholds;
- test agent-specific behavior without rebuilding the whole daemon.

This seems especially useful while current context-loss issues are being investigated.

## Migration Path

This does not need to be a big-bang rewrite.

A reasonable path could be:

1. Define a `ContextPolicy` / `ContextRewriteModule` contract.
2. Move existing hardcoded passes behind a built-in implementation.
3. Add trace output for every rewrite decision.
4. Add per-agent context policy selection.
5. Add deterministic module actions first.
6. Add `RequestLlmSummary` as a host-executed action.
7. Later add WASM/sidecar/plugin backends.

The first useful version could keep all current behavior, but route it through the new module interface.

## Question

Does this direction fit LibreFang's architecture?

If yes, I would like to discuss what the right boundary should be:

- extend the existing `ContextEngine`;
- add a new `ContextPolicy` layer above or beside it;
- make WASM/plugin modules a backend for that layer;
- or start by moving the current built-in passes behind a common interface.

### Alternatives Considered

_No response_

### Additional Context

_No response_

houko:
Closes #6264.

The proposal asked for pluggable context-rewrite modules. The trait + backends already shipped (`ContextEngine`, `ScriptableContextEngine`, `SidecarContextEngine`); the two genuine gaps are both implemented here.

## 1. Per-agent context-engine selection

The engine was chosen solely by the global `KernelConfig.context_engine`. An agent's manifest can now override it:

- New `context_engine: Option<ContextEngineTomlConfig>` on `AgentManifest`, alongside `proactive_memory` / `skill_workshop` / `compaction` (same serde shape, same `[agents.<name>.context_engine]` support in a `HAND.toml`) — the #5476 per-agent-override pattern.
- `context_engine_for_agent` resolves **agent override > kernel config > compiled default**; a malformed override falls back to the global engine. Per-agent engines are built lazily and deduplicated by their canonical-JSON config key (collision-free), and the plugin-allowlist gate checks the resolved engine.
- Registered in `PER_AGENT_OVERRIDE_KEYS` (a misplaced `[agents.x.context_engine]` in `config.toml` gets the #5476 warning) and covered by the exhaustive override-key drift-guard test.

## 2. `request_llm_summary` — host-run summarization for scriptable compact hooks

Previously a scriptable `compact` hook had to return the final compacted messages itself, with no LLM access (the host owns the driver). Now the hook may return:

```json
{ "request_llm_summary": { "prompt": "…", "summarize": [..], "keep": [..], "max_tokens": 2048 } }
```

and the **host** runs the LLM summary with its own driver, returning a `CompactionResult { summary: <LLM text>, kept_messages: <keep> }`. The script chooses *what* to fold and *how* (prompt, budget); the driver, credentials, and routing stay host-side. Handled at all three compact output sites (cache hit / miss / non-cached) via a shared `try_host_summary` helper. Missing/empty `summarize` or any LLM error falls back to the built-in compactor — never a panic. The legacy `messages` return path is unchanged.

## Verification

- `cargo check -p librefang-types -p librefang-kernel -p librefang-runtime` — clean (Docker dev image).
- `cargo clippy -p librefang-types -p librefang-kernel -p librefang-runtime --all-targets -- -D warnings` — zero warnings.
- Tests (all pass): per-agent resolution order + dedup (3 kernel tests, drift guards); `request_llm_summary` host-run + keep-list, no-directive passthrough, empty-summarize fallback (3 runtime tests).

## Reviewer note

Per-agent override engines are built once and held for the process lifetime (`Box::leak`, deduped by canonical-JSON config key) to keep the existing `Option<&dyn ContextEngine>` return type without a wider refactor. Alternative: `Arc<dyn ContextEngine>` + signature change — flagging the trade-off.


CHANGELOG.md
```diff
@@ -273,6 +273,8 @@ In-crate only; no cross-crate error-shape changes.
 
 ### Added
```
houko, line 275: CLAUDE.md prose rule: "CHANGELOG bullets — one sentence per line within a bullet."
The continuation line (275) packs four sentences together.
Split to one sentence per line:
```
  The context engine was chosen by a single global `config.toml: context_engine`.
  An agent's manifest (`agent.toml`, or a `HAND.toml [agents.<name>.context_engine]` block) can now override it, resolved as agent override > kernel config > compiled default — mirroring the existing per-agent `proactive_memory` / `skill_workshop` / `compaction` overrides (#5476).
  Per-agent engines are built lazily and deduplicated by config, and a malformed override falls back to the global engine rather than disabling recall/compaction.
  (The proposal's separate `RequestLlmSummary` hook-protocol change is deferred pending design sign-off.)
```

---

crates/librefang-types/src/agent.rs
```diff
@@ -1351,6 +1351,27 @@ pub struct AgentManifest {
...
+    ///    `[agents.<name>.context_engine]`) — when `Some(_)`
+    /// 2. `config.toml` `[context_engine]` — kernel-global default
+    /// 3. compiled-in `ContextEngineTomlConfig::default()` ("default" engine)
+    ///
+    /// Use case: one agent runs a `sidecar` recall engine while the rest
+    /// of the daemon stays on the default in-process engine, without
+    /// forcing every agent onto the heavier path. Like
+    /// `tool_exec_backend`, a manifest change is picked up on respawn —
+    /// the engine is built lazily per resolved config and cached for the
+    /// process lifetime (#6264).
```
houko, line 1372: CLAUDE.md: "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."
This 19-line block covers the resolution order, a use-case example, and the pick-up-on-respawn behaviour — all of which are either obvious from the code or better placed in CLAUDE.md/docs.
Condense to one line keeping only the non-obvious invariant:
```rust
/// Per-agent context-engine override (`agent.toml [context_engine]`); `None` inherits the kernel-global engine; resolution: agent manifest > kernel config > compiled default (#6264).
```

---

crates/librefang-kernel/src/kernel/mod.rs
```diff
@@ -734,6 +734,22 @@ pub struct LibreFangKernel {
...
+    /// Lazily-built per-agent context engines (#6264). Keyed by a stable
+    /// fingerprint of the agent manifest's `[context_engine]` override so
+    /// agents that select the same engine config share one instance. Each
+    /// distinct override is built at most once per process and stored as a
+    /// `'static` reference (the engine lives for the daemon's lifetime, like
+    /// the kernel-global `context_engine` above), so
+    /// [`Self::context_engine_for_agent`] can hand back a borrow that
+    /// outlives the lock and survives across the agent loop's `.await`.
+    /// Agents without an override are served the kernel-global engine and
+    /// never touch this map.
```
houko, line 746: Same rule — 10-line doc block on a private field.
The non-obvious part is the `Box::leak` lifetime contract; the rest is recoverable from the type and neighbouring comment on `context_engine`.
Condense to:
```rust
/// Per-agent engine cache; keyed by config fingerprint, `Box::leak`-ed to `'static` (bounded: one per distinct config per process).
```

---

crates/librefang-kernel/src/kernel/tools_and_skills.rs
```diff
@@ -1383,25 +1383,55 @@ impl LibreFangKernel {
...
+    /// the resolved override config so agents selecting the same engine share
+    /// one instance, and cached as a `'static` reference for the daemon's
+    /// lifetime (see [`Self::context_engine_overrides`]) — the kernel-global
+    /// engine has the same process lifetime, so handing back a borrow here is
+    /// sound across the agent loop's `.await`.
+    ///
+    /// Regardless of which engine is selected, the `allowed_plugins` allowlist
+    /// gate still applies: if the agent restricts plugins and the *resolved*
+    /// engine config names a plugin not in the allowlist, the agent gets no
+    /// context engine (`None`), exactly as before.
```
houko, line 1408: Three new functions in this file each carry multi-paragraph docs that violate the one-line-max rule.

**`context_engine_for_agent`** (this hunk, lines 1386–1408): condense to:
```rust
/// Returns the context engine for `manifest`: agent override > kernel config > compiled default; falls back to global if per-agent build fails; still gates on `allowed_plugins`.
```

**`per_agent_context_engine`** (lines 1449–1456): condense to:
```rust
/// Builds or fetches a per-agent engine by config fingerprint; `Box::leak`-ed to `'static`, bounded: at most one per distinct override config per process.
```
The two multi-line inline comments inside the body (lines 1470–1471 and 1489–1491) should each become one line.

**`context_engine_config_fingerprint`** (lines 1678–1685): condense to:
```rust
/// Within-process fingerprint of an override config (serde_json → DefaultHasher); cache key for `context_engine_overrides`; `None` on serialization failure.
```

---

crates/librefang-kernel/src/kernel/tests.rs
```diff
@@ -11218,6 +11218,182 @@ fn approve_create_leaves_all_skills_agent_unpinned() {
     kernel.shutdown();
 }
 
+// ----- #6264: per-agent context-engine selection -----
+
+/// Identity of the engine `context_engine_for_agent` resolves to, by data
+/// pointer. The kernel-global engine is one fixed allocation; a per-agent
+/// override builds (and leaks) a distinct one and dedups by config. Comparing
+/// the trait object's data address proves *which* engine an agent got without
+/// needing an LLM driver to observe behavioral differences.
```
houko, line 11227: Four doc blocks in the new test section each exceed the one-line limit.

**`engine_addr`** (this hunk): `/// Returns the data-pointer of the resolved engine; pointer identity proves which engine the agent got (leaked allocations are stable).`

**`context_engine_resolution_prefers_agent_override_then_kernel_config`** (line 11248–11254): `/// Proves resolution order: agent override > kernel config > compiled default; identical configs share one engine instance.`

**`context_engine_resolution_compiled_default_when_no_override_set`** (lines 11317–11322): `/// Proves that absent any non-default override, every agent resolves to the compiled-default kernel-global engine.`

**`context_engine_override_parses_from_agent_toml`** (lines 11359–11361): `/// Proves [context_engine] in agent.toml parses onto the manifest; absent section deserializes to None.`

The two-line inline comment at line 11275–11276 inside the first test body should also collapse to one line.

---

houko (commented):
Style-only review — 5 CLAUDE.md violations in the new prose, all the same class.

**Rule 1 — CHANGELOG sentence-per-line** (`CHANGELOG.md:275`): the continuation line packs four sentences; see inline comment for the split.

**Rule 2 — one-line-max for doc comments** (CLAUDE.md: "Never write multi-paragraph docstrings or multi-line comment blocks — one short line max"): every new `///` block in this PR exceeds the limit. Inline comments on each file carry condensed one-line replacements.

Affected locations:
- `crates/librefang-types/src/agent.rs` lines 1354–1372 (19-line field doc)
- `crates/librefang-kernel/src/kernel/mod.rs` lines 737–746 (10-line field doc)
- `crates/librefang-kernel/src/kernel/tools_and_skills.rs` lines 1386–1408, 1449–1456, 1470–1471, 1489–1491, 1678–1685 (three function docs + two inline comment blocks)
- `crates/librefang-kernel/src/kernel/tests.rs` lines 11223–11227, 11248–11254, 11275–11276, 11317–11322, 11359–11361 (four function docs + one inline comment)

No logic issues found; the resolution order, dedup, fallback-on-build-failure, and `allowed_plugins` gate all look correct.

---

Issue #6292, opened 2026-06-23
Workflow runs: show parameters, step errors, and allow re-run with same params


DaBlitzStein:
## Problem

When running a workflow from the dashboard UI:

1. **Parameters not visible in run history**: after launching a workflow, the runs list shows state and steps count but NOT what parameters were used. You can't see what sector, stage, etc. were passed without opening the run detail.

2. **No re-run with same parameters**: there's no way to re-launch a workflow with the same parameters as a previous run. You have to type them all again.

3. **No error shown in run list**: failed runs show state=failed but the error message is hidden in the run detail. The run list should show WHY it failed.

4. **No step-level errors in API**: the StepResult struct has no error field - only output. When a step fails, the error goes to the run-level error, not to the individual step that failed.

### API changes
- GET /api/workflows/{id}/runs: add input (the params used) and error to each run item
- StepResult struct: add error field, expose it in GET /api/workflows/runs/{run_id}

### Dashboard changes
- Add parameter preview column to the runs table (key: value per param)
- Add error column to the runs table (truncated error on failed runs)
- Add "Re-run with same params" button that pre-fills the run form from a previous run

## Related
- Observed during investor-outreach workflow testing on rodela
- The run detail endpoint already returns input and step_results with per-step output
- These are just not surfaced in the list UI

houko:
Backend for all three asks is now on `main`:

- **Step errors** — #6302 adds an `error: Option<String>` field to `StepResult`, surfaced in `GET /api/workflows/runs/{run_id}` and the synchronous `POST /api/workflows/{id}/run` response. Operator-node steps that fail without an LLM call (gate / transform / branch) now carry the cause in the dedicated `error` field instead of only stuffing it into `output`.
- **Re-run with same params** — #6302 adds `POST /api/workflows/runs/{run_id}/rerun`, which starts a fresh run of the same workflow with the original run's stored `input`, leaves the original untouched, and returns `202` with the new run id.
- **Show parameters** — #6293 exposes each run's `input` (and `error`) in `GET /api/workflows/{id}/runs`; that endpoint is now also scoped to the path workflow (#6298), so the run inputs are no longer leaked across workflows.

Remaining: the dashboard columns + re-run button (the UI half), tracked separately per the #6302 changelog. Leaving this open for that UI work.

PR #6280, opened 2026-06-22
chore(deps): bump vulnerable/yanked lockfile deps to clear global CI failures

houko:
## Problem

CI is red on **every branch** since ~23:00 UTC 2026-06-22 — this PR, the open dependabot PRs (cron #6278, wasmtime, cargo-minor-patch), and main's next scheduled run. It was green on main at the beta.22 push (10:57). Two independent upstream events, neither caused by any PR's content, wedged two separate required checks:

### 1. `cargo-deny (advisories)` — yanked crate
`crypto-bigint` **0.7.0 through 0.7.4 were all yanked** from crates.io. The lock pinned the yanked **0.7.3**, and `deny.toml` sets `yanked = "deny"`:
```
error[yanked]: detected yanked crate (try `cargo update -p crypto-bigint`)
```

### 2. `Security` (`cargo xtask deps --audit`) — new high-severity advisory
**RUSTSEC-2026-0185** (published 2026-06-22, **7.5 high**) against **quinn-proto < 0.11.15** — "Remote memory exhaustion from unbounded out-of-order stream reassembly". The lock pinned the vulnerable **0.11.14**:
```
error: 1 vulnerability found!
```

## Fix

Lock-only bumps to the fixed releases, plus the minimal compatible cascade each requires (all resolver-chosen, within existing semver ranges):

- `crypto-bigint` 0.7.3 → 0.7.5 — pulls `hybrid-array` 0.4.10 → 0.4.12 and `typenum` 1.19.0 → 1.20.1
- `quinn-proto` 0.11.14 → 0.11.15

No source, manifest, or feature changes. Merging this unblocks both `cargo-deny` and `Security` across all open branches once they rebase.

## Verification

- `git diff origin/main --stat`: only `Cargo.lock`, 8 lines each side (the four crate version+checksum pairs above).
- Resolver-chosen patch/minor bumps; full build/test covered by CI.

Issue #6266, opened 2026-06-21
Session mixing between LibreFang web UI and Telegram channel


DaBlitzStein:
When using both the LibreFang dashboard web UI and the Telegram channel to interact with the same agent, sessions get mixed — messages from one interface appear in the other, or the agent responds in the wrong channel. Agent sessions should be isolated per channel instance to prevent cross-contamination of conversation contexts.

DaBlitzStein:
## Findings 2026-06-21

### Async delegation callback delivery
- The mid-turn forward (send_channel_message in complete_async_task) DOES fire — confirmed via logs
- Response IS sent to Telegram adapter with correct user=34387719
- But the response is NOT delivered to the user on Telegram
- **Root cause**: Zombie Telegram sidecar processes survive daemon restart (systemd --user restart doesn't kill child processes). Old sidecar holds the long-poll connection, new sidecar can't receive or send messages.
- Workaround: kill -9 all sidecar processes + daemon restart

### Session mixing (Web UI ↔ Telegram)
- deannatroi responds in the web UI chat instead of forwarding to Telegram
- Agent sees the callback in session but doesn't push to the channel
- Related: session context from web UI is shared with Telegram session

### Long tasks
- Multi-step delegated tasks work (Coder: 5+ iterations with web_search + web_fetch)
- Deepseek API latency per iteration: ~20-30s
- No intrinsic timeout issue — just slow

houko:
Triage after #6267 (now on `main`):

**Fixed by #6267:**
- *Zombie Telegram sidecars surviving daemon restart* — `librefang-channels/src/sidecar.rs` previously just `break`-ed out of the supervisor loop on `Shutdown`/`ReceiverGone`, leaking the Python sidecar that held the long-poll connection. It now `kill().await`s the child before breaking in both shutdown arms. This was the actual cause of "response IS sent to the Telegram adapter but NOT delivered to the user."
- *Delegation result not pushed to the channel mid-turn* — `librefang-kernel/src/kernel/task_registry.rs` now forwards the completion text to the agent's home channel after injecting the mid-turn signal, **when the originating turn carries a channel `chat_id`**.

**Still open:**
[...]

PR #6253, opened 2026-06-20
feat(cli): localize TUI Onboarding Wizard and Agents screen

pavver:
## Type

- [ ] Agent template (TOML)
- [ ] Skill (Python/JS/Prompt)
- [ ] Channel adapter
- [ ] LLM provider
- [ ] Built-in tool
- [ ] Bug fix
- [ ] Feature (Rust)
- [x] Documentation / Translation
- [x] Refactor / Performance
- [ ] CI / Tooling
- [ ] Other

## Summary

This PR localizes the two major CLI TUI screens:
1. **Onboarding / Setup Wizard Screen (`tui/screens/init_wizard.rs`)**: Refactored the setup steps, provider selection (with localized environment keys and hints), API key verification, default model picker, smart routing configuration, and completion state launcher options.
2. **Agents Management Screen (`tui/screens/agents.rs`)**: Localized the full agents list (with table headers), agent detail view, custom agent builder (inputs for name, description, system prompt, and tools checklists), and template picker.

All user-facing text strings on these screens have been moved to Fluent-native localization files, supporting English, Ukrainian, and Simplified Chinese. In addition, both screens have been integrated into the `i18n_checks` test suite to prevent future localization regressions.

#### [MODIFY] [init_wizard.rs](file:///home/pavver/librefang-ukrainian-cli-i18n/crates/librefang-cli/src/tui/screens/init_wizard.rs)
- Modified `ProviderInfo` and static `PROVIDERS` list to use localizable Fluent keys for provider description hints (e.g. `tui-init-hint-*`).
- Localized all `draw_*` screens including migration detector, API key validation, model picker, smart routing tiers, and daemon launcher options.
- Cleaned up clippy compiler warnings by resolving needless borrows of reference variables during migration done screen rendering.

#### [MODIFY] [agents.rs](file:///home/pavver/librefang-ukrainian-cli-i18n/crates/librefang-cli/src/tui/screens/agents.rs)
- Localized sub-screen title headers, custom builder description, and prompt pre-fills.
- Replaced hardcoded text labels in `draw_detail` (e.g. ID, Name, State, Capabilities, Tags) and lists checklists (`draw_checkbox_list`).
- Modified `TOOL_OPTIONS` structure and `draw_tool_select` to dynamically fetch tool descriptions from Fluent translations.
- Adjusted tool checklist formatting (`format!("  {}{:<16}", check, name)`) to avoid having alphabetic letters in literals with spaces, passing the localization scanner.

[...]

PR #6310, opened 2026-06-24
release: v2026.6.24-beta.23

### LINKED ISSUE #6251: Bug: Consecutive `Role::User` Messages in Aggregated Developer Loop (Anthropic API 400)

### LINKED ISSUE #6264: Proposal: Pluggable Context Rewrite Modules

### LINKED ISSUE #6265: /new command does not work
The /new slash command in LibreFang does not work as expected. The expected behavior is to create a new agent or a new session. Needs investigation.

### LINKED ISSUE #6270: Windows desktop setup fails to launch: plugins.updater is null
### Summary
The Windows desktop app from the latest setup installer exits immediately on launch because the Tauri updater plugin configuration is bundled as null.

### Affected version
- LibreFang: 26.6.32241
- Release: v2026.6.22-beta.21
- Installer: LibreFang_26.6.32241_x64-setup.exe
- OS: Windows x64

### Reproduction
1. Install LibreFang using the latest Windows x64 setup exe.
2. Launch `D:\Programs\LibreFang\librefang-desktop.exe`.

### Actual result
The app exits immediately with exit code 101:

```text
Starting LibreFang Desktop...

thread 'main' panicked at crates\librefang-desktop\src\lib.rs:532:10:
Failed to build Tauri application: PluginInitialization("updater", "Error deserializing 'plugins.updater' within your Tauri configuration: invalid type: null, expected struct Config")
```

### Expected result
The desktop app should open normally.

### Additional context
This reproduced on multiple recent Windows desktop releases:
- 26.6.32179
- 26.6.32190
- 26.6.32241

The source contains updater config in `crates/librefang-desktop/tauri.desktop.conf.json`, but the release workflow appears to build with only target args and without explicitly passing/merging that desktop config overlay. The updater metadata URL also currently returns `Not Found`:

https://github.com/librefang/librefang/releases/latest/download/latest.json

houko:
## Release v2026.6.24-beta.23

_33 PRs from 5 contributors since v2026.6.22-beta.22._

### Added

- Localize TUI Onboarding Wizard and Agents screen (#6253) (@pavver)
- Pluggable context-rewrite modules — per-agent engine + host-run request_llm_summary (closes #6264) (#6287) (@houko)
- Add scriptable tool result transform hook (#6291) (@pavver)
- Per-step errors and re-run with same parameters (#6292) (#6302) (@houko)
- Code_search workspace regex search tool (#6295) (#6307) (@houko)

### Fixed

- Gate provider hourly token budget pre-call so exhaustion flags the fallback chain (#5980) (#5988) (@DaBlitzStein)
- Persist agent skill & MCP-server assignments to agent.toml (#6046) (@DaBlitzStein)
- Embed developer-loop placeholder in first result delivery (closes #6251) (#6254) (@maoxin1234)
- Kill sidecar child on shutdown + forward async delegation result to channel (#6267) (@DaBlitzStein)
- Complete dashboard i18n coverage (#6271) (@pavver)
- Vendor OpenSSL unconditionally so cross-compiled release targets link (#6279) (@houko)
- Merge updater plugin config into base tauri.conf.json (closes #6270) (#6283) (@houko)
- Handle /new in the TUI chat surfaces (closes #6265) (#6284) (@houko)
- Merge the dual [Unreleased] sections into one at the top (#6285) (@houko)
- Forward web-UI-initiated delegation results to the home channel (refs #6266) (#6286) (@houko)
- AUTH PLAIN fallback for SMTP + expose input/error in workflow runs list (#6293) (@DaBlitzStein)
- Scope GET /api/workflows/{id}/runs to the path workflow (#6298) (@houko)
- Block sandbox escape via intermediate-ancestor symlink on writes to non-existent dirs (#6299) (@houko)
- Reject newline/CR/NUL in secret key to prevent secrets.env line injection (#6300) (@houko)
- Graceful 400 for non-table [memory]/[proactive_memory] in PATCH /api/memory/config (#6301) (@houko)
- Apply api_key_env/base_url in PATCH /api/agents/{id}/config instead of dropping them (#6303) (@houko)
- Wire provider cooldown breaker into the LLM retry path and fix its probe gate (#6305) (@houko)
- Gate scriptable transform-hook tests behind cfg(unix) to unbreak Windows CI (#6306) (@houko)
- Gate use std::sync::Arc behind cfg(unix) to complete the Windows-red fix (#6308) (@houko)
- Install libdbus-1-dev in the release Bump Version job (#6309) (@houko)

### Performance

- JSON-aware token estimation for tool paths (#6281) (@maoxin1234)

<details>
<summary>Documentation, maintenance, and other internal changes</summary>

### Maintenance

- Token-estimation accuracy benchmark with multi-tokenizer baselines (#6269) (@maoxin1234)
- Bump the cargo-minor-patch group across 1 directory with 8 updates (#6276) (@app/dependabot)
- Bump wasmtime from 45.0.2 to 46.0.0 (#6277) (@app/dependabot)
- Bump cron from 0.16.0 to 0.17.0 (#6278) (@app/dependabot)
- Bump vulnerable/yanked lockfile deps to clear global CI failures (#6280) (@houko)
- Bump actions/checkout from 6 to 7 (#6296) (@app/dependabot)
- Bump actions/cache from 5.0.5 to 6.0.0 (#6297) (@app/dependabot)

</details>

---
**Full diff:** https://github.com/librefang/librefang/compare/v2026.6.22-beta.22...v2026.6.24-beta.23


houko (commented):
**CHANGELOG section ordering** — `[Unreleased]` has been pushed below the new `[2026.6.24]` section.

Keep a Changelog (which this repo follows per the CHANGELOG header) requires `[Unreleased]` to always be the first section:

```
## [Unreleased]          ← should appear here
…

## [2026.6.24] - 2026-06-24
…
```

The four post-cutoff entries under `[Unreleased]` (#6262, #6242, #6209, #3576) are not missing from the release notes — they legitimately landed after the cut — so this is ordering-only.

The bump-version script inserts the new release above `[Unreleased]` rather than below it; fixing the script would stop this from recurring on every version bump PR.

---

PR #6323, opened 2026-06-25
fix(channels): describe inbound images on the debounced channel path (#6321)

### LINKED ISSUE #6321: Bug: debounced channel path skips inbound image description ([media] image_description), text-only models lose vision
## Symptom

A channel bot (`bot-A`) with a message-coalesce window configured (`message_coalesce_window_ms = 3000`), routed to an agent (`agent-A`) on a **text-only** model (DeepSeek, `supports_vision = false`), never receives a description of inbound photos. The agent only sees the redacted placeholder `[image omitted: model has no vision support]` (the #6010 text-only redaction) and replies that it cannot see the image.

An otherwise identical bot **without** a coalesce window, same model, **does** get a Groq description and can answer about the image. The only differing variable is `message_coalesce_window_ms`.

`[media] image_description = true` is set globally; the Groq provider/key work (verified by calling the vision model directly with the same image — it returns a correct description).

## Root cause (read from current source, tag v2026.6.22-beta.22)

Inbound dispatch forks on the debounce window in `BridgeManager::start_adapter` (`crates/librefang-channels/src/bridge.rs:1435-1438`):

- **Fast path** (`debounce_ms == 0`) → `dispatch_message()`. Its image branch downloads the photo, builds an `ImageFile` block, **and calls the description step**:

```rust
// crates/librefang-channels/src/bridge.rs:3959
if let Some(desc_block) = maybe_describe_inbound_image(handle, Some(saved)).await {
    final_blocks.insert(0, desc_block);   // prepend Groq description
}
// ...
// crates/librefang-channels/src/bridge.rs:3966
dispatch_with_blocks(final_blocks, message, /* ... */).await;   // shared tail
```

- **Debounce path** (`debounce_ms > 0`) → pre-downloads images into `image_blocks` and dispatches via `flush_debounced()` → `dispatch_with_blocks()` (`bridge.rs:1223`). This branch builds the image blocks itself and **never calls `maybe_describe_inbound_image`**. `dispatch_with_blocks()` (`bridge.rs:6245`) contains no description call in its entire body.

So `maybe_describe_inbound_image` (defined `bridge.rs:5453`) is reachable **only** from `dispatch_message`. The `[media] image_description` flag is consequently only ever consulted on the fast path (the flag check lives in `describe_inbound_image` at `crates/librefang-api/src/channel_bridge.rs`, reached only via `maybe_describe_inbound_image`).

## Why it surfaces as it does

For a text-only model the un-described image block is then redacted to a placeholder by the #6010 vision-gate (`redact_images_for_text_only`, `crates/librefang-runtime/src/agent_loop/mod.rs`), so there is **no HTTP 400 / crash** — the failure is silent: the agent simply never receives any image content or description. Vision-capable models are unaffected (they would see the raw image either way). The bug is therefore invisible except as "the bot ignores photos" on text-only + coalesce setups.

## Proposed fix

The two paths already converge on the shared tail `dispatch_with_blocks()`. The divergence is only that the content→blocks conversion (which includes description) is duplicated, and the debounced copy omits the description step.

- **Minimal:** in `flush_debounced` (where the image `blocks` are assembled, around `bridge.rs:1073-1083`), mirror the `ImageFile`-extraction + `maybe_describe_inbound_image` prepend from `bridge.rs:3950-3963` before calling `dispatch_with_blocks`.
- **Proper (recommended):** factor the content→blocks conversion **including `maybe_describe_inbound_image`** into a single shared helper that both the immediate and the debounced paths call before the common `dispatch_with_blocks`. The fork should be only "coalesce or not" at ingest; all downstream media handling (description, future transcription/OCR) should live in one path so a parallel branch cannot drift out of parity again.

## Setup (for reference)

- v2026.6.22-beta.22
- Pattern: sidecar/channel bot with `message_coalesce_window_ms > 0`, default agent on a text-only model (`supports_vision = false`), `[media] image_description = true`.
- Repro: send a photo to a coalesced bot vs. a non-coalesced bot on the same model — only the non-coalesced one is described.
- Workaround: remove `message_coalesce_window_ms` for affected bots (forces the fast path), at the cost of losing message coalescing.

Happy to test a patch.

Refs: #6010 (text-only image redaction — closed; this is the adjacent description gap on the debounce path)

houko:
## Summary

Fixes #6321. Inbound channel images were only described on the immediate dispatch path; the debounced (message-coalesce) path never ran the description step, so text-only models lost vision on coalesced bots.

The kernel's image-description step (`maybe_describe_inbound_image`, gated by `[media] image_description`) was reachable only from `dispatch_message` (the `message_coalesce_window_ms == 0` fast path). The debounced path (`message_coalesce_window_ms > 0`) assembles its image blocks in `flush_debounced` and calls the shared `dispatch_with_blocks` tail directly, which contains no description call. So `[media] image_description` was only ever consulted when coalescing was off.

On a text-only model the un-described `ImageFile` block is then redacted to the `[image omitted: …]` placeholder by the #6010 vision gate, so the failure is silent — no HTTP 400, no crash, the agent simply replies that it cannot see the image. Vision-capable models were unaffected (they read the raw bytes either way), which is why this only surfaced on text-only + coalesce setups.

## Change

- Extract the per-image describe-and-prepend logic into one shared helper, `prepend_image_descriptions(handle, blocks)`, placed next to `maybe_describe_inbound_image`.
- Call it from **both** `dispatch_message` (immediate) and `flush_debounced` (debounced) right before `dispatch_with_blocks`, so the description step lives in one place and the two paths cannot drift out of parity again — the root cause was exactly that divergence.
- The helper describes **every** `ImageFile` block, not just the first, inserting each description immediately before its image. This matters because `MessageDebouncer::drain` coalesces the image blocks of several buffered messages into one list (`all_blocks.extend(...)`), so a debounced batch can legitimately carry multiple images.

The step self-gates inside `describe_inbound_image` (`[media] image_description`), so it remains a cheap no-op when the operator has not opted in, and a failed/timed-out description still degrades to the existing opaque `[Image description unavailable]` note rather than dropping the image.

### Behavior note (intentional)

On the immediate path the description used to be prepended at index 0 (before the caption); it is now inserted immediately before the image block (after the caption), matching the debounced path. The description is still adjacent to and precedes the image, so model behavior is unchanged; the ordering is now consistent across both paths. No test asserted the old absolute-front position.

## Verification

- `cargo check -p librefang-channels --lib` — clean
- `cargo clippy -p librefang-channels --all-targets -- -D warnings` — zero warnings
- `cargo test -p librefang-channels --lib` — **510 passed, 0 failed**
- `cargo fmt -p librefang-channels -- --check` — clean

New regression tests in `bridge::tests::inbound_image_description` (reuse the existing `DescribeHandle` fake):

- `prepend_describes_every_imagefile_before_its_block` — multi-image list gets a description before each image; both images hit the vision path.
- `prepend_is_noop_when_description_disabled` — `Ok(None)` (feature off) leaves blocks unchanged.
- `prepend_leaves_non_image_blocks_untouched` — a text-only list triggers no vision call.

The existing `maybe_describe_inbound_image` cases (success / disabled / provider-failure-doesn't-leak / timeout / non-image-mime) continue to pass.

## Out of scope

Live LLM round-trip verification (real Groq vision call against a coalesced bot) requires a running daemon and is left for a maintainer per the repo's human-only integration-testing policy.

### DIFF
--- a/crates/librefang-channels/src/bridge.rs
+++ b/crates/librefang-channels/src/bridge.rs
@@ -1082,6 +1082,13 @@ fn flush_debounced(
                 );
             }
 
+            // Parity with the immediate (`dispatch_message`) path: describe
+            // each inbound image so text-only models receive a description next
+            // to the `ImageFile` block. Coalesced (debounced) images used to
+            // skip this step entirely — the #6321 bug — because only the
+            // immediate path called it. Self-gates on `[media] image_description`.
+            let blocks = prepend_image_descriptions(&channel_handle, blocks).await;
+
             let ct_str = channel_type_str(&merged_msg.channel);
 
             // --- Input sanitization (prompt injection detection) ---
@@ -3935,33 +3942,15 @@ async fn dispatch_message(
                 ContentBlock::Image { .. } | ContentBlock::ImageFile { .. }
             )
         }) {
-            // We have actual image data.
-            //
-            // Optionally run `describe_inbound_image` so that text-only
-            // models receive a natural-language description next to the
-            // `ImageFile` block.  Vision-capable models ignore the text
-            // block and use the raw image bytes directly.
-            //
-            // Extract the (path, media_type) from the first `ImageFile`
-            // block so `maybe_describe_inbound_image` can pass it to the
-            // kernel's `MediaEngine`.  Inline `Image` blocks (base64
-            // fallback when the save failed) have no on-disk path, so
-            // skip description for those.
-            let saved_image: Option<(std::path::PathBuf, String)> = blocks.iter().find_map(|b| {
-                if let ContentBlock::ImageFile { path, media_type } = b {
-                    Some((std::path::PathBuf::from(path), media_type.clone()))
-                } else {
-                    None
-                }
-            });
-            let mut final_blocks = blocks;
-            if let Some(ref saved) = saved_image {
-                if let Some(desc_block) = maybe_describe_inbound_image(handle, Some(saved)).await {
-                    // Prepend the description so the model reads context
-                    // before the raw image bytes.
-                    final_blocks.insert(0, desc_block);
-                }
-            }
+            // We have actual image data. Optionally run `describe_inbound_image`
+            // for each inbound image so that text-only models receive a
+            // natural-language description next to the `ImageFile` block;
+            // vision-capable models ignore the text block and use the raw image
+            // bytes directly. Inline `Image` blocks (base64 fallback when the
+            // save failed) have no on-disk path and are left untouched. This is
+            // the same helper the debounced path uses, so the two cannot drift
+            // out of parity (refs #6321).
+            let final_blocks = prepend_image_descriptions(handle, blocks).await;
             // Send as structured blocks for vision
             dispatch_with_blocks(
                 final_blocks,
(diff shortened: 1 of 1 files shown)


crates/librefang-channels/src/bridge.rs
houko, line 1: CLAUDE.md: "Don't write multi-paragraph docstrings or multi-line comment blocks — one short line max."

Three new comment blocks added in this PR violate that rule:

1. The `///` doc on `prepend_image_descriptions` spans nine lines across four paragraphs.
2. The `//` block at the `flush_debounced` call-site is three lines.
3. The `//` block replacing the old comment in `dispatch_message` is three lines.

The WHY is non-obvious enough to warrant a comment on each — just collapse to one line:

```rust
/// Prepend a vision-description text block before every `ImageFile` in `blocks` so both inbound paths describe images identically; self-gates on `[media] image_description` (refs #6321).
```

```rust
// Parity with `dispatch_message`: describe inbound images for text-only models via the shared helper (refs #6321).
```

```rust
// Shared helper keeps both paths in parity (refs #6321).
```

---

houko (commented):
Code is correct and well-tested (three new unit tests cover the multi-image, disabled, and text-only cases). One comment-style issue to fix before merge — see inline note.

---

PR #6338, opened 2026-06-26
ci(mobile): put NDK bin on PATH so openssl-src finds the legacy ranlib symlink

houko:
## Problem

The Android Tauri builds — mobile-smoke `Android (debug, unsigned)` and release `Mobile / Android` — have been failing on `main` since OpenSSL was vendored (#6279).
#6335 was meant to fix this but only got halfway: the latest `main` push still fails with `aarch64-linux-android-ranlib: not found` (exit 127), e.g. [run 28226134802](https://github.com/librefang/librefang/actions/runs/28226134802/job/83618428640).

## Root cause

#6335 symlinked the legacy `<target>-ranlib` / `-ar` / `-nm` / `-strip` names to their `llvm-*` equivalents inside the NDK bin directory, but **never put that directory on `$PATH`**.

The CI log shows why the symlink alone is not enough — openssl-src invokes Configure as:

```
env -u CROSS_COMPILE AR="<abs>/llvm-ar" CC="<abs>/aarch64-linux-android26-clang" RANLIB="aarch64-linux-android-ranlib" perl ./Configure ...
```

It passes `AR` as an **absolute path** (resolves fine) but `RANLIB` as a **bare command name**. The bare name is resolved through a `$PATH` lookup, and the NDK bin directory — where the `aarch64-linux-android-ranlib` symlink lives — was not on `$PATH`. So `make install_dev` exits 127.

That `AR` resolves from exactly the `$NDK_BIN` directory we symlink into is what confirms the symlink itself is valid (not dangling); the only missing piece is PATH visibility.

## Fix

Append `$NDK_BIN` to `$GITHUB_PATH` in the existing "Symlink legacy NDK binutils" step, so the bare-name `RANLIB` resolves to the symlink.

- `.github/workflows/mobile-smoke.yml` — aarch64 job
- `.github/workflows/release.yml` — aarch64 + armv7 job

The symlinks themselves are unchanged; this only makes them findable.

## Verification

- The Android cross-compile is Linux-only (NDK r27 host toolchain) and cannot be reproduced locally, so the fix is grounded directly in the failing run's log: `RANLIB="aarch64-linux-android-ranlib"` (bare) vs `AR="<abs>/llvm-ar"` (absolute), with `AR` resolving from the same `$NDK_BIN` we symlink into.
- Both workflow files pass `yaml.safe_load`.
- The real verification is the Android jobs on this PR going green.

## Note (separate, transient)

The `Test / Security` job on the same `main` push failed independently with a transient crates.io download error (`OpenSSL SSL_read: SSL_ERROR_SYSCALL` while fetching `ed25519-dalek`) — infrastructure, not code. That run's failed jobs were re-run; no change required here.


houko (commented):
The fix itself looks correct — the root cause analysis (bare `RANLIB` resolved via PATH, absolute `AR` isn't) matches the CI log and the `echo "$NDK_BIN" >> "$GITHUB_PATH"` one-liner is the right fix.

One prose-style note (CLAUDE.md "Prose wrapping" rule): the paragraph this PR modifies in both workflow files column-wraps mid-sentence.
The project rule is one sentence per line in any prose block being touched.
The new last sentence currently spans four lines:

```yaml
      # "ranlib: not found" (exit 127). Symlink the legacy names to the llvm
      # equivalents AND put the NDK bin dir on PATH: openssl-src passes RANLIB
      # to Configure as a bare command name (unlike AR, which it passes as an
      # absolute path), so the symlink is only resolvable via a PATH lookup.
```

Sentence-based form would put `"ranlib: not found" (exit 127).` and the `Symlink…` sentence on separate lines (the earlier sentences in the unchanged portion of the block have the same issue, but those aren't touched by this PR so they're out of scope here).
This doesn't affect correctness — just flagging so the author is aware before the maintainer reviews style.


---

PR #6345, opened 2026-06-27
fix(channels): keep Telegram setup form available after describe timeout

pavver:
## Summary

The dashboard builds sidecar setup forms from schemas returned by each adapter's boot-time `--describe` probe.

Telegram is the first adapter in the catalog.
Its initial cold Python import can exceed the five-second probe timeout on slower systems.
Because Telegram had no static fallback schema, that one timeout left the setup drawer empty and disabled Save for the lifetime of the daemon.

Closing and reopening LibreFang Desktop did not necessarily retry discovery because the Tauri background process remained alive and retained the failed schema cache.

## Changes

- Add a compile-time Telegram fallback schema matching the Python and Rust adapters.
- Keep the Bot Token field required and secret.
- Keep Allowed User IDs and Clear done reaction under advanced settings.
- Use the fallback whenever Telegram `--describe` fails, so the setup form remains usable.
- Correct the unavailable-form guidance in English, Ukrainian, and Chinese.
- Stop claiming that channel reload refreshes schema discovery.
- Direct operators to restart the daemon because the schema cache is populated only at boot.
- Add unit coverage for the Telegram catalog fallback and its field contract.

## Verification

- `cargo fmt --check`
- `cargo test -p librefang-api --lib static_schema_tests`
  - 6 passed.
- `pnpm test`
  - 85 test files passed.
  - 839 tests passed.
- `pnpm build`
  - Production dashboard build succeeded.
- Built `librefang-desktop` in the repository's sanctioned `Dockerfile.rust-dev` image.
- Built and inspected a local Arch package:
  - `librefang-desktop-bin 2026.6.26_beta.24-2`
  - Package file list and runtime library resolution verified.
- Reproduced the original API response with:
  - `fields: []`
  - `schema_error: python3 ...--describe timed out after 5s`
- Confirmed that a genuine process restart allows the live Telegram schema to load.


crates/librefang-api/src/routes/channels.rs
```diff
@@ -219,6 +219,39 @@ struct SidecarCatalogEntry {
 /// or release notes. `webhook` is deliberately omitted: it still has an
 /// in-process entry in `CHANNEL_REGISTRY` and we must not show two
 /// "webhook" cards on the page.
+/// Compile-time field descriptors for the Telegram adapter.
+///
+/// Telegram is the first adapter probed at daemon boot. On slow disks its
+/// cold Python import can exceed the five-second describe timeout even though
+/// the embedded SDK is healthy. Keep this fallback aligned with both the
+/// Python and Rust Telegram schemas so the configure form remains usable.
+const TELEGRAM_STATIC_FIELDS: &[StaticSidecarField] = &[
```
houko, line 228: Doc comment merge: the three `///` lines that were already here (the "webhook" note) are the tail of the Feishu const's doc comment block, which continues immediately with `/// Compile-time field descriptors for the Feishu / Lark adapter.` further down in the file.

Inserting the new `TELEGRAM_STATIC_FIELDS` const — and its own `///` doc block — right between those "webhook" lines and the Feishu const's leading `///` lines merges all of them into one continuous `///` block.
Rust attaches a consecutive `///` block to the _next_ item, so the Telegram const ends up with a doc comment that opens with the webhook caveat, while the Feishu const silently loses the last three lines of its documentation.

The fix is to move the insertion point to _before_ the `/// or release notes…` / `/// in-process entry…` / `/// "webhook" cards on the page.` lines, so the Feishu doc comment remains intact:

```rust
/// Compile-time field descriptors for the Telegram adapter.
///
/// Telegram is the first adapter probed at daemon boot. …
const TELEGRAM_STATIC_FIELDS: &[StaticSidecarField] = &[ … ];

// ← existing Feishu doc comment block starts here (unmodified)
/// Compile-time field descriptors for the Feishu / Lark adapter.
/// …
/// or release notes. `webhook` is deliberately omitted: …
const FEISHU_STATIC_FIELDS: …
```

This doesn't affect compilation or tests (doc comments don't participate in either), but it leaves misleading documentation in the codebase.

---

houko (commented):
One doc comment placement issue found; everything else looks good — the fallback schema, tests, and locale string fixes are all correct.

---

houko (commented):
Two issues in `crates/librefang-api/src/routes/channels.rs`:

**1. Doc comment bleed into `TELEGRAM_STATIC_FIELDS`**

`TELEGRAM_STATIC_FIELDS` is inserted immediately after the last line of the preceding constant's doc block (`/// "webhook" cards on the page.`) with no blank line separating them. In Rust, all consecutive `///` lines before an item declaration merge into that item's doc comment — so the "webhook" text that was documenting an earlier constant (likely `SIDECAR_CATALOG`) will appear in `TELEGRAM_STATIC_FIELDS`'s rendered docs instead.

Fix: add one blank line before `/// Compile-time field descriptors for the Telegram adapter.`

**2. Multi-paragraph doc comment on `TELEGRAM_STATIC_FIELDS`** (CLAUDE.md: "one short line max")

```rust
/// Compile-time field descriptors for the Telegram adapter.
///
/// Telegram is the first adapter probed at daemon boot. On slow disks its
/// cold Python import can exceed the five-second describe timeout even though
/// the embedded SDK is healthy. Keep this fallback aligned with both the
/// Python and Rust Telegram schemas so the configure form remains usable.
const TELEGRAM_STATIC_FIELDS: ...
```

The WHY is worth keeping but needs to fit in one line, e.g.:
```rust
/// Compile-time fallback schema for Telegram; used when the `--describe` probe times out at daemon boot.
```

---

houko (commented):
One doc-comment issue in `crates/librefang-api/src/routes/channels.rs`.

The new `TELEGRAM_STATIC_FIELDS` constant is inserted immediately after the three `///` lines that ended the previous constant's doc block:

```rust
/// or release notes. `webhook` is deliberately omitted: it still has an
/// in-process entry in `CHANNEL_REGISTRY` and we must not show two
/// "webhook" cards on the page.
/// Compile-time field descriptors for the Telegram adapter.   ← inserted here, no separator
///
/// Telegram is the first adapter probed at daemon boot. …
const TELEGRAM_STATIC_FIELDS: &[StaticSidecarField] = &[…];
```

Because there is no blank `///` line between the two blocks, Rust treats all seven consecutive `///` lines as a single doc comment for `TELEGRAM_STATIC_FIELDS`. The webhook-exclusion text is now part of the Telegram constant's docs, where it reads as noise.

Fix — add one blank `///` separator before the Telegram-specific block:

```diff
 /// "webhook" cards on the page.
+///
 /// Compile-time field descriptors for the Telegram adapter.
```

That makes the webhook text float as a dangling comment (or re-attach to the preceding constant if the file structure supports it). Either way it removes the incorrect attribution from the Telegram constant.

---

PR #6351, opened 2026-06-29
fix(channels): enrich coalesced mixed-media batches on the debounced path (#6348)

### LINKED ISSUE #6348: Bug: coalesced mixed-media batches lose non-image enrichment (audio transcript / doc extract) on the debounced path
## Symptom

On a channel bot with a coalesce window (`message_coalesce_window_ms > 0`), when an inbound batch contains **an image together with other media** (e.g. a photo + a voice note sent within the debounce window), the non-image media is delivered to the agent as a bare text placeholder — `[Voice message (5s): <url>]`, `[File (<name>): <url>]` — instead of its transcript / extracted text. The agent never receives the voice transcription or document contents.

This is the same class of defect that #6321 fixed for images (PR #6323), but only images were made path-agnostic; audio/voice/document enrichment was not generalized.

## Scope (precise)

The gap is narrow but real — it depends on whether the coalesced batch contains an image:

| Case | Path | Result |
|---|---|---|
| Voice/doc alone, or in an image-free batch | `blocks == None` → `dispatch_message` | transcribed / enriched ✅ |
| Voice/doc coalesced **with an image** | `blocks == Some` → `dispatch_with_blocks` | placeholder text, **not** transcribed/enriched ❌ |

## Root cause (tag v2026.6.26-beta.24, `crates/librefang-channels/src/bridge.rs`)

1. The debounce ingest pre-downloads **only images** into blocks; all other media is pushed as raw `ChannelContent`:

```rust
// bridge.rs:1527
let image_blocks = if let ChannelContent::Image { ref url, ref caption, ref mime_type } = message.content {
    // download_image_to_blocks(...)
} else {
    None
};
let pending = PendingMessage { message, image_blocks };
```

2. On coalesce, `MessageDebouncer::drain` turns every non-image message into a **text placeholder** via `content_to_text` (bridge.rs:984-1020): `[Voice message (Ns): url]`, `[Audio (Ns): url]`, `[File (name): url]`.

3. The debounced flush only enriches images, then dispatches the assembled blocks:

```rust
// bridge.rs:1088  — image describe (the #6321 fix), no audio/doc step
let blocks = prepend_image_descriptions(&channel_handle, blocks).await;
// ... dispatch_with_blocks(blocks, ...)   // bridge.rs:~1228
```

So when `blocks == Some` (an image was in the batch), the path goes through `dispatch_with_blocks`, which runs **only** `prepend_image_descriptions`. The audio/doc enrichment steps live exclusively in `dispatch_message` and are never reached for that batch:

```rust
// dispatch_message — full per-type enrichment (the only place these run):
// bridge.rs:3947  prepend_image_descriptions(...)
// bridge.rs:4024  maybe_transcribe_inbound_audio(...)   // Voice arm
// bridge.rs:4087  maybe_transcribe_inbound_audio(...)   // Audio arm
// bridge.rs:5841  enrich_saved_file(...)                // File/doc arm
```

The `else` branch of the flush (no image in batch) calls `dispatch_message` directly, which is why standalone voice/docs still work — the gap is specifically the **mixed image + other-media** batch.

## Why this is the same divergence #6321 warned about

#6321/#6323 fixed image parity by extracting `prepend_image_descriptions` and calling it from both the immediate (`dispatch_message`) and debounced (`dispatch_with_blocks`) paths — explicitly so the two paths *cannot drift out of parity again*. But enrichment for the other media types was left only on the immediate path, so the divergence persists for audio and documents.

## Proposed fix

Generalize the #6321 approach to **all** media types: replace the image-only `prepend_image_descriptions` with a single, type-agnostic `enrich_media_blocks(handle, blocks)` that, for each media block in the final assembled list, runs the appropriate existing enrichment —

- `ImageFile` → `maybe_describe_inbound_image`
- audio/voice → `maybe_transcribe_inbound_audio`
- document/file → `enrich_saved_file`

— and call it from the single shared point before `dispatch_with_blocks`, used by **both** paths. The underlying enrichment helpers already exist; this is a consolidation, not new capability. It also requires the debounce ingest (bridge.rs:1527) to download non-image media into blocks too (mirroring `download_image_to_blocks`), so coalesced audio/docs carry a file path rather than a bare URL placeholder.

Net effect: media is enriched identically whether coalesced or not, and a mixed photo+voice+pdf burst reaches the agent as one turn with every attachment turned into text — which also benefits text-only models.

## Setup (for reference)

- v2026.6.26-beta.24
- Pattern: channel bot with `message_coalesce_window_ms > 0`; user sends an image and a voice note (or a document) within the coalesce window.
- Repro: the voice arrives as `[Voice message (Ns): <url>]` with no transcript; the same voice sent alone is transcribed.

Happy to test a patch.

Refs: #6321 / #6323 (image parity fix this generalizes)

houko:
Fixes #6348.

## Problem

On a channel bot with a coalesce window (`message_coalesce_window_ms > 0`), when an inbound batch contained an image **together with** other media (e.g. a photo + a voice note within the debounce window), the non-image media reached the agent as a bare text placeholder — `[Voice message (Ns): url]`, `[File (name): url]` — instead of its transcript / extracted text.

The gap was narrow but real:

| Case | Path | Pre-fix result |
|---|---|---|
| Voice/doc alone, or in an image-free batch | `blocks == None` → `dispatch_message` | transcribed / enriched ✅ |
| Voice/doc coalesced **with an image** | `blocks == Some` → debounced flush | placeholder text, **not** transcribed/enriched ❌ |

This is the same class of divergence #6321 / #6323 fixed for images — but only images were made path-agnostic. The debounce ingest pre-downloaded only images, and the debounced flush only ran `prepend_image_descriptions`; audio transcription and document extraction lived solely in `dispatch_message`'s per-type arms and were never reached for a mixed batch.

## Fix

Generalize the #6321 approach to **all** media types by funnelling both inbound paths through two shared helpers, so they cannot drift out of parity again:

- **`download_media_blocks(handle, adapter, content) -> Option<DownloadedMedia>`** — downloads any attachment (image / file / voice / audio / video) to disk and builds its blocks (caption / context headers, the `ImageFile` or `[File: …] saved to …` marker, and any document extraction already baked into `download_file_to_blocks`). It is I/O-bound only, so it is safe to run inline on the debounce ingest loop exactly as the original image-only pre-download was. It returns `None` for non-media content and for a failed download, so the caller falls back to the text placeholder — matching the immediate path's fall-through.
- **`enrich_media(handle, media) -> Vec<ContentBlock>`** — applies the deferred, LLM-backed enrichment recorded on the payload (image description via `prepend_image_descriptions`; audio transcription via `maybe_transcribe_inbound_audio`, inserted right after the context header) in the dispatch task, off the ingest loop.

Wiring:

- The debounce **ingest** now pre-downloads every media type (not just images) while the message's individual content is still available — after coalescing the debouncer only retains a merged text placeholder.
- The debounced **flush** enriches each coalesced attachment with the same `enrich_media` the immediate path uses, concatenates in arrival order, and prepends the coalesced text (which keeps media captions inside the sanitizer-checked text).
- `dispatch_message`'s five per-type media arms collapse into one `download_media_blocks` + `enrich_media` + `dispatch_with_blocks` call, preserving their exact output: context headers, transcription inserted after the header (index 1), and — deliberately — no image description for image-typed documents (matching the old File arm).

Net effect: a mixed photo+voice+pdf burst reaches the agent as one turn with every attachment turned into text, identically whether or not it was coalesced — which also benefits text-only models.

## Verification

- `cargo check -p librefang-channels --lib` — clean (no warnings).
- New `enrich_media` unit tests next to the existing `maybe_describe_inbound_image` / `maybe_transcribe_inbound_audio` fakes:
  - `enrich_media_describe_image_describes_each_imagefile` — `DescribeImage` inserts the description before the `ImageFile`.
  - `enrich_media_none_leaves_blocks_and_skips_vision` — `None` returns blocks verbatim with no kernel call (an image-typed document is not described).
  - `enrich_media_transcribe_audio_inserts_after_header` — `TranscribeAudio` inserts `[Transcription: …]` between the `[Voice message …]` header and the saved-path block (same order the immediate path's `transcription_block_lands_between_header_and_file_path` test pins).
  - `enrich_media_transcribe_audio_noop_when_disabled` — `Ok(None)` (feature off) leaves the [header, saved-path] pair unchanged.
- Existing debouncer drain tests updated for the new `Vec<DownloadedMedia>` return shape.
- clippy + full test suite (including the integration shards) run in CI.

Live LLM round-trip verification (a real coalesced photo+voice burst against a running daemon) is left for a maintainer per the repo's human-only integration-testing policy.


houko (commented):
**Style: multi-line comment blocks violate the project's one-line-max rule**

CLAUDE.md (and the project system prompt): *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."*

The following new items in `bridge.rs` violate this:

**Multi-paragraph doc comments** (need to be reduced to one line each):
- `DownloadedMedia` struct — 4-line block including an empty `///` separator
- `download_media_blocks` fn — 5-line block with two `///` paragraphs
- `enrich_media` fn — 4-line block
- `PendingMessage.media` field — 3-line doc comment

**Multi-line inline comment blocks** (each should collapse to ≤1 line or be removed):
- The 2-line block before `return Some((pm.message, pm.media.into_iter().collect()));` in `drain()`
- The 2-line block before `let mut media: Vec<DownloadedMedia>` in `drain()`
- The 2-line block before `for m in media {` in `flush_debounced()`
- The 3-line block before `let media = download_media_blocks(...)` in the debounce ingest loop
- The 3-line block before `if let Some(media) = download_media_blocks(...)` in `dispatch_message()`

Comments explaining *why* (non-obvious invariants, issue references) are welcome, but must fit in one line. Please condense.

---

Issue #6334, opened 2026-06-26
Publish official AUR packages for LibreFang releases


pavver:
# Publish official AUR packages for LibreFang releases

## Current status

PR #6314 added upstream-maintained AUR packaging files under `packaging/aur/`.
The package sources build locally, but the release packages are not available through AUR yet.

The intended official AUR packages are:

- `librefang-bin`
- `librefang-desktop-bin`
- `librefang-docker`

Current AUR state:

- `librefang-bin`: not published
- `librefang-desktop-bin`: not published
- `librefang-docker`: not published

There are existing community `-git` packages maintained by `wuxxin`:

- `librefang-git`
- `librefang-desktop-git`
- `librefang-whatsapp-gateway-git`
- `python-librefang-sdk-git`

Those packages are useful for source/git builds, but they do not provide the official release-pinned binary packages added in `packaging/aur/`.

## What was already verified

The package files added in the repository were tested locally:

- `makepkg -f` succeeds for `librefang-bin`
- `makepkg -f` succeeds for `librefang-desktop-bin`
- `makepkg -f` succeeds for `librefang-docker`
- `.SRCINFO` was checked against `makepkg --printsrcinfo`
- the staged native binary reports the expected LibreFang version
- the staged desktop binary starts with `--help`
- the Docker image was pulled and smoke-tested with `librefang --version`

## Goal

Make the official release packages installable from AUR, for example:

```bash
yay -S librefang-bin
yay -S librefang-desktop-bin
yay -S librefang-docker
```

Future releases should update the AUR package repositories automatically.

## Option 1: Maintainer manages AUR publishing directly

The maintainer can set up and maintain the AUR package repositories manually or with their own automation.

Required steps:

1. Create the AUR package repositories:

   - `ssh://aur@aur.archlinux.org/librefang-bin.git`
   - `ssh://aur@aur.archlinux.org/librefang-desktop-bin.git`
   - `ssh://aur@aur.archlinux.org/librefang-docker.git`

2. Copy the matching files from this repository:

   - `packaging/aur/librefang-bin/*`
   - `packaging/aur/librefang-desktop-bin/*`
   - `packaging/aur/librefang-docker/*`

3. Commit and push each package to AUR.

4. On every LibreFang release:

   - update `pkgver`
   - update release URLs if needed
   - update `sha256sums`
   - regenerate `.SRCINFO`
   - run `makepkg -f`
   - push the updated package files to AUR

[...]

PR #6366, opened 2026-06-30
fix(deps): clear cargo-deny advisory failures on main

houko:
## What

Restore a green `cargo-deny (advisories)` job on `main`. Two RustSec advisories began failing the audit once they were published to the live advisory DB — `main` went red with no source change of ours.

## Advisories

- **RUSTSEC-2026-0190** — `anyhow` 1.0.102 has an unsoundness in `Error::downcast_mut()` (UB when downcasting after `Error::context`). Fixed upstream in 1.0.103. Remediated by bumping the lockfile only (`cargo update -p anyhow`), so it applies to every dependency path, not just the `agent-client-protocol` chain that surfaced it.
- **RUSTSEC-2026-0192** — `ttf-parser` 0.25.1 is unmaintained (author has stated no further fixes). No safe upgrade is available, and it is transitive via `pdf-extract -> lopdf -> ttf-parser`. Added a scoped `ignore` entry to `deny.toml` next to the other unmaintained transitive crates, matching the existing policy and the precedent set by #6339 (the lopdf RUSTSEC fix).

## Changes

- `Cargo.lock`: `anyhow` 1.0.102 -> 1.0.103.
- `deny.toml`: ignore `RUSTSEC-2026-0192` with a linked, scoped reason.

## Verification

- `cargo deny check advisories` -> `advisories ok`.
- `cargo deny check` (all: advisories, bans, licenses, sources) -> `advisories ok, bans ok, licenses ok, sources ok`. Pre-existing wildcard `warn`s on workspace-internal path deps are unchanged.

## Out-of-scope findings (surfaced, not fixed here)

- **`Nix Build` failure on `main`** — `error: Nix daemon disconnected unexpectedly` during `librefang-acp` compilation. Infra/OOM flake, not a code issue; its failed job has been re-run on `main`.
- **Misplaced `## [Unreleased]` in `CHANGELOG.md`** — the single `## [Unreleased]` heading is buried between `[2026.6.24]` and `[2026.6.22]` (≈ line 136) instead of at the top above `[2026.6.29]`, and the section is ~480 lines with duplicated `### Fixed` / `### Added` subsections — pending entries accumulated across cycles without reorganization. Entries are not dropped today (there is only one `[Unreleased]`), but the next release will emit a release block in the wrong position and the section is disorganized. Reorganizing it needs a maintainer call on what was already shipped (e.g. #6334 appears both here and in `[2026.6.29]`), so it is intentionally left out of this CI-fix PR — and no CHANGELOG entry was added here, since adding one at the correct top position would create a second `[Unreleased]` and trip the duplicate guard.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -270,9 +270,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
 name = "approx"
--- a/deny.toml
+++ b/deny.toml
@@ -61,6 +61,7 @@ ignore = [
     { id = "RUSTSEC-2025-0081", reason = "unic-* unmaintained (kuchikiki/selectors chain). https://rustsec.org/advisories/RUSTSEC-2025-0081" },
     { id = "RUSTSEC-2025-0098", reason = "unic-* unmaintained (kuchikiki/selectors chain). https://rustsec.org/advisories/RUSTSEC-2025-0098" },
     { id = "RUSTSEC-2025-0100", reason = "unic-* unmaintained (kuchikiki/selectors chain). https://rustsec.org/advisories/RUSTSEC-2025-0100" },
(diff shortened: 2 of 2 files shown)


houko (commented):
Two things to address before merging:

**Missing CHANGELOG.md.** The PR description lists a `CHANGELOG.md` entry under `### Fixed` / `[Unreleased]`, but the diff only contains `Cargo.lock` and `deny.toml`. The pre-commit hook enforces `(@user)` attribution on any staged addition to `[Unreleased]`, so the entry needs to land in this PR (not a follow-up) to keep the changelog accurate.

**Overlap with dependabot PR #6362.** That PR also bumps `anyhow` 1.0.102 → 1.0.103 (same lockfile hunk). Whichever lands second will need a rebase; the maintainer may want to close #6362 once this one merges, or merge #6362 first and rebase this PR to just the `deny.toml` `RUSTSEC-2026-0192` ignore entry (which #6362 does not include).

The `deny.toml` entry itself is correct — format and reason string match the existing pattern.

---

PR #6386, opened 2026-07-04
docs: add Arch Linux pacman installation instructions

pavver:
## Summary

- Add Arch Linux installation instructions next to the existing Homebrew
  instructions in the main README, Getting Started guide, and CLI reference.
- Document the official signed LibreFang pacman repository, package-signing
  key, `/etc/pacman.conf` entry, and CLI/desktop installation commands.
- Explain that the project-maintained pacman repository is currently used
  because AUR account registration is temporarily unavailable.
- Document that `librefang-desktop-bin` is available on x86_64 only and point
  aarch64 users to `librefang-bin` and the web dashboard.
- Keep the Chinese documentation and all translated READMEs in sync with the
  English installation instructions.

This PR does not change the documentation toolchain.


houko (changes requested):
Thanks for the contribution — the direction is right, and the substance checks out: all repo links resolve, `<Note>` is globally registered for the docs MDX pages, and I verified that the key actually served at `https://packages.librefang.ai/librefang.gpg` carries exactly the pinned fingerprint `2C32 5B0F 8870 6ED9 9C45 E216 DD09 DC7D 3E70 E1E9` (uid `LibreFang Packaging <packaging@librefang.ai>`).
Requesting changes for one user-visible bug plus a few consistency items, all within this PR's docs scope.

**1. Desktop-only install path never syncs the new repo database (must fix)**

`docs/src/app/getting-started/page.mdx` (desktop section, ~line 219) and the zh counterpart say "After configuring the official repository above:" followed by `sudo pacman -S librefang-desktop-bin`.
A reader who only wants the desktop app performs the key import and the `pacman.conf` edit but never runs the CLI section's `pacman -Syu librefang-bin`, so the `[librefang]` database has never been downloaded and pacman fails with `error: database file for 'librefang' does not exist (use '-Sy' to download)`.
Fix: use `sudo pacman -Syu librefang-desktop-bin` in this section (both EN and zh pages).

**2. Make `packaging/arch-repo/README.md` the authoritative source for the fingerprint**

The fingerprint is now pinned 26 times across 13 files, but `packaging/arch-repo/README.md` — the doc every copy links to as the reference — does not contain it and prescribes a different verification flow (`pacman-key --finger packaging@librefang.ai`, confirm the printed value, then lsign).
So the canonical doc and the 13 copies diverge from day one, and a future key rotation has no single place to sync from.
Please also add the fingerprint to `packaging/arch-repo/README.md` in this PR and align its instructions with the pinned-fingerprint flow, so the canonical doc is the one authoritative statement.

**3. aarch64 guidance omits `librefang-docker`**

`packaging/arch-repo/README.md` lists `librefang-docker` as available on aarch64 and notes that `pacman -S librefang-desktop-bin` there reports "target not found".
The aarch64 notes added in this PR only mention `librefang-bin` plus the web dashboard; please also mention `librefang-docker` as the supported way to run the service on aarch64.
The combined README/i18n code block will error on its second line for Arch Linux ARM users who paste the whole block — the `# x86_64 only` comment mitigates this, so no change required there, but it is worth knowing.

**4. Prose wrapping convention**

Repo convention (root `CLAUDE.md`, "Prose wrapping" section): new Markdown prose is not hard-wrapped at a column count; lines break only at sentence boundaries, one sentence per line.
Several added paragraphs wrap mid-sentence at ~72–77 columns, e.g. `README.md` lines 80–81 (break after "currently") and `docs/src/app/getting-started/page.mdx` lines 223–224 (break after "install").
Please reflow the added prose (all 13 files) to one sentence per line.

houko (approved):
All four items from the previous review round are resolved.
Every install command across the 13 files now uses `pacman -Syu`, so the desktop-only path syncs the repo database correctly.
The canonical `packaging/arch-repo/README.md` now pins the fingerprint and its verification flow matches the copies; `--lsign-key <full fingerprint>` also keeps the security semantics of the old compare-then-sign flow.
The aarch64 guidance now covers `librefang-docker`, which I confirmed against `publish-arch-repo.sh` (published for both x86_64 and aarch64).
Added prose breaks only at sentence boundaries.

Also verified: the fingerprint is byte-identical in all 28 occurrences, both docs-site anchors resolve (h3 headings, covered by `rehypeSlugify`), the i18n blocks stay in sync with the English copy, splitting the README/i18n install commands into separate code blocks removes the paste-the-whole-block failure on aarch64, and the CHANGELOG entry follows the existing `### Documentation` section convention.

Two non-blocking nits, noting them for the record:

- `packaging/arch-repo/README.md`: the three new install sections restate the arch-availability matrix that the pre-existing "Available packages" list (and the aarch64 note below it) already carries, so the canonical doc now states the same facts three times.
- `docs/src/app/getting-started/page.mdx` (desktop section): capitalization drift between adjacent lines — "The Desktop package …" vs "The desktop package …".

PR #6397, opened 2026-07-07
fix(channels): convert Markdown to Slack mrkdwn in the Slack sidecar

### LINKED ISSUE #6396: bug(channels): Slack sidecar posts raw Markdown — owns_formatting()=true but slack.py has no Markdown→mrkdwn converter
## Summary

Since #5796, sidecar adapters own outbound formatting: the sidecar trampoline returns `owns_formatting() = true` (`crates/librefang-channels/src/sidecar.rs:1915`), so `send_response` skips the kernel-side `format_for_channel` and hands the sidecar **raw agent Markdown** (`crates/librefang-channels/src/bridge.rs:2922`). Telegram and Matrix honour their side of that contract (`telegram.py: markdown_to_telegram_html`, `matrix.py: markdown_to_matrix_html`). **The Slack sidecar never got a converter** — `slack.py: on_send()` posts `cmd.text` verbatim — so agent replies arrive in Slack with literal Markdown.

## Symptom

Agent messages to Slack show raw Markdown instead of Slack mrkdwn:

- `**bold**` → literal `**bold**` (Slack bold is `*bold*`)
- `## Heading` → literal `## Heading` (Slack has no headings)
- `- item` bullets, `[text](url)` links, and `| a | b |` tables all render raw

## Root cause

`sdk/python/librefang/sidecar/adapters/slack.py` — `on_send()` (~L857) sends `text = cmd.text or ""` straight to `_post_message()` with no Markdown→mrkdwn step, on both the plain-text and interactive paths. Compare `telegram.py` / `matrix.py`, which run a converter on the same path. Slack's own `_build_block_kit()` already declares its section text as `{"type": "mrkdwn"}`, confirming mrkdwn is the intended wire format — the conversion is simply absent.

## Expected

The Slack sidecar converts agent Markdown to Slack mrkdwn before posting, matching the post-#5796 "sidecar owns formatting" contract that Telegram/Matrix already implement.

## Notes

Slack mrkdwn differs from Markdown: bold is `*…*` (not `**…**`), links are `<url|text>`, strikethrough is `~…~`, and there is **no** heading or table syntax (headings are conventionally rendered as bold, tables as a fixed-width code block). Code spans (`` ` `` / ```` ``` ````) already render natively and should be preserved verbatim.

This is distinct from #1875 (the `channel_send` *tool* path); the bug here is the agent-reply `on_send` sidecar path.

## Environment

Reproduced on current `main`. Files: `sdk/python/librefang/sidecar/adapters/slack.py`, `crates/librefang-channels/src/{sidecar.rs,bridge.rs}`.

neo-wanderer:
## Type

- [x] Channel adapter
- [x] Bug fix

## Summary

Fixes #6396.

Since #5796, sidecar adapters own outbound formatting: the sidecar trampoline returns `owns_formatting() = true` (`crates/librefang-channels/src/sidecar.rs`), so `send_response` skips the kernel-side `format_for_channel` and hands the sidecar **raw agent Markdown** (`crates/librefang-channels/src/bridge.rs`). Telegram and Matrix implement their side of that contract (`telegram.py: markdown_to_telegram_html`, `matrix.py: markdown_to_matrix_html`). The Slack sidecar never got a converter — `slack.py: on_send()` posted `cmd.text` verbatim — so agent replies reached Slack with literal `**bold**`, `## headings`, `| tables |`, etc.

This adds the missing Markdown → Slack-mrkdwn conversion in the Slack sidecar, matching the established post-#5796 contract. `slack.py`'s `_build_block_kit()` already declares its section text as `{"type": "mrkdwn"}`, so mrkdwn is the intended wire format.

## Changes

- `sdk/python/librefang/sidecar/adapters/slack.py`: add `_markdown_to_mrkdwn()` (plus small helpers) and apply it in `on_send()` on both the text and interactive send paths:
  - `**bold**` / `__bold__` → `*bold*`
  - `#` .. `######` headers → bold (Slack has no headings)
  - `-` / `*` / `+` bullets → `•`; `[text](url)` → `<url|text>`; `~~strike~~` → `~strike~`
  - GitHub-style tables → aligned monospace code block (Slack has no table syntax)
  - fenced/inline code, `_italic_`, and `> quotes` pass through untouched (already valid mrkdwn)
- `sdk/python/tests/test_slack_adapter.py`: unit tests for the converter (bold / headers / bullets / links / strikethrough / table / code-preservation / italic+quote passthrough) and an `on_send` integration test asserting the posted body is converted.

## Attribution

- [x] Original work; no adapted prior code.

## Testing

- [x] `python -m pytest sdk/python/tests/test_slack_adapter.py` — converter unit tests and the `on_send` integration test pass (`asyncio_mode = auto`).
- [ ] `cargo clippy --workspace ...` — n/a, no Rust changed.
- [ ] `cargo test --workspace` — n/a, no Rust changed.

## Security

- [x] No new unsafe code
- [x] User input validated at boundaries — the converter is a pure string transformation and preserves code spans verbatim


sdk/python/librefang/sidecar/adapters/slack.py
```diff
@@ -922,6 +925,130 @@ async def on_send(self, cmd) -> None:
...
+    but sidecar adapters report `owns_formatting() = true`, so the kernel
+    now sends raw text and the adapter owns the conversion. Handles bold,
+    headers, bullets, links, and strikethrough — the cases that otherwise
+    leak literal `**` / `##` into Slack. Italic and blockquotes already
+    share syntax with Slack (`_x_`, `> q`) and pass through untouched;
+    fenced/inline code is preserved verbatim (Slack renders backticks).
+    """
+    if not text:
+        return text
+    parts = _CODE_SPLIT_RE.split(text)
```
houko: Splitting on code spans before applying inline Markdown means any inline construct that *wraps around* an inline code span silently fails to convert, leaking literal `**`/`[...]`  into Slack. Repro against this branch's `_markdown_to_mrkdwn`:

```
>>> _markdown_to_mrkdwn("**the `foo()` function** is broken")
'**the `foo()` function** is broken'   # bold markers leak through verbatim
>>> _markdown_to_mrkdwn("run `pip install x` then **restart**")
'run `pip install x` then *restart*'   # works only because bold doesn't cross the code boundary
```

Bold-wrapping-inline-code ("**the `get_user()` method**") is a fairly common pattern in LLM-generated text, so this isn't just a theoretical edge case.

Separately, header/bullet lines that contain inline code lose their line-level formatting after the first code span, e.g. `# Title \`code\` more` → `*Title*\`code\` more` instead of `*Title \`code\` more*` (cosmetic, but visibly wrong).

Root cause: `_CODE_SPLIT_RE.split(text)` chunks the whole message by code delimiters first, then each non-code chunk is converted independently — so any bold/link span or header/bullet line that straddles a code delimiter is only ever partially visible to the regexes. Fixing this well probably needs the code-span protection to happen per-match inside the inline substitutions (e.g., a placeholder/token swap for code spans before running the line-level and inline conversions, then restoring them after) rather than a top-level segment split. Flagging as a design question rather than pushing a fix, since it touches the core algorithm and I don't want to guess at the desired output for every edge case (nested spans, code spans that break across a header/bullet marker, etc.) — let me know if you'd like me to take a pass at reworking `_markdown_to_mrkdwn`.

---

houko (commented):
Otherwise the conversion coverage and test suite look solid — one design-level formatting gap flagged inline.

---

sdk/python/librefang/sidecar/adapters/slack.py
```diff
@@ -922,6 +925,130 @@ async def on_send(self, cmd) -> None:
...
+    """Convert common Markdown to Slack mrkdwn, leaving code spans intact.
+
+    The Rust `formatter::markdown_to_slack_mrkdwn` used to run kernel-side,
+    but sidecar adapters report `owns_formatting() = true`, so the kernel
+    now sends raw text and the adapter owns the conversion. Handles bold,
+    headers, bullets, links, and strikethrough — the cases that otherwise
+    leak literal `**` / `##` into Slack. Italic and blockquotes already
+    share syntax with Slack (`_x_`, `> q`) and pass through untouched;
+    fenced/inline code is preserved verbatim (Slack renders backticks).
+    """
```
houko, line 1057: CLAUDE.md compliance nit (prose wrapping, not a correctness issue): this docstring is hard-wrapped at a fixed column instead of breaking at sentence boundaries.

Repo rule: "The only legitimate line break inside a paragraph is at a complete sentence boundary... One sentence = one line, regardless of length... [applies to] Source code doc-comments (`//!`, `///`, `#`-style docstrings, ...) and any multi-line free-form prose comment block."

Here the sentence "The Rust `formatter::markdown_to_slack_mrkdwn` used to run kernel-side, but sidecar adapters report `owns_formatting() = true`, so the kernel now sends raw text and the adapter owns the conversion." is split across lines 1036-1038, and the closing sentence spans 1041-1042. Each should be a single line (however long), per the rule.

Same pattern recurs at several other spots this PR added, worth a single pass while touching this file:
- lines 877-878 (`on_send` comment)
- lines 928-930 (comment above `_CODE_SPLIT_RE`)
- lines 938-940 (comment above `_TABLE_DIVIDER_RE`)
- lines 969-970 (`_render_table` docstring)
- lines 1003-1004 (table comment inside `_convert_md_segment`)
- lines 1021-1022 (bullets comment)

---

houko (commented):
CLAUDE.md-compliance pass (not re-reviewing the conversion algorithm — the existing review already flagged the bold/inline-code-boundary correctness gap in `_markdown_to_mrkdwn`, which I agree is the substantive design question here).

One item worth a fixup pass: several of the new comments/docstrings in this diff hard-wrap prose at a fixed column instead of breaking at sentence boundaries, which the repo's CLAUDE.md calls out explicitly for doc-comments and free-form comment blocks. Left one inline note with the specific lines. No functional/security concerns beyond what's already flagged on this PR; commit messages are already in conventional-commit format.

---

sdk/python/librefang/sidecar/adapters/slack.py
neo-wanderer: Fixed in 2f29b7041 using the placeholder approach you sketched.
`_markdown_to_mrkdwn` now masks every code span (fenced or inline) with a `\x00<index>\x01` token before any other rule runs, converts the whole masked text (tables, headers, bullets, inline), and restores the spans at the end — the top-level `_CODE_SPLIT_RE.split` is gone.
Both repros now convert correctly: `` **the `foo()` function** is broken `` → `` *the `foo()` function* is broken ``, and `` # Title `code` more `` → `` *Title `code` more* `` (header stays one bold span).
Table cells restore their tokens with backticks stripped before column widths are computed (`_plain_cell` / `_restore_code_spans(strip_backticks=True)`), so the monospace grid still aligns on the real cell text.
Delimiter bytes are dropped from the input up front, so a pathological message can never alias a minted token.
New regression tests: `test_mrkdwn_bold_wrapping_inline_code`, `test_mrkdwn_bold_wrapping_multiple_code_spans`, `test_mrkdwn_bold_after_code_span`, `test_mrkdwn_header_with_inline_code_stays_one_bold_span`, `test_mrkdwn_bullet_with_inline_code`, `test_mrkdwn_link_text_with_inline_code`, `test_mrkdwn_strike_wrapping_inline_code`, `test_mrkdwn_table_cell_with_inline_code_stays_aligned`.
All 95 tests in `test_slack_adapter.py` pass (1892 across the SDK suite), including the pre-existing `test_mrkdwn_preserves_code_spans` verbatim-code guarantee.

sdk/python/librefang/sidecar/adapters/slack.py
neo-wanderer: Fixed in 2f29b7041 — all listed comment blocks and docstrings in the new code are now wrapped at sentence boundaries (one sentence per line), including the converter docstring, the regex-constant comments, and the new test comments this fixup added.
The pre-existing `_build_block_kit` docstring keeps its old wrapping per the "don't rewrap untouched paragraphs" clause.

sdk/python/librefang/sidecar/adapters/slack.py
```diff
@@ -922,6 +924,151 @@ async def on_send(self, cmd) -> None:
...
+            i = j
+            continue
+        # ATX headers (# .. ######) → bold (Slack has no headings).
+        m = _ATX_HEADER_RE.match(line)
+        if m:
+            content = m.group(2).strip()
+            out.append(f"{m.group(1)}*{_inline_md(content)}*" if content else "")
+            i += 1
+            continue
+        # Bullets (-, *, +) → •. Also stops a "* item" bullet from being mistaken for italic.
```
houko, line 1038: One leftover instance of the prose-wrap issue from the earlier review pass.
This single comment line crams two sentences together: "Bullets (-, *, +) → •." and "Also stops a \"* item\" bullet from being mistaken for italic."
CLAUDE.md's prose-wrapping rule requires one sentence per line in comments, with no exception for short sentences.
Suggested fix:
```python
        # Bullets (-, *, +) → •.
        # Also stops a "* item" bullet from being mistaken for italic.
```
I applied this exact change on my end and tried to push it to this branch (`git push` to `neo-wanderer/librefang:fix/slack-sidecar-mrkdwn`), but the push was rejected with a 403 — I don't have write access to this fork.
Leaving it here so it can be applied directly.

---

sdk/python/librefang/sidecar/adapters/slack.py
```diff
@@ -922,6 +924,151 @@ async def on_send(self, cmd) -> None:
...
+        if b:
+            out.append(f"{b.group(1)}•   {_inline_md(b.group(2))}")
+            i += 1
+            continue
+        out.append(_inline_md(line))
+        i += 1
+    return "\n".join(out)
+
+
+def _markdown_to_mrkdwn(text: str) -> str:
```
houko, line 1049: Design question, flagging for human judgment rather than fixing myself.
Slack's mrkdwn format requires literal `&`, `<`, and `>` in the text body to be HTML-entity-escaped (`&amp;`, `&lt;`, `&gt;`) before sending, otherwise a raw `<` can be parsed as the start of one of Slack's own special sequences (`<http://url>`, `<@USERID>`, `<!channel>`, etc).
`_markdown_to_mrkdwn` does not do this escaping anywhere in the pipeline, so an agent reply that happens to contain a literal `<@U012ABC>`-shaped substring, or a URL with an unescaped `&` in the query string, could render unexpectedly or ping an unintended user/channel once it reaches Slack.
This gap predates this PR (the old code sent `cmd.text` verbatim too), so it may be intentionally out of scope for a Markdown-conversion fix, but this PR is the first place actually assembling the final `mrkdwn` string, and CLAUDE.md's collaboration rules ask that in-scope fixes not be punted to a follow-up when they sit in the same file/domain already being touched.
Worth deciding explicitly: escape here (careful not to double-escape the `<url|text>` / `<@…>` sequences this converter itself generates for links), escape at the `on_send` call site instead, or confirm it's deliberately deferred and note why.

---

houko (commented):
Scheduled CLAUDE.md-compliance pass, scoped to the delta since the last review round (commit 2f29b704 already addressed the previously flagged bold/inline-code-boundary gap and most of the prose-wrap cleanup).
Two inline notes: one leftover two-sentences-on-one-line comment I fixed directly (push blocked by fork permissions, described inline with the exact diff), and one design question about Slack mrkdwn entity-escaping that needs a human decision rather than a mechanical fix.
One additional finding that isn't line-anchorable: the body of commit 55cdfb6 ("fix(channels): convert Markdown to Slack mrkdwn in the Slack sidecar") hard-wraps prose at a fixed column instead of breaking at sentence boundaries, e.g. "Post-#5796, sidecar adapters own outbound formatting (owns_formatting()" / "returns true in sidecar.rs), so send_response hands the sidecar raw agent" / "Markdown." split across three lines mid-sentence.
CLAUDE.md's prose-wrapping rule explicitly covers commit message bodies.
Fixing it would mean rewording a non-HEAD commit and force-pushing to rewrite this branch's history, which is not something I'll do on someone else's fork branch — leaving it here for the maintainer/author to decide whether it's worth a rebase before merge.
The second commit (2f29b704) already follows the one-sentence-per-line convention correctly in its own body.
Test suite (`python -m pytest sdk/python/tests/test_slack_adapter.py`, pytest 9.1.1 / Python 3.11.15) passes: 95 passed, including after my inline-comment fix.

---

Issue #6242, opened 2026-06-19
Proposal: Active Tool Output Optimization & Diagnostic Log Offloading


pavver:
## Background & Motivation

In the current implementation of LibreFang, the context size is kept in check primarily via the *compaction pass*, which runs periodically or when a threshold is met. However, developer tools (like `cargo test`, compiler builds, or dependency installers) can produce massive outputs (often exceeding 100 KB or 25,000+ tokens) in a single turn. 

If an agent runs `cargo test` on step 2, that massive output is immediately fed into the prompt for step 3, step 4, and so on, until a compaction pass eventually aggregates it. This causes:
1. **High Latency & Costs:** The model immediately incurs high prompt-processing latency and cost on the very next turns.
2. **Context Clutter:** The model's attention is filled with hundreds of repetitive `ok` lines, diluting the focus on the task.

We propose introducing **Active Tool Output Optimization**—a hot-path mechanism that processes and minimizes tool outputs *before* they enter the LLM context, while preserving full diagnostic data for troubleshooting.

---

### 1. Smart Output Pruning & Diagnostic Offloading (Recommended)
Instead of putting the entire stdout/stderr of a command into the conversation history, we offload the raw output to a session-specific diagnostic file.
* **Mechanism:**
  * The full raw output of the tool is written to a local log file, e.g., `.librefang/sessions/<session_id>/tool_logs/exec_<N>.log`.
  * The message appended to the LLM's context contains only a highly-compressed summary (e.g., a few lines) and a structured link pointing to the full log.
* **On-Demand Inspection:** 
  Since the agent is already equipped with file-reading and search tools (`view_file`, `grep_search`), it can inspect the full log *only when needed* (e.g., if a build fails or if it needs to check a specific warning).
* **Example Context Message:**
  ```
  Exit code: 0
  [Full stdout/stderr (1,240 lines) offloaded to file:///home/pavver/.librefang/sessions/abc-123/tool_logs/exec_4.log]
  Use `view_file` or `grep_search` on this path if you need to read the full logs or search for specific errors.
  ```

### 2. Output Line Filtering & Smart Parsing
We can implement lightweight output processors that scan stdout and keep only the "essential" lines based on simple rules:
* **For Successful Executions (`exit_code == 0`):**
  * Scan the output and keep only summary lines (e.g., lines starting with `test result:`, `Finished`, or showing test counts), dropping individual successful test lines (`test ... ok`).
* **For Failed Executions (`exit_code != 0`):**
[...]

PR #5149, opened 2026-05-16
fix(runtime): spill oversized MCP/tool results to artifact store before truncation

neo-wanderer:
## Problem

MCP tool results arrived at `execute_single_tool_call_inner` with no artifact spill applied, unlike `web_fetch`/`web_search` which spill at execution time inside their own runners. The shared chokepoint ran `sanitize_tool_result_content` (destructive truncation) **before** `ToolBudgetEnforcer::maybe_persist_result`, so:

1. Large MCP result arrives raw (e.g. 54 KB)
2. `sanitize_tool_result_content` truncates it to fit the context window (~12 KB)
3. Truncated text is now well below the 16 KiB spill threshold → no artifact created
4. LLM receives clipped output with **no `read_artifact(...)` recovery reference**
5. Original bytes are permanently lost

## Fix

Spill the full raw result to the artifact store inside the shared chokepoint **before** `sanitize_tool_result_content`, uniformly for all tools (MCP included). This covers both the streaming and non-streaming loops.

Already-compact results (web stubs already under 16 KiB) pass through `maybe_spill` as a no-op — no double-spill.

**Changes:**

- `tool_runner/spill.rs`: promote `resolve_spill_config` to `pub(crate)` (no duplicated constants)
- `tool_runner/mod.rs`: re-export `resolve_spill_config` at the crate boundary
- `agent_loop/tool_call.rs`: spill before truncation in `execute_single_tool_call_inner`
- `agent_loop/tests/sender.rs`: two ordering-invariant tests

## Verification

- `cargo check -p librefang-runtime` — clean
- `cargo clippy -p librefang-runtime --all-targets -- -D warnings` — zero warnings
- `cargo test -p librefang-runtime --lib` — 1689 passed, 1 pre-existing flaky failure (`plugin_runtime::overflow_kills_child`, `Text file busy` OS error in shared CI environment, unrelated to this change)
- Both new tests pass: `oversized_tool_result_spills_before_sanitize_preserves_full_bytes`, `small_or_already_spilled_result_passes_through_without_double_spill`

## Porting note

This rebases cleanly on `upstream/main` after the `#3710` god-crate split (`agent_loop.rs` → `agent_loop/tool_call.rs`, `tool_runner.rs` → `tool_runner/spill.rs`). The original fix was developed on the neo-wanderer fork (neo-wanderer/librefang#13) against a pre-split base; this PR ports it to the split module layout.

PR #6406, opened 2026-07-08
fix(runtime): never re-spill read_artifact results at the post-tool chokepoint

### LINKED ISSUE #6388: read_artifact output is re-spilled into a new artifact → oversized artifacts can never be paged in (infinite loop)
### Summary
When a tool result is spilled to the artifact store, calling `read_artifact` to page it back returns raw bytes that are themselves run through the generic post-tool spill. If the read exceeds `[tool_results] spill_threshold_bytes`, the `read_artifact` result is **re-spilled into a brand-new artifact** and the agent receives a fresh stub with a new handle. Reading *that* handle spills again → a new same-size artifact → forever. An artifact larger than the spill threshold is effectively unreadable.

### Repro (synthetic)
1. Any tool produces output larger than `spill_threshold_bytes` (default **16 KiB**, `librefang-types/src/config/types.rs:7343`) → spilled to `sha256:AAA…`, agent gets a stub.
2. Agent calls `read_artifact(handle="sha256:AAA…", length=65536)`. `length` max is 64 KiB (`MAX_READ_LENGTH`, `librefang-runtime/src/artifact_store.rs:26`) and the tool schema advertises `max 65536` (`librefang-runtime/src/tool_runner/definitions.rs:1234`).
3. The read returns e.g. ~18 KiB from `offset=0` — **correct**.
4. But that ~18 KiB result is > 16 KiB, so the generic spill at `librefang-runtime/src/agent_loop/tool_call.rs:1173-1189` re-spills it into `sha256:BBB…` and returns a new stub.
5. Agent reads `BBB…` → ~18 KiB → re-spilled to `CCC…` → …

Every iteration is a *different* handle read fully from `offset=0`, so from the agent's view "`read_artifact` always reads from offset 0 and never makes progress." The only escape is to stop using `read_artifact` and re-fetch the source in sub-threshold chunks.

### Root cause
`agent_loop/tool_call.rs:1173-1189` applies `artifact_store::maybe_spill` to **every** tool result. The code comment (`tool_call.rs:1167-1172`) reasons that a web stub is already below threshold so there's no double-spill — but a `read_artifact` result is not a stub; it's raw artifact bytes and routinely exceeds the threshold. The constants guarantee the collision: read cap (64 KiB) > spill threshold (16 KiB), so any read over 16 KiB re-spills. This path was introduced alongside the spill-on-every-result work in #6242.

### Suggested fix
Exempt `read_artifact` from the re-spill in `tool_call.rs:1173`:
```rust
let result_content = if tool_call.name == tool_name::READ_ARTIFACT {
    result_content // never re-spill a read_artifact result — it loops
} else {
    // existing maybe_spill(...)
};
```
Plus a regression `#[tokio::test]` that reads a > threshold artifact and asserts the result is returned verbatim (not converted to a new spill stub).

houko:
Fixes #6388.

## Summary

Fresh `read_artifact` output was re-spilled back into the artifact store, so paging in a spilled result larger than the active threshold produced a fresh stub that again pointed at `read_artifact` — a self-referential loop.
The loop existed at three independent gates; this PR exempts the page-in tool at all of them via one shared predicate (`tool_budget::respill_allowed`):

1. **Post-tool chokepoint** (`agent_loop/tool_call.rs`) — the original #6388 site; any read in the 16–64 KiB range looped here under the default 16 KiB `spill_threshold_bytes`.
2. **Per-result budget, Layer 2** (`tool_budget::maybe_persist_result`, both loop drivers) — runs on the same fresh content with the same threshold; an operator configuring `spill_threshold_bytes` below the ~10 KB sanitize cap recreated the loop here. The method now takes the tool name.
3. **Per-turn aggregate budget, Layer 3** (`tool_budget::enforce_turn_budget`) — fresh read pages are now the last-resort spill victims: non-exempt results are spilled first, and a page is spilled only when nothing else can bring the turn under budget (full exemption would leave a many-page turn unbounded). `ToolResultEntry` gained a `tool_name` field for this.

Two corrections to the issue as filed, verified against history: the chokepoint came from #5149 (not #6242), and sub-threshold reads (the default `length=4096`) always worked — the trap was that the schema advertises `max 65536`.

## Known limit (pre-existing, disclosed rather than changed)

`sanitize_tool_result_content` unconditionally clamps every tool result to ~10 KB (`strip_tool_result_details`, `session_repair.rs`), appending a `...[truncated from N chars]` note.
So a 64 KiB page-in now survives the spill gates but is still delivered clipped to ~10 KB; the read header and the truncation note together tell the agent the delivered length.
That cap predates this PR (it applied to ≤16 KiB reads all along) and protects the context window; raising or exempting it is a separate design decision — flagging it here for an explicit maintainer call rather than changing it silently.

## Changes

- `tool_budget.rs`: shared `respill_allowed` predicate; `maybe_persist_result(content, tool_use_id, tool_name)`; Layer-3 victim ordering `(exempt, Reverse(size))`; `ToolResultEntry.tool_name`.
- `agent_loop/tool_call.rs`: chokepoint logic extracted into a testable `spill_fresh_result(tool_name, content, cfg, artifact_dir)` used by `execute_single_tool_call_core`; Layer-3 wiring passes `tool_name`; comment corrected (the previous revision mischaracterized `tool_budget` as history compaction — both budget layers run on fresh results of the current turn).
- `agent_loop/{mod,run_streaming}.rs`: Layer-2 call sites pass the tool name.
- `tool_runner/definitions.rs` + `mod.rs`: `tool_name` const module raised to `pub(crate)` and re-exported.

## Tests

- `read_artifact_results_are_exempt_from_the_respill_chokepoint` (rewritten): drives the production `spill_fresh_result` — an 18 KiB page passes through verbatim with no artifact written, and the same bytes under `web_fetch` still spill to a recoverable stub. Reverting the chokepoint wiring now fails this test.
- `layer2_read_artifact_page_is_exempt`: an over-threshold page passes Layer 2 untouched.
- `layer3_read_artifact_page_is_last_resort_victim`: with the read page as the largest result, the smaller non-exempt result is spilled first and the page survives; a page-only over-budget turn still spills (safety valve).
- Existing spill/budget tests updated for the new signatures; all pass.

## Verification

- `cargo test -p librefang-runtime --lib tool_budget` — 11 passed
- `cargo test -p librefang-runtime --lib spill` — 11 passed
- `cargo test -p librefang-runtime --lib read_artifact_results` — 1 passed
- `cargo clippy -p librefang-runtime --all-targets -- -D warnings` — clean


crates/librefang-runtime/src/agent_loop/tool_call.rs
houko, line 1: Commit `c7ad449` ("docs(runtime): unwrap mid-sentence doc comment per CLAUDE.md prose rule") — the commit that touches this file — has git author identity `Claude <noreply@anthropic.com>` (`git log --format='%an <%ae>'` on this branch confirms it; the other commit, `bf678ec`, is authored by `Evan <tonymo2048@gmail.com>`).

CLAUDE.md's "Commit & PR hygiene" section states: "No Claude / Anthropic / AI attribution in commit messages, PR bodies, or issue comments... the rule exists because attribution pollutes `git log` and signals provenance the project does not want to imply." The commit-msg hook only pattern-matches the message body/trailers, so it doesn't catch the author field — but the author identity is exactly the kind of provenance signal the rule is meant to prevent, and it's visible in `git log`, `git blame`, and the GitHub UI.

Since this is an already-pushed commit, fixing it needs a history rewrite (`git commit --amend --author=...` or a rebase) followed by a force-push, which is outside what an automated review pass should do unilaterally. Flagging for the maintainer to re-author (or squash) before merge.

---

houko (commented):
Reviewed the read_artifact respill-chokepoint fix. Code, tests, and verification all check out (cargo check/clippy/fmt/test all clean, matching the PR's own claims). One finding left inline: a commit-author identity issue that needs a maintainer decision rather than an automated fix.

---

PR #6419, opened 2026-07-09
fix(api,dashboard): correlate chat turns with their WS terminal frames

### LINKED ISSUE #6390: bug(dashboard): inline post-turn memory extraction misbinds previous response to a new chat message
## Summary

When an agent has proactive memory (`auto_memorize`) enabled, the post-turn memory-extraction LLM call runs **inline**, before the WebSocket `"response"` terminal event is emitted.
If the user sends a second message in the dashboard chat while that extraction is still running, the first turn's delayed `"response"` binds to the **second** message's chat bubble — so the new prompt shows the previous answer, and the new message's real answer is dropped.
Once extraction has finished, the next message is handled correctly.

## Reproduce

1. Enable proactive memory with `auto_memorize = true` on a chat-facing agent (any `extraction_model` slow enough to take a few seconds — an inline extraction round-trip).
2. Open the dashboard chat, send message `A`, wait for the answer to finish streaming (you see the full text).
3. Immediately — before extraction completes — send message `B`.

**Expected:** `B` is processed; `B`'s bubble shows `B`'s answer.
**Actual:** `B`'s bubble is filled with `A`'s answer ("the last message is sent again"); `B`'s real answer never appears. A message sent *after* extraction completes works normally.

## Root cause

Two independent defects; both are required to produce the exact symptom.

### D1 — backend: post-turn extraction gates the terminal event and the recv loop

`auto_memorize` is `.await`ed inline inside the turn task, after the LLM answer is already complete and already streamed to the client.

- `crates/librefang-runtime/src/agent_loop/end_turn.rs` — `finalize_successful_end_turn` awaits `pm_store.auto_memorize(...).await` inline (around line 322). The turn task does not return until extraction finishes.
- `crates/librefang-api/src/ws.rs` — `handle_text_message` does `handle.await` on that turn task (around line 1279), then emits `typing: stop` (around line 1300) and `"response"` (around line 1364). So the user-facing terminal events fire **only after** extraction.
- `crates/librefang-api/src/ws.rs` — `handle_agent_ws` awaits `handle_text_message(...).await` inline in its receive loop (around line 845), so a message sent during extraction sits unread in the socket buffer until the previous turn (including extraction) fully returns.
- `crates/librefang-api/src/ws.rs` — the `"response"` frame (around line 1363) carries **content only**, with no turn / session / message correlation id.

### D2 — frontend: single per-agent listener, swapped on send, matched by closure only

- `crates/librefang-api/dashboard/src/pages/ChatPage.tsx` — on a new send, the previous turn is force-torn-down (around lines 953-958: `prevTurn.responded = true; prevTurn.cleanup()`), which removes the previous turn's `message` listener.
- The new `handleMessage` (registered around line 1203) is bound to the **new** bubble's closure `botMsg.id` and handles **any** `message` frame — there is no turn/message-id match.
- The `"response"` handler (around lines 1130-1143) patches `botMsg.id` (the new bubble) with the frame content unconditionally.

### The race

1. Message `A`: the answer streams in via `text_delta` (during the LLM loop). Extraction `A` then starts inline; `typing: stop` / `"response"` for `A` have **not** been sent, so bubble `A` is still `isStreaming` and its listener is still live.
2. User sends `B`. Frontend tears down `A`'s listener and installs `B`'s listener on the new bubble. `B` sits in the socket buffer because `handle_agent_ws` is still blocked on `A`'s inline extraction.
3. Extraction `A` finishes; backend emits `A`'s `"response"` (content-only, no id). The only live listener is now `B`'s, so `B`'s bubble is patched with `A`'s answer.
4. `B` is then read and processed, but its listener was already `cleanup()`'d by the `"response"` handler, so `B`'s real answer has no listener and is dropped.
5. A message sent after extraction has settled works, because no pending prior terminal frame races the new listener.

This is why the bug only appears with proactive memory enabled: without `auto_memorize`, the turn task returns as soon as the LLM answer is done, so the terminal `"response"` fires promptly and there is effectively no window for a second message to race it.

## Proposed fix

**Primary (backend) — decouple extraction from turn completion.**
Deliver the user-facing `"response"` first, then run `auto_memorize` as a detached `tokio::spawn` rather than awaiting it inline in `finalize_successful_end_turn`.
Effects: the `"response"` event fires immediately after the LLM answer; `handle_agent_ws` unblocks and reads the next message normally; the race window collapses to ~zero.
Extraction still runs; it just no longer sits on the turn's critical path. (The turn task must keep whatever handles/context the spawned extraction needs alive for its lifetime.)

**Secondary (defense in depth) — correlate terminal frames.**
Include a `message_id` (or turn id) on the `"response"` / `typing` / `silent_complete` frames (`ws.rs`) and have `handleMessage` (`ChatPage.tsx`) ignore any frame whose id does not match its `botMsg.id`.
This closes D2 permanently, so no late or duplicate frame can ever bind to the wrong bubble — independent of extraction timing.

Doing both fully closes the bug; the backend change alone removes the practical symptom, the frontend change removes the latent mis-binding class.

## Impact

Any deployment running the dashboard chat with `auto_memorize` enabled: a follow-up message sent during the post-turn extraction window is silently mis-answered (shows the prior answer) and its own answer is lost, until the user waits out extraction.

houko:
Fixes #6390.

## Summary

With proactive memory (`auto_memorize`) enabled, the post-turn extraction runs inline before the WebSocket `response` terminal frame is emitted (`end_turn.rs` → `ws.rs handle.await`).
A second message sent during that window raced the first turn's delayed `response`: the dashboard replaces its single per-turn socket listener on every send, and the `response` frame carried no turn identity, so turn A's answer was written into turn B's bubble and B's real answer was dropped (full verified chain in [the triage comment](https://github.com/librefang/librefang/issues/6390#issuecomment-4911472424)).

This implements the recommended fix from that triage (turn correlation — "Fix B"): it eliminates the misbinding class outright without touching memory semantics.
Detaching the extraction ("Fix A") is intentionally not done here — it regresses `memories_saved` delivery, the next-turn recall guarantee, and shutdown durability, per the trade-offs in the triage comment.

## How it works

- **Client → daemon**: the chat send frame now carries `message_id` (the client-generated bubble id — no new id scheme, no extra mapping).
- **Daemon** (`ws.rs`): the id is parsed (capped at 128 bytes), and echoed on every terminal frame of that turn — `response`, `silent_complete`, and all five `error` emission sites, including the pre-turn validation errors. Protocol docs at the top of `ws.rs` updated.
- **Dashboard** (`ChatPage.tsx` + `lib/chat.ts`): a foreign-terminal router at the top of the per-turn `handleMessage` — a terminal frame whose echoed id differs from the current bubble's id is **routed to the bubble that owns it** (response patch / silent removal / error mark on that bubble) and never touches the current turn's lifecycle (cleanup, `finishTurnIfCurrent`, session auto-pin).

Post-fix race replay: A's late `response` lands on A's bubble with its full content and memory chips; B's listener, stream, and terminal stay intact; nothing is dropped.

## Compatibility

- Old dashboard + new daemon: no `message_id` sent → nothing echoed → old behavior.
- New dashboard + old daemon: no `message_id` echoed → `terminalFrameOwner` returns `current` → old behavior.
- HTTP fallback path: inherently request/response-correlated, unchanged.

## Tests

- `ws.rs`: `stamp_message_id_echoes_the_turn_correlation_id` (echo + absent-id passthrough).
- `lib/chat.test.ts`: `terminalFrameOwner` (match / foreign / legacy-no-id / non-string) and `isTerminalFrameType` (exact terminal set; streaming frames unaffected).

## Verification

- `cargo test -p librefang-api --lib stamp_message_id` — 1 passed
- `cargo clippy -p librefang-api --lib -- -D warnings` — clean; `cargo fmt --check` — clean
- `cargo check -p librefang-api --lib` — clean
- Dashboard: `pnpm vitest run` — 847 passed (7 new); `pnpm tsc --noEmit` — clean; `eslint` on touched files — no new findings (8 pre-existing react-hooks warnings on untouched lines)

## Review follow-ups (5ee7b51f)

Rebased onto latest `main` (resolves the CHANGELOG `[Unreleased]` conflict — both #6418 and this entry are kept) and addressed both review comments:

- **Foreign-`response` session auto-pin (correctness).** The foreign-frame branch patched the owning bubble but never ran the `shouldAutoPinResolvedSession` gate the current-turn / HTTP paths run. If a still-unpinned chat's first turn resolved its `session_id` on a `response` that arrived as a foreign frame and the newer turn then ended as `silent_complete` / `error`, the resolved id was dropped and the chat stayed unpinned for the tab's life. The branch now runs the same gate against `data.session_id` and seeds the pre-navigation cache identically.
- **Router now unit-tested as a pure function (coverage).** The three foreign-frame transforms (response patch / `silent_complete` removal / `error` mark) are extracted into a pure `applyForeignTerminalFrame(messages, frame)` in `lib/chat.ts`; the hook delegates to it. `chat.test.ts` covers the two-bubble race directly: a foreign `response` patches turn A's bubble and returns turn B's bubble by reference (untouched), `silent_complete` removes only A, `error` marks only A, and an unknown owner is a no-op. This reaches the actual misbind-prevention effect the extracted boolean helpers could not.

Verification (this round): `pnpm typecheck` clean; `pnpm vitest run` — 852 passed (85 files); `eslint` on touched files — no new findings (8 pre-existing react-hooks warnings on untouched lines). No Rust changed this round, so `ws.rs` is byte-identical to the reviewed revision.


crates/librefang-api/dashboard/src/pages/ChatPage.tsx
```diff
@@ -990,6 +990,38 @@ function useChatMessages(
           resetFallbackTimer();
           try {
             const data = JSON.parse(event.data as string);
+            // #6390: a previous turn's terminal frame can land after this turn replaced the socket listener — post-turn memory extraction delays `response` past the next user send.
+            // The daemon echoes the `message_id` we sent, so route a foreign frame to the bubble that owns it and leave this turn's lifecycle (cleanup / finishTurn / auto-pin) untouched.
+            if (
+              isTerminalFrameType(data.type) &&
```
houko, line 996: Per CLAUDE.md's "Common Gotchas" rule — "Always write integration tests at the injection site, not just the implementation site" — the tests added here (`terminalFrameOwner`/`isTerminalFrameType` in `chat.test.ts`) only cover the two pure helpers extracted from this logic, not this foreign-terminal router itself inside `useChatMessages`/`handleMessage`. This is the exact wiring where the original #6390 misbind happened (a second turn's `handleMessage` closure receiving a first turn's late `response`), so it's the highest-value place to regress-test.

`TerminalPage.test.tsx` already establishes a `MockWebSocket` (`emitOpen`/`emitMessage`) + `render()` pattern for exactly this kind of WS-driven page in this same directory. A similar test here — render `ChatPage`, send turn A, send turn B before emitting A's `response` with A's `message_id`, then assert A's bubble (not B's) received the content and B's bubble/stream was untouched — would directly reproduce the #6390 race end-to-end instead of only exercising the extracted pure functions.

I did not write this myself since it requires mocking the surrounding hooks (`useSessionStream`, `useActiveHandsWhen`, etc.) the same way `TerminalPage.test.tsx` does, which is a judgment call on scope/mocking depth I'd rather leave to the PR author.

---

crates/librefang-api/dashboard/src/pages/ChatPage.tsx
```diff
@@ -990,6 +990,38 @@ function useChatMessages(
...
+                        ...m, content: data.content || m.content, isStreaming: false,
+                        tokens: { output: data.output_tokens, input: data.input_tokens },
+                        cost_usd: data.cost_usd,
+                        memories_saved: data.memories_saved,
+                        memories_used: data.memories_used,
+                        thinking: typeof data.thinking === "string" ? data.thinking : m.thinking,
+                        thinkingCollapsed: m.thinkingCollapsed ?? true,
+                      }
+                    : m
+                ));
```
houko: This foreign-`response` branch patches content/tokens/`memories_saved` onto the owning bubble but never runs the `shouldAutoPinResolvedSession` auto-pin logic that the current-turn path runs at line ~1202 (keyed off `data.session_id`).
That's fine when the *current* (newer) turn is the one that eventually resolves and pins the session, but consider the exact race this PR targets: user sends message A into a brand-new, unpinned chat (no `?sessionId=` yet), gets `typing:stop`, immediately sends message B before A's terminal frame lands.
If A's `response` is the one carrying the newly-resolved `session_id` (the first turn is usually the one that creates the session), that id is now silently dropped — this branch never reads `data.session_id` at all.
The URL only ends up pinned if turn B's own `response` also happens to carry (and re-resolve) a session id; if B ends up `silent_complete` or `error` instead, the chat can be left permanently unpinned for the rest of the tab's life even though the daemon already told the client the resolved session id once.
Is this an accepted tradeoff (per the comment above at line 994, auto-pin is intentionally scoped to "this turn" only), or should the foreign-response branch also read `data.session_id` and call `shouldAutoPinResolvedSession`/`onAutoPinSession` for `foreignId`'s own turn?

---

houko (commented):
Reviewed the #6390 correlation-id diff for CLAUDE.md compliance (prose-wrapping, dashboard data-layer rules, deterministic-ordering) and sanity-checked the compatibility/correlation claims.

Fixed directly (pushed to this branch): a prose-wrap violation — one new sentence in ws.rs was split across two comment lines, contrary to the repo's one-sentence-per-line convention.

Left one inline comment on a genuine (but ambiguous-severity) correctness question in the foreign-terminal-frame router: the `response` branch skips the session-auto-pin logic that the current-turn path runs, which can leave a brand-new unpinned chat's session id unresolved in a specific ordering of the race this PR fixes.

No dashboard data-layer violations found — the new WS message routing reuses the existing `updateAgentMessages` local-state pattern already used throughout this file for other frame types, not a query/mutation-hook bypass. No new routes, query keys, or HashMap/prompt-ordering concerns in this diff.

---

crates/librefang-api/dashboard/src/pages/ChatPage.tsx
houko: Good catch — this was a real drop, now fixed in 5ee7b51f.

The foreign-`response` branch in `useChatMessages` (ChatPage.tsx) now runs the same `shouldAutoPinResolvedSession` gate the current-turn and HTTP paths run, keyed off `data.session_id`, and seeds the pre-navigation cache the same way before calling `onAutoPinSession`.
So the exact race you described — A's `response` (the first turn, which creates the session) arriving as a foreign frame while B is in flight, then B ending as `silent_complete` / `error` — now pins from A's frame instead of leaving the chat unpinned for the tab's life.
The gate's own guards still hold: it only pins when the chat is still unpinned (`currentSessionId === null`, `urlSessionId` falsy) and the agent is unchanged, so a foreign frame never overrides a session the user pinned manually in the meantime.

crates/librefang-api/dashboard/src/pages/ChatPage.tsx
houko: Addressed in 5ee7b51f, though not via a full `ChatPage` render test — here's the trade-off I made.

I extracted the three foreign-frame transforms (response patch / `silent_complete` removal / `error` mark) out of `useChatMessages` into a pure `applyForeignTerminalFrame(messages, frame)` in `lib/chat.ts`; the hook now just delegates to it.
`chat.test.ts` covers the two-bubble race directly against that function: a foreign `response` patches turn A's bubble and returns turn B's bubble *by reference* (proving B's content and stream are never rewritten), `silent_complete` removes only A, `error` marks only A and defaults the message, and an unknown owner is a no-op.
That is the actual misbind-prevention effect the extracted boolean helpers couldn't reach, tested without depending on the surrounding hooks.

I deliberately did not mount `ChatPage` end-to-end: `useChatMessages` is not exported, the page is ~3.6k lines behind ~15 query/mutation/stream hooks, and a render test would be brittle relative to what it buys.
The residual sliver a mounted test would cover and this one does not: that the hook *returns early* after routing a foreign frame (skipping `finishTurnIfCurrent` / `cleanup`) and *calls* the auto-pin branch — three visible lines guarded by the already-tested `terminalFrameOwner` / `isTerminalFrameType`.
If you would rather have that wiring locked down end-to-end despite the brittleness, say so and I will add the `MockWebSocket` + render test modeled on `TerminalPage.test.tsx`.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project uses [Calendar Versioning](https://calver.org/) (YYYY.M.DD).
 ### Fixed
 
 - Uppercase the character after each separator when deriving the version-pinned Homebrew tap formula's class name, matching Homebrew's `Formulary.class_s`, so a pre-release tag yields `class LibrefangAT2026710Beta1` — the exact casing the `golden_gate` audit requires; #6416's `tr -cd '[:alnum:]'` removed the hyphen but left the channel word lowercase (`...beta1`), which `golden_gate` still rejects, so the next `-beta` / `-rc` release would have generated a versioned formula that fails `brew readall` / install (verified by running the real generation step for a stable and a beta tag → `brew readall` reports zero errors) (#6418) (@houko)
+- Correlate dashboard chat turns with the WebSocket terminal frames that own them: the client now sends a `message_id` on every chat message and the daemon echoes it on `response` / `silent_complete` / `error`, so a previous turn's late terminal frame — delayed past the next user send by post-turn proactive-memory extraction — lands on its own bubble instead of overwriting the newer message's bubble and dropping the newer answer (#6419) (@houko)
```
houko, line 19: Not about this line specifically, but flagging against the branch as a whole since there's no better diff anchor: three commits on this branch (`da31ba1`, `0f0fde0`, `0d34d6c`) have their git **author** identity set to `Claude <noreply@anthropic.com>` (committer is `Evan <tonymo2048@gmail.com>` on all three). The commit *messages* are clean — this isn't the "Co-Authored-By: Claude" footer the `commit-msg` hook scans for — but the author field itself is unambiguous AI attribution baked into permanent history, which the repo's "no Claude/Anthropic/AI attribution anywhere (commits, comments)" rule reads as covering more than just message text.

Since these are already pushed and this PR has been through one review round, fixing this means rewriting authorship on those three commits (e.g. `git rebase -i` with `--exec 'git commit --amend --no-edit --author="Evan <tonymo2048@gmail.com>"'`) and force-pushing this branch — something I'm not doing unilaterally on someone else's PR. Worth a maintainer decision before merge.

---

crates/librefang-api/src/ws.rs
```diff
@@ -2147,6 +2191,22 @@ mod tests {
         let _ = VerboseLevel::Off;
     }
 
+    /// #6390: terminal frames echo the client's `message_id` so the dashboard can bind a late frame to the turn that owns it.
+    /// Without an id the frame stays untouched — pre-correlation clients see the old shape.
+    #[test]
+    fn stamp_message_id_echoes_the_turn_correlation_id() {
```
houko, line 2200: Per CLAUDE.md's integration-testing mandate, a route/wiring change should get a `#[tokio::test]` against `TestServer` in `crates/librefang-api/tests/`, in addition to unit coverage. This PR's Rust-side coverage is only `stamp_message_id_echoes_the_turn_correlation_id`, a unit test on the pure helper — there's no end-to-end test that opens the `/api/agents/{id}/ws` route through `TestServer`, sends a `message` frame with `message_id`, and asserts the terminal frame echoes it back.

I checked the existing test suite before raising this: there's no existing precedent for driving a real WS connection in `crates/librefang-api/tests/` (no `tokio-tungstenite`/WS-client dev-dependency anywhere, `route_smoke.rs` only registers the route in the smoke-test router without connecting to it). So this may be a pre-existing gap in the harness rather than something introduced by this PR, and adding real WS-integration-test infra is a bigger lift than this fix warrants on its own — flagging for a maintainer call on whether the unit-test-only coverage is acceptable here, or whether it should block until a WS test harness exists.

---

PR #6437, opened 2026-07-11
chore(merge): upstream 2026-07-11 (68 commits) + surrealdb 3.2.1 retarget

GQAdonis:
Merges `upstream/main` (68 commits, `fbb5d5cad..77d896b47`) and retargets the three external runtime dependencies onto surrealdb 3.2.1.

## Conflicts resolved (9)

- **`crates/librefang-desktop/tauri.conf.json`** — kept BossFang `productName`, `ai.bossfang.desktop` identifier, the `frontendDist` key (a BossFang change from a57d7f508 that upstream never had), and the GQAdonis updater endpoint + minisign pubkey. Took upstream's version bump to `26.7.32119`.
- **`crates/librefang-llm-drivers/src/drivers/codex_cli.rs`** — took upstream's empty/bare-provider `None` guard, and **preserved the `codex-latest` → `gpt-5.5` rolling alias**. Upstream's rewrite dropped that alias, but it is asserted by live tests in two crates (`codex_cli.rs:525`, `model_catalog/tests.rs:1260`); our remaining match arms were identity mappings already covered by upstream's tail.
- **`crates/librefang-runtime/src/tts.rs`** — took upstream wholesale. It clears the provider keys under `ENV_LOCK` so the strict `expect_err` assertion is valid; our side used `.await` inside a plain `#[test]`, which no longer compiles.
- **`deny.toml`** — union of both advisory lists.
- **`sdk/{javascript,python,rust}`** — kept BossFang package names, took upstream's `2026.7.11` version. Those versions track upstream's own release commits, which is how the previous beta numbering arrived.
- **`.github/workflows/release.yml`** — kept the BossFang tap names. Upstream moved its stable CLI to homebrew-core; `bossfang` is not in homebrew-core, so adopting that line would print install instructions that do not work.
- **`Cargo.lock`** — took upstream's and re-resolved against our manifests.

## Dependency retarget

- **surrealdb `3.0.5` → `3.2.1`** across the family (client, core, types, strand, collections). Single copy in the lock, no duplicates.
- **`universal-agent-runtime`** → `GQAdonis/universal-agent-runtime` rev `0cd664ce` (branch `sync-gqadonis-8c7377a1`), rebasing upstream UAR `8c7377a1` onto the `.gitmodules` SSH→HTTPS rewrite. Pinned by `rev` rather than `branch`: UAR runs a `build.rs`, so a moving branch makes the build non-reproducible and lets an unrelated UAR push break CI.
  Pointing straight at Prometheus-AGS is **not** an option — its submodules are declared over `git@github.com:` SSH, which fails dep-resolution on CI runners with no deploy key. This was tried and reverted on 2026-06-17; see the note at `librefang-llm-drivers/Cargo.toml`.
- **UAR's own surrealdb pin bumped `=3.2.0` → `=3.2.1`** on that branch. Cargo cannot unify two exact `=` pins on one crate, and `uar-driver` is force-enabled via `librefang-kernel`, so a mismatch breaks **every** build rather than an opt-in one.
- **`surreal-memory` `7072370e` → `f9ab1c2`**, and its declaration is now byte-identical to UAR's (`…-server.git?rev=<sha>`). Cargo keys a git source on the literal URL + ref, so our previous `?branch=main` form resolved to the same commit through a different source key and compiled surreal-memory **twice**.

## Also included

`docs(uar)`: research + assessment for the reported "UAR cannot be found in the PATH" cloud bug (phase-10). **No code changes** — documentation and orchestrator artifacts only. Headline: the bug is a *missing capability*, not a misconfiguration. LibreFang never builds, ships, spawns, or supervises a UAR process anywhere. Full write-up in `docs/architecture/uar-runtime-integration-research.md`.

It also documents a factual error in `CLAUDE.md`: `uar-driver` is described as "off by default, opt-in", but `librefang-kernel/Cargo.toml:16` requests it unconditionally, so every shipped binary compiles UAR in.

## Verification

  Note: run with `SKIP_DASHBOARD_BUILD=1`. The dashboard `build.rs` needs corepack pnpm 10.33.0; a Volta-managed pnpm 8 on the local PATH cannot parse `dashboard/pnpm-workspace.yaml` (it uses `onlyBuiltDependencies`, a pnpm 10 feature). CI uses 10.33.0 and is unaffected.
- `python3 scripts/enforce-branding.py --check` — exit 0 (4 docs files patched by the fix pass first).
- Tauri desktop audit — 11/11 checks pass.
- No new upstream SQLite schema in the diff, so no `.surql` migration was needed.
- No new hardcoded upstream URLs in Rust.

## Out-of-scope follow-ups

- `GQAdonis/universal-agent-runtime` does not build or publish the `uar-sidecar` binary (0 refs in `release.yml`; `Dockerfile:225` builds only `--bin universal-agent-runtime`). Small additive change; gates end-to-end testing of the phase-10 spawn path.
- `CLAUDE.md`'s `uar-driver` "opt-in" claim should be corrected (tracked in phase-10 as G-10).


GQAdonis:
Opened against the wrong repository by mistake — this is a downstream fork's merge commit and does not belong upstream. Reopening against the fork. Apologies for the noise.

PR #6439, opened 2026-07-12
fix: second-pass security and correctness hardening from repo-wide audit

houko:
## Summary

A second-pass repo-wide audit (after the #6438 hardening sweep) surfaced a further set of independent defects across the kernel, runtime, memory, API, docker-sandbox, wire, subprocess, hands, channels and CLI crates. This PR fixes them, each with a regression test. The findings are unrelated to one another; the list below is grouped by severity for triage.

## High severity

- **Per-agent fallback chain bypassed the agent's configured model** (`kernel/llm_drivers.rs`). The `FallbackDriver` primary slot set its model-override to the provider name instead of an empty string, so a fallback-configured agent's primary request ran with `model = "anthropic"`, 404'd every turn, and silently failed over to the cheaper fallback. Now passes the request model through unchanged (matching the `boot.rs` sibling).
- **Marketplace hand could exfiltrate daemon secrets via env passthrough** (`kernel/hands_lifecycle.rs`, `runtime/subprocess_sandbox.rs`, `runtime/host_functions.rs`). A hand naming `LIBREFANG_VAULT_KEY` / a provider key in its HAND.toml passthrough had it injected into a shell/python subprocess. The reserved-secret denylist is now shared and enforced both when assembling `allowed_env` and defensively inside `sandbox_command`. Scoped to the actual secret set (vault key, provider keys, credential-word substrings) rather than the whole `LIBREFANG_*` namespace, so legitimate operator env passthrough keeps working.
- **Docker exec could OOM the daemon** (`runtime-sandbox-docker/lib.rs`). `exec_in_sandbox` buffered the whole container output via `cmd.output()` before truncating; a runaway command (`yes`, `cat /dev/zero`) exhausted host memory. Now spawns and byte-caps each stream. Also blocks `/run/docker.sock` (systemd `/var/run` symlink escape) and matches bind-mount containment by path component (no more over-blocking `/development` because of `/dev`).
- **Deleted session content stayed full-text searchable** (`memory/session.rs`). The retention/cleanup sweeps deleted `sessions` rows but left orphan `sessions_fts` rows, which the search endpoint still returned until restart. Each cleanup now deletes the FTS rows in the same transaction, mirroring `delete_session`.
- **User-role privilege escalation on approval endpoints** (`api/middleware.rs`). Unanchored `/approve` `/reject` suffix checks let a User bearer approve pending skills and trust A2A agents. The suffixes are now anchored to the `/api/approvals/` prefix.

## Medium severity

- **Per-agent budget PUT accepted negative/NaN caps** and silently disabled the quota (persisted); now rejected with 400 like the sibling handlers (`api/routes/budget.rs`).
- **AuxClient ignored the model catalog** (`runtime/aux_client.rs` + `kernel/boot.rs` + `config_reload_ops.rs`), so a dashboard-added custom cheap-tier provider never initialised and side tasks billed the primary. Now consults the catalog (snapshot refreshed at boot / config reload) like the primary path.
- **Trusted-peers TOFU store saved non-atomically** (`wire/trusted_peers.rs`); a torn write bricked OFP startup (fail-closed). `save()` now writes a `0600` temp file and renames.
- **Model aliases never persisted** (`api/routes/providers.rs`, `runtime/model_catalog.rs`); API-created/deleted aliases were lost on restart. A `save_aliases` path writes `aliases.toml` with rollback-on-failure.
- **Cron `Every{every_secs:0}` div-by-zero panic** crash-looped boot (`kernel/cron.rs`); the interval is clamped and `load()` re-validates deserialized `Every`/`Cron` jobs.
- **Audit anchor spurious tamper alarm after restart** (`runtime-audit/lib.rs`). The external anchor recorded the soft-cap-shrunken in-memory length; it now tracks the persisted SQLite row count for the anchor seq. The hash-chain tamper check is unchanged (still errors on a real hash mismatch).
- **Vector-store recall trusted the external backend for tenant isolation** (`memory/semantic.rs`); it now post-filters hydrated fragments against the `MemoryFilter` and enforces `peer_id` in the hydration query. `HttpVectorStore` also caps the response body (64 MiB) so a hostile backend cannot OOM the recall path (`memory/http_vector_store.rs`).

## Low severity

- `POST /api/hands/install` now runs the supply-chain scan on caller content (moved into the shared persisted-install helper), matching remote install (`hands/registry.rs`).
- Subprocess stderr drain uses the bounded `read_capped_line` instead of unbounded `lines()` (`subprocess/lib.rs`).
- `librefang init` reconciles a pre-existing `.gitignore` and skips the initial commit on failure, so `vault.enc`/`daemon.json` can't be committed (`cli/init.rs`).
- `ProactiveMemory::update` surfaces a re-embed failure instead of returning success with a stale vector (`memory/proactive.rs`).
- The canvas tool is gated by `config.canvas.enabled` (default-off now actually withholds it) (`kernel/tools_and_skills.rs`).
- `ChannelOverrides.typing_mode = "never"` now suppresses typing indicators (`channels/bridge.rs`).
- Workflow run eviction also drops the run's `cancel_notify` / `operator_resume_notify` entries (leak) (`kernel/workflow.rs`).
- Dashboard-login TOTP now checks the brute-force lockout and records failures, mirroring the approval path (`api/server.rs`).

## Deferred (out-of-scope for a hardening PR; tracked)

- `ResourceQuota.max_network_bytes_per_hour` (per-agent egress cap), `AutonomousConfig.heartbeat_channel` (status delivery), and `DockerSandboxConfig.reuse_cool_secs` (container pooling) are declared-but-unwired, but wiring each is a **new feature** in a different domain, not a fix. Flagged for follow-up rather than half-implemented.
- `CanvasConfig.max_html_bytes` / `allowed_tags` need `CanvasConfig` threaded into the runtime tool-execution context (`ToolExecContext` / kernel-handle `ToolPolicy`), a cross-crate change; only the `enabled` gate is wired here.
- A few audit candidates were dropped after verification (`hash_prompts` toggle is inert but hashing already runs; the `.env` unescape and `ContainerPool` config-hash gaps have no production caller).

## Verification

- Scoped `cargo nextest` across the touched crates passes. The one unit failure was a pre-existing test that used a `LIBREFANG_`-prefixed passthrough var as its fixture; resolved by scoping the env denylist to real secrets rather than the whole `LIBREFANG_*` namespace.
- Note: 10 heavy `agent_loop` integration/recovery tests SIGABRT with a stack overflow under the local nextest default thread stack (2 MB); they pass with `RUST_MIN_STACK=32M`, confirming a finite-depth stack-size artifact of the debug build (no recursion was introduced), not a logic regression — these tests are green in CI on main.

Each finding has a named regression test (e.g. `fallback_primary_slot_preserves_request_model`, `test_sandbox_command_strips_reserved_and_secret_env_vars`, `test_read_capped_bounds_unbounded_output`, `test_cleanup_expired_sessions_clears_fts`, `approval_suffixes_anchored_to_approvals_prefix`, `save_does_not_leave_a_torn_file`, `dashboard_login_locks_out_after_repeated_wrong_totp_codes`).

### DIFF
--- a/crates/librefang-api/src/routes/providers.rs
+++ b/crates/librefang-api/src/routes/providers.rs
@@ -492,6 +492,23 @@ pub async fn create_alias(
             .into_json_tuple();
     }
 
+    // Persist to disk. If the save fails, roll back the in-memory add so the
+    // catalog stays consistent with what's on disk — otherwise the caller
+    // sees "created" now but the alias vanishes on the next daemon restart.
+    let aliases_path = state.kernel.home_dir().join("aliases.toml");
+    if let Err(e) = state
+        .kernel
+        .model_catalog_load()
+        .save_aliases(&aliases_path)
+    {
+        tracing::warn!("Failed to persist aliases: {e}");
+        let alias_for_rollback = alias.clone();
+        state.kernel.model_catalog_update(&mut move |catalog| {
+            catalog.remove_alias(&alias_for_rollback);
+        });
+        return ApiErrorResponse::internal_scrub(e).into_json_tuple();
+    }
+
     (
         StatusCode::CREATED,
         Json(serde_json::json!({
@@ -508,13 +525,34 @@ pub async fn delete_alias(
     State(state): State<Arc<AppState>>,
     Path(alias): Path<String>,
 ) -> impl IntoResponse {
+    // Snapshot the mapped model ID before removing so we can restore it if
+    // the subsequent persist fails — keeps the in-memory catalog consistent
+    // with disk across failure paths.
+    let mut snapshot: Option<String> = None;
     let mut removed = false;
     state.kernel.model_catalog_update(&mut |catalog| {
+        snapshot = catalog.list_aliases().get(&alias.to_lowercase()).cloned();
         removed = catalog.remove_alias(&alias);
     });
     if !removed {
         return ApiErrorResponse::not_found(format!("Alias '{}' not found", alias))
             .into_json_tuple();
+    }
+
+    let aliases_path = state.kernel.home_dir().join("aliases.toml");
+    if let Err(e) = state
+        .kernel
+        .model_catalog_load()
+        .save_aliases(&aliases_path)
+    {
+        tracing::warn!("Failed to persist aliases: {e}");
+        if let Some(model_id) = snapshot {
+            let alias_for_rollback = alias.clone();
+            state.kernel.model_catalog_update(&mut move |catalog| {
+                catalog.add_alias(&alias_for_rollback, &model_id);
+            });
+        }
+        return ApiErrorResponse::internal_scrub(e).into_json_tuple();
     }
 
     (StatusCode::NO_CONTENT, Json(serde_json::json!(null)))
--- a/crates/librefang-runtime-sandbox-docker/src/lib.rs
+++ b/crates/librefang-runtime-sandbox-docker/src/lib.rs
@@ -7,6 +7,7 @@ use librefang_types::config::DockerSandboxConfig;
 use sha2::{Digest, Sha256};
 use std::path::Path;
 use std::time::Duration;
+use tokio::io::AsyncReadExt;
 use tracing::{debug, error, warn};
 
 /// SECURITY: Allowlist of Linux capabilities considered safe to grant back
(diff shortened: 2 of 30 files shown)


crates/librefang-api/src/routes/providers.rs
```diff
@@ -492,6 +492,23 @@ pub async fn create_alias(
             .into_json_tuple();
     }
 
+    // Persist to disk. If the save fails, roll back the in-memory add so the
+    // catalog stays consistent with what's on disk — otherwise the caller
+    // sees "created" now but the alias vanishes on the next daemon restart.
+    let aliases_path = state.kernel.home_dir().join("aliases.toml");
```
houko, line 498: `create_alias`/`delete_alias` now persist to `aliases.toml` with rollback-on-failure — a materially new behavior for these routes. Per CLAUDE.md's mandatory-integration-test rule, could you add a `#[tokio::test]` in `crates/librefang-api/tests/providers_routes_test.rs` exercising the persist/rollback path through the actual route (e.g. create an alias via the route, then assert it survives a re-read of `aliases.toml`)? The existing `add_alias_then_save_and_reload_restores_alias` in `model_catalog/tests.rs` only covers the lower-level catalog method, not the route's persist-or-rollback wiring.

---

crates/librefang-runtime-sandbox-docker/src/lib.rs
```diff
@@ -512,10 +585,30 @@ const BLOCKED_MOUNT_PATHS: &[&str] = &[
     "/sys",
     "/dev",
     "/var/run/docker.sock",
+    // On systemd hosts `/var/run` is a symlink to `/run`, so the real
+    // Docker socket path is `/run/docker.sock`; block both the socket
+    // and the `/run` runtime dir that shadows it (host-root escape).
+    "/run/docker.sock",
+    "/run",
```
houko, line 592: Blocking all of `/run` (rather than just `/run/docker.sock`) is a broad tightening beyond what's needed to close the systemd `/var/run` → `/run` symlink escape described in the comment above. This could reject legitimate bind mounts some operators already rely on (e.g. `/run/user/1000` for some workloads). The accompanying test (`test_validate_bind_mount_blocks_run_docker_sock`) only documents blocking the socket path, not the broader directory — could you confirm this breadth is intentional, or narrow the block to `/run/docker.sock` (plus maybe `/run/containerd` etc. if those are specific known-risky subpaths) rather than the whole tree?

---

houko (commented):
Automated CLAUDE.md-rules pass. One mechanical fix pushed as a follow-up commit (misattached doc comments in `crates/librefang-channels/src/bridge.rs`, where `typing_indicator_suppressed()`'s doc comment had absorbed `is_command_allowed()`'s trailing doc lines). Two remaining items below as inline comments, both about missing test coverage / scope of a security tightening rather than correctness bugs.

Everything else checked out clean: no missing `Default` impl entries for new config fields, no `HashMap`/`HashSet` in prompt-facing paths, no un-`serde(skip)`'d `Option<Arc<dyn Trait>>` fields, the `middleware.rs` approval-suffix anchoring fix is correct, and the `budget.rs`/dashboard-TOTP-lockout changes both have solid matching integration tests.

---

PR #6446, opened 2026-07-13
fix: fourth-pass security and correctness hardening from repo-wide audit

houko:
A fourth repo-wide audit pass (after #6438, #6439, #6441) fixes a further set of independent, verified defects across the runtime, API, channels, kernel, memory/skills, and driver layers. Each defect was surfaced by a fan-out audit and adversarially re-verified against the current tree (post-#6441); each fix ships with a regression test unless noted.

## Security

- **`process_start` leaked the daemon environment (critical).** `ProcessManager::start` spawned the child without `env_clear()` + the safe-var allowlist that every other exec path uses, so under the default `Allowlist` posture (`env` is a safe_bin) an agent could `process_start {"command":"env"}` then `process_poll` and read back `LIBREFANG_VAULT_KEY` and provider API keys. The child now inherits only the safe baseline plus the agent's resolved `allowed_env_vars`.
- **`process_poll` / `process_write` / `process_kill` had no owner check.** Keyed only on the globally-sequential, guessable `proc_N` id, so a second agent could drain another agent's stdout, inject into its stdin, or kill it. The caller's agent id is now threaded down; a mismatched owner is reported as "not found".
- **Persistent plugin hook subprocess inherited the full daemon env.** `PersistentProcess::spawn` re-added every `std::env::vars()` entry, so `[hooks] persistent_subprocess = true` handed the plugin `LIBREFANG_VAULT_KEY` + all provider keys. Both spawn paths now share a `hook_baseline_env` helper (safe baseline + `allowed_env_vars` filtered through `is_blocked_env_var`).
- **`GET /api/config/export` returned the raw unredacted `config.toml` to any role.** Reachable by any authenticated Viewer/User/Admin as a plain GET; a leaked master `api_key` re-presents as `Owner` (full escalation). Now Owner-gated via `min_role_for_privileged_get`.
- **Channel slash-commands bypassed the RBAC gate.** Command dispatch `return`ed above the chat-path `authorize_channel_user` gate, so an unauthorized user could `/approve` pending tool calls and run control-plane commands. Both command paths now authorize first (no-op when RBAC is off).
- **`webhook_triggers.max_payload_bytes` was dead config on `/hooks/*`.** Only the global 8 MiB cap applied. A `RequestBodyLimitLayer` sized from the config now bounds both the `/api/hooks/*` and unversioned aliases.

## Fixed

- **`FallbackChain` slept a provider's `Retry-After` verbatim** (up to ~1h RPH lockouts) before failover; now fails over immediately past a 10 s cap, matching the streaming path.
- **Attachment-URL fetch buffered the whole body before the 20 MB check** (remote OOM); now streams with a running cap and a Content-Length pre-check.
- **Workflow runs hung in `Running`** when a Conditional/Loop/DAG step referenced a deleted agent; all resolution-failure paths now mark the run `Failed` via a shared helper, and the DAG concurrent layer skips resolving a dependency-failed step's agent.
- **ClawHub/Skillhub skill-zip extraction was unbounded** (disk/inode exhaustion); now shares the marketplace decompression-bomb guards via `write_zip_entry_capped`.
- **Per-IP WS tracker guard-drop TOCTOU** let a single IP exceed `max_ws_per_ip`; the `Drop` impl now uses `remove_if` with a re-checked counter under the write lock (covers agent + terminal WS).
- **Config-reload false-positive change detection**: `field_changed` compared order-sensitive JSON strings of `HashMap`-bearing fields; now compares canonicalized `serde_json::Value`s.
- **Session auto-reset skipped for persistent channel sessions** under `session_mode = "new"`; the skip is now keyed on whether the session id was freshly minted this call.
- **ChatGPT/Codex driver dropped cached-prompt-token count** (over-metering); now reads `input_tokens_details.cached_tokens`.

## Changed (previously-ignored config now honoured / surfaced)

- **Canvas** `max_html_bytes` / `allowed_tags` are threaded into the tool (the `CANVAS_MAX_BYTES` task-local was never scoped); an empty `allowed_tags` still falls back to the built-in list.
- **Parallel-tools** `mcp_default_safety` / `mcp_readonly_allowlist` are now consumed by `plan_batch`.
- **`channels.file_upload_max_bytes`** doc corrected + boot WARN when tuned (unenforced since the sidecar migration; full enforcement is a sidecar-crate change — out of scope here).
- **`prompt_intelligence.hash_prompts`** doc corrected + boot WARN when set `false` (the `content_hash` is load-bearing for dedup, so it is always computed; gating it would break version dedup).

## Verification

- `cargo check --workspace --lib` clean.
- Scoped `cargo test -p <crate>` green across every touched crate, each new fix covered by a regression test. Exceptions: the session-reset fix is a logic-only change validated by the kernel suite; the two doc+WARN changes (file_upload_max_bytes, hash_prompts) have no behavior to unit-test; and the channel-command RBAC gate has no committed test — `dispatch_message` is private (so no external integration test can reach it) and a `ChannelAdapter` test double inside `librefang-channels/src` trips the sidecar-first policy guard (`cargo xtask channel-policy`), so it is verified by compile + by mirroring the already-tested chat-path gate at `bridge.rs` (an end-to-end test written during development passed before removal).
- `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt` clean.

## Notes / out of scope

- The `kernel_config_schema.golden.json` fixture is left untouched: it is a `#[ignore]`d, known-out-of-date reference (tracked by #4084), so the two config doc-comment changes here are intentionally not swept into it.
- `webhook_triggers.rate_limit_per_minute` enforcement and full sidecar `file_upload_max_bytes` enforcement are left for follow-ups (per-IP webhook limiter infra / a sidecar-crate change respectively); both are bounded today by the global limiter / sidecar hardcoded cap.


crates/librefang-channels/src/bridge.rs
```diff
@@ -3914,6 +3914,29 @@ async fn dispatch_message(
...
+            // slash-command. Commands are handled and `return` here — above the
+            // chat-path `authorize_channel_user` gate near the bottom of this
+            // fn — so without this an unauthorized user (rejected on any normal
+            // message when RBAC / an allowlist is configured) could still
+            // `/approve` pending tool calls, spawn/switch agents, reset another
+            // user's session, or run control-plane commands. The gate is a
+            // no-op when RBAC is disabled, so open deployments are unaffected.
+            // (#6441 follow-up)
+            if let Err(denied) = handle
+                .authorize_channel_user(ct_str, sender_user_id(message), "chat")
```
houko, line 3927: This RBAC gate (and its twin at the text/button command path around line 4340) is one of the two headline "Security" fixes in the PR body, but I don't see a `#[tokio::test]` anywhere in this diff that exercises `dispatch_message` with RBAC configured and an unauthorized sender to confirm `/approve` (or any other privileged command) is actually denied post-fix. The PR body explains the gap (`dispatch_message` is private so no external integration test can reach it, and a `ChannelAdapter` test double inside `librefang-channels/src` trips the sidecar-first policy guard), which is a reasonable constraint, but it means this fix currently ships on "compiles + mirrors the already-tested chat-path gate" rather than a direct regression test. Given this is exactly the kind of auth-bypass fix CLAUDE.md's integration-testing rule targets, is there an in-crate unit test feasible against `dispatch_message` directly (it's `pub(crate)`/private but still reachable from an in-module `#[cfg(test)]` block, unlike an external integration test)? If not, worth being explicit in the PR body that this is accepted as a known test gap rather than reading as "covered."

---

crates/librefang-kernel/src/workflow.rs
```diff
@@ -4274,8 +4301,14 @@ impl WorkflowEngine {
                     max_iterations,
                     until,
                 } => {
-                    let (agent_id, agent_name, agent_inherit) = agent_resolver(&step.agent)
-                        .ok_or_else(|| format_missing_agent_error(&step.name, &step.agent))?;
+                    let (agent_id, agent_name, agent_inherit) = match agent_resolver(&step.agent) {
+                        Some(v) => v,
+                        None => {
+                            let e = format_missing_agent_error(&step.name, &step.agent);
+                            mark_run_failed(&self.runs, &run_id, &e);
```
houko, line 4308: `mark_run_failed` is called identically from four sites (Conditional here at ~4230, Loop here at ~4308, the single-step DAG path at ~5286, and the DAG-concurrent path at ~5372), but the only regression test added (`missing_agent_on_conditional_step_marks_run_failed`) exercises the Conditional arm. The Loop and DAG arms are structurally similar so probably fine, but they're also exactly the kind of copy-pasted-four-times code where one arm silently diverges later (e.g. someone adds a fifth `StepMode` variant, or refactors DAG dependency resolution, and only patches 3 of 4 call sites) without a test catching it. Worth a parametrized test or at least one more case covering the DAG-concurrent `dep_failed` branch, since that one has extra logic (the `if dep_failed { ... } else { match agent_resolver ... } }` short-circuit) that the Conditional test doesn't touch at all.

---

crates/librefang-runtime/src/plugin_runtime.rs
```diff
@@ -1623,6 +1623,57 @@ fn expand_env_value(val: &str) -> String {
...
+    for var in &config.allowed_env_vars {
+        if crate::subprocess_sandbox::is_blocked_env_var(var) {
+            warn!(var = %var, "refusing to pass reserved/secret env var to persistent hook");
+            continue;
+        }
+        if let Ok(val) = std::env::var(var) {
+            env.push((var.clone(), val));
+        }
+    }
+    for (k, v) in &config.plugin_env {
```
houko, line 1671: `hook_baseline_env` correctly filters `config.allowed_env_vars` through `is_blocked_env_var` a few lines up, but `config.plugin_env` (added here, unfiltered) goes through `expand_env_value`, which does `${VAR_NAME}` → `std::env::var(VAR_NAME)` substitution with no blocklist check. A `plugin.toml` `[env]` entry like `SOME_KEY = "${LIBREFANG_VAULT_KEY}"` would inject the vault key's literal value into the persistent hook subprocess under an arbitrary key name, bypassing the exact protection this PR just added for `allowed_env_vars`. I believe this predates this PR (the non-persistent `run_hook_json` path a few hundred lines up does the same unfiltered `plugin_env` → `expand_env_value` without a blocklist check), so it's not a regression introduced here, and `plugin_env` is operator/plugin-author-authored config rather than the attacker-influenceable `allowed_env_vars` passthrough list the PR's threat model targets — but since this PR is explicitly hardening "what env reaches a hook subprocess," it might be worth either applying `is_blocked_env_var` to the *expanded* value's source var too, or documenting why `plugin_env` is considered a different trust tier than `allowed_env_vars`.

---

crates/librefang-api/src/routes/agents/attachments.rs
```diff
@@ -335,23 +350,45 @@ pub async fn resolve_url_attachments(
...
+                // fully buffers the body into memory *before* any size check,
+                // so a lying/absent Content-Length streaming an unbounded body
+                // could OOM the daemon (bounded only by the 30s timeout, i.e.
+                // multiple GB at high bandwidth). Accumulating with an in-loop
+                // cap bounds peak memory to MAX_ATTACHMENT_BYTES regardless of
+                // the server's send rate (mirrors the channels
+                // `fetch_url_bytes_unchecked` defence).
+                let mut resp = resp;
+                let mut data: Vec<u8> = Vec::new();
+                let collected = loop {
```
houko, line 371: The new regression test (`push_within_cap_rejects_over_cap_chunk`) only covers the pure `push_within_cap` helper, not this streaming loop itself (the Content-Length pre-check, the `resp.chunk()` loop, or the "abort and `continue`" behavior on cap-exceeded). Given the fix is specifically about avoiding `resp.bytes()` fully buffering an unbounded body, a test using something like `wiremock` (already a dev-dependency elsewhere in this PR) to serve an over-cap body and assert `resolve_url_attachments` skips it without buffering the whole thing would directly exercise the fixed code path rather than just the helper it happens to call. Not blocking, but worth noting since this is one of the weaker-covered claims relative to the other fixes in the PR (which mostly test the actual fixed function, not just an extracted pure helper).

---

houko (commented):
Reviewed the fourth-pass hardening diff against CLAUDE.md's conventions (deterministic-prompt BTreeMap rule, config-field completeness, is_public allowlist mechanics, Option<Arc<dyn Trait>> serde-skip rule) and cross-checked the "Security"/"Fixed" claims against their regression tests. The env-scrub logic in `ProcessManager::start` and `PersistentProcess::spawn` (the most safety-critical piece) is directionally correct — no inverted conditions or off-by-ones. No HashMap/HashSet or config-field/serde-derive issues found; the `/api/config/export` RBAC gate correctly uses `min_role_for_privileged_get` rather than touching `is_public`.

Pushed one small, mechanical commit (42d9a39) rewrapping a handful of new doc/inline comments in the security-critical paths (config field docs, `ProcessManager::start`, `hook_baseline_env`, `PersistentProcess::spawn`, `field_changed`, `min_role_for_privileged_get`) to the repo's one-sentence-per-line prose rule — comment text only, no logic changes; `cargo check`/`clippy`/`fmt --check` clean and the touched crates' existing test suites still pass. The prose-wrap issue is broader than these five files (most new doc comments across this PR are still hard-wrapped at ~72-80 cols), but I kept this fix scoped to what I could verify carefully rather than rewrapping the whole diff unattended.

Left four inline comments on things that look like real (but non-blocking, non-mechanical) gaps rather than fixing them directly, since they touch test coverage judgment calls or a security trust-boundary question I didn't want to unilaterally resolve:
- `bridge.rs`: the channel slash-command RBAC bypass fix (one of the two headline security items) ships with no direct regression test reaching `dispatch_message`; the PR body explains the constraint but it's worth being explicit that this is an accepted test gap.
- `workflow.rs`: `mark_run_failed` is called from four near-identical sites but only the Conditional arm has a regression test; Loop and the DAG-concurrent `dep_failed` branch are untested.
- `attachments.rs`: the attachment-OOM fix's only test covers the extracted `push_within_cap` helper, not the actual streaming loop in `resolve_url_attachments`.
- `plugin_runtime.rs`: `hook_baseline_env` filters `allowed_env_vars` through `is_blocked_env_var` but not `plugin_env`'s `${VAR}` expansion, so a plugin manifest could still reference `${LIBREFANG_VAULT_KEY}` under an arbitrary key name. Looks pre-existing (mirrors the non-persistent path) rather than a regression, but worth a decision on whether that's an accepted trust boundary.

---

Issue #6444, opened 2026-07-13
bug(channels/matrix-sidecar): every room is_group=true and was_mentioned never set — mention_only drops all matrix inbound, mentions included


neo-wanderer:
## Summary

The Matrix sidecar adapter marks **every** room message as a group message and never surfaces mention state in the metadata key the group gate reads. As soon as any `ChannelOverrides` applies to the receiving agent (see repro), `group_policy` defaults to `mention_only` and **all Matrix inbound — 1:1 DMs and explicit @-mentions included — is silently dropped** with `OB-06 group_gating_skip`. Nothing reaches the agent, the message journal, or the sender; from the user's side the bot just goes dead.

All references below are `main` @ `d243bcdec`.

## The two adapter gaps

**1. No DM detection — `is_group=True` unconditionally.**
`sdk/python/librefang/sidecar/adapters/matrix.py:1154` — the single `protocol.message(...)` call sets `is_group=True` for every `m.room.message`, regardless of room membership. A 1:1 DM room is indistinguishable from a 50-person room. (Contrast: the Telegram adapter flags private chats as non-group, so the same overrides leave Telegram DMs working.)

**2. `was_mentioned` is never set.**
The gate for `GroupPolicy::MentionOnly` reads only `metadata["was_mentioned"]`:

- `crates/librefang-channels/src/bridge.rs:2574` `should_process_group_message` → line 2594 `message.metadata.get("was_mentioned")`, defaulting to `false`
- skip log at line 2661: `reason = "mention_only_no_mention"` (OB-06)

The Matrix adapter forwards MSC3952 `m.mentions` user ids as `metadata["mention_names"]` (`matrix.py:1146`) — used for multi-agent routing (#5323) — but never sets `metadata["was_mentioned"]`, even when its own `user_id` is in the mention list. The Discord adapter does this correctly (`sdk/python/librefang/sidecar/adapters/discord.py:312-333` sets `metadata["was_mentioned"] = True`); Feishu likewise.

Net effect: under `mention_only`, a Matrix message can **never** pass the gate. It is not "respond only when mentioned" — it is "never respond".

## Repro (minimal, synthetic)

1. Configure the Matrix sidecar (`[[sidecar_channels]]` with the matrix adapter) with `default_agent = "A"`.
[...]

Issue #6443, opened 2026-07-12
bug(security/channels): channel_send account_id accepted unvalidated — cross-tenant dispatch across bot accounts


nevgenov:
## Summary

`channel_send`'s `account_id` parameter is accepted from the model with **zero validation** — an agent can dispatch a message to any configured channel account (i.e. any other bot/tenant instance registered on the daemon), not just the one it was invoked from. In a multi-tenant setup (distinct bot accounts serving different customers/organizations on the same daemon), this is a cross-tenant confidentiality boundary, not just a UX footgun.

This is the same bug *class* as #6117 (cross-chat dispatch via a mismatched `recipient` on the same channel) but a wider blast radius: #6117's fix explicitly scopes itself to same-channel re-targeting and states cross-channel dispatch "stays allowed" by design. `account_id` (cross-account, i.e. effectively cross-tenant) sits entirely outside that guard.

Verified against the current stable release **v2026.7.11** (pinned to the git tag, not `main`).

## Root cause

`crates/librefang-runtime/src/tool_runner/channel.rs`, inside `tool_channel_send()`:

```rust
let thread_id = trim_opt_string(input["thread_id"].as_str());
let account_id = trim_opt_string(input["account_id"].as_str());
```

`account_id` is extracted from the tool-call input and passed straight through to `kh.send_channel_media(...)` / the text-send equivalent with no allowlist check against the calling agent, no comparison to the turn's originating channel/account, and no capability gate beyond the coarse `channel.*` capability that merely grants/denies the tool as a whole.

The #6117 cross-chat guard (same file, a few lines above) only fires when:
```rust
if let (Some(explicit), Some(turn_channel)) = (explicit_recipient, sender_channel) {
    ...
    if turn_channel.eq_ignore_ascii_case(&channel) && explicit != expected { ... }
}
```
— i.e. it compares `channel` (the channel identity the request is being sent *on*, e.g. a specific bot instance) against the turn's inbound channel, and only rejects a mismatched *recipient* within the *same* channel. A `channel_send` call that targets an entirely different `channel`/`account_id` bypasses this check completely — which the issue that introduced it (#6117) explicitly calls out as intentional scope-narrowing ("a different-channel dispatch... stays allowed").

Related but orthogonal existing mechanism: `ChannelToolRule` (`crates/librefang-types/src/approval.rs`) gates which *tools* may run for requests *originating from* a channel (inbound). It has no notion of restricting which channel/account an agent's *outbound* `channel_send` may target.

[...]

Issue #6442, opened 2026-07-12
bug(kernel/agents): safe_path_component() collapses any all-non-ASCII agent name to the same fallback TOML path, breaking auto-discovery


nevgenov:
## Summary

`safe_path_component()` — used to resolve the fallback `agent.toml` path when `source_toml_path` is `None` in the DB — strips **any non-ASCII character**. If an agent's display name contains **zero** ASCII alphanumeric/`-`/`_` characters (e.g. an entirely Cyrillic, CJK, or accented-Latin name), the sanitized string is empty and the code falls back to a **shared literal path** `workspaces/agents/agent/agent.toml` for *every* such agent. TOML edits on disk for the real agent are then silently ignored — the loader looks in the wrong directory and finds nothing (or another agent's stale file).

Confirmed still present in the current stable release **v2026.7.11** (pinned against the git tag, not `main`) and reproduced live against a running v2026.7.11 daemon.

## Root cause

`crates/librefang-kernel/src/kernel/workspace_setup.rs`:

```rust
pub(super) fn safe_path_component(input: &str, fallback: &str) -> String {
    let sanitized: String = input
        .chars()
        .filter(|c| c.is_ascii_alphanumeric() || *c == '-' || *c == '_')
        .collect();
    if sanitized.is_empty() {
        fallback.to_string()
    } else {
        sanitized
    }
}
```

Called from three sites, all computing a *fallback* TOML path when `source_toml_path` isn't set in the DB:

- `boot.rs:1766` (boot-time TOML re-sync loop)
- `agent_state.rs:42` and `agent_state.rs:213` (runtime reload path)

All three build:
```rust
let safe_name = safe_path_component(&name, "agent");
cfg.effective_agent_workspaces_dir().join(safe_name).join("agent.toml")
```

For a name with zero ASCII chars, `safe_name == "agent"` — a **fixed literal**, not derived from the actual agent name. This means:

1. The fallback path silently points at the wrong directory, so TOML edits for that agent are never picked up (agent stays on whatever model/config is cached in the `agents.manifest` DB blob).
2. **Every** agent with an entirely non-ASCII name collides on the same fallback path `workspaces/agents/agent/agent.toml` — with two such agents, one can silently pick up the other's TOML on next boot-time re-sync.

## Steps to reproduce

[...]

Issue #6423, opened 2026-07-09
OpenRouter-specific runtime context-window issue


pavver:
### Description

I found another OpenRouter-specific runtime context-window issue.

This is visible in the chat UI context usage bar. After selecting `tencent/hy3:free`, clearing chat history, and sending one message, the UI shows something like:

`3,415 / 8,192 tokens (~41.7%)`

That denominator is wrong. OpenRouter’s `/models` response reports `context_length = 262144` for `tencent/hy3:free`.

The catalog parser itself appears correct: it maps OpenRouter `context_length` to `ModelCatalogEntry.context_window`, and maps `top_provider.max_completion_tokens` separately to `max_output_tokens`.

The issue seems to be in runtime/session context resolution. `GET /api/agents/{id}/session/context` calls `context_report_for_session`, and that currently resolves the model context with provider-blind lookup:

```rust
catalog.find_model(&entry.manifest.model.model)
```

For OpenRouter, live catalog entries are stored as `openrouter/{raw_model_id}`, while the agent manifest stores:

```toml
provider = "openrouter"
model = "{raw_model_id}"
```

For example, the catalog has `openrouter/tencent/hy3:free`, but the manifest has `tencent/hy3:free`. The provider-blind lookup misses the catalog entry and falls back to `UNKNOWN_MODEL_CONTEXT_WINDOW = 8192`.

This affects the chat context progress bar and likely any runtime/session budgeting path using the same provider-blind lookup. The fix should be to resolve context through provider-aware lookup, e.g. `find_model_for_provider(provider, model)`, or to route these paths through the existing `resolve_model_metadata` helper consistently.

houko:
Fixed in #6449 (commit 492dbf315), merged 2026-07-13.

The 8192 fallback you observed is real, but the root cause turned out to be a provider-prefix mismatch rather than provider-blind lookup alone: catalog entries are keyed `openrouter/{raw_id}`, while the dashboard model picker strips that prefix before persisting, so the manifest holds a bare `raw_id` that never matches the catalog and falls through to the unknown-model 8192 default.
The suggested `find_model_for_provider(provider, model)` would still miss, because it also requires an exact `id` match against the prefixed catalog key.

The fix adds `ModelCatalog::find_model_for_manifest(provider, model)`, resolving in order: exact `(provider, model)` → re-qualified `(provider, "{provider}/{model}")` → the legacy provider-blind `find_model` fallback, so nothing that resolved before regresses.
[...]

PR #6449, opened 2026-07-13
fix: resolve four reported bugs (#6423, #6442, #6443, #6444)

houko:
## Summary

Fixes four open bug reports, verified present on `main`:

- **#6443** (security) — `channel_send` `account_id` accepted unvalidated → cross-tenant dispatch
- **#6442** — `safe_path_component()` collapses any all-non-ASCII agent name to a shared fallback TOML path
- **#6444** — Matrix sidecar marks every room `is_group=true` and never sets `was_mentioned`, so `mention_only` drops all inbound
- **#6423** — OpenRouter context window resolves to the 8192 unknown-model fallback

Each was cross-checked against the current code before fixing; #6423 in particular is fixed differently from the reporter's suggestion (see below).

## #6443 — cross-account (cross-tenant) `channel_send` guard

`account_id` selects which registered bot instance a send routes through. It was extracted from the model's tool input and passed straight to the kernel send path with no validation. On a multi-tenant daemon (several bot accounts, each with its own `default_agent` serving a different customer) an agent induced via hallucination or prompt injection could pass another tenant's `account_id` and dispatch content into that customer's chat. The #6117 guard does not cover this — it only compares `recipient` within the same channel and runs before `account_id` is even parsed.

Changes:
- Kernel stamps the turn's originating account (`SenderContext.account_id`, already populated from the inbound adapter config name) into `manifest.metadata["sender_account_id"]`, next to the existing `sender_chat_id` stamp — both `messaging.rs` and `agent_execution.rs` sites.
- The runtime agent loop reads it back and threads it to `channel_send`. `execute_tool` keeps its historical positional signature (external API tool-run routes and ~57 unit tests untouched) and forwards `None` to a new `execute_tool_with_sender_account`; only the in-process agent loop threads the stamped account.
- `tool_channel_send` rejects an explicit `account_id` that differs from the turn's account on the same channel, mirroring the #6117 recipient guard's explicit-value-only scoping. Auto-filled sends carry no `account_id`; out-of-band callers (cron / triggers / API) leave the account `None` and the guard no-ops.
- The approval-deferred path is covered too: `DeferredToolExecution` gains an `account_id` field (serde-default, back-compatible with persisted rows) so an approved, deferred `channel_send` still enforces the guard.

**Question for maintainer — `/mcp` bridge parity.** #6117 fixed the in-process path and then the `/mcp` bridge (subprocess `claude-code` driver) in a separate commit. This PR fixes the in-process path for #6443 the same way. The `/mcp` bridge path (`network.rs` reads `x-librefang-current-*` headers forwarded by `claude_code.rs::write_mcp_config`) would need a `current-account-id` header threaded through `librefang-llm-drivers` + `librefang-api` for full parity — a different crate/domain than the rest of this PR. Should I extend it here, or land it as a follow-up mirroring how #6117 was split? Flagged rather than silently deferred.

## #6442 — non-ASCII agent name TOML fallback collision

`safe_path_component()` strips every non-ASCII character. An agent whose display name is entirely Cyrillic / CJK / accented-Latin sanitizes to the empty string, and the three fallback sites passed the literal `"agent"`, so every such agent collapsed onto `workspaces/agents/agent/agent.toml`: TOML edits were silently ignored and two such agents overwrote each other. The spawn-time resolver (`resolve_workspace_dir`) already uses the agent's UUID as the fallback, so the fallback sites disagreed with the directory spawn actually created.

Fix: the three fallback sites (`boot.rs` re-sync, `persist_manifest_to_disk`, `reload_agent_from_disk`) now pass `agent_id.to_string()` as the fallback, matching `resolve_workspace_dir`. Distinct agents never collide, and the fallback resolves the same directory spawn created (issue's suggested fix #2).

## #6444 — Matrix sidecar DM / mention detection

The adapter hardcoded `is_group=True` for every room and never set `metadata["was_mentioned"]` — the only mention signal the group gate reads. Under the default `GroupPolicy::MentionOnly`, 1:1 DMs and explicit @-mentions were both silently dropped (`OB-06 group_gating_skip`), so the bot appeared dead.

Fix (both in `matrix.py`):
- DM detection — a room is flagged non-group when it is listed in `m.direct` account data or its `/sync` summary reports two joined members. The verdict is cached (`_dm_rooms`) because the summary is sent lazily; a later summary-less sync does not flip a known DM back to group.
- `was_mentioned` — set to `True` when the bot's own `MATRIX_USER_ID` is in the MSC3952 `m.mentions` list, parity with the Discord / Feishu adapters.

## #6423 — OpenRouter context-window resolution

The symptom (8192 denominator) is real, but the reporter's diagnosis and suggested fix were both off:
- Root cause is a **provider-prefix mismatch**, not mere provider-blindness: catalog entries are keyed `openrouter/{raw_id}`, but the dashboard model-picker (`set_agent_model`) strips the prefix before persisting, so the manifest holds a bare `raw_id`.
- The suggested `find_model_for_provider(provider, model)` still requires an exact `id` match, so it would **also** miss the prefixed catalog id.
- `resolve_model_metadata` is not a drop-in — it has zero production callers and its prefix-stripping does not cover the `openrouter/` slash form.

Fix: a new `ModelCatalog::find_model_for_manifest(provider, model)` resolves in order — exact `(provider, model)`, then re-qualified `(provider, "{provider}/{model}")`, then provider-blind `find_model` (legacy fallback, so nothing that resolved before regresses). All eight provider-blind `find_model(&manifest.model.model)` lookup sites in the kernel — context report, live `ContextBudget` (this is not merely a display bug), compaction, and tool-support detection — now route through it.

## Verification

- `cargo check -p librefang-runtime -p librefang-kernel -p librefang-types --lib --tests` — clean (run natively in this worktree's isolated `target/`, no contention with the shared tree).
- `cargo clippy` on the touched crates — zero warnings.
- New tests:
  - `librefang-kernel workspace_setup::non_ascii_fallback_tests` — collision + spawn-parity for #6442.
  - `librefang-runtime tool_runner::tests` — five `channel_send` account-guard cases for #6443 (mismatch blocked, match/no-account/cross-channel/out-of-band pass).
  - `librefang-runtime model_catalog::tests::test_find_model_for_manifest_reconciles_openrouter_prefix` for #6423.
  - `sdk/python tests/test_matrix_adapter.py` — six cases for #6444 (`was_mentioned` set / not-set, DM via summary + `m.direct`, group via summary, cached verdict); 120 passed.

## Out-of-scope follow-ups

- `/mcp` bridge parity for #6443 — flagged above, pending maintainer call.


crates/librefang-runtime/src/model_catalog.rs
```diff
@@ -490,6 +490,51 @@ impl ModelCatalog {
         None
     }
 
+    /// Resolve a catalog entry for a manifest's `(provider, model)` pair,
```
houko, line 493: Minor, repo-rule nit: CLAUDE.md's prose-wrapping rule says doc comments should break only at sentence boundaries, not at a column width — "one sentence = one line, regardless of length." This doc comment (and quite a few others this PR adds — e.g. the `channel_send` account-guard comment block in `tool_runner/channel.rs:245-261`, the `#6442` fallback comments in `agent_state.rs`/`boot.rs`, several in `matrix.py`) hard-wraps mid-sentence at ~80 cols instead, e.g.:

```
/// Resolve a catalog entry for a manifest's `(provider, model)` pair,
/// reconciling the provider-prefix mismatch that OpenRouter (#6423) — and
/// any provider whose live catalog ids are stored as `{provider}/{raw_id}`
/// — exhibits.
```

That's one sentence split across 4 lines. I recognize this matches the pre-existing style already used nearby (e.g. the #6117 comments this PR extends), so it may just be established practice in this crate rather than an oversight — flagging in case it's actually intended to be enforced going forward, since the rule is explicit about applying to "any paragraph an agent is already touching." Not blocking either way.

---

crates/librefang-kernel/src/kernel/workspace_setup.rs
```diff
@@ -1259,3 +1259,78 @@ mod identity_write_tests {
         );
     }
 }
+
+#[cfg(test)]
+mod non_ascii_fallback_tests {
```
houko, line 1264: These regression tests cover the three fallback sites this PR fixes (`agent_state.rs`, `boot.rs`), but there's a fourth call site in this same file that looks like the same collision class and wasn't converted: `backfill_workspace_dir` (workspace_setup.rs:366-367), a few hundred lines up in this file:

```rust
let safe_hand = safe_path_component(hid, "hand");
let safe_role = safe_path_component(role, "agent");
```

If `hand_role` (read from a tag, potentially free text) sanitizes to empty, every such hand agent collapses onto the shared literal `"agent"` fallback — the same bug as #6442. `hid` is probably already a slug/id, so less likely to be non-ASCII, but `role` looks like it could plausibly be operator-supplied. Flagging in case it was simply missed rather than deliberately out of scope; not necessarily blocking.

---

houko (commented):
Reviewed the diff against CLAUDE.md (worktree/hook rules, deterministic-prompt-ordering, serde-default gotcha, conventional commits, prose wrapping) plus general correctness. Verified `cargo check -p librefang-kernel -p librefang-runtime -p librefang-types --lib --tests` and `cargo clippy -p librefang-kernel --all-targets -- -D warnings` clean in a fresh worktree on this branch.

One thing I fixed directly and pushed (58a4684): `manifest_helpers::global_thinking_backfill_allowed` still called the provider-blind `find_model(model)` — the exact lookup this PR's #6423 fix replaced at its other eight call sites in `agent_execution.rs` / `messaging.rs`. A bare OpenRouter manifest model would miss the prefixed catalog entry here too and fall back to the permissive "unknown model" default, silently backfilling `[thinking]` config onto a model that doesn't support it. Routed it through `find_model_for_manifest(provider, model)` like the rest, and added a regression case next to the existing #6398 test.

Two inline comments below on things I'm *not* confident are worth fixing unilaterally (both same-bug-class as things this PR does fix, but touch code this PR doesn't, so I've left them for your call rather than expanding the diff).

Also worth a quick look while you're in this area, though I couldn't leave it as a line comment since these files aren't part of the diff: `find_model_for_manifest` reconciles the OpenRouter prefix mismatch for context-window/tool-support/thinking-backfill lookups, but a few other provider-blind `find_model(&model)` calls resolving `reasoning_echo_policy` from a `(provider, model)` pair look like the same gap — they discard the provider and would miss a bare OpenRouter model too: `kernel/session_ops.rs:1204` and `kernel/triggers_and_workflow.rs:75` (both discard `resolution.resolved`'s `(String, String)` provider via `.map(|(_, m)| ...)`). `kernel/handles/catalog_query.rs:25` and `:38` (`lookup_reasoning_echo_policy` / `lookup_supports_vision`) have the same symptom but are behind the `CatalogQuery` trait shared with `librefang-kernel-handle` / drivers, so fixing those would mean a cross-crate signature change rather than a local one. Not blocking — just noting the fix may be more incomplete than the PR description's "all eight" implies, since these are a different symptom (reasoning-echo-policy / vision, not context window) that share the same root cause.

---

PR #6464, opened 2026-07-15
fix(runtime): scope env secret-name heuristic to hand-declared allowlists

### LINKED ISSUE #6458: shell_exec: is_blocked_env_var overrides operator's explicit exec_policy.allowed_env_vars, breaking credentialed CLIs
## Summary

`subprocess_sandbox::sandbox_command` applies the `is_blocked_env_var` name heuristic to the operator's **explicit** `exec_policy.allowed_env_vars` list, silently dropping any variable whose name contains or ends with `KEY` / `SECRET` / `TOKEN` / `PASSWORD` / `CREDENTIAL` / `PRIVATE`.
As a result an operator cannot pass a secret-shaped environment variable to a `shell_exec` child **even when they have explicitly allowlisted it in their own `agent.toml`**.
This makes it impossible to drive any credentialed CLI through `shell_exec` when that CLI reads its secret from an env var — a very common pattern (keyring-unlock passwords, API tokens, etc.).

Introduced in #6439 (commit `76fe32db2`).
Before that PR, `sandbox_command` re-added every name in `allowed_env_vars` unconditionally.

## Repro (minimal, synthetic)

`agent.toml`:

```toml
[exec_policy]
mode = "allowlist"
allowed_commands = ["printenv"]
allowed_env_vars = ["MYTOOL_KEYRING_PASSWORD"]
```

Start the daemon with `MYTOOL_KEYRING_PASSWORD=hunter2` in its environment, then have the agent run:

```
shell_exec: printenv MYTOOL_KEYRING_PASSWORD
```

**Expected:** the child sees `MYTOOL_KEYRING_PASSWORD=hunter2` (the operator explicitly allowlisted it).
**Actual:** the variable is absent; the daemon logs

```
WARN refusing to inject reserved/secret env var into sandboxed child var=MYTOOL_KEYRING_PASSWORD
```

Any name ending in a blocked word reproduces it — e.g. a real-world case is the `gogcli` file-keyring password env var (`GOG_KEYRING_PASSWORD`), which the CLI requires in non-interactive use; the CLI then fails with `no TTY available for keyring file backend password prompt; set GOG_KEYRING_PASSWORD` despite the operator having set and allowlisted it.

## Root cause

`crates/librefang-runtime/src/subprocess_sandbox.rs`:

- `is_blocked_env_var` (≈`:69`) matches `ends_with(sub)` **or** a word-boundary substring against `["KEY","SECRET","TOKEN","PASSWORD","CREDENTIAL","PRIVATE"]` (≈`:93-95`). So `MYTOOL_KEYRING_PASSWORD` / `FOO_API_KEY` / `BAR_TOKEN` all match.
- `sandbox_command` (≈`:149`) now filters the caller-supplied `allowed_env_vars` through `is_blocked_env_var` (≈`:170-181`) and `continue`s past any match.

The `allowed_env_vars` handed to `sandbox_command` on the `shell_exec` path comes from the operator's own `exec_policy.allowed_env_vars` (`tool_runner/dispatch.rs` `effective_allowed_env_vars`, ≈`:934` → `tool_runner/shell.rs::tool_shell_exec` ≈`:130`).
That is an intentional, operator-controlled grant in the operator's own `agent.toml` — categorically different from an attacker-controlled marketplace-hand `[[requires]]` list, which is the threat the function's SECURITY doc-comment actually describes.

## Why the blanket heuristic is wrong here

`allowed_env_vars` exists precisely so an operator can hand a specific secret to a specific tool.
Filtering that list through a broad name heuristic defeats the feature's only purpose: essentially every credential env var contains one of the blocked words, so the allowlist can never carry a credential.
The only feedback is a `WARN` in the daemon log; from the agent's side the CLI simply fails with a confusing downstream error, so this is hard to diagnose.

## Suggested fix

Keep the always-block list (`BLOCKED_ENV_EXACT` — the daemon's own provider keys, `LIBREFANG_VAULT_KEY`, etc.) unconditional, but do **not** apply the broad substring/`ends_with` heuristic to names the operator explicitly placed in `exec_policy.allowed_env_vars`.
Options:

1. Split the check: `BLOCKED_ENV_EXACT` always blocks (defends against smuggling the daemon's own secrets); the heuristic substring/suffix rule applies only to untrusted passthrough sources (marketplace-hand `[[requires]]`), not to the operator's `exec_policy.allowed_env_vars`.
2. Or pass a `trusted: bool` (operator-declared vs. hand-declared) into `sandbox_command` and skip the heuristic when `trusted`.
3. At minimum, surface the drop to the agent/operator (not just a `WARN`), and document that secret-shaped names can never be allowlisted — but (1) is the correct behavior.

Happy to send a PR for option (1) if that direction sounds right.

neo-wanderer:
Fixes #6458.

## Problem

`subprocess_sandbox::sandbox_command` filters every caller-supplied `allowed_env_vars` name through the `is_blocked_env_var` credential-word heuristic (introduced in #6439), including the operator's own `exec_policy.allowed_env_vars` from their agent.toml / config.toml.
Essentially every credential env var name contains `KEY` / `TOKEN` / `PASSWORD` / `SECRET`, so an operator could not hand any credential to a `shell_exec` or `process_start` child even after explicitly allowlisting it — which is the feature's only purpose.
The only feedback was a daemon-side WARN; the agent just saw the CLI fail downstream.

## Fix (option 1 from the issue)

`sandbox_command` now takes the allowlist's provenance as an explicit `EnvAllowlistSource` (new enum in `librefang-types`):

- `OperatorConfig` — a non-hand agent's `exec_policy.allowed_env_vars`, or the global config policy: the broad heuristic is skipped; only the reserved daemon secrets (`BLOCKED_ENV_EXACT`: vault key, provider keys) remain unconditionally blocked.
- `HandDeclared` — marketplace-hand `[[requires]]` / settings passthrough, or any field of a hand-authored manifest (including its `exec_policy`, which a HAND.toml fully controls): the full `is_blocked_env_var` check is kept, preserving every #6439 exfiltration defence.

Provenance is decided where the manifest is known — `manifest.is_hand` in the agent loop (`tool_call.rs`) and the `/mcp` bridge (`routes/network.rs`) — and threaded through `ToolExecContext` into `tool_shell_exec`, `tool_process_start` / `ProcessManager::start`, and `LocalBackend`.
It is persisted on `DeferredToolExecution` with a `HandDeclared` serde default, so approval payloads written before this field existed resume on the strict posture, and an approved resume filters the child env exactly as the live path would have.
`is_blocked_env_var` itself is unchanged (the WASM `env_read` host function and the kernel hand-activation filter keep the strict posture); the exact-name check is split out as a new `is_reserved_env_var`.

## Changes

- `librefang-types`: new `EnvAllowlistSource` enum (serde default `HandDeclared`); `DeferredToolExecution.env_allowlist_source` field with `#[serde(default)]`.
- `librefang-runtime`: `sandbox_command(cmd, vars, source)`; new `is_reserved_env_var`; provenance plumbed through `ToolExecContext`, `execute_tool_with_sender_account`, `tool_shell_exec`, `tool_process_start`, `ProcessManager::start`, `LocalBackend` / `build_backend`; the `execute_tool` shim passes `OperatorConfig` (its only production caller is the operator API session path with the global config policy).
- `librefang-kernel`: approval-resume `ToolExecContext` restores the persisted provenance.
- `librefang-api`: the `/mcp` bridge derives the source from the resolved caller entry's `manifest.is_hand` (strict default when the header is absent).

## Verification

- New regression tests: `test_sandbox_command_operator_allowlist_passes_secret_shaped_vars` (operator-allowlisted `*_KEYRING_PASSWORD` reaches the child env; `LIBREFANG_VAULT_KEY` still stripped), `test_is_reserved_env_var_exact_names_only`, and a legacy-payload serde-default assertion in `test_deferred_tool_execution_serialization`.
- Existing strict-posture test updated to pass `HandDeclared` and unchanged in its assertions.
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- `cargo fmt --all --check`: clean.
- Scoped tests green: `librefang-types --lib` (874), `librefang-runtime --lib` (2099), `librefang-kernel --lib approval` (98), `librefang-kernel --test rbac_m3_evaluate_tool_call` (9), `librefang-runtime --test tool_exec_backend_selection` (10), `librefang-api --test tool_exec_backend_selection` (13).


houko (changes requested):
Thanks for the thorough work here — the provenance plumbing is clean and the `is_reserved_env_var` split, the `EnvAllowlistSource` serde default for legacy deferred payloads, and the two regression tests are all exactly the right shape for this fix.

I'm requesting changes on one structural point rather than any defect in the code itself: **this PR overlaps with #6465, which is the maintainer's implementation of the same fix for the same issue.**

Both PRs are `Fixes #6458`, both split out `is_reserved_env_var`, both scope the credential-word heuristic by trust level, and both change the signature of the single function `subprocess_sandbox::sandbox_command`. That last point makes them mutually exclusive — merging either one forces a conflict on the other. #6458 is already closed by #6465 (commit `15284f0953`), and #6465 is rebased onto current `main` with CI running.

The two designs differ in how they carry provenance into `sandbox_command`:

- **#6465** takes two separate lists — `sandbox_command(cmd, operator_env_vars, untrusted_env_vars)` — so a single call can filter an operator list and a hand list side by side, each under its own rule, and it returns the refused names so the drop can be surfaced to the agent instead of only a daemon-log WARN.
- **#6464 (this PR)** takes one list plus an `EnvAllowlistSource` tag, so a given call's list is uniformly one trust level.

Given #6465 is the maintainer's own branch and already owns/closed the issue, I'd recommend closing this PR (or repointing the review effort at #6465) to avoid the two competing on the same function. Flagging for @houko to make the call rather than closing it myself.

One thing this PR does that #6465 currently does **not**, and that is worth preserving regardless of which branch lands: you persist the provenance on the deferred-approval path — `DeferredToolExecution.env_allowlist_source` with a `HandDeclared` serde default (`crates/librefang-types/src/tool.rs`), rehydrated in `kernel/mod.rs:1778`. That closes the approve-resume gap so a deferred `shell_exec` filters its child env identically to the live path. #6465 threads the source through the live path but not through `DeferredToolExecution`, so on the resume path it would fall back to the strict posture. If #6465 is the branch that lands, that persistence should be carried over into it — either by you in a follow-up or folded into #6465. Either way the coverage shouldn't be lost.

houko:
Correction to my earlier review — I need to withdraw one of the two points I made.

I claimed #6465 threads the source through the live path "but not through `DeferredToolExecution`, so on the resume path it would fall back to the strict posture." That is wrong, and I verified it against the merged code (#6465 landed as `627fa3cf4`).

#6465 does not carry a provenance tag at all — it carries trust in *which field* holds the name: the operator's list is `exec_policy.allowed_env_vars`, the untrusted hand passthrough list is `allowed_env_vars`. Both are already persisted on `DeferredToolExecution`, and `build_deferred_tool_exec_context` (`crates/librefang-kernel/src/kernel/mod.rs:1780` and `:1784`) feeds `deferred.allowed_env_vars` and `deferred.exec_policy` straight back into the same `tool_shell_exec` / `tool_process_start` code the live path uses. So a resumed `Allow once` filters the child env identically to the live path with no extra persisted field — there is no strict-posture fallback on resume.

I confirmed this end to end with a round-trip regression test (build a deferred payload → serde round-trip across the restart boundary → drive `sandbox_command` as the resume path does): the operator secret-shaped name survives and reaches the child, the hand passthrough secret-shaped name stays blocked, and the reserved daemon key is refused from either list. That test is up as #6467.

So the `env_allowlist_source` field this PR adds is not needed to close a resume gap under #6465's two-list model — it would be redundant with the field the list already lives in. My apologies for the inaccurate framing; that was a wrong inference from #6465 not touching `tool.rs`, which I should have checked against the resume path before writing.

The structural point stands unchanged: #6465 (the maintainer's branch, now merged) already resolved #6458, so this PR is superseded. Recommending closure remains with @houko.

crates/librefang-runtime/src/subprocess_sandbox.rs
```diff
@@ -95,6 +97,18 @@ pub fn is_blocked_env_var(name: &str) -> bool {
         .any(|sub| upper.ends_with(sub) || has_word_boundary_substring(&upper, sub))
 }
 
+/// Returns `true` if the env var name is one of the reserved daemon secrets
+/// (`BLOCKED_ENV_EXACT`) — the vault key and the daemon's own provider
+/// credentials.
+/// These are unconditionally withheld from every sandboxed child, no matter
+/// who allowlisted them; unlike [`is_blocked_env_var`] this does NOT apply
+/// the credential-word heuristic, so an operator-declared secret-shaped name
+/// (`MYTOOL_KEYRING_PASSWORD`) is not caught by it (#6458).
```
houko, line 106: CLAUDE.md's prose-wrapping rule applies to new doc-comments too: one sentence per line, no hard-wrap at ~70-80 cols. Suggested rewrap:
```rust
/// Returns `true` if the env var name is one of the reserved daemon secrets (`BLOCKED_ENV_EXACT`) — the vault key and the daemon's own provider credentials.
/// These are unconditionally withheld from every sandboxed child, no matter who allowlisted them; unlike [`is_blocked_env_var`] this does NOT apply the credential-word heuristic, so an operator-declared secret-shaped name (`MYTOOL_KEYRING_PASSWORD`) is not caught by it (#6458).
```

---

crates/librefang-runtime/src/subprocess_sandbox.rs
```diff
@@ -166,11 +195,18 @@ pub fn sandbox_command(cmd: &mut tokio::process::Command, allowed_env_vars: &[St
 
     // Re-add caller-specified allowed vars — but never a reserved secret or
     // internal-daemon var, even if the caller (e.g. an attacker-controlled
-    // HAND.toml passthrough list) explicitly asked for it.
+    // HAND.toml passthrough list) explicitly asked for it. Hand-declared
+    // lists additionally pass the secret-name heuristic; operator-declared
+    // lists are an intentional grant and skip it (#6458).
```
houko, line 200: Same prose-wrapping rule (one sentence per line) applies to this whole new doc block and the touched inline comment below it.

For the plain-sentence parts, straightforward rewrap:
```rust
/// SECURITY: caller-allowed names are re-filtered here as a last line of defence, scoped by `source` (#6458):
///
/// - [`EnvAllowlistSource::HandDeclared`] — the full [`is_blocked_env_var`] check. A marketplace hand's attacker-controlled `HAND.toml` can name arbitrary vars in `[[requires]]` (or in a hand-authored manifest's `exec_policy`), so even though the kernel filters that list at assembly time, we re-check here so no such caller can smuggle a secret-shaped var from the daemon's live environment into an untrusted child.
/// - [`EnvAllowlistSource::OperatorConfig`] — only the reserved exact-name list ([`is_reserved_env_var`]). The operator's own `exec_policy.allowed_env_vars` exists precisely to hand a specific credential to a specific tool; essentially every credential name matches the broad heuristic, so applying it here made the allowlist unable to carry a credential at all. The daemon's own secrets (`LIBREFANG_VAULT_KEY`, `ANTHROPIC_API_KEY`, …) stay blocked unconditionally.
```
(marking this part ambiguous only in the sense that CLAUDE.md's rule doesn't explicitly address bulleted lists — I read "one sentence = one line" as applying within each bullet, as shown above, but flagging in case you'd rather keep bullets terser.)

And for the inline comment right above the `for` loop (touched by this PR — a new sentence was appended to the existing one):
```rust
    // Re-add caller-specified allowed vars — but never a reserved secret or internal-daemon var, even if the caller (e.g. an attacker-controlled HAND.toml passthrough list) explicitly asked for it.
    // Hand-declared lists additionally pass the secret-name heuristic; operator-declared lists are an intentional grant and skip it (#6458).
```

---

crates/librefang-types/src/config/types.rs
```diff
@@ -2155,6 +2155,38 @@ impl Default for ExecPolicy {
...
+    OperatorConfig,
+    /// Names declared by a third-party artifact — a marketplace hand's
+    /// `[[requires]]` / settings passthrough, or any field of a hand-authored
+    /// agent manifest (including its `exec_policy`).
+    /// Untrusted: the full secret-name heuristic applies so a hand cannot
+    /// exfiltrate secret-shaped vars from the daemon's live environment.
+    /// This is the serde default so that deferred tool executions persisted
+    /// before this field existed resume with the strict posture.
+    #[default]
+    HandDeclared,
```
houko, line 2187: CLAUDE.md prose-wrapping rule (one sentence per line) — this new enum's doc-comments hard-wrap at ~80 cols. Suggested rewrap:
```rust
/// Provenance of an env-var allowlist handed to the subprocess sandbox.
///
/// Decides how strictly `subprocess_sandbox::sandbox_command` filters the caller-supplied `allowed_env_vars` before re-injecting them into a child process (#6458).
/// The reserved exact-name blocklist (the daemon's own vault key / provider keys) applies to BOTH variants; only the broad credential-word heuristic (`KEY` / `TOKEN` / `PASSWORD` / …) is scoped by provenance.
#[derive(
    Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize, schemars::JsonSchema,
)]
#[serde(rename_all = "snake_case")]
pub enum EnvAllowlistSource {
    /// Names the operator wrote into their own configuration — a non-hand agent's `exec_policy.allowed_env_vars` (agent.toml) or the global `config.toml` exec policy.
    /// The operator controls both the daemon environment and the allowlist, so an explicitly granted secret-shaped name (e.g. `MYTOOL_KEYRING_PASSWORD`) is an intentional grant and passes through; only the reserved daemon secrets are withheld.
    OperatorConfig,
    /// Names declared by a third-party artifact — a marketplace hand's `[[requires]]` / settings passthrough, or any field of a hand-authored agent manifest (including its `exec_policy`).
    /// Untrusted: the full secret-name heuristic applies so a hand cannot exfiltrate secret-shaped vars from the daemon's live environment.
    /// This is the serde default so that deferred tool executions persisted before this field existed resume with the strict posture.
    #[default]
    HandDeclared,
}
```

---

crates/librefang-types/src/tool.rs
```diff
@@ -148,6 +148,13 @@ pub struct DeferredToolExecution {
     pub allowed_tools: Option<Vec<String>>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub allowed_env_vars: Option<Vec<String>>,
+    /// Provenance of `allowed_env_vars` (and the `exec_policy` fallback list),
+    /// captured at defer time so the resume path filters the re-injected env
+    /// exactly as the live path would have (#6458).
+    /// The serde default (`HandDeclared`) keeps payloads persisted before this
+    /// field existed on the strict posture.
```
houko, line 155: CLAUDE.md prose-wrapping rule — one sentence per line:
```rust
    /// Provenance of `allowed_env_vars` (and the `exec_policy` fallback list), captured at defer time so the resume path filters the re-injected env exactly as the live path would have (#6458).
    /// The serde default (`HandDeclared`) keeps payloads persisted before this field existed on the strict posture.
```

---

crates/librefang-types/src/tool.rs
```diff
@@ -1380,6 +1388,19 @@ mod tests {
         let json = serde_json::to_string(&deferred).unwrap();
         let deserialized: DeferredToolExecution = serde_json::from_str(&json).unwrap();
         assert_eq!(deserialized.agent_id, "agent-1");
+        assert_eq!(
+            deserialized.env_allowlist_source,
+            crate::config::EnvAllowlistSource::OperatorConfig
+        );
+        // Payloads persisted before the field existed must deserialize to the
```
houko, line 1395: Same rule — this is one sentence, join it onto one line:
```rust
        // Payloads persisted before the field existed must deserialize to the strict default (#6458).
```

---

crates/librefang-runtime/src/tool_runner/dispatch.rs
```diff
@@ -51,6 +51,13 @@ pub struct ToolExecContext<'a> {
     pub web_ctx: Option<&'a WebToolsContext>,
     pub browser_ctx: Option<&'a crate::browser::BrowserManager>,
     pub allowed_env_vars: Option<&'a [String]>,
+    /// Provenance of the env allowlist in effect for this call —
+    /// `allowed_env_vars` when `Some`, else the `exec_policy` fallback list.
+    /// `OperatorConfig` only when the calling agent's manifest is
+    /// operator-authored (`!manifest.is_hand`); hand-authored manifests are
+    /// third-party input, so both their passthrough list and their
+    /// `exec_policy` stay on the strict heuristic (#6458).
```
houko, line 59: CLAUDE.md prose-wrapping rule — one sentence per line:
```rust
    /// Provenance of the env allowlist in effect for this call — `allowed_env_vars` when `Some`, else the `exec_policy` fallback list.
    /// `OperatorConfig` only when the calling agent's manifest is operator-authored (`!manifest.is_hand`); hand-authored manifests are third-party input, so both their passthrough list and their `exec_policy` stay on the strict heuristic (#6458).
```

---

crates/librefang-runtime/src/tool_runner/dispatch.rs
```diff
@@ -1515,6 +1525,13 @@ pub async fn execute_tool(
         spill_threshold_bytes,
         max_artifact_bytes,
         None,
+        // The only production caller of this shim is the operator API session
+        // path (`routes/tools_sessions.rs`), which passes the operator's own
+        // global `config.toml` exec policy — an operator-authored allowlist.
+        // Hand-agent calls flow through the agent loop, which calls
+        // `execute_tool_with_sender_account` directly with the source derived
+        // from `manifest.is_hand` (#6458).
```
houko, line 1533: CLAUDE.md prose-wrapping rule — one sentence per line:
```rust
        // The only production caller of this shim is the operator API session path (`routes/tools_sessions.rs`), which passes the operator's own global `config.toml` exec policy — an operator-authored allowlist.
        // Hand-agent calls flow through the agent loop, which calls `execute_tool_with_sender_account` directly with the source derived from `manifest.is_hand` (#6458).
```

---

crates/librefang-runtime/src/tool_exec_backend.rs
```diff
@@ -258,6 +258,10 @@ pub struct LocalBackend {
     /// Allowlisted env vars to pass through to spawned children.
     /// Mirrors the existing `tool_runner.rs` `allowed_env` plumbing.
     allowed_env: Vec<String>,
+    /// Provenance of `allowed_env` — decides whether the secret-name
+    /// heuristic applies on top of the reserved exact-name blocklist when
+    /// the child env is scrubbed (#6458).
```
houko, line 263: CLAUDE.md prose-wrapping rule — this is one sentence, join it onto one line:
```rust
    /// Provenance of `allowed_env` — decides whether the secret-name heuristic applies on top of the reserved exact-name blocklist when the child env is scrubbed (#6458).
```

---

crates/librefang-api/src/routes/network.rs
```diff
@@ -1549,6 +1549,16 @@ pub async fn mcp_http(
             cfg.tool_results.spill_threshold_bytes,
             cfg.tool_results.max_artifact_bytes,
             current_account_id.as_deref(), // sender_account_id (X-LibreFang-Current-Account-Id, #6443)
+            // #6458: mirror the agent-loop trust decision — a hand-authored
+            // manifest's env lists stay on the strict heuristic; an
+            // unresolved caller has no env list at all, so the strict
+            // default is a no-op there.
```
houko, line 1555: CLAUDE.md prose-wrapping rule — this is one sentence, join it onto one line:
```rust
            // #6458: mirror the agent-loop trust decision — a hand-authored manifest's env lists stay on the strict heuristic; an unresolved caller has no env list at all, so the strict default is a no-op there.
```

---

crates/librefang-kernel/src/kernel/mod.rs
```diff
@@ -1778,6 +1778,11 @@ impl LibreFangKernel {
             web_ctx: Some(&self.media.web_ctx),
             browser_ctx: Some(&self.media.browser_ctx),
             allowed_env_vars: deferred.allowed_env_vars.as_deref(),
+            // #6458: restore the provenance captured at defer time so the
+            // resumed execution filters the child env exactly as the live
+            // path would have. Pre-existing payloads without the field
+            // deserialize to the strict `HandDeclared` default.
```
houko, line 1784: CLAUDE.md prose-wrapping rule — one sentence per line:
```rust
            // #6458: restore the provenance captured at defer time so the resumed execution filters the child env exactly as the live path would have.
            // Pre-existing payloads without the field deserialize to the strict `HandDeclared` default.
```

---

crates/librefang-runtime/src/agent_loop/tool_call.rs
```diff
@@ -1099,6 +1099,15 @@ pub(super) async fn execute_single_tool_call_core(
             // #6443: thread the turn's originating account for the
             // cross-account dispatch guard in `channel_send`.
             ctx.sender_account_id,
+            // #6458: a hand-authored manifest is third-party input, so both
+            // its `hand_allowed_env` passthrough and its `exec_policy`
+            // allowlist stay on the strict secret-name heuristic; an
+            // operator-authored agent.toml allowlist is an intentional grant.
```
houko, line 1105: CLAUDE.md prose-wrapping rule — this is one sentence, join it onto one line:
```rust
            // #6458: a hand-authored manifest is third-party input, so both its `hand_allowed_env` passthrough and its `exec_policy` allowlist stay on the strict secret-name heuristic; an operator-authored agent.toml allowlist is an intentional grant.
```

---

houko (commented):
Automated CLAUDE.md compliance pass (no logic/security issues found). This PR's new code is otherwise clean against the project's rule set (CHANGELOG attribution present, no HashMap/HashSet ordering issues, no `KernelConfig` fields touched, no AI attribution, no dashboard-layer violations). The one recurring issue: every new multi-line comment/doc-comment in this diff hard-wraps at ~70-80 columns, but CLAUDE.md's prose-wrapping rule requires one sentence per line (applies to doc-comments and any multi-line free-form prose comment block, including new prose in a touched paragraph). Left line-anchored suggestions at each location — all mechanical rewraps, safe to apply as-is except the bulleted list in `sandbox_command`'s doc, flagged as a judgment call.

---

crates/librefang-api/src/routes/network.rs
```diff
@@ -1549,6 +1549,16 @@ pub async fn mcp_http(
...
+            match caller_entry.as_ref() {
+                Some(e) if !e.manifest.is_hand => {
+                    librefang_types::config::EnvAllowlistSource::OperatorConfig
+                }
+                _ => librefang_types::config::EnvAllowlistSource::HandDeclared,
+            },
```
houko, line 1561: CLAUDE.md's "MANDATORY: Integration Testing" section requires a `#[tokio::test]` against `TestServer` for any route/wiring change, specifically to catch things like kernel↔API type drift and un-deserialized/mis-derived fields at the boundary — this `match caller_entry.as_ref() { Some(e) if !e.manifest.is_hand => OperatorConfig, _ => HandDeclared }` branch is exactly that kind of security-relevant wiring (it decides whether the credential-word heuristic applies to a caller's env allowlist over the `/mcp` bridge) and has no test anywhere in this 23-file diff exercising it.

I checked whether an existing test incidentally covers it: `crates/librefang-api/tests/mcp_tools_list_allowlist_test.rs` (the file that would plausibly own this) has zero references to `EnvAllowlistSource`, `is_hand`, or `allowed_env`, and no other file in `crates/librefang-api/tests/` does either. The PR's own "Verification" list only names `librefang-runtime`/`librefang-types`/`librefang-kernel` unit and approval tests plus `tool_exec_backend_selection` — nothing that spins up the `/mcp` route itself.

Given this is the fail-secure branch (an unresolved or hand-authored caller must land on the strict `HandDeclared` default, an operator-authored caller must land on `OperatorConfig`), please add a `#[tokio::test]` that drives `POST /mcp` twice — once through a resolved hand-authored caller entry and once through an operator-authored one — and asserts the resulting child-env filtering behavior differs, per the pattern in `start_test_server()` / `tests/api_integration_test.rs`.

---

houko (commented):
Automated daily review pass. Checked existing reviews/comments first (1 CHANGES_REQUESTED review, 1 prior automated compliance-pass review, 11 existing line comments, 1 follow-up correction comment) to avoid duplication — all prose-wrapping violations and the #6465 supersession issue are already flagged, so not repeating those. Verified independently that `EnvAllowlistSource` correctly derives `Default` to `HandDeclared` (fail-secure) and that `DeferredToolExecution.env_allowlist_source` is threaded through every construction site (test and production) with `#[serde(default)]` — no silent-default regression there. One new, non-duplicate finding posted inline: the `/mcp` bridge provenance-derivation branch in `routes/network.rs` has no accompanying integration test per CLAUDE.md's mandatory testing rule for route/wiring changes.

---

crates/librefang-runtime/src/subprocess_sandbox.rs
houko, line 106: Prose-wrapping rule (CLAUDE.md): new doc-comments must break only at sentence boundaries, one sentence per line, no column-width wrap. This new `is_reserved_env_var` doc comment hard-wraps mid-sentence. Suggested rewrap (two sentences → two lines):

```
/// Returns `true` if the env var name is one of the reserved daemon secrets (`BLOCKED_ENV_EXACT`) — the vault key and the daemon's own provider credentials.
/// These are unconditionally withheld from every sandboxed child, no matter who allowlisted them; unlike [`is_blocked_env_var`] this does NOT apply the credential-word heuristic, so an operator-declared secret-shaped name (`MYTOOL_KEYRING_PASSWORD`) is not caught by it (#6458).
```

Same pattern recurs at several other locations touched by this PR (subprocess_sandbox.rs:157, subprocess_sandbox.rs:196, subprocess_sandbox.rs:1966, subprocess_sandbox.rs:1977, subprocess_sandbox.rs:2013, config/types.rs:2158, config/types.rs:2171, config/types.rs:2179, tool.rs:151, tool.rs:1395, tool_call.rs:1102, network.rs:1552, dispatch.rs:54, dispatch.rs:1528, tool_exec_backend.rs:261, kernel/mod.rs:1781) — leaving one comment per hotspot rather than one per line.

---

crates/librefang-runtime/src/subprocess_sandbox.rs
```diff
@@ -140,13 +154,28 @@ fn has_word_boundary_substring(haystack: &str, needle: &str) -> bool {
...
+///   time, we re-check here so no such caller can smuggle a secret-shaped
+///   var from the daemon's live environment into an untrusted child.
+/// - [`EnvAllowlistSource::OperatorConfig`] — only the reserved exact-name
+///   list ([`is_reserved_env_var`]). The operator's own
+///   `exec_policy.allowed_env_vars` exists precisely to hand a specific
+///   credential to a specific tool; essentially every credential name matches
+///   the broad heuristic, so applying it here made the allowlist unable to
+///   carry a credential at all. The daemon's own secrets
+///   (`LIBREFANG_VAULT_KEY`, `ANTHROPIC_API_KEY`, …) stay blocked
+///   unconditionally.
```
houko, line 173: Same prose-wrap issue as the comment on line 100-106 — the `SECURITY:` doc comment on `sandbox_command` (and its two bullet items) is hard-wrapped mid-sentence. Suggested rewrap:

```
/// SECURITY: caller-allowed names are re-filtered here as a last line of defence, scoped by `source` (#6458):
///
/// - [`EnvAllowlistSource::HandDeclared`] — the full [`is_blocked_env_var`] check.
///   A marketplace hand's attacker-controlled `HAND.toml` can name arbitrary vars in `[[requires]]` (or in a hand-authored manifest's `exec_policy`), so even though the kernel filters that list at assembly time, we re-check here so no such caller can smuggle a secret-shaped var from the daemon's live environment into an untrusted child.
/// - [`EnvAllowlistSource::OperatorConfig`] — only the reserved exact-name list ([`is_reserved_env_var`]).
///   The operator's own `exec_policy.allowed_env_vars` exists precisely to hand a specific credential to a specific tool; essentially every credential name matches the broad heuristic, so applying it here made the allowlist unable to carry a credential at all.
///   The daemon's own secrets (`LIBREFANG_VAULT_KEY`, `ANTHROPIC_API_KEY`, …) stay blocked unconditionally.
```

---

crates/librefang-runtime/src/subprocess_sandbox.rs
houko, line 200: Prose-wrap: this loop comment is hard-wrapped mid-sentence. Suggested:

```
    // Re-add caller-specified allowed vars — but never a reserved secret or internal-daemon var, even if the caller (e.g. an attacker-controlled HAND.toml passthrough list) explicitly asked for it.
    // Hand-declared lists additionally pass the secret-name heuristic; operator-declared lists are an intentional grant and skip it (#6458).
```

---

crates/librefang-runtime/src/subprocess_sandbox.rs
```diff
@@ -1924,4 +1960,67 @@ mod tests {
...
+    #[test]
+    fn test_sandbox_command_operator_allowlist_passes_secret_shaped_vars() {
+        // Regression for #6458: the operator's own `exec_policy.allowed_env_vars`
+        // is an intentional grant in the operator's agent.toml, not an
+        // attacker-controlled passthrough list. Filtering it through the broad
+        // secret-name heuristic made it impossible to hand any credential to a
+        // `shell_exec` child (essentially every credential name contains KEY /
+        // TOKEN / PASSWORD / …), defeating the feature's only purpose.
+        // Operator-declared names skip the heuristic; the reserved exact-name
+        // blocklist (the daemon's own secrets) still applies unconditionally.
```
houko, line 1973: Prose-wrap: test comment hard-wrapped mid-sentence. Suggested:

```
        // Regression for #6458: the operator's own `exec_policy.allowed_env_vars` is an intentional grant in the operator's agent.toml, not an attacker-controlled passthrough list.
        // Filtering it through the broad secret-name heuristic made it impossible to hand any credential to a `shell_exec` child (essentially every credential name contains KEY / TOKEN / PASSWORD / …), defeating the feature's only purpose.
        // Operator-declared names skip the heuristic; the reserved exact-name blocklist (the daemon's own secrets) still applies unconditionally.
```

And a couple lines below (the `SAFETY:` comment right before `set_var`) has the same issue:
```
        // SAFETY: the SANDBOXTEST_* name is unique to this test; the vault key is set and removed here and not read elsewhere in-test.
```

---

crates/librefang-runtime/src/subprocess_sandbox.rs
```diff
@@ -1924,4 +1960,67 @@ mod tests {
...
+            "reserved daemon secret must be stripped even from an operator allowlist"
+        );
+    }
+
+    #[test]
+    fn test_is_reserved_env_var_exact_names_only() {
+        // The reserved list blocks by exact (case-insensitive) name and does
+        // NOT apply the credential-word heuristic — that split is what lets an
+        // operator allowlist a secret-shaped var while the daemon's own
+        // secrets stay unreachable from any child.
```
houko, line 2016: Prose-wrap: same issue. Suggested single-line rewrap:

```
        // The reserved list blocks by exact (case-insensitive) name and does NOT apply the credential-word heuristic — that split is what lets an operator allowlist a secret-shaped var while the daemon's own secrets stay unreachable from any child.
```

---

crates/librefang-types/src/config/types.rs
```diff
@@ -2155,6 +2155,38 @@ impl Default for ExecPolicy {
     }
 }
 
+/// Provenance of an env-var allowlist handed to the subprocess sandbox.
+///
+/// Decides how strictly `subprocess_sandbox::sandbox_command` filters the
+/// caller-supplied `allowed_env_vars` before re-injecting them into a child
+/// process (#6458).
+/// The reserved exact-name blocklist (the daemon's own vault key / provider
+/// keys) applies to BOTH variants; only the broad credential-word heuristic
```
houko, line 2164: Prose-wrap (CLAUDE.md: doc-comments break only at sentence boundaries, one sentence per line): this new `EnvAllowlistSource` doc comment and its two variant docs (lines ~2171-2186) are hard-wrapped mid-sentence throughout. Suggested rewrap for the type-level doc:

```
/// Provenance of an env-var allowlist handed to the subprocess sandbox.
///
/// Decides how strictly `subprocess_sandbox::sandbox_command` filters the caller-supplied `allowed_env_vars` before re-injecting them into a child process (#6458).
/// The reserved exact-name blocklist (the daemon's own vault key / provider keys) applies to BOTH variants; only the broad credential-word heuristic (`KEY` / `TOKEN` / `PASSWORD` / …) is scoped by provenance.
```

And for the two variants:
```
    /// Names the operator wrote into their own configuration — a non-hand agent's `exec_policy.allowed_env_vars` (agent.toml) or the global `config.toml` exec policy.
    /// The operator controls both the daemon environment and the allowlist, so an explicitly granted secret-shaped name (e.g. `MYTOOL_KEYRING_PASSWORD`) is an intentional grant and passes through; only the reserved daemon secrets are withheld.
    OperatorConfig,
    /// Names declared by a third-party artifact — a marketplace hand's `[[requires]]` / settings passthrough, or any field of a hand-authored agent manifest (including its `exec_policy`).
    /// Untrusted: the full secret-name heuristic applies so a hand cannot exfiltrate secret-shaped vars from the daemon's live environment.
    /// This is the serde default so that deferred tool executions persisted before this field existed resume with the strict posture.
```

---

crates/librefang-types/src/tool.rs
houko, line 155: Prose-wrap: hard-wrapped mid-sentence. Suggested:

```
    /// Provenance of `allowed_env_vars` (and the `exec_policy` fallback list), captured at defer time so the resume path filters the re-injected env exactly as the live path would have (#6458).
    /// The serde default (`HandDeclared`) keeps payloads persisted before this field existed on the strict posture.
```

Same issue at line 1395 (test comment) — one-line fix: `// Payloads persisted before the field existed must deserialize to the strict default (#6458).`

---

crates/librefang-runtime/src/agent_loop/tool_call.rs
houko, line 1105: Prose-wrap: hard-wrapped mid-sentence. Suggested single-line rewrap:

```
            // #6458: a hand-authored manifest is third-party input, so both its `hand_allowed_env` passthrough and its `exec_policy` allowlist stay on the strict secret-name heuristic; an operator-authored agent.toml allowlist is an intentional grant.
```

---

crates/librefang-api/src/routes/network.rs
houko, line 1555: Prose-wrap: hard-wrapped mid-sentence. Suggested single-line rewrap:

```
            // #6458: mirror the agent-loop trust decision — a hand-authored manifest's env lists stay on the strict heuristic; an unresolved caller has no env list at all, so the strict default is a no-op there.
```

---

crates/librefang-runtime/src/tool_runner/dispatch.rs
houko, line 59: Prose-wrap: hard-wrapped mid-sentence. Suggested:

```
    /// Provenance of the env allowlist in effect for this call — `allowed_env_vars` when `Some`, else the `exec_policy` fallback list.
    /// `OperatorConfig` only when the calling agent's manifest is operator-authored (`!manifest.is_hand`); hand-authored manifests are third-party input, so both their passthrough list and their `exec_policy` stay on the strict heuristic (#6458).
```

Same issue at line ~1528-1533 in this file:
```
        // The only production caller of this shim is the operator API session path (`routes/tools_sessions.rs`), which passes the operator's own global `config.toml` exec policy — an operator-authored allowlist.
        // Hand-agent calls flow through the agent loop, which calls `execute_tool_with_sender_account` directly with the source derived from `manifest.is_hand` (#6458).
```

---

crates/librefang-runtime/src/tool_exec_backend.rs
houko, line 263: Prose-wrap: hard-wrapped mid-sentence. Suggested single-line rewrap:

```
    /// Provenance of `allowed_env` — decides whether the secret-name heuristic applies on top of the reserved exact-name blocklist when the child env is scrubbed (#6458).
```

---

crates/librefang-kernel/src/kernel/mod.rs
houko, line 1784: Prose-wrap: hard-wrapped mid-sentence. Suggested:

```
            // #6458: restore the provenance captured at defer time so the resumed execution filters the child env exactly as the live path would have.
            // Pre-existing payloads without the field deserialize to the strict `HandDeclared` default.
```

---

houko (commented):
Automated review pass on the env-allowlist-provenance fix.

**Security logic (the main point of this PR): looks sound.** Traced the trust boundary end to end:
- `is_reserved_env_var` / `is_blocked_env_var` split in `subprocess_sandbox.rs` is correct and the two new tests (`test_sandbox_command_operator_allowlist_passes_secret_shaped_vars`, `test_is_reserved_env_var_exact_names_only`) cover both the "was too broad, now scoped" case and the "still blocks the daemon's own secrets" regression case.
- Checked whether `ctx.manifest.is_hand` (the bit both `tool_call.rs` and `network.rs` key the provenance decision on) can ever be `false` while the manifest's `exec_policy` / `hand_allowed_env` metadata is nonetheless hand-authored (which would be a real bypass). Traced `hands_lifecycle.rs::activate_hand_with_id` and the boot-time re-sync in `boot.rs`: `manifest.is_hand = true` and the `hand_allowed_env` metadata key are only ever set together, on the same manifest object, for agents actually spawned by a hand. Per-agent `exec_policy` overrides from `config.toml`'s `[agents.<name>]` table are separately confirmed inert (`detect_misplaced_per_agent_overrides`), and a `HAND.toml`'s own `[agents.<role>]` table only configures the hand's own spawned roles (which get `is_hand = true`). So I could not construct a path where a non-hand (operator) agent's exec_policy/env list is actually hand-controlled — the `OperatorConfig` classification looks safe.
- Approval-resume path (`kernel/mod.rs`) and the `/mcp` bridge (`network.rs`) both correctly restore/derive the provenance rather than defaulting permissively.
- CHANGELOG entry has `(@neo-wanderer)` attribution and no AI/Claude mentions — fine as-is.

**Style-only findings (left as review comments, not committed):** every new/touched doc-comment and comment block in this diff is hard-wrapped at ~80 columns, which conflicts with this repo's CLAUDE.md prose rule ("no column limit; break only at sentence boundaries, one sentence per line"). Left one comment per hotspot with the suggested single-line rewrap (subprocess_sandbox.rs ×6, config/types.rs, tool.rs ×2, tool_call.rs, network.rs, dispatch.rs ×2, tool_exec_backend.rs, kernel/mod.rs). These are mechanical (no logic change) — happy to push the rewrap directly, but I'm holding off on committing anything in this session due to a git-identity issue flagged separately to the maintainer (this environment's global git author/committer resolves to an AI identity), so leaving them as suggestions here instead.

No functional/security changes requested — this is a comment-only review, not a request for changes.

---

houko:
Verified the base/ancestry relationship to #6465 directly against git history, to complement the code-level analysis in the existing comment above.

This PR's stated base (`00d1bd61c5847e799dbcf2125e8ef9bb81b83837`) is exactly `git merge-base pr-6464 origin/main` — confirmed by fetching `refs/pull/6464/head` and diffing against `origin/main`. That commit is `fix(security/channels): enforce cross-account channel_send guard through the /mcp bridge (#6443) (#6455)`, dated 2026-07-15 09:48 JST.

#6465 merged to `main` as `627fa3cf4f65ba162a53703b32c8b6e3bbd9ef23` (`fix(runtime): trust operator env allowlist in sandbox_command (#6458) (#6465)`), dated 2026-07-15 13:02 JST — after this PR's base commit.

`git merge-base --is-ancestor 627fa3cf4 <base>` → false. `git merge-base --is-ancestor <base> 627fa3cf4` → true. So this PR's base predates the #6465 merge; #6464 was branched before #6465 landed, and 627fa3cf4 sits on `main` strictly between #6464's base and the current tip (`git log --oneline <base>..origin/main` shows exactly one match for #6465/627fa3cf4, no others touching #6458).

Net: #6465 already closed #6458 on `main` via the two-list model (`exec_policy.allowed_env_vars` for the operator, `allowed_env_vars` for untrusted hand passthrough), and #6467 adds the regression test proving the defer→approve→resume path is already correct under that model without needing a provenance-tag field. This PR proposes the same fix again via a third, now-superseded mechanism (`EnvAllowlistSource` + `DeferredToolExecution.env_allowlist_source`) on a base that doesn't have #6465. Consistent with that, GitHub currently reports this PR's `mergeable_state` as `dirty` with a `has-conflicts` label, which is what's expected once `main` has moved past this PR's base with an overlapping fix.

Recommend @houko (or whoever owns triage here) confirm explicitly whether this PR should be closed as superseded by #6465, given #6467 is now the follow-up locking in the same resume guarantee this PR was trying to add a field for.

(Not re-litigating the inline review comments already on this PR — the CLAUDE.md prose-wrap findings there stand regardless of the outcome here.)

---

crates/librefang-runtime/src/subprocess_sandbox.rs
```diff
@@ -95,6 +97,18 @@ pub fn is_blocked_env_var(name: &str) -> bool {
...
 }
 
+/// Returns `true` if the env var name is one of the reserved daemon secrets
+/// (`BLOCKED_ENV_EXACT`) — the vault key and the daemon's own provider
+/// credentials.
+/// These are unconditionally withheld from every sandboxed child, no matter
+/// who allowlisted them; unlike [`is_blocked_env_var`] this does NOT apply
+/// the credential-word heuristic, so an operator-declared secret-shaped name
+/// (`MYTOOL_KEYRING_PASSWORD`) is not caught by it (#6458).
+pub fn is_reserved_env_var(name: &str) -> bool {
```
houko, line 107: `BLOCKED_ENV_EXACT` only has 9 names (vault key + `ANTHROPIC`/`OPENAI`/`GROQ`/`GEMINI` API keys + `GITHUB`/`NPM` tokens + AWS secret/session). `librefang-llm-drivers` actually reads `api_key_env` for roughly 40 providers — `OPENROUTER_API_KEY`, `MISTRAL_API_KEY`, `DEEPSEEK_API_KEY`, `COHERE_API_KEY`, `AZURE_OPENAI_API_KEY`, `XAI_API_KEY`, `TOGETHER_API_KEY`, and more (`crates/librefang-llm-drivers/src/drivers/mod.rs`).

Before this PR that gap didn't matter, because `is_blocked_env_var`'s substring heuristic (`KEY`/`TOKEN`/…) caught every one of those names regardless of whether they were in the exact list — `BLOCKED_ENV_EXACT` was genuinely "belt-and-suspenders" as the comment above `is_blocked_env_var` says.

Now that `EnvAllowlistSource::OperatorConfig` skips the heuristic and relies solely on `is_reserved_env_var`/`BLOCKED_ENV_EXACT`, this list has become the *sole* safety net for that path. An operator whose `default_model.provider` is e.g. `openrouter`, who allowlists `OPENROUTER_API_KEY` in their own `exec_policy.allowed_env_vars` for an unrelated reason (name collision, copy-paste from another tool's setup docs), would leak the daemon's live billing credential straight into a `shell_exec`/`process_start` child — unlike operators on Anthropic/OpenAI/Groq/Gemini/GitHub/NPM/AWS, who stay protected by the exact list.

Worth expanding `BLOCKED_ENV_EXACT` to the full provider-credential set (or deriving it from the driver registry so it can't drift) now that its completeness actually matters, rather than leaving it as the pre-#6458 belt-and-suspenders subset? Flagging as a question since it's a design/scope call (and `BLOCKED_ENV_EXACT` itself isn't touched by this diff) rather than a clear-cut bug.

---

crates/librefang-types/src/tool.rs
```diff
@@ -1380,6 +1388,19 @@ mod tests {
...
+        // strict default (#6458).
+        let legacy: DeferredToolExecution =
+            serde_json::from_str(&json.replace(r#""env_allowlist_source":"operator_config","#, ""))
```
houko, line 1398: Nit: this legacy-payload regression is coupled to serde_json's field-emission order via a raw string replace (`json.replace(r#""env_allowlist_source":"operator_config","#, "")`). It works today because `env_allowlist_source` happens to serialize right before `exec_policy` (which is `Some` here, so it isn't dropped by its own `skip_serializing_if`), but if a future PR reorders `DeferredToolExecution`'s fields, or `exec_policy` becomes `None` in this fixture, the trailing-comma assumption stops matching, the `replace` silently becomes a no-op, and the test would keep passing for the wrong reason (it'd just be re-parsing the field-present JSON, and `HandDeclared` wouldn't actually be exercised as "legacy/missing field").

Consider building the legacy JSON via `serde_json::Value` (parse, `.as_object_mut().remove("env_allowlist_source")`, re-serialize) instead of raw string surgery, so the test can't silently stop testing what it claims to test.

---

houko (commented):
Automated review pass against CLAUDE.md's rules (prose wrapping, security-sensitive env-allowlist logic). Verified the core provenance-splitting design carefully — traced every `EnvAllowlistSource::OperatorConfig` call site (the `execute_tool` shim's only production caller in `routes/tools_sessions.rs`, the agent-loop path keyed on `ctx.manifest.is_hand`, and the `/mcp` bridge's `caller_entry` check) and confirmed `manifest.is_hand` is set to `true` before `metadata["hand_allowed_env"]` is ever populated, so there's no path for a hand-declared (third-party) allowlist to be misclassified as operator-authored. Pushed a follow-up commit that reflows the new doc comments to the repo's one-sentence-per-line prose rule (no code/behavior changes). Left two comments below for judgment calls rather than direct edits.

---

houko:
Note on the automated review pass above: I found one mechanical, in-scope fix — several of this PR's new doc/inline comments (in `routes/network.rs`, `agent_loop/tool_call.rs`, `subprocess_sandbox.rs`, `tool_runner/dispatch.rs`, `tool_exec_backend.rs`, `kernel/mod.rs`, `config/types.rs`, `tool.rs`) wrap individual sentences across multiple lines instead of one-sentence-per-line, which the repo's prose convention calls for. Comment text itself is fine; only the line-wrapping needs to change.

I was not able to push that fix to this PR's actual branch: its head is `neo-wanderer/librefang:fix/allowed-env-vars-heuristic` (a fork), and my automation only has write access to `librefang/librefang`. I mistakenly pushed the reflow commit to a same-named branch on the base repo (`librefang/librefang:fix/allowed-env-vars-heuristic`) before realizing it wasn't the PR's head — that branch is disconnected from this PR and unrelated to it. I then tried to delete it to clean up but the deletion was also rejected, so it's still sitting there; a maintainer with delete rights on the base repo should remove `librefang/librefang:fix/allowed-env-vars-heuristic` when convenient. Apologies for the noise.

@neo-wanderer — if you'd like, enabling "Allow edits from maintainers" on this PR would let a maintainer push the reflow directly; otherwise it's a small enough change to fold in yourself (just line-break placement, no wording changes).

---

houko:
Closing as superseded by #6465, which is already merged to `main` and fixes the same issue (#6458).

#6465 trusts the operator's `exec_policy.allowed_env_vars` in `sandbox_command` via a two-list signature — `sandbox_command(cmd, operator_env_vars, untrusted_env_vars) -> Vec<String>` — and it already persists both the operator and the hand allowlists on `DeferredToolExecution`, so the approve→resume path filters correctly without a provenance tag (the regression test for that path landed in #6467).

Because both PRs re-signature the same `sandbox_command` function and update all of its call sites, they can't coexist; rebasing this branch onto current `main` would mean rewriting it onto the merged two-list design rather than resolving a conflict. The current CI red here is unrelated to your change — the Security lane is failing on the retired npm quick-audit endpoint (`pnpm audit` → 410), which affects every branch until CI is patched.

The scoping approach in this PR was correct — it just overlapped with parallel work on the same function. Thanks for the fix. If you'd like a follow-up, the one gap worth closing against the merged code is a `#[tokio::test]` covering the `/mcp` bridge's operator-vs-hand env provenance branch, which is currently only exercised by unit tests.

Issue #3866, opened 2026-04-28
WASM result_len + host_call params unbounded; serde_json deserializes guest output with no size/depth limit — guest can SIGSEGV the daemon


houko:
**Note (verified 2026-04-28):**
- **Accurate:** `result_len` in `sandbox.rs` is bounded only by guest memory (up to WASM linear memory size) with no explicit cap before calling `serde_json::from_slice`. Similarly `host_call` deserializes guest-provided JSON with no request-size cap. `host_kv_set` passes the entire `value` clone into `kernel.memory_store` without size limits. These are real concerns.
- **Inaccurate:** The issue cites `crates/librefang-runtime-wasm/src/host_functions.rs:507` as the location of the `host_kv_set` passthrough, but `host_functions.rs:507` is indeed where `host_kv_set` is defined — that part is correct. However, the issue also cites `sandbox.rs:325-344` for "guest result decode" and `sandbox.rs:425` for `host_call` request decode — these match the actual code. The concern about serde_json stack overflow from deeply nested JSON is real, but the stack overflow via 100k-deep nesting would require hitting serde_json's default recursion limit (128 levels on current versions); a ~128-level deep structure is far smaller than "100k-deep." The OOM risk from a large flat JSON string is more realistic.

---

## Location
- `crates/librefang-runtime-wasm/src/sandbox.rs` lines ~325–344 (guest result decode)
- `crates/librefang-runtime-wasm/src/sandbox.rs` line ~424–425 (`host_call` request decode)
- `crates/librefang-runtime-wasm/src/host_functions.rs` line ~507–529 (`host_kv_set` value passthrough)

## Problem

```rust
// sandbox.rs ~326–344
let result_ptr = (packed >> 32) as usize;
let result_len = (packed & 0xFFFF_FFFF) as usize;
// ... bounds check against memory.data().len() ...
let output: serde_json::Value = serde_json::from_slice(output_bytes)  // no size cap
    .map_err(|e| SandboxError::AbiError(...))?;
```

`result_len` has no explicit upper cap before the `serde_json::from_slice` call. The same pattern applies to `host_call` request decoding. `host_kv_set` clones and stores the full `value` in SQLite without any size limit.

## Impact
- A skill returning a very large JSON payload (within WASM memory limits) can cause significant allocation pressure on the daemon.
- `serde_json` has a recursion limit (~128 levels by default); a deeply nested structure near that limit could overflow the stack depending on platform.
[...]

Issue #3859, opened 2026-04-28
Dockerfile has no USER directive — fails CIS Docker Benchmark, blocks K8s runAsNonRoot


houko:
## Severity
**Medium** — fails CIS Docker Benchmark 4.1; cannot run on any platform that enforces `runAsNonRoot: true` (Kubernetes, Fly.io, hardened runtimes).

## Location
- `deploy/Dockerfile` — no `USER` instruction (`grep -c '^USER ' deploy/Dockerfile` = 0)
- `deploy/docker-entrypoint.sh:1` — root-only logic followed by `exec gosu node "$@"`
- ENTRYPOINT line: `deploy/Dockerfile:64`

## Finding

The Dockerfile leaves `USER` unset, so the container starts as root. The entrypoint does some chown/sed work as root, then drops privileges via `gosu node "$@"`. Functionally this works, but:

1. **CIS Docker Benchmark 4.1** (and most production hardening guides) require the final `USER` directive to be non-root. Auditors flag this on every scan.
2. **Kubernetes pod security standards `restricted`** profile enforces `runAsNonRoot: true`. Containers without a non-root `USER` are rejected at admission.
3. **Fly.io and similar PaaS** with rootless container runtimes refuse to schedule the image.
4. **Time spent as root before `gosu`** is non-trivial: chown -R on user data, sed -i to inject env values into config files. Any prompt-injection-style attack on `LIBREFANG_MODEL` env value (or any other operator-supplied env that hits sed) executes as root.

Issue #3556 already identified the entrypoint sed-injection; this is the orthogonal "even when sed is fixed, root-startup is the wrong default" angle.

## Why this matters now
The project's `deploy/docker-compose.yml` is the canonical self-host story, and `OBSERVABILITY.md` describes a multi-container stack. As soon as anyone tries to deploy that stack on managed K8s or a hardened cloud runtime, they hit a wall.

## Suggested fix

**Option A — split first-boot chown into an init container:**

```dockerfile
# deploy/Dockerfile
FROM ... AS final
COPY --chown=1000:1000 ... /app/
USER 1000:1000
ENTRYPOINT ["/app/librefang"]
```

For first-boot data setup, ship a `deploy/init-container.Dockerfile` that runs the chown/sed work as root in a separate container, sharing the volume:

```yaml
# docker-compose.yml
services:
  librefang-init:
    image: librefang/init:latest
    user: "0:0"
    volumes: [librefang-data:/data]
  librefang:
[...]

PR #6468, opened 2026-07-15
fix(channels): field-scope dm_policy/group_policy so a partial override stops silently gating groups, and expose them on [[sidecar_channels]] (#6445)

### LINKED ISSUE #6445: feat(channels): restore channel-level overrides for sidecars — partial [channel_overrides] silently flips group_policy to mention_only
## Summary

Post-sidecar-migration there is **no channel-level surface for behavioural `ChannelOverrides`** (`threading`, `group_policy`, `dm_policy`, `output_format`, …). The only place an operator can set them is the **agent-level** `[channel_overrides]` table in `agent.toml` (#3017) — which applies to every channel the agent serves, and, worse, **materializes the entire struct**: setting one field silently flips every other field from "absent" to its Rust default, changing runtime semantics that the operator never touched.

All references below are `main` @ `d243bcdec`.

## Current state of the three override surfaces

1. **Kernel / config.toml channel-level: gone.** `KernelApi::channel_overrides` is hardcoded to return `None` — `crates/librefang-api/src/channel_bridge.rs:1983`, with a comment noting per-channel overrides only ever lived on the retired in-process `[channels.<name>]` configs and "Sidecars have no equivalent per-instance override slot".
2. **`[[sidecar_channels]]` per-instance: partial.** `overrides_from_sidecar_config` (`crates/librefang-channels/src/sidecar.rs:1447`, #5841) projects only `command_policy` + `message_coalesce_window_ms`. Nothing else is reachable from the sidecar block.
3. **Env vars: WhatsApp-only.** Only the WhatsApp adapter reads `*_DM_POLICY` / `*_GROUP_POLICY` (`sdk/python/librefang/sidecar/adapters/whatsapp.py:520`, and even its `WHATSAPP_GROUP_POLICY` is documented inert at line 414). #6395 already cleaned up docs that claimed other adapters read these. There is no `SLACK_THREADING`-style knob; sidecar env is not an override surface.

So to enable, say, Slack thread replies (`threading = true`), the only working lever is the agent-level `[channel_overrides]` table.

## The footgun: partial overrides materialize full defaults

The bridge treats `overrides == None` as "no gating at all": `crates/librefang-channels/src/bridge.rs:3807-3808` runs the DM/group policy check only inside `if let Some(ref ov) = overrides`. But every surface that produces `Some` builds a **complete** `ChannelOverrides` with unstated fields at struct defaults:

- agent-level `[channel_overrides] threading = true` → `group_policy` becomes `MentionOnly` (`#[default]`, `crates/librefang-types/src/config/types.rs:41`) even though the operator never mentioned group policy;
- same in `overrides_from_sidecar_config`: configuring only a `command_policy = "blocklist"` on a sidecar constructs `ChannelOverrides::default()` around it (`sidecar.rs:1460`) — again flipping effective group handling from "process everything" to `mention_only`.

Concrete failure we hit: adding `threading = true` (for Slack) to an agent's `[channel_overrides]` silently killed **all Matrix inbound** for that agent, because the Matrix adapter marks every room `is_group=True` and never sets `was_mentioned` — see #6444. The gating flip itself, though, is channel-agnostic: any group traffic on any channel changes behaviour the moment an unrelated override field is set.

## Proposal

Two orthogonal pieces; either alone is already a big improvement:

1. **Field-wise override semantics.** Make unset fields behave identically whether the struct is absent or partially set — e.g. model overrides as `Option<_>` per behavioural field (or a merge step against a "neutral" baseline reflecting the `None` path), so `threading = true` cannot change group gating. At minimum this should hold for `group_policy` / `dm_policy`, whose defaults are the sharpest edges.
2. **Restore a channel-level override slot for sidecars.** Extend the `[[sidecar_channels]]` block (natural since #5841 already projects into `ChannelOverrides`) to accept the behavioural fields — `threading`, `group_policy`, `dm_policy`, `output_format`, … — scoped to that channel instance. Precedence stays as today: agent-level wins, channel-level is fallback (`bridge.rs:3783-3790`).

With (2), per-channel intent like "threads on Slack" stops requiring a per-agent, all-channel override; with (1), partial configuration stops rewriting unrelated policy.

houko:
Closes #6445 (author @neo-wanderer). Both proposals from the issue, in one PR.

## Problem

`ChannelOverrides.dm_policy` / `group_policy` were bare enums whose `#[default]` is `GroupPolicy::MentionOnly`. Every producer materializes the whole struct — the agent-level `[channel_overrides]` table via serde's container `#[serde(default)]`, and the sidecar projection. So writing a single unrelated field (e.g. `threading = true` for Slack) silently reflowed an unset `group_policy` to `MentionOnly`, flipping group traffic from the absent-`None` "process everything" behaviour to "mention-only" on every channel the agent serves.

The reporter set `threading = true` and lost all Matrix inbound. #6444 fixed the Matrix-adapter symptom; the materialization flip itself is channel-agnostic and was still live — `bridge.rs` gates group messages only inside `if let Some(ref ov) = overrides` (no `else`), so `None` means "no gating" while `Some(ChannelOverrides::default())` means "mention-only".

Separately, the sidecar migration left no channel-level surface for behavioural overrides at all — `overrides_from_sidecar_config` projected only command policy and coalescing.

## Fix — proposal 1 (sharp-field Option-ification)

`dm_policy` / `group_policy` become `Option<_>` with `#[serde(default, skip_serializing_if = "Option::is_none")]`, mirroring the sibling `output_format` field and the `CompactionOverrides::resolve` idiom. An unset policy stays `None` (no gating, identical to the historical whole-struct-absent path) and is distinct from an explicit `Some(MentionOnly)`.

- `should_process_group_message` keeps its `&ChannelOverrides` signature; the internal match gains `None | Some(GroupPolicy::All) => true`.
- Both gating paths (text + media), the reply-precheck `matches!`, and the DM-policy match handle `None` the same way.
- Hand-written `impl Default` sets both to `None`; the enums keep their own `Default`/`#[default]`.
- Cross-level `.or()` merge (agent table replaces channel table wholesale) is intentionally unchanged — existing semantics, out of scope. The other ~27 fields are untouched (the issue scopes proposal 1 as "at minimum `group_policy` / `dm_policy`").

## Fix — proposal 2 (sidecar channel-level override slot)

`SidecarChannelConfig` gains four `Option<_>` fields (`dm_policy`, `group_policy`, `threading`, `output_format`). `overrides_from_sidecar_config` widens its early-`None` guard so a behavioural-only config still yields an override, and projects ONLY the fields the operator explicitly set — an unset knob never materializes a policy they did not write (the same footgun, avoided by construction). Precedence unchanged: agent-level `[channel_overrides]` still wins.

`Option<_>` here (not the bare-enum-with-default of the existing `command_policy`) is deliberate: a sentinel default cannot distinguish "unset" from "explicitly set to the default value" — that IS the #6445 bug.

## Tests

- `none_group_policy_processes_all_but_some_mention_only_gates_6445` (bridge.rs) — the exact divergence: a struct with only `threading` set processes an unmentioned group message, while an explicit `Some(MentionOnly)` still gates it.
- `absent_group_and_dm_policy_deserialize_to_none_6445` (config/mod.rs) — a `[channel_overrides]` table with only `threading` leaves both policies `None`; `group_policy = "all"` round-trips to `Some(All)`.
- `overrides_projects_only_set_behavioural_fields_6445` / `overrides_threading_only_yields_override_6445` / `overrides_none_when_nothing_set_even_with_new_fields_6445` (sidecar.rs) — the projection.
- Updated every ripple literal/assert that relied on the old materialized default (`test_channel_overrides_defaults`, `test_channel_overrides_serde`, four `test_mention_only_*`, the `should_process_group_message_v2` module, the media gate test).

## Docs

Updated the configuration + integrations references (English + zh): the `[[sidecar_channels]]` field tables gain the four new fields, and the now-stale "not exposed on `[[sidecar_channels]]`" claims + the field-table defaults that listed the pre-fix materialized `respond` / `mention_only` are corrected.

## Verification

- `cargo test -p librefang-types --lib config::` — 158 passed
- `cargo test -p librefang-channels --lib` — 529 passed
- `cargo test -p librefang-channels --test bridge_integration_test` — 28 passed
- `cargo check --workspace --lib` — clean; `-p librefang-channels -p librefang-types -p librefang-api --no-run` — clean
- `cargo clippy -p librefang-types -p librefang-channels --lib -- -D warnings` — clean
- `cargo fmt --check` — clean

`SidecarChannelConfig` is only ever built by serde deserialization (no struct literals in the tree), so the added `Option` + `#[serde(default)]` fields are fully back-compatible.


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project uses [Calendar Versioning](https://calver.org/) (YYYY.M.DD).
 
 ## [Unreleased]
 
+### Added
+
+- Expose per-channel `dm_policy`, `group_policy`, `threading`, and `output_format` on `[[sidecar_channels]]`, restoring the channel-level override slot that was lost in the sidecar migration (#6445): each is `Option<_>`, so `overrides_from_sidecar_config` projects only the fields an operator explicitly set and an unset knob never materializes a policy they did not write. Precedence is unchanged — agent-level `[channel_overrides]` still wins over the per-channel value (#6468) (@houko)
```
houko, line 25: Commit `e7ffab9a41882182f9dd5602bf13733075e973ea` ("docs(channel-overrides): rewrap new prose to one sentence per line") has its author **and** committer identity set to `Claude <noreply@anthropic.com>`:

```
git log --format='%H | author=%an <%ae> | committer=%cn <%ce>' e7ffab9a41882182f9dd5602bf13733075e973ea
e7ffab9a... | author=Claude <noreply@anthropic.com> | committer=Claude <noreply@anthropic.com>
```

CLAUDE.md's Git Conventions section is explicit: "No AI / Claude attribution in commit messages, PR bodies, or comments — ... the rule exists because attribution pollutes `git log` and signals provenance the project does not want to imply." An AI-attributed author/committer on a commit is the same problem the `Co-Authored-By: Claude` trailer ban targets, just carried in the commit metadata instead of the message body — a `merge`/`rebase` merge of this PR will carry that identity permanently into `main`'s history.

This can't be mechanically fixed from a downstream review pass without rewriting history (the commit is 3 commits back from the current tip, so correcting the author would reshape `e7ffab9a`, `a2a1d3f8`, `4f3bf7c8`, and `401dc655` and require a force-push, which is out of bounds here). Recommend either squash-merging this PR (collapses individual commit authorship into the merge commit) or having the PR author interactively rebase to fix the author identity on `e7ffab9a` before merge.

---

houko (commented):
Automated daily review pass. Reviewed the full diff against origin/main for the CLAUDE.md conventions (deterministic-ordering / BTreeMap requirements, config field wiring, prose-wrapping, CHANGELOG attribution, integration-test mandate, AI-attribution ban). The channel-scoping fix itself (Option-ify dm_policy/group_policy, sidecar_channels projection, CHANGELOG/docs updates) is well-tested and correctly follows the repo's conventions — one finding below, left as a comment only since fixing it safely would require a history rewrite + force-push, both out of scope for this pass.

---

crates/librefang-channels/src/bridge.rs
```diff
@@ -3847,7 +3863,7 @@ async fn dispatch_message(
             // Reply-intent precheck: lightweight LLM classification for group
             // messages when group_policy is "all" and precheck is enabled.
             // Skipped for mentions and commands (already filtered above).
-            if ov.reply_precheck && matches!(ov.group_policy, GroupPolicy::All) {
+            if ov.reply_precheck && matches!(ov.group_policy, Some(GroupPolicy::All)) {
```
houko: `should_process_group_message` above (line ~2596) was deliberately changed so an unset `group_policy` (`None`) is treated identically to explicit `Some(GroupPolicy::All)` — its own comment says "byte-identical to the historical whole-struct-absent behaviour, same as an explicit `All`."

This `reply_precheck` gate wasn't updated to the same equivalence: it only fires for the literal `Some(GroupPolicy::All)`, not `None`. Concretely: an operator who sets `reply_precheck = true` but leaves `group_policy` unset — the common no-override config, which this PR just redefined to mean "process every group message" — gets *zero* reply-intent filtering; the bot replies unconditionally with no LLM-precheck safety valve. The same operator who instead writes the semantically-equivalent `group_policy = "all"` explicitly *does* get precheck. That's the opposite of "byte-identical," and it's a regression relative to the pre-fix state too (where unset materialized to `MentionOnly`, so this precheck branch was simply unreachable for the common case rather than silently skipped).

`crates/librefang-types/src/config/types.rs:139-141`'s doc comment on `reply_precheck` ("When true and group_policy is \"all\"") reflects the same pre-#6445 mental model and would need updating alongside a fix.

Note the fix isn't a pure `None` addition to the `matches!` — `group_trigger_patterns` set without an explicit `group_policy` resolves to an effective `MentionOnly` inside `should_process_group_message` (line ~2589), so `reply_precheck` should stay off in that sub-case. The condition needs something like `matches!(ov.group_policy, Some(GroupPolicy::All)) || (ov.group_policy.is_none() && ov.group_trigger_patterns.is_empty())`, ideally sourced from a shared helper rather than a third inline duplication of the resolution logic (the media-path twin at line 6382 has the identical gap).

---

crates/librefang-channels/src/bridge.rs
```diff
@@ -6362,7 +6379,7 @@ async fn media_dispatch_allowed(
             // classify an empty string: group_policy=all is opt-in to respond
             // broadly, and bare media always reached the agent before the media
             // gate existed, so proceeding preserves that behavior.
-            if ov.reply_precheck && matches!(ov.group_policy, GroupPolicy::All) {
+            if ov.reply_precheck && matches!(ov.group_policy, Some(GroupPolicy::All)) {
```
houko: Same asymmetry as the text-path check at line 3866: this only fires for the literal `Some(GroupPolicy::All)`, not the new `None`-means-process-all case that `should_process_group_message` now treats as equivalent. See the comment on line 3866 for the detailed scenario and a candidate fix — flagging here too since it's an independent duplicate of the same condition, not something a fix to the text path alone would cover.

---

CHANGELOG.md
houko, line 1: Not code-review, but worth a maintainer look before merge: commit `e7ffab9a41882182f9dd5602bf13733075e973ea` ("docs(channel-overrides): rewrap new prose to one sentence per line") in this PR's history is authored as `Claude <noreply@anthropic.com>`.

This repo's `CLAUDE.md` / commit-msg hook policy is "no Claude/Anthropic/AI attribution in commit messages, PR bodies, or comments" — that check scans message *content*, so it wouldn't flag an AI-looking *author* field, but it's the same category of provenance signal the policy exists to keep out of `git log`. I did not rewrite this since the branch is already under review and rewriting a merged-into-branch commit's authorship requires a rebase/force-push, which house rules reserve for pre-review branches only (and this PR has already had at least one round of code review per the later commit messages). Flagging for the maintainer to decide whether to `git commit --amend --author` / interactive-rebase this one commit before merge, or let it stand.

---

crates/librefang-kernel/src/kernel/handles/channel_sender.rs
```diff
@@ -82,15 +82,22 @@ impl kernel_handle::ChannelSender for LibreFangKernel {
...
+            .and_then(|ov| ov.output_format)
+            .unwrap_or_else(|| {
+                // wecom migrated to a sidecar; its formatting now happens inside
+                // the Python adapter (`librefang.sidecar.adapters.wecom`) which
+                // wraps every outbound chunk as `msgtype: "markdown"`. The
+                // generic `format_for_channel` path with the Markdown default
+                // (see `default_output_format_for_channel("wecom")`) gives the
+                // sidecar exactly that.
+                librefang_channels::formatter::default_output_format_for_channel(channel)
+            });
```
houko, line 99: Ambiguous — leaving as a comment rather than editing, since it's a testing-judgment call rather than a clear rule violation.

The commit message justifies skipping a dedicated test here ("the path has no existing harness and building a full mesh + mock adapter is disproportionate"). That's true for testing `send_channel_message` end-to-end, but the actual new logic — resolve `format` from `Option<ChannelOverrides>`, falling back to `default_output_format_for_channel(channel)` — doesn't touch the mesh or adapter at all once `channel_overrides()` has returned. It's a pure function of `(Option<ChannelOverrides>, &str) -> OutputFormat`.

Given `CLAUDE.md`'s MANDATORY integration-testing section is scoped to routes/wiring changes (this isn't a route), I'm not treating this as a hard requirement, but a maintainer might still want either: (a) extract the `adapter.channel_overrides()...unwrap_or_else(...)` resolution into a small named helper (e.g. `resolve_outbound_format(overrides: Option<&ChannelOverrides>, channel: &str) -> OutputFormat`) that's unit-testable without a mock adapter, or (b) accept the current inline form as-is since `sidecar.rs`'s `overrides_from_sidecar_config` projection tests already cover that `output_format` ends up `Some`/`None` correctly upstream of this call. Your call — not blocking.

---

crates/librefang-channels/src/bridge.rs
```diff
@@ -3847,7 +3856,7 @@ async fn dispatch_message(
             // Reply-intent precheck: lightweight LLM classification for group
             // messages when group_policy is "all" and precheck is enabled.
             // Skipped for mentions and commands (already filtered above).
-            if ov.reply_precheck && matches!(ov.group_policy, GroupPolicy::All) {
+            if ov.reply_precheck && matches!(ov.group_policy, Some(GroupPolicy::All)) {
```
houko: Possible second-order regression from the Option-ification — flagging rather than fixing since it's a behavioral judgment call, not a mechanical error.

`should_process_group_message` now treats `group_policy: None` (with no trigger patterns) as functionally equivalent to `Some(GroupPolicy::All)` — both let every group message through. But this reply-precheck gate (and its twin at line 6375 in `media_dispatch_allowed`) still checks `matches!(ov.group_policy, Some(GroupPolicy::All))` only, so a message that reached here because `group_policy` is unset (the new, corrected "process all" path) skips the reply-intent precheck classification entirely and always forwards to the LLM — whereas the same traffic under an *explicit* `group_policy = "all"` gets the precheck. Before this PR that distinction didn't exist in quite this shape (an unset policy previously materialized to `MentionOnly`, so it wouldn't have reached this branch as "all" traffic at all).

If the intent is "precheck applies whenever group messages are processed unconditionally" (which is what the existing comment above this line says: "when group_policy is \"all\""), this condition probably wants to become the same `effective_policy` resolution `should_process_group_message` uses (`None` without patterns, or explicit `Some(All)`) at both call sites. If the intent is narrower — precheck only for an *explicit* opt-in to "all" — then this is fine as-is and just needs a comment noting the deliberate asymmetry. Either way, worth a decision + a regression test (there isn't one covering precheck-firing under an unset `group_policy`) before merge.

---

houko (commented):
Automated daily review pass on #6468. Fixed a mechanical CLAUDE.md prose-wrapping violation (new comments/doc-comments reintroduced the ~80-column hard-wrap the PR's own earlier commit had already fixed once) in a follow-up commit, verified with scoped `cargo test -p librefang-channels --lib` (530 passed) and `-p librefang-types --lib` (875 passed) plus `rustfmt --check`. Left three inline comments on things that need a maintainer decision rather than a mechanical fix: a commit authored as `Claude <noreply@anthropic.com>` already in the branch history, a possible reply-precheck gating gap for the new `group_policy: None` "process all" path, and a testability suggestion for the new outbound `output_format` resolution. No approval/merge action taken.

---

houko:
Addressed the reply-precheck gating gap in b2e64fe03.

The three inline comments (bridge.rs text path #3859 / #3866, media path #6375) are the same defect: `should_process_group_message` resolves an unset `group_policy` (no trigger patterns) to process-all, but both reply-precheck gates matched only the literal `Some(GroupPolicy::All)`, so the common unset-policy config silently skipped the reply-intent filter while the semantically-identical explicit `group_policy = "all"` got it.

Fix, per the candidate in the review:

- Extracted a shared `group_reply_precheck_applies(&ChannelOverrides) -> bool` helper (bridge.rs, above `should_process_group_message`) so the text path and the media path resolve the state identically and cannot drift again — no third inline duplication. It returns true for explicit `all` and for `None` without trigger patterns, false under mention-only / commands-only / ignore and for `None` *with* patterns (which `should_process_group_message` resolves to mention-only, the patterns being the gate — so the precheck stays off there, matching the sub-case the review flagged).
- Both gates now call `ov.reply_precheck && group_reply_precheck_applies(ov)`.
- Updated the `reply_precheck` doc comment (`config/types.rs`) to describe both process-all triggers instead of only `group_policy = "all"`.
- Added `group_reply_precheck_applies_tests` (4 cases) locking the truth table, including the previously-uncovered "unset policy, no patterns → precheck applies" and "unset policy + patterns → precheck stays off" cases.

Verification (scoped, on the changed crates):

- `cargo test -p librefang-channels --lib` — 534 passed
- `cargo check -p librefang-channels -p librefang-types --lib` — clean
- `cargo clippy -p librefang-types -p librefang-channels --lib -- -D warnings` — clean
- `cargo fmt --check` — clean

The `output_format` testability comment on `channel_sender.rs` I'm leaving as-is (option b): it was flagged non-blocking, and the `overrides_from_sidecar_config` projection tests already cover that `output_format` lands `Some`/`None` correctly upstream of that call; extracting a kernel-side helper would cross into a different crate than this fix touches, so per the repo's scope rule I'm surfacing rather than bundling it.

Still open for a maintainer decision (me): the `Claude <noreply@anthropic.com>` author identity on `e7ffab9a` — and now `bdb04450` too, a second AI-authored rewrap commit is on the branch. Both need a history rewrite + force-push to correct, which the branch being under review argues against. I'll decide between squash-merge (collapses per-commit authorship) vs. an interactive rebase to fix both authors before merge separately.

houko:
Author-identity issue resolved. Interactively rebased the branch to reset the `Claude <noreply@anthropic.com>` author/committer on both `e7ffab9a` (now `17fc654c5`) and `bdb04450` (now `8087d6989`) back to my own identity; every commit on the branch is now authored by me. The rebase changed only author/committer metadata — the final tree is byte-identical to the pre-rebase tip (`git diff b2e64fe03 <new-head>` is empty). Force-pushed with `--force-with-lease`. `git log` no longer carries AI attribution, so no squash-merge is required on that account.

docs/src/app/configuration/channels/page.mdx
```diff
@@ -748,6 +749,14 @@ MY_API_TOKEN = "secret"
...
+| `group_policy` | string or null | `null` | Per-channel group policy (`all` / `mention_only` / `commands_only` / `ignore`), scoped to this sidecar instance (#6445). `null` inherits the bridge default (process all group messages). |
+| `threading` | bool or null | `null` | Enable thread replies for this sidecar instance (#6445). `null` inherits the bridge default (off). |
+| `output_format` | string or null | `null` | Output-format override (`markdown` / `telegram_html` / `slack_mrkdwn` / `plain_text`) for this sidecar instance (#6445). `null` inherits the channel default. |
+
+An unset (`null`) field is projected onto no override at all, so setting one knob here never materializes a policy you did not write — the footgun fixed in #6445.
+Agent-level `[channel_overrides]` in `agent.toml` still takes precedence over these per-channel values, and that precedence is whole-struct, not per-field: if the agent manifest declares a `[channel_overrides]` table at all, it replaces this sidecar's `dm_policy` / `group_policy` / `threading` / `output_format` entirely, even for the fields the agent table itself leaves unset.
```
houko, line 758: Good catch documenting this, but I want to flag the underlying behavior rather than just the doc.

`bridge.rs` resolves overrides with whole-struct `Option` precedence at both call sites (`dispatch_message` ~L3817-3821, `media_dispatch_allowed` ~L1228-1232):
```rust
let overrides = if let Some(aid) = early_agent_id {
    handle.agent_channel_overrides(aid).await.or(channel_overrides)
} else {
    channel_overrides
};
```
`agent_channel_overrides` returns `Some(_)` whenever `agent.toml` declares a `[channel_overrides]` table at all — regardless of which fields are set inside it. So the exact class of bug #6445 fixed within a single `ChannelOverrides` (an unrelated field silently materializing a policy the operator never wrote) still exists one layer up: an agent manifest that sets only `model` will silently discard a sidecar's `group_policy = "ignore"`, because the `.or()` treats "agent table present" as "use agent's values for every field," including the ones the agent left `None`.

Given `dm_policy`/`group_policy`/`threading`/`output_format` are now `Option<_>` on both structs, is a field-level merge (agent value if `Some`, else sidecar value, else default) the intended follow-up here, or is whole-struct agent precedence a deliberate simplicity/predictability choice? If deliberate, worth a one-line comment at both `.or()` call sites pointing at this doc paragraph so a future reader doesn't mistake it for an oversight. If not deliberate, this seems like the same bug class as #6445 and in the same crate/domain this PR is already touching.

---

houko (commented):
Reviewed the field-scoping fix, sidecar exposure, and docs (EN+zh). Left one inline comment on an architectural question (agent-vs-sidecar override precedence); pushed one small mechanical doc fix (prose-wrap) directly.

---

PR #6479, opened 2026-07-16
fix(dashboard/chat): distinguish context and budget limits

houko:
## Changes

- Add structured WebSocket error codes for context overflow, internal token budgets, provider rate limits, billing, authentication, model lookup, and other LLM failures.
- Make daemon-supplied semantic codes authoritative in the dashboard, while retaining text fallback for generic HTTP codes and legacy flat errors.
- Narrow legacy context matching so quota, 429, rate-limit, token-budget, tool-output-length, and other non-context errors no longer render as context exhaustion.
- Block automatic HTTP replay after a WebSocket error while retaining a 30-second listener window for a defensive late response with the same message id.
- Propagate error codes across late terminal-frame routing and clear stale error state when a response succeeds.
- Give quota errors actionable guidance for both agent/HAND resource limits and global config budgets.
- Update context-limit copy in all dashboard locales so it no longer describes provider quotas as session-length limits.
- Companion registry defaults change: https://github.com/librefang/librefang-registry/pull/102

## Verification

- `cargo test -p librefang-api --lib test_classify_streaming_error` — 5 passed on the final diff.
- `cargo test -p librefang-api` — all 825 library tests passed; the later integration phase received SIGABRT in `agents_routes_integration` without an assertion failure.
- `cargo test -p librefang-api --test agents_routes_integration -- --test-threads=1` — 51 passed, confirming the abort is concurrency/resource related.
- `pnpm test --run` — 85 files and 855 tests passed.
- `pnpm test --run src/pages/ChatPage.limit.test.ts src/lib/chat.test.ts` — 47 passed on the final diff.
- `pnpm typecheck`
- `pnpm lint` — 0 errors and 85 existing warnings.
- `pnpm exec vite build --outDir /tmp/librefang-dashboard-fix-build --emptyOutDir`
- `pnpm test:i18n-parity` reproduces the same eight pre-existing extra Ukrainian plural keys on clean main.
- `git diff --check`

## Out-of-scope follow-ups

- None.


crates/librefang-api/dashboard/src/pages/ChatPage.tsx
```diff
@@ -1726,7 +1720,15 @@ function CompactionSummaryBanner({ summary, isCompacting }: { summary: string |
   );
 }
 
-// No structured context-exhaustion signal exists on the chat surface; match the daemon / provider error string instead.
+const CONTEXT_LIMIT_CODES = new Set(["context_length_exceeded", "context_overflow"]);
```
houko: Checked the backend for producers of these code strings. `"context_overflow"` is never emitted anywhere in `crates/` — the WS classifier (`ws.rs`) uses `"context_length_exceeded"` for that category, and the `LlmErrorCategory::ContextOverflow` enum variant has no `rename_all` so it would serialize as `"ContextOverflow"` if it ever reached JSON, not `"context_overflow"`. `"quota_exceeded"` (line 1726) is similarly dead on this path: `ErrorCode::QuotaExceeded` is defined but never constructed, and the only live producer (`ExhaustionReason::QuotaExceeded::as_metric_label()`) feeds `GET /api/budget/providers`'s `exhaustion_reason` field, not the chat error `code`.

Meanwhile `POST /api/agents/:id/message` (the HTTP fallback path `sendViaHttp` uses, surfaced here via `ApiError.code`) can emit `agent_not_found`, `session_agent_mismatch`, and `message_delivery_failed` — none of which appear in either `CONTEXT_LIMIT_CODES` or `NON_CONTEXT_LIMIT_CODES`, so those fall through to the legacy string-matching fallback instead of being classified by code.

Is this intentional (i.e., are `context_overflow` / `quota_exceeded` reserved for a future/companion change, perhaps the linked librefang-registry#102), or should this list be trimmed to match what the backend actually emits today, with `session_agent_mismatch` / `message_delivery_failed` added for symmetry on the HTTP path?

---

crates/librefang-api/dashboard/src/pages/ChatPage.tsx
```diff
@@ -1115,25 +1119,15 @@ function useChatMessages(
...
               finishTurnIfCurrent(sendAgentId, botMsg.id);
-              // Don't cleanup the listener immediately — the agent may recover
-              // and send a final response. Shorten the inactivity window to
-              // 30s so the WS doesn't stay half-open forever if the failure
-              // is terminal.
-              if (turn.fallbackTimer) clearTimeout(turn.fallbackTimer);
-              turn.fallbackTimer = setTimeout(() => {
-                if (!turn.responded) { cleanup(); sendViaHttp(); }
-              }, 30_000);
+              cleanup();
```
houko: This removes the entire 30s HTTP-retry-after-WS-error fallback (previously: keep the WS listener alive and, if no late `response` arrived within 30s, re-send the turn over HTTP), not just an error-code classification tweak. The stated rationale — a server `error` frame is always terminal — is consistent with what I can see in `ws.rs` (every `"type": "error"` frame is sent from a branch that has already given up on the turn), but this is a behavior change with real user-facing consequences (no more automatic recovery if a WS error frame is ever sent for something that turns out to be transient), so flagging for a second pair of eyes rather than treating it as mechanical.

Is there a regression test (or was there manual verification) confirming no legitimate case exists where `ws.rs` sends an `error` frame and *later* still delivers a `response` for the same `message_id`? If such a case exists, this change would silently drop the recovery path instead of just fixing the misclassification the PR is about.

---

houko (commented):
Automated CLAUDE.md-compliance pass. Fixed one clear prose-wrapping violation directly (pushed as 9932742). Left two inline comments on judgment calls that go beyond the "distinguish context and budget limits" scope and deserve a second look rather than a unilateral fix.

---

crates/librefang-api/dashboard/src/pages/ChatPage.tsx
houko: Addressed in commits 2391c7d97 and 79430bf24. `CONTEXT_LIMIT_CODES` now contains only the live backend value `context_length_exceeded`; `context_overflow` was removed. Any daemon-supplied semantic code other than that explicit context code now returns false without string fallback, which covers `agent_not_found`, `session_agent_mismatch`, `message_delivery_failed`, budget, rate-limit, billing, and format errors. Only generic `HTTP_<status>` codes and legacy flat error strings retain text fallback for older daemons. The focused regression test now pins all three HTTP message codes plus the removed `context_overflow` value.

crates/librefang-api/dashboard/src/pages/ChatPage.tsx
houko: Addressed in commit 79430bf24 without restoring the duplicate HTTP request. On a WS `error`, the turn is now marked responded immediately so `onDrop` and the inactivity watchdog cannot replay it over HTTP, but the message listener remains attached for 30 seconds. A defensive late `response` for the same `message_id` therefore still reaches the existing response branch, clears `error` / `errorCode`, updates the bubble, and performs normal cleanup. `resetFallbackTimer` now exits once the turn is marked responded so unrelated socket frames cannot replace that 30-second cleanup timer. The final focused dashboard suite passes 47/47 and `pnpm typecheck` passes.

PR #6482, opened 2026-07-16
feat(sidecar-telegram): env opt-out (TELEGRAM_STREAMING) for the streaming path

### LINKED ISSUE #6469: Feature request: opt-out for Telegram sidecar streaming path (placeholder flash on fast responses)
## Symptom

On the Telegram sidecar, the streaming path produces a jarring visual flash for any answer that's ready before the 1s edit-debounce fires: a `…` placeholder is sent, then `StreamEnd` deletes it and sends a brand-new message with (often) identical text. For short/fast LLM responses there's no visible progressive edit at all — the user just sees the same message disappear and reappear.

There's currently no way to opt out of the streaming path for this adapter — `capabilities()` hardcodes `"streaming"` unconditionally.

## Root cause (read from current source, tag v2026.6.26-beta.24)

`sdk/rust/librefang-sidecar-telegram/src/adapter.rs:194-201`:

```rust
fn capabilities(&self) -> Vec<String> {
    vec![
        "typing".into(),
        "reaction".into(),
        "interactive".into(),
        "thread".into(),
        "streaming".into(),
    ]
}
```

No env var gates this — it's a fixed capability list. `crates/librefang-channels/src/bridge.rs:4795` checks `adapter.supports_streaming()` (which resolves to `has_cap("streaming")` per `crates/librefang-channels/src/sidecar.rs:1919-1920`) to decide whether to route through the streaming path or the plain `send_response` path. Since the capability is always present for this adapter, every send goes through streaming — there's no config knob to fall back to the simpler non-streaming flow.

Separately, `StreamEnd` finalizes via `finalize_as_new_message` (`adapter.rs` ~L142-186), which sends a fresh message and deletes the placeholder rather than editing it in place — this is what causes the visible flash even when the accumulated text hasn't changed since the last debounced edit.

## Why the observed behavior

- `StreamStart` → `sendMessage("…")`
- `StreamDelta` → `editMessageText` at most once per 1000ms (`STREAM_EDIT_INTERVAL_MS`)
- `StreamEnd` → `sendMessage(final_text)` + `deleteMessage(placeholder)`

For an LLM turn that completes in well under a second (common for short answers), there are zero `StreamDelta` edits — the user only sees `sendMessage("…")` immediately followed by `sendMessage(identical final text)` + delete. Net effect: the same message appears to vanish and reappear, with no intermediate content ever shown.

## Proposed fix

Add an opt-out env var (e.g. `TELEGRAM_STREAMING`, default `true` for backward compat) that gates the `"streaming"` entry in `capabilities()`:

```rust
fn capabilities(&self) -> Vec<String> {
    let mut caps = vec!["typing".into(), "reaction".into(), "interactive".into(), "thread".into()];
    if streaming_enabled() { // reads TELEGRAM_STREAMING, default true
        caps.push("streaming".into());
    }
    caps
}
```

With the capability absent, `bridge.rs`'s existing non-streaming path takes over automatically (it already sends `send_typing` unconditionally before branching on `supports_streaming()`, so the typing indicator is unaffected) — the bot would show a normal typing indicator and deliver one final message with a real push notification, no placeholder flash.

**Defensive (optional):** independently of the toggle, `StreamEnd` could finalize via `edit_with_fallback` (already implemented, used by `StreamDelta`) instead of `finalize_as_new_message`, removing the flash for streaming users too — at the cost of no push notification if the app was backgrounded before the placeholder arrived (Telegram doesn't push on message edits). This is a separate tradeoff from the toggle above and worth keeping as two independent options rather than conflating them.

## Setup (for reference)

- v2026.6.26-beta.24, Telegram sidecar (`librefang-sidecar-telegram`)
- Single bot, no `message_coalesce_window_ms`, streaming capability always on (unrelated to #6348/#6321 coalesce-path issues)
- No workaround currently available short of patching `adapter.rs` capabilities and rebuilding the sidecar

Happy to test a patch.

houko:
Closes #6469.

## Problem

`TelegramAdapter::capabilities()` unconditionally included `"streaming"`, so the daemon routed every send through the streaming path (`bridge` gates on `adapter.supports_streaming()` -> `has_cap("streaming")`).
For fast / short LLM turns this surfaces a visible `…` placeholder that flashes before the real answer, then gets replaced.

## Changes

- Add `streaming_enabled()` in `sdk/rust/librefang-sidecar-telegram/src/adapter.rs`: reads `TELEGRAM_STREAMING` live on each `capabilities()` call, defaulting to enabled. `"0"` / `"false"` / `"no"` / `"off"` (case-insensitive, trimmed) disable it; everything else — including an unset variable — leaves it enabled.
- `capabilities()` now builds the base vec `["typing","reaction","interactive","thread"]` and appends `"streaming"` only when `streaming_enabled()` is true. The base order is fixed and `"streaming"` is appended last, so the advertised capability list stays deterministic.
- Docs: add a `TELEGRAM_STREAMING` row (default `true`) to the sidecar env-var tables in `docs/src/app/architecture/rust-telegram-sidecar/page.mdx` and its zh translation.

`StreamEnd` / `finalize_as_new_message` behaviour is deliberately untouched — per the issue, the defensive edit-in-place change is a separate tradeoff and out of scope here.

## Tests

- `streaming_enabled_parsing_and_capability_gate` (unit test in `adapter.rs`): covers `streaming_enabled()` parsing for unset / `true` / `1` / `0` / `false` / `no` / `off` / ` OFF ` / `False` / `Off` / unknown, and asserts `capabilities()` drops `"streaming"` (while keeping the four base caps) when disabled and re-advertises it when enabled.
  Env-var mutation is process-global; this is the only test in the crate touching `TELEGRAM_STREAMING`, all cases run sequentially in one function, and an `EnvGuard` restores the original value on drop (even on panic).

## Verification

Cargo was not run locally: six sibling agents share this machine's `target/` and local compilation would thrash it (per the worktree workflow), so CI is the verification gate.
Reasoning checks performed by reading source: `BotClient::new` accepts any non-empty token (no network), `AllowList::from_env_value(Option<&str>)` matches the `None` call, the test module is a child of `adapter.rs` so it can construct `TelegramAdapter` with its private fields, and no other test asserts an exact telegram capabilities vector.

### DIFF
--- a/sdk/rust/librefang-sidecar-telegram/src/adapter.rs
+++ b/sdk/rust/librefang-sidecar-telegram/src/adapter.rs
@@ -42,9 +42,12 @@ macro_rules! tg_trace {
     ($($arg:tt)*) => { trace(format_args!($($arg)*)) };
 }
 
-/// Whether the streaming send path is advertised as a capability. Reads `TELEGRAM_STREAMING` live on each `capabilities()` call and defaults to enabled. `"0"` / `"false"` / `"no"` / `"off"` (case-insensitive, trimmed) disable it; everything else — including an unset variable — leaves it enabled.
+/// Whether the streaming send path is advertised as a capability.
+/// Reads `TELEGRAM_STREAMING` live on each `capabilities()` call and defaults to enabled.
+/// `"0"` / `"false"` / `"no"` / `"off"` (case-insensitive, trimmed) disable it; everything else — including an unset variable — leaves it enabled.
 ///
-/// When disabled the `"streaming"` capability is dropped, so the daemon routes every send through the plain (non-streaming) path. That trades incremental updates for a single final message, avoiding the visible placeholder flash on fast / short LLM turns.
+/// When disabled the `"streaming"` capability is dropped, so the daemon routes every send through the plain (non-streaming) path.
+/// That trades incremental updates for a single final message, avoiding the visible placeholder flash on fast / short LLM turns.
 fn streaming_enabled() -> bool {
     match std::env::var("TELEGRAM_STREAMING") {
         Ok(v) => !matches!(
@@ -211,8 +214,7 @@ impl SidecarAdapter for TelegramAdapter {
             "interactive".into(),
             "thread".into(),
         ];
-        // `"streaming"` is opt-out via `TELEGRAM_STREAMING`; appended last so the
-        // base capability order stays deterministic.
+        // `"streaming"` is opt-out via `TELEGRAM_STREAMING`; appended last so the base capability order stays deterministic.
         if streaming_enabled() {
             caps.push("streaming".into());
         }
@@ -469,8 +471,7 @@ impl SidecarAdapter for TelegramAdapter {
 mod tests {
     use super::*;
 
-    /// Restores `TELEGRAM_STREAMING` to its pre-test value on drop, so the
-    /// mutation stays contained even if an assertion panics mid-test.
+    /// Restores `TELEGRAM_STREAMING` to its pre-test value on drop, so the mutation stays contained even if an assertion panics mid-test.
     struct EnvGuard(Option<String>);
     impl Drop for EnvGuard {
         fn drop(&mut self) {
@@ -497,9 +498,8 @@ mod tests {
         }
     }
 
-    // `TELEGRAM_STREAMING` is process-global state; this is the only test in the
-    // crate that touches it and every case runs sequentially in one function, so
-    // nothing races with a parallel test. The guard restores the original value.
+    // `TELEGRAM_STREAMING` is process-global state; this is the only test in the crate that touches it and every case runs sequentially in one function, so nothing races with a parallel test.
+    // The guard restores the original value.
     #[test]
     fn streaming_enabled_parsing_and_capability_gate() {
         let _guard = EnvGuard(std::env::var("TELEGRAM_STREAMING").ok());
--- a/sdk/rust/librefang-sidecar-telegram/src/adapter.rs
+++ b/sdk/rust/librefang-sidecar-telegram/src/adapter.rs
@@ -42,6 +42,19 @@ macro_rules! tg_trace {
     ($($arg:tt)*) => { trace(format_args!($($arg)*)) };
 }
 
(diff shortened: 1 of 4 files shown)


sdk/rust/librefang-sidecar-telegram/src/adapter.rs
```diff
@@ -192,13 +208,17 @@ impl TelegramAdapter {
...
     fn capabilities(&self) -> Vec<String> {
-        vec![
+        let mut caps: Vec<String> = vec![
             "typing".into(),
             "reaction".into(),
             "interactive".into(),
             "thread".into(),
-            "streaming".into(),
-        ]
+        ];
```
houko, line 216: `TELEGRAM_CLEAR_DONE_REACTION` is both documented in the .mdx table and registered as a `Field` in `schema.rs` (the source the dashboard's configure form renders from, per the doc comment at the top of the .mdx page). `TELEGRAM_STREAMING` gets the doc-table row but no corresponding `Field::new("TELEGRAM_STREAMING", ...)` in `schema.rs`, so it won't show up in the dashboard's Telegram configure form even though the docs imply it should.

I didn't fix this myself: `Field` has no way to express "unset defaults to enabled" (no `default` on the struct), so a bare `FieldType::Bool` entry would render as an unchecked box while the real runtime default is *enabled* — that's actively misleading rather than a straightforward add. Worth a maintainer call: either add a `default` to `Field`/the dashboard form, word the label to make "unchecked = default (streaming on)" explicit, or intentionally leave this env var dashboard-invisible and trim the doc-table wording accordingly.

---

houko (commented):
Pushed a formatting-only commit (0e603c7) applying CLAUDE.md's one-sentence-per-line prose rule to the new doc comments in `adapter.rs` — no behavior change, `cargo test -p librefang-sidecar-telegram` (42/42) and `cargo clippy --all-targets -- -D warnings` still pass clean.

One more thing flagged inline below (dashboard schema gap), plus a repo-convention note: every other `[Unreleased]` entry in `CHANGELOG.md` carries a `(#issue) (@author)`-tagged bullet for behavior changes like this one, but this PR doesn't add one. Leaving that for the author since I can't attribute it correctly without knowing your GitHub handle.

---

Issue #6460, opened 2026-07-15
Per-user LLM provider credentials with per-owner usage attribution


houko:
Split out from #6447 (proposed by @sempervictus).

### Motivation

In team deployments, organizations want each human user's agents to consume that user's own LLM provider API key, so upstream usage tracking, quota, and chargeback attach to the human owner rather than to a shared daemon-wide credential.

### Current state

- LibreFang already has API-user identities with per-user keys, but those keys authenticate access to the LibreFang API itself (`crates/librefang-api/src/routes/users.rs`), not upstream LLM providers.
- LLM provider credentials are daemon-global (config / environment).
- Budget and metering are tracked per agent (`/api/budget/agents`), not per human owner.

### Proposed scope

- User-scoped provider credentials, stored encrypted via the existing vault mechanism (`LIBREFANG_VAULT_KEY`), never exposed back through the API in plaintext.
- Resolution order at driver-credential lookup: user-scoped key (of the human who owns the invocation) > daemon-global key.
- Inheritance: agents spawned or invoked on behalf of a user carry that user's provider identity for the duration of the invocation, including sub-invocations.
- Usage rollup by owner: extend metering so spend can be queried per user in addition to per agent.
- Integration tests covering resolution order and the inheritance boundary.

### Open questions

- How ownership is attributed for shared surfaces (channels, cron fires, agent-to-agent sends) where no single human initiated the turn.
- Interaction with existing `auth_profiles` key-rotation priorities.

### Out of scope

- The org-wide provider allowlist (tracked separately).
- Gateway-side enforcement upstream of the daemon.

sempervictus:
We use doubleword internally to track team members usage, handy for testing at it reports cost usage back to the caller.

houko:
Status update after scoping this out end-to-end.

**Metering-by-owner is already shipped** — this line item can be considered done independently of the credentials work. `UsageStore::query_user_hourly/daily/monthly` + `query_user_ranking`, the `usage_events.user_id` column, and the `/api/budget/users[/{id}]` routes already attribute spend to the human `UserId`, covered by `test_user_spend_rollup_per_window` and `test_user_ranking_excludes_anonymous_and_orders_by_daily`. So the genuinely-missing half was the **credential** side.

[...]

PR #6483, opened 2026-07-16
feat(kernel): per-user LLM provider credentials with per-owner usage attribution (initial)

houko:
## Summary

Refs #6460. The **credential storage + resolution-precedence core** of per-user LLM provider credentials. This is a complete, self-contained, reviewable slice: the storage and the owner-aware resolution are fully implemented and tested, and because every call site still passes `owner = None`, it changes no live behaviour. The remaining request-context / HTTP / dashboard work — and the design decisions it depends on — are enumerated under Deferred below, and tracked on the umbrella issue #6460.

In team deployments an org wants each human user's agents to consume that user's *own* upstream provider API key, so upstream usage / quota / chargeback attach to the human owner rather than to a shared daemon-wide credential. This PR lands the storage + resolution-precedence core and its tests; it is intentionally a minimal, coherent, compiling slice, with the larger request-context / HTTP / dashboard work explicitly deferred and listed below.

While scoping this I found the **per-owner metering rollup already exists** in the codebase (`UsageStore::query_user_hourly/daily/monthly` + `query_user_ranking`, the `usage_events.user_id` column, and the `/api/budget/users[/{id}]` routes, with tests `test_user_spend_rollup_per_window` and `test_user_ranking_excludes_anonymous_and_orders_by_daily`). So the genuinely missing half was the **credential** side and its resolution; that is what this PR adds.

## Substantive changes

- New `crates/librefang-kernel/src/user_provider_credentials.rs`:
  - `USER_PROVIDER_KEY_PREFIX` + `user_provider_vault_key(user_id, provider)` — a per-user, per-provider vault-key namespace (`user_provider_key/<user-uuid>/<provider>`), distinct from the `mcp_oauth/…` and `__sentinel__` namespaces in the shared `vault.enc`.
  - `resolve_provider_credential(user_scoped, global)` — the resolution precedence: user-scoped key wins, else fall back to the daemon-global credential (so global-only behaviour is unchanged when no user key exists).
  - `LibreFangKernel::{set,get,remove,list}_user_provider_key` — thin wrappers over the **existing** vault APIs (`vault_set` / `vault_get` / `vault_handle`), so credentials are stored **encrypted via the existing AES-256-GCM vault** (keyed by `LIBREFANG_VAULT_KEY` / OS keyring) — no new crypto. `list_user_provider_keys` returns provider **names only**, sorted (deterministic); the plaintext value never crosses an API/list boundary.
- `crates/librefang-kernel/src/kernel/llm_drivers.rs`: `resolve_driver` now delegates to a new owner-aware `resolve_driver_for_owner(manifest, owner: Option<UserId>)`. When `owner` is `Some` and that user has a stored key for the resolved provider, the user-scoped key wins over **both** the credential pool and the env-var global key (and bypasses the pool). When `owner` is `None` (every current call site) or the user has no key for the provider, resolution is **byte-identical** to the previous env-var / pool behaviour. An agent-pinned `api_key_env` (operator override) is not shadowed by a user key.
- `CHANGELOG.md`: `### Added` entry under `[Unreleased]`.

## Tests added

- `user_provider_credentials::tests::vault_key_is_namespaced_stable_and_isolated`
- `user_provider_credentials::tests::resolution_prefers_user_scoped_over_global`
- `user_provider_credentials::tests::provider_names_filter_by_user_sort_and_exclude_other_namespaces`
- `kernel::tests::per_user_provider_key_beats_global_and_falls_back` — a `#[tokio::test]` in the serialised `librefang_vault_key` group that boots a real kernel, stores a user key through the encrypted vault, and asserts: (a) **resolution order** — the user-scoped key beats the daemon-global key, and a provider with no user key falls back to global (global-only behaviour unchanged); (b) storage round-trips verbatim and is isolated per user / per provider; (c) `list_user_provider_keys` exposes names only, never the secret value; (d) removal drops only the targeted pair.

The **metering-by-owner query** is covered by the pre-existing `librefang-memory` tests noted above (`test_user_spend_rollup_per_window`, `test_user_ranking_excludes_anonymous_and_orders_by_daily`); this PR does not duplicate them.

## Deferred / Open questions (not hand-waved — concrete items)

1. **Request-context owner threading.** `resolve_driver_for_owner` exists and applies the precedence, but every current call site (`agent_execution.rs`, `messaging.rs` ×3) still passes `None`. Wiring the turn's human owner (available as `AuthenticatedApiUser.user_id` at the API boundary, and derivable via `UserId::from_name`) down through `send_message_full` / `execute_llm_agent` into `resolve_driver_for_owner` is the follow-up that makes this end-to-end. Kept out of this slice to avoid an invasive, hard-to-review call-graph change while the storage/resolution core is reviewed.
2. **HTTP + dashboard management surface.** There is no route yet to set / delete / list a user's provider keys (kept out because adding routes here would require regenerating `openapi.json` and would trip the `dead_route_audit` / `openapi_path_coverage` reflection tests, which needs a build this environment can't run). Proposed shape for the follow-up, mirroring `POST /api/users/{name}/rotate-key`: `PUT /api/users/{name}/provider-keys/{provider}` (body `{api_key}`), `DELETE …/{provider}`, `GET …/provider-keys` (names only), Owner-gated like the other `/api/users*` mutations.
3. **Ownership attribution for shared surfaces with no single human initiator** (channels, cron fires, agent-to-agent sends). Channel messages derive `SessionId::for_channel(...)` and cron fires synthesize a `"cron"` channel context — neither carries a human `UserId` today, so a per-user credential cannot be selected for them. Open question: fall back to daemon-global (current behaviour), attribute to a configured channel/cron "owner" user, or refuse. This PR leaves them on the global path.
4. **Full inheritance through nested sub-invocations.** Even once (1) lands, propagating the owner identity through forks / sub-agent spawns / `agent_send` for the duration of an invocation is a separate correctness concern (forks force `Persistent` sessions and reuse prompt cache; leaking an owner across a fork boundary could mis-attribute spend). Deferred deliberately rather than landed half-done.
5. **Interaction with `auth_profiles` key-rotation priorities.** `resolve_non_default_api_key_env` already layers operator-explicit `[provider_api_keys]` / `[auth_profiles]` over catalog / convention env-var names. A user-scoped key currently short-circuits *above* that env-var resolution entirely. Whether a user-scoped key should instead participate in (or be subordinate to) an operator's rotation-priority chain for the default provider is an open design question flagged for the maintainer.

## Verification

Rebased onto current `origin/main` (the only conflict was the `[Unreleased]` CHANGELOG adjacency, resolved keeping both `### Added` blocks). Verified via the repo's sanctioned dev image (`Dockerfile.rust-dev`, Linux, isolated named-volume `target/` — no contention with the host):

- `cargo check -p librefang-kernel --lib` — clean.
- `cargo test -p librefang-kernel --lib user_provider` — 4 passed (`vault_key_is_namespaced_stable_and_isolated`, `resolution_prefers_user_scoped_over_global`, `provider_names_filter_by_user_sort_and_exclude_other_namespaces`, `per_user_provider_key_beats_global_and_falls_back`).
- `cargo test -p librefang-kernel --lib resolve_driver_for_owner` — 1 passed (`resolve_driver_for_owner_drives_the_owner_scoped_key_path`).

All 4 `resolve_driver` production call sites remain unaffected — the public signature is preserved and each still delegates with `owner = None`. CI remains the gate for the full cross-platform matrix.


crates/librefang-kernel/src/kernel/llm_drivers.rs
```diff
@@ -205,15 +231,27 @@ impl LibreFangKernel {
...
+        // Per-user provider credential (#6460): resolve the human owner's own
+        // key for this provider, if any. A user-scoped key takes precedence
+        // over BOTH the credential pool and the env-var global key (and
+        // bypasses the pool below). Skipped when the agent pinned an explicit
+        // `api_key_env` — that operator-level override wins.
+        let user_scoped_key: Option<String> = if has_custom_key {
+            None
+        } else {
+            owner.and_then(|uid| self.get_user_provider_key(uid, agent_provider))
+        };
```
houko, line 242: Verified the two headline claims in the PR body by reading the full file: the `owner = None` path is byte-identical to pre-PR behaviour (`user_scoped_key` is unconditionally `None` when `owner.and_then(...)` has nothing to chain from, so both the `pool_opt` gate and the `api_key` `if let` fall straight through to the untouched historical branches), and `list_user_provider_keys` really does sort deterministically at the boundary (`provider_names_from_keys` sorts + dedups a `Vec` collected from `CredentialVault::list_keys()`'s `HashMap`-backed `entries`, so the nondeterministic iteration order never escapes). Vault reuse is also as claimed — `set/get_user_provider_key` route through the existing cached `vault_set`/`vault_get`/`vault_handle` (accessors.rs), no new crypto.

One gap: the doc comment on `resolve_driver_for_owner` (and this line's comment) asserts "An agent that pins an explicit `api_key_env` in its manifest is treated as an operator-level override and is NOT shadowed by a user-scoped key" — that's exactly what `has_custom_key` guards here — but neither new test (`per_user_provider_key_beats_global_and_falls_back`, `resolve_driver_for_owner_drives_the_owner_scoped_key_path`) sets `manifest.model.api_key_env` while an owner-scoped key also exists for that provider. That branch is reachable on the very first real call site that gets wired up (any agent with a pinned `api_key_env` whose owner also has a stored key), and it's the one place this PR's precedence claim could silently invert. Worth a third test case: owner has a stored key for the provider, manifest also sets `api_key_env` to an env var with a known value, assert the resolved driver used the env var's key, not the vault one.

---

houko (commented):
Automated daily review pass. No prior review comments existed on this thread. Verified the three specific claims flagged for scrutiny (deterministic sort in list_user_provider_keys, vault-API reuse with no new crypto, byte-identical owner=None resolution) by reading the full llm_drivers.rs and user_provider_credentials.rs against the diff — all three hold up. One gap noted inline: missing test coverage for the "operator-pinned api_key_env not shadowed by a user-scoped key" branch, which is a real, reachable code path with a doc-comment claim but zero assertions backing it. Everything else in the author's own Deferred/Open-questions list is legitimately out of scope for this slice and not re-raised here.

---

crates/librefang-kernel/src/user_provider_credentials.rs
houko, line 1: Daily CLAUDE.md-compliance pass on this WIP slice.

The four substantive items already listed in the PR's "Deferred / Open questions" section (request-context owner threading, HTTP/dashboard surface, ownerless-surface attribution, sub-invocation inheritance, auth_profiles rotation interaction) are correctly scoped and disclosed with reasoning — not flagging those per CLAUDE.md's conflict-resolution guidance (surface disagreement, don't punt unilaterally; these already surface the question to the maintainer).

Verified clean:
- `provider_names_from_keys` (this file, line ~40-52) uses a sorted+deduped `Vec<String>`, not a raw `HashMap` iteration, so the PR body's "sorted (deterministic)" claim for `list_user_provider_keys` holds — matches the repo's #3298 determinism rule.
- No plaintext credential is logged or returned through any listing path; `get_user_provider_key` is `pub(crate)`-scoped (commit `cb0aaaca8`) so the only plaintext getter can't leak out of the kernel crate.
- No `KernelConfig` field was added, so the config-wiring gotcha (struct field + `#[serde(default)]` + `Default` impl) doesn't apply here.
- `rustfmt --check --edition 2021` on both touched source files passes, matching the PR's "Verification" claim.
- CHANGELOG bullet is one long compound sentence (semicolons/em-dash, no embedded periods) — consistent with the prose-wrapping rule and with the sibling bullet already in the file.

One undisclosed finding — commit authorship, not code:

`git log` shows commit `729b20fab` (HEAD, "style(kernel): reflow #6460 doc comments to one-sentence-per-line") has **author identity `Claude <noreply@anthropic.com>`** (`git log --format='%an <%ae>'`). The other 4 commits on this branch are correctly attributed to `Evan <tonymo2048@gmail.com>`. CLAUDE.md's rule against AI attribution in "commit messages, PR bodies, or comments" exists because "attribution pollutes git log ... and signals provenance the project does not want to imply" — an author-identity field showing `Claude <noreply@anthropic.com>` in `git log`/`git blame` is exactly that pollution, and it slipped through because the `commit-msg` hook scans message *content*, not the author field, so no trailer string like `Co-Authored-By: Claude` was needed to trigger it.

I'm not fixing this myself: correcting it requires rewriting this commit's metadata (`commit --amend --author=...` or a rebase) followed by a force-push, and both are barred for me here (the harness's "always create new commits, never amend" rule, and this review's explicit "never force-push" instruction) — force-pushing over a commit whose SHA is already referenced in the open PR also isn't a decision I should make unilaterally. Recommend the maintainer (or the PR author, same-repo so it's their own branch) reword the commit's author metadata before this leaves draft, e.g. `git rebase -i --exec 'git commit --amend --no-edit --author="Evan <tonymo2048@gmail.com>"' HEAD~5` then force-push, or squash on merge if the maintainer prefers to just let the merge commit carry correct attribution.

---

crates/librefang-kernel/src/kernel/llm_drivers.rs
```diff
@@ -105,6 +105,23 @@ impl LibreFangKernel {
     pub(crate) fn resolve_driver(
         &self,
         manifest: &AgentManifest,
+    ) -> KernelResult<Arc<dyn LlmDriver>> {
+        // Owner-agnostic entry point — every current call site.
+        // Delegates with no owner, so credential resolution is byte-identical to the historical env-var / credential-pool behaviour.
+        // Owner-aware dispatch (per-user provider credentials, #6460) goes through `resolve_driver_for_owner`; wiring the human owner of a turn in from request context is a follow-up (see the PR's Deferred section).
+        self.resolve_driver_for_owner(manifest, None)
```
houko, line 112: Re-checking deferred item 1 ("Request-context owner threading") against the CLAUDE.md test — "would fixing it require touching a different crate or domain than the one you're already in? If no, fix it in this PR" — I think this deferral bundles two things that don't have the same answer to that test.

The PR body names the call chain as `AuthenticatedApiUser.user_id` (API boundary) → `send_message_full` → `execute_llm_agent` → `resolve_driver_for_owner`. `send_message_full` lives in `crates/librefang-kernel/src/kernel/messaging.rs:876` and `execute_llm_agent` lives in `crates/librefang-kernel/src/kernel/agent_execution.rs:333` — both in `librefang-kernel`, the exact crate this PR already touches (this file). Only the very first hop, reading `AuthenticatedApiUser.user_id` at the HTTP handler, lives in a different crate (`librefang-api`, `middleware.rs`).

So the intra-crate plumbing — adding an `owner: Option<UserId>` parameter to `send_message_full` / `execute_llm_agent` and threading it down to this `resolve_driver_for_owner` call, with every existing call site passing `None` (byte-identical behavior, same pattern already used here for `resolve_driver`'s own wrapper) — passes the "same crate" test and looks like it should have landed in this slice. The genuinely cross-crate piece (extracting the user id from `AuthenticatedApiUser` in `librefang-api` and passing it into the kernel entry point) is the legitimate part of the deferral.

The PR's stated reason for deferring the whole thing is reviewability ("invasive, hard-to-review call-graph change"), not a crate-boundary reason — worth the maintainer's explicit call on whether that's enough to defer the in-crate half too, per the CLAUDE.md scoping rule.

---

houko (commented):
Reviewed the diff against CLAUDE.md's engineering/collaboration rules (vault namespacing, HashMap/BTreeMap determinism, config Default-impl requirements, CHANGELOG attribution/prose-wrap, and the "fix what you found vs. legitimate deferral" test). The core storage + resolution slice (user_provider_credentials.rs, the resolve_driver_for_owner precedence, the vault-namespace tests) is clean: the vault-key namespace is collision-free and tested against the mcp_oauth/__sentinel__ namespaces, the plaintext getter is pub(crate)-scoped so it can't leak past the internal driver-resolution path, listing returns sorted provider names only (never values), DriverConfig already redacts api_key in its custom Debug impl so nothing new leaks to logs, no new HashMap lands in a prompt-affecting path, no new config/manifest struct fields needed a Default entry, and the CHANGELOG bullet is present, correctly attributed, and follows the one-sentence-per-line prose rule.

One inline comment on the "request-context owner threading" deferral (item 1 in the PR body) — the call chain it names spans two crates but the deferred work doesn't have to: `send_message_full` / `execute_llm_agent` are both in `librefang-kernel`, the same crate already being touched here, so the intra-crate parameter-threading (defaulting every current call site to `None`, same pattern as the `resolve_driver` wrapper added in this PR) looks like it passes CLAUDE.md's same-crate bar. Only the initial hop (reading `AuthenticatedApiUser.user_id` in `librefang-api`) is genuinely cross-crate. Flagging for the maintainer's call rather than assuming it's a non-issue.

The other four deferred items (HTTP/dashboard surface, channel/cron attribution, nested sub-invocation inheritance, auth_profiles interaction) all hold up as legitimate: they either cross into librefang-api/dashboard, are blocked on an environment constraint (openapi regen), or are explicitly surfaced as open design questions for the maintainer rather than silently-punted code work.

---

crates/librefang-kernel/src/kernel/llm_drivers.rs
```diff
@@ -105,6 +105,23 @@ impl LibreFangKernel {
...
+    }
+
+    /// Resolve the LLM driver for an agent turn on behalf of an optional human owner.
+    ///
+    /// Credential precedence (#6460): when `owner` is `Some` and that user has stored their own provider key for the resolved provider (see [`crate::user_provider_credentials`]), that user-scoped key wins over the daemon-global credential — both the credential pool and the env-var key.
```
houko, line 117: Nice, self-contained slice — storage goes through the existing vault correctly, `list_user_provider_keys` sorts before returning (verified `CredentialVault::entries` is a `HashMap`, so this sort is load-bearing per #3298, not decorative), and `get_user_provider_key` is correctly kept `pub(crate)` so the plaintext can't leak through a public API even by accident.

One gap not in the Deferred list: this precedence is only wired into the *primary* driver slot. The fallback-chain loop a bit further down in this file (`for fb in &effective_fallbacks`, around the `fb_api_key` resolution) never sees `owner` or `user_scoped_key` at all — each fallback slot always resolves its key via `fb.api_key_env` or `resolve_non_default_api_key_env`, independent of anything decided for the primary slot. So even after deferred item (1) threads a real owner in, an agent with `fallback_models` configured would apply the user-scoped key to the primary provider only; any fallback slot for a provider the user *also* has a stored key for would silently fall back to the daemon-global credential instead.

This may be intentional (fallback slots already don't inherit the primary's `api_key_env` override either, so it's consistent with existing behavior), but since the CHANGELOG/docstring here talk about "the resolved provider" without calling out the fallback chain explicitly, it seems worth a one-line note (here or in the #6460 tracking issue) so whoever lands item (1) doesn't assume this precedence already covers the whole driver chain.

---

houko (commented):
Reviewed the credential-storage + resolution-precedence core. Vault usage, determinism (sorted provider-name listing), and the pub(crate) plaintext-getter boundary all check out — one non-blocking note left inline about the fallback-chain not sharing this precedence.

---

crates/librefang-kernel/src/kernel/llm_drivers.rs
```diff
@@ -205,15 +222,24 @@ impl LibreFangKernel {
                 .unwrap_or_else(|| DriverConfig::default().max_retries),
         };
 
+        // Per-user provider credential (#6460): resolve the human owner's own key for this provider, if any.
+        // A user-scoped key takes precedence over BOTH the credential pool and the env-var global key (and bypasses the pool below).
+        // Skipped when the agent pinned an explicit `api_key_env` — that operator-level override wins.
+        let user_scoped_key: Option<String> = if has_custom_key {
```
houko, line 238: `user_scoped_key` is only consulted for the primary driver slot (here and the `pool_opt` bypass a few lines down). The fallback-chain builder further down this function (the `for fb in &effective_fallbacks` loop, `fb_api_key` resolution) always resolves fallback-slot credentials via env var / catalog only — it never looks up the owner's own key for `fb_provider`. So once request-context owner threading (Deferred item 1 in the PR body) lands, a turn that fails over from the primary provider to a fallback slot would silently drop back to the daemon-global credential for that slot, even though the primary slot honored the user-scoped key. This isn't one of the five deferred items called out in the PR body (those cover owner threading, the HTTP/dashboard surface, channel/cron attribution, nested sub-invocation inheritance, and auth_profiles interaction — not the fallback chain specifically), so flagging it separately: worth a conscious decision (and probably a code comment / follow-up issue) on whether fallback slots should also resolve a per-owner key for their own provider, or whether falling back to global-only is the intended behavior for fallback slots.

---

Issue #6459, opened 2026-07-15
Org-wide LLM provider allowlist (governance: constrain which providers agents may use)


houko:
Split out from #6447 (proposed by @sempervictus).

### Motivation

Workplace deployments need to constrain which LLM providers agents are allowed to call, so that prompts and proprietary data cannot accidentally flow to a vendor the organization has not approved.
Today the operator configures providers centrally, but nothing prevents an agent manifest or a runtime fallback chain from selecting any provider the daemon has credentials or endpoints for.

### Proposed scope

- A kernel-level allowlist (e.g. `[providers] allowed = [...]` in `config.toml`) restricting which provider families may be used by any agent, including aux-task and cheap-tier fallback chains.
- Fail-closed enforcement at driver resolution time, with a clear error surfaced to the agent and a `WARN` log for the operator.
- Classification in `build_reload_plan` (`crates/librefang-kernel/src/config_reload.rs`) so operators know whether the field hot-reloads, plus a row in `docs/operations/config-reload.md`.
- Dashboard visibility: show the active allowlist and flag configured-but-disallowed providers.
- Integration tests at the enforcement boundary per the repo testing policy.

### Out of scope

- Per-user provider credentials and usage attribution (tracked separately).
- Per-agent provider restrictions (the existing manifest already selects providers; this issue is the org-wide ceiling).

hegu-1:
Fail-closed at driver resolution is the right enforcement point. I would make the allowlist apply to the **resolved route**, not only the provider family string.

The decision input should include provider, endpoint/region, model, credential owner, data residency class, and whether the route came from primary, auxiliary, or fallback configuration. That prevents an allowed alias from resolving to an unapproved endpoint, and prevents a cheap-tier fallback from bypassing the organization policy.

Two CI fixtures are worth adding:

1. primary provider is allowed but its fallback is not — resolution must fail rather than silently downgrade;
2. provider name is allowed but endpoint/region is outside policy — fail closed before any prompt is sent.

The audit receipt should record configured route, resolved route, policy version, and decision. This model-routing template has explicit route/fallback/policy fields that may help shape the contract: https://github.com/hegu-1/enterprise-ai-os-architecture/blob/main/layers/04-brain/templates/model-routing.json

houko:
Implemented and shipped in #6484 (commit 235274b1). The org-wide provider allowlist landed as a `[providers] allowed = [...]` surface on `KernelConfig`, enforced fail-closed at driver resolution:

[...]

PR #6484, opened 2026-07-16
feat(kernel): org-wide LLM provider allowlist (fail-closed at driver resolution)

houko:
## Summary

WIP / initial slice of the org-wide LLM provider allowlist (governance).
Adds a `[providers] allowed = [...]` config surface and enforces it fail-closed at LLM driver resolution.
Backward compatible: an empty or absent allowlist means "no restriction — all providers allowed", so existing deployments are unaffected.

Refs #6459 (draft, partial — see Deferred below).

## What this does

- **Config**: new `ProvidersConfig` (`[providers] allowed: Vec<String>`) added to `KernelConfig` — struct field + `#[serde(default)]` + Default impl entry + Serialize/Deserialize/JsonSchema. Default is empty (`allow-all`).
- **Enforcement (fail-closed at driver resolution)**:
  - `LibreFangKernel::resolve_driver` (the per-turn primary path) rejects a disallowed resolved provider with `LibreFangError::CapabilityDenied` (a clear agent-facing error) and emits an operator `WARN`, before any driver is constructed — so the CLI-profile-rotation, credential-pool, single-key, and boot-default-fallback branches are all gated.
  - Disallowed providers in the agent/global fallback chain are skipped with a `WARN` (mirroring the existing init-failure skip), never added to the chain.
  - `AuxClient::build_driver` (aux-task / cheap-tier chains) skips a disallowed provider (`Err` → caller tries the next slot, exactly like a missing credential) with a `WARN`, so the side-task path cannot route around the gate.
- **Matching** is normalized: case-insensitive and `-`/`_` treated as equivalent (so `"Claude-Code"` in the list matches a resolved `claude_code`). The list names provider *identifiers* — the same strings used in `[default_model] provider`, `[[fallback_providers]] provider`, and `[llm.auxiliary]` specs.
- **Reload classification**: `providers` wired into `build_reload_plan` as **NOOP** (`classified_reload_fields()` + `docs/operations/config-reload.md` row, both drift-guarded by tests). Rationale: `resolve_driver` reads the allowlist live from `self.config.load()` every turn and the aux client is rebuilt on reload, so an allowlist edit takes effect on the next turn after the ArcSwap config swap — no restart, no explicit hot action.

## Tests added

- `crates/librefang-types` — `ProvidersConfig` decision function:
  - `provider_allowlist_empty_allows_everything`
  - `provider_allowlist_non_empty_is_fail_closed`
  - `provider_allowlist_matching_is_normalized`
  - `provider_allowlist_absent_section_is_backward_compatible`
  - `provider_allowlist_rejection_reason_is_actionable`
- `crates/librefang-kernel` — enforcement boundary:
  - `resolve_driver_enforces_provider_allowlist` (empty allowlist resolves; non-empty rejects a disallowed provider fail-closed; permits a listed local provider). Also exercises the live-read hot-reload path by swapping the allowlist on the ArcSwap config between assertions.

## Deferred (not attempted in this PR)

- **Dashboard visibility** of the active allowlist and flagging of configured-but-disallowed providers (e.g. a `[default_model]` / `[[fallback_providers]]` / `[llm.auxiliary]` provider that is not in `allowed`). This is a UI/read surface and is intentionally out of scope for this initial enforcement slice; it belongs in the dashboard domain (`crates/librefang-api/dashboard/`) and warrants its own PR.
- **Boot-time default-driver construction** is not gated (boot builds `default_driver` regardless). This is defense-in-depth only: every agent turn is already gated at `resolve_driver`, so a disallowed default provider cannot serve a real turn. Documented here for the follow-up.

## Verification

Per the parallel-worktree workflow, `cargo build` / `cargo check` / `cargo clippy` / `cargo test` were **not** run locally (six sibling agents share `target/`; local compilation thrashes the machine). `cargo fmt --check` was run (formatting only, no compilation) and passes. CI is the compile/test gate.


houko:
Correction to this PR's description (found in review): the two safety claims in the original body — "a disallowed default provider cannot serve a real turn" and "the side-task path cannot route around the gate" — were **both incorrect**.

The boot-time `default_driver` fallback chain (`boot.rs`) was built with no allowlist gate. That ungated driver seeds `aux.primary`, which the aux inherit path hands verbatim to side tasks (compaction, cron compaction, history fold) and to the two primary-turn shortcuts (CLI-profile early return; init-failure fallback). On the normal anthropic-exhausted condition the `FallbackDriver` could fail over to a disallowed provider, sending conversation content to a vendor the allowlist forbids.

Fixed in `c276973eb`: the boot `default_driver` fallback loop now applies the same fail-closed skip + WARN as the per-slot gate in `resolve_driver`, so no ungated driver is constructed or handed out. `4e8e10058` adds a source-grep regression test pinning both kernel-side gate sites and exposes the `[providers]` field in the dashboard config overlay.

Still deferred on this draft: the dashboard visibility of the active allowlist (flagging configured-but-disallowed providers), and a runtime-crate regression test for the aux/cheap-tier gate.

PR #6496, opened 2026-07-19
fix(ci): pin login_page.html to LF so the CSP-hash test passes on Windows (#6481)

houko:
## Problem

`main` has been red on the **Windows shard only** since #6486 last touched the login page. The failing test is:

```
FAIL middleware::tests::dashboard_login_page_script_is_allowed_by_csp_hash
thread panicked at crates/librefang-api/src/middleware.rs:3759
```

(Confirmed from run 29559606139 `--log-failed`; macOS and Ubuntu shards are green.)

### Root cause

`login_page.html` is embedded verbatim via `include_str!` (`middleware.rs:1730`), and its inline `<script>` is authorised by an exact SHA-256 hardcoded in the CSP (`middleware.rs:1731`, `sha256-TDA4xC…`). The file had **no `.gitattributes` rule**, so a Windows runner checkout with `core.autocrlf=true` materialises it with CRLF line endings. The `\r` bytes flow into the embedded string, the runtime `Sha256::digest(script)` no longer matches the hardcoded hash, and the assertion at `middleware.rs:3759` fails. The failure is invisible on macOS/Linux because they check the file out as LF.

Every open PR inherited this failure on its next CI run — the bot filed #6481 and re-attributed it across #6482 / #6485 / #6486 as they merged.

## Fix

One `.gitattributes` rule pinning the file to LF, mirroring the existing golden-fixtures rule right above it:

```
crates/librefang-api/src/login_page.html text eol=lf
```

The blob is already LF in the index (`git ls-files --eol` → `i/lf`), so `--renormalize` produced no content change — this only stops Windows checkouts from drifting to CRLF.

## Scope note

`login_page.html` is the **only** `include_str!`-embedded, byte-compared HTML in the tree (verified: the sole such reference is `middleware.rs:1730`), so no sibling files need the same treatment.

## Verification

- `git check-attr text eol -- crates/librefang-api/src/login_page.html` → `text: set`, `eol: lf`.
- No Rust code changed, so no behavioural test applies; the fix is a checkout-normalisation rule whose effect only manifests on a Windows runner. CI's Windows shard on this PR is the real verification — if it goes green, `main` is unblocked.

Refs #6481

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,9 +5,7 @@
 # explicit `eol=lf` here is a safety rail for tests that byte-compare.)
 crates/librefang-api/tests/fixtures/*.json text eol=lf
 
-# login_page.html is embedded verbatim via include_str! and its inline
-# <script> is pinned by an exact SHA-256 in the CSP (middleware.rs). A
-# Windows checkout with core.autocrlf=true rewrites it to CRLF, which
-# shifts the hash and fails dashboard_login_page_script_is_allowed_by_csp_hash
-# on Windows only. Pin it to LF so the embedded bytes match CI on every OS.
+# login_page.html is embedded verbatim via include_str! and its inline <script> is pinned by an exact SHA-256 in the CSP (middleware.rs).
+# A Windows checkout with core.autocrlf=true rewrites it to CRLF, which shifts the hash and fails dashboard_login_page_script_is_allowed_by_csp_hash on Windows only.
+# Pin it to LF so the embedded bytes match CI on every OS.
 crates/librefang-api/src/login_page.html text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,10 @@
 # tests would fail cross-platform. (Rust source + JSON are fine as-is; the
 # explicit `eol=lf` here is a safety rail for tests that byte-compare.)
 crates/librefang-api/tests/fixtures/*.json text eol=lf
+
+# login_page.html is embedded verbatim via include_str! and its inline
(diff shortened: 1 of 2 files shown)


.gitattributes
houko, line 1: The `.gitattributes` fix itself is correct and durable: `crates/librefang-api/src/login_page.html` is already stored LF-only in the blob (verified against `main`, zero `\r` bytes), so the Windows failure was purely a `core.autocrlf=true` working-tree rewrite on checkout, and pinning `eol=lf` here (mirroring the existing fixtures rule) fully closes that without needing a renormalize/re-add of the file itself. The new comment block also follows the one-sentence-per-line prose rule correctly.

One thing worth a look before merge: commit `6200715` ("style: rewrap .gitattributes comment to one sentence per line"), which is currently the branch HEAD, has both author and committer set to `Claude <noreply@anthropic.com>` (GitHub resolves it to the `claude` bot identity). CLAUDE.md's "No Claude / Anthropic / AI attribution" rule is written in terms of commit message trailers, but this bakes the same provenance signal into the commit's author/committer identity instead of the message body, so it slips past the `commit-msg` hook's string match while landing the same thing the hook exists to prevent. Since it's a metadata issue rather than content, fixing it means rewriting this commit (reauthor + force-push), which I'm not doing from an automated review pass — flagging so a maintainer can squash/reauthor before merge if the project wants to keep AI attribution out of `git log`.

---

houko (commented):
Daily automated review of PR #6496. No mechanical fixes applied — one finding left inline.

---

Issue #6225, opened 2026-06-19
bug(api): compaction-summary banner leaks across sessions — agent-scoped canonical summary shown on a freshly created session


neo-wanderer:
## Summary

The dashboard "Session summary (older messages compacted)" banner appears on a **freshly created session that was never compacted**, showing the summary of a *previous, unrelated* conversation. The compaction summary is persisted per-**agent**, but the banner is gated only on "is this the agent's active session?" — never on "was *this* session compacted?".

## Root cause

`compacted_summary` lives in the `canonical_sessions` table, keyed by `agent_id` (one row per agent):

```sql
-- crates/librefang-memory/src/migration.rs:498 (migration v5)
CREATE TABLE IF NOT EXISTS canonical_sessions (
    agent_id TEXT PRIMARY KEY,
    messages BLOB NOT NULL,
    compaction_cursor INTEGER NOT NULL DEFAULT 0,
    compacted_summary TEXT,
    updated_at TEXT NOT NULL
);
```

Compaction writes it via `store_llm_summary` with `ON CONFLICT(agent_id) DO UPDATE`
(`crates/librefang-memory/src/session.rs:1291`, persist at `:1757`). So the summary is **agent-scoped and outlives any individual session**.

The session GET handler `get_agent_session` exposes it whenever the requested session is the agent's *active* session, with no check that this session's own history was ever compacted:

```rust
// crates/librefang-api/src/routes/agents/sessions.rs:268
let compacted_summary: Option<String> = if target_session_id == entry.session_id {
    state.kernel.memory_substrate()
        .canonical_context(agent_id, None, Some(0))   // per-AGENT row
        .ok()
        .and_then(|(summary, _)| summary)
} else {
    None
};
```

`create_agent_session` (`sessions.rs:599` → `kernel.create_agent_session`) makes the new session the active one (updates `entry.session_id`) but does **not** clear `canonical_sessions.compacted_summary`. The new session therefore inherits the prior summary, and the dashboard renders `CompactionSummaryBanner` on message #1.

## Reproduce (synthetic)

1. Agent `A`. Hold a conversation in session `S1` long enough to trigger history compaction (or force `/compact`). → `canonical_sessions[A].compacted_summary` is set from `S1`.
2. Create a new session `S2` for `A` (`POST /api/agents/A/sessions`). `S2` becomes active; `entry.session_id == S2`.
[...]

Issue #6494, opened 2026-07-18
Knowledge graph (entities/relations) has no per-user scoping — only agent_id, unlike memories.peer_id


nevgenov:
## Summary

The knowledge graph (`entities` / `relations` tables) has no per-user scoping dimension at all — only `agent_id`. This is inconsistent with the `memories` table, which has a `peer_id` column and enforces it in the recall filter (`AND peer_id = ?`). On any agent shared by multiple distinct real users (differentiated per-message via `peer_id`, e.g. several Telegram users routed to one bot/agent), KG facts written by one user are immediately visible to every other user of that same agent — not as a leak-after-compaction like #6493, but structurally, on every read, since there is no user dimension to filter by in the first place.

## Where

Schema (`crates/librefang-memory` migrations):
```sql
CREATE TABLE entities (
    id TEXT PRIMARY KEY,
    entity_type TEXT NOT NULL,
    name TEXT NOT NULL,
    properties TEXT NOT NULL DEFAULT '{}',
    created_at TEXT NOT NULL,
    updated_at TEXT NOT NULL,
    agent_id TEXT NOT NULL DEFAULT ''
);
CREATE TABLE relations (
    id TEXT PRIMARY KEY,
    source_entity TEXT NOT NULL,
    relation_type TEXT NOT NULL,
    target_entity TEXT NOT NULL,
    properties TEXT NOT NULL DEFAULT '{}',
    confidence REAL NOT NULL DEFAULT 1.0,
    created_at TEXT NOT NULL,
    agent_id TEXT NOT NULL DEFAULT ''
);
```
No `peer_id` column on either table.

API (`crates/librefang-memory/src/knowledge.rs`):
```rust
pub fn add_entity(&self, entity: Entity, agent_id: &str) -> LibreFangResult<String>
pub fn add_relation(&self, relation: Relation, agent_id: &str) -> LibreFangResult<String>
pub fn query_graph_scoped(&self, pattern: GraphPattern, agent_id: Option<&str>) -> LibreFangResult<Vec<GraphMatch>>
```
Every write/read path takes `agent_id` only — there is no `peer_id`/`user_id` parameter anywhere in the KG API, so there is nothing for a caller to even pass through if it wanted per-user isolation.

## Impact

[...]

PR #6497, opened 2026-07-19
fix(memory): gate compacted summary by owning session to stop cross-user prompt leak (#6493)

### LINKED ISSUE #6493: canonical_sessions.compacted_summary leaks across users on multi-user agents (unconditional injection, no session-ownership check)
## Summary

`canonical_context()` filters the **recent message window** by `session_id` (per the fix that added `CanonicalEntry.session_id` tagging), but the **`compacted_summary`** produced by compaction is stored as a single value per `agent_id` and is returned/injected into the prompt **unconditionally**, regardless of which session's messages actually produced it. On any agent shared by multiple distinct real users (differentiated per-message via `peer_id`, not via separate agents), the compacted summary from user A's conversation leaks into user B's system prompt.

## Where

- `crates/librefang-memory/src/session.rs` — `canonical_context()`:
  ```rust
  pub fn canonical_context(
      &self,
      agent_id: AgentId,
      session_id: Option<SessionId>,
      window_size: Option<usize>,
  ) -> LibreFangResult<(Option<String>, Vec<Message>)> {
      let canonical = self.load_canonical(agent_id)?;
      // ... recent messages ARE filtered by session_id ...
      Ok((canonical.compacted_summary.clone(), recent))
  }
  ```
  The `compacted_summary` is returned as-is — no check against `compacted_summary_session_id`.

- `crates/librefang-kernel/src/kernel/agent_execution.rs` (~line 807) consumes it unconditionally:
  ```rust
  canonical_context: if stable_prefix_mode {
      None
  } else {
      self.memory
          .substrate
          .canonical_context(agent_id, Some(effective_session_id), None)
          .ok()
          .and_then(|(s, _)| s)
  },
  ```

- `CanonicalSession.compacted_summary_session_id` exists precisely to identify which session "owns" a given summary — per its doc comment it's currently only used to gate a dashboard UI banner, **not** used to gate prompt injection here.

## Reproduction

1. Configure a single agent/bot channel with multiple distinct `peer_id`s (e.g. several Telegram users allow-listed to the same bot/agent, each resolved to a different external identity via a per-message `user_id`/`peer_id` parameter passed to tools).
2. User A has a long-running conversation that crosses the agent's `compaction.threshold_messages`, producing a `compacted_summary` that describes User A's specific conversation content — tagged with `compacted_summary_session_id = <User A's session>`.
3. User B — a different real person sharing no context with User A — sends any message to the same agent/bot.
4. Building the prompt for User B's turn calls `canonical_context(agent_id, Some(effective_session_id_for_B), None)`. The recent-message window correctly excludes User A's messages, but the returned `compacted_summary` is User A's summary, and it gets injected into User B's system prompt unconditionally.

## Verified in production

On a live multi-user agent (3 distinct real users mapped via per-message `peer_id` to different external mailbox/CRM identities through an MCP server-side user-mapping config), `canonical_sessions.compacted_summary` currently contains a detailed summary of one specific user's document-editing session, with `compacted_summary_session_id` pointing at that user's own session — while the agent's `sessions` table rows all have an empty `peer_id` column (a possibly-related/separate issue in populating per-user session scoping for this channel type). Any of the other users messaging the bot next would receive the first user's summary injected into their turn's system prompt.

## Suggested fix

Gate the summary the same way `canonical_context()` already gates the recent-message window:

- Only include `compacted_summary` in the returned tuple when `compacted_summary_session_id` matches the requested `session_id`, OR is `None` (legacy/no-owner rows — mirrors the existing "untagged entries are chat-agnostic" rule already applied to recent messages).
- If cross-session/cross-channel continuity for the *same* real user is still desired (canonical sessions are documented as "cross-channel memory for one user"), compaction should probably be bucketed per distinct `peer_id`/user identity rather than a single agent-wide blob, since a single agent can legitimately be shared by several distinct real users differentiated only by `peer_id` at the tool-call layer (not by separate `AgentId`s).

## Related

- Adjacent to the fix that added `session_id` filtering for the recent-message window in canonical context (guarded by `canonical_chat_scoped_integration.rs`) — that fix covered recent messages but not the summary field, which is the one place cross-user content still leaks.

houko:
Fixes #6493.

## The leak

`SessionStore::canonical_context` (`crates/librefang-memory/src/session.rs`) is the reader on the **prompt-assembly path** — the kernel calls it at `messaging.rs:2579` and `agent_execution.rs:812`, both passing `Some(effective_session_id)`, and injects the returned summary at index 0 of the turn's prompt (`build_canonical_context_message`).

The `compacted_summary` is **agent-scoped**: `canonical_sessions` has one row per `agent_id` (no `peer_id`/user column). `canonical_context` filtered the *message list* by session but returned `compacted_summary` **unconditionally**, ignoring the `compacted_summary_session_id` owner column that #6225 added.

So on a multi-user agent (several `[[users]]` sharing one agent, each on its own `SessionId`), a summary built from **user A's** conversation was injected verbatim into **user B's** turn — up to a 4000-char excerpt of A's private messages, at prompt index 0.

#6225 added the owner column but wired the check into only the dashboard `/session` banner (`compacted_summary_for_session`, `substrate.rs:711`). The prompt-injection reader was left ungated. This PR closes that.

**Exposed by default:** the leak path runs whenever `stable_prefix_mode` is false — the default.

## The fix

`canonical_context` now applies the same ownership check `compacted_summary_for_session` already used:

```rust
let summary = match (&session_id, &canonical.compacted_summary_session_id) {
    (Some(want), Some(owner)) if owner != want => None,
    _ => canonical.compacted_summary.clone(),
};
```

- A caller that names session B gets `None` when the summary is owned by A.
- The owning session still gets its summary.
- A **legacy ownerless** row (pre-#6225) and an **unscoped `None`** read (agent-wide cross-channel read) keep the prior always-return behaviour, so existing deployments and the `None`-passing call sites are not regressed.

No signature change → the two kernel call sites are untouched.

## Tests

Two regression tests in `session.rs`:

- `test_canonical_context_summary_gated_by_owning_session` — summary owned by session A is withheld from B, still returned to A and to an unscoped `None` read.
- `test_canonical_context_legacy_ownerless_summary_still_returned` — an ownerless legacy summary is still returned (nothing to gate against).

All 10 `canonical` tests pass; clippy clean.

**Test-level note:** the regression is at the memory-crate boundary (`canonical_context`), which is the exact defective symbol. The two kernel call sites only forward its return value, so gating at the source gates the whole prompt path. A full kernel end-to-end test (drive a real user-B turn, assert the assembled prompt omits A's summary) would add a layer of confidence and can follow if reviewers want it.

## Out of scope (flagged, not fixed here)

The heuristic summary builder in `append_canonical` concatenates messages across sessions *before* stamping `compacted_summary_session_id = session_id` (`session.rs:1676-1712`). That means an owner's own summary can fold in earlier fragments from other sessions of the same agent. This is **owner-visible-only** — the read-path gate in this PR ensures it never reaches a *different* user — so it is not a cross-user leak, and fixing it would change compaction semantics (per-session summaries instead of a rolling agent-wide one). Left for a separate decision rather than bundled here.

## Verification

- `cargo check -p librefang-memory --lib` — clean (via `Dockerfile.rust-dev`, Linux).
- `cargo test -p librefang-memory --lib canonical` — 10 passed, 0 failed.
- `cargo clippy -p librefang-memory --lib -- -D warnings` — clean.


CHANGELOG.md
```diff
@@ -32,6 +32,10 @@ and this project uses [Calendar Versioning](https://calver.org/) (YYYY.M.DD).
...
+  The #6225 fix had scoped only the dashboard `/session` banner (`compacted_summary_for_session`) and left this injection reader ungated; `canonical_context` now applies the same owner check — a summary is surfaced only to the session that owns it, while a legacy ownerless row or an unscoped (`None`) agent-wide read is unchanged.
+  Exposed by default (the leak path runs whenever `stable_prefix_mode` is false, the default).
+  A follow-up observation, out of scope here: the heuristic summary builder in `append_canonical` concatenates across sessions before stamping the owner, so an owner's own summary can still fold in earlier same-agent session fragments — that is owner-visible-only (not cross-user) and would change compaction semantics to fix, so it is left for a separate decision.
```
houko: Pushing back on the "owner-visible-only (not cross-user)" characterization here, per CLAUDE.md's "fix what you found" rule (the bar to defer is "would fixing it require touching a different crate or domain?" — this is the same file/crate you're already in, `librefang-memory::session.rs`, so it shouldn't be deferred without explicitly asking).

`append_canonical`'s compacting slice (`session.rs:1676`, `let compacting = &canonical.messages[canonical.compaction_cursor..to_compact];`) is taken straight from `canonical.messages` with no filter on `entry.session_id`, and `session.rs:1712` then stamps whatever got folded in as owned by whichever session's turn happened to trigger the compaction threshold.

Per the existing `canonical_context_isolates_two_whatsapp_chats_for_same_agent` test (`tests/canonical_chat_scoped_integration.rs`), `canonical.messages` genuinely interleaves entries from different `SessionId`s belonging to different real people (Jessica's DM vs. a group chat) — which is exactly this PR's own leak scenario ("several `[[users]]` shared one agent, each on its own `SessionId`"). So a compaction triggered by session B's turn can fold in session A's message text, then get stamped `compacted_summary_session_id = Some(B)`. The read-side gate this PR adds correctly lets B see "its own" summary — but that summary can still contain A's raw conversation text, laundered in at write time. That's a cross-user leak through a different code path, not merely "owner's own summary."

Could you either:
1. Filter the compacting slice by `entry.session_id == session_id` (or otherwise scope the summary to the triggering session) in this same PR, since it's the identical file/crate, or
2. If per-session summaries is a real behavior change (rolling agent-wide → per-session) that needs a maintainer's explicit sign-off on compaction semantics, say so explicitly here and get that OK, rather than asserting unilaterally that this is "not cross-user."

Happy to be wrong if there's a session-scoping step upstream of `compacting` that I'm missing, but the interleaving test cited above contradicts the claim as currently written.

---

crates/librefang-memory/src/session.rs
```diff
@@ -1748,7 +1748,14 @@ impl SessionStore {
...
         let recent = filtered[start..].to_vec();
-        Ok((canonical.compacted_summary.clone(), recent))
+        // Gate the compacted summary by its owning session, mirroring MemorySubstrate::compacted_summary_for_session (#6225).
+        // The summary is agent-scoped (one row per agent, keyed by agent_id) but is built from one session's history, so returning it unconditionally on a caller that named a *different* session injects that session's content into another's prompt — a cross-user leak on a multi-user agent (#6493).
+        // When a specific session is requested, surface the summary only if it is the one that owns it; a legacy row with no recorded owner, or a caller that requested no particular session (None, e.g. a full agent-wide read), keeps the prior behaviour.
+        let summary = match (&session_id, &canonical.compacted_summary_session_id) {
+            (Some(want), Some(owner)) if owner != want => None,
+            _ => canonical.compacted_summary.clone(),
+        };
+        Ok((summary, recent))
```
houko, line 1790: Re: the PR description's note that "a full kernel end-to-end test... can follow if reviewers want it" — a kernel-level test lives in a different crate (`librefang-kernel` / `librefang-api` integration tests per the repo's `#[tokio::test]`-against-`TestServer` convention), so per CLAUDE.md's "different crate or domain" carve-out this one is fine to leave as a question rather than fix inline — and the PR already surfaces it as an explicit ask rather than silently punting, so no action needed here. Flagging only so it doesn't get lost: worth an explicit reviewer yes/no rather than merging silently on the assumption that no one wants it.

---

houko (commented):
Daily CLAUDE.md-compliance pass on this PR. One mechanical fix pushed (CHANGELOG prose-wrap); one substantive finding left as inline review comments below for maintainer/author input rather than silently fixed or silently accepted.

---

crates/librefang-memory/src/session.rs
```diff
@@ -1748,7 +1748,14 @@ impl SessionStore {
             .collect();
         let start = filtered.len().saturating_sub(window);
         let recent = filtered[start..].to_vec();
-        Ok((canonical.compacted_summary.clone(), recent))
+        // Gate the compacted summary by its owning session, mirroring MemorySubstrate::compacted_summary_for_session (#6225).
+        // The summary is agent-scoped (one row per agent, keyed by agent_id) but is built from one session's history, so returning it unconditionally on a caller that named a *different* session injects that session's content into another's prompt — a cross-user leak on a multi-user agent (#6493).
```
houko, line 1784: Re: the "out of scope" deferral in the PR body / CHANGELOG ("that is owner-visible-only, not cross-user") — I think this undersells the risk and want a second look before it's treated as a cosmetic follow-up.

`canonical.messages` is agent-scoped and cross-*channel* by design (doc comment at the `CanonicalSession` struct: "All channels contribute to it"). Each `CanonicalEntry` carries its own `session_id` independently, so on a multi-user agent the flat list interleaves entries from every user's session in whatever order their turns happened to append.

`append_canonical` (same file, not touched by this diff) is called unconditionally on every turn (`messaging.rs:3043`, `agent_execution.rs`) with `session_id = effective_session_id` — i.e. whichever session's *current* turn happens to push `canonical.messages.len()` over `DEFAULT_COMPACTION_THRESHOLD` (100). The `compacting` slice (`append_canonical`, around line 1676: `&canonical.messages[cursor..to_compact]`) is *not* filtered by `session_id` — it can and will contain `CanonicalEntry`s from other users' sessions if their turns interleaved chronologically with the triggering session's. The heuristic builder concatenates all of their text into `full_summary` regardless of origin, then stamps `compacted_summary_session_id = session_id` (around line 1712) — the *triggering* session, not the set of sessions whose content is actually embedded in the summary.

Concretely: agent shared by session A (user A) and session B (user B), messages interleave A1,B1,A2,B2,…. B's Nth message happens to cross the 100-message mark. The compacted slice covers a mix of A's and B's entries; the resulting summary (containing A's message text) gets stamped `owner = B`. After this PR's read-gate here, `canonical_context(agent, Some(B))` matches `owner == want` and returns that summary to B — verbatim, including A's content. That's the same cross-user leak class as #6493, just introduced at the write/stamping side instead of the read side this PR fixes, and it's reachable through ordinary multi-user usage (no `/compact` or explicit action needed) — not a rare edge case.

This is distinct from the LLM-based `compact_session` / `store_llm_summary` path (`agent_runtime.rs:437`), which operates on one session's own `session.messages` and is correctly scoped — this concern is specific to the heuristic canonical-compaction fallback in `append_canonical`.

Given the threat model this PR itself is fixing (multi-user agents sharing one agent via `[[users]]`), I'd push back on filing this purely as a "compaction semantics" nicety — worth tracking it as a proper security follow-up issue (not just a CHANGELOG aside), or gating the heuristic compactor to refuse mixing entries whose `session_id`s differ until it's fixed. Happy to be wrong if there's a mitigating constraint I'm missing (e.g. something upstream guarantees `canonical.messages` never actually interleaves across users in deployed configs) — but nothing in this file enforces that.

---

CHANGELOG.md
```diff
@@ -32,6 +32,10 @@ and this project uses [Calendar Versioning](https://calver.org/) (YYYY.M.DD).
 
 ### Security
 
+- Stop a compacted-session summary from leaking across users on a multi-user agent: `SessionStore::canonical_context` — the reader on the prompt-assembly path — returned the agent-scoped `compacted_summary` unconditionally, ignoring the `compacted_summary_session_id` owner column that #6225 added, so when several `[[users]]` shared one agent, a summary built from user A's conversation was injected verbatim into user B's turn (up to a 4000-char excerpt of A's private messages at prompt index 0) (#6493) (@houko).
+  The #6225 fix had scoped only the dashboard `/session` banner (`compacted_summary_for_session`) and left this injection reader ungated; `canonical_context` now applies the same owner check — a summary is surfaced only to the session that owns it, while a legacy ownerless row or an unscoped (`None`) agent-wide read is unchanged.
+  Exposed by default (the leak path runs whenever `stable_prefix_mode` is false, the default).
```
houko: Flagging: I don't think "owner-visible-only (not cross-user)" holds up. `append_canonical`'s heuristic compactor stamps `compacted_summary_session_id` with whichever session's turn happened to trigger compaction of the shared, cross-channel `canonical.messages` list — not with the set of sessions whose content actually got folded into the summary. On a multi-user agent those can differ, so the post-fix read-gate in this PR can still hand user A's compacted text to user B if B's turn was the one that crossed the compaction threshold. See my full comment on `session.rs` (the new `canonical_context` hunk) for the concrete scenario and line references. Worth a real security follow-up issue rather than a "left for a separate decision" CHANGELOG note.

---

houko (commented):
Reviewed the ownership-gating fix in `canonical_context` — the new match arm and both regression tests correctly implement the described gate (mismatch → None, unscoped/legacy → prior behaviour), and the CHANGELOG entry has the required `(@user)` attribution and follows the prose-wrap convention (one sentence per line). No mechanical fixes needed (no new HashMap/HashSet feeding a prompt, no new config fields, no prose violations).

One substantive concern on the PR's own "out of scope" deferral — left as inline comments rather than a fix since it's a design/semantics question, not a quick patch. Summary: I think the heuristic compactor in `append_canonical` can still leak one user's content to another user even after this fix, because the session it stamps as "owner" is the triggering session, not necessarily the session(s) whose content is in the summary. Details in the two line comments.

---

crates/librefang-memory/src/session.rs
houko, line 1: Re-derived the gating logic independently (not just trusting the PR description) and it checks out: `canonical_context` is the sole prompt-assembly reader (`crates/librefang-kernel/src/kernel/messaging.rs` and `agent_execution.rs` both call it with `Some(effective_session_id)`, no signature change needed), no other call site reads `compacted_summary` directly, and no parallel storage table exists. The `(Some(want), Some(owner)) if owner != want => None` match correctly withholds the summary only on an actual owner mismatch while preserving legacy/unscoped behavior. Good fix.

Two things worth a second look, neither of which I'm confident enough to edit myself:

**1. Correctness side-effect of the `append_canonical` fold filter (not a leak, but worth confirming is intended):** entries skipped by the new `(Some(got), Some(want)) if got != want => continue` filter still get swept away by the unconditional `canonical.messages = canonical.messages.split_off(to_compact)` a few lines below. Before this fix, a non-triggering session's messages in the compacted range at least survived (leaked) inside the shared summary text. After this fix, if they don't belong to the triggering session, they're now folded into nothing *and* trimmed from `messages` — i.e. permanently forgotten, with no representation anywhere. That's clearly better than leaking, but it is a new retention regression for the "other" session(s): their early conversation silently vanishes from the agent's memory whenever a *different* session happens to trigger compaction. Is that the intended tradeoff, or should the trim itself be restricted to entries that were actually folded (leaving other sessions' un-folded entries in `messages` until their own session triggers compaction)?

**2. Latent invariant, not exploitable today but worth hardening:** the per-entry filter `(Some(got), Some(want)) if got != want => continue` only skips cross-session entries when the *triggering* append's own `session_id` is `Some(_)`. If a future caller ever invokes `append_canonical` with `session_id: None` on an agent that already has per-session-tagged entries, every entry (including other sessions') would fold in unfiltered, and the result gets stamped `compacted_summary_session_id = None` — which the new read-gate in `canonical_context` treats as "legacy, return unconditionally," silently reopening the #6493 leak. I confirmed both current production call sites (`messaging.rs`, `agent_execution.rs`) always pass `Some(effective_session_id)` (the variable is typed `SessionId`, never `Option`), so this isn't live today — but there's no type-level guard or assertion preventing a future `None`-passing caller (e.g. a backfill/migration tool) from reintroducing it. Might be worth a comment or debug_assert here noting the invariant.

---

CHANGELOG.md
houko, line 1: Two small notes on this entry, neither blocking:

1. Per CLAUDE.md's prose-wrapping rule ("CHANGELOG.md bullets — one sentence per line within a bullet"), this new bullet should be split at sentence boundaries rather than left as one ~6-sentence line. I didn't fix it myself because every other bullet in this Unreleased section (including the #6494 entry right above it, from the same series) uses the same single-line-per-bullet style, so unilaterally reformatting just this one would look inconsistent with established practice here — flagging for a maintainer call rather than auto-fixing.

2. The PR description's "Out of scope (flagged, not fixed here)" section says the `append_canonical` cross-session fold issue was deliberately left unfixed ("owner-visible-only... not a cross-user leak"). That's stale — the actual diff (and this CHANGELOG bullet itself) shows it *was* fixed in this PR, with a dedicated regression test (`test_append_canonical_heuristic_summary_does_not_fold_other_sessions`). Worth updating the PR body for anyone reading it later, since it currently contradicts the shipped code.

---

houko (commented):
Daily automated review pass on this (already-merged) security fix. Re-derived the `canonical_context` gating logic and the call-site survey independently rather than trusting the description — the core fix is correct and complete for the prompt-injection path. Two non-blocking follow-up notes left inline (a retention side-effect worth a maintainer decision, and a latent invariant worth hardening against future regressions), plus a minor CHANGELOG note about the PR description being stale relative to the shipped diff. Nothing here rises to "block/revert" — all comment-only.

---

Issue #3863, opened 2026-04-28
Capability glob `*` matches across path/host/scheme separators — FileRead/NetConnect/ShellExec patterns leak grants


houko:
**Location**
- `crates/librefang-types/src/capability.rs:197-219` (`glob_matches`)

**Problem**
The `*` wildcard is byte-unbounded — it matches `/`, `:`, `.`, `@`, anything. There is no fnmatch-style FNM_PATHNAME equivalent.

```rust
pub fn glob_matches(pattern: &str, value: &str) -> bool {
    if pattern == "*" { return true; }
    if let Some(suffix) = pattern.strip_prefix('*') { return value.ends_with(suffix); }
    if let Some(prefix) = pattern.strip_suffix('*') { return value.starts_with(prefix); }
    if let Some(star_pos) = pattern.find('*') {
        let prefix = &pattern[..star_pos];
        let suffix = &pattern[star_pos + 1..];
        return value.starts_with(prefix) && value.ends_with(suffix) && value.len() >= prefix.len() + suffix.len();
    }
    false
}
```

**Impact**
Concrete escalations:
- `FileRead("/home/user/*")` matches `/home/user/../../etc/shadow` (the literal string passes the capability gate; combined with the canonicalize-after-check bug separately filed as #3457 / #3814, this is exploitable end-to-end).
- `NetConnect("api.openai.com:*")` matches `api.openai.com:443@evil.com:80` — the `:` in userinfo or path is wildcarded over.
- `ShellExec("git*")` matches `git; rm -rf /` and `/usr/bin/git-backdoor`. `host_shell_exec` reads `command` from guest and binds the entire string against this glob.
- `AgentMessage("billing-*")` matches `billing-x/../admin` if downstream namespacing allows `/`.

**Fix**
Either:
1. Make `*` non-traversing across separators (FNM_PATHNAME style: `*` does not match `/` for path-typed capabilities, does not match `:` for host:port, etc.).
2. Better: replace string globs with type-specific match logic — `Capability::FileRead` checks `requested_path.starts_with(allowed_path)` after path normalization; `NetConnect` parses host+port and matches each piece; `ShellExec` requires absolute argv[0] equality.

**Verified-not-duplicate-of:** #3457 / #3814 are about canonicalization order *after* the cap check. This issue is the cap-check itself: even if canonicalization were perfect, the glob algorithm leaks grants across separators.

---
[...]

PR #6498, opened 2026-07-19
fix(kernel): honour glob patterns in per-agent tool_allowlist/tool_blocklist (#6495)

### LINKED ISSUE #6495: tool_allowlist doesn't support glob — silently strips all MCP tools when set
## Summary

`tool_allowlist` in an agent manifest silently strips all MCP tools when set. Step 1 (`capabilities.tools`) uses `glob_matches`, Step 4 (`tool_allowlist`) uses exact `==`. MCP tool names are dynamic (`mcp_<server>_<tool>`), so enumerating them upfront is impractical — the result is that `tool_allowlist` and MCP are effectively mutually exclusive.

Silent failure mode: model receives only native tools, hallucinates about missing MCP capabilities.

## Repro

```toml
# agent.toml
mcp_servers = ["*"]
tool_allowlist = ["file_read", "web_search"]

[capabilities]
tools = ["file_read", "web_search"]
```

- `GET /api/tools` — kernel-wide catalog shows N MCP tools ✓
- Prompt sent to LLM — only `file_read`, `web_search`; zero `mcp_*` tools
- `tool_allowlist = ["file_read", "web_search", "mcp_*"]` — no change (glob not honored)

## Root cause

`crates/librefang-kernel/src/kernel/tools_and_skills.rs` in `available_tools()`:

- Step 1 (builtin filter): `.filter(|t| declared_tools.iter().any(|d| glob_matches(d, &t.name)))`
- Step 4 (final retain): `all_tools.retain(|t| tool_allowlist.iter().any(|a| a == &t.name))`

Asymmetric glob support in the same file.

## Suggested fix

Use `glob_matches(a, &t.name)` in Step 4 (both allowlist and blocklist) — smallest diff, mirrors Step 1. Users can then write `["file_read", "mcp_*"]` or `["mcp_notion_*"]`.

## Environment

- LibreFang `v2026.7.11`, verified at that exact tag
- Behavior identical in `v2026.6.29` — long-standing gap surfaced when an agent first tries to call an MCP tool

houko:
Fixes #6495.

## The bug

`available_tools` in `crates/librefang-kernel/src/kernel/tools_and_skills.rs` filters tools in stages. Step 1 (builtin/skill filtering) already matches with `glob_matches` (`:142`). But Step 4 — the per-agent `tool_allowlist` / `tool_blocklist` retain (`:302` / `:305`) — used exact string equality:

```rust
all_tools.retain(|t| tool_allowlist.iter().any(|a| a == &t.name));
```

MCP tool names are runtime-generated and namespaced `mcp__<server>__<tool>` (confirmed via `tool_classifier.rs:177` `name.starts_with("mcp__")`), so an operator cannot list them as static literals. The consequences under `==`:

- A non-empty `tool_allowlist` retained only the exactly-named **native** tools and **silently dropped every `mcp__*` tool** — no WARN.
- A `mcp__*` glob entry didn't help: `*` was a literal, so it matched nothing.

The result flows into the cached tool set that becomes the LLM prompt, so the model just silently loses all MCP tools whenever an allowlist is set.

## The fix

Both retains now use `glob_matches`, mirroring Step 1:

```rust
all_tools.retain(|t| tool_allowlist.iter().any(|a| glob_matches(a, &t.name)));
all_tools.retain(|t| !tool_blocklist.iter().any(|b| glob_matches(b, &t.name)));
```

`glob_matches` is already imported in this module. It keeps matching a bare tool name exactly (its `pattern == value` fast path), so existing exact-name allowlists are **not** regressed. Operators can now write `["file_read", "mcp__notion__*"]` and keep the MCP tools.

MCP names contain `__` but no `/`, so they route through `glob_matches_simple` where `*` matches freely — `mcp__notion__*` correctly spans `mcp__notion__search` etc., and the path-separator hardening in `glob_matches` (#3863/#3925) is irrelevant here.

## Tests

Three cases added to `mcp_tools_list_allowlist_test.rs`, using the **real double-underscore** MCP name shape (the existing tests used single-underscore names that pass even under exact matching, so they did not cover this):

- `tools_list_allowlist_supports_mcp_glob` — `["mcp__notion__*"]` retains the notion tools, drops `mcp__github__create_issue`.
- `tools_list_allowlist_exact_name_still_works` — an exact native name still matches (no regression).
- `tools_list_blocklist_supports_mcp_glob` — `["mcp__github__*"]` blocklist hides that namespace, leaves others visible.

All 7 tests in the file pass.

## Out of scope (noted, not done)

The reporter and I both note a secondary hardening: emit a WARN when a non-empty `tool_allowlist` reduces the MCP tool count to zero, so the silent-strip footgun surfaces in logs. That's an additive log line, not part of the correctness fix, and can follow if wanted.

## Verification

- `cargo check -p librefang-kernel --lib` — clean (via `Dockerfile.rust-dev`, Linux).
- `cargo test -p librefang-api --test mcp_tools_list_allowlist_test` — 7 passed, 0 failed.


houko (commented):
CLAUDE.md compliance pass on this PR:

- Fixed directly (pushed as `372a649ba`): the new `tool_allowlist`/`tool_blocklist` doc-comment in `crates/librefang-kernel/src/kernel/tools_and_skills.rs` and the four new doc-comments added to `crates/librefang-api/tests/mcp_tools_list_allowlist_test.rs` were hard-wrapped at ~72-80 columns. CLAUDE.md's prose rule requires breaking only at sentence boundaries for touched doc-comments, so these were reflowed (no logic change). Re-ran `cargo check -p librefang-kernel --lib`, `cargo check -p librefang-api --tests`, `cargo clippy -p librefang-kernel --lib -- -D warnings`, and `cargo test -p librefang-api --test mcp_tools_list_allowlist_test` (7/7 pass) afterward to confirm nothing broke.
- Left as-is (needs your call, not mine to force-push over): the original commit body (`e5c23a20a`, now carried forward) is also hard-wrapped at a column limit rather than one-sentence-per-line, which is the same CLAUDE.md rule. Fixing that requires rewriting an already-pushed commit message, which means a history rewrite/force-push — outside what an automated compliance pass should do unilaterally. Flagging so you can reword it in a follow-up amend/rebase if you'd like, entirely your call.

No other CLAUDE.md violations found (no new HashMap/HashSet reaching an LLM prompt, no config-field wiring gaps, no AI/Claude attribution, CHANGELOG entry already follows the existing one-line-per-bullet convention with `(@houko)` attribution, and the new tests already exercise a real router the same way the rest of this test file does).

---

crates/librefang-kernel/src/kernel/tools_and_skills.rs
```diff
@@ -298,11 +298,14 @@ impl LibreFangKernel {
             })
             .unwrap_or_default();
 
+        // Match with glob_matches (mirroring the Step 1 builtin filter above), not exact string equality.
+        // MCP tool names are runtime-generated and namespaced (`mcp__<server>__<tool>`), so they can never be enumerated as static literals in a manifest — a non-empty exact-match allowlist silently dropped every mcp__* tool, and a `mcp__*` glob entry was itself treated as a literal and matched nothing (#6495).
+        // glob_matches still matches a bare tool name exactly (`pattern == value`), so plain native-tool allowlists behave exactly as before.
         if !tool_allowlist.is_empty() {
-            all_tools.retain(|t| tool_allowlist.iter().any(|a| a == &t.name));
+            all_tools.retain(|t| tool_allowlist.iter().any(|a| glob_matches(a, &t.name)));
```
houko, line 305: Re: the "Out of scope (noted, not done)" item in the PR description (the WARN when a non-empty `tool_allowlist` zeroes out the MCP tool count) — per CLAUDE.md's "fix what you found, don't punt it to a follow-up" rule, the bar for deferring is whether the fix requires touching a different crate or domain than the one already being edited.

This WARN would live in this exact function, in this exact file, right next to the `retain` calls it's warning about — same crate, same domain. Read literally, that puts it inside "fix it in this PR", not "open a follow-up".

That said, I can see the counter-argument: it's additive observability, not a correctness fix, and the correctness bug is what #6495 asked for — so it's a genuine judgment call rather than a clear-cut miss. Flagging for the maintainer to decide rather than either committing it myself or waving it through silently.

---

houko (commented):
Automated daily review pass against CLAUDE.md conventions (deterministic prompt ordering, CHANGELOG attribution, prose-wrapping, config Default-impl, fix-vs-follow-up). One item flagged inline; everything else checks out.

---

PR #6503, opened 2026-07-19
fix(taint): stop over-blocking MCP arguments that carry a long numeric id (#6499)

### LINKED ISSUE #6499: bug(taint): opaque-token heuristic false-positives long attachment filenames as credentials
## Summary

The `OpaqueToken` branch of `detect_secret_rule_with_skip`
(`crates/librefang-types/src/taint.rs:457-463`) misclassifies long
**attachment filenames** as opaque credential tokens. When such a
filename is passed as an MCP tool argument (e.g. the `pattern` /
filename argument to a filesystem MCP server), the `Secret`-labelled
value is blocked at the MCP sink
(`crates/librefang-runtime-mcp/src/lib.rs:539`) with:

```
taint violation: sensitive value in MCP argument '<path>' (blocked by sink '<sink>')
```

This is a false positive: the string is a benign filename, not a secret.

## Root cause

The opaque-token heuristic flags any string that is:

- `len >= 32`
- charset limited to `[A-Za-z0-9-_./+=]`
- contains at least one letter **and** at least one digit, and is not pure-hex
- has no `YYYY-MM-DD` date component
- does **not** satisfy `looks_like_path`

The `looks_like_path` guard (`taint.rs:446-455`) only recognises a
string as a path when it starts with `/`, has a `C:\`-style drive
prefix, starts with `\\`, or **contains a `/`** followed by an
extension. A **bare filename with no directory separator** — the exact
shape of an attachment name — therefore is *not* recognised as a path,
so a long filename containing a run of digits (common in exported /
auto-generated attachment names) trips `looks_opaque` and is labelled
`Secret`.

The letters in the filename (`foo-…-….pdf`) satisfy `has_letter`, and
the long digit run satisfies `has_digit`, so `mixed_enough` is true even
though the "mixing" is really `letters-in-prefix + one-long-decimal-run`
rather than genuine token entropy.

(Note: the PII/credit-card regexes do **not** fire here — the
`tokenish_mixed` early-out in `detect_pii_rule_with_skip` short-circuits
any single-token string that contains a letter. The whole false positive
comes from `OpaqueToken`.)

## Reproduction (minimal, synthetic)

Add to the `taint.rs` test module:

```rust
#[test]
fn opaque_token_false_positives_long_filenames() {
    let sink = TaintSink::mcp_tool_call();
    let skip = std::collections::HashSet::new();

    // Long export-id-style attachment filename, no directory separator.
    let name = "note-6571000000000000000-1000000000.pdf"; // 39 chars, synthetic
    let hits = detect_outbound_text_violation_rules_with_skip(name, &sink, &skip);
    assert!(hits.is_empty(), "benign filename flagged as: {hits:?}");
}
```

Observed (measured against `detect_secret_rule_with_skip` directly):

| argument (synthetic)                          | len | fired rule    | blocked |
|-----------------------------------------------|-----|---------------|---------|
| `note-6571000000000000000-1000000000.pdf`     | 39  | `OpaqueToken`  | **yes** |
| `export-6512345678901234567-1000000000.pdf`   | 41  | `OpaqueToken`  | **yes** |
| `invoice-4123456789012345.pdf`                | 28  | none          | no (under 32) |
| `short-123.pdf`                               | 13  | none          | no      |

The same value passed as a filesystem-MCP `pattern`/name argument is
rejected before it reaches the server, so an agent cannot search for or
open the file by name.

## Impact

An agent trying to locate/open a stored attachment by its filename gets
its tool call blocked and cannot recover by retry (the name is
deterministic). If the agent has a consecutive-tool-failure guard, the
whole turn aborts. The filename is not sensitive data, so nothing is
actually protected.

## Suggested fixes (any one)

1. **Extend `looks_like_path`** to also treat a bare basename with a
   recognised file extension as a structured identifier (e.g. matches
   `^[\w.\-]+\.[A-Za-z0-9]{1,8}$` with no `/`), so filenames are
   excluded from the opaque-token heuristic — symmetric with the
   existing "path with `/` + extension" exclusion.
2. **Tighten `mixed_enough`** so a single long contiguous decimal run
   plus alpha prefix/suffix does not count as opaque-token entropy
   (require intra-string alternation of char classes, or a minimum
   count of distinct non-digit segments).
3. Both.

A per-path taint policy (skip-rules) exists as a runtime mitigation, but
the default heuristic should not require every deployment to allowlist
its own attachment filenames.

## Environment

- `crates/librefang-types/src/taint.rs` `detect_secret_rule_with_skip`, `OpaqueToken` branch
- `crates/librefang-runtime-mcp/src/lib.rs:539` (MCP outbound sink enforcement)
- Reproduced against current `upstream/main`.

houko:
Fixes #6499.

The outbound taint/DLP heuristic false-positive-blocked (fail-**closed**) legitimate MCP tool arguments that carry a long unformatted numeric id — it blocks valid traffic, it does not leak anything. Two rules, one root cause.

## OpaqueToken — `crates/librefang-types/src/taint.rs`

`looks_like_path` only recognized a string as structured (and thus not a credential) when it had a directory separator:

```rust
let looks_like_path = trimmed.starts_with('/')
    || (… C:\ …) || trimmed.starts_with("\\\\")
    || (trimmed.contains('/') && … extension after last '/' …);
```

So a bare export-attachment filename like `note-6571000000000000000-1000000000.pdf` (39 chars, no `/`) failed every branch and then satisfied every `looks_opaque` conjunct (`len >= 32` + mixed-alnum + non-hex + no date) → flagged `OpaqueToken`.

**Fix:** add a branch that excludes a bare basename with a **letter-initial** extension: `^[\w.\-]+\.[A-Za-z][A-Za-z0-9]{0,7}$`. The extension must start with a letter, so a numeric-suffixed opaque token (`abc123.9876543`) is **not** treated as a filename and is still scanned — keeping the false-negative surface narrow.

## PiiPhone

The phone regex makes every separator optional (`(?:\+…)?\(?\d{2,4}\)?[\s.\-]?\d{3,4}[\s.\-]?\d{3,4}`), so any contiguous digit run matched its digit-group floor and was flagged a phone number, e.g. a 19-digit id or a `WHERE id LIKE '%export-<id>%'` argument.

**Fix:** `phone_match_is_real` expands each regex match to the full contiguous digit run it sits in and rejects the match only when that run exceeds `MAX_PHONE_DIGITS` (16 — E.164 caps a subscriber number at 15). A 10-15 digit number is still treated as a possible phone; a 13-16 digit run is left to the credit-card rule rather than silently dropped.

## Scope honesty (two things I deliberately did *not* do)

- **Did not touch the credit-card rule.** A 19-digit id beginning `65..` matches the Discover BIN on its first 16 digits — that's the card rule slicing a long run, the same class of issue but a separate, higher-risk PII rule that #6499 didn't report. The regression tests use ids starting 7/8/9 to stay out of the 3/4/5/6 card-BIN ranges and keep this PR focused.
- **Did not try to disambiguate a bare 10-15 digit id from a phone number.** Those lengths are genuinely ambiguous, so they still fire the phone rule (conservative — the runtime `McpTaintPolicy` per-path skip-rules cover that case). Only `>16`-digit runs, which cannot be phone numbers, are let through.

An earlier iteration regressed the Amex-15 test (`340000000000009`, which had leaned on the phone rule for coverage since it doesn't match the 16-digit card regex); the run-length approach fixes that — surfaced by running the full suite, not just the new tests.

## Tests

`crates/librefang-types/src/taint.rs` `mod tests`:
- `test_long_export_filename_is_not_opaque_token_6499` — the reported filenames pass.
- `test_numeric_suffix_token_still_flagged_after_6499` — counter-test: `.<digits>`-suffixed token still flagged.
- `test_long_numeric_id_is_not_a_phone_number_6499` — >16-digit ids (bare / embedded / in SQL) pass.
- `test_real_phone_numbers_still_flagged_after_6499` — counter-test: `+1-555-123-4567`, `(555) 123-4567`, etc. still flagged.

## Verification (via `Dockerfile.rust-dev`, Linux)

- `cargo test -p librefang-types --lib` — 885 passed, 0 failed (incl. the pre-existing OpaqueToken / phone / credit-card counter-tests — no regression).
- `cargo clippy -p librefang-types --lib -- -D warnings` — clean.

## Security-posture note

Loosening a DLP heuristic trades fewer false positives for a slightly wider false-negative surface. Both changes are bounded (letter-initial extension only; >16-digit runs only) and every existing positive-detection test still passes. Happy to tighten further if you'd rather bound the filename exclusion by length too.


crates/librefang-types/src/taint.rs
```diff
@@ -531,6 +540,62 @@ fn detect_pii_rule_with_skip(
...
+/// long attachment filenames from the opaque-token heuristic (#6499). The
+/// extension must start with a letter and be 1-8 alphanumerics, so a token
+/// with a numeric suffix (`abc123.9876543`) is not treated as a filename and
+/// is still scanned as a potential opaque secret. The body allows only the
+/// filename charset (word chars, `.`, `-`), so a value with other punctuation
+/// is not mistaken for a filename.
+fn bare_filename_regex() -> &'static Regex {
+    static BARE_FILENAME: OnceLock<Regex> = OnceLock::new();
+    BARE_FILENAME.get_or_init(|| {
+        Regex::new(r"^[\w.\-]+\.[A-Za-z][A-Za-z0-9]{0,7}$")
```
houko, line 594: `bare_filename_regex` (`^[\w.\-]+\.[A-Za-z][A-Za-z0-9]{0,7}$`) validates only the *shape* of the extension (starts with a letter, 1-8 alphanumerics), not membership in an actual known-extension list. That makes the `OpaqueToken` exclusion bypassable by appending any short letter-led suffix to an otherwise-detected secret — not just a real numeric suffix, which the PR's counter-test (`test_numeric_suffix_token_still_flagged_after_6499`) already covers.

Verified locally (temporary test, not committed):
```rust
let secret_with_fake_ext = "0p3nai_sk_proj_abcXYZ1234567890qwertZXCV.log";
check_outbound_text_violation(secret_with_fake_ext, &TaintSink::mcp_tool_call())
// -> None (same secret without ".log" -> Some(...), see test_check_outbound_text_still_flags_opaque_token)
```
`.log`, `.data`, `.foo`, `.com`, etc. all match `[A-Za-z][A-Za-z0-9]{0,7}$` just as well as `.pdf`/`.docx` do, so this isn't limited to genuine attachment filenames — any opaque-token-shaped payload can dodge `OpaqueToken` by tacking on a plausible-looking short suffix. That directly contradicts the stated goal in the doc comment above ("keeps the false-negative surface narrow") and in the PR body ("keeping the false-negative surface narrow").

Note the pre-existing slash-based path branch just above (`trimmed.contains('/') && ... extension after last '/'`) has the same unbounded-extension property, so this isn't a new pattern introduced by this line alone — but this PR widens it from "look like a path" (needs `/`) to "look like *any* bare `name.ext`", which is a much larger slice of possible payload shapes (including tokens that were never meant to look like a filename at all).

Given the PR body already flags this exact tradeoff ("Happy to tighten further if you'd rather bound the filename exclusion by length too") — worth considering a bounded allowlist of real extensions (`pdf`, `docx`, `xlsx`, `csv`, `json`, `png`, `jpg`, `zip`, `txt`, `log`, …) rather than an open-ended shape match, since the shape alone doesn't actually distinguish "real filename" from "opaque token with a short suffix appended". Not blocking — flagging since it weakens a security control (DLP/taint heuristic) in a way that's easy to route around once known.

---

houko (commented):
Automated daily review pass. One finding worth a look — the rest of the diff (CHANGELOG formatting, comment style, phone-digit-run logic, test coverage) checked out clean against CLAUDE.md and this PR's own stated scope.

---

Issue #6492, opened 2026-07-17
Tool-approval gate broken via REST `/message` path: trusted_senders bypass, no turn-resume, queue zombies, CLI 415 (v2026.6.29 & v2026.7.11)


otaviobrum:
### Description

We rely on `[approval] auto_approve` + `trusted_senders` to let the operator's own trusted requests skip the approval gate, and on the approval flow to let a human clear anything else. Live end-to-end testing found this circuit broken in **four independent ways**, all reproducing over the REST endpoint `POST /api/agents/{id}/message` (the path used by our WhatsApp webhook bridge and equivalent integrations). The native chat UI also reproduces; this report focuses on the REST path since that is where `sender_id` is passed explicitly.

We re-ran the identical end-to-end checklist after upgrading from v2026.6.29 to v2026.7.11 and **all four defects are still present, unchanged**.

**Bug 1 — `trusted_senders`/`auto_approve` not honored on `POST /api/agents/{id}/message`.** A message with `sender_id` literally present in `trusted_senders` still gets gated, even for low-risk tools (`memory_recall`/`memory_store`). The gate does not appear to map the message's `sender_id`/`sender_name` to the `user_id` it compares against `trusted_senders` on this endpoint.

**Bug 2 — Approving a queued call does not resume the agent's turn.** Approving via `POST /api/approvals/{id}/approve` transitions `approval_audit` to `approved`, but the suspended turn never completes. The daemon logs `WARN No active adapter for the originating channel — agent reply produced but cannot be delivered; session has the reply persisted so the next user turn surfaces it channel="whatsapp"`. The result only surfaces attached to the *next unrelated* user turn — unusable for deferred/async channels where there may not be a next turn.

**Bug 3 — Live approvals queue does not evict terminal-state entries.** `approved`/`timed_out`/`expired` entries keep appearing in the live queue as if pending (diverging from `approval_audit`, which is correct), and re-resolving them throws `internal_error: Already <state> by <who>` instead of a no-op or typed 4xx. Zombies only clear on full daemon restart.

**Bug 4 — CLI `approvals approve` returns 415 and prints a false success message.** The CLI's underlying HTTP call is missing/mis-setting `Content-Type: application/json`; the server rejects it with `415 Unsupported Media Type`, yet the CLI prints `✔ approved` while the approval's real status stays `pending`. The same request done directly against the API with the correct header succeeds.

**Relevant config** (`[approval]` block, confirmed loaded at runtime via `GET /api/config/export`):

[...]

PR #6507, opened 2026-07-19
docs: explain trusted_senders vs [[users]] RBAC composition (#6492)

houko:
## Summary

Documentation-only change fulfilling the docs pass @houko committed to in #6492 "regardless" of the other bug decisions.

The security docs mentioned `[approval].trusted_senders` but never explained how it composes with `[[users]]` RBAC and the guest-gate. Users were confused because a `trusted_senders` entry does **not** bypass the RBAC guest-gate.

Adds a new subsection — "How `trusted_senders` composes with `[[users]]` RBAC" — to the approvals security page and its Simplified-Chinese mirror.

## What the subsection says

- `[approval].trusted_senders` and `[[users]]` RBAC are two separate trust surfaces. `trusted_senders` governs only the approval layer (waives approval for low-risk tools; high-risk tools stay gated). `[[users]]` RBAC governs the per-user tool gate.
- The RBAC gate is evaluated **first** and independently. When `[[users]]` is configured and the REST `sender_id` does not resolve to a registered user on the `api` channel, the sender falls to `guest_gate`, whose read-only allowlist does not include `memory_store` / `memory_recall` — so it returns "needs approval". That forced-approval verdict short-circuits (`force_approval || requires_approval_with_context(...)`) before the `trusted_senders` bypass is ever consulted.
- Consequence: an ID in `trusted_senders` that is not also a registered `[[users]]` on the `api` channel still has its low-risk `memory_*` calls gated.
- Concrete fix (with a minimal TOML snippet): register the operator as `[[users]]` bound to the `api` channel with a `tool_policy` allowing the tools it needs; the RBAC gate then returns `Allow` and the `trusted_senders` bypass applies on top as designed.
- With no `[[users]]` configured, RBAC is inactive and `trusted_senders` works standalone.

## Files changed

- `docs/src/app/security/approvals/page.mdx` — new subsection under Approval Policy.
- `docs/src/app/zh/security/approvals/page.mdx` — equivalent Simplified-Chinese subsection.
- `CHANGELOG.md` — `### Documentation` bullet under `[Unreleased]`.

## Verification

- Every behavioural claim verified against current source (`crates/librefang-kernel/src/auth.rs` `resolve_user_tool_decision` / `guest_gate`, `crates/librefang-runtime/src/tool_runner/dispatch.rs` `force_approval` short-circuit, `crates/librefang-kernel/src/approval.rs` `requires_approval_with_context` / `classify_risk`, `crates/librefang-api/src/routes/agents/mod.rs` `api` channel + `sender_id`). The cited paths had not drifted.
- TOML snippet validated with `tomllib`; MDX code-fences balanced in both files; internal `#hardening-checklist` anchor present in both.

## Scope

This is the committed docs pass on `trusted_senders` vs RBAC composition only. It does **not** resolve Bug 1 (needs the reporter's config to reproduce) or Bug 2 (needs a design decision) tracked in #6492.

Closes the docs commitment in #6492.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project uses [Calendar Versioning](https://calver.org/) (YYYY.M.DD).
 
 ## [Unreleased]
 
+### Documentation
+
+- Document how `[approval].trusted_senders` composes with `[[users]]` RBAC on the approvals security page (EN + zh mirror): the two are separate trust surfaces and the per-user RBAC gate is evaluated first, so an ID listed in `trusted_senders` that is not also a registered `[[users]]` on the `api` channel still has its low-risk tools (e.g. `memory_*`) gated by the `guest_gate`, because the forced-approval verdict short-circuits before the `trusted_senders` bypass is consulted; the new subsection gives the concrete fix (register the operator as a `[[users]]` bound to the `api` channel with a `tool_policy` covering the tools it drives) and notes that with no `[[users]]` configured `trusted_senders` works standalone (#6492) (@houko)
+
 ### Fixed
 
 - Pin `crates/librefang-api/src/login_page.html` to LF in `.gitattributes`: the file is embedded verbatim via `include_str!` and its inline `<script>` is authorised by an exact SHA-256 baked into the CSP (`middleware.rs`), so a Windows checkout with `core.autocrlf=true` rewrote it to CRLF, shifted the computed hash, and failed `dashboard_login_page_script_is_allowed_by_csp_hash` on the Windows shard only — turning `main` red on every merge since #6486 touched the page (#6481) (@houko)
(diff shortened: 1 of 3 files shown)


houko (commented):
Verified the anchor claim in the PR description against `docs/src/mdx/rehype.mjs`'s `rehypeSlugify()`: `@sindresorhus/slugify` strips CJK characters, so it falls back to a Unicode-aware slug that preserves the raw CJK text. That means the EN `## Hardening Checklist` heading really does slugify to `hardening-checklist`, but the zh `## 加固清单` heading slugifies to `加固清单` — not `hardening-checklist`. So the new zh subsection's `[加固清单](#hardening-checklist)` link was pointing at a non-existent anchor. Pushed a follow-up commit fixing it to `(#加固清单)`.

FYI (not blocking this PR): the same page's zh TOC already has several pre-existing links using the same wrong English-style slug pattern for other Chinese-only headings (e.g. `[审批策略](#approval-policy)` and `[恢复码](#recovery-codes)` near the top of `docs/src/app/zh/security/approvals/page.mdx`), while `[会话级批量审批](#会话级批量审批)` correctly uses the native-CJK anchor. That predates this PR, so I left it alone here, but it's probably worth a small separate cleanup issue/PR across the zh docs tree.

---

PR #6512, opened 2026-07-20
fix(llm-drivers): surface mid-stream provider errors instead of empty/garbled turns

houko:
## Summary

Streaming paths in `librefang-llm-drivers` silently mishandled a provider error that arrives **mid-stream over an already-HTTP-200 SSE body**. The pre-stream guards only inspect `resp.status()` before the body is consumed, so an error frame delivered *after* the 200 was never classified — the turn ended as a truncated/empty "success" (no retry, no failover, billed as success) or produced garbled concatenated output.

Found during a repo-wide bug audit; hardened after an xhigh code review (below).

## Changes

**Surface the error (drivers):**
- **anthropic** (`drivers/anthropic.rs`): an `event: error` frame (commonly `overloaded_error` under load) fell into the catch-all `_ => {}` and was dropped; the loop then returned `Ok(EndTurn, <partial/empty>)`. It now returns the typed `LlmError::Overloaded` / `RateLimited` / `Api` mapped from `error.type`.
- **openai** (`drivers/openai.rs`): OpenRouter/Groq send a terminal `data: {"error": …}` frame instead of a non-2xx status; that frame has no `choices`, so the loop hit `None => continue` and returned empty content. It now surfaces the frame as `LlmError::Api` classified via `openai_stream_error_code`.

**Don't corrupt output when re-running a provider (failover + retry layers):**
- **fallback** (`drivers/fallback.rs`): `FallbackDriver::stream` advanced to the next driver on the same caller channel after content was already forwarded, concatenating a second response. It now wraps the inner stream in the same `content_emitted` intercept relay `FallbackChain::stream` uses and propagates the error once content has reached the caller (and no longer health-penalizes a provider that served content before a transient error).
- **runtime** (`librefang-runtime/src/agent_loop/retry.rs`): `stream_with_retry` re-streamed a full second response onto the same `tx` on a retryable mid-stream error. It now tracks whether content reached the caller and surfaces the error instead of retrying once it has (retry still applies when the error precedes any content).

## Review hardening (xhigh review of the initial commit)

The initial "surface the error" change introduced two confirmed regressions, both fixed in the follow-up commit:

- **Retry-layer duplication** — the most severe: returning a *retryable* error mid-stream fed `stream_with_retry`, which had no content-emitted guard → duplicated user-visible output. Fixed by the runtime guard above.
- **OpenAI benign-error false positive** — the terminal check treated any non-empty `error` object as fatal, so a normal chunk carrying `"error": {"message":null,…}` aborted the stream and discarded valid content. The check now fires only on a real signal (non-null message/code/type, non-empty string, or bare number/array); `"error": null` and all-null objects are ignored, and non-object shapes are surfaced instead of dropped.

Also from the review: `openai_stream_error_code` now accepts OpenRouter's numeric `code` (`429`), and `LlmError::Api.status` is derived from the typed code (429/503/402/…) rather than a blanket 502.

## Tests (all green)

- `anthropic::tests::streaming_error_frame_surfaces_as_error`
- `openai::tests::streamed_error_frame_surfaces_as_error` (asserts status 429 via typed code)
- `openai::tests::streamed_numeric_error_code_classifies_rate_limit`
- `openai::tests::streamed_null_error_field_is_ignored`, `streamed_all_null_error_object_is_ignored`
- `fallback::tests::stream_does_not_fail_over_after_content_emitted` (+ health-penalty assertion)
- `retry::tests::no_retry_after_partial_content_on_overload` (the duplication regression)

## Verification

- `cargo test -p librefang-llm-drivers --lib` — **600 passed**
- `cargo test -p librefang-runtime --lib no_retry_after_partial_content_on_overload` — passed
- `cargo clippy -p librefang-llm-drivers -p librefang-runtime --all-targets -- -D warnings` — clean
- `rustfmt --check` — clean

## Crate footprint

`librefang-llm-drivers` + `librefang-runtime` (the retry-layer guard). No cross-crate contract changes — both use existing `LlmError` / `StreamEvent` types.

## Out-of-scope follow-ups (from the same audit)

Separately scoped; #6513 already lands the token-reservation RAII leak. Remaining: per-user usage attribution/budget bypass (kernel), alias-provider key lookup (kernel), session-lock namespace (kernel), browser config-reload no-op (kernel), Gemini/Vertex `response_format` drop (llm-drivers), MCP `knowledge_add_*` `agent_id=""` (5-crate).


crates/librefang-llm-drivers/src/drivers/fallback.rs
```diff
@@ -383,9 +384,55 @@ impl LlmDriver for FallbackDriver {
...
+                    }
+                    if tx_relay.send(event).await.is_err() {
+                        // Downstream caller dropped the receiver; close the
+                        // relay inbound so the wrapped driver's next send fails
+                        // and it aborts the upstream stream (#3769).
+                        tracing::debug!(
+                            "FallbackDriver(stream): downstream receiver dropped; cancelling inner driver"
+                        );
+                        intercept_rx.close();
+                        break;
```
houko: This content-emitted intercept relay (watch channel + mpsc relay task + `is_content` match) is a near line-for-line duplicate of the block already in `FallbackChain::stream` (`crates/librefang-llm-drivers/src/drivers/fallback_chain.rs:391-425`) — same variable names, same comments, same `is_content` match arm list. Verified: I diffed the two blocks and they differ only in which `entry`/`req` locals they close over.

Given both call sites now exist, worth considering extracting this into a shared helper (e.g. `fn spawn_content_emitted_relay(tx: Sender<StreamEvent>) -> (Sender<StreamEvent>, JoinHandle<()>, watch::Receiver<bool>)` or similar) so a future change to the content-classification set (e.g. if a new `StreamEvent` variant needs to count as "content") only has to be made once instead of drifting between the two copies. Not blocking this fix — flagging since it's the kind of duplication that's easy to miss keeping in sync later.

---

houko:
Routine review pass on this PR: diff logic checks out (see inline comment on `fallback.rs`), and verification was clean — `cargo check -p librefang-llm-drivers --lib`, `cargo clippy -p librefang-llm-drivers --all-targets -- -D warnings`, `cargo fmt --check` on the three changed Rust files, and `cargo test -p librefang-llm-drivers --lib` (598 passed, including the four new regression tests) all pass with zero findings.

One thing worth flagging separately: commit `2aa7802` ("docs(changelog): one sentence per line in the streaming-error entries") has both author and committer identity set to `Claude <noreply@anthropic.com>`. The repo's commit-msg hook only scans commit message text for AI attribution strings, so it doesn't catch attribution carried in the git identity metadata itself. Noting this rather than rewriting — the commit is already pushed and history rewrites on a branch already under review should be a maintainer call.

---

crates/librefang-llm-drivers/src/drivers/anthropic.rs
```diff
@@ -991,6 +991,41 @@ impl LlmDriver for AnthropicDriver {
...
+                                    retry_after_ms: 5000,
+                                },
+                                Some("rate_limit_error") => LlmError::RateLimited {
+                                    retry_after_ms: 5000,
+                                    message: Some(message),
+                                },
+                                _ => LlmError::Api {
+                                    status: 200,
+                                    message,
+                                    code: anthropic_error_code(kind, 200),
```
houko, line 1016: The catch-all arm (`anthropic.rs:1022-1026`) hardcodes `status: 200` for every mid-stream error type other than `overloaded_error`/`rate_limit_error` — i.e. `authentication_error`, `billing_error`, `not_found_error`, `invalid_request_error`, `api_error`. Two effects worth a second look:

1. `anthropic_error_code(kind, 200)` is called with the literal `200`, not a real status. The `invalid_request_error` branch inside `anthropic_error_code` (line 281-291) relies on `status == 413` to distinguish a context-length overflow from a generic bad request; with `200` hardcoded it can never take that branch, so `failover_reason()`'s `413 => ContextTooLong` check (`lib.rs:176`) is unreachable for this path — a mid-stream `invalid_request_error` will always classify as `HttpError` even in cases that would otherwise be `ContextTooLong`.
2. Any consumer reading `LlmError::Api.status` directly (logs/metrics) sees `200` stamped on an actual auth/billing/model-not-found error, which reads as "success" despite being an error path.

This is the same category of issue the OpenAI side of this same PR explicitly fixed via `status_for_stream_error_code` (deriving status from the typed code instead of a blanket value) — the Anthropic driver doesn't get the equivalent treatment. It may be intentional (the transport-level HTTP status genuinely was 200 for a mid-stream frame, and Anthropic likely doesn't emit a real mid-stream `invalid_request_error`/context-overflow in practice, since that's normally caught pre-stream), but flagging for a maintainer call since it's inconsistent with the sibling fix in this same PR.

---

houko (commented):
Reviewed the diff against CLAUDE.md and for correctness. Fixed and pushed one confirmed bug (bare empty-string mid-stream error frame incorrectly treated as a real signal in the OpenAI driver, contradicting this PR's own stated design). Left one inline comment on a status-code inconsistency between the Anthropic and OpenAI mid-stream error paths that I'm not fully confident warrants a code change — leaving it for maintainer judgment.

---

crates/librefang-llm-drivers/src/drivers/openai.rs
```diff
@@ -1933,6 +1987,62 @@ impl LlmDriver for OpenAIDriver {
...
+                            .get("code")
+                            .and_then(|c| {
+                                c.as_str()
+                                    .filter(|s| !s.is_empty())
+                                    .map(str::to_string)
+                                    .or_else(|| c.as_u64().map(|n| n.to_string()))
+                            })
+                            .or_else(|| err.as_u64().map(|n| n.to_string()));
+                        // A bare non-empty number/array error value is a signal even though it exposes no message/code field.
+                        let bare_scalar = matches!(err, serde_json::Value::Number(_))
```
houko, line 2022: Ambiguous edge case, not fixed here: `bare_scalar` only recognizes `Number` and non-empty `Array` as a real signal for a bare (non-object) `error` value.
A bare `"error": true` (or `false`) would fall through every branch — `message`/`err_type`/`err_code` are all `None` (none of the `.get()` calls apply to a `Value::Bool`), and `bare_scalar` is `false` — so the frame is silently ignored even though a boolean is arguably as much a "real signal" as a bare number.
I don't have evidence any OpenAI-compatible provider actually does this (it's speculative, unlike the empty-string case I fixed in a follow-up commit, which is a real, tested gap), so I'm leaving it as a comment rather than changing behavior.
If it's worth guarding, the fix would be extending `bare_scalar` (or renaming it) to also match `serde_json::Value::Bool(_)`.

---

houko (commented):
Automated review pass on the two-commit self-correction. Verified the content-emitted guard is applied consistently in both FallbackDriver::stream and stream_with_retry (same content-event set, same fall-through-to-generic-Err path), and that openai_stream_error_code's numeric-code handling covers 429/503/402/413/404/401/500 with the derived LlmError::Api.status matching the classified code (no blanket-502 gap for common codes).

Fixed directly and pushed (2 commits, within the 5-commit cap):
- f6d308a — rewrapped the #6512 comments in anthropic.rs, fallback.rs, and retry.rs to one-sentence-per-line (CLAUDE.md prose-wrapping rule); no logic changes.
- 0218fd3 — found and closed a real gap in the OpenAI benign-null-error check: the nested `error.message` / `error.code` / `error.type` extractions were not filtered for emptiness the way the bare `"error": ""` case already is, so `{"error":{"message":"","code":"","type":""},"choices":[...]}` would incorrectly abort a valid stream. Added `streamed_all_empty_string_error_object_is_ignored` regression test (passes), amended the CHANGELOG bullet, and rewrapped this file's remaining comments to match. `cargo check`/`clippy -D warnings`/`fmt --check` clean on both crates; full `cargo test -p librefang-llm-drivers --lib` (602 passed) and the new `agent_loop::retry` test in librefang-runtime both green.

Left one inline comment on a lower-confidence, unverified edge case (bare boolean `error` value) rather than fixing speculatively.

---

Issue #3739, opened 2026-04-27
running_tasks DashMap.insert overwrites previous task — stop_agent_run can't kill the orphaned LLM call


houko:
## Location
- `crates/librefang-kernel/src/kernel/mod.rs:6303-6309`

## Finding
```rust
if !is_fork {
    self.running_tasks.insert(
        (agent_id, effective_session_id),
        RunningTask { abort: handle.abort_handle(), ... });
}
```
`DashMap::insert` returns the old value and silently drops it. If two concurrent non-fork streaming invocations target the same `(agent_id, session_id)` key (which can happen with `session_mode = Persistent` when two callers both compute the same `effective_session_id`), the earlier entry's `abort_handle` is silently dropped.

`stop_agent_run` calls `running_tasks.remove(&(agent_id, session_id))` and aborts whichever entry is currently stored — typically the second invocation's handle. The first invocation continues running, burning tokens, with no way to stop it externally.

## When This Triggers
- `session_mode = Persistent` (default) + two concurrent calls through the streaming path that both resolve to the same session ID
- The trigger dispatcher's per-session mutex (`session_msg_locks`) serializes non-streaming calls but the streaming path only does `try_read` on `config_reload_lock`, not `agent_msg_locks` (see #3737)
- With the streaming bypass already present, a channel bridge + dashboard concurrent send can both land here simultaneously

## Impact
- User clicks "stop"; UI says stopped; LLM call continues; budget drains; no log explains why
- Impact severity depends on how often two concurrent streaming calls target the same session — uncommon in single-user single-session setups, more likely with channel bridges + dashboard open simultaneously

## Fix
- `entry(...).or_insert(...)` and return `Err("agent already running")` for the second insert, OR
- Store `Vec<AbortHandle>` and abort all in `stop_agent_run`
- Root cause fix: address #3737 (streaming bypasses `agent_msg_locks`) to prevent concurrent session access entirely

PR #6513, opened 2026-07-20
fix(kernel): release the token reservation on drop to stop a quota self-DoS

houko:
## Summary

`AgentScheduler::check_quota_and_reserve` pre-charges the agent's hourly `total_tokens` and returned a bare `u64`. The pre-charge was rolled back only by an explicit `settle_reservation` / `release_reservation` call, so when the **owning future was dropped mid-flight** the pre-charge leaked permanently:

- a follow-up message **supersedes** an in-flight streaming turn and `abort()`s the spawned task (#3739) before it reaches settle/release;
- a **`stop` / `kill`** of a running streaming turn does the same;
- a **trigger fire** wrapped in `tokio::time::timeout` is dropped when the timeout elapses, before `send_message_full_inner`'s inline settle/release runs.

Each leak inflates `total_tokens` by the estimate (`model.max_tokens`); enough of them push `total_tokens` past the limit, and `check_quota_and_reserve` then rejects **every** further turn with "Token limit would be exceeded" until the 1-hour window resets — a self-inflicted quota DoS, with no matching recorded spend for the operator to see. The USD sibling (`MeteringReservation`) already releases on drop via its own `Drop`; the token side had no such guard.

Found during a repo-wide bug audit (two separately-confirmed findings — streaming supersede/abort and trigger timeout — that share this one root cause and one fix).

## Fix

- Add `AgentScheduler::reserve_tokens`, returning a `#[must_use]` **`TokenReservation` RAII guard**. Its `Drop` releases the pre-charge unless the guard was explicitly `settle`d or `release`d, so a dropped/aborted future auto-releases — bringing the token side to parity with the USD side. No change is needed at the abort/timeout sites: the guard lives inside the dropped future.
- The settle/release math is extracted into shared free functions (`release_tokens_in` / `settle_tokens_in`) used by **both** the existing `&self` methods and the guard, so the arithmetic is unchanged and has one source of truth.
- `AgentScheduler.usage` is wrapped in `Arc` so the guard can reach the map on drop without a back-reference to the whole scheduler. All internal accesses are unchanged (DashMap is `&self`).
- A no-quota reservation carries `estimated_tokens == 0`, so its drop is a no-op — preserving the zero-reservation contract of `check_quota_and_reserve`.
- All send / streaming / trigger reservation sites in `messaging.rs` switch from `check_quota_and_reserve` + explicit `settle_reservation` / `release_reservation` to `reserve_tokens` + guard methods (2 reserve sites, 3 settle, 4 release). The existing `check_quota_and_reserve` / `settle_reservation` / `release_reservation` API and its 5 tests are **retained unchanged** as the underlying primitive, so they act as a regression net proving the math did not shift.

## Tests

Four new `scheduler` unit tests, all green:

- `dropped_reservation_releases_the_precharge` — the load-bearing regression: reserve, drop the guard **without** settle/release, assert `total_tokens` returns to 0 (this is exactly the abort/timeout path).
- `reservation_settle_records_actual_and_drop_is_noop` — settle corrects to actual + records the call, and the post-settle drop does not double-release.
- `reservation_release_rolls_back_and_drop_is_noop` — release rolls back without counting an LLM call.
- `reservation_without_quota_is_a_noop_guard` — a no-quota guard's drop does not disturb real recorded usage.

## Verification

- `cargo test -p librefang-kernel --lib` — **1305 passed, 0 failed, 1 ignored** (incl. 4 new + 5 retained reservation tests)
- `rustfmt --check` on both changed files — clean

## Related

Second PR from the same audit; the first is #6512 (llm-drivers streaming error handling). Independent change, no overlap.


houko (commented):
Automated review pass on this PR.

The core change (RAII `TokenReservation` guard, `Drop` releasing an unsettled pre-charge, `#[must_use]`, `Arc`-wrapped `usage` map) checked out cleanly: `cargo fmt --check -p librefang-kernel` and `cargo clippy -p librefang-kernel --lib -- -D warnings` are clean, and the full `cargo test -p librefang-kernel --lib` suite (1305 tests, including the four new `reservation_*` / `dropped_reservation_releases_the_precharge` regression tests) passes. The `settled` flag correctly prevents a double-release between `settle`/`release` and the subsequent `Drop`, and the no-quota (`estimated_tokens == 0`) path is a verified no-op.

Two small doc-comment nits I found while reading the diff, fixed and pushed directly (rebased cleanly on top of the two docs commits already on this branch):
- A dispatch-site comment in `messaging.rs` still said "Use check_quota_and_reserve" after this PR switched that call site to the new `reserve_tokens`.
- The new `TokenReservation`-related doc comments were hard-wrapped at ~80 columns, splitting sentences across lines; repo convention (CLAUDE.md) is one sentence per line with no fixed column wrap, so I rewrapped only the new/touched paragraphs.

One item I'm leaving for a maintainer decision rather than acting on unilaterally: two commits already on this branch (`a3069ed`, `3ac1ffa`) have `Author: Claude <noreply@anthropic.com>`. CLAUDE.md's git conventions say "No Claude / Anthropic attribution in commit messages, PR bodies, or comments" — the `commit-msg` hook checks message text but wouldn't catch attribution in the author identity itself. I did not rewrite this history (would require force-push over already-pushed commits), just flagging it here for a maintainer call on whether it needs a fixup.

---

houko (commented):
Automated daily review of #6513.

Correctness: verified `TokenReservation::settle` and `release` both consume `self` and set `settled = true` before any `Drop` could run, and `Drop` only releases when `!settled` — so no double-release path exists.
The 4 new tests in `scheduler.rs` (dropped-without-settle, settle-then-drop-is-noop, release-then-drop-is-noop, no-quota guard) match the fix's own description and exercise exactly the accounting this PR is meant to protect.

CHANGELOG: entry is present, references #6513, carries `(@houko)` attribution, and is split one sentence per line per CLAUDE.md's prose-wrapping rule.

Minor nit (not inline-commentable — outside the shown diff hunk): the comment above `token_reservation.settle(&result.total_usage);` at `crates/librefang-kernel/src/kernel/messaging.rs:3089` ("Lifecycle: emit TurnCompleted alongside settle_reservation.") still names `settle_reservation`, but that call site now goes through the `TokenReservation` guard's `settle` method, not `AgentScheduler::settle_reservation` directly. The PR's own follow-up commits (`f428278`, `a3069ed`) already fixed the same kind of stale-name drift at other comments — this one instance looks like it was missed. Not a functional bug, just worth a tidy-up pass.

Process finding, not code: five commits on this branch (`3ac1ffa`, `a3069ed`, `f428278`, `27db454`, `a9be8f0`) have their git author identity set to `Claude <noreply@anthropic.com>`. The message bodies are clean of AI-attribution text, but the author field itself carries the exact provenance signal CLAUDE.md's "No Claude / Anthropic / AI attribution" rule and the `commit-msg` hook exist to keep out of history. This reviewing session's own git identity is also configured as `Claude <noreply@anthropic.com>`, so reauthoring those commits from here would just swap one AI-attributed commit for another rather than fix it — flagging for the maintainer to reauthor/rebase before merge (or squash-merge under their own identity) instead.

---

Issue #5126, opened 2026-05-16
channel_send: owner-caller mirror path deadlocks per-agent message lock


houko:
## Problem
`SessionWriter::append_to_session` acquires `agent_msg_locks[agent_id]` via `block_in_place(|| lock.blocking_lock())`, keyed by the **channel owner**. When `channel_send` is invoked from inside A's own turn AND the target channel is owned by A, the mirror path tries to grab the same lock the outer `send_message_full(A)` is holding. Same-task deadlock — `block_in_place` does not release the held async mutex.

## Evidence
`crates/librefang-kernel/src/kernel/handles/session_writer.rs:25-40`:

```rust
fn append_to_session(&self, session_id, agent_id, message) {
    let lock = self.agents.agent_msg_locks
        .entry(agent_id)
        .or_insert_with(|| Arc::new(tokio::sync::Mutex::new(())))
        .clone();
    let _guard = tokio::task::block_in_place(|| lock.blocking_lock());
    // ...
}
```

Call chain: `send_message_full(A)` holds `agent_msg_locks[A]` (`messaging.rs:790`) → tool `channel_send` → `mirror_channel_send_to_session` → `resolve_channel_owner` returns A → `append_to_session` `blocking_lock`s the same `agent_msg_locks[A]`.

## Impact
Common production shape: an agent owns a channel (Telegram bot bound to a user) and uses `channel_send` to reply to that same channel from within its own turn. The mirror write deadlocks the worker thread silently. Multi-threaded tokio runtime parks the thread; no panic.

## Suggested fix
Detect the "agent already holds this lock" case before re-acquiring (task-local `HELD_AGENT_LOCKS: Cell<BTreeSet<AgentId>>`), and skip the mirror lock if the current task already owns it — the outer `send_message_full` will commit the session write on its own path. Or restructure mirror writes to enqueue rather than synchronously grab the lock.

Issue #5125, opened 2026-05-16
agent_send: transitive A→B→A cycle deadlocks per-agent message lock


houko:
## Problem
`tool_agent_send`'s self-send guard only catches direct cycles (`caller == agent_id`). A transitive cycle `A → B → A`, allowed by `max_agent_call_depth` (default 5), re-acquires `agent_msg_locks[A]` from the same async task. `tokio::sync::Mutex` is not reentrant — the task deadlocks against itself.

## Evidence
Self-send guard at `crates/librefang-runtime/src/tool_runner.rs:4334-4341`:

```rust
// Self-send guard: ... would attempt to acquire `agent_msg_locks[id]`
// while that lock is already held by the current turn (issue #3613).
if caller == agent_id { return Err("...".to_string()); }
```

Inner lock at `crates/librefang-kernel/src/kernel/messaging.rs:790`:

```rust
let lock = self.agents.agent_msg_locks.entry(agent_id)
    .or_insert_with(|| Arc::new(tokio::sync::Mutex::new(())))
    .clone();
let _guard = lock.lock().await;
```

`tool_agent_send` awaits `send_message_full` inside the same task via `AGENT_CALL_DEPTH.scope(...)` (`tool_runner.rs:4366-4380`).

Sequence: outer turn for A holds `agent_msg_locks[A]` → A calls `agent_send(B)` → `agent_msg_locks[B]` acquired → B calls `agent_send(A)` → blocks forever on `agent_msg_locks[A]` held by the same task.

## Impact
Any topology where A → B → A is reachable (multi-hand workflows, peer collaboration, callback agents) silently wedges. No panic, no timeout, the daemon worker thread parks until killed. Token spend stops mid-turn, no signal.

## Suggested fix
Track the lock-holding set per task (via `AGENT_CALL_DEPTH` chain or a dedicated task-local) and reject re-entrant acquisition explicitly. The error message should mention the cycle path so operators can diagnose.

Issue #4291, opened 2026-05-01
bug(kernel): forks read entry.session_id from registry — TOCTOU race with switch_agent_session breaks prompt-cache alignment


neo-wanderer:
## Symptom

A fork ("agent loop spawned to share the parent turn's context") can land on the wrong session if `entry.session_id` in the agent registry mutates between parent loop start and fork spawn. The mutation can come from `switch_agent_session` (`POST /api/agents/{id}/sessions/{session_id}/switch`), which any dashboard tab can trigger via the Sessions-page Play button or any other code path that wants to repoint an agent's "default" session.

Single-session deployments effectively never trip this. With multi-session UIs (#2959 era and after), and per-session liveness from #3172, it is much easier to trigger.

## Code

`kernel/mod.rs:6120`, in the session-resolution branch for fork loops:

```rust
_ if loop_opts.is_fork => entry.session_id,
```

The comment block above (lines ~6099-6111) is explicit about the invariant being defended:

> Fork calls always target the agent's canonical session — the whole point of fork mode is to share the parent turn's context (and therefore its prompt-cache prefix). An agent with `session_mode = "new"` would otherwise land on `SessionId::new()` here, producing a fresh empty session and breaking cache alignment. Force Persistent for forks regardless of manifest.

But "the agent's canonical session" is operationalized as `entry.session_id`, which is a registry pointer that any caller of `switch_agent_session` can flip via `registry.update_session_id(agent_id, session_id)` (`registry.rs:174`). There is no lock, no epoch, no "session id at parent-loop-start" snapshot.

So the invariant the comment claims — "fork lands on parent's session" — only holds if no one calls `switch_agent_session` during the parent turn. In a multi-session world that's an unsafe assumption.

## Reproduce (sketch)

1. Start a parent turn for agent A on session X. `entry.session_id == X`.
2. While that turn is mid-loop, `POST /api/agents/A/sessions/Y/switch` for some other session Y of A. Now `entry.session_id == Y`.
3. The parent loop emits a fork. `effective_session_id` for the fork resolves at `kernel/mod.rs:6120` → picks `entry.session_id` → Y.
4. The fork now appends to Y's history while the parent is still appending to X's. The prompt-cache prefix the fork was supposed to share is gone, and Y's history acquires messages that don't belong to its conversation.

## Why now is the right time

- #2959 (per-request `session_id` override) made it normal for an agent to have several concurrent sessions visible to users.
[...]

PR #6518, opened 2026-07-20
fix(kernel): serialize a canonical-session override on the per-agent lock to stop a lost-update race

houko:
## Summary

`send_message_full_inner` selected the dispatch serialization lock **purely on whether `session_id_override` was supplied**: `Some(sid)` → `session_msg_locks[sid]`, `None` → `agent_msg_locks[agent_id]`. The REST `/message` route forwards `req.session_id` verbatim as the override and accepts it as long as it belongs to the agent — **including when it equals the agent's canonical `entry.session_id`**.

So:
- **Task A** = REST `/message` with `session_id = <canonical>` → locks `session_msg_locks[canonical]`.
- **Task B** = a concurrent no-override persistent dispatch to the same agent (plain `send_message`, `send_message_as` without a `conversation_key`, a workflow step, or a `Persistent` trigger fire on an agent with no home channel) → locks `agent_msg_locks[agent_id]`.

Both resolve to `entry.session_id`, load it, append, and `save_session` (a blind last-writer-wins upsert). Holding **different** mutexes, they run concurrently and interleave writes to the same session history — **a lost update**. This is exactly the agent-lock-vs-session-lock hazard the cron compaction path already hand-guards (`cron_tick.rs` takes `agent_msg_locks` as an outer lock), but the general canonical-override path had no reconciliation. The streaming path (`send_message_streaming_with_sender_and_opts`) had the identical gap.

Found during a repo-wide audit and adversarially verified.

## Fix (narrow)

An override equal to the agent's canonical session collapses to the **per-agent** lock namespace, so both dispatches serialize on `agent_msg_locks[agent_id]`. Applied at both sites via a `session_lock_key` that maps `Some(canonical)` → `None` (agent lock), keeping the re-entrancy check (#5125) and held-lock registration (#5126) consistent with the lock actually taken. The raw `session_id_override` still flows into `resolve_dispatch_session_id`, so the **resolved session id is unchanged** — only the lock (and its tracking) is normalized.

### Design note (for reviewer)

A no-override **channel** dispatch keeps the per-agent lock — this is the *narrow* fix. The broader channel-derived-override variant (a REST override passing a channel-derived session id racing a channel dispatch to that session) is left unchanged; the cron-prune path already hand-guards its own instance of it. The *principled* alternative — "always lock on the resolved session id" — would close that variant too but would **bypass the `agent_msg_locks` outer lock the cron compaction path relies on**, so I deferred it rather than silently change cron's coordination. Happy to take the principled route (with the cron path taking both locks) if you prefer.

## Verification

- `rustfmt --check` — clean
- Re-entrancy / deadlock suite (`issue_5125_*`, `issue_5126_*`), `cross_task_serialization_on_agent_msg_lock_is_preserved`, and **262** messaging/session/cron/concurrency tests — pass.

A deterministic two-task race test is not feasible against this dispatch path (no injection point to force the interleaving); correctness rests on the lock-namespace unification plus the re-entrancy regression suite staying green. (The unrelated SIGABRT when running the *entire* kernel lib suite in one process is a known resource ceiling — CI shards these across 4 runners for that reason; reducing `--test-threads` runs them clean.)

## Related

Seventh PR from the same audit (#6512–#6517). Independent, kernel-only.


crates/librefang-kernel/src/kernel/messaging.rs
```diff
@@ -2795,7 +2814,19 @@ impl LibreFangKernel {
...
+        // `send_message_full_inner` site does: an override equal to
+        // `entry.session_id` writes the SAME session the no-override persistent
+        // path writes, so taking the per-session lock here while that path
+        // takes the per-agent lock would let the two write one session under
+        // two mutexes concurrently (lost update). `effective_session_id`
+        // already resolves to the canonical id in that case, so only the lock
+        // namespace changes; a no-override channel dispatch keeps the per-agent
+        // lock (narrow fix — the channel-derived variant is unchanged here).
+        let session_scoped_lock =
+            matches!(session_id_override, Some(sid) if sid != entry.session_id);
```
houko: The two "mirrored" sites aren't actually equivalent, and I think this one reintroduces a narrower version of the same hazard this PR is fixing.

At the non-streaming site (`send_message_full_inner`), the canonical id is re-read fresh right at the decision point:
```rust
let canonical_session_id = self.agents.registry.get(agent_id).map(|e| e.session_id);
```
`entry` there isn't fetched until later (line ~1130), after the lock is already held.

Here, `entry` is fetched once at line 2065 — before config/budget/quota work and the session-mode resolution block — and `entry.session_id` is compared against `session_id_override` ~760 lines later at line 2828, all inside one non-`async fn` that can still be preempted by other OS threads on the multi-threaded runtime (this function does real synchronous work: catalog lookups, budget reservation, quota checks). If `switch_agent_session` runs on another agent task in that window, `entry.session_id` is stale by the time we reach the lock decision — the exact hazard already called out in this same function at lines ~2280-2287 for the fork branch ("We read the parent session id from `loop_opts`, NOT from `entry.session_id`... mutable by `switch_agent_session`... #4291").

Concretely: if a caller passes `session_id_override == <old canonical>` and a concurrent `switch_agent_session` rotates the agent onto a new canonical session in this window, this comparison uses the stale `entry.session_id`, and could pick the per-session lock here while the no-override path (now writing the new canonical session) picks `agent_msg_locks[agent_id]` — reproducing the same two-different-mutexes-same-session race this PR closes at the other site, just gated on a rarer interleaving with `switch_agent_session`.

Suggest re-reading canonical fresh at the decision point here too (`self.agents.registry.get(agent_id).map(|e| e.session_id)`), mirroring the non-streaming site, rather than reusing the `entry` snapshot from line 2065 for this comparison specifically (it's still fine to use for the manifest/budget reads above, which don't need session-id freshness).

---

crates/librefang-kernel/src/kernel/messaging.rs
```diff
@@ -1009,7 +1009,26 @@ impl LibreFangKernel {
...
+        // `resolve_dispatch_session_id` below, so the resolved session id is
+        // unchanged; only the lock (and its re-entrancy / held-lock tracking)
+        // is normalized. (Read the canonical id once here; it is mutable via
+        // switch_agent_session, but a concurrent switch mid-dispatch is out of
+        // scope, as it already is for the resolution read below.)
+        let canonical_session_id = self.agents.registry.get(agent_id).map(|e| e.session_id);
+        let session_lock_key: Option<SessionId> = match session_id_override {
+            Some(sid) if Some(sid) != canonical_session_id => Some(sid),
+            _ => None,
+        };
```
houko, line 1055: The PR description says "a deterministic two-task race test is not feasible against this dispatch path," which I agree with for the actual concurrent-write race — but that doesn't cover this narrower, fully deterministic claim: *given `session_id_override == Some(entry.session_id)`, `session_lock_key` collapses to `None`* (and the mirror: *given a non-canonical override, it stays `Some(sid)`).

That's a pure function of `session_id_override` and the registry's current `session_id` — no interleaving required. A single-task test that spawns an agent, sends with `session_id_override = Some(entry.session_id)`, and asserts the dispatch took `agent_msg_locks[agent_id]` (e.g. via the existing `held_agent_locks` registration this PR already threads through — `agent_scoped` is observable, or check which of `session_msg_locks` / `agent_msg_locks` gained an entry) would pin down the exact behavior this PR introduces, independent of the true concurrency question. Right now the 262 passing tests cited in the PR description are regression coverage for pre-existing behavior (#5125/#5126 re-entrancy, general serialization) — none of them appear to exercise the new canonical-collapse branch itself, so a revert of the `session_lock_key` logic wouldn't be caught by the existing suite.

---

houko (commented):
Automated daily review: narrow kernel lock-namespace fix, correct in intent and scoped as described. Two findings left inline — both are judgment calls, not mechanical CLAUDE.md violations, so left as review comments rather than pushed fixes:

1. The streaming-path collapse (`send_message_streaming_with_sender_and_opts`) compares against a stale `entry.session_id` snapshot taken ~760 lines earlier, before this same function's own documented `switch_agent_session` staleness hazard (see its fork-branch comment, #4291) — unlike the non-streaming site, which re-reads canonical fresh at the decision point.
2. No test pins down the new canonical-collapse branch itself (deterministic, not requiring the true race) — the 262 passing tests cited are pre-existing regression coverage.

No CLAUDE.md mechanical violations found (formatting, HashMap/BTreeMap, Default-impl, attribution, etc. all clean); `cargo fmt -p librefang-kernel -- --check` is clean and `cargo check -p librefang-kernel --lib` was run.

---

crates/librefang-kernel/src/kernel/messaging.rs
```diff
@@ -1009,7 +1009,26 @@ impl LibreFangKernel {
...
+        // per-agent lock namespace so both serialize on the same mutex. The
+        // raw `session_id_override` is still passed to
+        // `resolve_dispatch_session_id` below, so the resolved session id is
+        // unchanged; only the lock (and its re-entrancy / held-lock tracking)
+        // is normalized. (Read the canonical id once here; it is mutable via
+        // switch_agent_session, but a concurrent switch mid-dispatch is out of
+        // scope, as it already is for the resolution read below.)
+        let canonical_session_id = self.agents.registry.get(agent_id).map(|e| e.session_id);
+        let session_lock_key: Option<SessionId> = match session_id_override {
+            Some(sid) if Some(sid) != canonical_session_id => Some(sid),
```
houko, line 1053: Test-coverage claim vs. diff: the PR description says correctness "rests on the lock-namespace unification plus the re-entrancy regression suite staying green" and lists `issue_5125_transitive_cycle_is_rejected_not_deadlocked`, `issue_5126_owner_caller_mirror_write_is_lockless_not_deadlocked`, and `cross_task_serialization_on_agent_msg_lock_is_preserved` as the coverage.

I checked those three tests (`crates/librefang-kernel/src/kernel/tests.rs:10509`, `:10566`, `:10636`) — none of them pass a `session_id_override` at all; every call site in those tests uses `None` for the override params, or (for the cross-task test) bypasses `send_message_full_inner` entirely and locks `agent_msg_locks` directly. So they exercise the *pre-existing* re-entrancy/lockless-mirror paths, not the new `session_lock_key`/`session_scoped_lock` normalization this PR adds (the `Some(sid) if Some(sid) != canonical_session_id` branch here, and its streaming twin at the `session_scoped_lock` site). Passing green doesn't tell us the collapse-to-agent-lock behavior actually happens for `Some(canonical)`.

I understand a full two-task interleaving test is hard to construct deterministically against this path (per the PR's note), but the *lock-selection* itself looks testable without any real race: spawn a task that acquires `agent_msg_locks[agent_id]` directly (as `cross_task_serialization_on_agent_msg_lock_is_preserved` already does), then call `send_message_full_inner`/`send_message_full` with `session_id_override = Some(entry.session_id)` from a second task and assert it blocks until the first releases — proving it took the *agent* lock, not a fresh `session_msg_locks[sid]` entry it would sail through concurrently. That would directly cover the bug this PR fixes, whereas the cited suite doesn't touch the new branch at all.

---

crates/librefang-kernel/src/kernel/messaging.rs
```diff
@@ -1009,7 +1009,26 @@ impl LibreFangKernel {
...
+        // write one session under two different mutexes concurrently — a lost
+        // update / corrupted history. Collapse such an override to the
+        // per-agent lock namespace so both serialize on the same mutex. The
+        // raw `session_id_override` is still passed to
+        // `resolve_dispatch_session_id` below, so the resolved session id is
+        // unchanged; only the lock (and its re-entrancy / held-lock tracking)
+        // is normalized. (Read the canonical id once here; it is mutable via
+        // switch_agent_session, but a concurrent switch mid-dispatch is out of
+        // scope, as it already is for the resolution read below.)
+        let canonical_session_id = self.agents.registry.get(agent_id).map(|e| e.session_id);
```
houko, line 1051: Minor: this `self.agents.registry.get(agent_id)` is a second lookup — `send_message_full_inner` already does another `self.agents.registry.get(agent_id).ok_or_else(...)` a bit further down (line 1130, after the lock is acquired) to get `entry` for `resolve_dispatch_session_id` / metering. Not a correctness bug — I traced it through and the per-agent lock key is `agent_id`-only (invariant to which session happens to be canonical), so a `switch_agent_session` racing between this read and the line-1130 read doesn't reopen the lost-update hazard this PR is closing; it would just mean this `canonical_session_id` is a dispatch-worth stale relative to the `entry.session_id` used later. CLAUDE.md calls out "redundant lookups" as an in-scope nit rather than follow-up material, so flagging it — though I'd rather leave the actual fix (e.g. deriving `canonical_session_id` from a single fetch) to you, since reordering the `entry` fetch in a lock-ordering-sensitive function like this one isn't something I want to do without dedicated test coverage backing the reorder.

---

houko (commented):
Automated review pass against CLAUDE.md conventions and the fix's own correctness/test-coverage claims. Pushed one mechanical formatting fix directly to the branch; left two inline judgment calls for the maintainer.

---

houko:
Good catch on the stale-snapshot race — confirmed and fixed in 6ace4aedb. The streaming site now re-reads the canonical id fresh at the decision point (`self.agents.registry.get(agent_id).map(|e| e.session_id)`), mirroring the non-streaming site and the fork branch that reads from `loop_opts` for the same #4291 reason, instead of reusing the `entry` snapshot from line ~2065. Reentrancy/serialization suite (issue_5125_*, issue_5126_*, cross_task_serialization) + clippy -D warnings stay green.

PR #6519, opened 2026-07-20
fix(memory): scope MCP knowledge_add_* writes to the calling agent, not agent_id=""

houko:
## Summary

`MemorySubstrate::add_entity` / `add_relation` hardcoded `agent_id = ""` — the write path for the MCP `knowledge_add_entity` / `knowledge_add_relation` tools. Meanwhile the proactive-extraction write path (`store_relations`) writes the same graph with the **real agent uuid**, and the agent-scoped read `GET /api/memory/agents/{id}/relations` scopes by that real uuid (`query_graph_scoped(pattern, Some(agent_id), peer)`).

So anything an agent created via its own `knowledge_add_*` tools was stored under `""` and:
- could **never** be returned by the agent-scoped relations endpoint or the dashboard relations view,
- was **never removed** by `delete_by_agent` (it orphaned on agent deletion),
- and left the graph **split-brained** — proactive triples under the real agent id, tool triples under `""`.

Found during a repo-wide audit and adversarially verified.

## Fix

The single-crate alternative (widen `query_graph_scoped` to also match `agent_id=""`) is wrong — it would surface shared rows to every agent and still leave `delete_by_agent` orphaning. The correct fix threads the caller agent id (already available at the tool dispatcher as `caller_agent_id`) through the trait contract:

- **librefang-types**: `Memory::add_entity` / `add_relation` gain an `agent_id: &str` param.
- **librefang-memory**: the substrate impl passes it to the store (was `""`).
- **librefang-kernel-handle**: `KnowledgeGraph::knowledge_add_entity` / `knowledge_add_relation` gain `agent_id: &str`.
- **librefang-kernel**: the handle impl threads it into `Memory::add_*`.
- **librefang-runtime**: the tool fns take the caller agent id and `dispatch.rs` passes `*caller_agent_id`; an absent caller id keeps the historical shared/unscoped (`""`) write.

The ~20 `KnowledgeGraph` test doubles get the matching signature (param ignored).

### Back-compat

Rows already written under `""` stay shared/unscoped — they cannot be retroactively attributed to an agent, so there is **no migration**; only new writes are agent-scoped. (These rows remain readable via an unscoped read, exactly as before.)

## Tests

- `librefang-memory` `substrate::tests::knowledge_add_scopes_the_row_to_the_caller_agent` (new): writes an entity/relation through the `Memory` trait with a real agent id and asserts that agent's scoped read returns it (was **empty** when written under `""`) while another agent's scoped read does not.

## Verification

- `cargo check --workspace --lib` and `cargo check --workspace --tests` — clean (all test doubles updated)
- `cargo test -p librefang-memory --lib` — **331 passed** (incl. the new regression)
- `cargo test -p librefang-runtime --lib knowledge` — 5 passed
- `cargo clippy -p librefang-types -p librefang-memory -p librefang-kernel-handle -p librefang-kernel -p librefang-runtime --lib -- -D warnings` — clean
- `rustfmt --check` — clean

## Scope

5-crate trait-contract change (the widest of this audit's fixes) — the `Memory` and `KnowledgeGraph` trait signatures change, rippling to their impls and ~20 test doubles. Kept mechanical; the load-bearing logic is the 6 production sites above.

## Related

Eighth and final PR from the same audit (#6512–#6518). Completes the 11 confirmed findings.


crates/librefang-runtime/src/tool_runner/knowledge.rs
```diff
@@ -96,7 +95,8 @@ pub(super) async fn tool_knowledge_add_entity(
     };
 
     let id = kh
-        .knowledge_add_entity(&entity, peer_id)
+        // Scope the row to the calling agent so agent-scoped reads / delete_by_agent see it; an absent caller id keeps the historical shared/unscoped write.
+        .knowledge_add_entity(&entity, agent_id.unwrap_or(""), peer_id)
```
houko, line 99: Judgment call, not fixing directly: `agent_id.unwrap_or("")` (same pattern at line 151 for `knowledge_add_relation`) silently falls back to the shared/unscoped `""` sentinel whenever `caller_agent_id` is `None` on the `ToolExecContext`. That's exactly the bug class this PR fixes, just gated on a different condition.

Every other tool in this module family that actually needs the caller's identity (`task_post`/`task_claim` in `task.rs`, `schedule_create`/`schedule_list` in `schedule.rs`, `cron_create`/`cron_list` in `cron.rs`, `event_publish` in `event.rs`) instead does:
```rust
let agent_id = caller_agent_id.ok_or_else(|| caller_agent_id_missing("task_post"))?;
```
surfacing `ToolError::MissingParameter("agent_id")` to the LLM rather than silently degrading.

`ToolExecContext::caller_agent_id` is `Option<&str>` and the struct doc explicitly says it's `None` for "legacy/test call sites" — so this isn't purely a test-only concern. Given the CHANGELOG text documents the `unwrap_or("")` fallback as intentional ("an absent caller id keeps the historical shared/unscoped write"), this may well be a deliberate compatibility choice rather than an oversight — but it's worth confirming explicitly: is there a real (non-test) dispatch path today where `caller_agent_id` is `None` for a `knowledge_add_*` call? If so, should these two follow the `caller_agent_id_missing` pattern instead, for consistency with the rest of the tool surface and to fully close the scoping gap rather than narrowing it?

---

crates/librefang-runtime/src/tool_runner/dispatch.rs
```diff
@@ -988,11 +988,14 @@ pub async fn execute_tool_raw(
...
-        // Knowledge graph tools. `sender_id` is the per-user peer identity
-        // (same as the memory tools use), so a multi-user agent's KG writes and
-        // reads are scoped to the calling user (#6494).
-        "knowledge_add_entity" => tool_knowledge_add_entity(input, *kernel, *sender_id).await,
-        "knowledge_add_relation" => tool_knowledge_add_relation(input, *kernel, *sender_id).await,
+        // Knowledge graph tools. `sender_id` is the per-user peer identity (same as the memory tools use), so a multi-user agent's KG writes and reads are scoped to the calling user (#6494).
+        // `caller_agent_id` additionally scopes `add_entity` / `add_relation` writes to the calling agent (#6519).
+        "knowledge_add_entity" => {
+            tool_knowledge_add_entity(input, *kernel, *caller_agent_id, *sender_id).await
+        }
```
houko, line 995: Judgment call, not fixing directly: none of the updated `KnowledgeGraph` test doubles (`DispatchCapture` in `tool_runner/tests/mod.rs`, and the ~15 other stub impls touched by this PR) assert on the `agent_id` argument's value — they all take `_agent_id: &str` and ignore it. That's fine for the stubs that just need to compile, but it means there's no dispatch-level test proving `caller_agent_id` (not `sender_id`, not a hardcoded value) actually reaches `KnowledgeGraph::knowledge_add_entity`/`knowledge_add_relation` end-to-end through `execute_tool_raw`. The new `librefang-memory` substrate test (`knowledge_add_scopes_the_row_to_the_caller_agent`) covers the storage layer well, but the dispatch wiring itself (this line and the one below) is unverified. Worth adding a small assertion in one of the `tool_runner` dispatch tests that a distinct `caller_agent_id` value is forwarded verbatim (e.g. have a capturing double record the received `agent_id` and assert it equals the context's `caller_agent_id`, not `sender_id`)?

---

houko (commented):
Automated CLAUDE.md-focused review pass. Pushed one mechanical fix commit (prose-wrap + stale comments); left two judgment-call findings as inline comments below rather than changing behavior unilaterally.

---

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project uses [Calendar Versioning](https://calver.org/) (YYYY.M.DD).
 
 ### Fixed
 
+- Scope entities/relations created by the MCP `knowledge_add_entity` / `knowledge_add_relation` tools to the calling agent instead of the empty-string sentinel, so tool-created knowledge is no longer split-brained from the rest of the graph; `MemorySubstrate::add_entity` / `add_relation` hardcoded `agent_id = ""`, while the proactive-extraction write path and the agent-scoped read (`GET /api/memory/agents/{id}/relations` → `query_graph_scoped(.., Some(agent_id), ..)`) both key on the real agent uuid — so anything an agent stored via its own `knowledge_add_*` tools could never be returned by the agent-scoped relations endpoint or the dashboard relations view, and `delete_by_agent` never removed it (it orphaned on agent deletion); the caller agent id (already available at the tool dispatcher as `caller_agent_id`) is now threaded through the `KnowledgeGraph` handle trait and the `Memory` trait into the store, and an absent caller id keeps the historical shared/unscoped (`""`) write; existing rows already written under `""` stay shared/unscoped (they cannot be retroactively attributed to an agent), while new writes are agent-scoped (#6519) (@houko)
```
houko: Judgment call, not fixing directly: the "no migration" claim here ("existing rows already written under `""` stay shared/unscoped ... new writes are agent-scoped") looks safe for the relations table, but I don't think it fully accounts for `entities`.

`KnowledgeStore::add_entity` (`crates/librefang-memory/src/knowledge.rs`) upserts on `ON CONFLICT(id, peer_id) DO UPDATE SET name = ?3, properties = ?4, updated_at = ?5` — it never updates `agent_id` on conflict. Before this PR every `knowledge_add_*` write shared `agent_id = ""`, so this was a no-op in practice. After this PR, entity ownership is first-writer-wins: if agent A calls `knowledge_add_entity` for a deterministic-id entity (e.g. a well-known org/person name) and agent B later calls `knowledge_add_entity` for the same entity, the row keeps `agent_id = "A"` — B's later relation referencing that entity by `id` still joins fine today (`query_graph_scoped`'s id-based join arm only requires `s.peer_id = r.peer_id`, not `s.agent_id = r.agent_id`), but if agent A is later deleted, `delete_by_agent("A")` removes the entity row entirely, silently orphaning agent B's still-live relations (the JOIN just stops matching, so B's data quietly vanishes from future reads — no error, no crash, just a disappearing row).

This isn't new code added by this PR (the `ON CONFLICT` clause is untouched), but the PR is what makes it reachable in practice, since before this fix every tool-written entity shared the same `""` owner and this couldn't happen. Given `knowledge.rs` isn't touched by this diff, out of scope to fix here — but is this a known/accepted tradeoff (shared-entity-by-convention, ownership doesn't matter) or worth a follow-up issue against `add_entity`'s `ON CONFLICT` handling?

---

CHANGELOG.md
houko, line 1: Process note, not a code issue: commit 5d98e41 ("docs: fix prose-wrap and stale comments in #6519 agent-scope change") is authored as `Claude <noreply@anthropic.com>`. CLAUDE.md's "No AI / Claude / Anthropic attribution in commit messages, PR bodies, or comments" rule is aimed at exactly this signal, and the existing `commit-msg` hook only scans message text, not the author identity, so it didn't catch it. Separately, the two existing review comments on this PR both carry a "_Generated by [Claude Code](https://claude.ai/code)_" footer, which is the same category of attribution the repo's convention asks reviewers/contributors to avoid in PR comments.

Not fixing either directly: rewriting the commit's author would require a rebase + force-push over already-pushed history, and I can't edit someone else's posted review comments. Flagging so a maintainer can decide whether to amend the author before merge (a force-push here would need explicit sign-off per CLAUDE.md's git-safety rules).

---

houko (commented):
Automated daily review pass. Verified: test-double signatures for the new `agent_id` param are consistent across all ~20 `KnowledgeGraph` doubles (all ignore it, which is fine — none of them back a test that specifically exercises `knowledge_add_*` scoping, so no tautological test found); the new `librefang-memory` regression test genuinely asserts agent-scoping (own-agent read sees the row, other-agent read doesn't); CHANGELOG entry has correct `(@houko)` attribution and reads as prose-wrapped one-sentence-per-line, matching the file's existing convention; commit messages are conventional. Two items left as comments below — one about a possible dangling-entity-reference edge case the "no migration" claim doesn't cover, one a process note about AI attribution already present in this PR's own commit history and review comments. The two pre-existing review threads (the `unwrap_or("")` fallback and the missing dispatch-level assertion) already cover real gaps and aren't duplicated here.

---

houko:
On the entities `ON CONFLICT` / `delete_by_agent` finding: agreed it's a real edge case, and agreed it's pre-existing and out of scope for this PR (this diff doesn't touch `knowledge.rs`'s conflict/delete SQL). Note the `entities` PRIMARY KEY is `(id, peer_id)` — not `agent_id` — so entities are inherently shared-by-id across agents; `agent_id` on an entity is just first-writer provenance, and the relation carries the load-bearing scoping. The real hazard you describe (delete_by_agent(A) removing a shared entity and orphaning B's relations) is a property of that shared-entity schema + delete semantics that predates this change; this PR only makes it reachable via tool writes. Leaving it out of scope here and filing a follow-up against `add_entity`'s ON CONFLICT / `delete_by_agent` for shared entities. The CHANGELOG 'no migration' note refers to attribution of existing `""` rows, not the conflict/delete semantics — I'll tighten the wording. (Also: the intermediate-commit `Claude` author identity is erased by squash-merge, which authors the single squashed commit as the PR author.)

PR #6528, opened 2026-07-21
fix(approval): honor auto_approve and return 409 on double-resolve (#6492)

### LINKED ISSUE #6492: Tool-approval gate broken via REST `/message` path: trusted_senders bypass, no turn-resume, queue zombies, CLI 415 (v2026.6.29 & v2026.7.11)

houko:
## Summary

Fixes the two defects from the #6492 approval-gate report that survived the earlier partial fixes (#6500 CLI 415 + status column, #6511 post-approval reply routing, #6507 docs). Both were confirmed by tracing the current code end-to-end, not just re-reading the report.

The report's Bug 2 (turn resume) and Bug 4 (CLI 415) are already fixed on `main`; Bug 3's `500 → typed 4xx` was fixed by #6441 but left non-deterministic. This PR closes the two remaining gaps.

## Bug 1 — `auto_approve = true` was a silent no-op

`ApprovalPolicy::apply_shorthands()` implements the `auto_approve` shorthand ("if set, clear `require_approval`"), but it was **only ever called from a unit test**. Both production install paths — the daemon-boot path (`ApprovalManager::new_with_db`, `boot.rs`) and the hot-reload path (`update_policy`, `config_reload_ops.rs`) — installed `config.approval` verbatim. So an operator who set `auto_approve = true` to disable gating got no effect, and the field's own doc comment ("clears the require list at boot") was false.

Fix: apply the shorthand at all three policy-install entry points (`new`, `new_with_db`, `update_policy`) by taking the policy `mut` and normalizing it before it is stored. Applying it at the install choke-points (rather than at each call site) means no future construction path can forget it. `auto_approve` now takes effect at boot and on `POST /api/config/reload`.

Not touched: the `trusted_senders` / `[[users]]` RBAC interaction (Candidate A in triage). Being listed in `trusted_senders` does not register you as a `[[users]]` entry, so with `[[users]]` configured the RBAC guest-gate still gates low-risk tools regardless of `trusted_senders`. That is a deliberate two-layer design, already documented on the approvals security page by #6507, and is out of scope for a behavior change here.

## Bug 3 — double-resolve returned a non-deterministic `400`/`404`

`ApprovalManager::resolve`'s "not pending" branch scanned only the in-memory `recent` ring (last 100 resolutions) to tell "already resolved" from "never existed". So re-resolving a terminal approval returned `400` (from "Already {decision} by {who}") while the entry was still buffered, and `404` (from "not found or expired") once it aged out or the daemon restarted — the same request, two different statuses depending on load and uptime.

Fix: `resolve` now falls back to the durable `approval_audit` log (the authoritative terminal-state record; every resolution writes a `decision != "pending"` row) when `recent` no longer holds the entry, so it deterministically recognizes an already-resolved id. The api boundary maps the "Already …" verdict to a new typed `LibreFangError::Conflict` → **`409 Conflict`, code `conflict`** — a stable, well-typed answer a client can act on. A genuinely-unknown id still returns `404`; a missing second factor still returns `400`.

`LibreFangError` is `#[non_exhaustive]`, so adding the `Conflict` variant does not break any external match (they already carry a `_` arm). `ErrorCode::Conflict` already existed; this completes the pair.

## Changes

- `librefang-types`: new `LibreFangError::Conflict(String)` variant; corrected `ApprovalPolicy::auto_approve` doc comment.
- `librefang-api`: `kernel_op_status` maps `Conflict → 409`, `kernel_op_error_code` maps `Conflict → ErrorCode::Conflict`; `approve`/`reject`/`modify` utoipa responses gain `404` + `409`; `openapi.json` regenerated.
- `librefang-kernel`: `apply_shorthands()` applied in `new` / `new_with_db` / `update_policy`; `resolve` consults the audit log via a new `db_lookup_resolved_decision` helper; `approval_gate.rs` maps `Already …` → `Conflict` (was `InvalidInput`), keeping `not found → AgentNotFound` and `TOTP required → InvalidInput`.

## Verification

- `cargo test -p librefang-kernel --lib` — new unit tests pass: `auto_approve_shorthand_applied_at_construction`, `..._at_construction_with_db`, `..._on_hot_reload`, `double_resolve_reports_already_resolved`, `double_resolve_stays_a_conflict_after_recent_eviction`, `resolve_genuinely_unknown_id_is_not_found` (6/6).
- `cargo test -p librefang-api --test approvals_routes_integration` — 23 passed, including new `double_approve_is_conflict` (asserts `409` + `code=conflict`).
- `cargo test -p librefang-api --test openapi_spec_test` — regenerates `openapi.json` (new `404`/`409` responses).
- `cargo check -p librefang-types -p librefang-kernel -p librefang-api --lib` — clean.
- `cargo clippy -p librefang-types -p librefang-kernel -p librefang-api --all-targets -- -D warnings` — clean.

Closes #6492.


crates/librefang-kernel/src/approval.rs
```diff
@@ -1081,14 +1107,19 @@ impl ApprovalManager {
                 Ok((response, pending.deferred))
             }
             None => {
```
houko, line 1109: Nice fix for the sequential double-resolve case, but there's still a narrow TOCTOU window for a genuinely *concurrent* double-resolve.

`self.pending.remove(&request_id)` (the atomic DashMap removal a few lines up) is what decides the winner, but the winner doesn't write to `recent` / the audit log until `push_recent(...)` runs — after TOTP-grace recording, session-approval caching, and response construction. If a second caller's `resolve()` enters this `None` branch in that window (winner already removed the entry, but hasn't reached `push_recent` yet), neither `recent` nor `approval_audit` has the row yet, and this falls through to "not found" (404) instead of "already resolved" (409) — for what is actually a legitimate double-resolve, just delayed by a few instructions rather than a full request round-trip.

This isn't introduced by this PR (the same window existed pre-#6492 against the `recent`-ring-only lookup) and it's a tight window, but since the PR's whole point is making double-resolve *deterministically* 409, it seems worth at least a comment noting the residual race, or moving the `push_recent` call (or just the audit-log write) to immediately after the `remove()` succeeds, before the other side effects. Happy to be told this is out of scope / acceptable given how narrow the window is — flagging for a maintainer call rather than changing the lock/ordering myself.

---

crates/librefang-kernel/src/approval.rs
```diff
@@ -1350,7 +1381,9 @@ impl ApprovalManager {
     }
 
     /// Update the approval policy (for hot-reload).
-    pub fn update_policy(&self, policy: ApprovalPolicy) {
+    pub fn update_policy(&self, mut policy: ApprovalPolicy) {
```
houko, line 1384: Per repo CLAUDE.md ("MANDATORY: Integration Testing"), any route/wiring change needs a `#[tokio::test]` against `TestServer` in `crates/librefang-api/tests/`, since that's what catches "config field parses but never actually wires up" bugs like Bug 1 itself. The PR does add solid kernel-unit coverage for the shorthand (`auto_approve_shorthand_applied_on_hot_reload` etc.), but I don't see a test in `crates/librefang-api/tests/config_routes_integration.rs` (or elsewhere) that goes through the actual `POST /api/config/reload` HTTP route with an `[approval] auto_approve = true` config and asserts the effect (e.g. that a subsequently-gated tool no longer requires approval, or that `GET /api/config` reflects the cleared `require_approval` list post-reload).

I didn't add one myself: `config_routes_integration.rs`'s existing reload tests write a `config.toml` to a tempdir and hit the route, but exercising "a tool call that used to be gated now isn't" would need to go through the approval-gate/tool-dispatch path too, and I wasn't confident I could wire that up correctly without risking a flaky or misleading test. Flagging this as a judgment call — is kernel-level coverage sufficient here since no route *shape* changed (only existing wiring behavior), or does the mandatory-integration-test rule still want an HTTP-level regression test for this specific bug?

---

crates/librefang-types/src/approval.rs
```diff
@@ -670,7 +670,8 @@ pub struct ApprovalPolicy {
     pub timeout_secs: u64,
     /// Auto-approve in autonomous mode. Default: `false`.
     pub auto_approve_autonomous: bool,
-    /// If `auto_approve = true`, clears the require list at boot.
+    /// When `true`, `apply_shorthands` clears `require_approval` so no tool is gated.
+    /// Applied whenever the policy is installed into an `ApprovalManager` (boot, hot-reload, construction) — see `#6492`.
     #[serde(default)]
     pub auto_approve: bool,
```
houko, line 676: `ApprovalPolicy` is nested under `KernelConfig.approval` (`crates/librefang-types/src/config/types.rs`), so this field's doc comment feeds `schemars`' generated `description` for this field in the `KernelConfig` JSON Schema — and that schema is snapshotted in `crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json`, which still has the *old* wording at its `auto_approve` entry ("If `auto_approve = true`, clears the require list at boot."). Per the schema-mirror invariant in `crates/librefang-types/CLAUDE.md` ("Re-run the kernel-config golden in `librefang-api` (CI will fail otherwise)"), a doc-comment content change on a `KernelConfig`-reachable field should regenerate that fixture.

In practice `kernel_config_schema_matches_golden_fixture` is currently `#[ignore]`'d for unrelated pre-existing drift (tracked under #4084), so this won't fail CI today — but it's additional drift on top of that, and whoever regenerates the fixture for #4084 should know this field's description also needs to be picked up. I didn't attempt to hand-patch the JSON myself since I can't run the regeneration command in this environment (no matching rustc toolchain / no docker daemon available) to confirm the exact generated string, and didn't want to guess at schemars' output.

---

houko (commented):
Daily automated review pass against CLAUDE.md conventions plus a correctness sanity-check of the approval double-resolve race fix.

Pushed one mechanical fix (comment reflow only, no behavior change) and left three inline comments on items that need a maintainer judgment call rather than a blind fix.

---

crates/librefang-kernel/src/approval.rs
```diff
@@ -112,7 +112,10 @@ impl ApprovalManager {
             .count()
     }
 
-    pub fn new(policy: ApprovalPolicy) -> Self {
+    pub fn new(mut policy: ApprovalPolicy) -> Self {
```
houko, line 115: Per CLAUDE.md's "MANDATORY: Integration Testing" section, a route/wiring behavior change like this needs a `#[tokio::test]` against `TestServer` that hits the endpoint and asserts the side effect, not just a kernel-level unit test.

Bug 3 (double-resolve → 409) gets that: `double_approve_is_conflict` in `crates/librefang-api/tests/approvals_routes_integration.rs` drives the real HTTP route twice and asserts `CONFLICT` + `code: "conflict"`.

Bug 1 (`auto_approve` shorthand) only has kernel-level unit tests here (`auto_approve_shorthand_applied_at_construction`, `..._with_db`, `..._on_hot_reload`). I don't see an equivalent API-level assertion that a client-visible effect actually changed — e.g. that a tool call which would otherwise create a pending approval no longer does once `auto_approve` is set, or that `POST /api/config/reload` with `auto_approve = true` observably stops gating.

I looked at whether this is straightforward to add with the existing harness and don't think it is: `approvals_routes_integration.rs`'s `boot()`/`seed_pending()` seed pending requests by calling `ApprovalManager::request_approval` directly, which bypasses the `requires_approval()` gate entirely (it always creates a pending entry regardless of policy) — and `routes/approvals.rs`'s own `create_approval` handler does the same (policy is only consulted for `timeout_secs`, not for gating). So there's no existing HTTP-reachable seam that would exercise "does `auto_approve` actually stop a tool from being gated" — that check happens upstream in the tool-execution/agent-loop path, which this test file's harness doesn't drive.

Given that, I'm not confident enough to write that test myself without either (a) asserting something that doesn't actually exercise the fixed code path, or (b) building new test scaffolding for the tool-execution path, which feels like a bigger design decision than this PR's scope. Flagging per CLAUDE.md's "Reviewers gate PRs on the presence of an integration test" rather than guessing — could a maintainer confirm the intended seam (e.g. exposing `requires_approval` through a small test-only hook, or is the kernel-level coverage considered sufficient for this one since it's a construction-time normalization rather than a new route)?

---

houko (commented):
Reviewed the approval-gate fix (auto_approve shorthand + deterministic 409 on double-resolve) against CLAUDE.md, focused on the mandatory integration-test rule for route/wiring changes. The 409-on-double-resolve case is well covered by a real TestServer test. Left one inline comment on a gap I didn't think was safe to fix myself: no API-level integration test for the auto_approve-honored case, and the existing test harness doesn't have an obvious seam to add one without either testing the wrong thing or building new scaffolding — asking for maintainer input rather than guessing. No other findings; auth allowlist, session_mode/concurrency, prose wrapping, commit format, and AI-attribution checks were all clean on the touched files.

---

Issue #6530, opened 2026-07-21
refactor(uploads): normalize on-disk upload naming to <uuid>.<ext> across the shared download dir


neo-wanderer:
### Summary

The shared upload/download temp directory (`channels.effective_file_download_dir()`, default `<temp>/librefang_uploads`) is written by several independent producers using **three different naming schemes**:

| Producer | On-disk name |
|---|---|
| Channel bridge (Telegram/Matrix) | `<uuid>.<ext>` (`crates/librefang-channels/src/bridge.rs`) |
| `image_generate` | `image_<uuid>.png` (`crates/librefang-runtime/src/tool_runner/media.rs`) |
| Dashboard/API upload, browser downloads, media API, base64-inline | bare `<uuid>` (`uploads.rs`, `browser_tools.rs`, `routes/media.rs`, `sessions.rs`) |

Consumers that reconstruct the path via `upload_dir.join(&file_id)` therefore only work for the bare-`<uuid>` producers.

### Why it matters: MIME/extension is load-bearing once an upload is persisted

The bare-`<uuid>` name discards the file's type at rest, which breaks any flow that **persists or re-dispatches** the uploaded bytes rather than just inlining them:

- **Extension-sniffing consumers** infer type from the path — `media_transcribe`, image readers, and any tool handed the raw path will mis-handle a bare-`<uuid>` file (no `.pdf`/`.png`/`.oga` to key on).
- **Persisting the attachment elsewhere** (copying it into a store/vault, archiving it, re-serving it) needs a correct extension so the persisted artifact keeps its type; a bare-`<uuid>` forces every persist site to separately thread the MIME through, and any that doesn't silently writes a type-less blob.
- `serve_upload` already has to *guess* (`image/png` fallback) for not-in-registry files precisely because the on-disk name carries no type.

This is the concrete motivation for parity: #6529 surfaces the *path*, but the path is only fully useful for persistence when it also carries the type via `<uuid>.<ext>`.

### Constraint

The `file_id` returned to clients and stored in `UPLOAD_REGISTRY` must stay a bare UUID — the path-traversal guard (`resolve_attachments`, `serve_upload`) and the #3361 owner-access check both do `uuid::Uuid::parse_str(&file_id)`. So the on-disk name diverges from `file_id`, and every reader must reconstruct the extension (deterministically from content-type, or from a basename stored in `UploadMeta`).

### Touchpoints to fix

Write sites (produce files in the shared dir):
[...]

PR #6531, opened 2026-07-21
fix(attachments): surface on-disk upload path to agents for every file type

### LINKED ISSUE #6529: bug(attachments): dashboard/API uploads don't surface the on-disk file path to the agent
### Summary

File uploads made through the **dashboard / HTTP API** (`POST /api/agents/{id}/upload`) reach the LLM **without the on-disk path** of the saved file. Uploads made through a **channel** (Telegram/Matrix) do surface it, via a `[File: <name>] saved to <path>` block. So an agent asked to act on an uploaded file (attach it somewhere, hand it to a file-reader/transcribe tool) has no way to locate the bytes after a dashboard upload — it has to list the upload directory and guess the file by size.

For **images on a non-vision model** it's worse: the image block is redacted to a placeholder that carries neither pixels nor path, so the agent is left with no reference to the file at all.

### Root cause

`resolve_attachments` (`crates/librefang-api/src/routes/agents/attachments.rs`) builds a content block per attachment:

- `image/*` → `ContentBlock::Image` (base64) — no path
- `application/pdf` → `[Attached PDF: <name>]` + extracted text — no path
- text-like → `[Attached file: <name>]` + body — no path
- everything else → skipped entirely

It reads the file from `upload_dir.join(&file_id)` (so it *has* the path) but never emits it. The channel bridge, by contrast, always pushes a `[File: <name>] saved to <path>` block after enrichment (`crates/librefang-channels/src/bridge.rs`, `FILE_SAVED_BLOCK_PREFIX`, #4448). The dashboard path — the one the channel enrichment module was even modeled on — is the one missing it.

Compounding it for images: `redact_images_for_text_only` (`crates/librefang-runtime/src/agent_loop/mod.rs`, #6010) and `MessageContent::strip_images` (`crates/librefang-types/src/message.rs`) both replace `Image`/`ImageFile` blocks with a text placeholder that drops the `ImageFile.path`, so even the channel path loses an image's location on a text-only model / after history trimming.

### Repro (synthetic)

1. Configure an agent whose model has **no vision support** (e.g. any text-only OpenAI-compatible model).
2. From the dashboard chat, upload a PDF (or any file) and ask the agent to save/attach it somewhere that needs a filesystem path.
3. The agent's prompt contains the extracted text / an `[image omitted: … no vision support]` placeholder, but **no path** — it cannot locate `<upload_dir>/<file_id>` without listing the directory and size-matching.

### Proposed fix

1. In `resolve_attachments`, append a standalone `[File: <name>] saved to <path>` **Text** block for **every** attachment type (image / PDF / text / binary). A standalone Text block (not folded into the image block) survives both image-redaction passes.
2. Make `redact_images_for_text_only` and `strip_images` keep the on-disk path in the placeholder when the block is an `ImageFile`.

Note this is a **breaking change** to the attachment message shape: agents will receive an extra content block per upload.

Out of scope (separate issue): the on-disk upload naming is inconsistent across producers (channel bridge writes `<uuid>.<ext>`, dashboard/browser/media write bare `<uuid>`, `image_generate` writes `image_<uuid>.png`); normalizing that shared temp namespace is a cross-cutting change tracked separately.

neo-wanderer:
Fixes #6529.

## Problem

Dashboard/API uploads (`POST /api/agents/{id}/upload`) reached the LLM **without the on-disk path** of the saved file, while channel (Telegram/Matrix) uploads surface it via a `[File: <name>] saved to <path>` block. An agent asked to act on a dashboard-uploaded file had to list the upload dir and guess by size. For an **image on a non-vision model** the redaction placeholder dropped both pixels and path, leaving no reference at all.

## Changes

- **`resolve_attachments`** (`librefang-api/.../attachments.rs`): append a standalone `[File: <name>] saved to <path>` text block for **every** resolved attachment type — image, PDF, text, and previously-skipped binary/unknown types (now surfaced with a note + path instead of being dropped). Mirrors the channel bridge's `FILE_SAVED_BLOCK_PREFIX`. New private helper `file_saved_block`. Kept as its own `Text` block so it survives image redaction.
- **`redact_images_for_text_only`** (`librefang-runtime/.../agent_loop/mod.rs`, no-vision path): keep the on-disk path in the placeholder when the block is an `ImageFile`. Base64 `Image` (no path) unchanged.
- **`MessageContent::strip_images`** (`librefang-types/.../message.rs`, history trimming): same path preservation for `ImageFile`.

## Breaking change

Agents now receive an additional `[File: ...] saved to ...` content block for every uploaded attachment. Downstream code asserting on exact attachment message content will observe the new block.

## Tests

- `attachments::tests::file_saved_block_has_expected_format` — pins the `[File: … saved to …]` format / prefix parity.
- `message::tests::test_image_file_strip_images` (extended) + `test_strip_images_base64_has_no_path_hint` — ImageFile keeps path; base64 Image fabricates none.
- `agent_loop::tests::utilities::redact_images_for_text_only_preserves_imagefile_path` — ImageFile keeps path under no-vision redaction; base64 Image does not.

## Verification

- `cargo clippy -p librefang-types -p librefang-api -p librefang-runtime --all-targets -- -D warnings` — clean.
- `cargo test -p librefang-types strip_images` (6 passed), `cargo test -p librefang-runtime redact_images_for_text_only` (3 passed), `cargo test -p librefang-api file_saved_block` (1 passed).

## Out of scope

On-disk upload naming is inconsistent across producers (`<uuid>.<ext>` vs bare `<uuid>` vs `image_<uuid>.png`). Normalizing that shared temp namespace is a cross-cutting change tracked separately in #6530.

### DIFF
--- a/crates/librefang-api/src/routes/agents/attachments.rs
+++ b/crates/librefang-api/src/routes/agents/attachments.rs
@@ -78,6 +78,33 @@ fn is_text_like_attachment(content_type: &str, filename: &str) -> bool {
     )
 }
 
+/// Build the `[File: <name>] saved to <path>` marker block that tells the
+/// agent where the raw upload bytes live on disk.
+///
+/// Emitted for EVERY resolved attachment type (image / PDF / text / other),
+/// so the on-disk path is available to the agent regardless of whether the
+/// content was also inlined. This mirrors the channel bridge's path block
+/// (`librefang-channels` `bridge.rs`, `FILE_SAVED_BLOCK_PREFIX`, #4448) so
+/// web/dashboard uploads reach the LLM with the same path affordance channel
+/// (Telegram/Matrix) uploads already have — previously the dashboard path
+/// omitted it, forcing agents to guess the file by listing the upload dir and
+/// size-matching.
+///
+/// Kept as its OWN `Text` block — never folded into the `Image` block — so it
+/// survives the two image-redaction passes (`redact_images_for_text_only` for
+/// no-vision models and `strip_images` for history trimming), both of which
+/// replace only `Image`/`ImageFile` blocks and leave `Text` untouched. That is
+/// what lets a non-vision model still learn where an uploaded image lives.
+fn file_saved_block(
+    filename: &str,
+    path: &std::path::Path,
+) -> librefang_types::message::ContentBlock {
+    librefang_types::message::ContentBlock::Text {
+        text: format!("[File: {}] saved to {}", filename, path.display()),
+        provider_metadata: None,
+    }
+}
+
 /// Resolve uploaded file attachments into content blocks.
 ///
 /// Reads each file from the upload directory and produces blocks the
@@ -91,7 +118,13 @@ fn is_text_like_attachment(content_type: &str, filename: &str) -> bool {
 ///     plus common code/data extensions) → `ContentBlock::Text` with a
 ///     `[Attached file: <filename>]` header. Read as UTF-8 lossy and
 ///     truncated at 200K chars.
-///   - everything else → skipped with a warn log.
+///   - everything else → a bare note that the file exists but its type is not
+///     inlined (the path block below still points the agent at the bytes).
+///
+/// In ALL cases a `[File: <name>] saved to <path>` block (see
+/// [`file_saved_block`]) is appended so the agent always knows the on-disk
+/// path of the raw upload — this is the parity fix that lets the agent attach
+/// the file to a vault / transcribe it / read it, for every file type.
 pub fn resolve_attachments(
     state: &AppState,
     attachments: &[AttachmentRef],
@@ -144,6 +177,10 @@ pub fn resolve_attachments(
                         media_type: content_type,
                         data: b64,
                     });
+                    // Sibling path block: survives no-vision redaction /
+                    // history image-stripping so the agent can still locate
+                    // and act on the raw image file.
+                    blocks.push(file_saved_block(&filename, &file_path));
                 }
                 Err(e) => {
                     tracing::warn!(file_id = %att.file_id, error = %e, "Failed to read image upload");
@@ -176,6 +213,7 @@ pub fn resolve_attachments(
                         text: format!("{header}\n\n{body}"),
                         provider_metadata: None,
                     });
+                    blocks.push(file_saved_block(&filename, &file_path));
                 }
                 Err(e) => {
                     tracing::warn!(file_id = %att.file_id, error = %e, "Failed to read PDF upload");
@@ -213,18 +251,40 @@ pub fn resolve_attachments(
                         text: format!("{header}\n\n{body}"),
                         provider_metadata: None,
                     });
+                    blocks.push(file_saved_block(&filename, &file_path));
                 }
                 Err(e) => {
                     tracing::warn!(file_id = %att.file_id, error = %e, "Failed to read text upload");
                 }
             }
         } else {
-            tracing::warn!(
-                file_id = %att.file_id,
-                content_type = %content_type,
-                filename = %filename,
-                "Attachment type not yet wired into the agent loop; skipping"
-            );
+            // Unsupported/binary type: we can't inline its content, but the
+            // file IS on disk and the agent may still want to act on it (attach
+            // to a vault, hand to a file-reader tool). Surface a note + the
+            // path block instead of silently dropping the attachment. Guarded
+            // on existence so we never advertise a path to a missing file.
+            if file_path.exists() {
+                tracing::info!(
+                    file_id = %att.file_id,
+                    content_type = %content_type,
+                    filename = %filename,
+                    "Attachment type not inlined; surfacing on-disk path only"
+                );
+                blocks.push(librefang_types::message::ContentBlock::Text {
+                    text: format!(
+                        "[Attached file: {filename} ({content_type})] — content not inlined for this type; the raw file is available on disk."
+                    ),
+                    provider_metadata: None,
+                });
+                blocks.push(file_saved_block(&filename, &file_path));
+            } else {
+                tracing::warn!(
+                    file_id = %att.file_id,
+                    content_type = %content_type,
+                    filename = %filename,
+                    "Attachment type not wired into the agent loop and file missing on disk; skipping"
+                );
+            }
         }
     }
 
@@ -423,6 +483,29 @@ mod tests {
     /// Regression: the streaming attachment fetch must bound peak memory by
     /// refusing a chunk that would exceed the cap (previously `resp.bytes()`
     /// buffered the whole body first, enabling remote OOM).
+    /// The `[File: <name>] saved to <path>` marker is the contract downstream
+    /// agents parse to locate an uploaded file on disk (and it must match the
+    /// channel bridge's `FILE_SAVED_BLOCK_PREFIX` format). Pin the exact shape.
+    #[test]
+    fn file_saved_block_has_expected_format() {
+        let block = file_saved_block(
+            "H1 report.pdf",
+            std::path::Path::new("/tmp/librefang_uploads/9ff3f16a"),
+        );
+        match block {
+            librefang_types::message::ContentBlock::Text { text, .. } => {
+                assert_eq!(
+                    text,
+                    "[File: H1 report.pdf] saved to /tmp/librefang_uploads/9ff3f16a"
+                );
+                // Prefix parity with librefang-channels bridge.rs.
+                assert!(text.starts_with("[File: "));
+                assert!(text.contains("] saved to "));
+            }
+            other => panic!("expected Text block, got {other:?}"),
+        }
+    }
+
     #[test]
     fn push_within_cap_rejects_over_cap_chunk() {
         let mut buf = Vec::new();
--- a/crates/librefang-api/src/routes/agents/attachments.rs
+++ b/crates/librefang-api/src/routes/agents/attachments.rs
@@ -78,23 +78,13 @@ fn is_text_like_attachment(content_type: &str, filename: &str) -> bool {
     )
 }
 
-/// Build the `[File: <name>] saved to <path>` marker block that tells the
-/// agent where the raw upload bytes live on disk.
+/// Build the `[File: <name>] saved to <path>` marker block that tells the agent where the raw upload bytes live on disk.
 ///
-/// Emitted for EVERY resolved attachment type (image / PDF / text / other),
-/// so the on-disk path is available to the agent regardless of whether the
-/// content was also inlined. This mirrors the channel bridge's path block
-/// (`librefang-channels` `bridge.rs`, `FILE_SAVED_BLOCK_PREFIX`, #4448) so
-/// web/dashboard uploads reach the LLM with the same path affordance channel
-/// (Telegram/Matrix) uploads already have — previously the dashboard path
-/// omitted it, forcing agents to guess the file by listing the upload dir and
-/// size-matching.
+/// Emitted for EVERY resolved attachment type (image / PDF / text / other), so the on-disk path is available to the agent regardless of whether the content was also inlined.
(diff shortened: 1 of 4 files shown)


crates/librefang-api/src/routes/agents/attachments.rs
```diff
@@ -423,6 +483,29 @@ mod tests {
     /// Regression: the streaming attachment fetch must bound peak memory by
     /// refusing a chunk that would exceed the cap (previously `resp.bytes()`
     /// buffered the whole body first, enabling remote OOM).
+    /// The `[File: <name>] saved to <path>` marker is the contract downstream
+    /// agents parse to locate an uploaded file on disk (and it must match the
+    /// channel bridge's `FILE_SAVED_BLOCK_PREFIX` format). Pin the exact shape.
+    #[test]
+    fn file_saved_block_has_expected_format() {
```
houko: **Bug: doc comment got attached to the wrong test.**

The new `file_saved_block_has_expected_format` test was inserted directly under the pre-existing doc comment for `push_within_cap_rejects_over_cap_chunk` (the "streaming attachment fetch must bound peak memory..." regression note), so the two merged into one comment on this function — and `push_within_cap_rejects_over_cap_chunk` further down (currently line 509) lost its doc comment entirely. Please split them back apart:

```suggestion
    /// The `[File: <name>] saved to <path>` marker is the contract downstream agents parse to locate an uploaded file on disk.
    /// It must match the channel bridge's `FILE_SAVED_BLOCK_PREFIX` format. Pin the exact shape.
    #[test]
    fn file_saved_block_has_expected_format() {
```

...and restore the orphaned comment at the `push_within_cap_rejects_over_cap_chunk` `#[test]` (currently line 509):

```
    /// Regression: the streaming attachment fetch must bound peak memory by refusing a chunk that would exceed the cap.
    /// Previously `resp.bytes()` buffered the whole body first, enabling remote OOM.
    #[test]
    fn push_within_cap_rejects_over_cap_chunk() {
```

(Note: I have this exact fix committed locally in a review worktree, but this session's egress policy blocks pushing to `neo-wanderer/librefang`, so I can't land it directly — leaving it here for you to apply.)

---

crates/librefang-api/src/routes/agents/attachments.rs
```diff
@@ -78,6 +78,33 @@ fn is_text_like_attachment(content_type: &str, filename: &str) -> bool {
...
+/// web/dashboard uploads reach the LLM with the same path affordance channel
+/// (Telegram/Matrix) uploads already have — previously the dashboard path
+/// omitted it, forcing agents to guess the file by listing the upload dir and
+/// size-matching.
+///
+/// Kept as its OWN `Text` block — never folded into the `Image` block — so it
+/// survives the two image-redaction passes (`redact_images_for_text_only` for
+/// no-vision models and `strip_images` for history trimming), both of which
+/// replace only `Image`/`ImageFile` blocks and leave `Text` untouched. That is
+/// what lets a non-vision model still learn where an uploaded image lives.
```
houko: **Prose wrapping: one sentence per line (CLAUDE.md).**

CLAUDE.md's prose-wrapping rule applies to doc-comments and any multi-line free-form comment block: "The only legitimate line break inside a paragraph is at a complete sentence boundary... One sentence = one line, regardless of length." This new doc comment (and its sibling code comments) are hard-wrapped at ~80 chars, splitting sentences mid-clause. Since this is new prose the rule applies (it's not grandfathered pre-existing text). Suggested fix for this block:

```suggestion
/// Build the `[File: <name>] saved to <path>` marker block that tells the agent where the raw upload bytes live on disk.
///
/// Emitted for EVERY resolved attachment type (image / PDF / text / other), so the on-disk path is available to the agent regardless of whether the content was also inlined.
/// This mirrors the channel bridge's path block (`librefang-channels` `bridge.rs`, `FILE_SAVED_BLOCK_PREFIX`, #4448) so web/dashboard uploads reach the LLM with the same path affordance channel (Telegram/Matrix) uploads already have — previously the dashboard path omitted it, forcing agents to guess the file by listing the upload dir and size-matching.
///
/// Kept as its OWN `Text` block — never folded into the `Image` block — so it survives the two image-redaction passes (`redact_images_for_text_only` for no-vision models and `strip_images` for history trimming), both of which replace only `Image`/`ImageFile` blocks and leave `Text` untouched.
/// That is what lets a non-vision model still learn where an uploaded image lives.
```

The same mid-sentence wrapping recurs at a few more spots this PR touches — please apply the same one-sentence-per-line reflow there too: the `resolve_attachments` doc addendum (currently ~L121-127), the "Sibling path block" comment in the image branch (~L180-182), and the "Unsupported/binary type" comment in the else branch (~L261-265).

---

crates/librefang-runtime/src/agent_loop/mod.rs
```diff
@@ -348,19 +348,30 @@ fn build_sender_prefix(manifest: &AgentManifest, sender_user_id: Option<&str>) -
 /// user sent a photo. Redacting upstream of the driver covers *every* entry
 /// path (WebUI, channels, sidecars, triggers), not just the OpenAI driver.
 ///
+/// For `ImageFile` blocks (which reference the image by on-disk path) the
+/// placeholder keeps that path, so a text-only agent can still read or attach
+/// the raw file even though it can't see the pixels. The inline base64
+/// `Image` variant has no path to keep.
```
houko: **Prose wrapping: one sentence per line (CLAUDE.md).** Same mid-sentence hard-wrap as noted on `attachments.rs` — this new doc comment should be:

```suggestion
/// For `ImageFile` blocks (which reference the image by on-disk path) the placeholder keeps that path, so a text-only agent can still read or attach the raw file even though it can't see the pixels.
/// The inline base64 `Image` variant has no path to keep.
```

---

crates/librefang-types/src/message.rs
```diff
@@ -252,23 +252,40 @@ impl MessageContent {
             MessageContent::Blocks(blocks) => {
                 let mut stripped = false;
                 for block in blocks.iter_mut() {
+                    // Capture the media type and, for on-disk `ImageFile`
+                    // blocks, the path — so the placeholder can point a later
+                    // turn at the raw bytes even after the pixels are dropped.
+                    // The inline base64 `Image` variant carries no path.
```
houko: **Prose wrapping: one sentence per line (CLAUDE.md).**

```suggestion
                    // Capture the media type and, for on-disk `ImageFile` blocks, the path — so the placeholder can point a later turn at the raw bytes even after the pixels are dropped.
                    // The inline base64 `Image` variant carries no path.
```

Same issue recurs a few lines down at "When the source was an `ImageFile`..." (currently ~L275-278) — please reflow that one too.

---

crates/librefang-runtime/src/agent_loop/tests/utilities.rs
```diff
@@ -2153,6 +2153,63 @@ fn redact_images_for_text_only_replaces_image_and_imagefile_blocks() {
...
+#[test]
+fn redact_images_for_text_only_preserves_imagefile_path() {
+    use super::super::redact_images_for_text_only;
+    use librefang_types::message::{ContentBlock, Message, MessageContent, Role};
+
+    let messages = vec![Message {
+        role: Role::User,
+        content: MessageContent::Blocks(vec![
+            // On-disk image: path must survive into the placeholder so a
+            // text-only agent can still read/attach the raw file.
```
houko: **Prose wrapping: one sentence per line (CLAUDE.md).**

```suggestion
            // On-disk image: path must survive into the placeholder so a text-only agent can still read/attach the raw file.
```

---

crates/librefang-api/src/routes/agents/attachments.rs
```diff
@@ -91,7 +118,13 @@ fn is_text_like_attachment(content_type: &str, filename: &str) -> bool {
...
 ///     truncated at 200K chars.
-///   - everything else → skipped with a warn log.
+///   - everything else → a bare note that the file exists but its type is not
+///     inlined (the path block below still points the agent at the bytes).
+///
+/// In ALL cases a `[File: <name>] saved to <path>` block (see
+/// [`file_saved_block`]) is appended so the agent always knows the on-disk
+/// path of the raw upload — this is the parity fix that lets the agent attach
+/// the file to a vault / transcribe it / read it, for every file type.
 pub fn resolve_attachments(
```
houko, line 115: **Possible gap: no `#[tokio::test]` against `TestServer` for this behavior change.**

Per CLAUDE.md's "MANDATORY: Integration Testing" section, a route/wiring change should get a `#[tokio::test]` in `crates/librefang-api/tests/` hitting the real router, not just a unit test on the helper function. `resolve_attachments` feeds directly into what `POST /api/agents/{id}/upload` injects into the agent's session, and this PR changes that content for every attachment type (the "Breaking change" called out in the PR description). The only existing upload-route integration test (`test_upload_over_limit_returns_413` in `api_integration_test.rs`) only exercises the 413-oversize path — nothing exercises the happy path end-to-end and asserts the injected session picked up the new `[File: ...] saved to ...` block (or the binary-type note) via `TestServer` + a follow-up read.

The unit tests added here (`file_saved_block_has_expected_format`, `redact_images_for_text_only_preserves_imagefile_path`, `test_strip_images_base64_has_no_path_hint`) are solid for the pure-function pieces, but per the stated policy they're a supplement to, not a replacement for, the `TestServer` integration test for the route itself. Flagging as a judgment call for the maintainer rather than fixing directly — writing that test means deciding how to fixture an upload + inspect injected session content, which felt like a design decision rather than a mechanical fix.

(Separately, confirmed no path-traversal issue: `file_path` is always `upload_dir.join(&att.file_id)` after `att.file_id` is validated as a UUID at L144-146 — `filename` is display-only, never used to construct a filesystem path.)

---

houko (commented):
Automated review of #6531. Logic looks sound (path-traversal already guarded via UUID validation on `file_id`; attachment ordering is a plain `Vec` iteration so the #3298 determinism rule is satisfied; no Docker/sandbox path concerns here).

Six inline findings, all left as comments rather than pushed directly: this session's egress policy blocks write access to `neo-wanderer/librefang` (the PR's head repo), so I couldn't land the fixes I'd otherwise have pushed. Summary:

- One real bug: a new test's doc comment got merged with an unrelated pre-existing test's doc comment during insertion, leaving one test unrelated docs and the other none.
- Several prose-wrapping violations (CLAUDE.md: one sentence per line) in new doc-comments/comments across `attachments.rs`, `agent_loop/mod.rs`, `message.rs`, and the new `utilities.rs` test — suggestions included.
- One design-level flag: the upload route's behavior change (new content block injected into every agent session, called out as a breaking change) doesn't yet have a `#[tokio::test]` against `TestServer` per the repo's mandatory integration-testing policy — only unit tests on the helper functions.

No Windows-specific path issues spotted (`path.display()` is informational text only, not used for further FS operations).

---

neo-wanderer:
Thanks for the review — pushed 47094db2d addressing the actionable findings:

- **Merged test doc comment (bug):** split apart — `file_saved_block_has_expected_format` keeps its own doc, and the "Regression: streaming attachment fetch must bound peak memory…" doc is restored onto `push_within_cap_rejects_over_cap_chunk`.
- **Prose wrapping (one sentence per line):** reflowed the new doc-comments/comments in `attachments.rs` (`file_saved_block` doc, `resolve_attachments` addendum, image-branch and else-branch comments), `agent_loop/mod.rs`, `message.rs` (both flagged blocks), and the new `utilities.rs` test line.

Comment-only changes; no behavior change.

On the `#[tokio::test]` against `TestServer` for the route-level behavior change: agreed it is the right coverage in principle. The honest end-to-end assertion (upload → resolved blocks injected into the session) has to either drive an agent turn or reach `resolve_attachments` through a currently-private module path, so wiring it reliably is a design decision rather than a mechanical fix — leaving that call to the maintainer as suggested. Happy to add it if you would like it in this PR.

PR #6532, opened 2026-07-21
feat(runtime-mcp): implement the MCP resources primitive (#6501)

### LINKED ISSUE #6501: feat(runtime-mcp): implement MCP resources primitive (resources/list + resources/read) and recognize resource_link tool content
## Summary

The MCP client in `librefang-runtime-mcp` is currently **tools-only**. It performs the `initialize` handshake, discovers/invokes tools (`tools/list` + `tools/call`), and declares the `roots` capability — but it does not implement the protocol's `resources` primitive (`resources/list`, `resources/read`, `resources/templates/list`), nor does it recognize the `resource_link` content type in tool results. This means a librefang agent cannot consume any MCP server whose data is exposed via resources rather than tools, and any `resource_link` a tool returns is silently flattened into an opaque JSON string.

This is a request for **protocol/client completeness** rather than a bug against a specific workflow. Filing it so the gap is tracked; happy to help scope or PR.

## Current behavior

Client capabilities are declared with only `roots` (`crates/librefang-runtime-mcp/src/lib.rs`, ~L854–L858):

```rust
let mut capabilities = rmcp::model::ClientCapabilities::default();
capabilities.roots = Some(rmcp::model::RootsCapabilities::default());
```

Tool-result handling extracts text-only content and stringifies everything else (`lib.rs`, ~L2502–L2514):

```rust
let texts: Vec<String> = result.content
    .iter()
    .filter_map(|item| item.as_text().map(|t| t.text.clone()))
    .collect();

let output = if texts.is_empty() {
    serde_json::to_string(&result.content)   // resource_link / embedded resource → opaque JSON
        .unwrap_or_else(|_| "No content".to_string())
} else {
    texts.join("\n")
};
```

Discovery only ever calls `tools/list`; there is no `resources/list` / `resources/read` path anywhere in the crate (grep for `resources/list`, `read_resource`, `ResourceContents`, `resource_link` returns nothing).

## Gap

1. **No resource discovery/read.** Servers that surface documents/data through `resources/*` are invisible to librefang agents. The only way such data reaches an agent today is if the server additionally wraps it in a tool.
2. **`resource_link` is not honored.** The 2025-06-18 spec revision lets a tool return `{ "type": "resource_link", "uri", "name", "mimeType" }` instead of inlining bytes — the intended escape hatch for large/binary payloads. Because the client only reads `as_text()`, a `resource_link` gets JSON-stringified into the tool output and no follow-up read/resolution happens.

## Proposed scope

Minimal, additive, on top of the `rmcp` 2.0 SDK already in use (which models resources natively):

- **Discovery:** `resources/list` (+ `resources/templates/list`) alongside the existing `tools/list`, surfaced to the agent as retrievable context or as synthetic read tools (design choice for maintainers).
- **Read:** `resources/read`, mapping `TextResourceContents` → text and `BlobResourceContents` (base64) → an appropriate handle/attachment path.
- **`resource_link` recognition** in `CallToolResult.content`: at minimum expose the URI + metadata as a first-class handle instead of stringifying it; optionally resolve `resources/read` on the referenced URI.
- Declare the `resources` client capability in the `initialize` handshake so servers advertise accordingly.

## Non-goals / notes

- No new transport needed — MCP carries resource contents in-band (text or base64 blob) over the existing JSON-RPC channel; this is purely handler-side wiring.
- `prompts/*` is a separate primitive and out of scope here.
- Backward compatible: declaring the `resources` capability and adding read paths does not change existing tool behavior.

Happy to genericize a minimal repro server and/or take the implementation if there's interest.

houko:
## Summary

Closes #6501. The `librefang-runtime-mcp` client was **tools-only** — it ran the `initialize` handshake, `tools/list` + `tools/call`, and declared `roots`, but never implemented the `resources` primitive. So an agent could not consume any MCP server that exposes its data as resources rather than tools, and a `resource_link` returned by a tool was silently flattened into an opaque JSON string.

This adds the resources primitive on both live transports, additive and backward-compatible.

## What's implemented

- **`resources/list`, `resources/read`, `resources/templates/list`** as methods on `McpConnection` — on the **rmcp** transport (stdio + streamable-HTTP) via `list_all_resources` / `read_resource` / `list_all_resource_templates`, and on the hand-rolled **SSE** transport via `sse_send_request`. **HttpCompat** has no resources concept and returns a clear "not supported" error.
- **Synthetic-tool surfacing (Option A).** When the server advertises the `resources` capability, the client registers two synthetic tools — `list_resources` and `read_resource(uri)` — that flow through the existing tool-call loop and are **intercepted before the transport `tools/call`**, so an agent can discover and read resources with no kernel changes. Interception is keyed on a per-connection `resource_ops` set, so a real server tool literally named `read_resource` is dispatched normally. The interception runs **after** the existing schema-validation and taint guards, so synthetic tools get the same argument checks as any other tool.
- **`resource_link` / embedded-resource recognition.** A tool result's `resource_link` block now renders as a first-class `[resource_link] name — uri (mime)` line instead of being JSON-stringified, and an embedded resource contributes its text. Binary blobs are **elided** (`[binary resource <mime>: N base64 bytes elided]`), never inlined into the prompt. Applied to both the rmcp and SSE tool-result extraction arms via shared pure helpers.
- **Capability gating.** rmcp path reads the server capability live from `peer_info().capabilities.resources`; SSE path captures `capabilities.resources` from the `initialize` response. No `resources` client capability is declared — it is a **server** capability and rmcp's `ClientCapabilities` has no such field (the issue's "declare the resources client capability" is a spec mismatch), so the client gates on the server's advertisement instead.
- **Determinism (#3298).** Resource / template lists are sorted by URI before they reach a tool result, keeping prompt-cache stability.

## Design note for reviewers

Surfacing uses **synthetic tools (Option A)** — fully self-contained in `librefang-runtime-mcp`, no kernel prompt-assembly changes. The scoping flagged Option A (synthetic tools) vs Option B (listable context in the prompt) as a maintainer design choice; Option A is the smaller, self-contained surface and the one the issue explicitly sanctions ("as synthetic read tools"). If you prefer Option B, the transport methods and renderers are reusable as-is and only the surfacing layer changes.

## Verification

- `cargo test -p librefang-runtime-mcp --lib` — **181 passed, 0 failed**, including 9 new tests:
  - pure helpers: `format_resource_link_includes_mime_when_present`, `render_resource_contents_inlines_text_and_elides_blobs`, `render_rmcp_content_block_handles_text_and_resource_link`, `render_json_content_block_mirrors_rmcp_rendering`, `parse_sse_resource_list_sorts_by_uri`, `parse_sse_read_resource_joins_contents`
  - `resources_list_read_deserialize_from_jsonrpc` (JSON-RPC shape)
  - `sse_resources_discovered_and_readable` (end-to-end against the SSE socket stub: capability advertised → synthetic tools registered → `read_resource` returns content)
  - `read_resource_without_uri_errors`
- `cargo clippy -p librefang-runtime-mcp --all-targets -- -D warnings` — clean.
- `cargo check -p librefang-runtime-mcp --lib --tests` — clean.

## Notes / non-goals

- `prompts/*` is a separate primitive and out of scope, as the issue states.
- Blob resources are elided rather than base64-inlined, to protect LLM context; a future PR can add an attachment/handle path if a concrete workflow needs the bytes.

### DIFF
--- a/crates/librefang-runtime-mcp/src/lib.rs
+++ b/crates/librefang-runtime-mcp/src/lib.rs
@@ -896,6 +896,18 @@ pub struct McpConnection {
     inner: McpInner,
     /// Current OAuth authentication state for this connection.
     auth_state: crate::mcp_oauth::McpAuthState,
+    /// Raw names of the synthetic resource tools this connection registered
+    /// (`list_resources` / `read_resource`), so `call_tool_with_caller` can
+    /// intercept them and route to the `resources/*` methods instead of a
+    /// `tools/call` — without shadowing a real server tool of the same name
+    /// (#6501).
+    resource_ops: std::collections::HashSet<String>,
+    /// Whether the server advertised the `resources` capability at handshake.
+    /// For the Rmcp path this is read live from `peer_info()`; for the
+    /// hand-rolled SSE path it is captured from the `initialize` response
+    /// (there is no stored capability object otherwise). Gates resource-tool
+    /// registration (#6501).
+    server_advertises_resources: bool,
 }
 
 /// Transport-specific connection handle.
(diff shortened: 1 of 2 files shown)


crates/librefang-runtime-mcp/src/lib.rs
```diff
@@ -2314,6 +2561,238 @@ fn json_type_name(v: &serde_json::Value) -> &'static str {
...
+                    .iter()
+                    .map(render_resource_contents)
+                    .collect::<Vec<_>>()
+                    .join("\n"))
+            }
+            McpTransportKind::Sse => {
+                let resp = self
+                    .sse_send_request("resources/read", Some(serde_json::json!({ "uri": uri })))
+                    .await?;
+                Ok(parse_sse_read_resource(resp))
```
houko, line 2689: `register_resource_tools` (and its caller `discover_and_register_resources`) registers the synthetic `list_resources` / `read_resource` tools unconditionally whenever the server advertises the `resources` capability — it never checks whether the server's real `tools/list` already declared a tool with one of those raw names.

The doc comment on `resource_ops` (field, ~line 899) and the comment at the interception site (`resource_ops.contains(&raw_name)`, ~line 2875) both explicitly claim "without shadowing a real server tool of the same name" / "so a real server tool literally named `read_resource` is dispatched normally." That guarantee doesn't actually hold: if a server exposes both the `resources` capability *and* a genuine tool literally named `read_resource` (or `list_resources`), `register_tool` is called twice for the same namespaced name (once for the real tool during `discovered_tools`/`sse_discover_tools`, once for the synthetic op here), and `resource_ops.insert("read_resource")` runs regardless of the collision. From then on, `call_tool_with_caller` will intercept *every* call to that namespaced tool name via `dispatch_resource_op`, so the real server tool becomes unreachable — the opposite of "unaffected."

This seems worth a real collision guard (e.g. skip registering a resource op — and skip the `resource_ops.insert` — if `self.tools`/`self.original_names` already contains a tool with that raw name), but the right resolution (skip only the colliding op vs. skip both vs. rename the synthetic tool) is a design call I didn't want to make unilaterally, so flagging rather than fixing directly.

---

houko (commented):
Reviewed against CLAUDE.md's stated rules (deterministic prompt ordering #3298, config-field conventions, prose wrapping, etc.). Pushed one mechanical fix (doc-comment rewrap to one-sentence-per-line per the prose-wrapping rule, no wording changes). One design-level concern left as an inline comment: the synthetic `list_resources`/`read_resource` tool registration doesn't actually implement the no-shadowing guarantee its own comments claim.

---

PR #6549, opened 2026-07-22
fix(skills): make skills/reload honest in frozen Stable mode (#6540)

### LINKED ISSUE #6540: skills/reload: silent no-op reported as success when registry frozen (Stable mode)
## Summary

`POST /api/skills/reload` unconditionally returns `200 OK {"status":"reloaded",...}` even when the skill registry is frozen (Stable mode) and no reload actually happened. New skill directories added to `~/.librefang/skills/` after boot are silently invisible to a running Stable-mode daemon, with no way to detect this from the API response — only a `restart` picks them up.

Confirmed present in both `v2026.7.11` and the current `v2026.7.21` release — the relevant functions are byte-identical between the two tags (0-line diff on `tools_and_skills.rs` / `skill.rs`).

## Root cause

- `crates/librefang-kernel/src/kernel/boot.rs:930-933` — when `config.mode == KernelMode::Stable`, the `SkillRegistry` is frozen once at boot via `skill_registry.freeze()`.
- `crates/librefang-kernel/src/kernel/tools_and_skills.rs:368-377` (`reload_skills`) — checks `registry.is_frozen()` and returns **immediately** without calling `load_all()`, only emitting a `warn!` to the daemon log:
  ```rust
  pub fn reload_skills(&self) {
      let mut registry = self.skills.skill_registry.write().unwrap_or_else(|e| e.into_inner());
      if registry.is_frozen() {
          warn!("Skill registry is frozen (Stable mode) — reload skipped");
          return;
      }
      ...
  }
  ```
- `crates/librefang-api/src/routes/skills/skill.rs:253-265` (`reload_skills` HTTP handler) — calls the kernel method above and **always** returns `200 OK` with `"status": "reloaded"`, regardless of whether the reload was actually skipped:
  ```rust
  pub async fn reload_skills(State(state): State<Arc<AppState>>) -> impl IntoResponse {
      state.kernel.reload_skills();
      let count = state.kernel.skill_registry_ref().read().map(|r| r.count()).unwrap_or(0);
      (StatusCode::OK, Json(serde_json::json!({"status": "reloaded", "count": count})))
  }
  ```
  The `warn!` never reaches the caller — only visible if you happen to be tailing `daemon.log`.

Notably, this is the same class of bug already fixed for `[browser]` config-reload in #6516 (`treat [browser] config-reload as restart-required, not a false hot-reload`), which introduced `ReloadPlan.restart_required` / `restart_reasons` so the reload report is honest instead of claiming success on a no-op. That fix didn't touch the skills-reload path, so the same failure mode remains there.

## Repro

1. Run with `mode = "stable"` in `config.toml`.
2. Add a new skill directory (with valid `skill.toml`) under `~/.librefang/skills/`.
3. `POST /api/skills/reload` → `200 {"status":"reloaded","count":N}` (N unchanged, new skill absent).
4. `GET /api/skills` → new skill still missing.
5. Only `librefang restart` picks it up (full reboot re-runs `boot()`'s `load_all()` before re-freezing).

## Impact

- Misleading success response makes this hard to diagnose without daemon.log access — looks like the new skill's manifest/content is wrong, not that reload was a no-op.
- Also affects legitimate content-refresh of an *already-loaded* skill in Stable mode: `reload_skills()` bails before reaching `load_all()`, even though the registry already exposes a freeze-safe single-skill path (`SkillRegistry::reload_skill(name)`, `crates/librefang-skills/src/registry.rs:496-500`, explicitly documented as "works even when frozen — it only refreshes an existing entry, never adds a new one"). The bulk reload endpoint doesn't use it.

## Suggested fix

Don't unfreeze the registry (that's an intentional Stable-mode safety boundary — a running production agent shouldn't gain new skills without an operator-initiated restart). Instead, make the reload honest and make the safe half of the operation actually work:

1. When frozen, iterate the *currently loaded* skill names and call the existing `SkillRegistry::reload_skill(name)` for each, to pick up on-disk content edits (already freeze-safe by design).
2. Diff on-disk skill directories against the currently loaded set to detect brand-new directories, and report them explicitly instead of silently dropping them.
3. Have the HTTP handler return an honest, partial-success shape, e.g.:
   ```json
   {
     "status": "partial",
     "refreshed": ["docx", "xlsx"],
     "new_skills_pending_restart": ["pdf-reader"],
     "restart_required": true,
     "reason": "registry frozen (Stable mode) — new skill directories require a daemon restart"
   }
   ```
   instead of the current unconditional `{"status":"reloaded"}`.

This mirrors the `restart_required`/`restart_reasons` pattern already established for config hot-reload (#6516) rather than inventing a new convention.

houko:
## Summary

Closes #6540. `POST /api/skills/reload` returned `200 {"status":"reloaded"}` even when the registry is frozen (Stable mode) and nothing reloaded: the kernel `reload_skills` bailed after a `warn!` when frozen, and the handler reported success regardless. New skill dirs added after boot were silently invisible, and even content edits to already-loaded skills were not refreshed — despite `SkillRegistry::reload_skill` being freeze-safe by design. Same honest-reload gap #6516 fixed for `[browser]` config, never applied to skills.

## Fix

The registry stays frozen (intentional Stable-mode boundary — a running agent must not gain new skills without an operator restart), but the operation is now honest and does the safe half of the work:

- kernel `reload_skills` returns `SkillReloadOutcome { frozen, refreshed, skipped_new, count }` instead of `()`. Frozen path refreshes each loaded skill via the freeze-safe `SkillRegistry::reload_skill` (picks up on-disk content edits) and diffs on-disk dirs against the loaded set via a new `SkillRegistry::unloaded_on_disk_dirs` to report brand-new dirs it deliberately did not load. Non-frozen keeps full `load_all()`.
- `unloaded_on_disk_dirs` canonicalizes paths so a symlinked skills dir doesn't misreport loaded skills as new.
- HTTP handler is honest: normal reload stays `{"status":"reloaded","count":N}`; frozen reload returns `{"status":"partial","frozen":true,"refreshed":[…],"skipped_new":[…],"count":N}` so a caller sees new skills need a restart.
- Also updates the `KernelApi::reload_skills` trait method (the API's actual call path) and the `skill_workshop` call site for the new return type; re-exports `SkillReloadOutcome`.

## Verification

Run with external `TMPDIR` (dev system disk was full):
- `cargo test -p librefang-skills` — 220 passed, incl. new `unloaded_on_disk_dirs_reports_only_new_skill_dirs`, `frozen_reload_skill_refreshes_existing_content`.
- `cargo test -p librefang-kernel --lib` — 57 passed, incl. `test_stable_mode_freezes_registry_and_skips_review_gate` (extended to assert the honest outcome) and `test_reload_skills_preserves_disabled_and_extra_dirs`.
- `cargo check -p librefang-api` clean; `cargo clippy -p librefang-skills -p librefang-kernel -p librefang-api --all-targets -- -D warnings` clean.


houko:
Automated CLAUDE.md compliance pass: the new prose added by this PR (doc-comments and multi-line `//` comment blocks in `tools_and_skills.rs`, `skill.rs`, `tests.rs`, `skill_workshop/mod.rs`, and `registry.rs`) was hard-wrapped mid-sentence at ~80 columns, which conflicts with the repo's "Prose wrapping: no column limit; break only at sentence boundaries" rule (applies to new prose, including doc-comments).

Pushed a comment-only reflow in f36a11f8 so each sentence is its own line; no code/logic changed, `cargo fmt --check -p librefang-kernel -p librefang-skills -p librefang-api` is clean.

Separately: GitHub is currently reporting this PR as having a merge conflict with `main` (`has-conflicts` label) — that's outside what an automated pass should resolve, so leaving it for a maintainer/author rebase.

---

houko:
Disclosure: commit f36a11f8 (the prose-reflow fix pushed above by this automated review) is authored as `Claude <noreply@anthropic.com>` — this session's sandboxed git identity, not something this routine is supposed to set given CLAUDE.md's own "no AI/Claude attribution" policy. The `commit-msg` hook only checks message content, not the author field, so it didn't catch this. The commit's content is a plain comment-only reflow with no logic change; flagging the identity issue for a maintainer to squash/amend if it matters before merge. No further automated commits will be made to this branch.

---

crates/librefang-api/src/routes/skills/skill.rs
```diff
@@ -251,17 +251,27 @@ pub async fn uninstall_skill(
...
+            "detail": if outcome.skipped_new.is_empty() {
+                "registry frozen (Stable mode); refreshed existing skills only"
+            } else {
+                "registry frozen (Stable mode); new skill directories require a daemon restart to load"
+            },
+        })
+    } else {
+        serde_json::json!({"status": "reloaded", "count": outcome.count})
+    };
+    (StatusCode::OK, Json(body))
```
houko, line 274: Per repo CLAUDE.md's MANDATORY Integration Testing section (and `crates/librefang-kernel/CLAUDE.md`: "Integration tests against a real router live in `librefang-api/tests/` — that's where `#[tokio::test]` against `TestServer` belongs, refs #3721") — this PR changes the HTTP response *shape* of `POST /api/skills/reload` (new `frozen`/`refreshed`/`skipped_new`/`detail` fields, and a new `status` value set of `"partial"`/`"refreshed"` alongside the existing `"reloaded"`), but the only new test is the kernel-level `test_stable_mode_freezes_registry_and_skips_review_gate` in `crates/librefang-kernel/src/kernel/tests.rs`, which exercises `LibreFangKernel::reload_skills()` directly and never touches this handler.

The existing HTTP-level coverage in `crates/librefang-api/tests/skills_routes_test.rs` (`skills_reload_picks_up_filesystem_drops`, line ~352) only asserts the non-frozen `{"status":"reloaded","count":N}` shape — nothing in the test suite boots a Stable-mode kernel through the router and asserts on the new frozen JSON body (in particular the `"partial"` vs `"refreshed"` status branch, or that `frozen`/`refreshed`/`skipped_new`/`detail` actually appear with the right values).

Could you add a `#[tokio::test]` in `skills_routes_test.rs` that boots a kernel with `mode = KernelMode::Stable` (mirroring the manual `LibreFangKernel::boot_with_config` + pre-boot skill install pattern already used in `test_stable_mode_freezes_registry_and_skips_review_gate`, then wrapping via `TestAppState::from_kernel`), drops a skill directory on disk before boot and another after boot, hits `POST /api/skills/reload`, and asserts the full frozen response body? That's the canonical replacement for "trust the kernel unit test" per #3721, and it would have caught any drift between `SkillReloadOutcome` and the handler's `serde_json::json!` mapping (e.g. the `"partial"` vs `"refreshed"` status selection, which is handler-only logic not covered by the kernel test at all).

---

crates/librefang-kernel/src/kernel/tools_and_skills.rs
```diff
@@ -365,15 +381,54 @@ impl LibreFangKernel {
...
+    /// Returns a [`SkillReloadOutcome`] describing what actually happened so callers (e.g. the `POST /api/skills/reload` handler) can report an honest result instead of claiming success on a no-op.
+    /// In Stable mode the registry is frozen at boot and never gains new skills without an operator-initiated restart; rather than silently skipping the whole reload, this refreshes the on-disk content of already-loaded skills (the freeze-safe `reload_skill` path) and reports any brand-new skill directories it is deliberately not loading (#6540).
+    /// Most callers invoke this for its side effect and ignore the return value.
+    pub fn reload_skills(&self) -> SkillReloadOutcome {
         let mut registry = self
             .skills
             .skill_registry
             .write()
             .unwrap_or_else(|e| e.into_inner());
         if registry.is_frozen() {
```
houko, line 394: `LibreFangKernel::reload_skills()` is shared by more than the `POST /api/skills/reload` handler this PR is titled around. `crates/librefang-kernel/src/kernel/config_reload_ops.rs` also calls `self.reload_skills()` from `HotAction::ReloadSkills`, which fires from `POST /api/config/reload` whenever the `skills` config block changes (documented as hot-reloadable, `H`, in `docs/operations/config-reload.md`).

Before this change, that path was a true no-op while frozen (Stable mode): `warn!` and return, no disk I/O, no state mutation. After this change, the same trigger now walks every already-loaded skill and re-reads + re-parses + re-scans (`reload_skill`) its `skill.toml`/`prompt_context.md` from disk, clears `prompt_metadata_cache.skills`, and bumps `skill_generation` — every time an operator changes an unrelated `skills.*` config field (e.g. `skills.disabled`) via config reload on a Stable-mode daemon.

Functionally this is probably fine (the freeze boundary itself is preserved — no new skills load), but it's a behavior change to a *second* call site that isn't mentioned in the PR description, the CHANGELOG entry, or `docs/operations/config-reload.md`'s `skills` row, and it turns a previously free operation into one with real disk I/O + cache invalidation cost on every config-reload cycle in Stable mode. Was this considered, and if so, is it worth a line in the CHANGELOG entry / config-reload doc so an operator investigating unexpected I/O on their Stable-mode daemon after a routine config reload can trace it back to this PR? If it wasn't considered, might be worth deciding whether `HotAction::ReloadSkills` in Stable mode should keep the old pure no-op behavior instead of routing through the now-heavier `reload_skills()`.

---

houko (commented):
Reviewed the diff against repo CLAUDE.md rules (integration-test mandate, config hot-reload classification, deterministic-prompt-ordering, prose-wrapping, commit hygiene). Two design-judgment items left as inline comments below; everything else checked out clean (see summary).

---

crates/librefang-kernel/src/kernel/tools_and_skills.rs
```diff
@@ -19,6 +19,22 @@
 use super::*;
 use super::{sanitize_reviewer_block, sanitize_reviewer_line, ReviewError};
```
houko, line 21: Not a code issue, but flagging for the record per CLAUDE.md's commit hygiene rule ("No Claude / Anthropic / AI attribution in commit messages, PR bodies, or issue comments... the rule exists because attribution pollutes `git log`"): commit `f36a11f` ("docs(skills): reflow #6540 prose comments to one-sentence-per-line"), which touches this file among others, is authored as `Claude <noreply@anthropic.com>` rather than the human contributor identity used on every other commit in this PR (`Evan <...>`). The commit *message* itself is clean, but the author identity is baked into permanent git history now that this PR is merged, which is the exact "signals provenance the project does not want to imply" case the rule calls out.

This can't be mechanically fixed post-merge without rewriting merged history (interactive rebase + force-push), which is out of scope for an automated review and against the repo's force-push policy — flagging so a maintainer can decide whether it's worth a `git commit --amend --author`-style follow-up or is acceptable as-is. Everything else in the diff (kernel `SkillReloadOutcome`, the frozen-mode HTTP response shape, the new kernel/registry/HTTP tests) looks correct and I verified `cargo check -p librefang-skills/-kernel/-api --lib` and all new/changed tests (`unloaded_on_disk_dirs_reports_only_new_skill_dirs`, `frozen_reload_skill_refreshes_existing_content`, `test_stable_mode_freezes_registry_and_skips_review_gate`, `skills_reload_frozen_stable_mode_reports_honest_result`) pass locally.

---

PR #6550, opened 2026-07-22
fix(runtime): stop lossily truncating tool results in the spill dead band (#6545)

### LINKED ISSUE #6545: bug(tool-results): 10-16 KB tool results lossily truncated instead of spilled to a recoverable artifact
## Summary

Two independent mechanisms bound tool-result size, and they are ordered so
that results in a **~6 KB dead band lose data irrecoverably**:

- **Artifact spill** (`spill_fresh_result`, `crates/librefang-runtime/src/agent_loop/tool_call.rs`)
  — runs at tool-exec time. If a result exceeds `spill_threshold_bytes`
  (**default `16_384`**, `default_spill_threshold_bytes`,
  `crates/librefang-types/src/config/types.rs`) the full bytes are written to
  the artifact store and replaced with a small, **recoverable** stub
  (`read_artifact`). Its own doc-comment states the intent:
  > "Spill a fresh tool result to the artifact store when it exceeds the
  > configured threshold, **preserving the full bytes before
  > `sanitize_tool_result_content` truncates them**."

- **Sanitizer size-cut** (`strip_tool_result_details`,
  `crates/librefang-runtime/src/session_repair.rs`) — runs later, when the
  result is assembled into the LLM message (`sanitize_tool_result_content`,
  `agent_loop/message.rs`). It hard-cuts at `max_len = 10_000` with a **lossy,
  non-recoverable** marker `"...[truncated from N chars]"`. No artifact, no
  handle.

Because the spill threshold (`16_384`) sits **above** the sanitizer cut
(`10_000`), a result whose size is in **`[10_000, 16_384)`** is:

- too small to spill → no recoverable artifact, **and**
- too big to survive the sanitizer → lossily truncated.

The bytes are gone. The spill mechanism was explicitly designed to preserve
them "before `sanitize_tool_result_content` truncates" — but the two
thresholds are ordered to defeat that intent on the happy path.

## The redundancy

`strip_tool_result_details` bundles three jobs:

```rust
let stripped = strip_base64_blobs(content);       // ① size-independent
let cleaned  = strip_injection_markers(&stripped);// ② size-independent
if cleaned.len() <= max_len { cleaned }           // ③ lossy size-cut (10_000)
else { format!("{}...[truncated from {} chars]", ...) }
```

- ① and ② are **security/noise stripping** — size-independent, must run inline
  on whatever text reaches the LLM (the stub, or a paged-in `read_artifact`
  chunk). Keep.
- ③ is **size-bounding** — and it is **redundant with artifact spill**. Spill
  already bounds context by replacing the payload with a tiny stub. Once spill
  has fired there is nothing left for ③ to cut; when spill has **not** fired
  (the 10–16 KB band), ③ destroys the exact bytes spill exists to preserve.
  ③'s only defensible role is a **fallback when the artifact write itself
  fails** (disk full / store error) — which the tool-budget layer already
  acknowledges elsewhere ("if the write fails, the content is truncated
  inline"). It should not fire on the happy path.

## Reproduce (synthetic)

1. Configure defaults (`spill_threshold_bytes = 16_384`).
2. Have an agent call any tool that returns a body of ~12 000 chars
   (e.g. `shell_exec` printing a 12 000-char string; any MCP tool whose useful
   payload sits past 10 KB works too).
3. Inspect the resulting tool-result message.

**Observed:** the result is inline-truncated to ~10 000 chars ending in
`...[truncated from 12000 chars]`. No artifact is written and there is no
`read_artifact` handle — the tail is unrecoverable.

**Boundary check:**
- `< 10_000` chars → passes verbatim.
- `> 16_384` chars → spills to a recoverable artifact stub (the 10 KB cut then
  sees only the small stub and does nothing). ✅
- `[10_000, 16_384)` → lossy truncation, **no artifact**. ❌

This is especially easy to hit with tools whose response leads with a large
fixed envelope (metadata/schema) and puts the caller-relevant fields at the
tail — the truncation lands exactly on the part the caller needs, while the
whole response is only ~12 KB, well under the spill line.

## Root cause

`default_spill_threshold_bytes()` returns `16_384` while
`strip_tool_result_details` hard-codes `max_len = 10_000`. The lossy cut is
gated *lower* than the recoverable spill, so the recoverable path never gets a
chance for results between the two.

## Proposed fixes (in preference order)

1. **Couple truncate → spill (addresses the redundancy directly).** When
   `strip_tool_result_details`/`sanitize_tool_result_content` would truncate
   (post-strip length exceeds the cap), spill the full bytes to an artifact and
   emit a **recoverable** stub instead of the lossy `...[truncated from N]`
   marker. This removes the dead band at *every* size, makes the size-bound
   lossless, and leaves ①/② running inline as today. The lossy inline cut
   remains only as the artifact-write-failed fallback.

2. **Failing that, enforce threshold ordering.** Guarantee
   `spill_threshold_bytes <= ` the sanitizer cap (clamp at config load, or
   derive one from the other), so spill always fires first and ③ only ever sees
   already-spilled stubs. Validate at config load and warn if an operator sets
   a spill threshold above the sanitizer cap.

3. **At minimum**, make the `10_000` cap a named, documented constant / config
   knob (it is currently an inline literal) and document the ordering
   requirement next to both sites.

Note: whatever the fix, the size-independent stripping (① base64, ② injection
markers) must continue to run inline regardless of spill, including on
`read_artifact` output.

## Related

- #3347 (tool-result context budget: cumulative cap, artifact spill, history
  fold) established the spill mechanism this issue extends.

## Env

- Affects `upstream/main` as of filing (verified: `session_repair.rs`
  `max_len = 10_000`; `config/types.rs` `default_spill_threshold_bytes = 16_384`;
  `tool_call.rs` `spill_fresh_result` doc-comment).

houko:
## Summary

Closes #6545. Two tool-result size mechanisms were ordered so a result in **[10_000, 16_384) bytes lost data irrecoverably**:

- **Artifact spill** (`spill_fresh_result`) fires above `spill_threshold_bytes` (default **16_384**) and replaces an oversized result with a small **recoverable** stub, preserving full bytes in the artifact store.
- **Sanitizer size cut** (`strip_tool_result_details`) later hard-cut at a hardcoded `max_len = 10_000` with a **lossy**, non-recoverable `...[truncated from N chars]` marker — no artifact.

A result too small to spill (< 16_384) but bigger than 10_000 was lossily truncated, defeating spill's documented intent to "preserve the full bytes before `sanitize_tool_result_content` truncates them".

## Fix

The sanitizer's base64-blob (①) and injection-marker (②) stripping are size-independent and still run unconditionally. The lossy size cut (③) is redundant with spill on the happy path — its only defensible role is a fallback when spill did **not** run (spill disabled, or the artifact write failed). Tie ③ to the spill threshold so the dead band vanishes:

- New `librefang_types::config::DEFAULT_SPILL_THRESHOLD_BYTES` const (16_384) backs `default_spill_threshold_bytes()`.
- `strip_tool_result_details` uses it as `max_len` instead of an independent `10_000`, so ③ only fires **above** the spill threshold, never in a band below it.

**Scope note:** the fallback cap is tied to the *default* spill threshold. An operator who sets a non-default `spill_threshold_bytes` below the default could still see a residual mismatch, but the default case the issue reports is fixed and the two no longer drift as independent literals. Threading the runtime-configured value into the sanitizer would need extra plumbing through `sanitize_tool_result_content` and is left as a follow-up if a non-default config surfaces it.

## Verification

Run with external `TMPDIR` (dev system disk was full):
- `cargo test -p librefang-runtime session_repair::tests` — 51 passed, incl. new `strip_preserves_dead_band_result_6545` (a ~12 KB result survives intact) and `strip_still_scrubs_dead_band_sized_content_6545` (base64/injection stripping still runs on dead-band-sized input), plus existing `test_strip_tool_result_details`.
- `cargo clippy -p librefang-runtime --all-targets -- -D warnings` clean.


crates/librefang-runtime/src/session_repair.rs
```diff
@@ -987,19 +987,25 @@ fn collect_tool_use_ids(messages: &[Message]) -> HashSet<String> {
...
+/// - As a last-resort fallback, lossily truncates content that exceeds the artifact-spill threshold
+///
+/// The base64/injection stripping is size-independent and always runs.
+/// The lossy size cut is a **fallback**, not the primary size bound: full-size results are meant to be preserved by artifact spill (`spill_fresh_result`, `spill_threshold_bytes`), which runs at tool-exec time and replaces an oversized result with a small *recoverable* stub before this sanitizer sees it.
+/// The cap here is therefore tied to `DEFAULT_SPILL_THRESHOLD_BYTES` so it only fires when spill did not run (spill disabled, or the artifact write failed) — never in a dead band *below* the spill threshold, which used to truncate 10–16 KB results irrecoverably even though spill exists to keep them (issue #6545).
 pub fn strip_tool_result_details(content: &str) -> String {
-    let max_len = 10_000;
+    // Fallback cap aligned with the spill threshold so there is no "too small to spill, too big to survive" band.
+    // Only reached when the recoverable spill path did not fire.
+    let max_len = librefang_types::config::DEFAULT_SPILL_THRESHOLD_BYTES as usize;
```
houko, line 1000: The PR description's "Scope note" defers threading the *configured* (non-default) `spill_threshold_bytes` into this fallback cap, citing the extra plumbing that would be needed through `sanitize_tool_result_content`.
That's a real blocker for the `compactor.rs:693` call site (no `ToolResultsConfig` in scope there), but it doesn't hold for the primary call site this issue is actually about: in `agent_loop/tool_call.rs` (around lines 1217-1232), the resolved `cfg: ToolResultsConfig` (from `ctx.opts.tool_results_config`) is already in scope a few lines above the call into `sanitize_tool_result_content` → `strip_tool_result_details`.
Per CLAUDE.md's "fix what you found" rule, deferring work that's reachable within the same crate you're already editing needs either doing it now or a stated reason a maintainer signs off on, not a unilateral follow-up note.
Would it be worth having `strip_tool_result_details` accept the resolved threshold (falling back to `DEFAULT_SPILL_THRESHOLD_BYTES` for callers like `compactor.rs` that don't have a `ToolResultsConfig` handy) instead of hardcoding the default unconditionally?
Not proposing this edit unilaterally since it touches a public fn signature and several call sites — flagging because the stated blocker doesn't actually apply to the main path the issue describes.

---

houko (commented):
Daily automated review pass. The fix itself is sound: it correctly closes the dead band by tying the sanitizer's fallback cap to the same constant that backs the spill threshold's default, keeps the size-independent base64/injection stripping unconditional, and adds direct regression coverage for the old dead band (`strip_preserves_dead_band_result_6545`, `strip_still_scrubs_dead_band_sized_content_6545`). Verification commands in the PR body are properly scoped (`cargo test -p librefang-runtime`, `cargo clippy -p librefang-runtime`), and the prior prose-wrap violation from an earlier commit in this branch was already self-corrected in `fa0d5943`.

One inline comment above on the "left as a follow-up" scope note — the stated blocker (extra plumbing) doesn't hold for the primary call site.

Separately, and not line-anchorable: commit `fa0d5943dc617606d9613fa49c543ad9cd86370f` on this branch ("docs: rewrap new doc-comment prose to one sentence per line (#6545)") has git author `Claude <noreply@anthropic.com>`. Repo CLAUDE.md prohibits Claude/AI attribution in commit history; the `commit-msg` hook checks message content but apparently doesn't catch attribution in the author field itself. I'm not rewriting this branch's history to fix it (force-push / rebase of someone else's pushed commits is out of scope for this automated pass) — flagging so a maintainer or @houko can re-author that commit (e.g. `git commit --amend --reset-author` + a normal, non-force-push-requiring path, or squash) before merge.

No merge/approve action taken; comment-only per policy.

---

crates/librefang-runtime/src/session_repair.rs
houko, line 1: Commit `fa0d5943` on this branch ("docs: rewrap new doc-comment prose to one sentence per line (#6545)", the commit that touches this file's doc comments) has git author identity `Claude <noreply@anthropic.com>`. CLAUDE.md's "No Claude / Anthropic / AI attribution" rule covers commit messages, PR bodies, and comments; this is the same signal showing up in the author field instead of the message body, which the `commit-msg` hook wouldn't catch since it only inspects message text, not author identity. Worth squashing/rebasing with a human author before merge so the merged history doesn't carry an AI-attributed commit.

---

PR #6577, opened 2026-07-26
fix(goals): treat blank parent_id / agent_id as absent instead of 404

### LINKED ISSUE #6562: Create a Goal, returns 404
### Description

on the goal page, I fill in the name and description and then hit Create and I get either 404 or Parent Goal can't be ""

http://192.168.0.1:4545/api/goals
POST Status Code 404 Not Found
Referrer Policy
strict-origin-when-cross-origin

### Expected Behavior

I would have expected it to create the goal, a multistep task, only guessing as it has never worked

### Steps to Reproduce

Open dashboard
select goal page
enter name and description
hit create the error
Also I have only seen the templates once, every subsequent visit to that page I have the same template waiting for me to create.

multiple restarts 

### LibreFang Version

2026.7.21

### Operating System

Linux (x86_64)

### Logs / Screenshots

_No response_

houko:
## Problem

Creating a goal from the dashboard has never worked. `POST /api/goals` returns `404 Parent goal '' not found` for every attempt.

The create form seeds its draft with `parent_id: ""` and `agent_id: ""` (`GoalsPage.tsx:119`) and `handleCreate` passed the whole draft straight to the mutation, so the request body carried both empty strings. On the backend, `create_goal` read the parent with a bare `as_str()`, which yields `Some("")` for an empty string, then ran the parent-existence check against a goal literally named `""` and 404'd.

The blank `agent_id` was persisted too. `create_goal` wrote it because the same `Some("")` passed its presence check, and `start_goal_run` later failed to parse it as a UUID and answered "Assign an agent to this goal before starting a run" — on a goal the user never assigned. Same root cause, so it is fixed here.

## Changes

- `crates/librefang-api/src/routes/goals.rs`
  - `optional_id_field` reads an optional id-like field and treats blank / whitespace-only as absent. Used for `parent_id` and `agent_id` in `create_goal`, so the stored document carries either a real id or omits the key.
  - `is_clear_signal` makes `""` mean the same as `null` on update — "clear this link". A form's reset `<select>` submits `""`, so that is the plain reading of the payload. Resolved once before the transaction so the validation branch and the mutation branch cannot disagree about what a blank string means.
  - Update also trims a non-blank id before storing it.
- `crates/librefang-api/dashboard/src/pages/GoalsPage.tsx` — `handleCreate` omits `parent_id` / `agent_id` from the payload when blank rather than sending `""`.

## Verification

- `cargo test -p librefang-api --test goals_routes_integration` — 31 passed, including five new cases:
  - `goals_create_treats_blank_parent_and_agent_ids_as_absent_6562`
  - `goals_create_treats_whitespace_parent_id_as_absent_6562`
  - `goals_create_with_real_parent_still_links_6562`
  - `goals_update_blank_parent_id_clears_link_6562`
  - `goals_update_blank_agent_id_clears_assignment_6562`
- `cargo fmt -p librefang-api -- --check` — clean.
- Dashboard: `pnpm typecheck`, `pnpm lint` (0 errors), `pnpm test --run` (865 passed, including the new `omits blank parent_id / agent_id from the create payload (#6562)` case), `pnpm build` — all pass.

## On the reporter's second observation

> Also I have only seen the templates once, every subsequent visit to that page I have the same template waiting for me to create.

That is current design, not a bug. `GET /api/goals/templates` returns a fixed catalog of six built-in templates (`goals.rs:605`), so the picker renders the same six on every visit. Applying one was never affected by this bug either: `handleApplyTemplate` (`GoalsPage.tsx:166`) posts `{title, description, status}` per template goal with no `parent_id`, which passed even before this change.

One real gap worth noting as an observation rather than a fix: template goals are created flat — the template shape carries no parent/child relationship, so applying "Product Launch" yields five sibling goals rather than a parent with four children. Changing that is a product decision about the template schema, outside this fix.

Fixes #6562


crates/librefang-api/dashboard/src/pages/GoalsPage.tsx
```diff
@@ -155,7 +155,15 @@ export function GoalsPage() {
     e.preventDefault();
     if (!createDraft.title.trim()) return;
     try {
-      await createMutation.mutateAsync(createDraft);
+      // Drop blank parent_id / agent_id instead of posting `""` (#6562): the
+      // form seeds both as empty strings, and an empty parent_id used to fail
+      // the backend's parent-existence check with "Parent goal '' not found".
+      const { parent_id, agent_id, ...rest } = createDraft;
+      await createMutation.mutateAsync({
```
houko: Per the dashboard data-layer rule in CLAUDE.md, API-shaping logic should live in the mutation hook (`src/lib/mutations/goals.ts: useCreateGoal`), not inline at a page call site.
Right now `handleApplyTemplate` a few lines below calls `createMutation.mutateAsync(goal)` directly with template-defined goals, so it does not get this same blank-omission treatment.
The backend already tolerates blank/whitespace `parent_id` / `agent_id` (`optional_id_field`), so this client-side omission is defense-in-depth rather than required for correctness today — but if it's worth doing at all, doing it once in `useCreateGoal`'s `mutationFn` would cover every caller instead of only the form submit path.
Is keeping it page-local intentional (e.g. because only the form has untrusted blank string state, while templates are assumed well-formed), or should this move into the hook?

---

houko (commented):
Automated daily review pass on PR #6562/#6577. Verified the 5 new Rust integration tests + 1 dashboard test actually exist and pass; verified `optional_id_field` / `is_clear_signal` handle whitespace-only strings correctly and consistently for create/update. Found and fixed one real inconsistency in `update_goal_by_id`: the self-parent guard and the parent existence/cycle-detection check compared the untrimmed `parent_id`, while the value is only trimmed at persist time — a whitespace-padded self-reference could slip past the guard, and a whitespace-padded but valid parent id could false-404. Pushed a fix + 2 regression tests. Left one inline comment on `GoalsPage.tsx` about whether the blank-omission logic belongs in `useCreateGoal`'s mutation hook instead of the page-level `handleCreate`, since `handleApplyTemplate` bypasses it.

---

Issue #4808, opened 2026-05-08
feat(types): add mcp_disabled field to AgentManifest — complement skills_disabled for MCP servers


DaBlitzStein:
## Summary

`AgentManifest` has `skills_disabled: bool` to explicitly disable all skills (since `skills = []` means "all skills"). There is no equivalent for MCP servers — `mcp_servers = []` means "all MCP servers" with no way to say "none".

## Problem

When configuring restrictive per-agent allowlists:
- Skills: `skills_disabled = true` → no skills. `skills = ["web-search"]` → only web-search. Works.
- MCP servers: `mcp_servers = []` → ALL servers. No way to say "none" without a sentinel value hack.

## Proposed fix

Add `mcp_disabled: bool` (default false) to `AgentManifest`, mirroring `skills_disabled`:

```rust
/// Explicitly disable all MCP server tools for this agent.
#[serde(default)]
pub mcp_disabled: bool,
```

When `mcp_disabled = true`, the agent receives zero MCP tools regardless of the global MCP server configuration.

## Files

- `crates/librefang-types/src/agent.rs` — add field
- `crates/librefang-kernel/src/kernel/tools_and_skills.rs` — check flag before injecting MCP tools
- `crates/librefang-kernel/src/kernel/prompt_context.rs` — skip MCP summary when disabled

PR #6578, opened 2026-07-26
fix(dashboard/agents): collect new_name on clone, reflect mcp_servers grants in Tools tab

### LINKED ISSUE #6565: bug(dashboard): Tools tab does not reflect MCP servers granted via mcp_servers allowlist
## Symptom

Agent detail → **Tools** tab (added in #5677) shows an MCP server as "AVAILABLE / click to assign" (0 tools recognized) even when that server is fully granted to the agent and its tools are actively called by the LLM in production.

Setup pattern that reproduces it:
- `agent-A`: MCP server `server-x` granted via the server-level `mcp_servers` allowlist (`mcp_servers: ["server-x"], mcp_servers_mode: "allowlist"`), with `capabilities_tools` containing only builtin tool names (no `mcp_server-x_*` entries), and one specific tool from `server-x` listed in `tool_blocklist` as an exception.
- `agent-B`: same server granted through the dashboard's per-tool picker (#5259), which additionally expands every `mcp_<server>_<tool>` name into `capabilities_tools`.

Result: Tools tab for `agent-A` shows `server-x` under "AVAILABLE", tools counter reflects only the builtin count — even though `agent-A`'s system prompt references and calls `mcp_server-x_*` tools successfully at runtime (the `tool_blocklist` exception only makes sense if the rest of the server is granted).

## Root cause (read from current source, v2026.7.21 / git_sha 7835a8c)

`crates/librefang-api/dashboard/src/pages/AgentsPage.tsx`:

```tsx
// line ~1750
const declared = agentToolCfg?.capabilities_tools ?? [];
...
const draft = toolsDraft ?? [];
const draftSet = new Set(draft);

// line ~1757
const getGroupStatus = (groupTools: ToolDefinition[]): "full" | "partial" | "none" => {
  const count = groupTools.filter((t) => draftSet.has(t.name)).length;
  if (count === groupTools.length) return "full";
  if (count > 0) return "partial";
  return "none";
};

// line ~1824
const assignedGroups = sortedGroups.filter(([, tools]) => getGroupStatus(tools) !== "none");
const availableGroups = sortedGroups.filter(([, tools]) => getGroupStatus(tools) === "none");
```

`getGroupStatus` (and therefore the whole assigned/available split rendered on the Tools tab) derives group status **exclusively** from `capabilities_tools` (via `toolsDraft`/`declared`). It has no awareness of the agent's `mcp_servers` allowlist field at all. Kernel-side, a server can be granted either by listing it in `mcp_servers` (whole-server grant, with `tool_blocklist` as the exception mechanism) or by explicit per-tool entries in `capabilities_tools` (the picker flow from #5259). The Tools tab only accounts for the latter.

## Why the observable behavior happens

Two independent, both-valid grant paths exist server-side, but only one is rendered. An agent configured through the "MCP servers" tab (whole-server grant) or through direct API calls that only set `mcp_servers`/`tool_blocklist` will always show its granted MCP servers as unassigned/0-tools on the Tools tab, regardless of whether the tools are actually reachable by the LLM at runtime.

## Proposed fix

**Minimal:** in the Tools tab data prep (around `AgentsPage.tsx:1750`), when computing `declared`/`draftSet`, also treat any server present in `agent.mcp_servers` (mode `allowlist`) as contributing its full known tool set minus `tool_blocklist` entries, instead of relying solely on `capabilities_tools`.

**Defensive (optional):** alternatively, normalize on save — whenever an MCP server is added via the MCP tab, also expand and persist its tool names into `capabilities_tools` (mirroring what the per-tool picker from #5259 already does), so there's a single source of truth for "assigned" state across both flows.

## Setup (for reference)

- LibreFang `v2026.7.21` (git_sha `7835a8c`), confirmed via `/api/version`
- Verified end-to-end: compared two real agents where one was granted an MCP server via `mcp_servers`-only path and the other via the dashboard per-tool picker; confirmed via `GET /api/agents/{id}` and `GET /api/agents/{id}/tools` that `capabilities_tools` diverges from `mcp_servers` in the former case, and confirmed via the agent's own system prompt + a blocklist entry that the "unassigned" server's tools are in active use.
- Not a duplicate of #5677 (built the tab), #5259 (built per-tool picker), #6045 (persistence-to-disk bug), or #6495 (glob bug in `tool_allowlist`) — those are adjacent but don't cover this display gap.

Happy to test a patch.

Refs: #5677, #5259, #6045, #6495 (related, not duplicates)

### LINKED ISSUE #6566: bug(dashboard): Clone button always fails — new_name never collected or sent
## Symptom

Clicking the **Clone** button on an agent's detail page in the dashboard always fails. Reproduced directly against the API: `POST /api/agents/{id}/clone` with the exact body the dashboard sends returns `422 Unprocessable Entity` — `missing field new_name` — for every agent, every time.

## Root cause (read from current source, v2026.7.21 / git_sha 7835a8c)

Backend requires `new_name` with no default:

```rust
// crates/librefang-api/src/routes/agents/cloning.rs
#[derive(serde::Deserialize, utoipa::ToSchema)]
#[serde(deny_unknown_fields)]
pub(crate) struct CloneAgentRequest {
    pub new_name: String,                              // required, no #[serde(default)]
    #[serde(default = "default_clone_true")]
    pub include_skills: bool,
    #[serde(default = "default_clone_true")]
    pub include_tools: bool,
}
```

Dashboard API client sends an empty body and never collects a name:

```ts
// crates/librefang-api/dashboard/src/api.ts:1543
export async function cloneAgent(agentId: string): Promise<ApiActionResponse> {
  return post<ApiActionResponse>(`/api/agents/${encodeURIComponent(agentId)}/clone`, {});
}
```

```tsx
// crates/librefang-api/dashboard/src/pages/AgentsPage.tsx:~2740
onClick={async () => {
  try {
    await cloneMutation.mutateAsync(detailAgent.id);   // no new_name collected anywhere
  } catch (err) {
    addToast(toastErr(err, t("agents.clone_failed", { defaultValue: "Failed to clone agent" })), "error");
  }
}}
```

There is no dialog/prompt anywhere in the Clone button's click handler (or upstream) to collect a name for the new agent before calling `mutateAsync`.

## Why the observable behavior happens

The mutation always posts `{}`. The backend correctly rejects it with 422 since `new_name` has no default. The click handler catches the error and shows a generic "Failed to clone agent" toast, so from the user's perspective the button "just doesn't work" with no indication of why.

## Proposed fix

**Minimal:** add a name-input dialog (e.g. reusing the existing `confirmDialog`/modal pattern already used elsewhere on this page, such as the reset-session confirmation) that collects `new_name` before calling `cloneMutation.mutateAsync({ agentId, new_name, include_skills, include_tools })`, and update `cloneAgent()` in `api.ts` to accept and forward those fields.

## Setup (for reference)

- LibreFang `v2026.7.21` (git_sha `7835a8c`)
- Verified via direct `POST /api/agents/{id}/clone` call reproducing the exact 422, plus reading both the backend request struct and the dashboard's mutation/click-handler code — this isn't a recent regression, the one commit touching `cloning.rs` is an unrelated file-split refactor (#5975); the frontend appears to have never sent `new_name`.
- Not a duplicate of #5639 (missing backend semantic test coverage for clone/reload/push routes) — that's about test debt, not this frontend/backend contract mismatch.

Happy to test a patch.

Refs: #5639 (related, not duplicate)

houko:
Two contract mismatches between the agents dashboard and the kernel. Both live in `AgentsPage.tsx`, so they ship together rather than as two PRs racing on the same file.

## #6566 — Clone always fails

`POST /api/agents/{id}/clone` requires `new_name` (`CloneAgentRequest` has no `#[serde(default)]` on it) and the struct is `#[serde(deny_unknown_fields)]`. The client posted `{}` and nothing in the click handler — or upstream of it — ever collected a name, so the backend correctly answered 422 and the handler swallowed it into a generic "Failed to clone agent" toast. From the user's side the button just didn't work.

Changes:
- `api.ts` — `cloneAgent(agentId, payload)` takes a `CloneAgentPayload`. The type is documented as exactly `new_name` / `include_skills` / `include_tools` because `deny_unknown_fields` makes any extra key a 422.
- `lib/mutations/agents.ts` — `useCloneAgent` takes `{ agentId, payload }`.
- `AgentsPage.tsx` — Clone opens a dialog pre-filled with `<source>-copy`, with an editable name and the two include toggles the endpoint already supported. Submit is disabled while the name is blank or the mutation is in flight.

## #6565 — Tools tab ignores `mcp_servers` grants

The tab derived every group's assigned/available state from `capabilities_tools` alone. But the kernel has **two** grant paths (`tools_and_skills.rs`, `available_tools`):

- Builtin tools are filtered by `capabilities.tools`.
- MCP tools are granted by the agent's `mcp_servers` allowlist and are explicitly *not* filtered by `capabilities.tools`. The kernel says so at the call site: "MCP tools are NOT filtered by capabilities.tools … capabilities.tools governs builtin tools only — MCP tool names are dynamic and unknown at agent-definition time. Use tool_blocklist to restrict specific MCP tools if needed."

So an agent granted a whole server through `mcp_servers` always rendered as "AVAILABLE / click to assign" with its tool count excluded, while the LLM was calling those tools successfully — exactly the report.

Changes:
- `api.ts` — `AgentDetail` declares `mcp_servers`, `mcp_servers_mode`, `tools_disabled`, `skills_disabled`. **The backend already emitted all four** (`routes/agents/lifecycle.rs: get_agent`); only the dashboard type was missing them, so no API change was needed.
- New `lib/toolGrants.ts` — the kernel's grant semantics as tested pure functions: `toolPatternMatches` (mirror of `librefang_types::capability::glob_matches` for the separator-free tool-name branch), `normalizeMcpName` (mirror of `librefang_runtime::mcp::normalize_name`), `resolveMcpGrantMode` (mirror of `routes::agents::mcp_servers_mode`, used as a fallback when an older daemon omits the field), `isMcpServerGranted`, `isToolBlocked`.
- `AgentsPage.tsx` — MCP group state is `granted(server) && !blocklisted(tool)`; builtin groups keep using the `capabilities_tools` draft. Blocklisted tools inside an expanded MCP group are labelled, which is what makes the reporter's "one tool as an exception" setup legible. The `usesAll` view no longer labels an ungranted MCP server "included".

### MCP groups are now read-only on this tab

`PUT /api/agents/{id}/tools` carries only `capabilities_tools` / `tool_allowlist` / `tool_blocklist` — there is no `mcp_servers` field. The old group toggle therefore wrote `mcp_<server>_<tool>` names into `capabilities_tools`, where the kernel ignores them for grant purposes: the click appeared to work and changed nothing reachable. Rather than leave a control that lies, MCP groups render their real grant state and point at the MCP servers tab, and `handleCustomize` no longer seeds `mcp_*` names into the allowlist draft.

This is the reason I did not take the issue's "defensive" alternative (expand MCP tool names into `capabilities_tools` on save). That would persist names the kernel discards, making `agent.toml` describe a grant the runtime does not honour, with `mcp_servers` still the real authority. Making the display truthful is the smaller and more accurate change. Moving MCP grant editing onto this tab would mean a new field on the tools endpoint or a second `PATCH /api/agents/{id}` call from the same Save button — a backend/API decision worth its own issue, not a silent add-on here.

## Verification

- `pnpm typecheck` — clean.
- `pnpm lint` — 0 errors (85 pre-existing warnings, unchanged).
- `pnpm test --run` — 884 passed across 87 files, up from 864/86:
  - 20 new cases in `lib/toolGrants.test.ts` covering glob edge cases (prefix / suffix / interior wildcard, non-overlapping prefix+suffix, wildcard-free patterns being exact-only), blocklist globs, MCP name normalization, and all three grant modes.
  - `useCloneAgent` case rewritten out of the shared `describe.each` (its argument shape differs now) to assert the exact three-key payload reaches `cloneAgent`.
- `pnpm build` — succeeds.
- New i18n keys added to `en` / `ko` / `uk` / `zh`; the `locale-parity` and `en-locale-coverage` suites enforce this and pass.

Note on test placement: `renderToolsTab` is a closure inside `AgentsPage` (the page's only exported sub-component is `SystemPromptSection`), so it isn't independently mountable. That is precisely why the grant logic moved into `toolGrants.ts` — the decision that was wrong is now covered by unit tests instead of being re-derived inline. No integration test is included because neither fix changes route shape; both endpoints are unchanged.

Fixes #6566
Fixes #6565


crates/librefang-api/dashboard/src/pages/AgentsPage.tsx
```diff
@@ -1754,25 +1778,65 @@ export function AgentsPage() {
...
 
-    const getGroupStatus = (groupTools: ToolDefinition[]): "full" | "partial" | "none" => {
-      const count = groupTools.filter((t) => draftSet.has(t.name)).length;
+    // #6565: MCP tools are granted by the agent's `mcp_servers` allowlist, not
+    // by `capabilities_tools` — the kernel explicitly skips the declared-tools
+    // filter for them (`tools_and_skills.rs`, Step 3). Reading MCP group state
+    // off `capabilities_tools` reported a whole-server grant as "AVAILABLE /
+    // click to assign" while the agent was actively calling those tools.
+    const blocklist = agentToolCfg?.tool_blocklist ?? [];
+    const mcpMode = agent.tools_disabled
```
houko: `mcpMode` only short-circuits to `"none"` on `agent.tools_disabled`, then falls back to `resolveMcpGrantMode(agent.mcp_servers, agent.mcp_servers_mode)`.
That misses the separate `mcp_disabled` manifest flag (`librefang_types::agent::AgentManifest::mcp_disabled`, default `false`, added in #4808).
In the kernel, `tools_and_skills.rs` gates MCP tool inclusion with `if !mcp_disabled && !mcp_allowlist.is_empty()` (around line 235) — independently of `tools_disabled`.
So an agent with `mcp_disabled = true` and a non-empty `mcp_servers` allowlist gets **zero** MCP tools from the kernel, but this tab would still render the group as granted, since `mcp_disabled` isn't in `AgentDetail` / not returned by `GET /api/agents/{id}` (`lifecycle.rs::get_agent`) at all.
Given `mcp_disabled` defaults to `false` this is a narrow edge case, but it directly contradicts the PR's stated goal ("reflect mcp_servers grants... accurately") for the agents that do set it.
Fixing it needs a backend change (expose `mcp_disabled` on `get_agent`'s JSON + the `AgentDetail` type) in addition to the frontend derivation, so flagging rather than fixing inline here.

---

crates/librefang-api/dashboard/src/lib/toolGrants.ts
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * - **Builtin tools** are filtered by `capabilities.tools` (the dashboard's
+ *   `capabilities_tools`). An empty list — or one containing `"*"` — means
+ *   unrestricted.
+ * - **MCP tools** are granted by the agent's `mcp_servers` allowlist and are
+ *   explicitly NOT filtered by `capabilities.tools`: "MCP tool names are
+ *   dynamic and unknown at agent-definition time. Use tool_blocklist to
+ *   restrict specific MCP tools if needed." An empty `mcp_servers` grants no
+ *   server; `["*"]` grants every connected server (#5855).
+ *
```
houko, line 7: This docstring (and `isMcpServerGranted`/`isToolActive` in `AgentsPage.tsx`) accounts for `tool_blocklist` but not the separate `tool_allowlist` override.
In `tools_and_skills.rs`, Step 4 applies both after MCP tools have already been added in Step 3: `if !tool_allowlist.is_empty() { all_tools.retain(|t| tool_allowlist.iter().any(|a| glob_matches(a, &t.name))); }` runs on `all_tools`, which by that point includes the MCP tools from Step 3 — so a non-empty `tool_allowlist` that doesn't include a matching pattern (e.g. `mcp_*` or `*`) filters MCP tools out entirely, the same way `tool_blocklist` does.
`agentToolCfg.tool_allowlist` is already fetched and used elsewhere on this page (e.g. the allowlist editor around line 3048), so the data is available — `isToolActive`/`isMcpGroupGranted` in `AgentsPage.tsx` (~1786-1800) just don't consult it.
Net effect: an agent with a restrictive `tool_allowlist` set would show its MCP group as "granted via mcp_servers" here while the kernel actually serves none of those tools to it.
`toolPatternMatches` is already exported and could mirror this (same shape as `isToolBlocked`), but wiring it in touches the render tree and adds another per-tool state (allowed-but-not-in-allowlist) that isn't in this PR's screenshots/description, so raising it for a call on scope rather than adding it unreviewed.

---

houko (commented):
Reviewed the diff against CLAUDE.md's dashboard data-layer rules and the kernel's actual tool-grant semantics.

Verification performed on the `fix-dashboard-agent-tools-clone` branch (pnpm install, node 22 / pnpm 10):
- `pnpm typecheck` — clean, no errors.
- `pnpm test` (full dashboard suite) — 87 files / 884 tests passed.
- `pnpm test:i18n-parity` — ko/zh clean; uk.json reports 8 extra keys, but none of them are touched by this PR (confirmed against `origin/main`'s uk.json — pre-existing drift, unrelated to #6565/#6566).
- `pnpm lint` — 0 errors; the only warnings in `AgentsPage.tsx` are pre-existing `react-hooks/exhaustive-deps` items outside the lines this PR touches.

Cross-checked `CloneAgentPayload` / `deny_unknown_fields` against `crates/librefang-api/src/routes/agents/cloning.rs::CloneAgentRequest` (matches exactly: `new_name` required, `include_skills`/`include_tools` default `true`), and `toolGrants.ts`'s glob/normalize helpers against `librefang_types::capability::glob_matches` and `librefang_runtime_mcp::normalize_name` (both mirrored correctly for tool names, which never contain `.` so the dot-separator branch in `glob_matches` is unreachable here). i18n key parity across en/ko/uk/zh is exact for the keys this PR adds.

Left two inline comments on gaps in the "reflect mcp_servers grants" logic that the diff doesn't cover: the per-agent `mcp_disabled` flag (independent of `tools_disabled`, and not even exposed by `GET /api/agents/{id}` today) and the `tool_allowlist` override (which, like `tool_blocklist`, filters MCP tools in Step 4 of `tools_and_skills.rs` but isn't consulted by `isMcpGroupGranted`/`isToolActive`). Both are narrower than the bug this PR fixes and need a backend field or a render-tree/i18n change respectively, so raising for a scope call rather than fixing unreviewed. No other issues found; no mechanical fixes were needed.

---

crates/librefang-api/dashboard/src/lib/toolGrants.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/** Mirror of `librefang_types::capability::glob_matches` for tool names.
+ *
+ * Tool names carry no `/`, `\` or `.`, so only the separator-free branch
+ * (`glob_matches_simple`) is reachable for them: `*` matches any run of
+ * characters, a bare `*` matches everything, and a pattern with no wildcard
+ * must match exactly.
+ */
+export function toolPatternMatches(pattern: string, value: string): boolean {
```
houko, line 25: `toolPatternMatches` diverges from the kernel's `glob_matches_simple` (`crates/librefang-types/src/capability.rs`) for patterns with **two or more** `*`. The kernel's multi-wildcard branch only strips a literal prefix before the first `*` (via `strip_prefix`), it doesn't search for the next literal segment — so a gap between the first and second wildcard fails to match on the Rust side. This TS version instead does a proper `indexOf` search for every interior segment, which is more "correct" glob semantics but not what the kernel actually does.

Concretely, with the kernel's real matcher:
```rust
glob_matches_simple("mcp_*_search_*_v2", "mcp_brave_search_web_v2") // => false
glob_matches_simple("get_*_v*_data", "get_helper_v2_extra_data")    // => false
```
but this file's `toolPatternMatches` returns `true` for both (verified with a standalone repro of both algorithms).

Since this function backs `isToolBlocked`, a `tool_blocklist` entry with two-plus wildcards would show as "blocked" in the Tools tab while the kernel still lets the agent call that tool — the opposite of the #6565 bug this PR fixes (understating a grant), but the same class of dashboard/kernel state mismatch. `tool_blocklist` entries with a single wildcard (the only shape covered by the new tests) are unaffected.

Worth either mirroring the kernel's actual (limited) recursion exactly, or fixing `glob_matches_simple` itself in the same PR if the "any characters" reading is the intended behavior — cross-crate, so flagging rather than changing it myself here.

---

houko (commented):
Verified the fix end-to-end: `pnpm typecheck`/`lint`/`test --run` all green (884/884), and the backend `CloneAgentRequest` contract matches the new `CloneAgentPayload` exactly. One edge-case finding left inline on `toolGrants.ts` — not blocking, just flagging a kernel/dashboard glob-matching divergence for multi-wildcard blocklist patterns.

---

crates/librefang-api/dashboard/src/pages/AgentsPage.tsx
```diff
@@ -1754,25 +1775,56 @@ export function AgentsPage() {
...
       (draft.length !== declared.length || draft.some((n) => !declared.includes(n)));
 
-    const getGroupStatus = (groupTools: ToolDefinition[]): "full" | "partial" | "none" => {
-      const count = groupTools.filter((t) => draftSet.has(t.name)).length;
+    // #6565: MCP tools are granted by the agent's `mcp_servers` allowlist, not by `capabilities_tools` — the kernel explicitly skips the declared-tools filter for them (`tools_and_skills.rs`, Step 3).
+    // Reading MCP group state off `capabilities_tools` reported a whole-server grant as "AVAILABLE / click to assign" while the agent was actively calling those tools.
+    const blocklist = agentToolCfg?.tool_blocklist ?? [];
+    const mcpMode = agent.tools_disabled
+      ? "none"
+      : resolveMcpGrantMode(agent.mcp_servers, agent.mcp_servers_mode);
```
houko: `mcpMode` only checks `agent.tools_disabled` before falling back to `resolveMcpGrantMode(agent.mcp_servers, agent.mcp_servers_mode)`. It doesn't account for `AgentManifest.mcp_disabled` (`crates/librefang-types/src/agent.rs:1155`, #4808), which is a separate kill-switch documented specifically for this case: "guarantee no MCP tools even if `[\"*\"]` or a non-empty list is set" (agent.rs:1131-1132). The kernel gate at `crates/librefang-kernel/src/kernel/tools_and_skills.rs:236` is `!mcp_disabled && !mcp_allowlist.is_empty()` — so an agent with `mcp_disabled = true` and `mcp_servers = ["*"]` gets **zero** MCP tools from the kernel, but with this PR's change the Tools tab will still render every MCP group as "granted via mcp_servers" (mcpMode resolves to `"all"` since `agent.mcp_disabled` is never consulted).

This needs three small additions to close: emit `"mcp_disabled": entry.manifest.mcp_disabled` alongside `mcp_servers_mode` in `crates/librefang-api/src/routes/agents/lifecycle.rs` (~line 1070), add `mcp_disabled?: boolean` to `AgentDetail` in `api.ts` (next to `mcp_servers_mode` at line 1271), and fold it into the ternary here, e.g. `agent.tools_disabled || agent.mcp_disabled ? "none" : resolveMcpGrantMode(...)`.

---

crates/librefang-api/dashboard/src/lib/toolGrants.ts
```diff
@@ -0,0 +1,75 @@
...
+/** Mirror of `librefang_runtime::mcp::normalize_name`. */
+export function normalizeMcpName(name: string): string {
+  return name.toLowerCase().replace(/-/g, "_");
+}
+
+/** Mirror of `librefang_types::capability::glob_matches` for tool names.
+ *
+ * Tool names carry no `/`, `\` or `.`, so only the separator-free branch (`glob_matches_simple`) is reachable for them: `*` matches any run of characters, a bare `*` matches everything, and a pattern with no wildcard must match exactly.
+ */
+export function toolPatternMatches(pattern: string, value: string): boolean {
```
houko, line 25: This mirrors `glob_matches_simple`, but the backend's `tool_blocklist` matching (`crates/librefang-kernel/src/kernel/tools_and_skills.rs:324`) actually calls the general `glob_matches` (`crates/librefang-types/src/capability.rs:215`), not `glob_matches_simple` directly. `glob_matches` special-cases patterns/values containing `.` by routing to `glob_matches_with_separator(pattern, value, '.')` (segment-by-segment dot matching, capability.rs:242-244) before falling through to the simple substring logic used here.

For plain tool names (no dots) this is a distinction without a difference, but if any MCP tool name or a `tool_blocklist` entry ever contains a `.` (e.g. a namespaced tool id), this TS mirror and the Rust source of truth will disagree on whether it's blocked. Worth either asserting tool names never contain `.` (and documenting that assumption in the comment above `toolPatternMatches`), or adding the dot-segment branch to match `glob_matches` exactly.

---

Issue #4898, opened 2026-05-11
Per-session model override (dashboard model picker silently overwrites agent default)


neo-wanderer:
## Problem

The dashboard's chat model picker is presented in a session-scoped UI (it sits inside the per-session chat view, next to the agent/provider badge), but it is wired to agent-scoped state. Selecting a model while viewing session A overwrites the agent's default model, and session B (same agent) picks up the change on its next LLM call.

For users with multiple long-lived sessions per agent — e.g. one persistent session per channel, one per cron job, plus interactive chats — this is surprising and lossy: there is no way to say "use a cheap model for this background session, keep the agent default for everything else" without forking a new agent.

## Current behavior — evidence

Frontend sends only `agentId`, no `sessionId` — `crates/librefang-api/dashboard/src/pages/ChatPage.tsx:2121-2123`:

\`\`\`ts
await modelConfigMutation.mutateAsync({
  agentId,
  config: { model: model.id, provider: model.provider },
});
\`\`\`

Endpoint is \`PATCH /api/agents/{id}/config\` (or \`/api/agents/{id}/hand-runtime-config\` for hand-derived agents).

Backend writes to the agent manifest — \`crates/librefang-kernel/src/kernel/agent_state.rs:75\` \`set_agent_model(...)\` mutates \`manifest.model.{model,provider}\` and persists via \`persist_manifest_to_disk\`. Nothing session-scoped is touched.

Session schema has no model field — \`crates/librefang-memory/src/session.rs:72-90\`:

\`\`\`rust
pub struct Session {
    pub id: SessionId,
    pub agent_id: AgentId,
    pub messages: Vec<Message>,
    pub context_window_tokens: u64,
    pub label: Option<String>,
    pub messages_generation: u64,
    pub last_repaired_generation: Option<u64>,
}
\`\`\`

Runtime read path always pulls from the agent manifest — every LLM dispatch in \`crates/librefang-runtime/src/agent_loop.rs\` reads \`manifest.model.model\` / \`manifest.model.provider\` with no session-level fallback.

## Prior art in this codebase

Cron's \`AgentTurn\` action already has a per-fire override at \`crates/librefang-types/src/scheduler.rs:124-125\`:

\`\`\`rust
/// Optional model override for this turn.
model_override: Option<String>,
\`\`\`

So the resolution-order pattern (per-turn > per-session > agent default) already exists at one boundary; this proposal generalizes it to interactive sessions.

## Proposed solution

[...]

PR #6580, opened 2026-07-26
fix(kernel): honour agent.toml context_window on the paths that run a turn

### LINKED ISSUE #6568: agent.toml context_window ignored — falls back to 8192 for unknown models
## Description

When an agent uses a model not in the built-in catalog (e.g. `deepseek-v4-flash` via a custom `provider_urls` override), the `context_window` field in `agent.toml` is ignored and the runtime falls back to a conservative 8192 tokens.

The warning says:
```
Model not in catalog — falling back to conservative context window. 
Set `model.context_window` in agent.toml to silence this warning.
```

But setting `context_window` in `[model]` does NOT silence the warning or change the actual context window used.

## Steps to Reproduce

1. Add a provider URL override in `config.toml`:
```toml
[provider_urls]
deepseek = "https://proxy.example.com/v1"
```

2. Set `DEEPSEEK_API_KEY` in the environment

3. Configure an agent with:
```toml
[model]
context_window = 131072
provider = "deepseek"
model = "deepseek-v4-flash"
max_tokens = 131072
```

4. The model IS in the catalog (`librefang models list --provider deepseek` shows `deepseek-v4-flash` with 1000000 context)

5. But the runtime still says "Model not in catalog" and falls back to 8192

## Expected Behavior

`context_window` in agent.toml should override the catalog lookup, OR the catalog lookup should work correctly with `provider_urls` overrides.

## Actual Behavior

- `context_window` in agent.toml is ignored
- Runtime falls back to 8192
- Warning persists despite the field being set
- Agent becomes very slow due to aggressive message trimming at 8192 tokens

## Environment

- LibreFang version: 2026.7.21
- OS: Debian 12
- Model: `deepseek-v4-flash` via Polza proxy (OpenAI-compatible)
- Provider override: `provider_urls.deepseek = "https://polza.ai/api/v1"`

## Workaround

None — the context window is hardcoded to 8192 when the catalog lookup fails.

## Theory

The catalog lookup might be checking `provider/model` format (e.g. `deepseek/deepseek-v4-flash`) but the catalog entry uses just `deepseek-v4-flash` under the `deepseek` provider. When `provider_urls` overrides the base URL, the catalog might not be re-indexed with the new provider configuration.

houko:
## Confirmed: the field was inert on every path that runs a turn

The runtime tells the operator exactly what to do:

```
Model not in catalog — falling back to conservative context window.
Set `model.context_window` in agent.toml to silence this warning.
```

No execution path read that field. All three resolved the window from the catalog alone and passed `None` on a miss, leaving the agent loop's `UNKNOWN_MODEL_CONTEXT_WINDOW` (8192) in place:

- `agent_execution.rs:1042` — `execute_llm_agent`, the non-streaming path.
- `messaging.rs:651` — the ephemeral `/btw` path.
- `messaging.rs:2433` — the streaming path.

The one place that *did* honour it was `agent_runtime.rs:539`, a read-only context report — and its comment claimed to mirror "the same precedence chain the agent loop uses". That claim is what let the divergence sit unnoticed: the dashboard showed the operator's value while every real turn ignored it and trimmed history at 8192. The comment is corrected in this PR.

## Changes

- **New `manifest_helpers::resolve_context_window(catalog, model, session_hint)`** — one implementation of the documented chain: `agent.toml` override (0 filtered out as "unset") → catalog lookup (provider-aware / prefix-reconciling per #6423, 0 filtered out so image/audio entries don't poison budget math) → persisted session value. Returns `None` when nothing resolves, so the caller keeps owning its own fallback (8192 in the agent loop, 200 000 in the compaction gate) and the existing warning still fires.
- **Five call sites converted**: the three execution paths, the context report, and the compaction gate at `agent_runtime.rs:362` (which was also catalog-only, so a manifest override never reached the compaction trigger either).

### Two ordering problems this surfaced

Both only start to matter once the override is live, so they belong here rather than in a follow-up:

1. **`apply_session_model_override_to_manifest` now clears `model.context_window`** when the override actually changes provider or model. That number annotates the *manifest's* model; carrying it onto a different model would size the budget from the wrong one. A no-op override (re-selecting the same model) and a rejected override both keep the value.
2. **The streaming path resolves the window after the session override is applied.** It previously read `entry.manifest`, i.e. the pre-override manifest, so a `/model` switch sized the budget from the manifest's original model. That was a latent #4898 gap; moving the resolution below the override fixes it and is required for point 1 to take effect on that path.

## Scope — what this does and does not settle

The report says the model *is* in the catalog (`librefang models list --provider deepseek` shows it with a 1M window) while the runtime says it is not. I did not verify that half. The plausible explanation is that `models list` can surface live provider-advertised models that the static catalog files don't carry, in which case `find_model_for_manifest` legitimately misses — but I have not confirmed it, and whether a `provider_urls` override should also feed catalog indexing is a separate question worth its own issue.

What this PR settles is that the documented escape hatch works: setting `context_window` in `agent.toml` now sizes the real context budget, silences nothing spuriously, and is respected identically by the dashboard report, the compaction gate, and all three turn paths.

## Verification

- `cargo test -p librefang-kernel --lib` — **1321 passed, 0 failed** (with `--test-threads=1`), including 8 new `manifest_helpers::context_window_tests`: override wins for an unknown model (the reported case), override beats the catalog, catalog used without an override, `0` override ignored, `0` catalog window falling through to the session hint, session hint ranked last, `0` session hint ignored, and nothing-resolves returning `None`.
  - Note on the default parallel run: `cargo test -p librefang-kernel --lib` at full thread count aborts with SIGABRT on this macOS box with **no test failure reported**, and the same binary passes 1321/1321 single-threaded. That looks like local resource contention in the suite (many tests each opening a SQLite-backed temp substrate), not something this change touches — the diff adds a pure function and swaps five lookups. CI runs these under nextest's process-per-test isolation. Flagging it rather than burying it.
- `cargo test -p librefang-runtime --lib agent_loop::model` — 13 passed, including 4 new cases for the override clearing / preserving `context_window`.
- `cargo clippy -p librefang-kernel -p librefang-runtime --all-targets -- -D warnings` — clean.

No integration test: no route shape changed. The behaviour is kernel-internal resolution, covered by the unit tests above at the point where the decision is made.

Fixes #6568


crates/librefang-kernel/src/kernel/agent_runtime.rs
```diff
@@ -352,24 +352,21 @@ impl LibreFangKernel {
...
+        // Resolve the agent's actual context window BEFORE the gate so the
+        // gate's token-trigger comparison uses the real per-agent window (e.g.
+        // 128K GPT-4o, 1M Gemini) instead of the global 200K default that
+        // `CompactionConfig::from_toml_…` would otherwise hand us.
+        //
+        // Same precedence as every other consumer (#6568): agent.toml override,
+        // then the catalog, then the persisted session value. `resolve_context_window`
+        // filters 0 out so image/audio entries (no context window) fall back to
+        // the 200K default instead of feeding 0 into compaction math.
+        let agent_ctx_window = super::manifest_helpers::resolve_context_window(
```
houko, line 372: This call (and the `context_report_for_session` one below, around what's now line 542) resolves the window from `entry.manifest.model`, not from a manifest with `session.model_override` applied.

The three turn-execution paths this PR fixes (`execute_llm_agent`, the streaming turn, and now the read-only report's stated intent) all clone the manifest and call `apply_session_model_override_to_manifest` before resolving the context window, specifically so a `/model` switch mid-session sizes the budget from the model actually in use. This function already has `session` in scope (fetched a few lines up), so a session running under an active `/model` override will have its compaction trigger threshold sized from the pre-override model's window instead of the one that's actually generating the turn — the same class of bug this PR fixes elsewhere, just left unaddressed at this call site (and at `context_report_for_session`, which feeds the dashboard's context-usage display).

Is this intentionally out of scope (e.g. because compaction/reporting are considered "manifest-level" concerns rather than per-turn ones), or worth applying the same clone+override pattern here for consistency with the other four call sites?

---

houko (commented):
Automated review pass against CLAUDE.md conventions. One design question left inline; no mechanical rule violations found in the diff itself worth a direct fix.

---

PR #6581, opened 2026-07-26
fix(skills): install and search from the synced registry checkout

### LINKED ISSUE #6569: FangHub marketplace broken: skill search/install returns 404/422
## Description

The FangHub marketplace is non-functional in v2026.7.21. All skill/plugin install methods fail.

### CLI
```bash
librefang skill search "code"        # 422 Unprocessable Entity
librefang skill install web-search   # 404 Not Found
librefang skill install https://github.com/user/repo.git  # Wrong URL construction
```

### Dashboard API
```bash
GET /api/plugins                      # {"items":[],"total":0}
POST /api/plugins/install             # "Plugin not found in registry (404)"
```

### Local registry
```bash
librefang skill install ~/.librefang/registry/skills/code-reviewer
# "No skill.toml found" — local skills use SKILL.md format, not skill.toml
```

## Expected Behavior

- `librefang skill search` returns available skills from FangHub
- `librefang skill install <name>` installs a skill
- Dashboard shows installable plugins
- Local skills can be installed without `skill.toml`

## Actual Behavior

- CLI: 422 on search, 404 on install
- Dashboard: empty plugin list, 404 on install
- Local: requires `skill.toml` which doesn't exist in `registry/skills/`
- Git URL install: constructs wrong URL (`librefang-skills/<url>`)

## Environment

- LibreFang version: 2026.7.21
- OS: Debian 12
- `registry_host` in config.toml: `https://codeberg.org`

## Notes

The 62 skills in `~/.librefang/registry/skills/` use `SKILL.md` format. They can be manually copied to agent workspaces but cannot be installed via CLI/dashboard. The marketplace at librefang.ai appears to be a SPA with no public API for installation.

houko:
## Root cause: the marketplace pointed at an organization that does not exist

```
$ gh api /orgs/librefang-skills
{"message":"Not Found","status":"404"}
```

`MarketplaceConfig::default()` hardcodes `github_org: "librefang-skills"`, and both remote paths build on it. That single fact explains both reported status codes precisely:

- **`search` → 422.** It requests `/search/repositories?q=<query>+org:librefang-skills`. GitHub answers `422 Unprocessable Entity` for an `org:` qualifier naming a missing org — not an empty result set.
- **`install <name>` → 404.** It requests `/repos/librefang-skills/<name>/releases/latest`.

Meanwhile the skills were on disk the whole time. `registry_sync` clones `librefang/librefang-registry` into `~/.librefang/registry` and honours `registry.registry_host`, so the reporter's Codeberg configuration works:

```
$ gh api /repos/librefang/librefang-registry/contents/skills --jq '[.[]|select(.type=="dir")]|length'
61
$ gh api /repos/librefang/librefang-registry/contents/skills/code-reviewer --jq '.[].name'
SKILL.md
```

`web-search` — the exact skill in the report — is one of the 61. And two endpoints already read that checkout correctly: `POST /api/skills/install` copies from it, and `GET /api/skills/registry` enumerates it including SKILL.md frontmatter. The CLI and `GET /api/marketplace/search` were the ones that didn't.

## Changes

**`crates/librefang-skills/src/marketplace.rs`**
- `MarketplaceConfig::registry_dir` — the synced checkout, injected by the caller. (`librefang-skills` can't call the kernel's `librefang_home()` without inverting the dependency, so the CLI/API pass it in.)
- `search_registry(query)` — scans `<registry_dir>/skills/`, matching name and description case-insensitively, reading **both** on-disk shapes (`SKILL.md` frontmatter and `skill.toml`), sorted so output doesn't depend on `read_dir` order. An unsynced registry returns an error naming the missing path rather than masquerading as an empty catalog.
- `install(name, target)` — resolves the checkout first, then falls back to GitHub releases. The registry leg copies the directory, normalizes the manifest through the existing `ensure_skill_manifest` (so `SKILL.md` becomes a native `skill.toml`), and runs the **same supply-chain audit** as a remote bundle — a checkout is still third-party content pulled off a forge.
- `install_from_git(url, target)` — shallow-clones into a staging sibling, drops `.git`, normalizes, audits, and cleans up the staging dir on every exit path so a failed clone leaves nothing behind.
- `looks_like_git_url` / `skill_name_from_git_url` — remote detection and directory naming, the latter validated through the existing `is_safe_component` so a crafted URL can't traverse.
- The copy helper skips symlinks: a link inside a checkout would otherwise be reproduced pointing outside the installed skill, where the audit can't reason about it.

The GitHub-org `search` and the releases leg of `install` are kept for a deployment that genuinely publishes one repo per skill under a configured org.

**`crates/librefang-cli/src/commands/skill.rs`**
- `skill search` uses `search_registry`. Registry rows print the ready-to-run `librefang skill install <id>` line instead of a meaningless `(0)` star count.
- `skill install ./dir` accepts a `SKILL.md`-only directory. The registry loader already auto-converts that shape (`registry.rs: load_all`), but the CLI checked only `skill.toml` then OpenClaw's `package.json`, so it rejected every registry skill with "No skill.toml found" — the reporter's third symptom.
- `skill install <git-url>` routes to `install_from_git`. `cli.rs`'s `Install` `long_about` has always advertised `librefang skill install https://github.com/user/skill.git`; nothing implemented it, so the URL fell through to the name-based install and was pasted into `{org}/{name}` — the reporter's "wrong URL construction".

**`crates/librefang-api/src/routes/skills/skill.rs`**
- `GET /api/marketplace/search` reads `SKILL.md` as well as `skill.toml`, sorts its rows, and returns `install_id` (the directory name, which can differ from the frontmatter name and is what the install endpoints take). Since the registry ships SKILL.md exclusively, this endpoint previously returned an empty list on every install.

## On the `/api/plugins` half of the report

> `GET /api/plugins` → `{"items":[],"total":0}`, `POST /api/plugins/install` → "Plugin not found in registry (404)"

Both are correct behaviour, not bugs, so nothing changed there:

- `GET /api/plugins` lists **installed** plugins. Empty is the right answer on a fresh install.
- `/api/plugins/install` resolves against the registry's `plugins/` directory, which holds 9 entries (`auto-summarizer`, `context-decay`, `conversation-logger`, `episodic-memory`, `guardrails`, `keyword-memory`, `mempalace-indexer`, `sentiment-tracker`, `todo-tracker`). `web-search` is not among them — it is a *skill*, and after this PR `librefang skill install web-search` is the working command for it.

## Verification

- `cargo test -p librefang-skills --lib` — 235 passed. 12 new marketplace cases: registry listing of SKILL.md entries, name/description/case-insensitive matching, `skill.toml` entries, skipping manifest-less directories, unsynced-registry error, unconfigured-dir error, install preferring the checkout **and** producing a converted `skill.toml` plus `source_kind: "registry"` metadata, traversal name rejection, the two git-URL helpers, and symlink skipping.
- `cargo test -p librefang-api --test skills_routes_test` — 21 passed, 4 new `#[tokio::test]` cases against `TestServer` for `GET /api/marketplace/search`: SKILL.md discovery with `install_id`, description matching, a genuine miss returning zero rows, and sorted output.
- `cargo clippy -p librefang-skills -p librefang-api -p librefang-cli --all-targets -- -D warnings` — clean.

Not covered by automated tests: the `install_from_git` clone itself shells out to `git` against a live remote, so its happy path is unverified here — the URL parsing, naming, and safety checks around it are unit-tested, and a failed clone is surfaced with git's stderr. Worth one manual `librefang skill install <some-repo>.git` from someone set up for it.

Fixes #6569


crates/librefang-api/src/routes/skills/skill.rs
```diff
@@ -744,34 +744,78 @@ pub async fn marketplace_search(
...
+    });
     let total = results.len();
     Json(serde_json::json!({"results": results, "total": total}))
 }
 
+/// Read `(name, description)` for one registry skill directory.
+///
+/// Handles both on-disk shapes — `SKILL.md` with YAML frontmatter (what the registry ships) and the native `skill.toml` — and falls back to the directory name when the metadata is present but unparsable.
+/// Returns `None` for a directory holding neither, so unrelated directories are skipped.
+fn read_registry_skill_entry(dir: &std::path::Path, dir_name: &str) -> Option<(String, String)> {
```
houko, line 790: `read_registry_skill_entry` here duplicates `read_registry_skill_meta` in `crates/librefang-skills/src/marketplace.rs` (private fn, same file) almost line-for-line: both check `SKILL.md` first via a YAML-frontmatter parse, fall back to `skill.toml`, and fall back further to the bare directory name. Since this PR already touches both `librefang-skills` and `librefang-api`, consider exposing the `librefang-skills` version as `pub` and having this route call it directly instead of re-implementing the same on-disk-shape logic — it removes a place where the two can silently drift (e.g. one gets a bugfix, the other doesn't), and `librefang-api` already depends on `librefang_skills` elsewhere in this file. Not blocking, just flagging the duplication since it's new code in this same PR.

---

crates/librefang-skills/src/marketplace.rs
```diff
@@ -101,19 +101,131 @@ fn urlencoded(s: &str) -> String {
...
+    } else {
+        None
+    }
+}
+
+/// Name + description read from a registry skill directory.
+///
+/// Accepts both on-disk shapes: `SKILL.md` with YAML frontmatter (what the registry actually ships — 61 of 61 entries at the time of writing) and the native `skill.toml`.
+/// Returns `None` for a directory holding neither.
+fn read_registry_skill_meta(dir: &Path) -> Option<(String, String)> {
```
houko, line 201: `read_registry_skill_meta` (and the API route's equivalent `read_registry_skill_entry`) checks `SKILL.md` **before** `skill.toml` when both are present in a directory. But `ensure_skill_manifest` (line ~1020 in this file, the function the actual install path calls) checks `skill.toml` **first** and returns early without touching `SKILL.md` if it exists. So for a hypothetical registry directory that ships both files, `librefang skill search` / `GET /api/marketplace/search` would display the `SKILL.md` frontmatter's name/description, but `librefang skill install` would install using the `skill.toml` content instead — a silent mismatch between what search shows and what install actually installs. Today this is likely moot (the PR description says all 61 registry skills are one shape or the other, not both), but it's a latent inconsistency worth either documenting as an assumption or aligning the precedence order between the two code paths.

---

crates/librefang-skills/src/marketplace.rs
```diff
@@ -295,6 +479,146 @@ impl MarketplaceClient {
...
+        target_dir: &Path,
+    ) -> Result<String, SkillError> {
+        let skill_dir = resolve_skill_dir(target_dir, skill_name)?;
+        if skill_dir.exists() {
+            std::fs::remove_dir_all(&skill_dir)?;
+        }
+        std::fs::create_dir_all(&skill_dir)?;
+
+        copy_dir_recursive(source, &skill_dir)?;
+        ensure_skill_manifest(&skill_dir)?;
```
houko, line 500: In both new install paths (`install_from_registry_dir` here, and `install_from_git` at the equivalent `ensure_skill_manifest(&skill_dir)?;` call further down), if `ensure_skill_manifest` returns `Err` (no loadable `SKILL.md`/`skill.toml`/OpenClaw shape found), the `?` bails out immediately and `skill_dir` — already created and populated by `copy_dir_recursive` — is left on disk. Only the `supply_chain::scan` failure branch right below explicitly does `remove_dir_all(&skill_dir)`. This mirrors a pre-existing gap in the older GitHub-releases `install()` above (same pattern there too), but this PR is the one propagating it into two new code paths rather than fixing it. Low severity — the debris gets cleared out by the `if skill_dir.exists() { remove_dir_all }` at the top of the next install attempt for the same name — but worth a `let _ = std::fs::remove_dir_all(&skill_dir);` before the `?` (or restructuring with the same closure/cleanup pattern used for the scan failure) so a failed install doesn't leave a manifest-less directory sitting in `~/.librefang/skills/` until someone retries.

---

PR #3197, opened 2026-04-26
fix(desktop): wrap with wrapGAppsHook3 so tray icon resolves on NixOS

houko:
## Summary
- Tray icon never appeared in `librefang-desktop` on NixOS, while the AppImage build worked on the same machine (#3192).
- Root cause: `tray-icon` → `libappindicator-sys` 0.9.0 `dlopen`s `libayatana-appindicator3.so.1` at runtime with no `DT_NEEDED` entry. The previous fix (#3052) added the lib dir via `patchelf --add-rpath`, but modern patchelf writes `DT_RUNPATH`; glibc's `ld.so` only consults `DT_RUNPATH` for `DT_NEEDED` deps — **never for `dlopen()` string-name lookups** — so the RPATH was inert and the tray plugin silently failed to load.
- Fix: switch the desktop derivation to `wrapGAppsHook3` and inject `pkgs.libayatana-appindicator` on `LD_LIBRARY_PATH` via `gappsWrapperArgs`. As a bonus this also wires the GTK runtime env (`XDG_DATA_DIRS`, `GIO_MODULE_DIR`, `GSETTINGS_SCHEMA_DIR`) that the webview wants.

User confirmed in the issue that adding `libayatana-appindicator` to `environment.systemPackages` doesn't help — that's expected, NixOS has no global `ld.so.conf`, the `dlopen` simple-name lookup never sees system packages.

## References
- crate dlopen path: https://github.com/tauri-apps/libappindicator-rs/blob/main/sys/src/lib.rs
- `ld.so(8)` on `DT_RUNPATH`: *"Such directories are searched only to find those objects required by `DT_NEEDED` entries"*
- Same-symptom fix in a downstream fork (different codebase, same root cause): RightNow-AI/openfang#1063

## Test plan
- [ ] `nix build .#librefang-desktop` succeeds on a Linux runner (CI).
- [ ] On NixOS with the resulting build, launch `librefang-desktop` and verify the tray icon appears (right-click menu shows Show Window / Quit / etc).
- [ ] Verify the wrapper exports the lib dir: `cat $(nix path-info .#librefang-desktop)/bin/.librefang-desktop-wrapped` should contain `LD_LIBRARY_PATH=…libayatana-appindicator…`.
- [ ] Verify CLI build still green: `nix build .#librefang-cli` (we did not touch that scope).
- [ ] Smoke-test on a non-NixOS Linux distro (Ubuntu) to confirm `wrapGAppsHook3` doesn't regress AppImage / cargo build paths.

[...]

Issue #3156, opened 2026-04-25
Nix Flake does not set up LibreFang .desktop entry and icon


FrantaNautilus:
### Description

Nix Flake does not set up LibreFang .desktop entry and icon, therefore the `librefang-desktop` cannot be started from app menu.

### Expected Behavior

The `librefang-desktop` is a desktop app and therefore is expected to provide .desktop entry.

### Steps to Reproduce

1. Install `librefang-desktop` using flake.
2. The LibreFang is missing from app menu

### LibreFang Version

v2026.4.24-beta5

### Operating System

Linux (x86_64)

### Logs / Screenshots

Does not apply.

houko:
Fixed by #3157 (merged 2026-04-25). The Nix flake now installs the `.desktop` entry + 128×128 icon and runs `update-desktop-database` so LibreFang shows up in the application menu after `nix profile install`.

Issue #3052, opened 2026-04-24
Librefang-desktop creates empty window (AppImage and NixOS)


FrantaNautilus:
### Description

The Librefang desktop app currently does not open a functional window upon starting. When started `librefang-desktop` produces an blank white window. When started from terminal `librefang-desktop` does not produce any errors.

This behavior was reproduced with multiple configurations:
- AppImage  build on computer running  BluefinOS (based on Fedora SilverBlue 43, Gnome 49)
- AppImage build on NixOS 25.11 with KDE Plasma 6.5.5
- NixOS build on NixOS unstable (26.05) with Gnome 49 (see details below)

Similar problems can be tackled using different acceleration and rendering settings via environment variables. I tried to apply different settings of `WEBKIT_USE_GL_STACK`  and `GSK_RENDERER`  variables, but to no success.

On the NixOS I could not test the `librefang desktop` since the build is again failing due to the dlopen not finding `libayatana-appindicator`. This is presumably caused by recent commit that has removed `libayatana-appindicator` from `desktopBuildInputs` and the `rpath` fix which was proposed by @houko in PR [https://github.com/librefang/librefang/pull/2974](https://github.com/librefang/librefang/pull/2974).
For this reason the `librefang-desktop` I tested had to be locally modified to build.

The AppImages were tested in three scenarios: start from command line, start from `librefang-cli`, start using Gearlever. In all cases the result is the same, blank window, see attached screenshot.

### Expected Behavior

The Librefang desktop app is expected to produce a functional and readable window.

### Steps to Reproduce

1. Run `librefang-desktop` (assuming the AppImage is installed)
2. Observe blank window.

### LibreFang Version

v2026.4.23-beta4

### Operating System

Linux (x86_64)

### Logs / Screenshots

Blank window screenshot:
<img width="2880" height="1800" alt="Image" src="https://github.com/user-attachments/assets/64925a61-b8bd-490d-b3b9-2bd60bb11e4e" />

FrantaNautilus:
The NixOS build suffers from this issue also on version v2026.4.24-beta5. BUt is produces some error output now.

```
$ ./result/bin/librefang-desktop 
2026-04-24T13:23:31.421920Z  INFO librefang_desktop: Starting LibreFang Desktop...
2026-04-24T13:23:31.643787Z  INFO librefang_desktop: LibreFang Desktop window created

[...]

PR #2974, opened 2026-04-22
fix(flake.nix): resolve 2937, build of both librefang-cli and librefang-desktop on NixOS

FrantaNautilus:
## Type

- [ ] Agent template (TOML)
- [ ] Skill (Python/JS/Prompt)
- [ ] Channel adapter
- [ ] LLM provider
- [ ] Built-in tool
- [ ] Bug fix
- [ ] Feature (Rust)
- [ ] Documentation / Translation
- [ ] Refactor / Performance
- [ ] CI / Tooling
- [x] Other

## Summary

Fixes #2937
The PR fixes the directory organization in flake.nix for successful build of librefang-cli and librefang-desktop on NixOS

## Changes

Changed flake.nix
```
-        # Filter source to only include Rust-relevant files
-        src = craneLib.cleanCargoSource ./.;
+               # Filter source to only include Rust-relevant files
+        # Include locale/files needed for i18n compile-time embedding
+        cleanSrc = craneLib.cleanCargoSource ./.;
+        src = pkgs.runCommand "librefang-src" {} ''
+          cp -a ${cleanSrc} $out
+          chmod -R u+w $out
+          # Restore locale dirs stripped by cleanCargoSource
+          rm -rf $out/crates/librefang-types/locales
+          cp -a ${./.}/crates/librefang-types/locales $out/crates/librefang-types/
+          # Restore packages dir stripped by cleanCargoSource
+          rm -rf $out/packages
+          cp -a ${./.}/packages $out/packages
+          # Restore static dir (locales, logo, favicon) stripped by cleanCargoSource
+          rm -rf $out/crates/librefang-api/static
+          cp -a ${./.}/crates/librefang-api/static $out/crates/librefang-api/
+          # Restore login_page.html stripped by cleanCargoSource
+          mkdir -p $out/crates/librefang-api/src
+          cp -a ${./.}/crates/librefang-api/src/login_page.html $out/crates/librefang-api/src/
+          # Restore CLI templates dir
+          rm -rf $out/crates/librefang-cli/templates
+          cp -a ${./.}/crates/librefang-cli/templates $out/crates/librefang-cli/
+          # Restore CLI locales dir
+          rm -rf $out/crates/librefang-cli/locales
+          cp -a ${./.}/crates/librefang-cli/locales $out/crates/librefang-cli/
+          # Restore Tauri desktop files (config, capabilities, icons)
+          rm -rf $out/crates/librefang-desktop/tauri.conf.json
[...]

PR #6582, opened 2026-07-27
feat(nix): first-class NixOS deployment and deepin/Debian distro awareness

houko:
## Why

NixOS was a package-only target. The flake produced binaries, but there was no declarative way to run the daemon — a NixOS operator had to hand-write a systemd unit, and to get it right they first had to discover that the obvious `ExecStart=librefang start` forks. The Nix path was also never gated before merge: `nix-build.yml` runs on push-to-main only, and its own header comment records five breakages (#2937, #2974, #3052, #3156, #3197) that "slipped through because regular CI doesn't exercise the Nix path".

deepin had no representation in the repo at all — a full-tree grep for it returns only the unrelated `deepinfra` LLM provider.

## Flake outputs

- `nixosModules.default` / `nixosModules.librefang`, backed by the new `nix/nixos-module.nix`, exposing `services.librefang` with `port`, `openFirewall`, `user`, `group`, `stateDir`, `environmentFile` and `extraEnvironment`.
- `overlays.default`, so nixpkgs consumers can pull `librefang-cli` / `librefang-desktop` into their own package set. It reads the system from `prev` rather than `final`, so the nixpkgs fixed point does not become self-referential.
- Both sit **outside** `flake-utils.lib.eachDefaultSystem`. A nixosModule nested per-system lands at `nixosModules.<system>.default`, which `nixos-rebuild` and `lib.nixosSystem` never read — the flake evaluates fine and the module is silently unusable.

## The unit

`ExecStart` is `librefang start --foreground`, and the flag is load-bearing. The bare `librefang start` takes the `!spawned && !foreground` branch at `crates/librefang-cli/src/commands/daemon.rs:347` into `spawn_detached_daemon`, which re-execs with `--spawned`, calls `libc::setsid()` in `pre_exec` (`daemon.rs:110-118`), then returns from the parent after a health poll. systemd would see the main process exit immediately and tear the setsid'd child down with it. `Type = "exec"` rather than `notify`, because `grep -rn "sd_notify\|NOTIFY_SOCKET" crates/` returns nothing.

`port` reaches the daemon as `LIBREFANG_LISTEN=127.0.0.1:<port>`, which `Kernel::boot_with_config` applies over `config.api_listen` (`crates/librefang-kernel/src/kernel/boot.rs:47-49`). The module deliberately does **not** generate `config.toml`, because the daemon and the dashboard write that file themselves — the MCP migrator rewrites it, and `POST /api/config/set`, the budget `PUT` and the `[default_model]` writers all go through `crate::atomic_write`. A read-only store path would break all of them.

`pkgs.git` is on the unit `PATH`, because `init_git_if_missing` spawns `git` by bare name on every boot (`crates/librefang-kernel/src/kernel/workspace_setup.rs:203-207`). `RestrictAddressFamilies` allows `AF_UNIX` alongside the two inet families, because the API binds a unix socket for ACP (`crates/librefang-api/src/acp_uds.rs:119-141`). `MemoryDenyWriteExecute` is left `false`, mirroring `deploy/librefang.service:31` — the WASM sandbox needs W+X.

Assertions fire at evaluation time for the failure modes that would otherwise surface long after `nixos-rebuild` reported success:

- an `environmentFile` inside `builtins.storeDir`, which would make provider keys world-readable
- a `stateDir` under `/home`, `/root` or `/run/user`, which `ProtectHome = true` makes inaccessible and empty
- a `stateDir` with an empty final component (systemd refuses to load the unit) or a shared FHS parent (the tmpfiles rule would chown it)
- an `extraEnvironment.LIBREFANG_HOME` that would desynchronise from `StateDirectory` / `ReadWritePaths`
- `openFirewall` with a non-loopback bind and no auth source, mirroring `check_bind_auth_safety` (`crates/librefang-api/src/server.rs:312-330`)

## CI

`checks` gains `librefang-desktop` on Linux — it was previously reachable only through `packages`, so a `wrapGAppsHook3` or desktop-item regression passed `nix flake check` — and `nixos-module-eval`, which instantiates a throwaway `lib.nixosSystem` and asserts thirteen properties of the generated unit. That derivation deliberately holds no reference to the rendered unit text, because that text embeds `${librefang-cli}/bin/librefang` and depending on it would turn the PR eval into a full cold compile.

`nix-build.yml` gains a `nix-eval` job on `pull_request` that runs `nix flake check --no-build` plus two assertion steps. It instantiates derivations and never compiles one, so it costs minutes rather than the 80-95 documented for a cold matrix leg. The expensive matrix stays gated behind `github.event_name != 'pull_request'`.

## Installer

`detect_distro` reads `/etc/os-release` (`ID` and `ID_LIKE`) with `/etc/NIXOS` as the authoritative NixOS marker, and degrades silently to `unknown` where neither exists — it must never fail the install on macOS or in a minimal container.

On NixOS the glibc fallback is now suppressed. That binary can never execute there for want of an ELF interpreter at the FHS path it hardcodes, so offering it as a fallback downloaded ~60 MB, failed the post-install `--version` check, rolled back, and printed a generic "The new binary failed to run" that told a NixOS user nothing. The suppression is expressed as a single `effective_platform_fallback` helper consulted at each point of use, rather than by mutating `PLATFORM_FALLBACK`, so the outcome cannot change with the order `detect_distro` and `detect_platform` run in.

Debian-family hosts get an apt dependency hint gated on a **runtime probe** — pkg-config for `webkit2gtk-4.1` then `webkit2gtk-4.0` — rather than a hardcoded claim that any given distro ships a given package.

## Doctor

A Linux-only soft check probes pkg-config for the desktop webview stack and maps `/etc/os-release` to a per-family remediation hint (apt / pacman / dnf / nix / generic). It deliberately suggests a package **search** rather than concrete package names, because the package shipping a given pkg-config module differs between distributions and between releases of one distribution.

It can never fail the command: `Error` is the only severity that clears `all_ok` (`crates/librefang-cli/src/commands/doctor_cmd.rs:1221`), missing libraries are `Warn`, and an unexecutable pkg-config is `Info`. A CLI-only install is a fully supported configuration.

## Docs

New `docs/operations/nixos.md`, linked from the README NixOS section. New README and docs-site sections for NixOS and Debian / Ubuntu / deepin in both English and Chinese. `docs/src/mdx/rehype.mjs` learns the `nix` shiki language so the module snippets highlight instead of rendering as plain text.

The deepin desktop story is documented **honestly**: whether deepin's repositories carry `libwebkit2gtk-4.1` rather than `4.0` is not something this project can verify, so the docs say so and point at `librefang doctor` to check the local machine. No claim is made in either direction.

## Verification

Run locally, all green:

- `cargo check --workspace --lib` — exit 0
- `cargo clippy -p librefang-cli --all-targets -- -D warnings` — exit 0
- `cargo test -p librefang-cli` — all pass, including the three i18n gates `test_no_untranslated_strings`, `test_no_dead_locale_keys` and `test_locales_cover_used_i18n_keys`
- `cargo fmt --check` — exit 0
- `sh scripts/tests/install_sh_test.sh` — all pass, including four new NixOS / distro cases
- `python3 scripts/check-changelog-attribution.py --all-unreleased` — OK

Nix evaluation was verified with a real `nix` 2.34.1:

- `nix eval --json '.#nixosModules' --apply builtins.attrNames` → `["default","librefang"]` (module names, not system names — confirms top-level placement)
- `nix eval --json '.#overlays' --apply builtins.attrNames` → `["default"]`
- `nix eval --json '.#checks.x86_64-linux' --apply builtins.attrNames` → includes `librefang-desktop` and `nixos-module-eval`
- `nix eval --json '.#checks.aarch64-darwin' --apply builtins.attrNames` → `["librefang-cli","librefang-clippy","librefang-fmt"]`, confirming the Linux-only additions are correctly gated
- `nix eval --raw '.#checks.x86_64-linux.nixos-module-eval.drvPath'` → instantiates, so all thirteen unit assertions hold

## Not verified

**Nothing has booted the service.** Evaluation asserts the generated unit's shape; it cannot prove the daemon survives being started by systemd.

There is now a check that closes this, added in the second commit: `checks.<system>.nixos-vm-test` uses `pkgs.testers.runNixOSTest` to boot a NixOS guest with `services.librefang.enable = true`, wait for the unit, wait for port 4545, request `/api/health`, and confirm `StateDirectory` created the state dir with a non-zero `MainPID`. Run it on a Linux host with KVM:

```
nix build .#checks.x86_64-linux.nixos-vm-test -L
```

**I have not run it** — this is a macOS host, and a NixOS VM test needs a Linux kernel. What I did verify is that it instantiates and that having it in `checks` costs the PR lane nothing: `nix flake check --no-build --all-systems` reports `checking derivation checks.x86_64-linux.nixos-vm-test...` and `checks.aarch64-linux.nixos-vm-test...`, performs **zero** builds, and finishes in 33s. So a module rename that broke the test expression fails the existing lane, while no lane ever compiles the VM. The darwin check sets correctly omit it.

`librefang-desktop` on NixOS remains the rough path — it is now in `checks` so a regression is caught, but the GTK closure and the appindicator `dlopen` are unchanged by this PR.

`librefang-desktop` on NixOS remains the rough path — it is now in `checks` so a regression is caught, but the GTK closure and the appindicator `dlopen` are unchanged by this PR.

## For maintainer decision

1. **Does the Nix lane get to gate PRs at all?** `nix-build.yml:3-10` records a deliberate decision *not* to run on pull requests, for runner-minute reasons. The `nix-eval` job added here is designed around that reason rather than against it — evaluation only, minutes not hours, and not wired into any required-checks gate — but the call is yours. Reverting just that job leaves everything else in this PR intact.

2. ~~**`librefang service install` on macOS writes a LaunchAgent, not a LaunchDaemon.**~~ **Resolved separately in #6584.** A LaunchAgent only runs after a user logs in, so a Mac that reboots to the login window never starts the daemon — the same always-on gap this PR closes for NixOS. Fixing it touches a different domain (service installation, not dependency diagnostics or Nix packaging), so it is still deliberately **not** bundled here; #6584 adds `service install --system` with its own tests. Nothing in this PR depends on it, and the two can be reviewed and merged in either order.

3. **Should `nixos-vm-test` get a CI lane, or stay opt-in?** The check itself is in this PR (second commit) — that part is not a question. What is a question is whether it should ever run in CI. Building it compiles `librefang-cli` and boots a VM, so it cannot join the `nix-eval` lane that finishes in 43s by never compiling anything; it would have to go on the push-to-main matrix, which `nix-build.yml:34-36` already documents as 80-95 minutes cold against the 5 GB cache cap at `:68`. It also needs the runner to nest KVM for a NixOS guest, which nothing in this repository establishes — so no lane here claims it. As shipped, CI instantiates the expression on every PR (verified: zero builds, 33s) and never runs the VM, and running it for real is one command on a Linux box. If you want a lane, it is a self-contained follow-up; if you would rather it stay opt-in, nothing needs to change.


web/public/install.sh
```diff
@@ -38,6 +46,40 @@ is_enabled() {
...
+# --- Distro detection -----------------------------------------------------
+
+# Print the unquoted, lowercased value of the $1 field of the os-release file, or nothing when the file or the field is absent.
+# Parsed rather than sourced: os-release is shell-syntax but sourcing it would execute a root-owned file inside the installer and would trip `set -u` on any field the caller does not read.
+os_release_field() {
+    [ -r "$OS_RELEASE_FILE" ] || return 0
+    grep -E "^$1=" "$OS_RELEASE_FILE" 2>/dev/null \
+        | head -n 1 \
+        | cut -d '=' -f 2- \
+        | tr -d "\"'" \
```
houko, line 58: `tr -d "\"'"` deletes every quote character in the value, not just a matching wrapping pair — a value like `PRETTY_NAME="Bob's Distro"` would come out as `Bobs Distro` (the internal apostrophe gets eaten too, not just the surrounding double quotes).

`crates/librefang-cli/src/doctor.rs`'s `unquote_os_release_value` in this same PR does the equivalent parse correctly — it strips one matching layer of `"`/`'` and falls through unchanged otherwise. Worth aligning the shell implementation with that, or leaving a comment here explaining why the simpler (but lossier) approach is acceptable for the fields this script actually reads (`ID`, `ID_LIKE`, which the os-release spec says never contain quotes internally). Not filing this as a definite bug since `DISTRO`/`DISTRO_LIKE` don't hit this path in practice today, but flagging the inconsistency between the two parsers for a decision either way.

---

nix/nixos-module.nix
```diff
@@ -0,0 +1,305 @@
...
+        message = ''
+          services.librefang binds the non-loopback address ${effectiveListen} with no authentication source configured, and the daemon refuses to start in that configuration (crates/librefang-api/src/server.rs:1915 calling `check_bind_auth_safety`, crates/librefang-api/src/server.rs:312-330).
+          Set services.librefang.environmentFile to a file exporting an API key, or accept the risk explicitly with extraEnvironment.LIBREFANG_ALLOW_NO_AUTH = "1".
+        '';
+      }
+      {
+        # Both consumers of `stateDir` need a dedicated directory with a real final component, and neither failure surfaces during evaluation.
+        # A trailing slash leaves `StateDirectory=` empty and systemd rejects the unit at load time — long after `nixos-rebuild` reported success — while a shared FHS parent would have the tmpfiles rule below chown that parent to the service user.
+        assertion = lib.last (lib.splitString "/" (toString cfg.stateDir)) != ""
+          && !(lib.elem (toString cfg.stateDir) [ "/var" "/var/lib" "/srv" "/opt" "/etc" "/usr" "/home" ]);
```
houko, line 164: The "shared FHS directory" exact-match list here (`/var`, `/var/lib`, `/srv`, `/opt`, `/etc`, `/usr`, `/home`) and the `ProtectHome` exact-match list a few lines up (`/root`, `/run/user`) between them cover every tree this unit treats specially — except `/tmp`. Setting `stateDir = "/tmp"` (or `/run`, `/dev`) passes every assertion here: it's not under `/var/lib` so it takes the `systemd.tmpfiles` + `ReadWritePaths` branch, and with `PrivateTmp = true` the unit's `/tmp` is already namespace-private, so the tmpfiles rule would `chown`/`chmod 0700` a directory with FHS-mandated semantics (`1777`, world-writable-sticky) inside that private namespace.

Probably a low-traffic edge case (nobody sets `stateDir` to a bare top-level directory on purpose), but since the other two special-cased trees each got their own assertion, wondering if `/tmp` was deliberately left out (e.g. because `PrivateTmp` genuinely makes it safe) or just not considered. Not fixing this myself since it's a design call about which paths deserve an evaluation-time assertion versus being left to the operator's judgment.

---

houko (commented):
Reviewed against CLAUDE.md's conventions (prose-wrapping, CHANGELOG attribution, config-field pattern, determinism rule) and re-ran the verification the PR description claims locally: `cargo check -p librefang-cli --bins`, `cargo clippy -p librefang-cli --all-targets -- -D warnings` (both clean), `cargo test -p librefang-cli` (268 + 8 tests pass, including the new `LinuxDesktopDepsCheck` cases and the three i18n gates across all four locales), and `sh scripts/tests/install_sh_test.sh` (all pass, including the new NixOS/distro cases). No CLAUDE.md-rule violations found — prose is already one-sentence-per-line throughout the new docs/comments, the CHANGELOG entry carries `(@houko)` attribution with no duplicate `[Unreleased]` header, and no new config fields need the struct+serde+Default pattern. Left two ambiguous, non-blocking observations inline for your judgment call rather than fixing them unilaterally.

---

nix/nixos-module.nix
```diff
@@ -0,0 +1,305 @@
...
+
+      # `init_git_if_missing` spawns `git` by bare name on every boot (crates/librefang-kernel/src/kernel/workspace_setup.rs:203-207, called from crates/librefang-kernel/src/kernel/boot.rs:847) to version-control the state dir.
+      # A systemd unit does not inherit the system profile's PATH, so git has to be declared here.
+      path = [ pkgs.git ];
+
+      environment = {
+        # `librefang_home()` prefers `LIBREFANG_HOME` over the user's home directory (crates/librefang-kernel/src/config.rs:535-542); without it the daemon would resolve `<home>/.librefang`.
+        LIBREFANG_HOME = cfg.stateDir;
+        # `dirs::home_dir()` reads `$HOME` before consulting the passwd entry, and the first-start init path exits 1 when it resolves to nothing (crates/librefang-cli/src/commands/init.rs:13-19).
+        HOME = cfg.stateDir;
```
houko, line 243: The assertion block a few lines up (`!(cfg.extraEnvironment ? LIBREFANG_HOME)`) exists specifically because letting `extraEnvironment` override a module-derived env var would "desynchronise it from StateDirectory / ReadWritePaths" — but that same protection isn't applied to the bare `HOME` set right here.

`environment = { LIBREFANG_HOME = cfg.stateDir; HOME = cfg.stateDir; LIBREFANG_LISTEN = listenAddress; } // cfg.extraEnvironment;` merges `cfg.extraEnvironment` last, so `services.librefang.extraEnvironment.HOME = "/something/else";` silently wins over this line with no assertion and no warning. Per the comment two lines above, `dirs::home_dir()` reads `$HOME` before the passwd entry, and `crates/librefang-cli/src/commands/init.rs:13-19` exits 1 when it resolves to nothing usable — so an operator who sets `extraEnvironment.HOME` (plausible copy-paste from an unrelated unit, or an attempt to fix some other tool's home lookup) desynchronises `HOME` from `LIBREFANG_HOME`/`StateDirectory` exactly the way the existing assertion prevents for `LIBREFANG_HOME` itself, just via the other variable that gates the same code path.

Worth either adding `!(cfg.extraEnvironment ? HOME)` to the existing assertion (or its own), or folding `HOME` into the same message as `LIBREFANG_HOME` since they protect the same invariant.

---

houko (commented):
One finding from today's automated pass. The rest of the module/module-eval/install.sh/doctor.rs logic checked out against the PR body's claims (verified against actual hunks, not just the description) — including the two points already raised in the existing review threads (install.sh quote-stripping, nixos-module.nix's /tmp gap in the stateDir assertion), which I did not duplicate.

---

Issue #983, opened 2026-03-17
[Bug] Custom model with non-default provider is incorrectly routed


houko:
---
name: Bug Report
about: Report a bug or unexpected behavior
labels: ["bug", "area/runtime"]
---

## Description
When adding a custom model assigned to a specific provider in custom_models.json, the model is not correctly routed to the specified provider. Instead, the UI displays an incorrect provider prefix, suggesting the routing logic is ignoring the provider field.

**Reference**: This is fixed in openfang#581.

## Expected Behavior
The model should be routed through the specified provider, and the UI should display the model as `provider:modelname` or simply `modelname`.

## Steps to Reproduce
1. Add custom model entry to custom_models.json:
```json
{
  "id": "MiniMax-M2.5",
  "provider": "qwen",
  "tier": "custom",
  ...
}
```
2. Switch to the custom model in the UI
3. Observe incorrect provider prefix displayed

## Additional Context
**Proposed fix**: Fix the routing logic to respect the provider field in custom model configuration.

PR #6583, opened 2026-07-27
feat(cli): add `librefang models connect everyapi` and an EveryAPI wiring doctor check

houko:
## What

Adds `librefang models connect everyapi [--set-default]`, a one-shot command that registers an [EveryAPI](https://github.com/everyapi-ai/everyapi) gateway account as a LibreFang LLM provider, plus a `doctor` check that reports whether such a gateway is present and which route is actually in effect.

EveryAPI is an AI API gateway that aggregates many upstream providers behind one OpenAI/Claude/Gemini-compatible endpoint. Its own CLI already has `everyapi use <tool>` for pointing Claude Code / Codex / Gemini CLI at the gateway, but that mechanism injects environment variables and execs a child process — which does nothing for a long-lived daemon whose HTTP drivers read `base_url` from config, never from the environment. So the wiring belongs on this side.

No changes to `librefang-llm-drivers` were needed. `create_driver()` already treats an unregistered provider that carries a `base_url` as an OpenAI-compatible endpoint, so the gateway is reachable purely through a synthesised provider catalog entry.

## Changes

- **`crates/librefang-cli/src/commands/everyapi.rs`** (new) — reads `~/.config/everyapi/credentials.json` (honouring `$XDG_CONFIG_HOME`), fetches `GET /v1/models`, synthesises a `ModelCatalogFile`, stores the relay key in `~/.librefang/.env` via `dotenv::save_env_key`, and persists the provider through `POST /api/registry/content/provider` when the daemon is up (which hot-reloads the catalog in-process) or by writing `~/.librefang/providers/everyapi.toml` directly when it is not.
- **`crates/librefang-cli/src/doctor.rs`** — new `EveryApiWiringCheck`.
- **`crates/librefang-cli/src/cli.rs`, `main.rs`, `commands/{mod,prelude}.rs`** — subcommand declaration, dispatch, module wiring.
- **`crates/librefang-cli/locales/{en,zh-CN,uk,ko}/main.ftl`** — 38 new keys per locale.
- **`crates/librefang-cli/tests/i18n_checks.rs`** — allowlist the new command name.

## Design notes

**The relay key is never printed, logged, or formatted into any message.** It goes from the credentials file into `.env` and nowhere else. The doctor check inspects it only for non-emptiness and drops the value.

**`base_url` is derived as `api_base` (trailing slash trimmed) + `/v1`.** `OpenAIDriver` builds `{base_url}/chat/completions`, so the version segment has to live in `base_url`; `AnthropicDriver` builds `{base_url}/v1/messages` and would need the opposite. Picking the OpenAI shape keeps this to pure configuration and avoids overriding `provider_urls["anthropic"]`, which would hijack the gateway onto every agent configured for the real Anthropic API.

**Synthesised entries are `ModelTier::Balanced`, not `Custom`.** `find_model` returns the first `Custom` match immediately so user-defined models beat builtins (#983), and `merge_catalog_file` dedupes on `(id, provider)` rather than id alone. Since several gateway ids collide with upstream ones (`claude-sonnet-5`, `claude-opus-5`, `gemini-3.5-flash`), tagging them `Custom` would have made every provider-blind lookup resolve to the gateway copy — silently re-pricing `catalog.pricing()` and re-answering `effective_capabilities_for()` for agents that never opted in. The gateway copy stays reachable through the provider-scoped `find_model_for_provider`, which is the path an agent configured for `everyapi` actually takes.

**A Text model with no snapshot metadata is skipped loudly rather than emitted with zeros.** The gateway publishes no max-output field and pricing for nothing, so `max_output_tokens` is resolvable only from the compiled-in OpenRouter snapshot. `ModelCatalogEntry::validate()` rejects a Text entry with a zero `context_window` or `max_output_tokens`, and `merge_catalog_file` then drops it with nothing but a `warn!` — which would have looked like a successful run that quietly lost most of the listing. Where pricing does not resolve, `pricing_known = false` is written explicitly, because that field defaults to `true` on deserialize and a bare `0.0` would assert the model is free and poison every budget figure derived from it.

**`--set-default` avoids the `openai-response`-only family.** Those models reject non-streaming requests, and `driver.complete()` is on the compactor, proactive-memory, skill-workshop and web-augment paths — pinning the default there would leave the main chat loop working while those four fail.

**Models are sorted by id before emission** so the generated TOML is byte-stable across runs (#3298).

**The doctor check's env scan deliberately does not reuse `is_proxied_via_env`.** That helper returns on the first variable merely *set*, so an `ANTHROPIC_BASE_URL` left at the official host masks a proxied `ANTHROPIC_API_URL`. That is correct at its own call sites, which ask "can this CLI still be treated as the official provider"; here it would be a false negative in exactly the case the check exists to catch. The four variables scanned are the ones EveryAPI actually injects — `GEMINI_API_BASE` / `GOOGLE_AI_BASE_URL` are omitted because `everyapi use gemini` is a native-OAuth passthrough that sets no base-URL variable at all.

When both routes are live at once — an env variable redirecting CLI-subprocess drivers and a provider entry redirecting API drivers — the check warns and names both, since which one applies depends on the agent's driver and neither surface mentions the other.

## Verification

All commands scoped per `CLAUDE.md`. Docker was unavailable on the dev box, so these ran against the native toolchain with `CARGO_TARGET_DIR` pinned to the worktree's own `target/` — the shared-`target` contention the Docker rule guards against does not apply.

```
cargo check -p librefang-cli --bins              exit 0
cargo test -p librefang-cli                      exit 0   297 passed / 0 failed (was 265; 32 new)
cargo clippy -p librefang-cli --all-targets -- -D warnings   exit 0
cargo fmt -p librefang-cli -- --check            exit 0
cargo check --workspace --lib                    exit 0
```

No `#[allow]` suppressions were added.

The 32 new tests cover credentials parsing (valid, missing/blank `relay_key`, unknown future fields, malformed JSON), `base_url` derivation with and without a trailing slash, modality inference, snapshot lookup (exact, case-insensitive, `-`/`.` version-tail normalised, genuine miss), the skip rule, `pricing_known` correctness, byte-identical output across input orderings, the `--set-default` choice, clap parsing, and a round-trip asserting the generated TOML reloads through `ModelCatalogFile` with every Text entry passing `validate()`.

Against the real 18-model listing the pipeline registers 16 (8 Text, 5 Video, 1 Image, 2 Audio) and skips 2 (`gemini-3-flash`, `gemini-3.1-pro-low`) with both named in the command output.

The gateway's protocol compatibility was confirmed with live requests before this was written: `POST /v1/chat/completions` with a bearer token, `POST /v1/messages` with the `x-api-key` header that `AnthropicDriver` actually sends, and a streaming tool-call round-trip whose `content_block_start` / `input_json_delta` / `stop_reason: tool_use` sequence passes through the gateway unchanged.

## Not covered

No live end-to-end run of an agent turn through the gateway — that needs a release build and a long-lived daemon on port 4545, which the repo's hooks reserve for a human.


houko:
## Docs

Added in a follow-up commit on this branch:

- `docs/src/app/configuration/providers/platforms/page.mdx` + the `zh/` mirror — a full EveryAPI entry. It is the only provider on that page with no fixed model list, so the entry explains why (the gateway aggregates; available models are a property of the account, not of LibreFang) and documents the three behaviours an operator would otherwise have to reverse-engineer from the command output: text models with unresolvable metadata are skipped rather than silently discarded at catalog load, streaming-only models are excluded from `--set-default`, and the doctor check warns when an env route and a provider entry are live at once.
- `docs/src/app/integrations/cli/commands/page.mdx` + the `zh/` mirror — `connect` added to the `librefang models` subcommand list.
- `CHANGELOG.md` — `[Unreleased] / Added`.

`cd docs && pnpm install --frozen-lockfile && pnpm build` exits 0 locally, so the docs deploy workflow should be green.

crates/librefang-cli/src/commands/everyapi.rs
```diff
@@ -0,0 +1,1448 @@
...
+                    "everyapi-connect-provider-write-failed",
+                    &[("error", &e)],
+                ));
+                std::process::exit(1);
+            }
+        }
+    }
+
+    report_models(&synthesis);
+    handle_default_model(&synthesis, daemon.as_deref(), set_default);
```
houko: `handle_default_model` is always given `daemon.as_deref()` here, not gated on `via_daemon` (line 630-632).

Walk through the case where a daemon *is* running (`find_daemon()` returns `Some`) but `write_provider_via_daemon` returns `false` — e.g. a transient network error, the daemon process having just died, or any other non-2xx/error-shaped response from `POST /api/registry/content/provider`. In that branch the code falls back to `write_provider_file`, writes `providers/everyapi.toml` directly to disk, and tells the user via `everyapi-connect-restart-required` that a daemon restart is needed before the new provider is live.

But then, a few lines later, if `--set-default` was passed, `handle_default_model` still receives `daemon.as_deref() = Some(base)` and will go ahead and `POST {base}/api/providers/everyapi/default` against that same (still-not-reloaded, or possibly now-unreachable) daemon. Since the daemon's in-process catalog never got the provider (the write happened directly to disk, bypassing `model_catalog_update`/`load_catalog_file`), this call is very likely to fail — producing a confusing `everyapi-connect-default-failed` message directly underneath "restart required," rather than the clearer `everyapi-connect-default-needs-daemon` message that's shown when no daemon is detected at all.

Suggest threading `via_daemon` through so `handle_default_model` only attempts the live daemon call when the provider write actually went through the daemon, e.g.:

```rust
let default_daemon = via_daemon.then(|| daemon.as_deref()).flatten();
...
handle_default_model(&synthesis, default_daemon, set_default);
```

so the "daemon not actually wired up yet" case reuses the same needs-daemon messaging as the no-daemon case, instead of attempting (and failing) a live call.

---

houko (commented):
Reviewed against the PR's design-rationale claims, cross-checking the actual implementation (not just the description) in `crates/librefang-cli/src/commands/everyapi.rs` and `crates/librefang-cli/src/doctor.rs`, plus the upstream types/logic they depend on:

- **Deterministic ordering (#3298)**: confirmed real, not just claimed — `synthesize_catalog` sorts `entries`/`skipped`/`streaming_only` before returning, and `output_is_byte_identical_regardless_of_input_order` exercises it.
- **Relay key never leaked**: verified. `EveryApiCredentials` has no `Debug` derive, `relay_key` is only read into the `Authorization: Bearer` header (`fetch_gateway_models`) and `dotenv::save_env_key` (which itself doesn't log the value). `doctor.rs`'s `EveryApiState`/`relay_key_present` only ever returns a bool.
- **`pricing_known = false` handling**: correct — `ModelCatalogEntry::pricing_known` defaults to `true` via `default_true()` in `librefang-types`, and the PR's fallback `ModelMetadata` explicitly sets `pricing_known: false` when no snapshot hit resolves, which `model_request_value` also always sends explicitly to the daemon route.
- **`ModelTier::Balanced` vs `Custom` claim**: verified directly against `librefang-runtime::model_catalog`'s `find_model` / `find_model_for_provider` / `merge_catalog_file` — `find_model` does return the first `Custom` match immediately, and `merge_catalog_file` dedupes on `(id, provider)` case-insensitively, exactly as described.
- **`XDG_CONFIG_HOME` handling**: safe — falls back to `dirs::home_dir()?.join(".config")` via `?`, no panic on a missing/unset home dir.
- **`is_proxied_via_env` divergence claim** (doctor's env-route check): confirmed against `librefang_llm_drivers::drivers::is_proxied_via_env`, which does return on the first *set* var and can mask a proxied secondary var — the doctor check's own `detect_cli_driver_env_route` correctly avoids that short-circuit.
- **i18n**: all 34 new keys (`doctor-everyapi-*`, `everyapi-connect-*`) are present and identical across `en`/`ko`/`uk`/`zh-CN`; the `i18n_checks.rs` allowlist addition for `"models connect"` is correct.

One real finding, left as an inline comment on `everyapi.rs`: `cmd_models_connect` passes `daemon.as_deref()` into `handle_default_model` unconditionally, rather than gating on whether `write_provider_via_daemon` actually succeeded (`via_daemon`). When a daemon is detected but the provider write via daemon fails for any reason (the fallback-to-file path), `--set-default` will still attempt a live `POST .../default` against a daemon whose in-process catalog was never updated, likely producing a confusing failure message instead of the clearer "needs daemon restart" messaging already used for the no-daemon case.

Native `cargo check -p librefang-cli --bins` couldn't complete in this environment (missing system `dbus-1` dev headers, unrelated to this PR's diff), so compilation was verified by careful manual reading against the actual upstream implementations rather than a full local build.

---

crates/librefang-cli/src/commands/everyapi.rs
```diff
@@ -0,0 +1,2280 @@
...
+                    model_name: model_name.to_string(),
+                    per_token: !per_call,
+                    model_ratio: item
+                        .get("model_ratio")
+                        .and_then(serde_json::Value::as_f64)
+                        .unwrap_or(0.0),
+                    completion_ratio: item
+                        .get("completion_ratio")
+                        .and_then(serde_json::Value::as_f64)
+                        .unwrap_or(0.0),
```
houko: `parse_pricing_rows` reads `model_ratio` / `completion_ratio` straight off the gateway's `/api/pricing` feed with no sign or finiteness guard, unlike the equivalent daemon-side parser this same PR adds.

Compare `crates/librefang-api/src/everyapi_catalog.rs::parse_pricing_entries`, which filters both fields with `.filter(|r| r.is_finite() && *r >= 0.0)` before using them. Here, a `model_ratio` of e.g. `-5.0` (a malformed or compromised `/api/pricing` response — this feed is fetched over the network and is exactly the kind of untrusted input the rest of this module is careful about) flows unfiltered into `PricingRow::input_cost_per_m()` / `output_cost_per_m()` (`self.model_ratio * RATIO_USD_PER_MILLION`), producing a negative cost. Since the row is `per_token` (not per-call), `pricing_known` is set `true` for that entry, so the negative price gets written into `providers/everyapi.toml` as a confidently-known figure rather than being treated as unknown.

That's a worse outcome than the `0.0` case this file's own doc comments repeatedly warn about ("a bare 0.0 would assert the model is free and poison every budget figure derived from it") — a negative `input_cost_per_m`/`output_cost_per_m` doesn't just look free, it actively corrupts any downstream sum/subtraction against a budget cap.

Suggest mirroring the API crate's guard here, e.g. `.and_then(serde_json::Value::as_f64).filter(|r| r.is_finite() && *r >= 0.0).unwrap_or(0.0)` for both fields.

---

crates/librefang-cli/src/doctor.rs
```diff
@@ -318,6 +319,245 @@ impl AuditCheck for ConfigTomlSchemaCheck {
...
+/// primary var redirected"); here a masked redirect is a false negative in
+/// exactly the scenario this check exists to catch, so we skip past official
+/// and empty values instead of stopping. Do not "fix" this back.
+fn detect_cli_driver_env_route() -> Option<(String, String)> {
+    for &(var, official_host) in CLI_DRIVER_BASE_URL_VARS {
+        let Ok(raw) = std::env::var(var) else {
+            continue;
+        };
+        let normalized = raw.trim().trim_end_matches('/').to_lowercase();
+        if normalized.is_empty() || normalized.contains(official_host) {
```
houko: `detect_cli_driver_env_route`'s "is this the official host" test is a raw substring check (`normalized.contains(official_host)`), so it can be defeated by any host that merely contains the official hostname as a substring — e.g. `ANTHROPIC_BASE_URL=https://api.anthropic.com.attacker.example/v1` normalizes to a string containing `"api.anthropic.com"`, so this loop treats it as the *official* host, skips it, and `detect_cli_driver_env_route` returns `None` as if no redirection were configured. `grade_everyapi_wiring` then reports `doctor-everyapi-provider-only` (Pass) or `doctor-everyapi-absent`/`doctor-everyapi-not-connected` instead of ever warning that a CLI-subprocess driver is actually pointed at an attacker-controlled host.

This is the exact scenario the check exists to catch (the module doc block above even frames the whole feature as "which one applies depends on the agent's driver... saying so explicitly is the whole point of this check"), so a false negative here defeats the check's purpose rather than just being a cosmetic gap.

Note `librefang_llm_drivers::drivers::is_proxied_via_env` has the same `val.contains(host)` shape, so this isn't a new pattern in the codebase — but this PR is writing a brand-new function for a brand-new security-relevant audit check, and it reproduces the weak match instead of comparing against the actual parsed host (`url_host(&normalized)`, which this same function already computes a few lines below for the *return* value, just not for the official-host comparison).

---

houko (commented):
Two findings from a daily automated pass, both concrete code-level issues rather than style nits. Skipped duplicating the existing open thread on `handle_default_model`/`via_daemon`.

---

houko (commented):
Reviewed against CLAUDE.md. Pushed one mechanical fix: `CHANGELOG.md`'s two new "Fixed" bullets had their continuation lines flush-left instead of indented two spaces like every other bullet in the file — normalized to match convention.

Everything else checked out clean: #3298 deterministic-ordering claim verified true (`everyapi_catalog.rs:443`, `everyapi.rs:655` both sort by id, with a dedicated regression test), no new config-struct fields, relay key never reaches `println!`/`format!`/`tracing::*!`/error strings anywhere in `everyapi.rs`/`doctor.rs`, i18n parity holds across all four locales, `cargo fmt` clean, no `unwrap()`/`expect()` outside `#[cfg(test)]`, docs prose-wrapping is sentence-per-line, and the host-parsing fix in `doctor.rs` genuinely closes the substring-match spoofing issue it claims to.

One thing worth a maintainer's attention rather than something I should fix myself: three commits in this branch's history (`e92644a`, `e51683a`, `864bdbd`) are authored/committed as `Claude <noreply@anthropic.com>`, while the rest of the branch is `Evan Hu`. The repo's `commit-msg` hook only greps message *text* for AI attribution, so it didn't catch the author/committer identity field, but it's the same substance the rule is guarding against. Fixing this means rewriting already-pushed history and force-pushing, which is outside what this automated review will do unilaterally — flagging for you to rebase with `git rebase -i --exec 'git commit --amend --no-edit --reset-author'` (or equivalent) before merge if you'd like it cleaned up.

---

crates/librefang-cli/src/commands/everyapi.rs
houko: Resolved on the current head (b5b7602f) — `handle_default_model` no longer receives the daemon handle unconditionally:

```rust
let default_daemon = via_daemon.then_some(daemon.as_deref()).flatten();
handle_default_model(&synthesis, default_daemon, set_default);
```

So the branch this flagged — daemon present, `write_provider_via_daemon` returns `false`, code falls back to `write_provider_file` and emits `everyapi-connect-restart-required` — now passes `None`, and the default-model write goes to the file rather than being `POST`ed at a daemon that has not reloaded the provider.

crates/librefang-cli/src/commands/everyapi.rs
houko: Resolved on the current head (b5b7602f). Both ratio fields now carry the same guard as the daemon-side `parse_pricing_entries`:

```rust
let model_ratio = item
    .get("model_ratio")
    .and_then(serde_json::Value::as_f64)
    .filter(|r| r.is_finite() && *r >= 0.0);
```

and `completion_ratio` has the identical `.filter(|r| r.is_finite() && *r >= 0.0)`.

Worth noting the fix went one step past what this comment asked for: `per_token` is now `!per_call && model_ratio.is_some()`. Filtering alone would have left a rejected row falling through to `model_ratio.unwrap_or(0.0)` while still claiming `per_token` — asserting the model is *known to be free* rather than unknown, which is a quieter version of the same corruption.

crates/librefang-cli/src/doctor.rs
houko: Resolved on the current head (b5b7602f). The substring test is gone; the comparison is now against the parsed authority, in a helper split out so the bypasses are unit-testable without mutating process env:

```rust
fn env_route_redirect_host(value: &str, official_host: &str) -> Option<String> {
    let normalized = value.trim().trim_end_matches('/').to_lowercase();
    if normalized.is_empty() {
        return None;
    }
    let host = url_host(&normalized).unwrap_or(normalized);
    (host != official_host).then_some(host)
}
```

The doc block above it names `https://api.anthropic.com.attacker.example/v1` as the case being defended against and carries a "Do not 'fix' this back" note. It also records that `librefang_llm_drivers::drivers::is_proxied_via_env` still has the weak `contains` shape — pre-existing and out of scope for that PR, but flagged there rather than lost.

PR #6584, opened 2026-07-27
feat(cli): add `service install --system` for a boot-time LaunchDaemon on macOS

houko:
## Why

`librefang service install` has always registered a *per-user* service on every platform: a systemd user unit on Linux, a LaunchAgent on macOS, an `HKCU\…\Run` entry on Windows. A per-user service starts when that user logs in.

On macOS that means a Mac which reboots and stops at the login window — or at the FileVault unlock screen — never starts the daemon. For an always-on install the only options were to leave a user logged in permanently or to hand-write a LaunchDaemon plist, and getting that plist right is not obvious: the first thing everyone writes is `ExecStart`-equivalent `librefang start`, which forks and lets the parent exit, and launchd tears the job down.

This came up while documenting NixOS deployment in #6582, where `services.librefang` gives Linux operators exactly the boot-time supervision macOS was missing.

## What `--system` does

```bash
sudo librefang service install --system
sudo librefang service uninstall --system
librefang service status              # reports both, no flag needed
```

It writes `/Library/LaunchDaemons/ai.librefang.daemon.plist`, the directory launchd loads at boot before any login.

Four things in the generated job are load-bearing, and each is a way to get this wrong:

- **`start --foreground`.** Plain `librefang start` takes the `!spawned && !foreground` branch at `crates/librefang-cli/src/commands/daemon.rs:347` into `spawn_detached_daemon`, which re-execs, calls `libc::setsid()`, and returns from the parent after a health poll. launchd would see the job exit immediately and reap the detached child with it.
- **`UserName`.** Without it the daemon runs as root against a user's state directory, and every file it creates becomes root-owned.
- **`HOME` and `LIBREFANG_HOME` as two separate values.** `HOME` must be the account's home so `dirs::home_dir()` resolves — the first-start `librefang init` path (`daemon.rs:296-318`) exits 1 when it does not (`init.rs:13-19`). `LIBREFANG_HOME` is the state directory. Deriving one from the other passes for the default `~/.librefang` and breaks the moment `sudo -E` carries `LIBREFANG_HOME=/opt/librefang`, where the state dir's parent is `/opt`. There is a regression test for exactly that case.
- **Ownership of the whole state tree.** launchd opens `StandardOutPath` *before* dropping to `UserName`, so the log is created as root and the daemon cannot write it. The install creates the state dir and log and then hands the entire tree over, recursively — the second commit fixes this, because the first version chowned only the directory node. `~/.librefang` almost always exists already (the user ran `librefang init` or the LaunchAgent first), so `create_dir_all` is a no-op and its contents keep whatever uid created them; a state dir previously written by a `sudo librefang start` would leave the LaunchDaemon hitting `EACCES` at an arbitrary depth long after `service status` reported the job registered. The walk records symlinks but never descends them (`DirEntry::file_type` describes the entry, not its target, and the chown is `lchown`), so a link out of the state dir cannot redirect the handover onto an unrelated tree.

The plist is also chmodded to 0644: launchd refuses to load a LaunchDaemon that is writable by anyone other than its owner.

## Resolving who to run as

`SUDO_USER` is the only signal that identifies the human behind a root process, so `--system` requires being invoked through `sudo` rather than from a root login. A missing or `root`-valued `SUDO_USER` is an error, not a silent default — a LaunchDaemon quietly running as root against `/var/root/.librefang` is worse than a refusal.

The home directory, uid and gid come from `getpwnam` rather than `dirs::home_dir()`, which under `sudo` returns root's home.

## What did not change

The pre-existing root guard on the per-user path is intact. Plain `librefang service install` still refuses to run under `sudo` and exits 1 — that guard exists because a LaunchAgent installed as root is registered for the root account and never starts for the invoking user. It now additionally points at `--system` when it fires, which is the thing the user probably wanted.

`--system` on Linux or Windows is a clear error naming the alternatives (the `services.librefang` NixOS module, a hand-written `/etc/systemd/system/` unit, or the existing user service plus `loginctl enable-linger`) rather than a silent no-op.

## Verification

- `cargo check --workspace --lib` — exit 0
- `cargo clippy -p librefang-cli --all-targets -- -D warnings` — exit 0
- `cargo test -p librefang-cli` — all pass, including the three i18n gates, four new plist tests and the ownership-handover walk test (259 tests in the bin target)
- `cargo fmt --check` — exit 0
- `python3 scripts/check-changelog-attribution.py --all-unreleased` — OK
- All 15 new i18n keys confirmed present in all four locale files (`en`, `zh-CN`, `ko`, `uk`)

**The non-macOS arms are verified by CI, not locally.** `#[cfg(not(target_os = "macos"))]` code never compiles on a macOS host, and that is where a real hazard lives: a trailing `return` after those arms' `process::exit` would be unreachable code, which `-D warnings` turns into a build failure on precisely the platforms a macOS developer cannot see. Each arm therefore terminates on its own — the macOS arm returns, the others exit — rather than falling through to a shared `return`, and there is a comment at the branch saying why. I attempted `cargo check --target x86_64-pc-windows-msvc` to prove it and could not: this host's `cargo` is the Homebrew build, which does not see targets installed through `rustup`, and switching toolchains would trigger a full recompile across two active worktrees. The Windows and Linux CI lanes cover it.

The plist renderer is a pure function gated with `#[cfg_attr(not(test), cfg(target_os = "macos"))]`, so its tests compile and run on every platform's test build rather than depending on someone repeating a privileged macOS run.

## Not verified

**No LaunchDaemon was actually installed or booted.** That needs `sudo` on a macOS host and a reboot to prove the boot-time path, and it writes to `/Library/LaunchDaemons`. The tests cover the plist contents and the platform gating; they do not cover launchd accepting the job.

The sequence to confirm it, on a Mac you are willing to reboot:

```bash
sudo librefang service install --system
sudo launchctl print system/ai.librefang.daemon | head -20
curl -s http://127.0.0.1:4545/api/health
# reboot, do not log in, then from another machine or after logging in:
curl -s http://127.0.0.1:4545/api/health
sudo librefang service uninstall --system
```

Related: #6582 adds the equivalent boot-time supervision for NixOS via a `services.librefang` module.


crates/librefang-cli/src/commands/maintenance.rs
```diff
@@ -255,6 +285,323 @@ pub(crate) fn service_install_macos(binary: &std::path::Path, librefang_home: &s
...
+            ui::hint(&i18n::t("maintenance-service-system-sudo-hint"));
+            std::process::exit(1);
+        }
+    };
+
+    // Honour an explicitly forwarded LIBREFANG_HOME (`sudo -E`), otherwise derive it from the target
+    // account rather than from root's home, which is what `cli_librefang_home()` would return here.
+    let librefang_home = match std::env::var("LIBREFANG_HOME") {
+        Ok(h) if !h.is_empty() => std::path::PathBuf::from(h),
+        _ => target.home.join(".librefang"),
```
houko, line 460: `std::os::unix::fs::chown` follows symlinks (it wraps `libc::chown`, not `lchown`). Both paths handed to it here are attacker-reachable before this code ever runs as root: `librefang_home` defaults to `target.home.join(".librefang")`, i.e. a path inside the target account's own home directory, which that account fully controls.

Concretely: a standard (non-admin) user pre-creates `~/.librefang` as a symlink to `/etc` (or any other root-owned path) before an administrator — or an MDM/fleet-management script, which is exactly the always-on/managed-Mac scenario this PR is written for — runs `sudo librefang service install --system` with `SUDO_USER` set to that account. `create_dir_all` on an existing symlink-to-directory is a no-op, so the `chown` a few lines down follows the symlink and hands ownership of `/etc` itself to the standard user's uid/gid.

Use `std::os::unix::fs::lchown` and/or refuse to proceed (`fs::symlink_metadata(path)?.file_type().is_symlink()`) if either path is already a symlink.

---

crates/librefang-cli/src/commands/maintenance.rs
```diff
@@ -255,6 +285,323 @@ pub(crate) fn service_install_macos(binary: &std::path::Path, librefang_home: &s
...
+
+/// Resolve who the LaunchDaemon should run as.
+///
+/// `SUDO_USER` is the only signal that identifies the human behind a root process. Its absence means
+/// the command was run from a real root login rather than through `sudo`, and there is no way to guess
+/// which account's `~/.librefang` the daemon should serve — so that is an error rather than a default.
+#[cfg(target_os = "macos")]
+fn resolve_system_service_target() -> Option<SystemServiceTarget> {
+    let user = std::env::var("SUDO_USER").ok().filter(|u| !u.is_empty())?;
+    if user == "root" {
```
houko, line 342: This only rejects the literal username `"root"`, not uid 0. `passwd_entry_for` already returns the resolved `uid` — nothing here checks it. If `SUDO_USER` resolves (via `getpwnam`) to any account with uid 0 that isn't spelled exactly `"root"` (a duplicate/alias superuser entry, which some systems do have), this function accepts it, `UserName` in the generated plist gets set to that account, and the LaunchDaemon keeps running as root — silently defeating the entire "drop privileges to a real account" goal the CHANGELOG and PR description describe. Should be `if user == "root" || uid == 0 { return None; }` (check after the `passwd_entry_for` call, or reorder).

---

crates/librefang-cli/src/commands/maintenance.rs
```diff
@@ -255,6 +285,323 @@ pub(crate) fn service_install_macos(binary: &std::path::Path, librefang_home: &s
...
+    <string>{librefang_home}/daemon.log</string>
+</dict>
+</plist>
+"#,
+        binary = binary.display(),
+        user = user,
+        // `HOME` is the account's real home, not the state dir: `dirs::home_dir()` reads it, and the
+        // first-start `librefang init` path exits when it resolves to nothing.
+        home = account_home.display(),
+        librefang_home = librefang_home.display(),
```
houko, line 411: None of `binary`, `user`, `home`, or `librefang_home` are XML-escaped before being interpolated into this plist template. `librefang_home` in particular is honored verbatim from an unsanitized `LIBREFANG_HOME` env var a few lines below in `service_install_macos_system` (`Ok(h) if !h.is_empty() => std::path::PathBuf::from(h)`). A value containing `</string><key>ProgramArguments</key><array><string>...` would inject additional keys/elements into a file that gets written to `/Library/LaunchDaemons` and loaded as a root-privileged, `RunAtLoad`+`KeepAlive` boot daemon — i.e. it can redefine what runs as root at every boot. This is reachable via `sudo -E` forwarding a hostile `LIBREFANG_HOME` from the invoking user's own shell environment into an admin/MDM script that runs this installer, not just via a deliberately malicious direct invocation. Please escape `&`, `<`, `>`, `"` (at minimum) before interpolating any of these four values into the XML.

---

crates/librefang-cli/src/commands/maintenance.rs
```diff
@@ -39,14 +39,44 @@ pub(crate) fn resolve_binary_path() -> std::path::PathBuf {
...
 
-pub(crate) fn cmd_service_install() {
+pub(crate) fn cmd_service_install(system: bool) {
+    // The two modes have opposite privilege requirements, so the root check has to branch before anything else runs.
+    // A per-user LaunchAgent / systemd user unit installed while root would be registered for the root account and never start for the invoking user, which is the mistake the original guard was written to catch.
+    // A LaunchDaemon writes to /Library/LaunchDaemons, which only root can do.
+    // Each arm terminates on its own — the macOS arm returns, the others exit — rather than falling
+    // through to a shared `return`. A trailing `return` after the non-macOS `process::exit` would be
+    // unreachable code, which `-D warnings` turns into a build failure on exactly the platforms a
+    // macOS developer never compiles.
```
houko, line 49: CLAUDE.md's prose-wrapping rule ("no column limit; break only at sentence boundaries... one sentence = one line... applies universally... source code doc-comments... and any multi-line free-form prose comment block") is violated by several of the new comments in this file. Here, "Each arm terminates on its own — the macOS arm returns, the others exit — rather than falling through to a shared `return`." is split across lines 46-47, and the next sentence spans 47-49 — both hard-wrapped at ~100 columns instead of one sentence per line.

Same pattern recurs elsewhere in this diff's new doc comments, e.g. lines 295-298 (`SystemServiceTarget` doc), 306-309 (`passwd_entry_for` doc), and 358-364 (`macos_system_plist` doc) all split single sentences across multiple lines. Since this is all new prose added by this PR, the rule applies in full — not asking for a pass over every line, but flagging the pattern since it's the opposite of what the file's existing (pre-PR) comments already do correctly.

---

houko (commented):
Daily automated review. Focused on the privileged install path (root-owned LaunchDaemon writing to `/Library/LaunchDaemons`, chown to a resolved account). A few correctness/security gaps and one repo-convention violation below; nothing else stood out.

---

PR #6585, opened 2026-07-27
fix(cli): give the TUI a process-lifetime runtime (startup panic + silent session-summary loss)

houko:
## Problem

Two bugs with one root cause: the TUI event loop is synchronous, but the kernel underneath it is async.

**1. Panic on startup.** `librefang chat` and `librefang tui` died as soon as the in-process kernel finished booting:

```
thread 'main' (11884342) panicked at crates/librefang-kernel/src/kernel/accessors.rs:630:22:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

`main()` is a plain `fn main()`, so `handle_kernel_ready` ran on a runtime-free main thread and called `spawn_approval_sweep_task`, which opens with `Handle::current()`. The only runtime in the picture was the throwaway `Runtime::new()` in `spawn_kernel_boot`'s worker thread, dropped as soon as the boot closure returned — so even reaching the spawn through it would have aborted the sweep loop immediately. Both TUI entry points were affected (`chat_runner.rs:213`, `mod.rs:1294`).

**2. Silent session-summary loss on `/new`.** Every other TUI async op also built a per-call `Runtime::new()` and dropped it right after `block_on`. Dropping a runtime aborts its tasks, and `reset_session` deliberately *detaches* the summary write — `save_session_summary` does `Handle::try_current()` + `handle.spawn(...)` so the aux-LLM digest happens off the reset's return path. Correct on the daemon, whose runtime outlives every request. On the TUI the runtime died first.

## Why the data loss stayed invisible

It is config-dependent. With no `[llm.auxiliary] session_summary` chain, `build_session_summary` early-returns a trivial digest — no `.await` before the write, so the task finishes inside the scheduler window `reset_session`'s remaining async work provides and the row lands anyway. Configure an aux chain and the same task awaits an LLM round-trip, the drop aborts it, and the summary is gone.

No panic, no warning, and the `try_current()` branch was taken either way. It hits exactly the operators who configured good summaries.

I found this while verifying the fix: my first attempt at a regression test asserted the row was *absent* on a throwaway runtime and it failed — the row was there. Chasing that down is what surfaced the aux-chain dependency, and both tests below now encode it.

## Fix

A process-lifetime multi-thread runtime behind a `OnceLock` in `tui::event`, exposed two ways:

- `spawn_kernel_background_tasks()` — enters it just long enough to spawn the sweep loop.
- `block_on_tui()` — replaces all four throwaway `Runtime::new()` sites: the in-process stream driver, the dashboard auto-dream status pull, and the two `/new` reset paths.

Kernel boot runs under the same runtime, so tasks spawned during boot (`set_self_handle`'s prune, `publish_event`) also survive the boot thread.

`block_on` panics only when called from a thread already driving tokio tasks. Every TUI caller is the main thread or a `std::thread::spawn`'d worker, so none qualify; that constraint is documented on the wrapper.

## Correction to the first commit

`bdd37fb9` and its code comments claimed a loop-wide `EnterGuard` would break `/new` with "Cannot start a runtime from within a runtime". **That is wrong.** On tokio 1.53 an `EnterGuard` on a non-worker thread does not trip that panic — only actually running on a runtime worker thread does. I verified both directions with a standalone probe before writing the follow-up.

The narrow guard scope is still correct (nothing else on the main thread needs a runtime context), but the stated reason was not. Code comments are fixed in the second commit, which also carries the correction note. The earlier commit message is left as-is since it sits under a merge.

The test that had encoded the false claim (`tui_runtime_guard_does_not_leak_to_the_caller`) was vacuous — its assertions held with or without the guard — and is replaced by one that tests the real property.

## Breadth scan

`spawn_task_board_sweep_task`, `spawn_session_stream_hub_gc_task`, and `spawn_key_validation` share the same `Handle::current()` contract, but every caller is inside the daemon or desktop server's runtime — the TUI was the only sync caller. The requirement is now documented on all three rather than relaxed to `try_current()`: a spawner asking to be called from a runtime context is the authoritative contract, and degrading it would turn a loud panic into a sweep loop that quietly never runs.

## Verification

Released `2026.7.21` binary as the before case, local debug build as the after case, both driven under a pty.

| Entry point | Old binary | New binary |
| --- | --- | --- |
| `librefang chat` | panics at `accessors.rs:630` | boots clean, reaches `LibreFang kernel booted successfully` |
| `librefang tui` (in-process menu) | panics at `accessors.rs:630` | boots clean, same |

```
$ script -q /dev/null /opt/homebrew/bin/librefang chat </dev/null
thread 'main' (12676405) panicked at crates/librefang-kernel/src/kernel/accessors.rs:630:22:
there is no reactor running, must be called from the context of a Tokio 1.x runtime

$ (sleep 6; printf '\r'; ...) | script -q /tmp/tui-old2.txt /opt/homebrew/bin/librefang tui
thread 'main' (12985526) panicked at crates/librefang-kernel/src/kernel/accessors.rs:630:22:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

After, both paths: 0 panic lines in the pty capture, `logs/tui.log` reaches the boot line.

Tests:

```
test tui::event::tests::tui_runtime_tasks_outlive_the_spawning_scope ... ok
test tui::event::tests::block_on_tui_detached_tasks_survive_the_call ... ok
test result: ok. 2 passed; 0 failed

test summary_survives_reset_on_a_long_lived_runtime ... ok
test trivial_summary_path_is_not_where_the_loss_shows_up ... ok
test result: ok. 2 passed; 0 failed
```

`session_summary_runtime_lifetime_test` drives the real `reset_session` the way the TUI does and asserts the summary row lands in `kv_store`. Its companion test asserts only what is invariant on a throwaway runtime — the reset's caller-visible result and the synchronous transcript clear. Whether the *row* lands there is a scheduling race the detached write usually wins in the aux-less config, which is exactly why the original bug was invisible by default; that is documented in the module comment rather than asserted, since asserting either direction would flake on a loaded machine. A test that configures a real aux chain would need a live LLM, so the abort semantics themselves are pinned at the primitive by `block_on_tui_detached_tasks_survive_the_call`.

### End-to-end, live turn

Unit tests don't execute `spawn_inprocess_stream` or the `/new` reset paths, so one pty session drives the whole chain against the real `claude-code` provider — send a message, let it stream, then `/new`:

```
$ sqlite3 ~/.librefang/data/librefang.db "select count(*) from kv_store where key like 'session_%'"
0

$ (sleep 8; printf 'Reply with exactly: PONG\r'; sleep 60; printf '/new\r'; ...) \
    | script -q /tmp/turn.txt .../debug/librefang chat

# turn streamed:
INFO ...agent_loop::end_turn: Streaming agent loop completed agent=assistant iterations=1 tokens=7
INFO ...claude_code: Claude Code CLI subprocess completed successfully model=sonnet

# after /new — the summary the old code would have raced away:
$ sqlite3 ... "select key, length(value) from kv_store where key like 'session_%'"
session_39d98ddf-f0a7-4520-bdf7-4fa9d5fef6c6|186

$ sqlite3 ... "select substr(value,1,90) from kv_store where key like 'session_%'"
"Session on 2026-07-27 (auto-summary, aux LLM unavailable)\n\n- Turns: 1 user / 1 assistant
```

0 panics in the pty capture and in `tui.log`. The session id in the row matches the one in the turn's log lines, and the dashboard site was already exercised by the `librefang tui` run reaching the main phase — so all four `block_on_tui` call sites ran.

Both TUI tests were mutation-checked — swapping the shared runtime back for a throwaway makes them fail:

```
panicked at crates/librefang-cli/src/tui/event.rs:3111:9:
task spawned on the TUI runtime was aborted after its spawning scope ended

task detached during block_on_tui was aborted when the call returned
```

Daemon path unaffected:

```
test kernel::tests::test_spawn_approval_sweep_task_is_idempotent ... ok
```

`cargo clippy -p librefang-cli -p librefang-kernel --all-targets -- -D warnings` is clean.

## Notes

- Concurrency was checked before shipping: two threads each holding `block_on_tui` over a 2s future complete in ~2.0s, not ~4.0s, so a stream turn does not freeze the dashboard tick for its duration. (`Runtime::block_on` and `Handle::block_on` behave identically here.)
- `spawn_inprocess_stream` loses its graceful runtime-creation failure path (`StreamDone(Err(...))`) in favour of `tui_runtime()`'s `.expect(...)`. Runtime creation essentially never fails, so this is a note rather than a defect — it is also why `tui-event-stream-runtime-error` is now unreferenced. The translations are left in place rather than deleted.
- Quitting the TUI immediately after `/new` can still lose the summary — that is a process-lifetime bound, not this bug.


crates/librefang-cli/src/tui/event.rs
```diff
@@ -311,13 +311,51 @@ pub fn spawn_daemon_detect(tx: mpsc::Sender<AppEvent>) {
     });
 }
 
+/// Process-lifetime tokio runtime for the TUI's long-lived background tasks.
```
houko, line 314: This PR (both the original sweep-task fix and the `route every TUI async op through the shared runtime` follow-up) fixes a user-facing panic in `librefang chat` / `librefang tui` plus a silent session-summary data-loss bug, but there's no `CHANGELOG.md` bullet under `[Unreleased]`.
Every other recently-merged PR in the log has a `### Fixed` entry with `(@handle)` attribution, and the repo's `pre-commit` hook / CI job specifically check CHANGELOG attribution on `[Unreleased]` additions.
Flagging rather than adding it myself since I don't want to guess the attribution handle — worth a bullet along the lines of "Fix TUI (`librefang chat`/`librefang tui`) panicking on kernel-ready with 'there is no reactor running' and drive `/new`'s detached session-summary write off the shared runtime instead of a throwaway one that aborted it silently."

---

houko (commented):
Reviewed the runtime-lifetime fix (both the original sweep-task commit and the follow-up that routes every TUI async op through `block_on_tui`). The core mechanism checks out: `EnterGuard`/`Handle::current()` context is thread-local, so `tui_runtime().enter()` on the boot thread and inside `spawn_kernel_background_tasks` never leaks into the main event loop, and the two new regression tests (`tui_runtime_tasks_outlive_the_spawning_scope`, `block_on_tui_detached_tasks_survive_the_call`) actually pin the failure mode rather than just checking happy-path behavior. Ran `cargo check -p librefang-kernel --lib`, `cargo clippy -p librefang-cli --all-targets -- -D warnings`, and `cargo clippy -p librefang-kernel --all-targets -- -D warnings` — all clean.

Pushed two small direct fixes (mechanical, no behavior change):
- Rewrapped the new doc comments in `tui/event.rs`, `tui/chat_runner.rs`, `tui/mod.rs`, and `kernel/accessors.rs` to the repo's one-sentence-per-line prose convention (they were hard-wrapped mid-sentence at ~80 cols), and filled in the `# Panics` doc on `spawn_key_validation`, which shares the same `Handle::current()` contract as the other sweep spawners but was left undocumented.
- Ran `cargo fmt` on the new `session_summary_runtime_lifetime_test.rs` — two spots didn't match the `max_width = 100` config.

Left one comment on the missing CHANGELOG entry — everything else looked solid.

---

crates/librefang-cli/src/tui/event.rs
```diff
@@ -311,13 +311,51 @@ pub fn spawn_daemon_detect(tx: mpsc::Sender<AppEvent>) {
...
+///
+/// The `OnceLock` is never dropped (statics aren't), so tasks spawned onto this handle live until the process exits.
+fn tui_runtime() -> &'static tokio::runtime::Runtime {
+    static RT: OnceLock<tokio::runtime::Runtime> = OnceLock::new();
+    RT.get_or_init(|| {
+        tokio::runtime::Builder::new_multi_thread()
+            .enable_all()
+            .thread_name("librefang-tui-bg")
+            .build()
+            .expect("Failed to create Tokio runtime")
```
houko, line 327: `.expect("Failed to create Tokio runtime")` here means that if runtime construction ever fails (e.g. thread-creation exhaustion) on the *first* caller to touch `tui_runtime()`, the whole TUI process panics — where before this PR, `spawn_inprocess_stream`'s per-call `Runtime::new()` failure was handled gracefully (sent an `AppEvent::StreamDone(Err(...))` via the now-removed `tui-event-stream-runtime-error` message).

This is a low-probability trade-off and arguably acceptable — a broken tokio runtime makes the rest of the TUI unusable anyway — but it's a real behavior change from "that one operation fails" to "the process panics," worth a conscious call rather than a silent one. Up to you whether to keep the panic or fall back to a user-visible error on first-init failure.

---

houko (commented):
Reviewed against CLAUDE.md. Pushed one mechanical fix: the new `session_summary_runtime_lifetime_test.rs` doc comments were hard-wrapped mid-sentence, missing the sentence-per-line rewrap that `b8c521498` already applied to `event.rs`/`chat_runner.rs`/`mod.rs`/`accessors.rs` — reflowed to match.

Verified clean: `block_on_tui`'s documented panic contract matches every actual call site, the `bdd37fb9` correction is applied consistently with no leftover incorrect claims elsewhere, the two new regression tests are real (poll an `AtomicBool` set by a detached/spawned task, not vacuous assertions), locale cleanup is consistent across all four `main.ftl` files, and `cargo fmt` is clean.

One more thing worth a maintainer's attention rather than something I should fix myself: commits `b8c52149` and `804159` in this branch are authored/committed as `Claude <noreply@anthropic.com>`, unlike the rest of the branch. The repo's `commit-msg` hook only greps message text, so it didn't catch the author/committer identity field, but it's the same substance the "no AI attribution" rule guards against. Fixing it means rewriting pushed history and force-pushing, which this automated review won't do unilaterally — flagging for you to reset the author before merge if you'd like it cleaned up.

---

PR #6586, opened 2026-07-27
feat(dashboard): add an EveryAPI connect action to the Providers page

houko:
## Why

Registering the EveryAPI gateway required `librefang models connect everyapi` in a terminal. The dashboard had no path to it at all — and not because the flow was merely missing a button.

EveryAPI is not a built-in provider. Until a registry entry exists it is absent from `GET /api/providers` altogether rather than being listed as unconfigured, and the Add picker's catalog is exactly `useProviders()` filtered to unconfigured entries (`ProvidersPage.tsx`, "the unconfigured catalog lives behind the Add picker"). Nothing the picker reads could ever contain it, so no amount of searching in that drawer would surface EveryAPI.

## What this adds

The picker footer offers a connect action while no `everyapi` entry exists, opening a two-field drawer for the relay key and an optional gateway root.

It keys off the entry **existing**, not off it being usable. An entry with a missing key is already reachable through the normal configure flow, and re-offering "connect" there would overwrite it. One of the tests pins this by using an unconfigured `everyapi` entry, so loosening the condition to "configured" fails the suite.

`useConnectEveryApi` writes the provider entry first and the key second, because the key endpoint addresses a provider by id and there is nothing to attach a key to until the entry exists.

## Why the entry is registered with no models

Deliberately, and this is the part worth reviewing:

`catalog_needs_initial_refresh` in `crates/librefang-api/src/everyapi_catalog.rs` is true exactly when the provider is configured **and** has no live models. So writing the entry with an empty `models` array is what arms the daemon's `refresh_if_missing_in_background` (wired at `routes/providers.rs:859`), which fetches `/v1/models` and `/api/pricing` and synthesises the catalog itself.

The alternative — replicating the CLI's synthesis in the browser — would mean a second copy of the pricing rules to keep in sync with `parse_pricing_entries`, and the gateway is not CORS-open to the dashboard in any case.

## Cross-language contract

`EVERYAPI_PROVIDER` (id, display name, key env var, default gateway root) is exported alongside the mutation and documented as mirroring `PROVIDER_ID` / `PROVIDER_DISPLAY_NAME` / `API_KEY_ENV` in `crates/librefang-cli/src/commands/everyapi.rs`. Drift there would register a provider id the daemon never refreshes. The test mock repeats the literal for the same reason, so a change on either side surfaces as a failing placeholder assertion rather than a silently dead feature.

## Relationship to #6583

#6583 has merged (`c2a11aa6`), so the daemon-side catalog synthesis this flow relies on is now on `main` and the feature is complete end to end: the UI registers the entry with an empty `models` array, and `everyapi_catalog.rs` backfills the catalog in the background.

This branch is rebased onto that tip. The two changes touch disjoint files — this one is dashboard-only, #6583 was Rust-only — and the only conflict rebasing was the `CHANGELOG.md` `[Unreleased]` entry, resolved by placing this entry above #6583's at the top of `### Added`.

The endpoints called here (`POST /api/registry/content/provider` and `POST /api/providers/{id}/key`) predate #6583 and are unchanged by it.

## Conventions

- Data access goes through `lib/mutations/providers` per the dashboard data-layer rule; no inline `fetch()`
- `invalidateQueries` uses the `providerKeys` / `modelKeys` factories, colocated in the mutation hook, so both the configured grid and the picker refetch
- Strings in all four locales (`en` / `zh` / `ko` / `uk`), 9 keys each
- Three drawer states covered: required-field error on the key, inline `role="alert"` off the mutation error, spinner plus disabled submit while pending
- `lucide-react` icons, no emoji; Tailwind with the existing token classes; TypeScript strict, no `any`

## Verification

- `pnpm typecheck` — exit 0
- `pnpm test` — **87 files, 888 tests, all passing**. `ProvidersPage.test.tsx` goes 10 → 13.
- `pnpm lint` — 0 errors, 85 warnings, all pre-existing. The two files this PR touches do not appear in the lint report at all.
- `node scripts/i18n-parity.mjs` — the 9 new keys introduce no drift (`zh.json` reports parity with `en.json`; the drift the script does report is pre-existing `*_few` / `*_many` plural keys missing from `ko` / `uk`, none of them `everyapi_*`)
- `python3 scripts/check-changelog-attribution.py --all-unreleased` — OK, re-run after the rebase
- `npx playwright test e2e/everyapi-connect.spec.ts` — 2 passed. Added in the second commit; see below.
- Post-rebase, the dashboard payload is byte-identical to the pre-rebase revision (`git diff 986d2406 HEAD -- crates/librefang-api/dashboard/` is empty); the rebase changed `CHANGELOG.md` only, so the test results above still describe the pushed tree.

**The E2E suite caught a real bug on its first run.** The reset-on-close effect named `connectMutation` as a dependency, and `useMutation` returns a fresh object every render, so the effect re-ran on every render and its `setState` calls triggered the next — React aborted the drawer with "Maximum update depth exceeded". The unit tests structurally could not see it, because a mocked hook returns a reference-stable literal. It is fixed by reading the mutation through a ref, and the suite now asserts the warning count is zero.

One thing the mock change is load-bearing for: `ProvidersPage.test.tsx` replaces the whole `lib/mutations/providers` module with `vi.mock`, so adding an import to the page without adding it to the mock would have broken all ten existing tests. They were re-run before the new cases were written, specifically to confirm that.

## Not verified

**No daemon ran this.** A browser does now run it — `e2e/everyapi-connect.spec.ts` drives the real drawer in Chromium and asserts the two request bodies and their order. What it cannot assert is the other side of the wire: every backend call is fulfilled by `page.route`, so nothing here confirms that `POST /api/registry/content/provider` *accepts* this body shape, or that the daemon then picks the entry up and backfills the catalog. That needs a live daemon plus a real EveryAPI relay key, and it is the step worth doing before merge:

```
# with a daemon on 4545 and the dashboard open
# Providers → Add → "Connect EveryAPI gateway" → paste relay key → Connect
curl -s http://127.0.0.1:4545/api/providers | python3 -c "import sys,json; print([p['id'] for p in json.load(sys.stdin)])"
```

The empty-`models` design also rests on reading `everyapi_catalog.rs` — now on `main` — rather than on running it.

## A `[Unreleased]` structure problem found while rebasing — fixed in #6587

`c2a11aa6` inserted its new `### Fixed` heading in the middle of the existing `### Added` list, so four entries that are additions ended up under `### Fixed`: `librefang service install --system`, the NixOS deployment surface, the MCP `resources` primitive, and the `librefang_media_understanding_failures_total` counter. Since `release.yml` copies those headings verbatim into the published notes, they would ship listed as bug fixes.

It is not fixed here — the move touches a ~30-line span of the one file every open branch appends to, which would conflict with all of them from inside a dashboard PR. #6587 does it on its own branch instead. Placing this PR's entry at the top of the genuine `### Added` is independent of that fix and correct either way.


crates/librefang-api/dashboard/src/lib/mutations/providers.ts
```diff
@@ -127,6 +127,72 @@ export function useSetDefaultProvider() {
...
+      const trimmedBase = (baseUrl ?? "").trim();
+      await createRegistryContent("provider", {
+        id: EVERYAPI_PROVIDER.id,
+        display_name: EVERYAPI_PROVIDER.displayName,
+        api_key_env: EVERYAPI_PROVIDER.apiKeyEnv,
+        base_url: trimmedBase || EVERYAPI_PROVIDER.defaultBaseUrl,
+        key_required: true,
+        models: [],
+      });
+      await setProviderKey(EVERYAPI_PROVIDER.id, trimmedKey);
```
houko, line 180: `useConnectEveryApi` only wires `onSuccess` (line 189), not `onError`/`onSettled`. If `createRegistryContent` (line 179) succeeds but this `setProviderKey` call throws, the mutation fails as a whole but `providerKeys`/`modelKeys` are never invalidated — so the UI's cached `everyApiPresent` check on the Providers page still shows "Connect EveryAPI gateway" even though the daemon now has a keyless `everyapi` entry. A retry would then re-run `createRegistryContent` against an id that already exists.

Worth invalidating on `onError`/`onSettled` too (or checking-before-create) so a retry sees the entry already exists and only re-attempts the key write.

---

crates/librefang-api/dashboard/src/lib/mutations/providers.ts
```diff
@@ -127,6 +127,72 @@ export function useSetDefaultProvider() {
...
+ * `crates/librefang-api/src/everyapi_catalog.rs` is true exactly when the provider is
+ * configured but has no live models, and `refresh_if_missing_in_background` then fetches
+ * `/v1/models` and `/api/pricing` and synthesises the entries. Duplicating that synthesis in
+ * the browser would mean a second implementation of the pricing rules to keep in sync, and the
+ * gateway is not CORS-open to the dashboard anyway.
+ *
+ * Ordering matters: the provider entry is written first, because the key endpoint addresses a
+ * provider by id and there is nothing to attach a key to until the entry exists.
+ */
+export function useConnectEveryApi() {
```
houko, line 159: Minor CLAUDE.md nit: the new doc comment above this function (and the one added around `ProvidersPage.tsx`'s picker footer) hard-wraps multi-sentence paragraphs at a fixed column rather than one sentence per line, per the repo's prose-wrapping rule. Not blocking — it matches the pre-existing `useTestProvider` comment a few lines up, which has the same issue and predates this PR — just flagging since this PR does add fresh paragraphs of its own that could have followed the rule.

---

houko (commented):
Reviewed against CLAUDE.md house rules. Data layer, query-key, mutation-invalidation, i18n parity, and no-`any`/no-emoji checks all pass. Two non-blocking ambiguous points left inline for a human call; no mechanical fixes needed.

---

crates/librefang-api/dashboard/src/lib/mutations/providers.ts
houko: Fixed in 986d2406 — switched to `onSettled`, which matches `useTestProvider` a few lines above (it already uses `onSettled` for the same "the side effect happened even though the call reported failure" reason).

Two regression tests in `src/lib/mutations/memory-providers-workflows.test.tsx`, under `describe("useConnectEveryApi")`:

- `invalidates provider and model caches when the key write fails after the entry was created` — mocks `setProviderKey` to reject, asserts `createRegistryContent` was still called and that both `providerKeys.all` and `modelKeys.lists()` are invalidated anyway. Confirmed this one fails against the previous `onSuccess` wiring.
- `invalidates provider and model caches on the success path` — guards against a later change that trades the failure path for the success one.

Went with `onSettled` over check-before-create because it fixes the stale-cache half directly, and the retry-collision half follows from it: once the caches are invalidated, the page re-reads the provider list and the connect action is gone, so there is no second attempt to collide.

crates/librefang-api/dashboard/src/lib/mutations/providers.ts
houko: Rewrapped in 986d2406. Four paragraphs, all of them prose this PR introduced:

- `providers.ts` — the `useConnectEveryApi` doc block.
- `ProvidersPage.tsx` — the `EveryApiConnectDrawer` doc block, the `resetRef` render-loop comment, and the `everyApiPresent` comment.

Left `useTestProvider`'s comment alone, as you noted it predates this PR and the rule says not to rewrap untouched paragraphs.

Issue #5855, opened 2026-05-29
Empty mcp_servers = [] leaks all globally-connected MCP tools into agent


nevgenov:
# Empty `mcp_servers = []` leaks ALL globally-connected MCP tools into agent

## Summary

When an agent's manifest has `mcp_servers = []` (empty list), the agent ends up with **every MCP tool from every globally-connected MCP server** in its tool list — not zero tools as the field name suggests. The empty list is interpreted as "no filter / wildcard" rather than "no servers".

This is the exact opposite of operator intent and causes downstream failures: the model sees tools it has no runtime access to, calls them, they fail, and after 3 consecutive failures the streaming loop exits with `Repeated tool failures: 3 consecutive iterations with 1 errors` (visible to the end user as "Something went wrong: please try again").

## Affected version

- `main @ cc75e55c491b672f9adef7a171fd1db3e3c7e284`
- Confirmed observed on installed `2026.5.12-beta.11` in production.

## Root cause

`crates/librefang-kernel/src/kernel/tools_and_skills.rs`, around lines 189–218.

```rust
let mcp_allowlist = e.manifest.mcp_servers.clone(); // = [] for affected agents
// ...
if !mcp_disabled {
    if let Ok(mcp_tools) = self.mcp.mcp_tools.lock() {
        // ...
        let mut mcp_candidates: Vec<ToolDefinition> = if mcp_allowlist.is_empty() {
            mcp_tools.iter().cloned().collect()   // <-- ALL global MCP tools
        } else {
            // filter by configured server names
        };
        // ...
        for t in mcp_candidates {
            // MCP tools are NOT filtered by capabilities.tools.
            // mcp_candidates is already scoped to the agent's allowed servers
            // (via mcp_allowlist above), so no further declared_tools filtering
            // is needed.
            all_tools.push(t);
        }
    }
}
```

`mcp_tools` is a single global `Mutex<Vec<ToolDefinition>>` populated by every connected MCP server across the kernel (`crates/librefang-kernel/src/kernel/subsystems/mcp.rs:51`). The "empty allowlist == wildcard" branch therefore yields the union of all servers' tools, regardless of which agent is being prepared.

The comment in Step 3 explicitly says "mcp_candidates is already scoped to the agent's allowed servers (via mcp_allowlist above), so no further declared_tools filtering is needed" — that scoping assumption holds for non-empty allowlists only, but the empty case silently widens to everything.

## Reproduction (multi-agent setup)

[...]

PR #6592, opened 2026-07-27
fix(kernel): stop trusted_senders from waiving approval on MCP tools, and document the EveryAPI MCP bridge

houko:
## What

Documents the MCP bridge between LibreFang and [EveryAPI](https://github.com/everyapi-ai/everyapi), and fixes a kernel bug the documentation work exposed: `[approval] trusted_senders` silently waived human approval on every tool a connected MCP server exposes.

## The security fix

`ApprovalManager::classify_risk` matches a closed list of built-in tool names — `shell_exec`, `file_write`, `agent_spawn` and friends — so anything namespaced `mcp_*` fell into the `Low` default. The trusted-sender branch of `requires_approval_with_context` returns as soon as a tool is not high-risk, and that return happens *before* the operator's `require_approval` globs are consulted. So those globs were dead config for exactly the tools whose behaviour the kernel cannot audit: server-side tool code is third-party and its effects are not derivable from a name.

This is not hypothetical. The EveryAPI MCP server documented in this PR exposes `everyapi_seller_withdraw`, which moves money, and `everyapi_seller_add_key`, which takes a plaintext upstream API key as an argument and is the only write tool with no confirm-token friction of its own. An operator who added their own account to `trusted_senders` and gated the write tools with globs got no gating at all.

`mcp_*` now grades as high risk, keeping the bypass scoped to the built-ins it was written for. Someone who genuinely wants an MCP tool running unattended can still allow it through a channel rule or a per-user policy; what they can no longer do is disable the gate for every MCP tool at once without saying so.

Two adjacent bypasses are documented rather than changed, because both are deliberate features rather than oversights: a channel `allowed_tools` rule short-circuits ahead of the `require_approval` list, and a `hand:`-tagged agent auto-approves unless per-user RBAC set `force_human`. The security section names them so nobody reads the glob config as a hard block.

## The docs

A new "Connect the EveryAPI MCP server" section on the MCP/A2A integrations page, in both locales. The integration is genuinely two config stanzas and no code, since EveryAPI ships an MCP stdio server in its own binary and LibreFang is an MCP stdio client. The section covers:

- The `[[mcp_servers]]` entry and the agent-side `mcp_servers` allowlist, including that an empty or omitted list means **no** servers rather than all of them (#5855) — an agent that omits the key gets silence, not tools.
- Why `env = []` is correct here: the stdio transport forwards a fixed safe set including `HOME` and `XDG_CONFIG_HOME`, which is all `everyapi mcp` needs to read its own credentials file, so no credential is copied into LibreFang's config or vault.
- The full 15-tool inventory split read versus write, with `everyapi_seller_add_key` called out specifically.
- That tool names are double-prefixed. LibreFang namespaces as `mcp_{server}_{tool}` and EveryAPI's tools are already `everyapi_*`, so `everyapi_status` becomes `mcp_everyapi_everyapi_status`. An approval glob written against the single-prefix form matches nothing at all — a silent failure worth stating outright.
- Four `require_approval` globs that gate all eight writes and leave the seven reads free, which is what keeps the unattended balance-check use case working. A blanket `mcp_everyapi_*` would stall it on an approval prompt.

## The unparseable examples

Twelve `[[mcp_servers]]` stanzas across six page pairs were in a pre-`transport` schema. `McpServerConfigEntry` is `deny_unknown_fields`, so those are not stylistic slips that degrade gracefully — a user copying one got a hard parse error that takes the entire `config.toml` down, and a daemon that will not boot. Three distinct broken shapes were present, all copy-paste descendants of an older schema.

`crates/librefang-extensions/tests/docs_mcp_servers_examples.rs` now walks every `.mdx` under `docs/src/app`, extracts each stanza from the fenced TOML, and deserializes it through the real type. New pages are covered with no list to maintain.

## Verification

Docker was unavailable on this box, so these ran against the native toolchain with `CARGO_TARGET_DIR` pinned to the worktree's own `target/` — the shared-`target` contention the Docker rule guards against does not apply.

```
cargo test -p librefang-kernel --lib approval::     95 passed / 0 failed
cargo test -p librefang-extensions                  exit 0
cargo clippy -p librefang-kernel --all-targets -- -D warnings   exit 0
cargo fmt --all -- --check                          exit 0
cargo check --workspace --lib                       exit 0
cd docs && pnpm build                               exit 0
```

`cargo test -p librefang-kernel --lib` aborts with SIGABRT when run multi-threaded. That is pre-existing and unrelated to this change: it reproduces identically on a clean tree with the approval.rs edit stashed, and the same suite passes 1323/1323 under `--test-threads=1`. Worth a separate issue; not touched here.

## Not covered

The EveryAPI side of the bridge — a `librefang` target for `everyapi mcp install` — lives in the other repository and needs its own PR.


CHANGELOG.md
```diff
@@ -61,6 +61,9 @@ and this project uses [Calendar Versioning](https://calver.org/) (YYYY.M.DD).
 
 ### Fixed
 
+- Stop `[approval] trusted_senders` from waiving human approval on every tool a connected MCP server exposes.
+`classify_risk` matched a closed list of built-in names, so any `mcp_*` tool fell into the `Low` default and the trusted-sender branch of `requires_approval_with_context` returned before the operator's `require_approval` globs were consulted — the globs were dead config for exactly the tools least able to be audited, since server-side tool code is third-party and its effects are not enumerable from a name.
+Real MCP servers ship tools that move money or take a plaintext credential as an argument, so `mcp_*` now grades as high risk and the bypass stays scoped to the built-ins it was written for; an operator who wants an MCP tool unattended can still allow it per-channel or per-user (#6589) (@houko)
```
houko: This bullet's issue reference looks off. `(#6589)` is the docs-only PR that already merged into `main` — and its own body states the *pre-fix* behavior ("`trusted_senders` waives it for every MCP tool") as the documented, shipped fact. This bullet is describing the kernel fix that only exists in this PR (#6592); citing #6589 attributes the security fix to a PR that shipped without it.

Worth double-checking whether this should read `(#6592)` instead (or whichever issue is actually tracking the kernel change), so the CHANGELOG doesn't point a future reader at the wrong PR when they go looking for the fix.

---

houko (commented):
Automated review pass. Ran cargo check/clippy/fmt and the scoped test suites for librefang-kernel and librefang-extensions against this branch — all green, including the new regression test and the docs-drift guard tests.

Pushed two mechanical fixes directly (both verified via cargo check/clippy/fmt and scoped tests before pushing):
- Moved the `MCP_TOOL_PREFIX` const out of the middle of the `use` block in `approval.rs` so imports stay grouped.
- (Note: while working, the branch was force-pushed/rebased onto latest `main` — I reset onto the new tip. In the process I'd also independently found and drafted a fix for the zh mirror's `trusted_senders` paragraph, which asserted the *pre-fix* vulnerable behavior as current; the rebase already carries an equivalent correction, so no separate commit was needed there.)

One item left as an inline comment: a CHANGELOG issue-number citation that looks like it points at the wrong PR.

---

crates/librefang-kernel/src/approval.rs
houko, line 1: Commit `d7f8cf3d` on this branch has author `Claude <noreply@anthropic.com>` (visible in `git log` and in the commit view). Repo `CLAUDE.md` requires no Claude/Anthropic/AI attribution "in commit messages, PR bodies, or comments" — this is that attribution living in commit metadata rather than the message body, so the `commit-msg` hook's string-match wouldn't have caught it. Purely cosmetic change (moving `MCP_TOOL_PREFIX` below the `use` block) but the authorship needs a human call: squash/rebase with a non-AI author before merge, or confirm this is acceptable as-is.

---

houko (commented):
Daily automated review pass. No changes made to the approval-logic fix or docs — verified correct as-is. One process finding flagged inline (commit authorship).

---

crates/librefang-kernel/src/approval.rs
```diff
@@ -1411,6 +1417,11 @@ impl ApprovalManager {
             }
             "file_write" | "file_delete" | "apply_patch" => RiskLevel::High,
             "web_fetch" | "browser_navigate" => RiskLevel::Medium,
+            // Tools from an MCP server are third-party code whose effects this table cannot enumerate.
+            // The arms above are a closed list of built-ins, so before this arm existed every `mcp_*` tool fell into the `Low` default and the `trusted_senders` bypass waived approval on all of them — including ones that move money or upload a plaintext credential.
+            // The operator's `require_approval` globs were never consulted, because the bypass returns before that check.
+            // Grading them High keeps the bypass to the built-ins it was written for; an operator who genuinely wants an MCP tool unattended can still allow it per-channel or per-user.
+            name if name.starts_with(MCP_TOOL_PREFIX) => RiskLevel::High,
```
houko, line 1424: This closes the `trusted_senders` bypass, but there is a second, wider hole for MCP tools that this fix does not touch: `is_high_risk_tool` in `crates/librefang-acp/src/permission.rs` (used to decide whether the ACP editor's permission modal offers "Allow always").
It is a separate hard-coded list — `shell_exec` / `file_write` / `file_delete` / `apply_patch` / `skill_evolve_*` — and does not know about `mcp_*` at all, so it still classifies every MCP tool as low-risk.
That means an ACP client can offer (or a misbehaving client can just echo) `allow_always` for something like `mcp_everyapi_everyapi_seller_withdraw`, and `remember_decision` persists that into `ApprovalManager::remembered` with no risk check on the write side.
`requires_approval_with_context_for` consults that cache and short-circuits to "no approval needed" *before* it ever calls `requires_approval_with_context` — i.e. before the `classify_risk`/`is_high_risk` check this PR just fixed even runs.
So a single "Allow always" click permanently waives approval for that `(agent_id, tool_name)` pair, independent of `trusted_senders` entirely — the same underlying bug (a hard-coded high-risk list unaware of `mcp_*`), reached through the ACP remember-cache instead of the trust bypass.
Suggest either having `is_high_risk_tool` delegate to `ApprovalManager::classify_risk`, or adding an `mcp_*` arm there too.
Flagging rather than fixing directly since it's a different crate (`librefang-acp`) touching different control flow than this PR's change.

---

houko (commented):
Daily automated review pass on top of the two prior ones already on this thread.
Fixed one mechanical CLAUDE.md nit directly (pushed as 6fc370fd): three new comments in approval.rs (the MCP_TOOL_PREFIX doc comment, the classify_risk match-arm comment, and the new test's doc comment) hard-wrapped mid-sentence; repo convention is one sentence per line, no column limit, so I joined each back onto a single line with no code/behavior change.
The trusted_senders/classify_risk fix itself checks out: is_high_risk (used by both requires_approval_with_context and is_tool_denied_with_context) now correctly routes every mcp_* name to High via the new match arm, so the operator's require_approval globs are consulted for MCP tools again.
Left one finding as an inline comment: a second, broader MCP-tool approval gap in crates/librefang-acp/src/permission.rs that this PR doesn't close (see the inline comment for detail) — reached through the ACP "Allow always" remember-cache rather than trusted_senders, so it bypasses the classify_risk fix entirely once triggered.
Not re-raising the two things already flagged and still open on this thread (the CHANGELOG issue-number citation, and Claude-authored commit metadata) — noting only that the same commit-authorship pattern (author "Claude <noreply@anthropic.com>") recurred on the newest commit at the tip (933e3805) as well as d7f8cf3d, so that's now two commits needing the same authorship cleanup before merge.

---

crates/librefang-kernel/src/approval.rs
```diff
@@ -18,6 +18,12 @@ use totp_rs::{Algorithm, Secret, TOTP};
 use tracing::{debug, info, warn};
 use uuid::Uuid;
 
+/// Namespace prefix every MCP tool name carries.
+///
+/// Mirrors `format_mcp_tool_name` / `is_mcp_tool` in `librefang-runtime-mcp`.
+/// Duplicated as a literal rather than imported because `classify_risk` is a pure string classifier and pulling the MCP crate into this module's dependency path for one prefix is not worth the coupling.
+const MCP_TOOL_PREFIX: &str = "mcp_";
```
houko, line 25: Process finding, not a code-correctness issue — flagging for a maintainer decision before merge.

This branch now has three commits with git author `Claude <noreply@anthropic.com>`:
- `d7f8cf3d` "chore(kernel): move MCP_TOOL_PREFIX const out of the use-statement block"
- `933e38058` "docs: fix PR number self-references in new CHANGELOG entries"
- `6fc370fde` "docs(kernel): unwrap mid-sentence line breaks in new approval.rs comments"

CLAUDE.md requires no Claude/Anthropic/AI attribution "in commit messages, PR bodies, or comments," and the repo's own `commit-msg` hook exists to enforce that — but it string-matches commit *message* content, and this attribution lives in the git *author* field, so it isn't message text the hook would ever see. All three are cosmetic/docs-only changes (no logic touched beyond the fix already in `e9d3066a1`), so there's nothing to re-review functionally, but the authorship on a security-relevant PR still needs an explicit maintainer call: squash/rebase these three with a non-AI author before merge, or confirm the project is fine shipping them as-is. Noting it once here rather than on each commit individually since it's the same root cause.

---

houko, line 25: Separate note on this same constant: `MCP_TOOL_PREFIX` is intentionally a duplicated literal rather than an import of `librefang_runtime_mcp::is_mcp_tool` / `format_mcp_tool_name` — reasonable, since `librefang-kernel` doesn't currently depend on `librefang-runtime-mcp` and the doc comment explains the tradeoff. But nothing pins the two crates' definitions of the MCP namespace prefix together. If `is_mcp_tool`'s convention in `librefang-runtime-mcp` ever changes (different prefix, added version segment, etc.), this constant silently stops matching real MCP tool names and the `trusted_senders` bypass this PR just closed reopens with no failing test anywhere, since the two crates have no dependency edge to hang a shared check on. Worth a workspace-level test asserting the two stay in sync (e.g. `is_mcp_tool(&format!("{MCP_TOOL_PREFIX}x"))` is true, exercised from a crate that already depends on both, or hoisting the prefix into a shared low-level crate both import) — same spirit as the `#3298` deterministic-ordering drift-guard tests CLAUDE.md calls out for other cross-boundary invariants.

---

crates/librefang-kernel/src/approval.rs
houko, line 1424: Correctness check on this fix's completeness (please confirm intentional — this looks security-relevant).

Traced both call sites: `requires_approval_with_context` and `is_tool_denied_with_context` both gate their trusted-sender bypass on `!Self::is_high_risk(tool_name)`, and `is_high_risk` is `classify_risk(...) >= RiskLevel::High`, so this new `mcp_*` arm correctly closes the bypass in both places (the channel-deny bypass too, not just the `trusted_senders` path named in the title) — confirmed against `RiskLevel`'s declared `Low < Medium < High < Critical` ordering. `cargo check -p librefang-kernel --lib` passes clean on this branch.

There is a second, independent risk classifier this PR does not touch: `is_high_risk_tool` in `crates/librefang-acp/src/permission.rs` (gates whether the ACP editor's permission modal offers "Allow always"):

```rust
fn is_high_risk_tool(tool: &str) -> bool {
    matches!(tool, "shell_exec" | "file_write" | "file_delete" | "apply_patch")
        || tool.starts_with("skill_evolve_")
}
```

It has no `mcp_*` awareness, so it still classifies every MCP tool as low-risk. A human (or a misbehaving ACP client) can click "Allow always" on e.g. `mcp_everyapi_seller_withdraw`, which calls `remember()` and persists an `Approved` decision keyed on `(agent_id, tool_name)`. `requires_approval_with_context_for` checks that cache via `recall()` *before* it ever calls into `requires_approval_with_context` — i.e. before the `classify_risk`/`is_high_risk` check this PR fixes even runs. So a single "Allow always" click permanently waives approval for that tool, independent of `trusted_senders` entirely: same underlying bug (a hard-coded high-risk list unaware of `mcp_*`), reached through the ACP remember-cache instead of the trust bypass.

Flagging rather than fixing directly since it's a different crate (`librefang-acp`) with different control flow than this PR's change — but given the PR's stated goal ("the bypass stays scoped to the built-ins it was written for"), this looks like the same class of hole through a different door. Worth a decision: have `is_high_risk_tool` delegate to `ApprovalManager::classify_risk`, add an `mcp_*` arm there too, or confirm it's tracked separately.

---

houko (commented):
Reviewed the kernel fix (`e9d3066a1`) plus the three follow-up commits already on this branch.

Core fix: traced both `requires_approval_with_context` and `is_tool_denied_with_context` — the new `mcp_*` → `RiskLevel::High` arm in `classify_risk` correctly closes the `trusted_senders` bypass in both, and `cargo check -p librefang-kernel --lib` / `-p librefang-extensions --tests` both pass clean on this branch. The CHANGELOG issue-number bullet I initially flagged (#6589 vs #6592) was already fixed upstream in `933e38058` before I could post it, so no comment needed there — pushed one small mechanical fix myself (`fe3dd7922`, CHANGELOG indentation on the same bullet, rebased on top of the two upstream doc fixes).

Left three inline comments for a human decision:
1. Three commits on this branch (`d7f8cf3d`, `933e38058`, `6fc370fde`) carry git author `Claude <noreply@anthropic.com>` — attribution living in commit metadata rather than message text, so the repo's `commit-msg` hook wouldn't catch it. Needs a maintainer call on squashing/rebasing before merge.
2. No cross-crate test pins `MCP_TOOL_PREFIX` (kernel) to `is_mcp_tool`/`format_mcp_tool_name` (runtime-mcp) staying in sync — a future rename of the MCP namespace convention would silently reopen this exact bypass with nothing failing.
3. `crates/librefang-acp/src/permission.rs::is_high_risk_tool` is a second, independent risk classifier (gates the ACP "Allow always" button) that still has no `mcp_*` awareness, so the same class of bypass is reachable through the ACP remember-cache instead of `trusted_senders`. Different crate/control-flow than this PR, flagging rather than fixing.

---

houko (commented):
Reviewed the security fix in `crates/librefang-kernel/src/approval.rs` in detail (diff + fresh checkout).

**The fix itself is correct.** `classify_risk` (line 1415) now grades any `mcp_` prefixed tool name as `RiskLevel::High` via the new `MCP_TOOL_PREFIX` const, and `is_high_risk` (line 1447, `>= RiskLevel::High`) is the sole gate the `trusted_senders` bypass in `requires_approval_with_context` (line 723: `if policy.is_trusted_sender(sid) && !Self::is_high_risk(tool_name)`) checks before returning early — so the fix lands exactly on the code path the PR description identifies, and does so *before* the `require_approval` glob check, matching the described bug ("the globs were dead config").
The `mcp_` prefix matches `librefang_runtime_mcp::is_mcp_tool` (`name.starts_with("mcp_")`) and `format_mcp_tool_name`'s `mcp_{server}_{tool}` output, so the classification boundary is consistent with how tool names are actually namespaced.

**Regression test is present and correct**: `test_trusted_sender_cannot_bypass_an_mcp_tool` (approval.rs ~2652) reproduces the exact scenario from the PR description (a trusted sender + `mcp_everyapi_seller_withdraw` / `mcp_everyapi_seller_add_key`) and also pins that a built-in low-risk tool (`web_search`) keeps its exemption, so the fix is scoped correctly rather than disabling the bypass wholesale.

**Verified locally** (fresh worktree, native toolchain after installing `libdbus-1-dev`/`pkg-config` — Docker daemon wasn't reachable in this sandbox):
- `cargo test -p librefang-kernel --lib approval:: -- --test-threads=1` — 95 passed / 0 failed, including the new test

**One related bypass this PR does not close** (left as a line comment attempt on `crates/librefang-acp/src/permission.rs:302`, but GitHub rejected it because that file isn't part of this PR's diff, so noting it here instead):

`is_high_risk_tool` in `crates/librefang-acp/src/permission.rs` is a second, independently-maintained closed list (`shell_exec`, `file_write`, `file_delete`, `apply_patch`, `skill_evolve_*`) that does not include `mcp_*`.
It gates whether the ACP editor's permission modal offers "Allow always" and whether a client-echoed `allow_always` gets persisted (`sanitize_remember`).
A "yes" there calls `ApprovalManager::remember_decision`, and `requires_approval_with_context_for` (approval.rs ~691-705) checks the resulting `recall()` cache *before* it ever reaches the `classify_risk`/`is_high_risk` check this PR just fixed.
So an ACP editor user who clicks "Allow always" once on an MCP write tool today persists a blanket bypass for that `(agent_id, tool_name)` — the same "third-party tool code the kernel cannot audit" risk this PR closes for `trusted_senders`, reached through a different call site.
I confirmed via grep that `remember_decision` / an "always" option only exists on the ACP path — dashboard and CLI approval flows have no equivalent — so this is the one other place that needs to agree MCP tools are high-risk.
I'm not confident enough in the `librefang-acp` ↔ `librefang-kernel` dependency direction to propose the exact fix (shared helper vs. inline prefix check) myself, so flagging for the author/a maintainer rather than guessing on a security-relevant path.

No other findings: no `HashMap`/`HashSet` touching an LLM prompt in this diff, no config-field additions (the new item is a `const`, not a config field), commit messages are conventional (`fix(kernel):`, `chore(kernel):`) with no AI attribution, and the CHANGELOG/docs edits read as accurate corrections of the pre-existing "documented bypass" language now that the bypass is fixed.

---

PR #6588, opened 2026-07-27
release: v2026.7.27

houko:
## Release v2026.7.27

_33 PRs from 2 contributors since v2026.7.21._

### Highlights

- **EveryAPI integration** — connect EveryAPI as a model provider via `librefang models connect everyapi` from the CLI or the new Providers page connect action, with a built-in wiring doctor check
- **macOS boot-time service** — `service install --system` installs a LaunchDaemon so LibreFang starts automatically at login without a running user session
- **NixOS & additional Linux distros** — first-class NixOS deployment support with deepin/Debian distro awareness out of the box
- **Audit log integrity** — audit entries are now preserved (WORM-protected) when an agent is purged, preventing accidental evidence loss
- **Tool result and context fidelity fixes** — stops lossily truncating tool results in the spill dead band, and correctly honors `context_window` from `agent.toml` on all turn-execution paths

### Added

- Emit a failure metric for media-understanding (vision/STT) (#6538) (#6551) (@houko)
- First-class NixOS deployment and deepin/Debian distro awareness (#6582) (@houko)
- Add `librefang models connect everyapi` and an EveryAPI wiring doctor check (#6583) (@houko)
- Add `service install --system` for a boot-time LaunchDaemon on macOS (#6584) (@houko)
- Add an EveryAPI connect action to the Providers page (#6586) (@houko)

### Fixed

- Override sharp to >=0.35.0 to clear the libvips high advisory (#6546) (@houko)
- Include openrouter-models.snapshot.json in the flake source (#6547) (@houko)
- Anchor credential-prefix secret patterns to stop false positives (#6541) (#6548) (@houko)
- Make skills/reload honest in frozen Stable mode (#6540) (#6549) (@houko)
- Stop lossily truncating tool results in the spill dead band (#6545) (#6550) (@houko)
- Bump next to 16.2.11 to clear the App Router security advisories (#6557) (@houko)
- Don't delete audit_entries when purging an agent (WORM integrity) (#6553) (#6558) (@houko)
- Apply busy_timeout to every pooled connection in modify() test helper (#6561) (@houko)
- Treat blank parent_id / agent_id as absent instead of 404 (#6577) (@houko)
- Collect new_name on clone, reflect mcp_servers grants in Tools tab (#6578) (@houko)
- Resolve callback message_id from either metadata shape or the native id (#6579) (@houko)
- Honour agent.toml context_window on the paths that run a turn (#6580) (@houko)
- Install and search from the synced registry checkout (#6581) (@houko)
[...]

PR #6602, opened 2026-07-28
fix: audit sweep over #6577-#6601 — EveryAPI seam, skill-install symlinks, macOS LaunchDaemon, NixOS bind auth

houko:
An audit sweep over the last 20 merged PRs (#6577–#6601). This is deliberately one PR rather than one-per-issue: it is a single review pass, and splitting nine two-to-twenty-line fixes across nine branches would cost more review attention than it saves. Bullets are grouped by area below.

Method: eight reviewers covered the range (the EveryAPI cluster as a cross-PR seam, the four dependabot bumps, #6584 and #6582 individually, release artifacts, the six ordinary fix PRs, and a mechanical omissions sweep against the repo's own invariants). Every candidate finding was then handed to an independent verifier prompted to refute it, defaulting to refuted when uncertain, and checked against current source rather than the diff. Fifteen candidates went in, ten survived (two of which were the same defect found independently by two reviewers), five were dropped — listed at the bottom so the reasoning is on the record.

## EveryAPI cross-PR seam (#6583, #6586)

- **Dashboard connect stored a base URL that cannot work.** `EVERYAPI_PROVIDER.defaultBaseUrl` was `https://api.everyapi.ai`, sitting under a doc comment asserting "the drivers append the path themselves". No driver does: `openai.rs` builds `{base_url}/chat/completions` and `everyapi_catalog.rs` builds `{base_url}/models`, and nothing on the write path normalizes the value. Connecting from the dashboard with the gateway field blank therefore produced a registered provider whose model fetch 404s. Because that failure is only `warn!`-logged and then throttled per base URL, the entry sat configured-with-zero-models indefinitely — while the identical `librefang models connect everyapi` flow, whose `derive_base_url` appends `/v1`, worked. Two UIs for one feature, one of which never yields a usable provider.
- **Neither channel read path refreshed the EveryAPI catalog.** `list_models_by_provider` and `list_models_text` in `channel_bridge.rs` each refresh for `openrouter` and fell through for `everyapi`, so a Telegram or Slack user listing or picking a model saw a stale snapshot while every `providers.rs` read path refreshed correctly. `list_models_text` spans every provider and so refreshes both catalogs unconditionally; the per-provider picker only refreshes the one being asked about. (The reviewer's finding named only the picker — the second site came out of a broad scan of the file.)

The user-typed gateway field is deliberately still stored verbatim — a self-hosted gateway mounted at a different path is legitimate, and a conditional `/v1` append would be a second, divergent normalization rule.

## Skill installation (#6581)

- **Symlink hardening covered one of three installers.** #6581 hardened `librefang_skills::marketplace::copy_dir_recursive`, but the API route's helper (`routes/skills/mod.rs`) and the CLI's (`commands/common.rs`) read the same registry checkout. The former dereferenced links through `std::fs::copy`; the latter branched on `Path::is_dir()`, which follows them. A symlink planted in a registry skill still copied the target's real contents — or an entire external tree — into the installed skill. Both now skip symlinks, matching the hardened helper, each with a regression test covering the file and the directory case.

## Dashboard tool grants (#6578)

- **Tools tab ignored `tool_allowlist`.** The tab mirrored `tool_blocklist` only. The kernel's Step 4 filter (`tools_and_skills.rs`) runs *after* MCP tools are pushed into the candidate set, so a non-empty allowlist naming no `mcp__*` glob strips the whole server (#6495). An allowlist-filtered server rendered as fully granted — the UI asserted an agent could call tools the kernel had already removed. Adds `isToolAllowed` mirroring the kernel's empty-list-is-unrestricted semantics, with tests. The four locale strings that described MCP tools as filtered "by tool_blocklist" are corrected too; that sentence is where the wrong mental model was written down.

## macOS service install (#6584)

- **`--system` did not check for the per-user LaunchAgent.** Both carry the `ai.librefang.daemon` label and the same state directory, so installing the LaunchDaemon over an existing agent produced a login-time respawn loop: the agent's `start --foreground` finds the daemon already holding the port, exits non-zero, `KeepAlive` relaunches it. The install now refuses with a remediation hint, and does so before the ownership handover and the plist write, so a refused install leaves nothing behind.
- **Neither plist renderer escaped XML metacharacters.** APFS permits every byte but `/` and NUL, so a volume named `Backup & Media` or a `LIBREFANG_HOME` containing `<` produced an ill-formed plist that launchd rejects wholesale — while the install path still reported it as written. Both the LaunchAgent and LaunchDaemon renderers now escape; tests assert no raw metacharacter survives in any `<string>` node.

## NixOS module (#6582)

- **The bind-auth assertion accepted the wrong evidence.** It treated `environmentFile != null` as proof of authentication, but `any_auth_configured` reads `api_key` from `<stateDir>/config.toml` only — no environment variable feeds it. The only credentials the unit environment can supply are `LIBREFANG_DASHBOARD_USER` / `LIBREFANG_DASHBOARD_PASS`. The documented off-host recipe, whose environment file is annotated "must export an API key", therefore passed `nixos-rebuild` and deployed a unit the daemon refuses to start. The assertion message now names what actually works, a new `authConfiguredExternally` option covers a `config.toml` maintained out-of-band, and the docs recipe is corrected.
- **`LIBREFANG_ALLOW_NO_AUTH` was a presence check.** `cfg.extraEnvironment ? LIBREFANG_ALLOW_NO_AUTH` accepted `"0"`, which `allow_no_auth_env()` rejects — eval passes, boot fails. Now checked against the same five spellings the daemon accepts.

Nix expressions were not evaluated locally (no `nix` on this host); CI covers them.

## Release artifacts (#6588)

- **The v2026.7.27 article shipped with no YAML front matter.** Both publish paths key off it: `devto-publish.yml` skips an article with no `title`, and `release-notify.yml` extracts the Discussion body with `awk '/^---$/{c++;next} c>=2{print}'`, which prints nothing when the counter never reaches two. `scripts/changelog-to-article.sh` emits the block correctly — this article was written without it, so the fix is the article, not the generator.
- **A highlight inverted the new service's behaviour**, saying `service install --system` starts LibreFang "at login without a running user session". It is a boot-time LaunchDaemon; the authoritative entry a few hundred lines above already said so. Corrected in `CHANGELOG.md` and the article. This does not retroactively fix the published GitHub Release body, which would be a manual `gh release edit`.

## CHANGELOG hygiene

The `[Unreleased]` section had accumulated **16 subsection headers where 5 distinct kinds exist** — seven `### Added`, three `### Fixed`, three `### Changed`, two `### Documentation` — because each PR appended its own block instead of merging into the existing one. #6587 fixed one instance of this by hand last week.

Same-named blocks are now concatenated into a single header each, ordered Added / Fixed / Changed / Security / Documentation to match the layout every released section from 2026.6.11 onward already uses. No entry text was edited, no entry was reordered within its group, and nothing was dropped: the multiset of 285 content lines and 143 bullets is identical to before the merge, verified independently against `git show HEAD:CHANGELOG.md` rather than only inside the script that performed the rewrite.

Only `[Unreleased]` was touched. Released sections are left alone, including the older ones whose subsection order predates the current convention.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test -p librefang-cli --bins` — 346 passed
- `cargo test -p librefang-api --lib` — 858 passed
- dashboard `vitest run` — 894 passed across 87 files; `tsc --noEmit` clean; `eslint` 0 errors, warning count unchanged from base
- `scripts/check-changelog-attribution.py` — clean

The dashboard's Playwright suite under `crates/librefang-api/dashboard/e2e/` is **not run by any CI workflow** (`test-web.yml` drives `web/`, a different directory). This PR edits `everyapi-connect.spec.ts`, so it was run locally: both cases fail — but they fail identically on an unmodified tree, at `getByRole("link", { name: "Providers" })` during navigation, well before reaching any assertion this PR touches. That is a pre-existing breakage in an unwired suite and is left as-is rather than folded into an audit PR; worth a separate issue on whether the suite should be in CI at all.

Nix expressions were not evaluated locally (no `nix` on this host); CI owns them.
- `cargo fmt --all --check` — clean

New tests: `copy_dir_recursive_skips_symlinks` in both `routes/skills` and `commands/common`; `xml_escape_escapes_ampersand_before_angle_brackets` and `macos_system_plist_escapes_xml_metacharacters_in_paths`; `isToolAllowed` suite in `toolGrants.test.ts`.

## Findings dropped during verification

Recorded so the same ground is not re-covered:

- `psl` 2.1.219 → 2.1.223 was claimed to widen the MCP OAuth same-eTLD+1 rule via dropped shared-hosting suffixes — not substantiated against the actual list diff.
- The `--system` ownership handover was claimed to follow a symlinked state root; `collect_ownership_handover_paths` already uses `DirEntry::file_type` and `lchown`, both non-following.
- #6589 was flagged as missing from the 2026.7.27 changelog section despite being an ancestor of the tag — it is a docs-only PR and the omission is intentional.
- #6580 was claimed to have left two compaction callers on the old `context_window` contract — both were verified to route through `resolve_context_window`.
- `install_id` was claimed to be missing from `GET /api/skills/registry` — the dashboard does not read that field from that endpoint.

Existing `providers/everyapi.toml` files already written by the dashboard with the suffix-less base URL stay broken — this fixes the write path, not previously-written state. A repair-on-load migration is a genuinely separate change.

The four dependabot bumps (`base64` 0.23, `jsonwebtoken` 11, `serial_test` 4, and the ten-crate minor group) were reviewed specifically for silent behaviour change at call sites — vault-key decoding, JWT validation defaults, and test serialization grouping. Nothing survived verification there.


crates/librefang-api/src/channel_bridge.rs
houko, line 1: This adds the `everyapi_catalog::refresh_if_stale` call to both `list_models_text` and `list_models_by_provider`, mirroring the existing `openrouter_catalog` call — logic looks correct and matches `everyapi_catalog.rs`'s real `refresh_if_stale` signature.
No test in `crates/librefang-api/tests/` (or a `#[cfg(test)]` module in this file) exercises either function for either provider, so this specific wiring fix — and the sibling `openrouter` call it mirrors — has no regression coverage; a future refactor of `ChannelBridgeHandle` could silently drop the refresh again.
Given the CLAUDE.md rule that wiring changes should carry a regression test, worth a follow-up unit test against a mocked `KernelApi`/model catalog asserting both catalogs get refreshed. Not blocking since it mirrors a pre-existing (also untested) pattern, but flagging since this PR's own changelog entry describes exactly the kind of silent-drift bug a test here would catch.

---

houko (commented):
Automated audit sweep of PR #6602. Cross-checked every code claim in the diff and CHANGELOG against the actual source (xml_escape coverage in both macOS plist renderers, symlink-skip consistency across the three skill installers, NixOS assertion wording against server.rs, EveryAPI base-URL/`/v1` contract against the CLI and OpenAI driver, PR citations in the merged CHANGELOG section). Pushed one small fix; left one non-blocking test-coverage note inline.

---

CHANGELOG.md
```diff
@@ -58,9 +58,42 @@ and this project uses [Calendar Versioning](https://calver.org/) (YYYY.M.DD).
...
+- Add opt-in completion notification for `process_start` background processes via the #4983 async-task tracker: a call with `notify_on_completion: true` registers a new `TaskKind::Process { pid }`, and when the process exits on its own or is killed the kernel injects a `TaskCompletionEvent` (exit code + tail of the captured output) back into the originating session, so a long-running process no longer has to be polled to learn it finished — reusing the same delivery path (mid-turn injection / wake-idle) as workflow / delegation tracking with no new config surface; note the `[async_tasks]` `default_timeout_secs` auto-kill is deliberately NOT applied to processes (a background server is meant to run indefinitely), only the completion-delivery machinery is shared (#6471) (@houko)
```
houko, line 65: Lines 64-65 are two near-duplicate `process_start` completion-notification bullets, both tagged `(#6471) (@houko)`.
They pre-date this PR (already present as duplicates on `origin/main` under two separate `### Added` headers, per `git show origin/main:CHANGELOG.md`), so the header-merge pass correctly preserved them byte-for-byte per the PR body's stated invariant ("no entry text was edited... nothing was dropped").
Flagging rather than fixing because removing one would touch that explicit verification claim — worth a maintainer call on whether to collapse to the more detailed line 65 version (it additionally documents that `default_timeout_secs` auto-kill does not apply to processes) as a follow-up, or leave as-is since it predates this sweep.

---

houko (commented):
Automated review pass over the diff against merge-base, applying CLAUDE.md's prose-wrap, CHANGELOG-hygiene, deterministic-ordering, and attribution rules. One finding left for a maintainer call (inline comment on CHANGELOG.md); everything else checked out (headers deduped correctly, no duplicate `### ` blocks, attribution present on every new/touched bullet, no new HashMap/HashSet on prompt-facing types, no AI attribution). Also pushed a small fix commit for a few mid-sentence hard-wraps this PR's own new comments introduced.

---

Issue #6595, opened 2026-07-27
`max_iterations` in a HAND.toml agent silently converts a reactive hand into a 30-second autonomous tick loop


nevgenov:
## Symptom

A hand whose manifest declares `schedule` reactive (or declares no schedule at all) starts firing `[AUTONOMOUS TICK]` turns every 30 seconds after activation. Nothing in the HAND.toml asks for autonomy — no `[autonomous]` section, no continuous schedule. The materialized workspace manifest ends up with an `[autonomous]` block the author never wrote.

The only field responsible is `max_iterations` on the agent, which reads as a per-turn tool-loop cap. The bundled `researcher` hand states that intent explicitly in its own comment next to the value:

```toml
[agents.main]
max_iterations = 80
# Raise the history cap above the kernel default. Deep research workflows do
# extensive web_search → web_fetch → summarize
# loops with multi-source synthesis: 80 iterations × ~4 messages each
# easily produces 200+ messages per user turn.
max_history_messages = 120
```

That is a loop-depth cap, and `docs/operations/config-reload.md` documents the operator-facing equivalent the same way — "Operator override for the agent-loop iteration cap". Nothing suggests it enables autonomous scheduling.

## Root cause (read from current source)

**Step 1** — `crates/librefang-hands/src/lib.rs:327-330`: presence of `max_iterations` constructs an `AutonomousConfig`:

```rust
autonomous: legacy.max_iterations.map(|max_iter| AutonomousConfig {
    max_iterations: max_iter,
    ..Default::default()
}),
```

`..Default::default()` brings `heartbeat_interval_secs: 30` (`crates/librefang-types/src/agent.rs:159`).

**Step 2** — `crates/librefang-kernel/src/kernel/hands_lifecycle.rs:344-355`: a `Some(autonomous)` then rewrites the schedule:

```rust
// Autonomous scheduling: only override if agent doesn't already have
// a non-default schedule (respect agent-level schedule config)
if manifest.autonomous.is_some() && matches!(manifest.schedule, ScheduleMode::Reactive)
{
    manifest.schedule = ScheduleMode::Continuous {
        check_interval_secs: manifest
            .autonomous
            .as_ref()
            .map(|a| a.heartbeat_interval_secs)
            .unwrap_or(60),
    };
}
```

[...]

Issue #6594, opened 2026-07-27
Hand activation hardcodes exec_policy.mode = Full, overriding operator's global `deny` and skipping the approval gate


nevgenov:
## Symptom

Operator config sets a deliberately restrictive shell posture:

```toml
[exec_policy]
mode = "deny"

[approval]
auto_approve = false
auto_approve_autonomous = false
require_approval = ["shell_exec", "file_delete", "apply_patch"]
```

Every regular agent on the install resolves to `deny` / `allowlist` / no policy, as expected. But an activated Hand runs `shell_exec` **unrestricted and with no approval prompt** — its materialized manifest carries `mode = "full"`, which appears nowhere in operator config and nowhere in the Hand's own `HAND.toml`.

In our case a stale Hand instance issued ~320 `shell_exec` calls over one morning with zero approval prompts, including repeated `<daemon-cli> stop` / `start` / `restart`. Each restart bounced the shared daemon and every channel adapter on it; some adapters came back with a broken outbound path, so agent turns completed while nothing was actually delivered.

## Root cause (read from current source)

**1. Hand activation hardcodes `Full`** — `crates/librefang-kernel/src/kernel/hands_lifecycle.rs:357-365`:

```rust
// Shell exec policy: only set if agent doesn't already have one
if manifest.exec_policy.is_none() && def.tools.iter().any(|t| t == "shell_exec") {
    manifest.exec_policy = Some(librefang_types::config::ExecPolicy {
        mode: librefang_types::config::ExecSecurityMode::Full,   // <-- global [exec_policy] never consulted
        timeout_secs: 300,
        no_output_timeout_secs: 120,
        ..Default::default()
    });
}
```

`exec_policy.is_none()` is treated as "grant Full", not "inherit the operator's global policy". A Hand whose definition lists `shell_exec` therefore escalates past a global `mode = "deny"` silently.

**2. `Full` disables command validation entirely** — `crates/librefang-runtime/src/subprocess_sandbox.rs:701-708`:

```rust
ExecSecurityMode::Full => {
    tracing::warn!(
        command = crate::str_utils::safe_truncate_str(command, 100),
        "Shell exec in full mode — no restrictions"
    );
    Ok(())
}
```

**3. `Full` additionally short-circuits the approval queue** — `crates/librefang-runtime/src/tool_runner/dispatch.rs:1620-1621` and `1708-1712`:

```rust
let shell_exec_full_mode = tool_name == "shell_exec"
[...]

PR #6610, opened 2026-07-28
fix(hands): require an explicit [autonomous] or schedule declaration to start a hand loop

houko:
## What this corrects

This corrects the `#6595` half of #6603 (merged as `c4d4473b`), which did not fix its own issue.

That change derived a hand role's wake-up schedule from `[metadata] frequency` ANDed with `manifest.autonomous.is_some()`.
The second half of that predicate is true for a role that declared nothing but `max_iterations`, because `From<LegacyHandAgentConfig> for AgentManifest` synthesizes an `AutonomousConfig` purely to carry the loop cap.
So the merged predicate is the original bug's predicate merely ANDed with a second condition.

Concretely, on main today the bundled `researcher` hand declares `[metadata] frequency = "continuous"` and `[agents.main] max_iterations = 80` — with an author comment sizing `max_history_messages` off "80 iterations × ~4 messages each", so the field is plainly a per-turn loop-depth cap — and it still materializes `ScheduleMode::Continuous { check_interval_secs: 30 }`.
That is verbatim the symptom #6595 reports, on the hand the issue names.

#6595 also states that `[metadata] frequency` is never consumed by the kernel and is catalog-display metadata only.
#6603 promoted it to a load-bearing scheduling control; this change reverts that promotion and implements the issue's own prescription instead.

The `#6594` exec_policy half of #6603 is correct and is untouched here.

## Accepted semantics

- An explicit `schedule` on the role wins, honoured verbatim, and remains the only route to the `Periodic` (cron) and `Proactive` variants.
- An explicit `[autonomous]` block yields `ScheduleMode::Continuous` at that block's own `heartbeat_interval_secs` (30s if the block omits it).
- `max_iterations` alone is a loop-depth cap — the ceiling on tool-call iterations within a single turn — with no scheduling effect whatsoever.
- `[metadata] frequency` is catalog-display metadata for the marketplace listing and does not control scheduling, at any value.
- Writing both `schedule` and `[autonomous]` is allowed and the `schedule` decides the cadence; `schedule = "reactive"` alongside an `[autonomous]` block is how an author keeps the guardrails while switching the wake-up cycle off.

## Why the mechanism sits at parse time

`manifest.autonomous` cannot answer "did the author ask for autonomy?" — it is also the carrier for the flat `max_iterations` field, so the declared block and the synthesized block are the same `Option<AutonomousConfig>` after deserialization.
The raw TOML table can answer it: an `autonomous` key exists only if somebody wrote `[autonomous]`.
So the resolution lives in `librefang-hands::apply_explicit_autonomous_schedule`, called from `parse_single_agent_section` — the single funnel every entry point converges on (`parse_hand_toml`, the `HandTomlWrapper` form, `parse_hand_toml_with_agents_dir`, `install_from_content`, `reload_from_disk`, and `impl Deserialize for HandDefinition` itself).

Reading the raw table is also what makes the behaviour identical across the two agent formats.
A fix threaded through `LegacyHandAgentConfig` would cover only the flat shape; a role with an `[agents.X.model]` sub-table takes the `AgentManifest::deserialize` branch instead.
Both are covered by tests, and gating the call on `!has_model_table` was mutation-checked to confirm the nested test is non-redundant.

`hands_lifecycle.rs` no longer writes `manifest.schedule` at all — the whole `frequency`-driven derivation block is deleted, 61 lines.
Hand roles now honour `schedule` exactly the way `spawn_agent_inner` already honours it for a standalone agent.

## The loop cap still resolves

`autonomous.max_iterations` remains the carrier, so #6595's own "minimal" suggestion of leaving `manifest.autonomous` as `None` was **not** followed.
`librefang_runtime::agent_loop` resolves the cap from `manifest.autonomous.max_iterations` and no other manifest field carries it, so dropping the block to stop the ticking would silently discard the cap the hand author asked for.
Every reactive-schedule test in this PR asserts the surviving cap as its second assertion, precisely to rule that shortcut out.

## Removals

`HandFrequency::tick_interval_secs`, `HandFrequency::declares_wake_up_cycle` and `PERIODIC_TICK_INTERVAL_SECS` were introduced by #6603 and are dead under this design.
`git grep` confirms their only consumers on `cb7608d9` were `librefang-hands/src/lib.rs` itself and the deleted `hands_lifecycle.rs` block; no third party appeared, and nothing dangling remains in code or docs.
The only test removed is the one covering the two deleted methods.

The `reactive` serde alias on `HandFrequency::OnDemand` survives, with its test.
#6595 records an operator reaching for that spelling while hunting the ticks and hitting a TOML parse error that failed the whole hand over a synonym.
Its doc comment and assertion message were reworded to stop claiming scheduling relevance, since the alias only ever fixed the parse.

## BEHAVIOUR CHANGE

Hands that were implicitly ticking stop ticking.
This is a behaviour change, not only a bug fix.

Migration for a role that genuinely wants a loop — add one of these to its HAND.toml:

```toml
[agents.main.autonomous]
max_iterations = 80
heartbeat_interval_secs = 30
```

```toml
[agents.main.schedule.continuous]
check_interval_secs = 30
```

Nothing else changes: the loop cap, quiet hours, restart caps and every other `[autonomous]` field keep working exactly as before whether or not the role is scheduled.

## Registry impact

Measured against the pinned snapshot at `crates/librefang-runtime/tests/fixtures/registry/hands/` — 17 hands, 59 roles.
Not one role declares an `[autonomous]` block or a `schedule` today; 20 carry a flat `max_iterations`.

**14 roles change behaviour relative to main (`cb7608d9`):**

| hand/role | `[metadata] frequency` | interval on main | after |
|---|---|---|---|
| `analytics/main` | continuous | 30s | reactive |
| `apitester/main` | continuous | 30s | reactive |
| `browser/main` | continuous | 30s | reactive |
| `collector/main` | continuous | 30s | reactive |
| `devops/main` | continuous | 30s | reactive |
| `devops/implementer` | continuous | 30s | reactive |
| `lead/main` | continuous | 30s | reactive |
| `predictor/main` | continuous | 30s | reactive |
| `researcher/main` | continuous | 30s | reactive |
| `strategist/main` | continuous | 30s | reactive |
| `trader/main` | continuous | 30s | reactive |
| `reddit/main` | hourly | 3600s | reactive |
| `twitter/main` | hourly | 3600s | reactive |
| `linkedin/main` | daily | 86400s | reactive |

**20 roles change behaviour relative to release `2026.7.27`**, since #6603 is unreleased: the 14 above plus the 6 that #6603 already stopped — `clip/main`, `creator/main`, and all four `wiki` roles (`main`, `ingestor`, `analyst`, `linter`), all `frequency = "on-demand"` and all ticking at 30s before #6603.
The CHANGELOG bullet uses 20 because that is the delta a released-version reader sees; both baselines are named here so the two documents do not look contradictory.

**Does anything start ticking?** No role in the registry.
The new condition (explicit `[autonomous]` or explicit `schedule`) is unsatisfied by all 59.
The rule is not a strict subset of main's in the abstract — a flat-format role that *did* write an explicit `[autonomous]` block on an `on-demand` hand would newly tick, because main's predicate required `frequency != on-demand` and the new one does not, and because that block used to be dropped by the flat-format parse entirely (fixed by the #6594 passthrough in #6603).
No fixture has that shape, so nothing is affected in practice.

**Three roles are indeterminate from this repo:** `devteam/pm`, `devteam/engineer`, `devteam/qa` reference `base = "planner"` / `"coder"` / `"code-reviewer"`, and only `hello-world` exists under `fixtures/registry/agents/`.
At the fixture level the fact is stronger than "indeterminate": `parse_multi_agent_entry` hard-errors when the template file is unreadable, so in the in-repo snapshot `devteam` fails to parse and is skipped by `reload_from_disk` altogether.
Whether the shipped templates declare `max_iterations` or `[autonomous]` has to be checked in the registry repository.

**The shipped registry is a separate repository**, so adding the explicit declarations to those manifests is a follow-up there rather than something this PR can carry.
One related item for whoever picks that up: `devops/HAND.toml:644` prompt prose asserts "The Hand is already `frequency = \"continuous\"`, so this Phase fires once per turn while gates pass" — that sentence needs revisiting alongside the manifest, since the field no longer drives anything.

## Semantics note: hand roles vs standalone agents

An explicit `[autonomous]` block implies `Continuous` for a hand **role**, but not for a standalone agent, whose schedule still comes only from `schedule` (`spawn_agent_inner` derives nothing from `[autonomous]`).
So the same `agent.toml` ticks when a hand names it as a `base =` template — the merged table reaches `parse_single_agent_section` with the `autonomous` key intact — and does not tick when spawned directly.
This is a semantics note, not a regression: those base-template roles already ticked on main via `autonomous.is_some()` + `frequency = "continuous"`, and no role starts ticking as a result.
Extending the rule to standalone agents would start loops for agents that have never had one, which #6595 does not ask for, so it is documented as a deliberate scope boundary in `apply_explicit_autonomous_schedule` rather than smuggled in.

## Verification

| Command | Outcome |
|---|---|
| `cargo fmt --check` | exit 0 |
| `cargo check --workspace --lib` | exit 0, no diagnostics |
| `cargo clippy --workspace --all-targets -- -D warnings` | exit 0, zero `warning:` / `error:` lines |
| `cargo test -p librefang-hands` | exit 0 — 98 lib + 2 `registry_smoke`, all passed |
| `cargo test -p librefang-kernel --lib hand_activation` | exit 0 — 13 passed (includes all five #6594 exec_policy tests, unchanged) |
| `cargo test -p librefang-api --test hands_routes_integration -- --test-threads=1` | exit 0 — 36 passed, 173s |

Kernel lib test count goes 1335 → 1336 (net +1 kernel test).

**The unfiltered kernel lib suite cannot complete on this machine, on this branch or on the base — and that is environmental, not this change.**
`cargo test -p librefang-kernel --lib` aborts with `signal: 6, SIGABRT`, no panic output and no failing test name, which is not the shape of an assertion failure.
Evidence it is pre-existing: `git switch --detach cb7608d9`, genuine recompile of `librefang-kernel` from main's source, same command — aborted after 704 passing tests, against 705 on this branch. `--test-threads=2` does not avoid it; both figures are from that flag.

Evidence it is resource starvation rather than a defect: `hands_routes_integration` aborted the same way at default parallelism after 24 of 36 tests, then passed 36/36 at `--test-threads=1`.
Every one of these tests boots a full kernel, and the host was at load average 34.8 with a live `librefang start --foreground` daemon of its own during the runs.
Reduce the concurrency and the same binaries go green.

CI's nextest lanes run each test in its own process and are the gate for the full suite.

**No other crate is implicated.** `crates/librefang-runtime/tests/` contains no `ScheduleMode` reference and no test that reads the hands fixture registry; the fixture registry is seeded into `librefang-api` tests, and the `ScheduleMode` assertions in `agents_routes_integration.rs` are all on standalone agents created through the API, not on registry-derived hand roles.

### Tests inverted (they encoded the rejected behaviour)

| Old name | New name |
|---|---|
| `hand_activation_derives_continuous_schedule_from_hand_frequency` | `hand_activation_keeps_reactive_schedule_for_loop_cap_on_a_continuous_hand` |
| `hand_activation_derives_fixed_cadence_from_hourly_and_daily_frequency` | `hand_activation_schedule_ignores_hand_frequency` |
| `hand_activation_honours_explicit_continuous_schedule` | `hand_activation_honours_explicit_schedule_over_explicit_autonomous` |

The first now asserts `Reactive` plus a surviving `max_iterations == 80`, on a fixture reshaped to `researcher`'s exact shape.
The second is a table over `continuous / hourly / daily / periodic / on-demand / no [metadata] block`, all asserting `Reactive` plus the surviving cap — exhaustive over `HandFrequency` on purpose, so adding a variant without deciding what it means for scheduling shows up there.
The third keeps its 900s assertion but its fixture's flat `max_iterations` is replaced by an `[autonomous]` block with a conflicting 60s heartbeat, so the winner is a defined choice rather than an absence.

### Tests added

- `librefang-hands::explicit_autonomous_block_yields_continuous_schedule_in_flat_format` — 900s, non-default on purpose (30 is both the `AutonomousConfig` default and the rejected derivation's output, so a 30 assertion would pass under the old behaviour).
- `librefang-hands::explicit_autonomous_block_yields_continuous_schedule_in_nested_model_format` — with a `system_prompt` fixture-sanity assert proving it took the nested branch.
- `librefang-hands::explicit_reactive_schedule_wins_over_explicit_autonomous_block` — the one input the `schedule`-key guard discriminates; without this case the guard looks redundant and would be deleted by a future reader.
- `librefang-hands::schedule_resolution_is_idempotent_across_a_serialize_reparse_round_trip` — guards the fact that `AgentManifest::schedule` has no `skip_serializing_if`; adding one would silently reintroduce #6595 through a re-parse.
- `librefang-kernel::hand_activation_honours_explicit_autonomous_block` — route 2 end-to-end through `install_from_content` → `activate_hand_with_id` → the agent registry, proving the resolved schedule survives activation's manifest rewriting.

### Mutations checked

Source restored bit-identically after each (verified with `cmp` against a pre-mutation snapshot).

| Mutation | Expected to die | Result |
|---|---|---|
| `!table.contains_key("autonomous")` → `manifest.autonomous.is_none()` (restores #6603's predicate) | the crux distinction | hands 2 failed, kernel 4 failed |
| delete `\|\| table.contains_key("schedule")` | the explicit-reactive case only | `explicit_reactive_schedule_wins_over_explicit_autonomous_block` + round-trip failed; the kernel `..._over_explicit_autonomous` test correctly did not move, since guard 3 already catches it |
| hardcode `check_interval_secs: 30` | both 900s assertions | 3 hands + 1 kernel failed |
| gate the call on `!has_model_table` (flat only) | the nested test only | exactly 1 failed |

## Issue state — a request, not an action

#6595 was auto-closed as `COMPLETED` by #6603's merge while its symptom persisted on main.
It currently shows as `REOPENED`, so that has already been handled — flagging it only so the sequence is on the record.

This PR deliberately uses `Refs #6595` rather than a closing keyword, so merging it will not auto-close the issue.
Closing it is left to you after verifying `researcher` no longer ticks, since a closing keyword is exactly what produced the wrongly-closed state last time.

## Out of scope — questions for you

Two items surfaced by the `#6594` review that live in `librefang-runtime`, a different crate from the one this PR touches. @nevgenov has since answered both in the PR thread — endorsing the first and correcting my premise on the second — so these are now questions of whether to take them, not of whether they are real:

- `crates/librefang-runtime/src/tool_runner/dispatch.rs:1620-1621` and `1708-1712`: `exec_policy.mode == Full` sets `shell_exec_full_mode`, and `skip_approval_for_full_exec` then short-circuits the approval queue for `shell_exec`. The `!force_approval` conjunct preserves a user-policy `NeedsApproval`, but a global `require_approval` entry for `shell_exec` is bypassed whenever the effective mode is `Full`. Is that the intended contract, or should #6594's "never hand out a stronger posture than configured" reasoning extend here?
- `crates/librefang-runtime/src/dangerous_command.rs:123-126`: the dangerous-pattern list catches `\b(python[23]?|perl|ruby|node)\s+-[ec]\s+` but has no entry for the daemon lifecycle verbs, so `librefang stop` / `start` / `restart` pass unimpeded. Worth adding them, plus direct writes to the daemon's own SQLite? **Corrected:** an earlier revision of this section claimed `subprocess_sandbox.rs:701-708` performed no dangerous-pattern screening under `Full`. That was wrong — it is the allowlist validator, which under `Full` correctly has nothing to do. The screen is the separate gate at `tool_runner/dispatch.rs:769-773`, whose comment states it "Runs in Manual mode for all exec policies (including Full)", and @nevgenov confirmed from the #6594 incident that it does fire there. So the gap is the pattern list's coverage, not the absence of a mechanism, and adding to the existing list beats introducing a second one.

One item this change made more pressing, now resolved:

- `crates/librefang-kernel/src/heartbeat.rs`: **done in this PR** rather than asked about, after @nevgenov confirmed it. The monitor skipped agents on `autonomous.is_none()` while its own comment said it skips passive agents; `autonomous.is_some()` was only ever a proxy, and it held while activation derived `Continuous` from the mere presence of the field. This branch makes `autonomous` mean "carries a loop cap, and possibly guardrails", so the proxy inverts and every one of the 20 roles carrying `max_iterations` would be reported `BecameUnresponsive` for behaving as configured. The predicate is now `matches!(manifest.schedule, ScheduleMode::Reactive)`. My earlier note called this a different crate — it is not, `heartbeat.rs` is in `librefang-kernel`, the same crate this PR already changes.

## Credit

Diagnosis by @nevgenov, who filed #6595 with the `researcher` reproduction, identified `max_iterations` as the trigger, and pointed out that `[metadata] frequency` is not consumed by the kernel — the fact #6603 got backwards.


nevgenov:
Reading the raw TOML table to tell an author-written `[autonomous]` block from the synthesized one is the right seam — it is the only place the distinction still exists, and it drops the "was this explicit?" flag that would otherwise have to be threaded through two structs. Taking your three questions in order.

## 1. `Full` short-circuiting the approval queue — yes, #6594's reasoning should extend

Our position, and it is the one we put in #6594's proposed fixes rather than something new: "no allowlist filtering" and "no human confirmation" are orthogonal properties that `Full` currently fuses.

The operator on our host had both halves configured — `require_approval = ["shell_exec"]`, `auto_approve = false`, `auto_approve_autonomous = false` — and a single field they never set (`exec_policy.mode`, fabricated by activation) silently voided both. #6603 fixes the fabrication, which removes our specific incident. But the coupling survives it: an operator who *deliberately* sets `Full` for some hand still loses their global `require_approval` for that agent, with nothing anywhere saying so. "I trust this agent to run arbitrary commands" and "I want to be asked first" are positions a person can reasonably hold at the same time.

The `!force_approval` conjunct already establishes the shape of the answer — a user-policy `NeedsApproval` is preserved through `Full`. A global `require_approval` entry is an operator statement of the same kind and reads like it deserves the same treatment.

## 2. No dangerous-pattern screening under `Full` — the always-on denylist you are proposing already exists

Worth checking before you add one, because we think the premise is off by one layer. `subprocess_sandbox.rs:701-708` is the *allowlist validator*, and under `Full` it correctly has nothing to do. The dangerous-pattern screen is a separate gate in `tool_runner/dispatch.rs:770-800`, and its own comment states the contract you are asking about:

```rust
// Dangerous command detection gate.
//
// Runs in Manual mode for all exec policies (including Full) because
// even explicitly-trusted agents should not silently execute commands
// like `rm -rf /` or fork bombs.
```

It demonstrably fires under `Full`: during the incident behind #6594 and #6595, the hand was running with the fabricated `Full` policy and its `python3 -c "import platform; ..."` calls were blocked twice with `shell_exec blocked: dangerous command detected (script execution via -e/-c flag)`.

So the gap is not that screening is absent — it is what the pattern list covers. `dangerous_command.rs:123-126` catches `\b(python[23]?|perl|ruby|node)\s+-[ec]\s+`, while `librefang stop` / `start` / `restart` pass unimpeded. On a host with several agents and channel adapters on one daemon, that is inverted: a diagnostic one-liner scoped to the calling agent is blocked, and a command that terminates every other agent and channel on the box is not. That is the escalation route the hand actually took, ten times in one morning.

Adding the daemon lifecycle verbs (and direct writes to the daemon's own SQLite) to the existing list looks like the smaller and better-placed change than introducing a second mechanism.

## 3. `heartbeat.rs:86` — yes, and the comment above it already argues your case

The code reads:

```rust
// Skip non-autonomous agents — they are passive (wait for messages)
// and should not be flagged as unresponsive by the heartbeat monitor.
if entry_ref.manifest.autonomous.is_none() { continue; }
```

The stated intent is "skip passive agents". `autonomous.is_some()` was only ever a proxy for that, and it is the same conflation this PR removes everywhere else — after this change the field means "carries a loop cap and possibly guardrails", which says nothing about whether the agent wakes itself. `!matches!(manifest.schedule, ScheduleMode::Reactive)` expresses the comment directly.

We have a concrete instance waiting for this. Our sanitized copy of the `researcher` hand currently has `max_iterations` deleted outright, because that was the only way to stop the ticks — which also silently dropped its loop cap from the author's 80 to the library default of 50. Once this lands we intend to restore `max_iterations = 80` with no `[autonomous]` block, i.e. exactly the "guardrails present, schedule reactive" shape, and on current `heartbeat.rs` that agent would be flagged `BecameUnresponsive` for behaving as configured.

Doing it in this PR seems reasonable given it is the change that makes the shape common, but we have no stake in where it lands.

## Testing

Happy to exercise this once it is in a release — six sidecar channels and the `researcher` hand on this host, which is where both defects showed up. Building a PR branch on that box is not something we want to do while it is carrying production bots, so we would be testing the released artefact rather than the branch.

crates/librefang-hands/src/lib.rs
```diff
@@ -395,15 +395,50 @@ fn parse_single_agent_section(value: &toml::Value) -> Result<AgentManifest, Stri
...
             .or_else(|_| AgentManifest::deserialize(value.clone()))
-            .map_err(|e| format!("Failed to parse [agent] section: {e}"))
+            .map_err(|e| format!("Failed to parse [agent] section: {e}"))?
+    };
+
+    apply_explicit_autonomous_schedule(value, &mut manifest);
+    Ok(manifest)
+}
+
+/// Turn an author-written `[autonomous]` block into a continuous schedule (#6595).
```
houko, line 413: CLAUDE.md-compliance note, not a design objection.

CLAUDE.md's Architecture Notes documents a resolution order for `session_mode` (per-trigger/per-job override > agent manifest default > historical persistent) as a named, findable rule. This PR introduces an analogous resolution order for hand-role `ScheduleMode` (explicit `schedule` > explicit `[autonomous]` > reactive), resolved here in `apply_explicit_autonomous_schedule`, and it's easy to conflate the two by name even though they're different mechanisms (one derives a `SessionId`, this one decides whether a background loop starts).

Given the PR's own documented scope — 14-20 bundled roles change behaviour, and the PR body flags `heartbeat.rs:86`'s `autonomous.is_none()` check as now stale against this new resolution — should CLAUDE.md get an Architecture Notes bullet for this resolution order, the way session_mode/cron and trigger-dispatch-concurrency got one? Not asking for a design change, just flagging that the project's own convention of documenting resolution orders isn't followed here yet, and leaving to you whether that's this PR's job or a fast follow-up doc PR.

---

houko (commented):
One more CLAUDE.md item that doesn't have a diff line to anchor on: CLAUDE.md's prose-wrapping rule ("Commit message bodies... One sentence = one line") applies to commit bodies, and both commits on this branch (`f72780f`, `e96993c`) write hard-wrapped paragraphs with several sentences per physical line rather than one sentence per line (verified with `git show --format=%B -s <sha> | cat -A`). I'm not fixing this myself — the branch is already pushed and has an existing comment on it, so rewriting these commit messages needs an amend + force-push, which is outside what I'll do without your explicit sign-off per this repo's own force-push rule. Flagging so you can decide whether to reword before merge (squash-merge would sidestep it entirely, if that's the intended merge strategy).

No other CLAUDE.md violations found: prose-wrapping in the diff's new doc comments and docs/*.mdx pages, the CHANGELOG bullet's `(@houko)` attribution and single `[Unreleased]` header, and the absence of any AI/Claude attribution all check out. No new config struct fields were added, so the `#[serde(default)]`/`Default` gotcha doesn't apply here. `cargo fmt --check` and `cargo check -p librefang-hands -p librefang-kernel --lib` both pass clean on this branch.

---

houko:
Thanks — item 2 corrects a mistake of mine, and I have taken item 3 into this PR.

## 2 — you are right, and my citation was in the wrong layer

I checked before writing this. `subprocess_sandbox.rs:701-708` is the allowlist validator, exactly as you say, and under `Full` it correctly has nothing to do. The dangerous-pattern screen is the separate gate at `tool_runner/dispatch.rs:769-773`, whose own comment states the contract:

```rust
// Dangerous command detection gate.
//
// Runs in Manual mode for all exec policies (including Full) because
// even explicitly-trusted agents should not silently execute commands
// like `rm -rf /` or fork bombs.
```

So screening is present under `Full` and my out-of-scope question was posed against the wrong mechanism. I have corrected that section of the PR body rather than leaving the wrong claim in the record.

The real gap is the pattern list, and your framing of it is the right one: `dangerous_command.rs:123-126` catches `\b(python[23]?|perl|ruby|node)\s+-[ec]\s+` while the daemon lifecycle verbs pass. A diagnostic one-liner scoped to the calling agent is blocked; a command that stops every other agent and channel adapter on the host is not. Adding the verbs to the existing list is smaller and better-placed than a second mechanism, so that is what I will propose — as its own change against `librefang-runtime`, since it is a different crate from this PR and touches a security denylist that deserves review on its own diff.

## 3 — done in this PR

`heartbeat.rs` is in `librefang-kernel`, the same crate this PR already changes, so my "different crate, which is why I am asking" was also wrong. The predicate is now `!matches!(manifest.schedule, ScheduleMode::Reactive)`, which says what the comment above it always said.

Mutation-checked rather than assumed: reverting the predicate to `autonomous.is_none()` fails the new test with `expected only the continuous agent to be checked, got ["autonomous-agent", "capped-reactive-agent"]`, `left: 2 right: 1` — your scenario precisely. Four pre-existing heartbeat tests constructed their agents with `AutonomousConfig::default()` and the default reactive schedule; two of them expected the agent to be checked and would have failed, and two expected it skipped and would have started passing for the wrong reason. All four now set a continuous schedule so they still exercise what their names claim. That is a changed expectation on pre-existing tests, so it is called out in the PR body too.

Your restore path should work as intended after this: `max_iterations = 80` with no `[autonomous]` block gives you the author's cap back, a reactive schedule, and no `BecameUnresponsive`.

## 1 — agreed in principle, and it needs its own diff

The `!force_approval` conjunct preserving a user-policy `NeedsApproval` is the precedent I would build on too, and "I trust this agent with arbitrary commands" and "ask me first" being independently holdable is the right way to put it. It is a change to when approval is required on a security path in `librefang-runtime`, and it makes the system more restrictive for anyone currently relying on `Full` to run unattended, so it wants its own PR and its own review rather than riding along here. Flagging it as the maintainer's call on strictness, not mine.

## Testing

Testing the released artefact rather than a branch is the right call on a box carrying production bots — please do not build a PR branch there. Worth knowing for when you do exercise it: several `librefang-api` integration binaries `SIGABRT` on macOS at the default thread count, with `wasmtime::runtime::vm::sys::unix::machports::handler_thread → abort` as the faulting thread and no LibreFang frame on the stack. It reproduces on untouched test binaries, so it predates all of this, and Linux is unaffected because it uses POSIX signal handlers. Not something you would hit running a release build, but if you ever do run the test suite locally, `--test-threads=1` is the workaround.

crates/librefang-kernel/src/kernel/hands_lifecycle.rs
```diff
@@ -351,64 +351,11 @@ impl LibreFangKernel {
...
-                             in the hand's `[metadata] frequency` and `[autonomous]` \
-                             guardrails on this role, or an explicit \
-                             `[agents.<role>.schedule.continuous]`"
-                        );
-                    }
-                    None => {}
-                }
-            }
+            // A role now leaves `ScheduleMode::Reactive` only if its own manifest says so: an explicit `schedule` (the only route to the `Periodic` cron and `Proactive` variants) or an explicit `[autonomous]` block, which `librefang_hands::apply_explicit_autonomous_schedule` resolves into `Continuous` at parse time, where the raw TOML still distinguishes an author-written block from the synthesized one.
+            // Honouring `schedule` verbatim is also what `spawn_agent_inner` does for a standalone agent, so hand roles and standalone agents no longer disagree about what the field means.
```
houko, line 358: Backward-compat/observability question, not a request to change the schedule logic: the deleted block here used to emit an `info!("Hand agent stays reactive; ...")` line explaining *why* a role with `[metadata] frequency` and/or `[autonomous]`-carrying `max_iterations` stayed reactive. That's gone now, and nothing replaces it — grep confirms there's no `warn!`/`info!` anywhere in `librefang-hands` or here tied to this resolution.

Combined with `b3d2447`'s heartbeat fix (which is correct on its own terms — these agents genuinely shouldn't be flagged `BecameUnresponsive`), an operator upgrading now has *zero* runtime signal that a previously-ticking hand (20 of 59 bundled roles, per the CHANGELOG) has gone quiet: no activation-time log, and no heartbeat alert either. The only notice is the CHANGELOG entry, which only reaches someone who reads it before/after upgrading.

The CHANGELOG bullet itself is excellent (bolded behavior-change callout, exact affected-role count, migration path). I'm flagging this as a judgment call rather than fixing it myself: is CHANGELOG-only notice sufficient here, or should activation log a one-time `WARN` when a role carries `autonomous.is_some()` (i.e. declares `max_iterations`) but resolves to `Reactive` — mirroring the existing `detect_misplaced_per_agent_overrides`-style targeted-WARN pattern used elsewhere in the kernel for exactly this "config parses fine but silently doesn't do what the author probably expects" situation?

---

houko (commented):
Reviewed the full diff plus the current in-repo state (branch is at 9e01f55, tip of origin/fix/hands-explicit-autonomy). This PR is in unusually good shape already — five commits, with two prior review rounds already folded in (CHANGELOG attribution restore, prose-wrap fixes, and a follow-up commit fixing the heartbeat monitor's own use of the same predicate).

Checked against the review criteria:
- **Consistency of the new gate**: traced every path that can start a hand's background loop (`background_lifecycle.rs` boot scan, `agent_state.rs` schedule-change restart, the hand-activation route in `routes/skills/mod.rs`, and `background.rs::start_agent`'s own `ScheduleMode::Reactive => {}` no-op). All of them key off `manifest.schedule`, and the only place that manifest field is decided for a hand role is `apply_explicit_autonomous_schedule`, which every branch of `parse_single_agent_section` runs through before returning. No bypass found.
- **Test coverage**: both directions are covered — explicit `[autonomous]` in both flat and nested formats yields `Continuous` (`explicit_autonomous_block_yields_continuous_schedule_in_{flat,nested_model}_format`, `hand_activation_honours_explicit_autonomous_block`), and declaration-absent stays `Reactive` across every `HandFrequency` value (`hand_activation_schedule_ignores_hand_frequency`) plus an idempotent serialize/re-parse round trip.
- **CHANGELOG**: correctly attributed `(@houko)`, one-sentence-per-line throughout, and the behavior change is called out in bold with the affected-role count and migration path.
- **Confirmed dead code removal is clean**: `tick_interval_secs`/`declares_wake_up_cycle`/`PERIODIC_TICK_INTERVAL_SECS` have zero remaining references anywhere in the tree.

One inline comment left on `hands_lifecycle.rs` — the deleted scheduling block used to log *why* a role stayed reactive, and nothing replaces that logging now, so an operator upgrading gets no runtime signal (log or heartbeat alert — the latter is now deliberately suppressed for these agents too) that a previously-ticking hand went quiet, only the CHANGELOG entry. That's a judgment call on whether CHANGELOG-only notice is sufficient, so I left it for the maintainer rather than adding logging myself.

No mechanical fixes pushed — nothing met the ≥90%-confidence bar for a direct edit.

---

PR #6615, opened 2026-07-28
fix(api): merge partial identity PATCHes and flag tool_allowlist entries that cannot grant

### LINKED ISSUE #6608: PATCH /api/agents/{id}/identity replaces instead of merging — omitted fields are nulled, unlike the sibling PATCH /api/agents/{id}/config
`PATCH /api/agents/{id}/identity` is registered as a PATCH (`crates/librefang-api/src/routes/agents/mod.rs`, `axum::routing::patch(update_agent_identity)`), annotated `patch` in its `utoipa::path`, and documents its body as "Identity fields to update". All six request fields are `Option<String>`, three of them with `#[serde(default)]`, which is the shape of a partial update.

The handler does a full replacement instead (`crates/librefang-api/src/routes/agents/identity.rs`):

```rust
let identity = AgentIdentity {
    emoji: req.emoji,
    avatar_url: req.avatar_url,
    color: req.color,
    archetype: req.archetype,
    vibe: req.vibe,
    greeting_style: req.greeting_style,
};
```

No read of the current identity, no merge. `PATCH {"emoji": "🤖"}` therefore nulls `avatar_url`, `color`, `archetype`, `vibe` and `greeting_style` if they were set — silently, with a 200.

## The same API already does this correctly next door

`PATCH /api/agents/{id}/config` (`crates/librefang-api/src/routes/agents/config.rs`) writes the *same* identity fields and merges them explicitly:

```rust
// Update identity fields (merge — only overwrite provided fields)
...
// Read current identity, merge with provided fields
let merged = AgentIdentity {
    emoji: req.emoji.or(current.emoji),
    avatar_url: req.avatar_url.or(current.avatar_url),
    color: req.color.or(current.color),
    ...
```

So two PATCH endpoints on the same resource, touching the same six fields, have opposite semantics. A client that partially updates via `/config` keeps its other fields; the identical partial update via `/identity` destroys them.

## Reachability

Not reachable from the current dashboard — the shipped UI always sends the full identity object. This affects direct API consumers, and the endpoint is in the generated OpenAPI surface, so "the UI does not hit it" is not much comfort.

## Suggested fix

Apply the same read-merge-write the sibling handler already uses. If full replacement is actually intended here, the honest shape would be `PUT` plus a docs note, but given the sibling's behaviour and the `Option`/`serde(default)` field types, merge looks like the intended semantics.

## Environment

- Deployed `v2026.7.21`; verified on current `main` (`cb7608d9`, post-#6602/#6603/#6604) — handler unchanged.
- Found while re-auditing dashboard/API consistency after #6596.

Happy to test a patch.

### LINKED ISSUE #6609: PUT /api/agents/{id}/tools accepts tool_allowlist entries that can never take effect — 200 OK, clean GET round-trip, zero runtime grant
This is not a request to change the allowlist semantics — those look deliberate. It is about the API accepting, persisting and echoing back a value that provably cannot do anything, with no signal to the caller.

## Behaviour

`PUT /api/agents/{id}/tools` takes `capabilities_tools`, `tool_allowlist` and `tool_blocklist` as independent fields. Sending only `tool_allowlist` validates, persists, returns 200, and a subsequent GET returns exactly what was sent. Everything looks like a successful grant.

At runtime the tool is not granted, and cannot be.

## Why

Tool resolution in `crates/librefang-kernel/src/kernel/tools_and_skills.rs` applies the two in a fixed order:

- **Step 1** builds the candidate set by filtering all builtins against `declared_tools` (i.e. `capabilities.tools`).
- **Step 4** then applies the per-agent allowlist to whatever survived:

```rust
if !tool_allowlist.is_empty() {
    all_tools.retain(|t| tool_allowlist.iter().any(|a| glob_matches(a, &t.name)));
}
```

`retain` can only remove. A tool name that was never in `capabilities.tools` was already dropped in Step 1, so no allowlist entry can bring it back. The in-code comment at Step 4 ("These are separate from capabilities.tools and act as additional filters") describes exactly this, so the narrowing behaviour reads as intended — the gap is that the write API gives no hint of it.

## Why it matters in practice

The failure is silent in all three places an operator would look: the PUT succeeds, the GET round-trips, and the agent's tool list in the dashboard reflects the stored value. The only way to discover that nothing was granted is to inspect a session export and notice the `tool_use` never happens.

We hit this directly: a tool was granted to a production agent by sending `tool_allowlist` alone, the API reported success, and the agent went on to *report* a successful test of a tool that was never in its trace. Diagnosis took a session export to settle. An operator scripting agent configuration has no reason to suspect the field they just set is inert.

## Suggested fix

Any of these would remove the trap without touching resolution semantics:

1. Reject the request (400) when `tool_allowlist` contains an entry that cannot match anything in the effective `capabilities.tools` set — the strictest option, and the one that fails loudly at the point of the mistake.
2. Accept it but return a warning in the response body naming the inert entries.
3. If neither is wanted, document on the endpoint that `tool_allowlist` only narrows `capabilities.tools` and cannot grant — the OpenAPI description currently says nothing about the interaction, and the field name suggests the opposite of what it does.

Option 1 or 2 seem preferable: the set needed to evaluate "can this entry ever match" is already available at write time.

## Environment

- Deployed `v2026.7.21`; verified on current `main` (`cb7608d9`) — Step 1 / Step 4 ordering unchanged. Note the allowlist match was since switched to `glob_matches` for #6495, which does not affect the narrowing-only property.

Happy to test a patch.

houko:
Fixes #6608
Fixes #6609

Both issues are write-path semantic defects on `/api/agents/{id}/*`, both were found in the same audit sweep, and they land in the same two files — `crates/librefang-api/src/routes/agents/identity.rs` and `crates/librefang-api/src/routes/agents/config.rs`.
#6608's correct reference implementation already lived in `config.rs`, which is also where #6609's handler lives, so splitting them would have put two branches in the same file.

Diagnosis, reproduction and the resolution-order analysis for both: @nevgenov.

## #6608 — `PATCH /api/agents/{id}/identity` replaced instead of merging

`crates/librefang-api/src/routes/agents/identity.rs` built a fresh `AgentIdentity` from the request body alone, with no read of the stored one.
A body of `{"emoji": "X"}` therefore nulled `avatar_url`, `color`, `archetype`, `vibe` and `greeting_style`, persisted that, and returned `200`.

The fix adopts the read-merge-write the sibling `PATCH /api/agents/{id}/config` has always used: read the stored identity out of the registry, overlay only the fields the caller provided, write the result.
The two endpoints write the same six fields and now agree on what a partial body means.

**The merge is factored into one shared helper rather than duplicated.**
`merge_agent_identity` lives in `crates/librefang-api/src/routes/agents/mod.rs` and both handlers call it.
Two PATCH endpoints on one resource with opposite semantics is the actual defect here, and a per-handler copy of the six-line `incoming.field.or(current.field)` merge is exactly what produced it — so a second copy would have fixed the symptom and left the drift mechanism in place.
`mod.rs` was the natural home: `identity.rs` and `config.rs` are siblings that already `use super::*`, and `resolve_lang` / `kernel_err_to_status` / `effective_default_model` already live there, so there is no new module and no import churn.
The helper is pure, which makes a transposed-field bug unit-testable without booting a kernel.
`config.rs` keeps its `has_identity_field` guard, so a `/config` PATCH carrying no identity fields still skips the read-merge-write entirely.

**Finding on clearing a field: neither endpoint can set one back to `null`, and that is unchanged by this PR.**
`null` is already spoken for by "not provided", so there is no encoding left for "clear" without an `Option<Option<T>>` or a sentinel value, neither of which is worth inventing here.
`config.rs` had this same limitation before this change (it was already `req.field.or(current.field)`), so matching it is not a new regression — but it is now a documented contract rather than an accident.
The closest available operation is to store an empty string: `Some("")` passes both handlers' `color` / `avatar_url` validators (each guarded by `!x.is_empty()`) and is stored as `Some("")`, which `GET /api/agents/{id}` then reports as `""` rather than `null`.
The OpenAPI request-body description and `merge_agent_identity`'s doc comment both say exactly that, in those words — "stores an empty string", not "clears", because `""` is what a reader of the spec will actually get back.
No in-repo caller needs `null`-as-clear: the dashboard never calls `PATCH /identity` at all, and the only callers are generic SDK passthroughs (`sdk/go/librefang.go`, `sdk/python/librefang/librefang_client.py`, `sdk/rust/src/lib.rs`).

## #6609 — `PUT /api/agents/{id}/tools` accepted `tool_allowlist` entries that cannot grant

**The resolution semantics are unchanged, deliberately.**
`LibreFangKernel::available_tools` (`crates/librefang-kernel/src/kernel/tools_and_skills.rs`) filters the candidate set by `capabilities.tools` in Step 1 and then applies `tool_allowlist` in Step 4 as `all_tools.retain(...)`.
`retain` can only remove, so `tool_allowlist` narrows what `capabilities.tools` already admits and never grants anything.
That is the intended design and this PR does not touch it — no Step reordering, no change to the `retain`, no change to how any tool resolves at runtime.
The only kernel-side edit is widening `LibreFangKernel::is_evolve_tool` from `pub(crate)` to `pub` (plus doc lines), so the API layer can reuse the kernel's list of eight self-evolution tool names instead of hardcoding a copy that would drift.

What this PR fixes is that the **write API** used to accept, persist and echo an allowlist entry that could never do anything, with no signal at all: `PUT` returned `200`, a subsequent `GET` round-tripped the value, and the only way to discover the tool was never granted was to inspect a session export.
`PUT /api/agents/{id}/tools` now adds an optional `warnings` array of strings to its `200` body, one message per inert entry, naming the entry and saying to add it to `capabilities_tools` instead.
The key is omitted entirely when there is nothing to report, so the response is byte-identical to today's for the common case.
`Vec<String>` under a `warnings` key follows the existing precedent in `routes/config/migration.rs` and `routes/config/system.rs`; the `[{severity, message}]` shape used by the skillhub routes exists because a scanner emits multiple severities, which does not apply here.

### How inert entries are detected

`inert_tool_allowlist_entries` in `crates/librefang-api/src/routes/agents/config.rs` reports an entry only when **all five** of these hold, so that only *provably* inert entries are named:

- `capabilities_tools` is restricted — non-empty and free of a `*` entry.
- The entry contains no `*`.
- The entry is not MCP-namespaced (`librefang_kernel::mcp::is_mcp_tool`), because MCP tools join the candidate set in Step 3 *without* being filtered by `capabilities.tools`, and an `mcp_*` allowlist entry is how a caller selects among them (#6495).
- The entry is not a self-evolution tool (`LibreFangKernel::is_evolve_tool`), because Step 1's post-filter injects those regardless of what the manifest declares.
- No declared pattern glob-matches the entry.

**The check evaluates globs with the same matcher the runtime resolution uses, not string comparison.**
The last clause calls `librefang_types::capability::glob_matches(declared, entry)` — same function, same argument order as the kernel's Step 1 builtin filter and Step 2 skill filter (#6495).
That is the load-bearing part: `capabilities_tools = ["file_*"]` with `tool_allowlist = ["file_read"]` is a working configuration, and a string compare would flag it as inert.
Restricting the check to wildcard-free entries is what removes any need for glob-vs-glob intersection reasoning — `declared = ["file_*"]` and `entry = "*_read"` overlap on `file_read` even though neither pattern matches the other's literal text.
`glob_matches` supports only `*` and `**` (no `?`, no character classes), so `!entry.contains('*')` really does capture every wildcard form, and a wildcard-free pattern reduces to exact equality through its `pattern == value` arm.
The cost is false negatives — a genuinely inert glob entry goes unreported — which is the safe direction, since a later skill install or MCP connect can make a glob live.

**An unbounded `capabilities.tools` is skipped entirely, in both spellings.**
`[]` and `["*"]` both make Step 1's filter a no-op, so every builtin reaches the candidate set and nothing is provably inert.
The first clause mirrors the kernel's own `tools_unrestricted` check (`declared_tools.is_empty() || any(|t| t == "*")`) and returns no warnings in that case, so a wildcard grant never draws a warning on a configuration that works.

**When the check runs.**
It runs when the request submits `tool_allowlist` — the obvious case — and also when it submits `capabilities_tools`, because narrowing the grant surface is itself a way to render a stored allowlist entry inert.
An operator who sends `{"capabilities_tools": ["file_read"]}` against an agent whose stored allowlist is `["web_search"]` is the person who just silenced `web_search`, and gating on a submitted allowlist alone would hand them a clean `200` — the exact "succeeds, round-trips, grants nothing" experience #6609 is about.
A request that submits only `tool_blocklist` touches neither side and stays quiet about whatever was already stored.
The check is evaluated against the registry entry read back **after** the write, so both the declared set and the allowlist it judges are post-write values.

## OpenAPI description changes

The field is named `tool_allowlist`, which reads as a grant to anyone who has not traced `available_tools`, so the spec now says otherwise at every surface a caller looks at.
`openapi.json` and `xtask/baselines/openapi.sha256` are regenerated in this commit.

- `PUT /api/agents/{id}/tools` request body: `capabilities_tools` is the grant surface; the allowlist and blocklist only narrow it, because the kernel applies them afterwards as a retain; an allowlist entry naming an excluded builtin or skill tool grants nothing; MCP tools are the stated exception.
- `PUT /api/agents/{id}/tools` `200`: documents the optional `warnings` array, that the key is absent when empty, and which request shapes trigger the check.
- `GET /api/agents/{id}/tools` `200`: the same narrowing note, since this is the read surface an operator inspects when auditing.
- `SetAgentToolsRequest` per-field docs rewritten for all three fields — they propagate verbatim into the schema, so this is where an SDK user actually reads them.
- `PATCH /api/agents/{id}/identity` request body and the `UpdateIdentityRequest` schema description: PATCH semantics, that a partial body never nulls what it omits, and that an empty string stores an empty string rather than clearing the field.

The four doc comments that propagate into the spec are written one sentence per line, so the generated descriptions no longer carry mid-sentence `\n`.
The purely internal doc comments in the changed files keep the surrounding files' existing column wrapping — they have no downstream artifact, and hand-reflowing the 45-line false-positive argument is a word-drop risk no test would catch.
Say the word and I will reflow those too.

## Verification

All commands run in the worktree on this branch, rebased on `origin/main`.

| Command | Result |
|---|---|
| `cargo check --workspace --lib` | exit 0 |
| `cargo check -p librefang-api --all-targets` | exit 0 (`--lib` does not compile the `#[cfg(test)]` mods or the integration test binaries) |
| `cargo clippy --workspace --all-targets -- -D warnings` | exit 0, zero warning/error lines |
| `cargo fmt --all -- --check` | exit 0 |
| `cargo test -p librefang-api --lib` | 874 passed, 0 failed |
| `cargo test -p librefang-api --lib -- inert_allowlist_tests merge_agent_identity` | 13 passed, 0 failed |
| `cargo test -p librefang-api --test agents_routes_integration -- --test-threads=1` | 57 passed, 0 failed |
| `cargo test -p librefang-api --test agents_capabilities_files_integration` | 21 passed, 0 failed |
| `cargo test -p librefang-api --test dead_route_audit_test` | 2 passed, 0 failed |
| `cargo test -p librefang-api --test openapi_path_coverage_test` | 1 passed, 0 failed |
| `cargo test -p librefang-api --test openapi_spec_test` | 1 passed — this is what regenerates `openapi.json` |
| `cargo test -p librefang-kernel --lib -- --test-threads=4` | 1334 passed, 0 failed, 1 ignored |
| `shasum -a 256 -c xtask/baselines/openapi.sha256` | `openapi.json: OK` |
| `python3 scripts/codegen-sdks.py` | all four SDKs rewritten, `git status` clean afterwards — `scripts/codegen-sdks.py` never reads `description`, so a description-only spec change cannot move them |

`dead_route_audit_test`, `openapi_path_coverage_test` and `openapi_spec_test` all pass, so no route or schema shape moved.

### The regression tests fail against unmodified code

Each was run against the pre-fix source to confirm it bites, not just asserted to.

#6608, in `crates/librefang-api/tests/agents_routes_integration.rs` — reverting only `identity.rs` and re-running gives `0 passed; 3 failed`:

- `test_patch_identity_merges_instead_of_replacing` — fails on "avatar_url must survive a partial PATCH", `None` vs `Some("https://example.invalid/avatar.png")`. It seeds all six fields first and then patches one, and also pins that the patched field actually changed, so a no-op handler fails it too.
- `test_patch_identity_and_patch_config_agree_on_partial_updates` — fails on "avatar_url diverged". Six pairwise comparisons between the two routes plus four absolute pins, so it cannot pass by both routes being wrong identically.
- `test_patch_identity_empty_string_clears_a_single_field` — fails on "clearing one field must not disturb the others".

Unit tests for the helper, in `mod.rs`: `merge_agent_identity_preserves_fields_the_caller_omitted`, `merge_agent_identity_maps_each_field_to_itself` (per-field distinguishable values, so a transposed pair is caught), `merge_agent_identity_treats_empty_string_as_an_explicit_clear`, `merge_agent_identity_with_no_incoming_fields_is_a_noop`.

#6609, in `crates/librefang-api/tests/agents_capabilities_files_integration.rs`:

- `test_tools_put_warns_about_inert_allowlist_entry` — fails pre-fix, there was no `warnings` key. Sends both fields in one request, so it also catches a check evaluated before the write, and asserts the value is still persisted and round-trips through `GET`.
- `test_tools_put_warns_when_narrowing_capabilities_makes_a_stored_entry_inert` — the pin for the capabilities-only request shape. Verified by reverting the gate to `body.tool_allowlist.is_some()` alone: `8 passed; 1 failed`, on "narrowing capabilities_tools must report the stored entry it silenced", `left: 0 right: 1`. It also asserts that widening back to `["*"]` goes quiet again.
- `test_tools_put_does_not_warn_about_a_live_allowlist_entry` — the false-positive guard (`capabilities_tools = ["file_*"]`, `tool_allowlist = ["file_read"]`). Verified non-vacuous by mutating the glob clause to string equality: this test plus `entry_admitted_by_a_declared_glob_is_not_inert` and `only_inert_entries_are_reported_and_order_is_preserved` fail.
- `test_tools_put_does_not_warn_when_capabilities_tools_is_unbounded` — both `["*"]` and `[]`.
- `test_tools_put_does_not_warn_about_mcp_or_glob_entries`.
- `test_tools_put_only_warns_about_submitted_allowlist_entries` — a blocklist-only follow-up must stay quiet about a stored inert entry.

Nine unit tests for the pure helper in `config.rs::inert_allowlist_tests` cover each clause in isolation, including `mcp_namespaced_entry_is_never_reported` (the integration test's `mcp_*` entry is caught by the wildcard clause first, so the MCP clause needs its own case) and `only_inert_entries_are_reported_and_order_is_preserved`.

### Why the integration binaries were run at a reduced thread count

Some `librefang-api` integration binaries `SIGABRT` on macOS when run at the default thread count, and I chased it down rather than assume it was environmental noise.

The abort is in wasmtime's macOS Mach-port trap handler, not in any LibreFang code.
The faulting thread in the crash report (`~/Library/Logs/DiagnosticReports/*.ips`, `asi: {"libsystem_c.dylib": ["abort() called"]}`) is:

```
libsystem_c.dylib                 abort
agents_routes_integration         wasmtime::runtime::vm::sys::unix::machports::handler_thread
agents_routes_integration         wasmtime::runtime::vm::sys::unix::machports::TrapHandler::new::{{closure}}
```

Each test boots a kernel, each kernel brings up the WASM sandbox, and each of those installs a process-global Mach exception handler; enough of them concurrently in one process and the handler thread aborts.
Three things establish it is not this change:

- `api_integration_test`, which this diff does not touch at all, aborts with the identical signature at both the default thread count and `--test-threads=4`.
- No LibreFang frame appears on the faulting stack — it is wasmtime's own handler thread, spawned before any route code runs.
- `agents_routes_integration` at `--test-threads=1` is **57 passed, 0 failed, exit 0**, including all three new `test_patch_identity` cases.

It is macOS-only: the Linux CI shards use POSIX signal handlers rather than Mach exception ports, so CI is unaffected.
Worth its own issue if the macOS lane matters to you — say the word and I will open one.

Separately, a real `OPENAI_API_KEY` in the local environment makes `media_routes_integration`'s "no provider configured" tests hit the live OpenAI API; `env -u OPENAI_API_KEY` gives 18 passed, 0 failed.
Neither issue is reachable from this diff.

## Out-of-scope — questions for the maintainer

**Should an inert `tool_allowlist` entry be rejected with `400` instead of warned about?**
The argument for it: any caller sending one is already silently failing, so a `400` breaks nothing that currently *works*, and it moves the failure to the moment the mistake is made.
The argument against it: it converts a silent no-op into a hard failure in an existing deploy pipeline, so a config-management run that has been "succeeding" for months starts erroring on the first apply after upgrade — and unlike the warning, that is not additive.
This PR keeps the write permissive because the strictness call belongs to you, not to the PR that discovered the problem.
If you want it tightened, should it be unconditional or behind a config flag?

**Should the dashboard surface `warnings`?**
`AgentToolsResponse` in `crates/librefang-api/dashboard/src/api.ts` has no `warnings` field, and `updateAgentTools` feeds the raw response straight into `setQueryData`, so the warning is invisible in the UI today.
Confirmed harmless — the dashboard uses plain TS interfaces with no runtime schema validation, and the one vitest case that touches this path (`src/api.test.ts`) asserts against its own mocked response, so the extra key cannot break anything.
Adding `warnings?: string[]` to the type alone would be dead code; surfacing it properly means a toast in the save flow plus i18n keys across the locale files.
That is TS/i18n work in a different domain from this Rust write path, and the reporter's access pattern is the direct API, so the fix lands for them either way — but I did not want to decide unilaterally to skip it. Fold it into this PR, or a follow-up?

**A real tempfile race in `crates/librefang-api/src/routes/sidecar_toml.rs` — needs its own issue.**
`upsert_sidecar_block` (line 142) and `remove_sidecar_block` (line 184) each hold their *own* function-local `static SEQ: AtomicU64 = AtomicU64::new(0)`, and both build `parent.join(format!(".config.toml.tmp.{}.{seq}", std::process::id()))` — same pid, same directory, two counters that each start at 0.
A concurrent upsert and remove therefore both mint `.config.toml.tmp.<pid>.0`; one renames it away and the other fails with `ENOENT` or writes a torn file.
This is not test-only — the comment above the counter says it exists specifically to guard "two HTTP handlers racing on the same config file", so the mitigation is incomplete for exactly the cross-function case, which two concurrent sidecar API calls hit.
The fix is small (one shared module-level counter, or the function name in the temp path) but it is a different module and domain and needs its own regression test, so bundling it would break one-PR-one-issue. Want me to open the issue?

**`mcp__` vs `mcp_` is two namespaces or one typo — blocked on that question.**
`format_mcp_tool_name` (`crates/librefang-runtime-mcp/src/lib.rs`) emits `mcp_{server}_{tool}` with a single underscore, and `is_mcp_tool` tests `starts_with("mcp_")`.
But `crates/librefang-runtime/src/tool_classifier.rs` branches on `name.starts_with("mcp__")`, which under LibreFang's own naming can never match, so `mcp_default_safety` is effectively dead for LibreFang-generated MCP tool names.
The comment at `tools_and_skills.rs` likewise tells operators to write an `mcp__*` glob, which `glob_matches` will not match against `mcp_github_create_issue`.
`claude_code.rs` documents `mcp__<server>__<tool>` as the *Anthropic CLI* convention, so these may be two genuinely distinct namespaces rather than one typo — fixing any single site would create new inconsistency.
It does not affect the #6609 check either way, since `mcp_` prefixes `mcp__`. Which is it meant to be?


nevgenov:
On the `mcp__` vs `mcp_` question at the end — production data from this host, which may settle it.

All 67 MCP tools exposed to an agent here carry a **single** underscore, matching `format_mcp_tool_name`:

```
mcp_office_tools_docx_create
mcp_notion_afina_api_get_user
mcp_ms_ivt_router_...
```

Zero tools with `mcp__`. Seven configured MCP servers, five of them third-party stdio servers, none of which influence the name shape — the prefix is minted by LibreFang, not by the server.

So on the evidence they are two distinct namespaces, not one typo:

- `mcp_{server}_{tool}` — what LibreFang generates and what `is_mcp_tool` tests for.
- `mcp__{server}__{tool}` — the Anthropic CLI convention that `claude_code.rs` documents, which reaches LibreFang only for tools originating on that side.

Which makes `tool_classifier.rs`'s `starts_with("mcp__")` correct for its own namespace and inert for LibreFang-generated names, exactly as you describe — so `mcp_default_safety` never applies to the 67 tools above. Whether that is the intent is the part we cannot answer from here; if it is meant to cover both, the check needs to accept the single-underscore form too, and if it is CLI-only then a comment saying so would stop the next reader filing it as a bug.

The operator-facing half is less ambiguous: the `tools_and_skills.rs` comment tells operators to write an `mcp__*` glob in `tool_allowlist`, and on this install that pattern matches nothing at all — every tool it would need to match is single-underscore. That one looks like a straightforward doc error regardless of how the namespace question resolves.

Happy to dump the full 67-name list if useful.

Issue #6596, opened 2026-07-27
`GET /api/config` omits config fields that `PUT` accepts — write-only settings are invisible in the dashboard


nevgenov:
## Symptom

Several settings that live in `config.toml` and are accepted by the config write path are never returned by `GET /api/config`. The dashboard therefore renders them blank even when a value is set and in effect.

Concrete case that surfaced this: `[browser] cdp_endpoint`. Our `config.toml` has

```toml
[browser]
enabled = true
cdp_endpoint = "ws://127.0.0.1:9222/"
```

and the endpoint is genuinely in use (we attach to a long-running local CDP server). The dashboard shows the CDP Endpoint field empty, which reads as "not configured" — the risk being that an operator re-enters or "fixes" a setting that was never actually missing.

## Root cause (read from current source)

`crates/librefang-api/src/routes/config/manage.rs` builds the response by hand-enumerating fields per section rather than serializing the config struct. The `browser` block lists seven fields:

```rust
set!("browser", {
    "headless": config.browser.headless,
    "viewport_width": config.browser.viewport_width,
    "viewport_height": config.browser.viewport_height,
    "timeout_secs": config.browser.timeout_secs,
    "idle_timeout_secs": config.browser.idle_timeout_secs,
    "max_sessions": config.browser.max_sessions,
    "chromium_path": config.browser.chromium_path,
});
```

`BrowserConfig` (`crates/librefang-types/src/config/types.rs:~820-850`) additionally has `enabled`, `auto`, `cdp_endpoint`, and `cdp_token_env`. Nothing scrubs them: `SCRUB_SUFFIXES` only blocks `*.api_key`, `*.password`, `*.bypass`, `*.admin`, `*.owner`, and none of the omitted names match. The write side is happy to take them — `browser.` is in `SECTION_PREFIXES` in `is_writable_config_path` (`routes/config/mod.rs:~709`), so `browser.cdp_endpoint` is a writable path.

The net effect is a set of **write-only** settings: `PUT` accepts them, `GET` never shows them.

## Scope — not just `browser`

Diffing our `config.toml` against the `GET /api/config` payload (secrets excluded) turns up 8 paths that are writable per `SECTION_PREFIXES` but absent from the read response:

| Path | Section prefix writable? |
|---|---|
| `browser.cdp_endpoint` | yes |
| `browser.enabled` | yes |
| `browser.auto` | yes |
| `auto_dream.auto` | yes |
| `channels.file_download_dir` | yes |
[...]

Issue #3703, opened 2026-04-27
OIDC login does not require email_verified — allowed_domains can be impersonated


houko:
## Severity
**High** — domain-restricted SSO can be impersonated by anyone who can register an unverified email at the IDP.

## Location
- `crates/librefang-api/src/oauth.rs:841-862` (allowed_domains check)
- `crates/librefang-api/src/oauth.rs:94` (`IdTokenClaims.email_verified: Option<bool>` — already deserialized, never consulted)

## Finding
The `allowed_domains` enforcement only checks the email's domain suffix:

```rust
if !provider.allowed_domains.is_empty() {
    if let Some(ref email) = claims.email {
        let domain = email.rsplit('@').next().unwrap_or("");
        if !provider.allowed_domains.iter().any(|d| d == domain) {
            return ... FORBIDDEN ...
        }
    }
    // ← falls through; claims.email_verified is never inspected
}
```

`email_verified` is parsed into `IdTokenClaims` (`oauth.rs:94`) but **never read**. An IDP that returns `email_verified: false` (or omits it) is treated identically to one that returns `email_verified: true`.

## Why this is exploitable
The intended threat model of `allowed_domains` is "only people inside `company.com` can log in." That model assumes the IDP has verified the email address actually belongs to the user. But:

1. Many SaaS IDPs (Auth0 trial tenants, self-hosted Keycloak with default settings, several OIDC-as-a-service offerings) allow account signup without email verification, or with verification trivially skippable.
2. An attacker registers `boss@company.com` at such an IDP.
3. They authenticate to LibreFang via the OIDC flow. The IDP returns an ID token with `email: "boss@company.com"`, `email_verified: false`.
4. LibreFang accepts the login, attributes the session to a `company.com` user.

Even with the company's own IDP, an attacker who compromises a low-privilege account can change their profile email to `ceo@company.com` if the IDP allows self-service email change without re-verification, then re-authenticate.

## Suggested fix
Add a strict check in the same block:

```rust
if !provider.allowed_domains.is_empty() {
    if claims.email_verified != Some(true) {
        return Err(error_with_lang(
            StatusCode::FORBIDDEN,
            "OIDC provider did not assert email_verified=true; refusing domain-restricted login",
            lang,
        ));
    }
    if let Some(ref email) = claims.email {
        // existing domain check
    } else {
        return Err(...FORBIDDEN, "no email claim"...);
    }
}
```

[...]

PR #6618, opened 2026-07-28
fix(api): expose non-writable security config on read and guard writable paths against missing fields

### LINKED ISSUE #6605: Two config read/write gaps that survive #6596's parity guard: a non-writable-but-unreadable security flag, and a writable path with no backing field
Follow-up to #6596 / PR #6604. That PR fixed the field inventory and — more usefully — added `config_read_write_parity_tests::every_writable_config_leaf_is_readable`, which turns future drift into a build failure. Re-checking against current `main` after the merge, two gaps remain, and each escapes that guard for a different structural reason. Neither is a regression from the PR; both predate it.

## Gap 1 — `external_auth.require_email_verified` is invisible on read, and the guard cannot see it

`ExternalAuthConfig.require_email_verified` (`crates/librefang-types/src/config/types.rs`, default `true`) is the documented mitigation for #3703 — it prevents `allowed_domains` impersonation via unverified addresses.

It is deliberately excluded from `EXACT`/`SECTION_PREFIXES` in `is_writable_config_path`, with a comment explaining why: flipping it post-auth would let an Owner-role attacker skip email verification. That exclusion is correct and should stay.

But it is *also* absent from `GET /api/config`'s `external_auth` object, and nothing documents that as intentional. The consequence is asymmetric in the wrong direction: an operator has no way — dashboard or API — to confirm whether the protection is currently active. The only ways to check are reading raw `config.toml` or the `Default` impl.

The new parity guard enforces `writable ⊆ readable`. A field that is intentionally non-writable but should still be readable falls outside that invariant by construction, so no amount of guard coverage will surface it.

Same applies to five `OidcProvider` fields, also absent from GET: `auth_url`, `token_url`, `userinfo_url`, `jwks_uri`, `audience`, plus the per-provider `require_email_verified: Option<bool>` override.

**Repro:**
```
curl -s -H "Authorization: Bearer <key>" http://localhost:4545/api/config \
  | python3 -c "import json,sys; print('require_email_verified' in json.load(sys.stdin)['external_auth'])"
# -> False
```

**Suggested fix:** emit these on the read side and keep them non-writable. If some should stay hidden, a comment saying so would prevent the next person re-reporting it.

## Gap 2 — `ui.theme` / `ui.locale` / `ui.timezone` / `ui.language`: writable paths with no backing struct field at all

These four are in the `EXACT` writable list (`crates/librefang-api/src/routes/config/mod.rs`, comment "UI / locale (no security impact)") and are pinned by a unit test asserting they are writable (`issue_3458_writable_path_allowlist`).

There is no `UiConfig` struct and no `pub ui:` field anywhere in `crates/librefang-types/src/config/`, and no `ui` key in the generated config schema golden fixture. Searched current `main` after PR #6604 — still absent.

The write path validates the dotted path against `is_writable_config_path` and then edits `config.toml` through `toml_edit` keyed by that path; it does not require the path to correspond to an existing field. So a write is accepted, a `[ui]` table lands on disk, and `KernelConfig` (no `deny_unknown_fields`) discards it silently on the next load. The value is never applied, never read back, and `GET /api/config` reports `ui` as null.

This is a distinct class from #6596: not "readable side fell behind the struct" but "writable side references something that was never in the struct". The parity guard walks paths derived from the serialized config, so a writable path with no corresponding field is structurally invisible to it — the guard cannot fail on a leaf that does not exist in the value being walked.

Looks like leftover scaffolding from a UI-preferences feature that was never wired into `KernelConfig`.

**Note on evidence:** this one is traced through the code, not live-tested — we did not want to write to a production `config.toml` to confirm the 200. Happy to test on a scratch instance if useful.

**Suggested fix:** either drop the four entries from `EXACT` (and the assertions in `issue_3458_writable_path_allowlist`), or add the backing `UiConfig` if the feature is still wanted. A second guard direction — every entry in `EXACT`/`SECTION_PREFIXES` resolves to a real field — would catch this class and complements the one PR #6604 added.

## Environment

- Deployed `v2026.7.21`; both gaps re-verified against current `main` (post-#6604 merge), not just the released tags.
- Found while re-auditing dashboard/API consistency after #6596 was fixed.

Happy to test a patch.

houko:
Fixes #6605.

Diagnosis and both gap reports by @nevgenov.

## Context: follow-up to #6596 / PR #6604, and neither gap is a regression from it

PR #6604 fixed the read side of `GET /api/config` and added `every_writable_config_leaf_is_readable`, a guard enforcing `writable ⊆ readable`.
Both gaps here predate that PR — neither is a regression from it — and each escapes its guard for a different structural reason.

- A field that is intentionally **non-writable** but still worth reading falls outside `writable ⊆ readable` by construction: the invariant says nothing about paths that are not writable in the first place.
- A **writable path with no backing field** is invisible to that guard because it derives its candidate paths *from* a serialized `KernelConfig` value, so a path naming a field that does not exist can never appear in its input.

The new guard added here is the mirror direction of #6604's: `writable ⊆ readable` checks that everything writable can be read back, and `every_writable_allowlist_entry_has_a_backing_config_field` checks that everything writable corresponds to a field that exists at all.

## Gap 1 — non-writable `external_auth` fields are now readable

`crates/librefang-api/src/routes/config/manage.rs`, `redacted_config_json`:

- `external_auth.require_email_verified` (top level).
- The six missing `OidcProvider` fields: `auth_url`, `token_url`, `userinfo_url`, `jwks_uri`, `audience`, and the per-provider `require_email_verified` override.

**`require_email_verified` remains non-writable by design; only its readability changed.**
It is the #3703 mitigation — it rejects a login whose ID token does not carry `email_verified = true`, which is what stops an unverified address in an `allowed_domains` domain from inheriting that domain's authorization — and it stays absent from the `POST /api/config/set` allowlist so an Owner-role caller with a leaked API key cannot switch it off.
Omitting it from the *read* side too was the wrong asymmetry: an operator had no way to confirm the protection was active without shell access to read `config.toml`.
The new integration test asserts these paths still return 403 from `POST /api/config/set`.

The per-provider override is `Option<bool>` and `None` is emitted as `null`, so "inherit the global setting" stays distinguishable from an explicit `false`.

Nothing newly exposed is secret-bearing, and nothing was newly redacted.
`client_secret_env` names the environment variable the secret is read from rather than the secret, `client_id` is the public half of the OAuth client registration, and both were already emitted.
Neither struct holds a secret value at all — the secret never lives in config.

Census was mechanical rather than trusting the report: `OidcProvider` has exactly 14 fields and all 14 are now emitted with JSON key equal to the struct field name; `ExternalAuthConfig` was missing only `require_email_verified`.

**Public-route exposure, checked and judged to add no new class.**
`GET /api/config` sits in `PUBLIC_ROUTES_DASHBOARD_READS` (`crates/librefang-api/src/middleware.rs:1177`), so on an install with no auth configured at all `require_email_verified` becomes anonymously readable alongside the already-exposed `allowed_domains`.
That combination is not a new exposure class: `require_auth_for_reads` defaults to `None` (`crates/librefang-types/src/config/types.rs:3335`) and derives from configured auth, so any install with OIDC providers configured also has an api_key or dashboard credentials and the endpoint is bearer-gated; an install with no auth at all already exposes the entire config, including `allowed_domains`, `api_listen`, and `home_dir`.

## Gap 2 — the four dead `ui.*` write paths are removed

`ui.theme`, `ui.locale`, `ui.timezone`, and `ui.language` sat in `WRITABLE_EXACT_PATHS` accepting writes that were silently discarded, from #4113 (`ad96842c`, 2026-05-01) until now.

Evidence that decided it — nothing writes them and nothing ever did:

- There is no `UiConfig` struct and no `pub ui:` field anywhere in `crates/librefang-types/src/config/`.
- `ui` is absent from the 129 root properties of `crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json`.
- The dashboard persists all four client-side: `crates/librefang-api/dashboard/src/lib/store.ts` uses zustand `persist` under the store name `librefang-ui-storage`, with `partialize` covering `theme` / `language` / `isSidebarCollapsed`; `SettingsPage.tsx` reads `useUIStore((s) => s.theme)` and calls `toggleTheme()` with no config-set call; the `locales/en.json` help text reads "Stored in browser localStorage; per-device."
- The only remaining repo references were the allowlist itself, its unit test, and two integration tests using `ui.theme` as a throwaway payload for unrelated failure modes.

Because `KernelConfig` does not set `deny_unknown_fields`, neither the post-edit `toml::from_str::<KernelConfig>` check nor the subsequent reload could reject the write: the path validated against the allowlist, `toml_edit` dropped a `[ui]` table into `config.toml`, and the next load discarded it.
The caller received a success status for a change that was never applied and never read back.

## New guard

`writable_allowlist_backing_field_tests::every_writable_allowlist_entry_has_a_backing_config_field` in `crates/librefang-api/src/routes/config/manage.rs`.

- **Enumerates the real lists rather than restating them.** `EXACT` / `SECTION_PREFIXES` / `DEPTH_2_ONLY_PREFIXES` moved from function-local `const`s inside `is_writable_config_path` to module scope as `WRITABLE_EXACT_PATHS` / `WRITABLE_SECTION_PREFIXES` / `WRITABLE_DEPTH_2_ONLY_PREFIXES`. The relocation is mechanical, verified by a normalized line-multiset diff: the only content changes are the three renames, the three reference updates, and the four removed entries.
- **The oracle is a live `schemars::schema_for!(KernelConfig)`**, not a serialized config value and not the golden schema fixture. Trade-off that decided it: a value walk cannot see a field whose `#[serde(skip_serializing_if = …)]` predicate holds for its default, and `crates/librefang-types/src/config/types.rs` carries 63 of those attributes — several on writable paths (`exec_policy.allowed_env_vars`, `budget.providers`, `tool_invoke.allowlist`) — so a value-based oracle would report real fields as dangling. Generating live also means the oracle tracks the struct automatically and cannot go stale the way a checked-in fixture can.
- **No exclusions.** With the four `ui.*` entries gone, all 29 remaining exact paths and all 54 section prefixes resolve.
- **Loud-failure guards**, because the failure mode of a path resolver is over-permissiveness rather than emptiness: a count floor (`29 + 54 > 50`), an assertion that the schema's `definitions` block is non-empty so a schemars draft-2020-12 upgrade — which renames it to `$defs` and would leave every `$ref` dangling — fails on the shape change instead of on 83 simultaneous misses, and four negative controls (`ui.theme`, bare `ui`, `nonexistent_section.nonexistent_field`, `log_level.nested_under_a_scalar`).
- **Stated scope limit:** section prefixes are checked at their base only. A prefix admits arbitrary leaves beneath it, so "this section exists" is the strongest claim available without enumerating what a caller might post. This is a real limitation and it is why the `channels.` question in Out-of-scope below is not caught by the guard.
- Second test `every_depth_2_only_prefix_is_also_a_section_prefix`: `WRITABLE_DEPTH_2_ONLY_PREFIXES` is consulted only from inside the `WRITABLE_SECTION_PREFIXES` loop, so an entry missing from the latter restricts nothing — same defect shape as a dangling path, a rule that reads as a restriction but has no effect.

## Tests added

- `writable_allowlist_backing_field_tests::every_writable_allowlist_entry_has_a_backing_config_field` (`crates/librefang-api/src/routes/config/manage.rs`).
- `writable_allowlist_backing_field_tests::every_depth_2_only_prefix_is_also_a_section_prefix` (same file).
- `config_read_write_parity_tests::non_writable_but_operator_visible_fields_are_readable` (same file). Pinned by name because this direction has no machine-derivable membership test: "non-writable but should be readable" has no general oracle, since the correct answer for `api_key` is "redacted" and for `dashboard_pass` is "absent". It asserts the gate plus all 14 provider keys are present and non-null, with the config set to the non-default `false` so a hardcoded literal in the response builder cannot pass it.
- `get_config_exposes_non_writable_external_auth_fields` (`crates/librefang-api/tests/config_routes_integration.rs`). Boots through a new `boot_router_with_config` hook (a four-line refactor of `boot_router_with_api_key`; existing callers unchanged), asserts every endpoint URL round-trips as an exact string, asserts both `require_email_verified` values come back `false`, and asserts the same paths still 403 on `POST /api/config/set`.

The 403 assertions are reachable rather than short-circuiting: `config_set` runs only `validate_config_key_path` (path shape, no type check) before `is_writable_config_path`, so posting `{"value": true}` against the `String`-typed `external_auth.audience` cannot fall out early as a 400.

## Pre-existing test expectations that changed

**One real expectation change.** `routes::config::searxng_config_parse_tests::issue_3458_writable_path_allowlist` previously asserted `ui.theme` and `ui.locale` **writable**; it now pins all four `ui.*` paths **closed**. This is the deliberate behaviour change of the PR: the old expectation pinned a write that returned 200 and was discarded by the next load.

**Two payload swaps that are not expectation changes.** `config_set_rejects_missing_value_field` and `config_set_requires_auth_when_key_set` posted `ui.theme` and now post `log_level`; both still assert the same 400 and 401 respectively, because both statuses are produced before the allowlist gate is reached (the 400 from the missing `value` field, the 401 from the auth middleware). They were changed only so they no longer reference a path that does not exist.

## Verification

| Command | Result |
|---|---|
| `cargo check --workspace --lib` | exit 0 |
| `cargo clippy --workspace --all-targets -- -D warnings` | exit 0, zero warnings |
| `cargo fmt --check -p librefang-api` | exit 0 |
| `cargo test -p librefang-api --lib -- routes::config` | 24 passed, 0 failed |
| `cargo test -p librefang-api --lib -- writable_allowlist_backing_field_tests` | 2 passed, 0 failed |
| `cargo test -p librefang-api --test config_routes_integration` | 36 passed, 0 failed |
| `cargo test -p librefang-api --no-fail-fast` | see the note at the end — four binaries fail for host-environment and parallel-load reasons, each re-verified passing in isolation |

`cargo clippy --workspace --all-targets` is the load-bearing gate for this change, not `cargo check --workspace --lib`: every added test lives in a `#[cfg(test)]` module or a `tests/` target, neither of which `--lib` compiles.

`cargo test -p librefang-types` was not run because that crate is untouched — `git diff origin/main..HEAD --stat` is `CHANGELOG.md` plus exactly three files under `crates/librefang-api/`. The dashboard suite was not run because `crates/librefang-api/dashboard/` is untouched.

Generated artifacts are correctly unchanged: `git diff origin/main..HEAD --stat -- openapi.json sdk/ xtask/baselines/ crates/librefang-api/tests/fixtures/` is empty, re-measured against the rebased base rather than the pre-rebase one. The route returns `Json<serde_json::Value>`, so no utoipa schema references the new keys, and no `KernelConfig` field was added or renamed.

Rebased onto `origin/main` at `1b9518fd`; the only conflict was `CHANGELOG.md` (both sides appending under `### Fixed`, resolved by keeping both), and `Cargo.lock` was unaffected.

## Out-of-scope — questions, not decisions

**1. `channels.` is a live instance of the same defect, one level below what the guard inspects. Which repair do you want?**

After the sidecar migration `ChannelsConfig` (`crates/librefang-types/src/config/types.rs:7066-7088`) declares only three depth-1 scalars — `file_download_max_bytes`, `file_download_dir`, `file_upload_max_bytes` — and no per-vendor sub-tables.
`WRITABLE_DEPTH_2_ONLY_PREFIXES = ["channels."]` (`crates/librefang-api/src/routes/config/mod.rs:806`) therefore rejects all three real fields, which sit at depth 1, and accepts depth-2 paths such as `channels.telegram.enabled` that resolve to no field at all — byte-for-byte the `ui.*` shape this PR removes.
The visible half is the mirror image: `ui_sections_overlay` declares a `channels` section (`mod.rs:378`) and `ConfigPage` posts `` `${sectionKey}.${fieldKey}` `` (`crates/librefang-api/dashboard/src/pages/ConfigPage.tsx:648,815,826`), so the dashboard renders editors for those three real scalars and every save is rejected with 403 today.
The new guard cannot see this: it checks a section prefix at its base, and `channels` *is* a real `KernelConfig` field.
Both candidate repairs change the write *surface* rather than the allowlist's internal consistency, which is why this is left untouched here — dropping `channels.` makes the surface match the struct and the 403 honest but leaves the dashboard section unsavable, while moving it to depth 1 makes the three scalars writable and retires the depth-2 protection should a vendor sub-struct ever return.
Recorded as a factual note at the decision site with no TODO marker, and cross-referenced from the `assert!(is_writable_config_path("channels.telegram.enabled"))` in `issue_3458_writable_path_allowlist` so that assertion no longer reads as intentional without explanation.
Which repair, and should the guard grow depth-2 coverage for prefixes on that list?

**2. Should runtime write-path field validation land here or as its own PR?**

The durable fix for this whole class is rejecting a dotted path that resolves to no field at request time, not only at build time.
It is not implemented here because of a concrete regression: `DefaultModelConfig.extra_params` is `#[serde(default, flatten)]` (`crates/librefang-types/src/config/types.rs:6681`, verified on `origin/main` at `1b9518fd`) and `default_model.` is a writable section prefix, so `default_model.<custom_provider_param>` is legitimately absorbed today and a strict properties check would start rejecting it.
`channels.<vendor>` `OneOrMany` shapes are a second risk.
The trade-off: it closes the class at runtime, but needs a flatten / dynamic-map escape hatch designed and tested — which is more surface than #6605 asks for.
Own PR, or fold it in?

**3. `sidecar_toml.rs` mints colliding temp filenames across two functions — should this get its own issue?**

Found while auditing the config write paths.
`upsert_sidecar_block` (`crates/librefang-api/src/routes/sidecar_toml.rs:144`) and `remove_sidecar_block` (`:186`) each declare their **own** `static SEQ: AtomicU64 = AtomicU64::new(0)` and both format `.config.toml.tmp.{pid}.{seq}` into the target's parent directory.
Two separate counters over one namespace: an upsert and a remove running concurrently in the same process against different files in the same directory both produce `.config.toml.tmp.<pid>.0`, so one renames the other's temp file away — producing `rename … No such file or directory`, or content landing on the wrong target.
The comment at `:137-141` says the counter exists to guard "concurrent threads within this process", which is exactly the case it fails to cover.
Repo-internal callers share `~/.librefang/` as the parent, so a concurrent sidecar configure and remove on a live daemon can hit this, not only the test suite.
Different module and domain from #6605, and the fix is a behaviour change to a config write path, so it is flagged rather than touched.

## Notes on the full `cargo test -p librefang-api` run

Four test binaries fail on this machine for reasons unrelated to this diff. Each was root-caused by re-running it in isolation rather than assumed pre-existing, and none is in a module this diff touches — the two production functions this PR changes, `redacted_config_json` and `is_writable_config_path`, are reachable only from `GET /api/config` (`manage.rs:18`) and `POST /api/config/set` (`manage.rs:1223`), and none of these binaries exercises either route:

- `media_routes_integration` — host environment. `OPENAI_API_KEY` is set in the shell, so tests asserting "no configured provider" reach the real API. `env -u OPENAI_API_KEY cargo test -p librefang-api --test media_routes_integration` passes 18/18.
- `agents_routes_integration` — SIGABRT under parallel load. `--test-threads=1` passes 57/57.
- `api_integration_test` — SIGABRT under parallel load; the log shows a registry-sync `git fetch` to `github.com:443` timing out after 75s, and this environment has no egress. `--test-threads=1` passes 88/88.
- `sidecar_toml_test` — the concurrency bug in Out-of-scope item 3, and the one failure here that is a genuine product defect rather than an environment artifact. 3 of 11 fail in parallel and all 11 pass with `--test-threads=1`, in 0.01s either way, so no I/O timing is involved — which is what corroborates the colliding-temp-filename root cause rather than load.

CI does not see the three load-related failures because it shards these binaries across four Ubuntu runners; they surface here only because 80+ test binaries run concurrently on one machine.

Separately worth knowing: `kernel_config_schema_matches_golden_fixture` (`crates/librefang-api/tests/config_schema_golden.rs:90-99`) is `#[ignore]`d **on main** for a pre-existing 490-byte drift. This branch neither caused it nor depends on it — the new guard generates its schema live rather than reading that fixture.


crates/librefang-api/src/routes/config/mod.rs
```diff
@@ -579,232 +579,244 @@ pub fn ui_options_overlay(
...
+// A depth-2 write such as `channels.telegram.enabled` is still accepted, still lands a `[channels.telegram]` table in config.toml, and is still discarded by the next load — the same silent-no-op shape #6605 removed the `ui.*` entries for.
+// `every_writable_allowlist_entry_has_a_backing_config_field` cannot see it: the guard checks a section prefix at its base, and `channels` is a real `KernelConfig` field, so the dangling part sits one level below what the guard inspects.
+// The visible half is the mirror image: `ui_sections_overlay` declares a `channels` section and `ConfigPage` posts `<section>.<field>`, so the dashboard renders editors for the three real scalars and every save of one is rejected with 403.
+// Both candidate repairs (drop `channels.` so the surface matches the struct, or move it to depth 1 so the three real scalars become writable) change the write surface rather than the allowlist's internal consistency, so they are a maintainer call, not a mechanical consequence of #6605.
+const WRITABLE_DEPTH_2_ONLY_PREFIXES: &[&str] = &["channels."];
```
houko, line 806: Flagging rather than fixing, since this is a security-relevant judgment call: the comment above this const documents that `channels.` is now a section prefix with no backing depth-2 field on `ChannelsConfig` post-sidecar-migration (only three depth-1 scalars remain: `file_download_dir`, `file_download_max_bytes`, `file_upload_max_bytes`). A `channels.telegram.enabled`-style write is still accepted, still lands a `[channels.telegram]` table in `config.toml`, and is still silently discarded on next load — the exact same defect shape this PR just removed the four `ui.*` entries for, and `every_writable_allowlist_entry_has_a_backing_config_field` structurally cannot see it (the guard only checks a section prefix at its base, and `channels` itself is a real field).

Per CLAUDE.md's "Fix what you found — don't punt it to a follow-up" (the bar being: does it require touching a different crate/domain than the one already being worked in? here, no — same file, same PR), this reads like something that should be resolved in this PR rather than left as a "maintainer decision." I understand the two candidate repairs (drop `channels.` entirely, vs. move it to depth-1) both change the write surface rather than just tidying the allowlist, which is presumably why it was deferred — but that's exactly the kind of security-adjacent call this review is supposed to surface rather than silently punt. Given the PR title explicitly promises "guard writable paths against missing fields," leaving one known-dangling path undocumented-as-a-gap in the guard itself (vs. documented-only-in-a-comment) seems like it undercuts that promise. Worth a maintainer call before merge: either close the gap now, or explicitly scope it out in the PR description/changelog as a known follow-up limitation (rather than only a code comment).

---

houko (commented):
Daily automated review of the config read/write gap fix. Summary:

**Code correctness — looks solid.** Walked the actual guard logic rather than trusting the PR description:
- All 14 `OidcProvider` fields and all 11 `ExternalAuthConfig` fields are now present in `redacted_config_json`'s `external_auth` output; cross-checked against the struct definitions in `librefang-types/src/config/types.rs` and every field already carries a correct `#[serde(default)]` (or default fn) from before this PR, so no new struct-field/Default-impl gap was introduced.
- Nothing newly exposed is secret-bearing: `client_secret_env` is the env var name (not the secret value), `client_id` is the public OAuth client id — matches the PR's own claim.
- `require_email_verified` (both the global bool and the per-provider `Option<bool>`) stays correctly absent from `WRITABLE_EXACT_PATHS` / `WRITABLE_SECTION_PREFIXES`, so the read-exposure doesn't create a write regression — confirmed by the new `get_config_exposes_non_writable_external_auth_fields` integration test asserting 403 on write.
- `every_writable_allowlist_entry_has_a_backing_config_field`'s schema-walk (`schema_declares_path`) correctly handles `$ref`, `allOf`/`anyOf`/`oneOf`, and `additionalProperties`; the `ui.*` negative controls and the >50 sanity floor are appropriate guards against a silently-neutered test.
- This PR doesn't add any new writable/hot-reloadable field, so no `build_reload_plan` / `docs/operations/config-reload.md` update was needed here — confirmed `external_auth` classification there predates this PR and is unaffected.
- CHANGELOG additions and new doc comments follow the one-sentence-per-line prose rule.

**One inline comment left** on `crates/librefang-api/src/routes/config/mod.rs` (the `channels.` depth-2-only prefix) — a known dangling write path the PR's own comments document but explicitly defer as a "maintainer decision." Given CLAUDE.md's "fix what you found" bar (same file/PR, not a different crate), flagging for a maintainer call rather than silently agreeing it should stay deferred.

**Separate concern, not code-related:** commit `a89cd61` in this PR's history ("fix(api): break merged sentences in new #6605 comments onto separate lines") is authored as `Claude <noreply@anthropic.com>`. The repo's own `commit-msg` hook and CLAUDE.md policy prohibit Claude/Anthropic attribution on commits reaching this repo, and this is a case of it landing in the author field itself (not just the message body), on a security-adjacent PR. I'm not attempting a rebase/force-push to scrub it — that's a history rewrite on someone else's branch/commits, out of scope for an automated reviewer and requires explicit maintainer sign-off — but wanted to make sure this doesn't get merged without the maintainer being aware. Worth squashing or re-authoring that commit before merge.

No mechanical fixes applied this run (nothing met the ≥90%-confidence bar for an automated edit on a security-sensitive PR); everything above is left for maintainer judgment.

---

crates/librefang-api/src/routes/config/mod.rs
```diff
@@ -579,232 +579,244 @@ pub fn ui_options_overlay(
...
+// A depth-2 write such as `channels.telegram.enabled` is still accepted, still lands a `[channels.telegram]` table in config.toml, and is still discarded by the next load — the same silent-no-op shape #6605 removed the `ui.*` entries for.
+// `every_writable_allowlist_entry_has_a_backing_config_field` cannot see it: the guard checks a section prefix at its base, and `channels` is a real `KernelConfig` field, so the dangling part sits one level below what the guard inspects.
+// The visible half is the mirror image: `ui_sections_overlay` declares a `channels` section and `ConfigPage` posts `<section>.<field>`, so the dashboard renders editors for the three real scalars and every save of one is rejected with 403.
+// Both candidate repairs (drop `channels.` so the surface matches the struct, or move it to depth 1 so the three real scalars become writable) change the write surface rather than the allowlist's internal consistency, so they are a maintainer call, not a mechanical consequence of #6605.
```
houko, line 805: This PR's own audit surfaces (and documents in detail, thank you) that `channels.telegram.enabled` and every other `channels.<vendor>.<field>` path is accepted by `is_writable_config_path`, lands a `[channels.<vendor>]` table in `config.toml`, and is silently discarded by the next load — `ChannelsConfig` no longer has any per-vendor sub-table post-sidecar-migration, only the three depth-1 scalars (`file_download_dir`, `file_download_max_bytes`, `file_upload_max_bytes`).

That's the exact same silent-no-op shape this PR just removed the four `ui.*` entries for, and `every_writable_allowlist_entry_has_a_backing_config_field` can't catch it because the guard only checks the section prefix's base (`channels`, which is real), not what's beneath it.

Per CLAUDE.md's "fix what you found" rule this is arguably in-scope (same crate, same PR's own guard, same defect class), but I'm flagging rather than fixing because the two candidate repairs you outline change the write surface, not just allowlist bookkeeping:
- drop `channels.` from `WRITABLE_SECTION_PREFIXES` (closes a currently-"writable"-but-inert path, no functional change since it was already a no-op), or
- promote the three real scalars to depth-1 `WRITABLE_EXACT_PATHS` entries (makes them actually writable, which they currently aren't despite the dashboard rendering editors for them per your note).

That's a product/security call about what the config-write surface should look like, not a mechanical consequence of the parity-guard work — leaving it to a maintainer to pick, per the guidance in `WRITABLE_DEPTH_2_ONLY_PREFIXES`'s own comment.

---

houko (commented):
Reviewed the config read/write exposure changes against CLAUDE.md. The core change (exposing pre-existing, already-`#[serde(default)]`-backed `OidcProvider`/`external_auth` fields as read-only, plus the allowlist parity/backing-field guards and the shared sidecar tempfile-sequence fix) looks correct and well-tested. One mechanical prose fix applied directly; one pre-existing dead-write-path finding (surfaced by this PR's own audit) left as a comment for maintainer judgment since it's a write-surface security call, not a mechanical fix.

---

crates/librefang-api/src/routes/config/mod.rs
houko, line 805: Given this PR's own mission is closing exactly this defect class (`ui.*`), should `channels.` be resolved here rather than left open?
The comment documents that `channels.telegram.enabled` is a live instance of the same silent-no-op shape the PR just removed four `ui.*` entries for: `ChannelsConfig` has no per-vendor sub-table post-sidecar-migration, so the write is accepted, lands a `[channels.telegram]` table in `config.toml`, and is discarded on next load.
The new `every_writable_allowlist_entry_has_a_backing_config_field` guard can't catch it because it only checks the section prefix's base (`channels`, which is a real field), not the depth-2 leaf.
Since the test at line ~1432 pins this as still-writable rather than closing it, this looks like a conscious deferral of an in-scope fix to a future PR — worth either picking one of the two documented repairs now (dropping `channels.` vs. promoting the three real scalars to depth 1) or explicitly confirming with a maintainer that the security tradeoff needs sign-off before landing.

---

PR #6620, opened 2026-07-28
fix(channels): base the status indicator on supervisor liveness instead of shared per-type traffic

### LINKED ISSUE #6606: Channels status dot reports shared per-type traffic, not per-bot liveness — a dead bot shows green while any sibling on the same channel type is alive
The dashboard's per-channel status indicator cannot distinguish a healthy bot from a dead one. Two independent defects compound, and the second makes the first considerably worse than it looks.

## Defect 1 — the dot is a traffic proxy, not health

`ChannelsPage.tsx` derives the indicator as `msgs_24h > 0 ? "running" : "idle"`.

Real liveness signal already exists and is maintained by the sidecar supervisor — `crates/librefang-channels/src/types.rs` `ChannelStatus { connected, started_at, last_message_at, messages_received, messages_sent, last_error }`, with `connected` set false when the process dies. None of it is read by the UI, and `connected`/`last_error` are not present in the `GET /api/channels` payload at all.

So a healthy-but-quiet channel renders grey, and a channel that died after receiving traffic renders green.

## Defect 2 — `msgs_24h` is a per-channel-*type* aggregate presented as a per-channel number

`crates/librefang-api/src/routes/channels.rs`:

```rust
let m = msgs_24h
    .get(channel_type)              // "telegram"
    .or_else(|| msgs_24h.get(name)) // per-instance name — never reached
    .copied()
    .unwrap_or(0);
row["msgs_24h"] = serde_json::json!(m);
```

The map comes from `crates/librefang-memory/src/usage.rs` `channels_msgs_24h_bulk`:

```sql
SELECT channel, COUNT(*) FROM usage_events
WHERE channel IS NOT NULL AND channel != '' AND timestamp >= ?1
GROUP BY channel
```

`usage_events.channel` stores the channel *type*, not the instance name. On our host that column contains only `api`, `autonomous`, `cron`, `telegram` — no per-bot names — so the `channel_type` lookup always hits and the `name` fallback is unreachable dead code.

**Live evidence:** this host runs six distinct Telegram sidecar channels with very different real traffic (one is a primary personal assistant, another is used a few times a week). All six report the identical `msgs_24h` value, equal to the total across all Telegram channels:

```
$ sqlite3 librefang.db "SELECT channel, COUNT(*) FROM usage_events
    WHERE channel != '' AND timestamp >= datetime('now','-24 hours') GROUP BY channel"
api|12
autonomous|1
cron|5
telegram|35        <-- every one of the six bots reports 35
```

## Combined effect

Because the dot is `msgs_24h > 0` and `msgs_24h` is shared across every channel of the same type, **the indicator is green for all six bots whenever any single one of them has had traffic in the last 24h**. A bot that has been dead for more than a day still shows green as long as one sibling is alive. The indicator conveys "this deployment has some Telegram traffic", not "this bot works".

This is not hypothetical for us: we recently had individual sidecar channels stop delivering outbound messages while the daemon considered turns complete. The dashboard showed nothing unusual, and this indicator is exactly what an operator would check first.

## Suggested fix

1. Key the traffic map by channel instance, or drop the unreachable `name` fallback and label the number honestly as per-type.
2. Base the indicator on `ChannelStatus.connected` / `last_error` / `started_at`, which the supervisor already tracks — and surface those fields in `GET /api/channels` so the UI can read them. That is the signal operators think the dot already represents.

## Environment

- Deployed `v2026.7.21`; cited code byte-identical at `v2026.7.27` and on current `main`.
- Six sidecar channels of the same type on one daemon.

Happy to test a patch.

houko:
Fixes #6606

Diagnosis and reproduction are @nevgenov's.
The operational consequence they documented: individual sidecar channels stopped delivering outbound messages while the dashboard showed nothing unusual.

## The two defects, and why they compounded

**1. The status indicator was a traffic proxy, not a health signal.**
The channel card derived its state from `msgs_24h > 0 ? "running" : "idle"`.
`ChannelStatus { connected, started_at, last_message_at, messages_received, messages_sent, last_error }` was maintained by the sidecar supervisor all along — `connected = true` at `crates/librefang-channels/src/sidecar.rs:922`, `connected = false` at `1340` / `1352` / `1390` / `1762` / `1872` — and was reachable from the API layer through `kernel.channel_adapters_ref()`, but only the QR handler ever read it, and only for `.qr`.
None of those six fields appeared on `GET /api/channels` at all.

**2. The traffic number was a per-channel-type aggregate presented as per-bot.**
`usage_events.channel` stores the channel *type*, so `sidecar_channel_rows`' `msgs_24h.get(channel_type).or_else(|| msgs_24h.get(name))` always hit on the first lookup and the per-instance fallback was unreachable dead code.
Six Telegram sidecars therefore all reported the same number: the total across all six.

**Compounded, these are worse than either alone.**
Because the indicator was *derived from* the shared aggregate, one busy bot turned all six cards green.
The reporter's dead `tg-alerts` did not merely fail to go red — it was actively painted green by its healthy siblings' traffic, for over a day.
Neither the colour nor the number carried any information about the instance the operator was looking at.

## Changes

- `crates/librefang-api/src/routes/channels.rs` — configured rows now carry `connected`, `started_at`, `last_message_at`, `messages_received`, `messages_sent`, `last_error` per instance, plus a `supervised` flag saying whether an adapter is registered for that name at all.
  Adapters are keyed by the sidecar instance name (`SidecarAdapter::new` sets `name: config.name.clone()` at `sidecar.rs:1565`; `channel_bridge.rs` inserts under `adapter.name()`), which is the same key this loop iterates, so the lookup is per-bot.
- `crates/librefang-api/dashboard/src/lib/channelLiveness.ts` (new) — the state mapping, shared by both pages that render this payload.
- `crates/librefang-api/dashboard/src/pages/ChannelsPage.tsx` — cards, details drawer and card `aria-label` driven from the mapping; per-instance `N in / N out` sub-line; error state for a failed fetch.
- `crates/librefang-api/dashboard/src/pages/CommsPage.tsx` — the Channels tab's tile extracted to a `ChannelTile` component and its badge switched from `configured` to the shared mapping.
- `crates/librefang-cli/src/commands/channel.rs` — `librefang channel list` gains CONNECTED and IN/OUT columns plus a per-channel sticky-error footnote, fed the raw supervisor fields rather than a second copy of the state mapping.
- `crates/librefang-memory/src/usage.rs` — `UsageStore::channels_msgs_24h_bulk` renamed to `channel_type_msgs_24h_bulk`, doc comment stating the grouping.
  The SQL and its `GROUP BY channel` are byte-identical.
- `crates/librefang-api/dashboard/src/api.ts` — `ChannelItem` gains the seven new fields with the scope of each spelled out.
- Locales: 12 new `channels.*` entries in each of en/ko/uk/zh — `card_aria`, `configured_label`, `last_error`, `last_error_sticky_hint`, `last_message_at`, a 7-state `liveness` group, `load_error`, `msgs_24h_by_type`, `not_supervised_hint`, `started_at`, `traffic_in_out`, `traffic_since_start` — with `channels.msgs_24h` removed; 6 new CLI keys in each of en/ko/uk/zh-CN with `channel-header-msgs-24h` removed.

## The state mapping

Seven states, evaluated in this order:

| state | colour | condition |
|---|---|---|
| `not_supervised` | amber | `!supervised` |
| `failed` | red | `!connected && last_error` |
| `stopped` | red | `!connected && started_at` |
| `starting` | amber | `!connected`, no error, never spawned |
| `degraded` | amber | `connected && last_error` |
| `active` | green | `connected`, traffic > 0 |
| `connected` | green | `connected`, no traffic |

Three non-obvious decisions, with the evidence that decided them:

- **`connected` with a `last_error` is amber, not red.**
  `last_error` is sticky: the supervisor sets it on a sidecar `error` event, a failed spawn (`sidecar.rs:1762`) and a circuit-break (`1389-1392`), and never clears it — not even on the `connected = true` of the successful respawn that follows.
  Trade-off: red would over-report (a channel that failed once six weeks ago and has run fine since is not broken), green would under-report (the operator would never learn it failed).
  Clearing the field on reconnect was not available — the circuit-break regression test at `sidecar.rs:3548` depends on the persistence.
- **A configured channel with no registered adapter is amber, not grey.**
  Trade-off: grey is the visually honest colour for "we don't know", but grey reading as "fine, just quiet" is precisely the misreading this PR exists to remove.
  The API layer genuinely cannot distinguish "start failed and the registration was rolled back" from "added to config.toml and never reloaded", so amber with the label "Not started" states what is known without asserting health.
- **A quiet connected channel stays green.**
  `active` and `connected` differ by label only, never by colour.
  Colouring a healthy-but-quiet channel differently is the original bug in its other direction.

The state text is visible rather than `sr-only` and is interpolated into the card's `aria-label` (`channels.card_aria`), because the card's own label overrides its inner text for assistive tech — without that, colour would be the only carrier of meaning.
The Badge's decorative dot is `aria-hidden`.

## What happened to the traffic figure

**It remains per-channel-type. The keying did not change, so there is no migration and no transitional period.**
`usage_events.channel` stored the channel type before this PR and stores the channel type after it; the `usage.rs` diff is a rename plus a doc comment, and the SQL and `GROUP BY channel` are byte-identical.
There is no mix of type-level and instance-level rows to reconcile, and no window in which numbers read wrong while old rows age out.

Re-keying that column per instance is not available.
It is written from `SenderContext.channel`, which the bridge derives from `channel_type_str(&ChannelMessage.channel)` — `ChannelMessage` carries only a `ChannelType`, so the sidecar instance name never reaches the write site.
`SenderContext.channel` also feeds `SessionId::for_channel(agent, channel)` and the auth `identify(&channel, …)` binding, so re-pointing it at the instance name would silently re-derive every existing channel session and auth binding.

So the UI stops implying per-bot instead:

- The JSON key is now `msgs_24h_channel_type`, published alongside the `channel_type` it covers. The old `msgs_24h` key and its unreachable per-instance fallback are removed entirely (asserted absent by test).
- The figure moved off the card into the details drawer, labelled `channels.msgs_24h_by_type` = "24h messages (all {{type}} channels)". The CLI column header is "24H MSGS (BY TYPE)".
- The card's sub-line now reads `telegram · 34 in / 21 out` from the supervisor's own per-instance counters. Those are since-adapter-creation (they live on the adapter's `Arc<Mutex<ChannelStatus>>` and survive supervised restarts, so they are neither a 24h figure nor since-`started_at`); the drawer row says so, and the card carries the scope on a `title` since it has no room to spell it out and that position previously held a 24h number.
- `UsageStore::channels_msgs_24h_bulk` → `channel_type_msgs_24h_bulk`, so the grouping cannot be misread at the call site.

## Also in this PR

The mapping lives in `src/lib/channelLiveness.ts` rather than on a page because two pages render the same `useChannels()` payload.
The Comms page's Channels tab painted its badge from `configured` alone (`configured ? "Online" : "Setup"`), so the reporter's dead bot kept a green ONLINE badge — and a green icon ring — on that page even once the Channels page reported it correctly.
It now renders the same verdict from the same helper, with the icon ring following the badge variant so the two cannot disagree.
That tab lists unconfigured catalog rows (the Channels page filters them out), and those carry none of the supervisor fields, so they would read `not_supervised` — "broken" for something that was simply never set up.
`ChannelTile` therefore branches on `configured` before consulting the mapping and keeps the "Setup" badge for catalog rows; both halves are pinned by tests.

A failed `GET /api/channels` used to fall through to the "no channels yet" CTA, because `channels` is `[]` on error — an unreachable daemon looked like a clean install on the page whose whole job is now health signalling.
It renders `ErrorState` instead.
Gated on `isError && channels.length === 0` rather than `isError` alone: the query polls every 30s and retains its last successful data across a failed refetch, so the bare flag would blank a working list on one transient blip.

New prose in the touched files (the `sidecar_channel_rows` doc block, the `usage.rs` doc comment, the `ChannelItem` JSDoc, the new module's doc block) breaks only at sentence boundaries per CLAUDE.md's prose-wrapping rule.
Pre-existing paragraphs written under the old column-wrap convention were left alone.

## Verification

| command | outcome |
|---|---|
| `cargo check --workspace --lib` | clean, exit 0 |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean, exit 0 |
| `cargo fmt --check` on the changed Rust files | clean |
| `cargo test -p librefang-api --test channel_liveness_test` | 4 passed |
| `cargo test -p librefang-api --test dead_route_audit_test` | 2 passed |
| `cargo test -p librefang-api --test openapi_path_coverage_test` | 1 passed |
| `cargo test -p librefang-api --lib` | 874 passed |
| `cargo test -p librefang-memory --lib` | 333 passed |
| dashboard `npx tsc --noEmit` | clean |
| dashboard `npx vitest run` | 917 passed / 89 files |
| dashboard `npx eslint` on the changed files | 0 errors (7 warnings, all pre-existing on untouched lines) |

Tests added:

- `crates/librefang-api/tests/channel_liveness_test.rs` — `same_type_sidecars_report_independent_liveness` (seeds 35 real `UsageRecord`s through the real SQLite substrate and asserts two same-type sidecars report independent `connected` / counters while sharing the type figure), `channel_with_traffic_but_dead_process_reports_disconnected` (the reporter's case), `configured_sidecar_without_adapter_reports_unsupervised`, `liveness_is_keyed_by_instance_name_not_channel_type`.
- `crates/librefang-api/dashboard/src/lib/channelLiveness.test.ts` — 9 mapping cases plus `livenessLabel` exhaustiveness (a state added without a label would fall through and render an empty badge, i.e. colour-only).
- `crates/librefang-api/dashboard/src/pages/CommsPage.test.tsx` (new file) — 5 cases: dead sidecar reads `failed` while a live sibling reads `active`; an unconfigured catalog row keeps `common.setup` and does **not** borrow `not_supervised`; a configured-but-unsupervised row reads `not_supervised`; the sticky error rides the badge tooltip; name and description still render.
- `crates/librefang-api/dashboard/src/pages/ChannelsPage.test.tsx` — 6 render cases including the issue's exact case (`connected: false` with 34/21 traffic and `msgs_24h_channel_type: 35` renders `stopped`, not `active`), plus 2 new error-state cases (empty + `isError` shows the error rather than the clean-install CTA; cached data + `isError` keeps rendering the list).

**Test relocation — no coverage was dropped.**
`ChannelsPage.test.tsx` shows ~88 deleted lines: the 9 pure `channelLiveness` unit cases moved verbatim to `src/lib/channelLiveness.test.ts` when the helper moved out of the page, with no assertion changed.

**Pre-existing expectations changed: one.**
`msgs_24h: 12` left the `makeChannel` fixture in `ChannelsPage.test.tsx`, replaced by the new field set, because the key no longer exists on the payload.
No assertion was altered.
`common.running` / `common.idle` (dropped from the card) are still referenced by `Pill.tsx`, `AgentsPage.tsx` and `SessionsPage.tsx`, so the dead-key test stays green; `channels.msgs_24h` is removed from all four dashboard locales and `channel-header-msgs-24h` from all four CLI locales.

**Generated fixtures unchanged, deliberately.**
`openapi.json`, `sdk/` and `xtask/baselines/` are untouched: `list_channels` returns the untyped `crate::types::JsonObject`, and the added row-shape note sits in a plain `//` block that never reaches the utoipa operation description.
`dead_route_audit_test` and `openapi_path_coverage_test` both pass.

**Full-crate suite — three pre-existing failures, none from this change.**
`cargo test -p librefang-api --no-fail-fast` exits 101 locally: `api_integration_test` and `agents_routes_integration` SIGABRT under parallel kernel/router boots on macOS (both reproduced on the stashed baseline; 88/88 and 54/54 with `--test-threads=1`), `media_routes_integration` fails 3 when `OPENAI_API_KEY` is present in the shell (18/18 with `env -u OPENAI_API_KEY`), and `sidecar_toml_test` fails 3-4 on a temp-path collision (11/11 with `--test-threads=1`, root cause below).
CI sees none of these — nextest runs process-per-test.

## Out-of-scope — questions for the maintainer

- `crates/librefang-api/src/routes/sidecar_toml.rs:142` and `:184` declare two *distinct* function-local `static SEQ: AtomicU64` counters, one in `upsert_sidecar_block` and one in `remove_sidecar_block`, so a concurrent upsert + remove in one daemon process both stage `.config.toml.tmp.<pid>.0` in the same directory: one `fs::write` truncates the other (observed as a torn write dropping an array key), then the loser's `fs::rename` fails ENOENT.
  The comment at `:137-141` claims the counter guards concurrent threads within the process, which it does not across the two functions.
  This is a real config-corruption race in the config *write* path, unrelated to the channels *read* path this PR touches, and it needs its own concurrent-upsert/remove test.
  Should it get its own issue rather than being folded in here?
- The sticky `last_error` has no path back to green: nothing clears it, including the `connected = true` of a successful respawn (`sidecar.rs:922-924`), so a channel that fails one spawn and then runs healthily for weeks reads amber `degraded` for the daemon's lifetime.
  A clean fix needs `last_error_at` on `ChannelStatus` compared against `started_at` — a change in `librefang-channels`, and `sidecar.rs:3548` depends on the current persistence.
  Do you want that crate change, and in this PR or a separate one?
- After "Reload channels", `reload_channels_from_disk` re-registers the adapter map while `connected` only flips true when the child announces ready (`sidecar.rs:922`), and `useChannels` polls at 30s (`lib/queries/channels.ts:27`) — so every card can read amber `starting` for up to 30s after a reload the operator just watched succeed.
  Would you like a short post-mutation refetch on `useReloadChannels`, or is the transient window acceptable?
- `crates/librefang-api/dashboard/src/locales/zh.json` — `channels.liveness.failed` is `失败` rather than `启动失败` ("start failed"), because the state covers runtime exits (`"sidecar exited with status 1"`) and not only spawn failures, and en/ko/uk all read plain "Failed" / "실패" / "Збій".
  That overrides a translator's wording on a semantic reading rather than fixing a bug — does it read right to you?
- The card sub-line's leading token changed from `c.category` (always the literal `"sidecar"` for these rows) to `c.channel_type` (`"telegram"`).
  Not requested, and it is one line — keep it or revert it?


houko (commented):
Automated CLAUDE.md compliance pass over this diff (excluding the "Out-of-scope — questions for the maintainer" items in the PR body, which are for a human).

Checked: prose-wrapping (doc comments, CHANGELOG, locale files), HashMap/BTreeMap at LLM-prompt boundaries (n/a here — the touched maps are dashboard/CLI payload data, not anything reaching an LLM prompt), dashboard data-layer rule (no inline `fetch`/`api.*`, all via `useChannels()`), config-field Default-impl pattern (n/a, no `KernelConfig` fields added), commit/PR attribution.

**One finding, not auto-fixed:**

The commit body (`754aab0`) violates the "Commit message bodies" clause of CLAUDE.md's prose-wrapping rule — each paragraph is a single long line containing multiple sentences (e.g. paragraph 2: three sentences on one line) instead of one sentence per line. This is normally a mechanical fix, but correcting it requires rewriting the existing commit (`git commit --amend`) and force-pushing the branch, which this review session is under a hard instruction never to do. Flagging for @houko or a maintainer to reword before merge (or it may be moot if the repo squash-merges with an editable message).

Everything else in the diff — `crates/librefang-api/src/routes/channels.rs`, `crates/librefang-memory/src/usage.rs`, the new `channelLiveness.ts`/tests, `ChannelsPage.tsx`/`CommsPage.tsx`, `channel.rs` (CLI), and all four locale files — reads as compliant: new doc-comment prose breaks at sentence boundaries, no HashMap sits at a documented determinism boundary, and no inline API calls were added outside the existing query-hook layer. No inline line comments needed.

---

PR #6622, opened 2026-07-28
fix(runtime): let the global require_approval list survive exec_policy Full, and guard daemon lifecycle commands

houko:
Refs #6594

Both items here come from #6594's "Defensive (optional)" list.
They were held back from #6603 and #6610 as different-crate work, and @nevgenov answered the open questions on them in the #6610 thread before they were picked up here.
#6594 is already closed by #6603, so this is a Refs rather than a closing keyword.

## Change A — `exec_policy.full_mode_skips_approval`

New `bool` on `ExecPolicy`, default `true`, readable on `GET /api/config` and writable through `POST /api/config/set`.
`mode = "full"` waived the global `approval.require_approval` list for `shell_exec` in addition to skipping allowlist validation, so an operator who set `Full` for one agent silently lost `require_approval = ["shell_exec"]` for it with nothing on any surface saying so.
With the flag off, `Full` waives only command validation and `[approval]` decides who must confirm, exactly as under `allowlist`.

**This makes a documented decision overridable rather than fixing an oversight.**
The pre-existing `// SECURITY:` comment at the gate described the coupling as intentional, so that comment was edited rather than appended to — as written it asserted something this change makes false.
The rewritten comment states the asymmetry and why.

**The default is deliberately not flipped, and I would like a maintainer call on whether it should be.**
Two facts are the trade-off, and they are load-bearing only together:

- `ApprovalPolicy::default()` ships with `shell_exec` in `require_approval`.
- `Kernel::spawn` promotes any standalone agent whose `capabilities.tools` contains `shell_exec` or `*` and which declares no `exec_policy` to `mode = "full"` (`kernel/spawn.rs:237`).

So on a stock install this waiver is the only reason ordinary agents run shell commands unattended, and flipping the default would make every default install prompt on every shell command rather than expressing a new operator intent.
The question for the maintainer: is "secure by default, noisy on upgrade" the right trade here, or is opt-in correct?
I did not decide it unilaterally.

`safe_bins_skip_approval` (#6000) is deliberately left unconditional.
It is by its own name an explicit approval opt-out, so honouring `require_approval` on top of it would leave the flag meaning nothing; `Full` is a command-validation mode that never claimed to speak for `[approval]`.
A per-user RBAC `NeedsApproval` still forces the approval queue in either position of the new flag — the `!force_approval` conjunct is unchanged and is covered by a test run at the default `true` on purpose, since with the flag off the global list would queue the call anyway and that conjunct would never be exercised.

**Reload caveat, now documented.** `exec_policy` is classified restart-required (`docs/operations/config-reload.md:151`) and is copied into each agent's manifest at spawn/restore (`kernel/spawn.rs:237`, `kernel/boot.rs:2165`, `kernel/hands_lifecycle.rs:444`), and the agent loop reads the manifest copy (`agent_loop/tool_call.rs:1060`).
So setting `false` and calling `POST /api/config/reload` leaves already-running agents untouched — kill the agent and let it respawn, or restart the daemon.
For a security opt-in that silent no-op reads as "nothing needed approval", which is the worst available failure mode, so it is called out in all four locale pages and the CHANGELOG.

### Is the new writable leaf an escalation?

No. It reaches no permissiveness state `POST /api/config/set` could not already reach: `exec_policy.mode` is itself writable, and setting it to `full` already waives the global require list today.
Setting the new leaf `true` restores a state a config write could already produce; the only *novel* reachable state is the restrictive `false`.
Writability was proven empirically rather than by reading — removing the `GET /api/config` read entry makes `every_writable_config_leaf_is_readable` fail naming exactly `exec_policy.full_mode_skips_approval`.

### Correcting the record from #6610

The #6610 out-of-scope note cited `subprocess_sandbox.rs:701-708` as evidence that no dangerous-pattern screening happens under `Full`. That was wrong.
That function is `validate_command_allowlist` (`subprocess_sandbox.rs:696`); under `ExecSecurityMode::Full` it logs a warn and returns `Ok(())`. It is the allowlist validator and performs no pattern screening in *any* mode, so it was never the screen.

The screen is a separate, unconditional block at `dispatch.rs:770-789`, whose own comment reads "Runs in Manual mode for all exec policies (including Full) because even explicitly-trusted agents should not silently execute commands like `rm -rf /` or fork bombs."
It sits outside the `if let Some(policy)` exec-policy branch and outside the `if !is_full_exec` guards that *do* gate the metacharacter check (`dispatch.rs:741`) and the taint check (`dispatch.rs:758`).
@nevgenov confirmed from the #6594 incident that it fires under `Full`.
So this PR extends an existing list; it adds no mechanism and no new gate.

### The read/write line

Both entries are written against **mutating statement forms, not bare verbs**, and are scoped to `librefang.db` rather than any `.db`:

- `insert into`, `replace into`, `update <table> set`, `delete from`, `drop table|index|view|trigger`, `alter table`, matched against `librefang.db`.
- Output redirection that would truncate the file.

A bare `\bdelete\b` would block `… WHERE tool = 'delete'`, and a bare `replace` would block the `replace(x,y,z)` scalar function in ordinary `select`s.
This matters because a match is a hard block in `manual` mode: **the reporter's own read-only `sqlite3` diagnostic from #6606 is covered verbatim by a negative test** (`read_only_daemon_database_queries_stay_safe`), since blocking it would have blocked the investigation that produced that report.
`select`, `.schema`, `.dump`, a redirect into a separate backup file, and a `select` that merely mentions `'delete'` as a value are all pinned as safe.

The redirect anchor is `([\s;&|)]|$)` rather than `\b`: `\b` is satisfied by the `.` in `librefang.db.bak` and would block an ordinary `.dump` backup, while a bare `(\s|$)` let `> librefang.db; echo done` walk past. A test caught the latter.

### Lifecycle entry

Matches `start` / `stop` / `restart` by bare name, by path (`target/release/librefang`, `/usr/local/bin/librefang`), with the Windows `.exe` suffix, through the `gateway` alias subcommand (`cli.rs:1383-1423` — same daemon, same blast radius), and behind the CLI's one global option with its separate value token.

That last form is the review fix in the second commit. The original pre-verb group `(-[^\s]+\s+)*` absorbs a flag token then requires the next token to be another flag or the verb, so `librefang --config ~/.librefang/config.toml stop` did not match — `--config` is the only `global = true` option (`cli.rs:50-51`) and its value is a separate token, i.e. the denylist missed a documented invocation of the exact command it exists to block.
The group now optionally absorbs one non-flag value token per iteration while still **requiring a leading `-` at the start of every iteration**, which is what keeps it from eating a subcommand name and matching `librefang <subcommand> … start`.
The test that claimed to cover this asserted on `librefang --json restart`, which clap rejects (`--json` is not global; `Restart` declares only `tail` / `foreground`, `cli.rs:90-99`) — replaced with the two real forms.

`librefang status` and every other read-only subcommand stay unmatched. `ServiceCommands` is `Install` / `Uninstall` / `Status` only (`cli.rs:1688-1712`), so there is no `service stop` to cover.

### Known false positives, stated plainly

The gate is `shell_exec`-only, so Grep/Read tools are unaffected, and `allow_for_session("stop/restart the LibreFang daemon")` is a recoverable escape.
Still, searching or editing this repo's own prose trips the lifecycle entry: `rg "librefang start" .` and `grep -rn "librefang restart" docs/` are blocked.
This is the same class as the pre-existing `\bdelete\s+from\b` blocking `grep "delete from"`, and consistent with the file's substring-matching design, but the trigger phrase appears in prose far more often than SQL does.
The widening extends that surface: the `librefang --config <path> stop` line in this PR's own CHANGELOG entry now matches too.
A sweep for `librefang --` across `crates/**/*.rs` returns only `dangerous_command.rs` itself (one comment, two test strings) — nothing that reaches `check` at runtime.

Documented non-catches, in the code comment: the SQL entry requires filename-before-statement (the canonical `sqlite3 <db> <sql>` form), so `sqlite3 -cmd "insert into …" librefang.db` and the stdin-redirect sibling `sqlite3 librefang.db < mutate.sql` are unmatched, as are SQLite conflict-clause forms (`INSERT OR IGNORE INTO`, `UPDATE OR REPLACE <t> SET`). The SQL entry also spans a whole command line rather than one invocation, so it over-matches across a `&&` chain — conservative, and a scope-precision limit rather than a defect.

## Docs

`configuration/page.mdx`, `configuration/features/page.mdx` and both `zh` twins: two rows on the `[exec_policy]` table (`safe_bins_skip_approval` was undocumented in every locale until now), prose that `mode` and approval are separate decisions, why the default is not flipped, that the denylist runs under every mode including `full`, and the reload caveat above.

`security/sandboxing/page.mdx` and its `zh` twin: this diff takes `DANGEROUS_PATTERNS` from 39 to 42 entries, which de-synced the "**39 regular-expression patterns** grouped into 11 categories" figure that page stated — that number was accurate on `main` and this branch is what broke it, so it is corrected to 42 / 13.
Making the stated count true also required adding the rows the table was missing, since 13 over an 11-row table is the same defect: the new daemon-database-mutation row, and the container-privilege-escalation row that was already absent before this branch.
The source pointer said `dangerous_commands.rs`; the file is `dangerous_command.rs`.
Two example-trigger tokens in the Service management row are dropped because I verified no pattern implements them: there is no `service` pattern, and `kill -9 1` is not matched (`kill -9 -1` is, under Process termination).

All four/six locale twins are updated symmetrically. Pure markdown tables and paragraphs, no JSX; no docs build was run.

## Verification

Build phase (first commit, `d2a90c47`):

- `cargo check -p librefang-types --lib`, `-p librefang-runtime --lib`, `-p librefang-api --lib`, then `cargo check --workspace --lib` — all clean.
- `cargo clippy -p librefang-types -p librefang-runtime -p librefang-api --all-targets -- -D warnings` — clean.
- `cargo test -p librefang-runtime --lib dangerous_command` (29 ok), `--lib approval` (21 ok), `--lib full_mode` (10 ok); `cargo test -p librefang-types --lib exec_policy` (8 ok); `cargo test -p librefang-api --lib config_read_write_parity` (3 ok), `--test config_schema_overlay` (4 ok), `--test config_routes_integration -- --test-threads=1` (35 ok).

Review phase (second commit, `89a63087`) changed code in **`librefang-runtime` only** — the other seven files are CHANGELOG and docs prose. So verification was re-run on that one crate rather than repeating the three-crate form:

- `cargo test -p librefang-runtime --lib dangerous_command` — 30 ok.
- `cargo clippy -p librefang-runtime --all-targets -- -D warnings` — clean.

**Do not read this as workspace coverage.** Clippy was scoped to the crates touched, never `--workspace`, and `librefang-kernel`'s 1335-test lib suite was deliberately not run because the diff does not touch that crate. CI covers both.

New tests: `exec_policy_full_mode_skips_approval_defaults_true_on_upgrade` (pins the deserialization path real installs upgrade through, which struct-literal tests do not exercise); `test_shell_exec_full_mode_skips_approval_by_default`; `test_shell_exec_full_mode_honours_require_approval_when_flag_disabled`; `test_shell_exec_full_mode_flag_disabled_skips_when_not_in_require_list`; `test_safe_bins_skip_approval_unaffected_by_full_mode_flag`; `test_shell_exec_full_mode_respects_rbac_needs_approval`; `librefang_daemon_lifecycle`; `librefang_read_only_subcommands_stay_safe`; `librefang_subcommand_names_are_not_absorbed_as_flags`; `mutating_sql_against_daemon_database`; `read_only_daemon_database_queries_stay_safe`; `redirect_clobbering_daemon_database`; `mutating_sql_against_other_databases_is_not_this_entry`.

Mutations checked, each reverted byte-exact and re-run green:

| Mutation | Result |
|---|---|
| `default_full_mode_skips_approval()` → `false` | `test_shell_exec_full_mode_skips_approval_by_default` and `exec_policy_full_mode_skips_approval_defaults_true_on_upgrade` both FAILED |
| drop `&& p.full_mode_skips_approval` from the gate | `test_shell_exec_full_mode_honours_require_approval_when_flag_disabled` FAILED — command executed instead of queuing |
| remove all three new `dp!` entries | the 3 new positives FAILED; the negatives passed both before and after, as expected |
| remove the `GET /api/config` read entry | `every_writable_config_leaf_is_readable` FAILED naming the leaf |
| revert the lifecycle pre-verb group to `(-[^\s]+\s+)*` | `librefang_daemon_lifecycle` FAILED on precisely the `--config <path> stop` assertion — the review fix is load-bearing, not cosmetic |
| redirect anchor `(\s|$)` | `redirect_clobbering_daemon_database` FAILED on `> librefang.db; echo done` |

Test 3 needed a mock able to express "tool not in `require_approval`" — `ApprovalKernel::requires_approval` returned `tool_name == "shell_exec"` unconditionally. Added `requires_approval_override: Option<bool>`, `None` preserving existing behaviour, at all 21 `Arc::new(ApprovalKernel {` sites.

Not regenerated, stated rather than skipped: `kernel_config_schema.golden.json` — its test is `#[ignore]`d with pre-existing drift on `main`, so CI does not run it and this change cannot newly break it; regenerating would mix unrelated drift into the diff. I am **not** claiming the fixture is consistent with the new field — it is not. `openapi.json` contains zero occurrences of `ExecPolicy`. `docs/operations/config-reload.md` gets no row: its table is top-level `KernelConfig` fields (`exec_policy` is already there as `R`) and `doc_reload_table_matches_classified_reload_fields` asserts set equality both ways, so a nested-leaf row would fail its `extra_in_doc` assertion.

## Out-of-scope — questions, not decisions

- **macOS service-manager forms.** `launchctl stop com.librefang.daemon` and `launchctl unload ~/Library/LaunchAgents/ai.librefang.daemon.plist` both bounce the daemon and match nothing — the same blast radius as the entry added here, on the reporter's own platform, and `librefang service install` writes exactly that LaunchAgent (`cli.rs:1691`). The pre-existing `systemctl` entry covers the Linux equivalents. I did not add it: a new pattern family needs its own negative set and false-positive derivation, which would invite a second review round on a diff that has already had one. Add it here, or as a follow-up?
- **`rm librefang.db`** is caught by nothing (existing `rm` patterns need `-r` or a leading `/`). Different verb family from this PR's brief. Worth an entry?
- **The sandboxing page's Example Triggers column is unaudited beyond the row I rewrote.** I corrected the two false tokens in Service management because I was rewriting that cell, and stopped there — so `shred` (Filesystem deletion), `killall` (Process kill) and `> /boot/` (System file overwrites) remain, and a grep for each returns zero implementing patterns. Should the whole column be audited in its own PR, or do you want the remaining three fixed here?
- **Two further claims on that same page that I did not verify and did not touch:** it documents a `[shell] dangerous_commands` config key, and a `POST /api/shell/allowlist` endpoint. `grep -rn "dangerous_commands" crates/ --include="*.rs"` and `grep -rn "shell/allowlist" crates/librefang-api/src/` both return zero hits. Since I just fixed the source-filename typo on that page, the asymmetry is deliberate rather than an oversight — rewriting the Approval Modes and allowlist sections is a distinct job. Own issue?
- **SQL entry gaps** listed under "Known false positives" above (conflict clauses, stdin redirect, flag-first invocation, `&&`-chain span). Tighten now, or leave as documented limits?

## Credit

@nevgenov diagnosed #6594, answered the open questions on both of these in the #6610 thread, and corrected the `subprocess_sandbox.rs` premise that would otherwise have sent Change B after the wrong code path.


houko (commented):
Automated review pass on this branch found one item worth a human's attention, and made one mechanical fix directly.

**Mechanical fix pushed (512cfe406, "docs: wrap remaining new prose comments at sentence boundaries"):** the new `//`/`///` comment blocks added in `dangerous_command.rs`, its test module, and the `serde_compat.rs` upgrade-path test packed multiple sentences onto single hard-wrapped lines, and one paragraph added to `docs/security/sandboxing/page.mdx` was hard-wrapped mid-sentence at ~80 columns. Both violate this repo's one-sentence-per-line prose convention (CLAUDE.md). A sibling commit already on this branch (43e51dff8) applied the same fix to the `exec_policy` docs prose but didn't reach these source comments — this commit is a reflow-only follow-up covering the rest. No code or test logic changed; every changed line is a comment.

**Flagging, not fixing — commit `43e51dff8` ("docs: wrap new exec_policy prose at sentence boundaries") is authored as `Claude <noreply@anthropic.com>`.** That's AI/Anthropic attribution in commit metadata, which this repo's own convention (CLAUDE.md → "No Claude / Anthropic / AI attribution in commit messages, PR bodies, or issue comments") prohibits. The `commit-msg` hook only inspects the message body — which is clean here — not the author field, so it didn't catch this one. Fixing it means rewriting already-pushed history (`git commit --amend --author=... ` + force-push), which isn't something I'll do unilaterally on a shared branch; leaving it for @houko or a maintainer to correct before merge.

Rest of the diff (the `exec_policy.full_mode_skips_approval` decoupling and the two new dangerous-command denylist entries) looks solid: the regression tests explicitly cover the `Full` + flag-on / flag-off / RBAC-override / safe-bins-independence matrix, the new config field follows the struct-field + `#[serde(default)]` + `Default` impl ritual, and the `GET /api/config` read/write parity test was updated. No AI attribution found elsewhere in the diff.

---

PR #6623, opened 2026-07-28
fix(runtime): wire MAX_CONTENT_CHARS into the page-extraction script

nevgenov:
## Type

- [x] Bug fix

## Summary

`MAX_CONTENT_CHARS` documents the cap on how much extracted page text reaches the model. Nothing reads it.

```rust
#[allow(dead_code)]
const MAX_CONTENT_CHARS: usize = 50_000;
```

The cap that actually applies is a literal written twice inside the extraction script's JS string:

```js
if (content.length > 50000) content = content.substring(0, 50000) + '\n... (truncated)';
```

So editing the constant changes nothing, and the `#[allow(dead_code)]` suppresses the single warning that would have said so. Anyone tuning the limit — the obvious reason to touch that constant — gets a silent no-op.

This is reachable in practice rather than theoretical: a Wikipedia article measured here yields 73,447 characters of page text and hits the truncation, so the value is load-bearing on ordinary pages, not just pathological ones.

## Changes

- The script becomes `EXTRACT_CONTENT_JS_TEMPLATE` with a `__MAX_CONTENT_CHARS__` placeholder, substituted once through a `LazyLock<String>` exposed as `EXTRACT_CONTENT_JS`.
- `#[allow(dead_code)]` dropped, since the constant is now read.
- Call sites take `&EXTRACT_CONTENT_JS`; substitution happens once per process, not per call.

No behaviour change — the substituted value is the same `50000`.

## Attribution

## Testing

- [x] `cargo test -p librefang-runtime` and `cargo test -p librefang-types` pass (scoped per CLAUDE.md; 2147 and 891 tests)
- [ ] Live integration tested (if applicable)

`test_extract_content_js_uses_max_content_chars` asserts three things: the placeholder is gone from the built script, the built script contains the constant's value, and the *template* still contains the placeholder. That last assertion is the one that matters — without it, deleting the placeholder from the template would restore exactly the silent decoupling this fixes, with the other two assertions still passing.

No live browser run: this changes only which string is sent, and the string is identical to what shipped before.

## Security

- [x] No new unsafe code

### DIFF
--- a/changelog.d/fixed/6623-wire-max-content-chars.md
+++ b/changelog.d/fixed/6623-wire-max-content-chars.md
@@ -1,0 +1,4 @@
+Substitute `MAX_CONTENT_CHARS` into the page-extraction script instead of leaving the cap hard-coded in the JavaScript, so the constant that documents the limit is the one that enforces it.
+`MAX_CONTENT_CHARS` was declared, marked `#[allow(dead_code)]`, and read by nothing; the script truncated at a literal `50000` written twice inside the JS string.
+Editing the constant to change how much page text reaches the model therefore did nothing at all, and the `#[allow]` suppressed the one warning that would have said so.
+The script is now a template with a `__MAX_CONTENT_CHARS__` placeholder substituted once through a `LazyLock`, and a test asserts both that the placeholder is gone from the built script and that it is still present in the template — the second half is what keeps the two from silently drifting apart again (#6623) (@nevgenov)
--- a/crates/librefang-runtime/src/browser.rs
+++ b/crates/librefang-runtime/src/browser.rs
@@ -17,7 +17,7 @@ use librefang_types::config::BrowserConfig;
 use serde::{Deserialize, Serialize};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU64, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 use std::time::{Duration, Instant};
 use tempfile::TempDir;
 use tokio::io::AsyncBufReadExt;
@@ -34,7 +34,7 @@ const CDP_CONNECT_TIMEOUT_SECS: u64 = 15;
 const CDP_COMMAND_TIMEOUT_SECS: u64 = 30;
 const PAGE_LOAD_POLL_INTERVAL_MS: u64 = 200;
 const PAGE_LOAD_MAX_POLLS: u32 = 150; // 30 seconds
-#[allow(dead_code)]
+/// Cap on the extracted page text handed back to the model.
 const MAX_CONTENT_CHARS: usize = 50_000;
 const BROWSER_ENV_ALLOWLIST: &[&str] = &[
     "PATH",
@@ -661,7 +661,7 @@ impl BrowserSession {
     }
 
     async fn cmd_read_page(&self) -> BrowserResponse {
-        match self.cdp.run_js(EXTRACT_CONTENT_JS).await {
+        match self.cdp.run_js(&EXTRACT_CONTENT_JS).await {
             Ok(val) => {
                 let parsed: serde_json::Value = val
                     .as_str()
@@ -767,7 +767,7 @@ impl BrowserSession {
 
         let content_val = self
             .cdp
-            .run_js(EXTRACT_CONTENT_JS)
+            .run_js(&EXTRACT_CONTENT_JS)
             .await
             .unwrap_or_default();
         let content_obj: serde_json::Value = content_val
@@ -1077,7 +1077,10 @@ impl BrowserManager {
 // ── Embedded JavaScript ────────────────────────────────────────────────────
 
 /// JavaScript to extract readable page content as markdown.
-pub(crate) const EXTRACT_CONTENT_JS: &str = r#"(() => {
+/// Page-extraction script, with `__MAX_CONTENT_CHARS__` still to be substituted.
+///
+/// Use [`EXTRACT_CONTENT_JS`] rather than this template.
+const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
     const title = document.title || '';
     const url = location.href || '';
     const body = document.body;
@@ -1120,9 +1123,14 @@ pub(crate) const EXTRACT_CONTENT_JS: &str = r#"(() => {
     walk(root);
 
     let content = lines.join('\n').replace(/\n{3,}/g, '\n\n').trim();
-    if (content.length > 50000) content = content.substring(0, 50000) + '\n... (truncated)';
+    if (content.length > __MAX_CONTENT_CHARS__) content = content.substring(0, __MAX_CONTENT_CHARS__) + '\n... (truncated)';
     return JSON.stringify({title, url, content});
 })()"#;
+
+/// Page-extraction script with the cap substituted in, built once.
+pub(crate) static EXTRACT_CONTENT_JS: LazyLock<String> = LazyLock::new(|| {
+    EXTRACT_CONTENT_JS_TEMPLATE.replace("__MAX_CONTENT_CHARS__", &MAX_CONTENT_CHARS.to_string())
+});
 
 // ── Tests ──────────────────────────────────────────────────────────────────
 
@@ -1283,6 +1291,26 @@ mod tests {
         assert!(
             !paths.is_empty(),
             "Should have platform-specific candidates"
+        );
+    }
+
+    /// The cap the model sees must be the one the constant declares.
+    ///
+    /// The script used to hard-code `50000` while `MAX_CONTENT_CHARS` sat unused, so changing the constant changed nothing.
+    #[test]
+    fn test_extract_content_js_uses_max_content_chars() {
+        assert!(
+            !EXTRACT_CONTENT_JS.contains("__MAX_CONTENT_CHARS__"),
+            "placeholder must be substituted before the script is sent"
+        );
+        assert!(
+            EXTRACT_CONTENT_JS.contains(&MAX_CONTENT_CHARS.to_string()),
+            "the script must truncate at MAX_CONTENT_CHARS"
+        );
+        // Guards against the placeholder being dropped from the template entirely, which would leave the two silently independent again.
+        assert!(
+            EXTRACT_CONTENT_JS_TEMPLATE.contains("__MAX_CONTENT_CHARS__"),
+            "the template must keep the placeholder"
         );
     }
 


CHANGELOG.md
```diff
@@ -69,6 +69,10 @@ and this project uses [Calendar Versioning](https://calver.org/) (YYYY.M.DD).
 
 ### Fixed
 
+- Substitute `MAX_CONTENT_CHARS` into the page-extraction script instead of leaving the cap hard-coded in the JavaScript, so the constant that documents the limit is the one that enforces it.
+  `MAX_CONTENT_CHARS` was declared, marked `#[allow(dead_code)]`, and read by nothing; the script truncated at a literal `50000` written twice inside the JS string.
+  Editing the constant to change how much page text reaches the model therefore did nothing at all, and the `#[allow]` suppressed the one warning that would have said so.
+  The script is now a template with a `__MAX_CONTENT_CHARS__` placeholder substituted once through a `LazyLock`, and a test asserts both that the placeholder is gone from the built script and that it is still present in the template — the second half is what keeps the two from silently drifting apart again (@nevgenov)
```
houko: CLAUDE.md requires a changelog entry to be a new file under `changelog.d/`, not a direct edit to `CHANGELOG.md`:

> Changelog entry = a new file under `changelog.d/`, NOT an edit to `CHANGELOG.md`.
> Appending to the single `## [Unreleased]` section conflicts with every other open PR that does the same, in a file where the conflict carries no information — both sides are correct and the resolution is always "keep both".

This bullet should instead be `changelog.d/fixed/6623-wire-max-content-chars.md` containing the bullet body (no leading `- `, continuation lines indented two spaces).

Separately, even taken as-is this bullet is missing the PR reference: it ends `...drifting apart again (@nevgenov)` with no `(#6623)`. The required closing shape is `(#PR) (@your-github-login)` — see `changelog.d/README.md`.

---

houko (commented):
Reviewed the actual change (`crates/librefang-runtime/src/browser.rs` + `CHANGELOG.md`, commit 6e31087) against `CLAUDE.md`. The fix itself is sound: `EXTRACT_CONTENT_JS_TEMPLATE` + `LazyLock`-substituted `EXTRACT_CONTENT_JS` correctly wires `MAX_CONTENT_CHARS` into the script, `test_extract_content_js_uses_max_content_chars` exists and genuinely pins the substitution (it asserts the placeholder is gone from the built script, the built script contains the constant's value, *and* the template still contains the placeholder — so the two can't silently drift apart again). No `HashMap`/ordering concerns apply here (single scalar substitution, not a prompt-reaching collection), no config-field or dashboard-data-layer changes in scope.

One concrete finding, left inline: the changelog entry edits `CHANGELOG.md` directly instead of adding a `changelog.d/fixed/` fragment, and the bullet as written is also missing its `(#6623)` reference.

Process note (not blocking correctness, flagging per repo policy): the PR's Testing section states `cargo test --workspace` was run. `CLAUDE.md` forbids the unscoped, workspace-wide `cargo test` locally — only `cargo test -p <crate>` is permitted, since the full run contends with other sessions on the shared `target/` directory. Worth a scoped rerun (`cargo test -p librefang-runtime`) for the record, though it doesn't change the correctness assessment above.

(Noting for context, not requesting action: the PR is currently based on a commit several merges behind `main` and GitHub shows it as `has-conflicts`/`dirty` — will need a rebase before it's mergeable, independent of the above.)

---

CHANGELOG.md
houko, line 1: Per CLAUDE.md's Git Conventions ("Changelog entry = a new file under `changelog.d/`, NOT an edit to `CHANGELOG.md`"), this bullet should be a fragment file instead of a direct edit here — appending to the shared `## [Unreleased]` section conflicts with every other open PR doing the same. The canonical target path is literally `changelog.d/fixed/6623-wire-max-content-chars.md` (that's CLAUDE.md's own worked example filename), with the bullet body minus the leading `- `/indentation and ending `(#6623) (@nevgenov)`.

I attempted to apply this fix directly and push it, but this PR's head branch lives on a fork (`nevgenov/librefang`) that I don't have write access to, so I can't land the commit on your branch. Suggested fragment content for `changelog.d/fixed/6623-wire-max-content-chars.md`:

```
Substitute `MAX_CONTENT_CHARS` into the page-extraction script instead of leaving the cap hard-coded in the JavaScript, so the constant that documents the limit is the one that enforces it.
`MAX_CONTENT_CHARS` was declared, marked `#[allow(dead_code)]`, and read by nothing; the script truncated at a literal `50000` written twice inside the JS string.
Editing the constant to change how much page text reaches the model therefore did nothing at all, and the `#[allow]` suppressed the one warning that would have said so.
The script is now a template with a `__MAX_CONTENT_CHARS__` placeholder substituted once through a `LazyLock`, and a test asserts both that the placeholder is gone from the built script and that it is still present in the template — the second half is what keeps the two from silently drifting apart again (#6623) (@nevgenov)
```

...and drop the corresponding hunk from `CHANGELOG.md` in this PR. Separately, note this PR is based on `main` at `1b9518f`, which predates `#6628` (the changelog.d migration) — a rebase onto current `main` is needed anyway before this will apply cleanly.

---

houko (commented):
Found one CLAUDE.md conventions violation (changelog fragment location); flagged inline with the exact fix. No other issues found in the diff (prose wrapping, doc comments, and the added unit test all look correct for this internal, non-route change).

---

nevgenov:
Rebased onto current `main` and moved the entry to `changelog.d/fixed/6623-wire-max-content-chars.md`, using the filename from CLAUDE.md's worked example and ending `(#6623) (@nevgenov)`. `has-conflicts` is gone.

On the scoped-test point: fair, and the checklist now reports `cargo test -p librefang-runtime` (2147 passed) rather than the unscoped run. Supporting evidence for the rule, from this branch — running the workspace suite unscoped filled the disk outright, since it links a separate test binary per crate; `target/` went from ~100 MB to past 60 GB before the run died.

Thanks for confirming the substitution test pins what it claims to. That third assertion was the point of the exercise: without it, deleting the placeholder from the template restores exactly the silent decoupling this fixes, with the other two assertions still green.

crates/librefang-runtime/src/browser.rs
```diff
@@ -1286,6 +1294,27 @@ mod tests {
...
+        assert!(
+            !EXTRACT_CONTENT_JS.contains("__MAX_CONTENT_CHARS__"),
+            "placeholder must be substituted before the script is sent"
+        );
+        assert!(
+            EXTRACT_CONTENT_JS.contains(&MAX_CONTENT_CHARS.to_string()),
+            "the script must truncate at MAX_CONTENT_CHARS"
+        );
+        // Guards against the placeholder being dropped from the template entirely,
+        // which would leave the two silently independent again.
```
houko: CLAUDE.md's prose-wrapping rule: the only legitimate line break inside a comment/doc-comment is at a sentence boundary, not a column limit. This one sentence is split across two `//` lines.

```suggestion
        // Guards against the placeholder being dropped from the template entirely, which would leave the two silently independent again.
```

---

houko (commented):
One small CLAUDE.md prose-wrapping nit found; everything else (changelog fragment, `LazyLock` usage, doc comments, attribution) checks out clean.

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1120,10 +1123,15 @@ pub(crate) const EXTRACT_CONTENT_JS: &str = r#"(() => {
     walk(root);
 
     let content = lines.join('\n').replace(/\n{3,}/g, '\n\n').trim();
-    if (content.length > 50000) content = content.substring(0, 50000) + '\n... (truncated)';
+    if (content.length > __MAX_CONTENT_CHARS__) content = content.substring(0, __MAX_CONTENT_CHARS__) + '\n... (truncated)';
```
houko, line 1126: Not a blocker on this PR (the truncation logic itself is unchanged — you've correctly parameterized a pre-existing literal, not introduced new truncation behavior), but since this is the line being touched, flagging a latent edge case for a maintainer to weigh in on:

`content.length` / `content.substring()` operate on UTF-16 code units. A character outside the BMP (e.g. many emoji, some CJK Extension B+ code points) is stored as a surrogate pair. If `MAX_CONTENT_CHARS` happens to land the cut exactly between the high and low surrogate of such a pair, `substring(0, N)` produces a string containing a lone surrogate. When that gets round-tripped through `JSON.stringify` → CDP → `serde_json::from_str` on the Rust side (see `cmd_read_page`/`page_info`), a lone surrogate in a JSON string escape is something `serde_json` cannot decode into a valid Rust `String` (which must be valid UTF-8/USVs) — it would surface as a deserialize error, silently absorbed by the `.and_then(...).unwrap_or(val)` fallback pattern used at each call site, rather than a panic.

Given `MAX_CONTENT_CHARS` is a fixed constant, this is a rare, hard-to-hit boundary rather than something end users will usually notice, but it's worth a maintainer decision: is it acceptable as-is (silent fallback to raw value), or would it be worth snapping the cut point to a code-point boundary (e.g. checking `content.charCodeAt(N-1)` is not a high surrogate before slicing)? Not asserting this is a bug that needs fixing in this PR — just surfacing it since the substitution here is the natural place to notice it.

---

houko (commented):
Reviewed against CLAUDE.md's checklist for this repo (changelog.d fragment format, commit/PR attribution, prose-wrapping rule, truncation correctness). The fix itself is clean and well-tested: changelog fragment is present and correctly formatted (`changelog.d/fixed/6623-wire-max-content-chars.md`, one sentence per line, ends with `(#6623) (@nevgenov)`), no direct `CHANGELOG.md` edit, no AI/Claude attribution anywhere, and the new doc comments don't hard-wrap mid-sentence. One non-blocking observation left inline about a pre-existing UTF-16 surrogate-pair edge case in the truncation line this PR touches — flagged for maintainer judgment, not asserted as a bug introduced by this change.

---

houko:
Reviewed against `origin/main` at `3680db29`. This is correct and I would merge it as-is.

The dead-code finding checks out: `MAX_CONTENT_CHARS` was `#[allow(dead_code)]` and read by nothing, while the JS truncated at a literal `50000` written twice. The `#[allow]` is the part worth calling out — it is what kept the compiler from reporting a constant that documented a limit it did not enforce.

Three things I verified rather than took on faith:

**The `pub(crate)` type change is contained.** `EXTRACT_CONTENT_JS` goes from `&'static str` to `LazyLock<String>`, which is a breaking change for any consumer. `git grep 'EXTRACT_CONTENT_JS' -- crates/` returns hits only inside `browser.rs`, and both call sites are updated to `&EXTRACT_CONTENT_JS`. `run_js(&self, expression: &str)` takes `&str`, so the deref coercion from `&LazyLock<String>` resolves. Nothing dangling.

**The third assertion is the one that earns its place.** `EXTRACT_CONTENT_JS_TEMPLATE.contains("__MAX_CONTENT_CHARS__")` is what stops someone from "fixing" a future failure of the first assertion by deleting the placeholder from the template, which would restore exactly the silent independence this PR removes. The first two assertions alone would still pass in that world, because `MAX_CONTENT_CHARS.to_string()` is `"50000"` and a hard-coded `50000` in the template contains it. That is a real mutation and the test catches it.

**`LazyLock` is the right tool here** — the substitution is a pure function of a compile-time constant, so it could arguably be a `const fn` or a build-time concat, but neither is expressible for a runtime `String` replace and the once-per-process cost is a single small allocation.

One observation, not a change request: this makes the constant load-bearing, which is the precondition for making the cap operator-tunable. I have proposed exactly that on #6624 — `BrowserConfig.max_content_chars` defaulting to the current 50_000 — and it should land on top of this rather than compete with it, since it would replace the constant as the substitution source and turn the `LazyLock` into a per-session build. Sequencing it that way is why this should merge first.

CI is green across all 8 checks. Nice catch on the `#[allow(dead_code)]` — that attribute suppressing the one warning that would have surfaced the bug is the generalizable lesson.

nevgenov:
Rewrapped, and the commit body with it. Thanks for the close read on the type change and the third assertion.

On the UTF-16 surrogate edge case: agreed it's real and agreed it's out of scope here — the truncation semantics are byte-for-byte what shipped, this PR only moves where the number comes from. Worth recording the shape of it, though, because your trace of the failure mode is the interesting part: it doesn't panic, it deserialises to an error that the `.and_then(...).unwrap_or(val)` fallback swallows, so the operator sees a page that quietly read back as something other than its content rather than a reported failure.

That makes it a candidate for the follow-up you sketched on #6624 rather than a separate issue: the moment `max_content_chars` becomes operator-tunable, the cut lands wherever the operator puts it instead of at one fixed offset, so a boundary that is currently a fixed rare coincidence becomes reachable by configuration. Cutting on a code-point boundary — stepping back one unit when `content.charCodeAt(N-1)` is a high surrogate — is a couple of lines in the same script, and it belongs with the change that makes the offset variable. I'll fold it in there unless you'd rather see it standalone.

changelog.d/fixed/6623-wire-max-content-chars.md
houko, line 1: The code and the fragment content itself are both correct: `MAX_CONTENT_CHARS` wiring checks out (no off-by-one, same `50_000` value, both `run_js` call sites updated, `cargo check -p librefang-runtime --lib` is clean), and this file matches CLAUDE.md's worked example filename and format, ends `(#6623) (@nevgenov)` on its own line, and passes `scripts/check-changelog-attribution.py --all-unreleased`.

One thing worth flagging before merge: on the branch state fetched via `git fetch origin fix/wire-max-content-chars`, this file was added in a second commit (`chore(changelog): move #6623 entry into a changelog.d fragment`) whose git author is `Claude <noreply@anthropic.com>`, not a human contributor. That's AI attribution baked into commit metadata, which this repo's CLAUDE.md explicitly disallows ("No AI / Claude attribution in commit messages, PR bodies, or comments"). I can't fix this myself — correcting an author on an already-pushed commit needs a rebase + force-push, which I'm not authorized to do unilaterally. Flagging so a maintainer can decide: squash-merge would absorb it under the PR author regardless, or the commit can be amended with `--reset-author` before merge if a clean git history matters here.

---

houko (commented):
Reviewed the MAX_CONTENT_CHARS wiring and the changelog.d fragment against CLAUDE.md. The code change and the fragment are both correct — no fixes needed there. One finding left as a line comment: a commit already on the fetched branch has AI attribution in its git author field, which the repo's no-AI-attribution rule disallows; I can't rewrite pushed history myself, so surfacing it for a maintainer decision rather than force-pushing a fix.

---

crates/librefang-runtime/src/browser.rs
houko, line 1126: Good fix for the dead-constant problem, but this line still truncates on `content.length` / `substring`, which count UTF-16 code units, not Unicode scalar values or bytes. If the cut lands between the two halves of a surrogate pair (any character outside the BMP — most emoji, some CJK Extension B ideographs, mathematical alphanumeric symbols), `substring(0, __MAX_CONTENT_CHARS__)` splits the pair and leaves a lone surrogate in `content`.

I traced what that does downstream: `JSON.stringify` emits the lone surrogate as an escaped `\uD83D`-style sequence (verified in Node — a 49999-`a` string plus one trailing emoji, truncated at 50000, round-trips fine inside V8's own `JSON.parse`, but the emitted JSON text contains that unpaired escape). On the Rust side, `cmd_read_page` (`val.as_str().and_then(|s| serde_json::from_str(s).ok())` around line 668) and `page_info` (same pattern around line 775) both feed that text through `serde_json::from_str`, which does strict surrogate-pair validation and errors on a lone leading surrogate. The `.ok()` swallows that error and the `.unwrap_or(val)` / `.unwrap_or(content_val)` fallback silently substitutes the *raw unparsed string* for the expected `{title, url, content}` object — so `content_obj["content"].as_str()` in `page_info` comes back `None` and the caller gets empty content with no error surfaced, instead of the truncated page text.

This isn't new in this PR (the same `substring` call existed at the hardcoded `50000` before), but the PR is the one touching this exact line to parameterize it, and now that `MAX_CONTENT_CHARS` is a constant someone can actually tune, it's worth cutting at a safe boundary while it's being edited anyway. A minimal fix that stays inside the same template:

```js
let cut = __MAX_CONTENT_CHARS__;
if (content.length > cut) {
    if (content.charCodeAt(cut - 1) >= 0xD800 && content.charCodeAt(cut - 1) <= 0xDBFF) cut -= 1; // don't split a surrogate pair
    content = content.substring(0, cut) + '\n... (truncated)';
}
```

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1286,6 +1294,26 @@ mod tests {
...
+    ///
+    /// The script used to hard-code `50000` while `MAX_CONTENT_CHARS` sat unused, so changing the constant changed nothing.
+    #[test]
+    fn test_extract_content_js_uses_max_content_chars() {
+        assert!(
+            !EXTRACT_CONTENT_JS.contains("__MAX_CONTENT_CHARS__"),
+            "placeholder must be substituted before the script is sent"
+        );
+        assert!(
+            EXTRACT_CONTENT_JS.contains(&MAX_CONTENT_CHARS.to_string()),
```
houko, line 1307: This test (and the two assertions around it) only checks that `MAX_CONTENT_CHARS.to_string()` appears as a substring of the built script and that the placeholder is gone — it doesn't execute the JS or otherwise exercise the truncation logic itself. It would pass unchanged if, say, the comparison were `>=` instead of `>`, or the cut length were off by one, or `substring` were swapped for something else entirely, as long as the literal number `50000` still shows up somewhere in the text. It does correctly guard the specific regression this PR fixes (the placeholder never being substituted at all), so it's not asking for a browser-driven integration test — just flagging that "the cap the model sees must be the one the constant declares" (per the doc comment) isn't actually verified end-to-end by this test; only string presence is.

---

houko (commented):
Cross-checked that MAX_CONTENT_CHARS is now genuinely wired through EXTRACT_CONTENT_JS (confirmed via LazyLock substitution and the crate's own test), left two follow-up notes inline: a pre-existing UTF-16 surrogate-splitting risk on the truncation boundary this PR happens to touch, and a note on what the new test does/doesn't verify.

---

crates/librefang-runtime/src/browser.rs
houko, line 1: Not yet raised on this thread: both commit bodies on this branch hard-wrap at a ~72-76 column limit rather than breaking only at sentence boundaries.

CLAUDE.md's prose-wrapping section is explicit that this applies to commit message bodies (subject lines keep git's own ~72-char truncation convention, but bodies don't get a column wrap):

`6e310879` ("fix(runtime): wire MAX_CONTENT_CHARS...") splits sentences mid-clause, e.g. "The constant was declared, marked #[allow(dead_code)], and read by\nnothing." and "twice inside the JS string, so editing MAX_CONTENT_CHARS to change how\nmuch page text reaches the model did nothing, ...".

`d4d720c0` ("chore(changelog): move #6623 entry...") does the same: "Editing CHANGELOG.md's Unreleased section directly conflicts with every\nother open PR doing the same; this repo's convention is one fragment\nfile per entry under changelog.d/, ...".

Neither commit is fixable in place without rewriting an already-pushed commit (amend + force-push), which isn't something I'm authorized to do unilaterally — flagging for a maintainer to fold in if/when this branch gets rebased or squash-merged (the squash-merge commit message is the natural point to apply sentence-boundary wrapping instead of carrying the column-wrapped bodies forward).

---

PR #6628, opened 2026-07-29
chore(changelog): assemble release notes from per-PR fragments instead of a shared [Unreleased] section

houko:
## Problem

Every PR that writes a changelog entry appends a bullet to the same `## [Unreleased]` section of `CHANGELOG.md`, so any two concurrent PRs conflict there — and the conflict carries no information, because both sides are correct and the resolution is always "keep both".

The traffic makes it routine rather than occasional.
Twelve commits landed on `main` on 2026-07-28; six of them edited `CHANGELOG.md`, three within 44 minutes of each other (#6604 at 14:49, #6603 at 15:10, #6602 at 15:33).
Every one of those merges invalidated the file for every open branch cut from an earlier tip.

It lands hardest on fork PRs, where the maintainer cannot rebase the contributor's branch.
`git merge-tree --write-tree --name-only origin/main <pr-head>` against the three currently-conflicting fork PRs — #6617, #6619, #6623, all from @nevgenov — reports `CHANGELOG.md` as the **only** conflicting path in each.
Their actual code (`crates/librefang-runtime/src/browser.rs`, `crates/librefang-types/src/config/types.rs`) auto-merges cleanly.
Three contributions are blocked on a file that has nothing to do with what they change.

## Why not "stop writing changelog entries"

**Correcting the record, because a later maintainer should not inherit a wrong belief.**
An earlier revision of this PR's rationale — and the brief that produced it — claimed that abandoning changelog entries would empty every release body.
That was wrong.
Release bodies are built from merged **PR titles** by `generate_classified_output` in `xtask/src/changelog.rs`, and until the second half of this PR they never carried a single hand-written `[Unreleased]` bullet.
Dropping changelog entries would not have emptied a release body; it would have left the bodies exactly as they already were.

The real reasons the file cannot be abandoned:

- **Release automation consumes it.** `.github/workflows/release.yml:327` slices `CHANGES=$(awk '/^## \['"$VERSION"'\]/{found=1; next} found && /^## \[/{exit} found{print}' CHANGELOG.md)` into the GitHub release body; `release-notify.yml:186` reuses the same slice `| head -20` for the release Discussion; `scripts/changelog-to-article.sh` slices it for the dev.to article.
- **It is the project's history.** 3,043 of the file's 3,252 bullets carry a `(@login)` attribution, under the same block rule `bullet_block_has_attribution` in `scripts/check-changelog-attribution.py` uses. That is 4,530 lines of credited provenance, not a scratch file.
- **Three gates enforce it.** The `scripts/hooks/pre-commit` hook plus the `CHANGELOG Attribution` CI jobs fail a PR that adds an unattributed bullet, so the format is a contract rather than a convention.

This PR puts fragments exactly where hand-written bullets already go, and nowhere else.
That is what makes append-not-replace and backward compatibility coherent, and it keeps the file shape the `awk` extractors depend on byte-identical.

## Layout

```
changelog.d/
  README.md                     format, sections, worked example
  added/     fixed/     changed/     security/     documentation/
```

One directory per `### ` heading of `## [Unreleased]`, each with a `.gitkeep` so the empty directory stays tracked.

## Fragment format

One file, one bullet, holding the bullet body **without** the leading `- `, one sentence per line, continuation lines indented two spaces, ending `(#PR) (@your-github-login)`.

`changelog.d/fixed/6623-wire-max-content-chars.md`:

```markdown
Honour `max_content_chars` on the streaming path, which read the compiled-in default and ignored the per-agent override entirely.
The value was resolved once at kernel boot and captured into the driver, so an `agent.toml` edit took effect only after a restart.
It is now resolved per turn from the manifest, falling back to the kernel config and then the compiled default (#6623) (@houko)
```

Assembles under `### Fixed` in `## [Unreleased]` as:

```markdown
- Honour `max_content_chars` on the streaming path, which read the compiled-in default and ignored the per-agent override entirely.
  The value was resolved once at kernel boot and captured into the driver, so an `agent.toml` edit took effect only after a restart.
  It is now resolved per turn from the manifest, falling back to the kernel config and then the compiled default (#6623) (@houko)
```

Two PRs never touch the same file, so the conflict is structurally impossible rather than merely rarer.

## Assembly

`cargo xtask collect-fragments` folds every fragment into `## [Unreleased]`, deletes the files it consumed, and prints the count.
`cargo xtask release` runs it immediately **before** `changelog::run` cuts the dated `## [VERSION]` section, so fragments are ordinary `[Unreleased]` bullets by the time a release heading exists and no workflow has to learn about them.

- Bullets are sorted by file name within a section, and sections render in the order the repo already uses (Added, Fixed, Changed, Security, Documentation) — an unsorted `read_dir` would make the assembled file depend on the order fragments happened to be created in (#3298).
- A fragment in a section directory assembly has no heading for is **left on disk** and warned about, never deleted.
  The gate that rejects the typo is `scripts/check-changelog-attribution.py`, which runs per-PR and on every push to `main`, long before a tag is cut.
- A fragment that cannot be deleted after the fold fails the command **naming the survivors**, because `CHANGELOG.md` is already written by then and a silent partial delete would fold the same entry in twice on the next run.
- The release commit stages `changelog.d` as a directory, which is what gets those deletions into the commit — a per-file entry is a no-op, since the staging loop skips a path that no longer exists.
  Without this the release would ship the folded bullets while leaving the fragments tracked on `main`, and the next release would fold them in again.

## Second half: curated `[Unreleased]` prose now reaches the release body

Fixing the conflict exposed that the section being fought over was **write-only**.

`cargo xtask changelog` built the entire `## [VERSION]` section from PR metadata — `git log` for the numbers, `gh pr view` for the titles — and inserted it *below* `## [Unreleased]` without reading or touching that section.
The `awk` extractors slice only the dated section.
So 160 hand-written bullets spanning #983 → #6609 sat on `main` and reached nothing: the part that explains *why* a change was made rather than restating its title, and the part three gates enforce attribution on, was consumed by no automation at all.
Making fragments the way to write an entry while that hole stayed open would have industrialised writing prose into a dead end.

**What now happens.** The release cut lifts the `[Unreleased]` body out and composes the dated section as stats, breaking changes, highlights, the curated prose, then the generated entries.
The curated text is a verbatim slice, so `### ` subsections and the order a human chose survive untouched.
Generated entries fill only the gaps: a PR whose number appears in the trailing `(#N)` group of a curated bullet gets no title line, so each PR is described exactly once — in the contributor's words where they wrote them, from the PR title otherwise.
The reference is read from the bullet's **last non-empty line** only, which keeps a mid-bullet cross-reference (`the latter via #6441`) from being mistaken for the PR the bullet documents and suppressing a real entry, and it accepts the `(#6594, #6595)` form one existing bullet uses for two PRs.

**`[Unreleased]` survives as an empty heading.** In-flight PRs append under it and `fold_fragments` targets it, so both break the moment it disappears.
The drain leaves the heading plus one blank line — not cosmetic, since `fold_fragments` inspects the last line before inserting a new `### ` subsection to decide whether it still owes a separator.

**Two no-loss guards, both aborting rather than writing.**
Draining is destructive and a lossy CHANGELOG surfaces only once a tag exists, at which point the words are in no commit at all.

- `verify_no_curated_bullet_lost` is the first statement after composing the body and ahead of every `fs::write`. It compares the composed body against the section as it stood on disk and names the missing bullets. Its parse is deliberately a *different* one from the drain's: it ends the section at the next **dated** heading while the drain stops at the next `## [` of any kind, so the two disagree exactly where a drain would truncate — on a bullet continuation line starting in column 0 with a release heading — rather than restating what the drain happened to take.
- A **second** guard covers regenerating a `## [VERSION]` section that already exists, which the first one is structurally blind to. Three steps after `changelog::run` hard-fail (`codegen --openapi`, `codegen-sdks.py`, `schema-check gen`) and the preflight then refuses a dirty tree with "Commit or stash changes first", so committing the drained file and re-running is the natural recovery — and on that re-run `[Unreleased]` is already empty, so the first guard derives its expectation from nothing and passes on a zero-length slice while the replace-in-place branch rewrites the section from PR titles alone. Regenerating over attributed prose the new body does not contain now aborts, naming the bullets and both ways out (move the prose back under `## [Unreleased]`, or delete the dated section and cut it again).

Both guards compare **whole bullets**, not `- ` marker lines, and attribution is detected anywhere in a bullet's block.
The sentence-per-line prose rule makes multi-line bullets the norm — 67 of the 160 carry their `(@login)` on a continuation line, none on the marker — so a marker-line rule would have called a bullet preserved while every sentence after its first was dropped, and would have named 93 bullets while reporting a count that read as if the other 67 were safe.
Both parsers route through the existing `split_bullets`, whose doc comment already pointed at the Python validator that spells out why per-line is wrong.

**Fail-safe on a bullet with no `(#N)`.**
There is no way to know which generated entry such a bullet replaces, so that PR keeps its generated line and the bullet is named in a warning — a duplicated entry is cosmetic, a silently dropped PR is not.
The fallback is **per bullet**: three bullets out of 160 carry no reference, and they no longer disarm the other 157.

**What the cut now reports about the section's shape.**
Nothing prunes `[Unreleased]`, so it accumulates across releases.
Of the 160 bullets only **13** reference a PR in `v2026.7.27..HEAD`, **144** reference only older PRs, and **3** reference none; 52 of the 78 distinct referenced PRs already appear in a published dated section.
The cut prints that breakdown per bullet against the in-range PR set and warns on the already-shipped count, because a dated section that re-announces shipped work contradicts the `Full diff` compare link published beside it.
This PR does **not** prune the backlog — see the questions below.

Two shape notes, so neither reads as a bug:

- One release body can carry `### Added`, `### Fixed`, and `### Changed` **twice** — the curated block first, the generated block second — because merging them would mean re-grouping bullets whose order a human chose. The curated `### Documentation` sits above the fold while the generated one is inside the `<details>` block, and the curated `### Security` has no generated counterpart. The `awk` slice is indifferent to all of it.
- **Size, as a measurement and not a cap claim.** The `[Unreleased]` body is 133,657 characters against 3,598 for `## [2026.7.27]`, a 37x jump in what ships. It flows verbatim into the GitHub release body, the release PR body, and `articles/release-<v>.md`. I did not verify a hard API limit, and `## [2026.5.2]` measures 208,212 characters today — so a historical section already exceeded anything I could name. Worth the maintainer's own check before the first cut.

## Validation

`scripts/check-changelog-attribution.py` now holds a fragment to the same standard as an `[Unreleased]` bullet in all four of its modes (diff, `--staged`, `--all-unreleased`, `--full`), reusing the single `bullet_block_has_attribution` predicate so there is one copy of the `(@user)` rule.
It additionally rejects a fragment outside a recognised section directory, because assembly has no heading to render it under and the entry would vanish silently.
Diff and staged modes read the post-image and filter `--diff-filter=ACMR`, so the release commit that *deletes* fragments is not failed for it.

- `scripts/hooks/pre-commit`: the attribution check was gated on `CHANGELOG.md` being staged, so a fragment-only commit — the normal case from now on — skipped it entirely.
  Now gated on `^(CHANGELOG\.md$|changelog\.d/)`; the duplicate-`[Unreleased]` guard stays CHANGELOG-only.
- `.github/workflows/ci.yml`: `scripts/tests/test_check_changelog_attribution.py` is wired into the `CHANGELOG Attribution (full [Unreleased])` job — it existed and nothing ran it.
  No check names changed, so no branch-protection churn.

The release PR stays green on the attribution job after a drain: `added_lines_in_unreleased` bounds added lines to the post-image `[Unreleased]` range, and the moved bullets land below the `## [VERSION]` heading, while `--all-unreleased` sees a bare heading with zero bullets.

## Backward compatibility

Editing `## [Unreleased]` directly still works and is still documented as supported.
Assembly **appends** to a `### ` subsection that already exists and creates a missing one in the canonical position; it never rewrites what is there.

The 160 existing `[Unreleased]` bullets are deliberately **not** migrated.
Converting them would rewrite the exact lines the open PRs are conflicting on, making every one of those PRs strictly worse.

An empty `[Unreleased]` section still cuts a byte-identical release, which is the path every release before this took.

## Scope limit

This eliminates the conflict for PRs opened after it lands.
It does nothing for the PRs already open:

- #6621 and #6622 (same-repo) each edit `CHANGELOG.md` and are mergeable today; whichever lands first may invalidate the other.
- #6617, #6619 and #6623 are fork PRs already conflicting on `CHANGELOG.md` and need a one-time manual resolution — their author rebasing, or the maintainer using the web conflict editor.
  A maintainer cannot rebase a fork branch.

## Verification

Only `xtask` was compiled, and that is the honest extent of the coverage.
The diff touches `xtask`, one Python script, `ci.yml`, Markdown, and the `changelog.d/` tree — no library crate, no `Cargo.toml`, no dependency — so a workspace-wide command would cost many minutes and cover nothing this change can break.
Nothing here was exercised against a real `cargo xtask release` run: that needs `gh`, a tag, and a live LLM for the highlights block, so the drain is verified by unit and file-level tests plus a full cut against a copy of this repo's real `CHANGELOG.md`, not by a rehearsed release.

| Command | Result |
| --- | --- |
| `cargo check -p xtask` | exit 0 |
| `cargo clippy -p xtask --all-targets -- -D warnings` | exit 0, zero warnings |
| `cargo test -p xtask` | 106 passed, 0 failed |
| `cargo fmt -p xtask -- --check` | exit 0 |
| `python3 scripts/check-changelog-attribution.py --all-unreleased` | exit 0 |
| `python3 scripts/tests/test_check_changelog_attribution.py` | OK |

Tests for the fragment mechanism:

- `changelog::tests::fragment_assembly_is_deterministic_across_creation_order` — two trees, fragments created in opposite orders, byte-equal output, and `first < second < third` pinned separately so it does not pass on already-sorted names.
- `changelog::tests::fold_appends_to_existing_subsection_and_creates_missing_one_in_canonical_order`
- `changelog::tests::release_cut_stays_sliceable_by_the_awk_extractor` — folds, cuts a dated section, then asserts exact line equality for each curated and generated bullet plus a terminating `## [`, i.e. the shape the extractor depends on. Its earlier assertions asserted the defect (`!slice.contains("Existing fixed bullet")`) and were rewritten.
- `changelog::tests::folds_into_the_repos_own_changelog` — folds into a copy of the real 4,530-line file and asserts the `## [` and `### ` heading counts are unchanged.
- `changelog::tests::an_undeletable_fragment_is_reported_instead_of_silently_left`
- `changelog::tests::fragments_in_an_unrecognised_section_are_left_in_place`
- `changelog::tests::multiline_fragment_keeps_its_two_space_continuation_indent`, `render_fragment_bullet_indents_an_unindented_continuation`, `render_fragment_bullet_strips_a_marker_the_author_wrote_anyway`
- `changelog::tests::fragment_sections_match_the_python_validator` — compares the Rust section list against the `frozenset` literal in the validator, a cross-language contract previously enforced only by a comment.
- `release::tests::release_staging_carries_the_fragment_deletions` — a scratch repository, a real commit, a real fold, then an assertion that the index carries both the `CHANGELOG.md` modification and the fragment's deletion.

Tests for the drain:

- `curated_unreleased_prose_reaches_the_dated_section`
- `draining_leaves_unreleased_a_bare_heading_fragments_still_fold_into`
- `a_curated_pr_reference_suppresses_only_its_own_generated_entry`
- `curated_pr_reference_is_the_last_group_on_the_last_line`
- `a_curated_bullet_with_no_pr_reference_fails_open_only_for_itself`
- `an_empty_unreleased_section_cuts_a_byte_identical_release`
- `a_curated_bullet_the_drain_cannot_reach_aborts_the_write`
- `curated_multiline_bullets_keep_their_continuation_indent`
- `regenerating_a_section_aborts_rather_than_dropping_its_curated_prose`
- `attribution_on_a_continuation_line_still_marks_a_bullet_as_prose`
- `drains_the_repos_own_unreleased_section_without_tripping_the_guard` — cuts a full release from a copy of this repo's real `CHANGELOG.md` and asserts all 160 curated bullets come out of the `awk` slice as whole verbatim blocks, with a separate assertion that some of them are multi-line so the block comparison is not silently testing marker lines.

Mutation probes, each reverted afterwards with the suite green:

| Mutation | Result |
| --- | --- |
| Delete the `verify_no_curated_bullet_lost(...)?` call site in `write_changelog` | **FAILS** `a_curated_bullet_the_drain_cannot_reach_aborts_the_write` at `expect_err` — the write succeeded |
| Drop the curated body's `push_str("\n\n")` terminator | **FAILS 2** — the curated bullet is glued onto the generated `### Fixed` and is no longer a line of the release body |
| Omit `curated` from `GeneratedSections::body` | **FAILS 6**, the first being the guard firing: `refusing to write CHANGELOG.md: 2 hand-written [Unreleased] bullet(s) would not reach the [2026.2.2] section` |
| Revert attribution detection to the `- ` marker line only | **FAILS 1**, exactly `attribution_on_a_continuation_line_still_marks_a_bullet_as_prose` |
| Disable the re-release abort (`if false && !dropped.is_empty()`) | **FAILS 1**, exactly `regenerating_a_section_aborts_rather_than_dropping_its_curated_prose` |
| Restore all-or-nothing PR references (clear the set when any bullet is unreferenced) | **FAILS 1**, exactly `a_curated_bullet_with_no_pr_reference_fails_open_only_for_itself` |
| Drop `changelog.d` from the release staging set | **FAILS** `release_staging_carries_the_fragment_deletions` — index held `M CHANGELOG.md` and no `D changelog.d/fixed/6623-probe.md` |
| Add a section to the validator's `frozenset` only | **FAILS** `fragment_sections_match_the_python_validator`, naming both lists |
| Neuter the `- ` case in `strip_list_marker` | **FAILS** `render_fragment_bullet_strips_a_marker…` with `- - Fix the thing.` |
| Remove `paths.sort_by` in `read_section_fragments` | determinism test **FAILS 3/3** — order came out `Second, Third, First` |
| `let end = lines.len()` (drop the `[Unreleased]` boundary) | **FAILS** — the bullet landed inside `## [2026.1.1]` |
| `## [{}] - {}` → `## {} - {}` in the release heading | **FAILS** — the `awk` slice came back empty |
| `remove_file(path)?` instead of collecting survivors | **FAILS** with a bare `Permission denied (os error 13)` instead of the named error |

An honest negative: the weaker form of the terminator mutation (`push('\n')` instead of `push_str("\n\n")`) failed **nothing**.
Each line stays terminated and CommonMark still parses an ATX heading directly after a list item, so a single-newline terminator is cosmetically different but not structurally broken.
No test guards the blank line specifically, and I judged that not worth one.

Validator invocations, all against this branch's tree:

| Invocation | Result |
| --- | --- |
| `check-changelog-attribution.py` (bare, diff `e8728b18..HEAD`) | **0** |
| `--all-unreleased` | **0** |
| `--full` | **1**, 201 violations, **0** of them under `changelog.d/`; `origin/main`'s copy of the script on the same tree also reports 201, so the `Violation` refactor adds no false positives |
| `--staged`, code change staged, no fragment | **0** |
| `--staged`, valid fragment staged | **0** |
| `--staged`, fragment missing `(@user)` | **1** — `changelog.d/fixed/9999-no-attribution.md:1: missing (@user) attribution: This fragment forgot its attribution entirely (#9999)` |
| `--staged`, fragment at `changelog.d/fix/9999-typo-section.md` | **1** — `'fix' is not a recognised changelog.d section (expected one of: added, changed, documentation, fixed, security); assembly would silently drop this entry` |
| `--staged`, fragment carrying a leading `- ` marker | **0** — the validator accepts it (it is otherwise valid), which is why the renderer strips it |

`--staged` was also confirmed to still fail an unattributed `CHANGELOG.md` bullet in the old format, `CHANGELOG.md:<line>`.

This PR's own changelog entries are fragments, `changelog.d/added/6628-changelog-fragments.md` and `changelog.d/fixed/6628-unreleased-reaches-the-release-body.md`, dogfooding the mechanism.

## Out-of-scope, as questions

- A fragment whose *second* line starts with `- ` at column 0 currently renders as a nested bullet.
  Should a multi-bullet fragment be rejected outright instead, or is nesting the right reading?
- Once the open PRs land, should the existing 160 `[Unreleased]` bullets stay as they are and drain at the next release, or is a separate migration PR wanted?
- **The backlog, now that the drain makes it consequential.** 144 of the 160 bullets reference no PR in `v2026.7.27..HEAD`, so the first release after this lands would announce work that shipped in v2026.7.10, v2026.7.21 and v2026.7.27 as new. The tool now prints the breakdown and warns, and this PR deliberately prunes nothing — deleting 144 bullets of other contributors' credited history is a maintainer call, and I cannot verify offline which of them shipped where. Prune before the next cut, or accept a one-time catch-up release body?
- Three `[Unreleased]` bullets carry no `(#N)` anywhere, so their PRs will each appear twice in the next release body — as curated prose and as a generated title line. Worth hand-adding the references, or is a duplicate on three entries acceptable? (This is now a cosmetic three-entry issue rather than the blocking item an earlier draft of this description claimed: the per-bullet fail-safe means those three do not disarm suppression for the other 157.)
- Should the curated and generated `### Added` / `### Fixed` / `### Changed` blocks ever be merged? Doing so means re-grouping bullets whose order a human chose, which is why this PR does not.
- `render_changelog`'s insert path now rewrites the file with LF via `kept.join("\n")`, so a CRLF checkout of `CHANGELOG.md` would be normalised — the replace path already did this, and `.gitattributes` has no `text eol=lf` entry for the file. Releases are cut from `main` by a maintainer, so impact is low. Worth an explicit `.gitattributes` entry?
- The suppression rule fails in the dangerous direction with no guard: if a curated bullet ever *ends* on a parenthesised cross-reference to some other PR, that PR's generated entry is suppressed and it vanishes from the notes entirely. I checked all 160 bullets against the current file: the rule resolves correctly in every case, including all four whose last line carries more than one `(#N)` group — `(#6445)` mid-line resolves to #6468, `(#3054)` mid-line resolves to #6461, a duplicated `(#6499)` resolves to #6499, and one bullet naming `(#6460)`, `(#6459)`, `(#6460)` resolves to #6460, which leaves #6459 a generated line (a duplicate, i.e. the safe direction). The `\(#` anchor also correctly ignores the `(see #N)` and bare `#6441` forms. Should the cut warn when a suppressed number is in range but the same bullet also names a different in-range PR?
- An empty-but-existing `CHANGELOG.md` renders into `""` and loses the `# Changelog` title. Pre-existing, unchanged by this diff, untested.
- `report()` raises `BrokenPipeError` when its stdout is piped to `head` — pre-existing on `main` and unrelated to this diff. Worth silencing?


xtask/src/changelog.rs
```diff
@@ -14,6 +14,29 @@ pub struct ChangelogArgs {
     pub base_tag: Option<String>,
 }
 
+#[derive(Parser, Debug)]
+pub struct CollectFragmentsArgs {}
+
+/// Directory under `changelog.d/` → the `### ` heading its fragments render under.
+///
+/// The order is the canonical order in which a *missing* subsection is created
```
houko, line 23: CLAUDE.md's prose-wrapping rule ("no column limit; break only at sentence boundaries") says it applies universally, explicitly listing "Source code doc-comments (`//!`, `///`, `#`-style docstrings, ...)", and that it covers "new prose and any paragraph an agent is already touching" — pre-existing untouched paragraphs are exempt, new ones are not.

This doc comment is entirely new but is still 80-column hard-wrapped mid-sentence (e.g. "is created\ninside `## [Unreleased]`" splits one sentence across two lines). The same pattern repeats through most of the new `///` comments added to this file (e.g. `strip_list_marker` around line 485, `fold_fragments` around line 585) and carries into the new comments in `.github/workflows/ci.yml` (~387, ~416) and `scripts/hooks/pre-commit` (~83-88).

Worth noting this is inconsistent within the PR itself: the new `changelog.d/README.md`, the `CONTRIBUTING.md`/`CLAUDE.md` prose additions, and the fragment file `changelog.d/added/6628-changelog-fragments.md` all correctly follow the new one-sentence-per-line convention — only the Rust/shell/YAML comments were left in the old style (presumably matching the pre-existing style already in these files, but that pre-existing style isn't what's being added here).

Not blocking — this is a style-only nit with zero functional risk — but flagging since the project's own CLAUDE.md is explicit about it applying to source comments too. Not auto-fixing this myself since a mechanical reflow risks misreading sentence boundaries around inline code spans and version numbers (e.g. `v2026.4.27`) across ~15 comment blocks; a careful human pass seems safer here than a bulk automated one.

---

houko (commented):
Reviewed against CLAUDE.md's changelog-fragment workflow, prose-wrapping, commit, and Rust conventions.

Verification performed on this worktree (native rustc was too old for the lockfile — updated via `rustup update stable` to 1.97.1, no Docker daemon available here):
- `cargo check -p xtask` — clean
- `cargo clippy -p xtask --all-targets -- -D warnings` — clean, zero warnings
- `cargo test -p xtask` — 95 passed, 0 failed (matches the count claimed in the PR description)
- `python3 scripts/tests/test_check_changelog_attribution.py` — OK
- `python3 scripts/check-changelog-attribution.py --all-unreleased` — OK, no violations

The mechanism itself (fragment format, section-directory mapping, deterministic file-name ordering, the fold/stage seam between `collect_fragments_in` and `stage_release_files`, and the Python validator's fragment checks) is thorough and well-tested — I did not find a functional defect worth fixing. Left one style note inline (new Rust/shell/YAML comments in this diff don't follow the repo's one-sentence-per-line prose rule the way the new Markdown content in the same PR does); not auto-fixing it given the scale and risk of a bulk mechanical reflow across ~15 comment blocks.

---

PR #6637, opened 2026-07-29
fix(hands): re-render saved settings into the live agent prompts, and report the whole approval policy on config read

### LINKED ISSUE #6636: Hand settings never reach the agent's system prompt — [[settings]] values are silently ignored (v2026.7.27)
### Description

Configuring a hand through the dashboard Settings tab has no effect on the agent.
The agent's system prompt always contains the `default` values declared in HAND.toml,
regardless of what is saved in the UI.

Reproduced on three independent installations across two operating systems:
- Windows 11, existing installation
- Windows 11, clean installation (full uninstall incl. ~/.librefang, PATH, credential manager)
- Arch Linux

Using the built-in Trading Hand (`id = "trader"`, registry version 1.1.0).

Settings saved in the dashboard:

    trading_mode     = live
    market_focus     = crypto
    strategy_style   = scalping
    scan_schedule    = 15m
    watchlist        = BTC,ETH
    initial_capital  = 100
    approval_mode    = false

Agent's answer when asked to report its configuration:

    trading_mode:    Paper Trading (paper)
    initial_capital: 10,000 USD
    watchlist:       SPY, QQQ, AAPL, MSFT, NVDA, BTC, ETH
    approval_mode:   Enabled

These are exactly the `default` values from registry/hands/trader/HAND.toml.

The cause appears to be structural. In `schema.toml`, `[hand.sections.settings]`
defines only `key`, `label`, `description`, `setting_type` and `default` — there is
no field for a user-selected value. And `[hand.sections.agent.fields.system_prompt]`
is a plain required string with no templating mechanism defined anywhere in the schema.

The `## User Configuration` block that the Trading Hand's prompt instructs the agent
to read is therefore literal text inside `system_prompt`. Nothing ever substitutes
saved settings into it.

### Expected Behavior

Values saved via the dashboard (or `librefang hand set`) should be reflected in the
agent's runtime configuration.

The Trading Hand's own prompt depends on this:
- Phase 0: "Read User Configuration section for trading_mode, market_focus, risk settings, watchlist"
- Phase 1: `"initial_capital": <from settings>`
- Phase 6: branches on trading_mode ("analysis" / "paper" / "live")

As it stands, every [[settings]] block in every hand is decorative. For the Trading
Hand specifically this is safety-relevant: a user who selects "Live Trading" and
disables approval_mode gets neither, with no warning or error.

### Steps to Reproduce

1. Activate the Trading Hand
2. Hands → Trading-Hand → Settings → set Initial Capital to 100 → Save settings
3. Open Chat with `trader-hand · coordinator`, start a new session (Clear Chat)
4. Ask: "Report your current configuration: trading_mode, initial_capital, watchlist, approval_mode"
5. Agent reports 10000 (the HAND.toml default) instead of 100

Environment: LibreFang 2026.7.27 (CLI + Desktop), Windows 11 and Arch Linux,
hand `trader` v1.1.0 from librefang/librefang-registry

### LibreFang Version

2026.7.27

### Operating System

Windows

### Related observations

a) Local edits to ~/.librefang/registry/hands/trader/HAND.toml are overwritten on
   daemon start, even with the file set read-only. No config option disables the
   registry sync — [registry] exposes only cache_ttl_secs, registry_host and
   registry_mirror, and registry_mirror is documented as a GitHub proxy prefix,
   not an alternate source.

b) Editing the system prompt directly (Hands → hand → Agent tab) DOES take effect
   for a new session, but does not survive a hand deactivate/activate cycle.
   This is currently the only way to configure a hand at all.

c) The Agent tab's system prompt textarea does not refresh when a different agent
   is selected from the dropdown — easy to overwrite the wrong agent's prompt.

d) Configuration → Security → Approval: `require_approval` (string[]) is not
   persisted. The per-field save appears to succeed, config.toml is unchanged.

e) Configuration → Security → Approval: `cache_approvals_per_session` displays as
   false while config.toml contains true.

f) Several field descriptions are wrong — exec_policy.mode and reload.mode both
   show "Kernel operating mode — stable, default, or dev".

houko:
Closes #6636.

## The bug

A hand's `[[settings]]` values only reach an LLM as the rendered `## User Configuration` tail on each role's `model.system_prompt`, and that tail is materialized by `activate_hand_with_id`.

`PUT /api/hands/{id}/settings` called `hands().update_config(..)` and `persist_hand_state()` and stopped there. It never touched the live agents, so a save took effect on the **next daemon restart** — `start_background_agents` replays every persisted hand through the activation path, which does re-render from `hand_state.json` — while the running agent kept answering from the HAND.toml defaults, with no error to explain the discrepancy. The reporter hit this on the Trading Hand, whose prompt branches on `trading_mode` and `approval_mode`: selecting Live Trading and disabling approval silently got neither.

The issue text suggests the mechanism is missing entirely — "every `[[settings]]` block in every hand is decorative", and the `## User Configuration` block "is therefore literal text inside `system_prompt`". Neither holds. `resolve_settings` → `apply_settings_block_to_manifest` exists and renders correctly at activation and at boot, and the registry's `trader` v1.1.0 prompt contains **no** `## User Configuration` heading at all — only two prose references to the section (Phase 0's "Read **User Configuration** section for trading_mode, market_focus, risk settings, watchlist" and Phase 6's "Read trading_mode from User Configuration"). I checked the live registry copy against `crates/librefang-runtime/tests/fixtures/registry/hands/trader/HAND.toml` and the two are byte-identical, so there is no duplicate heading for the agent to read the wrong one of, and no risk of the fenced marker truncating an author-written phase. `settings_marker_ignores_prose_references_to_the_section` pins that shape.

What was missing is only the wiring from the save endpoint to the live agents.

The reporter's `initial_capital: 10,000 USD` and `watchlist: SPY, QQQ, AAPL, ...` are consistent with this: `resolve_settings` emits `- Initial Capital: 10000` and `- Watchlist: SPY,QQQ,AAPL,MSFT,NVDA,BTC,ETH` from the schema defaults, and the model reformatted them in its answer.

## Changes

**Kernel — `LibreFangKernel::update_hand_config`** (`kernel/hands_lifecycle.rs`)
- Takes a **delta**, not a full map, and merges it onto the authoritative read under the per-instance lock. A caller that merged first would put the read-modify-write outside the lock, so two concurrent saves would each write their own snapshot and the later would silently revert the earlier — now visibly, because the revert reaches the LLM immediately.
- Rollback is all-or-nothing across both halves. Each role's previous prompt and allowlist are snapshotted before it is rewritten, so a failure part-way through restores the roles already touched as well as the config; otherwise the persisted file says `paper` while a live agent runs on `live`. Every rollback step logs on failure rather than discarding its `Result`.
- A role recorded on the instance but absent from the reloaded definition (a rename plus `POST /api/hands/reload`, which does not respawn) is skipped with a WARN rather than rendered against another role's roster and playbook — matching `clear_hand_agent_runtime_override`.
- Existence is checked **before** the lock is minted, so an unknown UUID no longer inserts an entry into `hand_instance_locks` that nothing removes; the authoritative re-read under the lock still returns `InstanceNotFound` if the instance vanished in the window.
- Roles are driven off `instance.agent_ids` (per-instance), not a `hand:<id>` tag scan. A role with no live agent (paused hand, failed spawn) is skipped — the next activation renders from the config just persisted.
- No SQLite write. Hand agents are skipped by the boot restore loop (`if entry.is_hand { continue; }`) and rebuilt from `hand_state.json` plus HAND.toml, so their persisted agent blob is never read back. `persist_manifest_to_disk` is deliberately not called either: a hand agent's `source_toml_path` is the workspace `hands/<hand>/hand.toml` (a `HandDefinition`), and serializing an `AgentManifest` over it would clobber the mirror.

**Kernel — locking**
- `hand_runtime_override_locks` → `hand_instance_locks`, and `activate_hand_with_id` / `deactivate_hand` now take it: both write the `AgentRegistry` entries a settings save is rewriting. Activation also re-reads the instance under the lock, so it cannot render from a pre-save snapshot and overwrite a save that already returned 200 — #6636 from the other side. Activation acquires it once the instance exists (the id is not known before); deactivate drops the guard before removing the map entry.
- pause/resume stay unlocked and the doc-comment says why: they touch scheduling and instance status, never the prompt.

**Kernel — `rerender_hand_prompt_tails`** (`kernel/manifest_helpers.rs`)
- One renderer, three call sites. It assembles the base prompt plus the settings, reference-knowledge and team tails from pure renderers in a single pass, and activation, the boot TOML-drift loop and the settings save all go through it.
- The three `apply_*` helpers it replaces each located their own tail by searching the whole prompt for a marker, which is only sound while nothing downstream of that marker is author-controlled. It is not: a SKILL.md playbook may legitimately contain a `---` rule followed by a `## Your Team` heading, and once the reference tail was appended the team helper's search found *that* copy and truncated the playbook there — silently dropping everything after it from the prompt. They also could not be composed: run in sequence against a live manifest, each strip discarded the tail the previous call had just written.
- Assembling in one pass never searches appended content, so the class is gone rather than avoided. The helpers are deleted so nothing reaches for them again; their tests move onto the surviving API. `earliest_rendered_tail_idx` (shared with `manifest_for_diff`) still does the one strip that is safe — back to the author-written base, before anything is appended.

**Kernel — env passthrough**
- `resolve_hand_allowed_env` assembles the allowlist (resolved settings' `provider_env` / `env_var` plus `[[requires]]` env-var entries, all filtered through the secret blocklist) and is what `rerender_hand_prompt_tails` returns, so activation, boot and the save cannot drift. Unifying the boot loop closed a second gap: it re-derived the allowlist from the settings alone, dropping every `[[requires]]`-declared name on every restart until a manual re-activation — the same silent disappearance #5137 fixed for the settings half.
- `set_hand_allowed_env` and `AgentRegistry::update_hand_rendered_prompt` both **remove** `metadata["hand_allowed_env"]` when the list is empty, so a settings change that drops the last `provider_env` narrows the passthrough instead of leaving the previous, wider list in place.

**CLI** (`commands/hand.rs`) — both `hand` settings subcommands were broken, which is why the CLI was no help in diagnosing this:
- `hand set` posted `{"config": {…}}` at a handler that deserializes a flat map, so it stored a literal `config` key holding the whole object and never touched the named setting.
- `hand settings` read `body["config"]` from a response whose keys are `hand_id` / `settings` / `current_values`, so it printed "no configurable settings" for every hand, configured or not.
- It now renders `effective_values` — what the resolver will actually use — rather than re-deriving the coercion rules client-side, which is how the first version ended up reporting a stored non-scalar the resolver was ignoring. A stored value that lost to the default is called out explicitly, and `unknown_values` surfaces saved keys the schema does not declare so `hand set <typo> <value>` no longer vanishes without trace. Four new locale strings, all four locales.

**Settings resolution** (`librefang-hands`)
- `effective_setting_value` is the single definition of what a setting resolves to, shared by the prompt renderer and the HTTP layer. It honours a stored value only when it coerces to a scalar **and**, for a `Select` with declared options, names one.
- Scalar coercion alone would have removed a guard the old `as_str()` had by accident: a non-string always fell back to the author-declared default, which for a `Select` matches an option by construction. Without the second condition `{"trading_mode": true}` rendered `- Trading Mode: true (true)` for the LLM and, matching no option, dropped that option's `provider_env` from the allowlist — which the live agent then loses the instant the save returns 200.
- Arrays / objects / null still fall back: they have no meaningful rendering against a string-valued schema.

**HTTP surface** (`routes/skills/hands.rs`)
- Both settings handlers resolve the instance through the new `HandRegistry::active_instance` (backed by the deterministic `active_index`) rather than a `list_instances` scan whose DashMap order varies per process, so a write and the read after it cannot land on different instances of the same hand.
- `GET` reports `effective_values` and `unknown_values` alongside `current_values`.
- Errors: `HandError::InstanceNotFound` (a concurrent deactivate) → 404; persist / registry failures → 500. Previously every error was 400, which told the operator their input was invalid and sent them to retype a valid setting instead of looking at the host.

**Dashboard** (`lib/mutations/hands.ts`)
- `useUpdateHandSettings` invalidates the agent and overview caches, since the save now rewrites every live agent's system prompt — the same reason `useUpdateHandManifest` does ("hands surface as agents").

**Config read surface — issue observation (e)** (`routes/config/manage.rs`)
- The `approval` section declares no explicit `fields` list in `ui_sections_overlay`, so `ConfigPage` renders a control for whatever the derived schema says `ApprovalPolicy` has, while `redacted_config_json` enumerated seven of its fourteen fields by hand. The other seven rendered blank and read back as their JSON zero value.
- `cache_approvals_per_session` is the reported one: it defaults to `true`, so the dashboard showed the box unchecked even against a `config.toml` that said `true`. `trusted_senders`, `channel_rules`, `timeout_fallback`, `routing`, `totp_tools`, and `audit_retention_days` had the same gap and are fixed with it — leaving six siblings broken next to the line being fixed is how this class of bug keeps coming back (#6596, #6605, #6618).
- All seven stay **non-writable**. Only `auto_approve`, `auto_approve_autonomous`, and `totp_grace_period_secs` are in `WRITABLE_EXACT_PATHS`, deliberately, so an Owner-role caller with a leaked API key cannot relax approval policy over HTTP. That is also why `every_writable_config_leaf_is_readable` never caught this — it enforces `writable ⊆ readable` and is blind to a non-writable field by construction.
- `approval_policy_fields_are_all_readable` guards the other direction, using the serialized `ApprovalPolicy` as its oracle rather than restating a list, and comparing values rather than key presence. Verified as a negative control: removing the `cache_approvals_per_session` line makes it fail.
- Nothing added is secret-bearing — operator-chosen user ids, channel names, tool globs, notification recipients; credentials never live in `ApprovalPolicy`.

## Verification

- `cargo check --workspace --lib` and `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- `cargo test -p librefang-kernel --lib` — 1340 passed.
- `cargo test -p librefang-hands --lib` — 103 passed. New: `resolve_settings_rejects_a_select_value_that_matches_no_option` (bool / number / wrong-case string), `resolve_settings_keeps_free_form_select_without_options`, `undeclared_keys_are_reported_sorted`, `resolve_settings_coerces_non_string_scalars`, `resolve_settings_falls_back_for_non_scalar_json`.
- `cargo test -p librefang-api --lib` — 879 passed. New: `approval_policy_fields_are_all_readable`.
- `cargo test -p librefang-api --test hands_routes_integration` — 38 passed. `update_hand_settings_rerenders_live_agent_prompts` now asserts each role's own base prompt and a roster naming the *other* role, and round-trips the env allowlist widen-then-narrow. New: `update_hand_settings_skips_roles_absent_from_the_reloaded_definition`, which drives a role rename through `PUT manifest` + reload.
- `cargo test -p librefang-api --test config_routes_integration` — 37 passed. New: `get_config_exposes_non_writable_approval_fields`.
- Dashboard: `pnpm typecheck` clean, `pnpm lint` 0 errors, `pnpm test --run` 933 passed, `pnpm build` succeeds.

Note for anyone reproducing locally on macOS: `cargo test` at full default parallelism makes both the `librefang-kernel` lib binary and the `hands_routes_integration` binary abort with a silent SIGABRT and zero test failures. That reproduces on `origin/main` without this change (3/6 runs on the API binary, 3/3 on the kernel binary); the kernel suite passes cleanly on repeat runs and the hands suite passes serially. CI is unaffected — nextest runs each test in its own process.

## Not in this PR

The issue lists six further observations; (e) is fixed here, the rest are not. (a) is a request for a config knob to disable registry sync — a new knob in a different subsystem. (b) is by design (activation re-renders from the definition), and the fix here removes the need for the workaround it describes. The remaining four are all in the dashboard / config-route surface rather than hands, and each needs a decision rather than a mechanical fix:

- (c) The Agent tab's system-prompt textarea does not reset when a different agent is selected from the dropdown, so it is easy to overwrite the wrong agent's prompt. Real data-loss bug; SPA domain.
- (d) `approval.require_approval` is deliberately absent from `WRITABLE_EXACT_PATHS` in `routes/config/mod.rs` (only `auto_approve_autonomous`, `auto_approve`, and `totp_grace_period_secs` are writable, to keep an Owner-role API key from relaxing approval policy). The defect is therefore that the dashboard renders an editable control for a read-only field, not that the write is dropped — but whether the fix is "make it writable" or "render it read-only" is a policy call.
- (f) `ConfigPage.tsx` resolves field descriptions with `t(\`config.desc_${fieldKey}\`)` — keyed on the leaf name only — so `exec_policy.mode` and `reload.mode` both pick up `config.desc_mode`, which describes the kernel operating mode. Needs a section-qualified key with a leaf fallback plus the missing locale strings.

Happy to take any of these on in a follow-up; flagging rather than bundling them so this PR stays one issue's worth of change.


crates/librefang-kernel/src/kernel/hands_lifecycle.rs
```diff
@@ -785,17 +760,124 @@ impl LibreFangKernel {
...
+                allowed_env.clone(),
+            ) {
+                // Roll the config back so the operator's next read shows the
+                // values the live agents are actually running on.
+                let _ = self
+                    .skills
+                    .hand_registry
+                    .update_config(instance_id, previous.config);
+                let _ = self.persist_hand_state_result();
+                return Err(KernelError::LibreFang(e));
```
houko: For a multi-agent hand, this loop re-renders and writes each role's live prompt one at a time, but the rollback on failure only reverts the persisted `HandInstance.config` — it does not undo the rendered-prompt writes already applied to earlier roles in this same iteration.
If `update_hand_rendered_prompt` fails for role N (its only failure mode is the agent having vanished from the registry between the `.get()` check above and this call, so a narrow window), roles `1..N-1` are left running on the *new* settings while the persisted config and the API's error response both say the save failed and reverted.
That's the same class of live/persisted divergence this PR is fixing for the single-agent case, just reintroduced across roles.

Two ways to close it that stay inside this function: (1) resolve every role's new prompt first, confirm every entry still exists, and only then apply the writes (turns "vanished mid-loop" from a partial-apply into a clean before/after failure); or (2) on failure, walk back and reapply the pre-change rendered prompt to the roles already written in this call before returning `Err`.
Given the failure mode is a genuine race rather than a validation error, plan (1) reads simpler.

Not blocking if you'd rather track it separately, but worth a decision since it's the same bug shape #6636 targets.

---

houko (commented):
Reviewed the diff against `origin/main` and CLAUDE.md's checklist (changelog.d fragment, deterministic-prompt-ordering, config-field wiring, integration-test coverage, commit hygiene).
No mechanical fixes needed — the changelog fragment, `BTreeMap` usage, and the claimed regression test all check out; see the inline note for the one design question worth a second look.

Verified:
- `changelog.d/fixed/6637-hand-settings-reach-live-prompts.md` exists, correctly formatted, and passes `scripts/check-changelog-attribution.py --staged`.
- `HandInstance.agent_ids` is already `BTreeMap<String, AgentId>`; no new `HashMap`/`HashSet` was introduced on a path that stringifies into an LLM prompt (settings lookups are point `.get()` calls against an already-ordered `settings: &[HandSetting]` iteration, so ordering isn't at risk).
- `update_hand_settings_rerenders_live_agent_prompts` exists in `crates/librefang-api/tests/hands_routes_integration.rs` and covers exactly what's described: multi-agent hand, settings save, live-prompt re-render, tail-ordering (`## Reference Knowledge` / `## Your Team` survive), and the GET round-trip. `cargo test -p librefang-api --test hands_routes_integration update_hand_settings_rerenders_live_agent_prompts` passes.
- `cargo check -p librefang-kernel -p librefang-hands -p librefang-api -p librefang-cli --lib` and `cargo clippy` (same scope, `-D warnings`) are clean.
- `cargo test -p librefang-hands --lib` (100 passed) and `cargo test -p librefang-kernel --lib manifest_helpers` (37 passed) are green, including the new `rerender_*` and `allowed_env_*` cases.
- No config.toml field additions, so the `#[serde(default)]`/`Default` wiring rule doesn't apply here.
- Commit history is clean of AI/Claude/Anthropic attribution.

---

crates/librefang-kernel/src/kernel/hands_lifecycle.rs
```diff
@@ -785,17 +760,124 @@ impl LibreFangKernel {
...
+            }
+        }
```
houko, line 939: Not a CLAUDE.md-named rule, but worth a maintainer look given the doc comment's own claim: if this loop rewrites role A's live prompt successfully and then fails on role B's `update_hand_rendered_prompt`, the rollback here restores `HandInstance.config` (and re-persists `hand_state.json`), but role A's `AgentRegistry` entry is left on the *new* rendered prompt — it isn't reverted. So the invariant the doc comment states ("the persisted file and the live registry cannot disagree") holds for the hand-instance config vs. disk, but not for the already-updated `AgentRegistry` manifests vs. the now-rolled-back config, in this partial-failure interleaving. Likely a narrow window (an agent would have to disappear from the registry mid-loop under the same instance lock), so may not be worth a full snapshot/restore of every role's prior prompt — flagging so the tradeoff is a conscious call rather than an oversight, per the "surface the question, don't decide unilaterally" guidance in CLAUDE.md's PR-hygiene section.

---

houko (commented):
Reviewed against CLAUDE.md's conventions (hot-reload classification, agent.toml-vs-config.toml scoping, deterministic prompt ordering #3298, changelog-fragment format, prose wrapping, injection-site regression tests). The PR is thoroughly compliant: `agent_ids` is already a `BTreeMap` so prompt-tail rendering per role is order-independent; both changelog fragments are correctly placed under `changelog.d/fixed/`, sentence-per-line, ending `(#6637) (@houko)`; new doc comments and the fragments themselves already follow the sentence-boundary wrap rule; `resolve_settings` only does key lookups against the config `HashMap` (never iterates it) so no ordering hazard reaches the prompt; the injection site (`update_hand_settings` handler) has a direct regression test (`update_hand_settings_rerenders_live_agent_prompts`); this PR doesn't touch `KernelConfig`/`config_reload.rs` or introduce any `[agents.*]`-style per-agent config, so those two rules don't apply here. One inline comment left on a narrow partial-failure edge case in the new rollback logic — not a rule violation, just worth a maintainer's conscious call.

---

crates/librefang-kernel/src/kernel/hands_lifecycle.rs
houko: The per-role loop's rollback-on-failure doesn't undo work already done by earlier iterations.

`update_hand_config` documents the invariant "Rolls the config back and re-persists if either step fails, so `hand_state.json` and the live registry cannot disagree." That holds for the `persist_hand_state_result()` failure above, but not here: if `update_hand_rendered_prompt` fails for e.g. the *second* role of a multi-agent hand, the first role's live prompt has already been rewritten to the new settings (that `AgentRegistry` write already landed and cannot be un-sent), while this branch only rolls `instance.config` and `hand_state.json` back to `previous.config`. The result is exactly the split-brain state the doc comment says can't happen: the persisted config (and the next `GET .../settings`) shows the old values, but the first role's running agent is already answering from the new ones.

`update_hand_rendered_prompt` can only fail via `with_entry_mut`'s `AgentNotFound` (the entry was just fetched a few lines above via `self.agents.registry.get(*agent_id)` in the same synchronous scope, so this needs a concurrent removal of that specific agent — e.g. a direct `DELETE /api/agents/{id}` racing this hand-settings save without going through `hand_instance_lock`), so it's a narrow window rather than a mainline path. Given multi-agent hands are an explicit feature this PR adds coverage for, though, it seems worth either (a) snapshotting each role's previous `(system_prompt, hand_allowed_env)` before the loop so a mid-loop failure can restore every already-mutated agent, not just the config, or (b) at least downgrading the doc comment's "cannot disagree" claim to note this residual gap, so a future reader doesn't rely on a guarantee the code doesn't fully provide.

---

houko (commented):
Automated daily CLAUDE.md-driven review of this PR. No mechanical/unambiguous issues found — the deterministic-ordering rule (#3298), the per-agent vs per-hand config-override distinction (#5476), changelog.d fragments, and the "MANDATORY: Integration Testing" rule are all satisfied (both `hand_instance.agent_ids` and `HandDefinition.agents` are already `BTreeMap`, `resolve_settings` iterates the ordered `[[settings]]` slice rather than the instance-config `HashMap`, two well-formed `changelog.d/fixed/` fragments are present, and both the settings-rerender and config-read-gap fixes ship `#[tokio::test]` coverage against `TestServer`). One architecturally-notable gap is left as an inline comment on `hands_lifecycle.rs` rather than fixed directly, since closing it is a design decision (snapshot-based rollback) rather than a mechanical patch.

---

crates/librefang-kernel/src/kernel/hands_lifecycle.rs
houko, line 939: For a multi-role hand, rollback here is config-only, not per-role.

If `update_hand_rendered_prompt` fails for role N after role 1..N-1 already succeeded, the `catch` only resets `previous.config` (and re-persists it) — it doesn't revert the prompt/`hand_allowed_env` metadata already written to roles 1..N-1's live `AgentRegistry` entries. That leaves e.g. `lead` rendered against the *new* config while `worker` and the persisted `hand_state.json` fall back to the *old* one — exactly the drift the doc comment above (`update_hand_config`) says can't happen ("rolling the config back if either step fails so the persisted file and the live registry cannot disagree").

The realistic trigger is narrow — `update_hand_rendered_prompt` only fails on `AgentNotFound` via `with_entry_mut`, i.e. a concurrent deactivate/removal racing the `self.agents.registry.get(*agent_id)` check a few lines up — but for a ≥2-role hand it's a real, currently untested state. `hands_routes_integration.rs::update_hand_settings_rerenders_live_agent_prompts` only exercises the all-succeed path.

Worth either (a) rendering all roles' new prompts first (pure, can't fail) and only then writing all of them, still checking each write but knowing a mid-loop failure means none of the *renders* were bad — doesn't fully close the race but simplifies reasoning — or (b) snapshotting each role's previous `(system_prompt, hand_allowed_env)` before the loop and rolling every already-written role back to its own snapshot on failure, not just the config. Flagging rather than patching since it's a rollback-semantics change in a locking path, not a mechanical fix.

---

crates/librefang-api/tests/hands_routes_integration.rs
```diff
@@ -409,6 +409,168 @@ default = "15"
...
+// ---------------------------------------------------------------------------
+// PUT /api/hands/{hand_id}/settings — saved values reach the live prompts
+// ---------------------------------------------------------------------------
+
+/// Regression test for #6636: saving settings must re-render the `## User Configuration` tail on every live agent of the hand.
+///
+/// Before the fix, the handler wrote the instance config and persisted `hand_state.json` but never touched the agents, so a running agent kept answering from the HAND.toml defaults until the daemon restarted — boot replays hands through `activate_hand_with_id`, which does re-render.
+/// The hand here is multi-agent and ships skill content so the test also pins the tail-ordering hazard: re-rendering the settings block on a prompt that already carries `## Reference Knowledge` and `## Your Team` must not truncate them away.
+#[tokio::test(flavor = "multi_thread")]
+async fn update_hand_settings_rerenders_live_agent_prompts() {
```
houko, line 421: This test is a good regression for the prompt-tail re-render, but it doesn't assert on `hand_allowed_env` — the other thing `update_hand_config` rewrites via `AgentRegistry::update_hand_rendered_prompt` per its own doc comment ("writing them apart would leave a window where the prompt advertises a setting whose env var the subprocess sandbox still refuses to pass through"). `resolve_hand_allowed_env` itself is unit-tested in isolation (`allowed_env_covers_selected_option_and_requirements`), but nothing here exercises the wiring: that a settings save which adds a `provider_env` (e.g. selecting the `live` option in the trading-mode style fixture) actually widens `entry.manifest.metadata["hand_allowed_env"]` on the live agent, and that dropping the last one removes the key rather than leaving a stale wider allowlist in place. Given the security note on that code path (blocking daemon-secret exfiltration through a widened passthrough), this seems worth an assertion at the integration level per the repo's "mandatory integration test for wiring changes" rule, not just a kernel-crate unit test.

---

houko (commented):
Automated daily review pass on this large PR. Focused on the multi-role rollback semantics of the new `update_hand_config`, the shared env-allowlist filter (`resolve_hand_allowed_env`), and the config-read/write surface (`WRITABLE_EXACT_PATHS`). Two inline notes below — both narrow-window / test-coverage observations rather than blockers; no CLAUDE.md violations found (changelog fragments, prose wrapping, BTreeMap ordering, Default-impl requirements, auth allowlist, and read/write config asymmetry all check out).

---

crates/librefang-kernel/src/kernel/boot.rs
```diff
@@ -1936,87 +1936,70 @@ impl LibreFangKernel {
...
+                                                    // recognise, and the activation path will
+                                                    // re-stamp the tags on the next
+                                                    // `hand activate`.
                                                     let role_opt = entry
                                                         .manifest
                                                         .tags
                                                         .iter()
                                                         .find_map(|t| t.strip_prefix("hand_role:"))
                                                         .map(|s| s.to_string());
                                                     if let Some(role) = role_opt {
```
houko, line 1953: Possible narrowing regression here, worth a second look before merge.

Before this change, the settings tail (`apply_settings_block_to_manifest`) was re-rendered at boot whenever `def.settings` was non-empty, independent of whether `role_opt` resolved — only the skill/team tails were gated on the `hand_role:<role>` tag being present.

Now `rerender_hand_prompt_tails` requires a `role: &str` (it needs it for the per-role skill override and the team-peer exclusion), so all three tails — including the settings tail, which never needed a role — are gated behind `role_opt.is_some()`. For a hand-derived agent whose `hand_role:<role>` tag is missing (the case the `else` branch already logs a `debug!` for as a known, pre-existing degradation), settings that previously survived a restart via the role-independent path now silently stop rendering too, until the operator re-runs `hand activate`.

This is a narrow edge case — the tag is stamped unconditionally at activation (`hands_lifecycle.rs` around the `format!("hand_role:{role}")` call), so it should only bite pre-existing/legacy persisted state from before role tagging existed. But it is a real behavior change from a refactor whose stated goal is "rendering is now one path, not three," and the settings-tail case doesn't actually need that unification (only skill/team do). Worth a comment (or a quick call to `render_settings_block` directly in the `else` branch) confirming this is accepted, rather than an unintended side effect of collapsing the three call sites into one signature.

---

houko (commented):
Reviewed the diff against CLAUDE.md (deterministic prompt ordering, config/Default rules, prose wrapping, changelog convention). Overall this is a careful, well-tested fix — settings-env allowlist resolution correctly stays Vec-ordered (no HashMap iteration reaching a prompt), `effective_setting_values` deliberately uses `BTreeMap` for stable HTTP/CLI output, `undeclared_setting_keys` sorts its HashMap-derived output before returning, and the two changelog.d/fixed fragments are present and correctly formatted ending in (#6637) (@houko). Pushed one small mechanical fix (rewrapped three new doc-comments that were hard-wrapped mid-sentence instead of at sentence boundaries, per the repo's prose-wrapping rule). Left one inline comment on a possible narrowing regression in the boot-time re-render path that I wasn't confident enough to fix myself, since the right resolution is a maintainer call.

---

houko:
Daily automated review pass — scope: CLAUDE.md compliance only. This is a large PR; skimmed for the specific rules rather than a full design re-review (a fuller pass is running separately given the size). One finding so far:

The tip commit, `674bc8ee` ("docs(kernel): rewrap three new hand-lock doc-comments to sentence boundaries"), carries `Claude <noreply@anthropic.com>` as the git commit **author** identity, not just in message text. CLAUDE.md's "no AI attribution" rule and the repo's `commit-msg` hook both scan commit message text, so neither catches attribution living in the author field — the same gap previously flagged on #6504, and also present on #6644's tip commit today.

Not fixing this myself: it's the single tip commit here with nothing stacked after it, so amending its author is mechanically trivial, but doing so changes the SHA and needs a force-push, which this review pass's own rules forbid regardless of how safe the rewrite would be. Maintainer call: amend-and-push (safe here since it's the last commit), or leave it — a squash-merge would discard individual commit authorship anyway.

A more complete CLAUDE.md pass over the rest of the diff is in progress and I'll follow up separately if it turns up anything else.

---

houko:
Follow-up to the full pass: pushed `a229edf9`, which rewraps the remaining `//` comment blocks in `hands_lifecycle.rs` and `boot.rs` that were still hard-wrapped at a column width (the `///` doc comments on `hand_instance_lock`/`hand_instance_locks`/`parse_hand_with_settings` were already fixed in `674bc8ee`, but several plain explanatory comments were missed). Comment-only, no behavior change — `cargo fmt --check -p librefang-kernel` is clean.

Everything else checked out: no new `HashMap`/`HashSet` reaching an LLM prompt (`resolve_settings` iterates the ordered `Vec<HandSetting>`, `effective_setting_values` deliberately returns a `BTreeMap`), both new changelog fragments are correctly named/formatted and end `(#6637) (@houko)`, no new config-struct fields so the `#[serde(default)]`/Default-impl rule doesn't apply, and `useUpdateHandSettings` invalidates via the existing hand/agent/overview factory keys correctly.

One behavioral question worth a maintainer look, not a style nit: in `boot.rs`, before this PR the settings tail was re-rendered on boot regardless of whether the agent had a `hand_role:<role>` tag (only the skill/team tails were gated on it). After unifying all three tails into `rerender_hand_prompt_tails` (which needs the role to render), the call now sits entirely inside `if let Some(role) = role_opt`, so an agent that's hand-tagged but missing the role tag — an artifact of an incomplete/interrupted activation — now gets *none* of the three tails re-rendered on boot, where it previously still got settings. The `else` branch logs a debug message about the degradation but doesn't render anything. Narrow edge case, not mentioned in the PR body, not covered by a test — flagging for a call on whether that's acceptable collateral of the unification or worth a narrower fallback.

---

houko (commented):
Automated review pass (diff against `origin/main`, focused on CLAUDE.md conventions — deterministic prompt ordering #3298, config-field conventions, changelog.d format, commit conventions).

**Deterministic prompt ordering (#3298):** no regressions found. `HandInstance.agent_ids` is a `BTreeMap`, the three rendered tails are assembled from a fixed-order array (not a hash-keyed structure), `undeclared_setting_keys` explicitly sorts before returning, and `effective_setting_values` returns a `BTreeMap` keyed by setting for the same reason. `resolve_settings` / `rerender_hand_prompt_tails` both iterate the schema's `Vec<HandSetting>` in declared order. Nothing here reintroduces a `HashMap`/`HashSet` on a path that reaches an LLM prompt.

**Changelog fragments:** both `changelog.d/fixed/6637-*.md` files match the documented format (section dir, PR-number-prefixed filename, no leading `- `, trailing `(#6637) (@houko)`).

**Config fields:** no new `KernelConfig` struct fields were added — `manage.rs` only widens which existing `ApprovalPolicy` fields the response serializes, so the struct/`#[serde(default)]`/`Default` checklist doesn't apply here, and the new `approval_policy_fields_are_all_readable` test is a good guard against the same class of gap recurring.

**Locking/rollback in `update_hand_config`:** the per-instance mutex acquisition order (check existence → lock → re-read authoritative snapshot), the all-or-nothing rollback across config + per-role prompt rewrite, and the `active_instance`-vs-`list_instances` scan fix all look correct against the race they describe, and are backed by targeted tests (`rerender_preserves_reference_and_team_tails`, `skill_content_containing_a_team_heading_survives_intact`, `update_hand_settings_skips_roles_absent_from_the_reloaded_definition`).

**One minor nit (not a functional issue):** CLAUDE.md's prose-wrapping rule ("one sentence = one line... applies universally... Commit message bodies") isn't followed in this PR's commit bodies — e.g. `f75d6d6` and `2f4284a` have multi-sentence paragraphs on a single line rather than one sentence per line. Non-blocking since rewriting history on an already-pushed, same-repo branch isn't something to do lightly mid-review; flagging only so a future amend (if one happens for another reason) can pick it up.

No mechanical fixes pushed — the diff is clean on every dimension this pass checked, including the two things that are easiest to get wrong here (hash-map iteration reaching a prompt, and settings-save atomicity).

---

Issue #6631, opened 2026-07-29
security: Admin can execute plugin-controlled dependency installers as daemon user


whatnick:
## Summary

Admin-role callers can install attacker-controlled plugins and invoke plugin dependency installation under the LibreFang daemon identity.
This crosses the intended Admin/Owner privilege boundary and permits arbitrary code execution through package lifecycle scripts or malicious build dependencies.

## Affected code

- `crates/librefang-api/src/middleware.rs`: Admin is allowed write routes that are not explicitly Owner-only.
- `crates/librefang-api/src/routes/plugins.rs`: plugin installation and dependency-installation routes are exposed through that policy.
- `crates/librefang-runtime/src/plugin_manager/install.rs`: Git plugin sources are cloned without establishing publisher trust.
- `crates/librefang-runtime/src/plugin_manager.rs`: npm, pip, Bundler, and Composer dependency installers run under the daemon user and inherit its execution context.

## Attack path

1. An attacker obtains Admin credentials.
2. The attacker installs a Git-backed plugin containing a malicious package lifecycle script or build dependency.
3. The attacker invokes the plugin dependency-installation endpoint.
4. The dependency tool executes attacker-controlled code as the daemon user.

## Impact

The attacker can read daemon-accessible files and credentials, modify LibreFang state, access inherited environment secrets, and execute arbitrary commands with the daemon account permissions.

This resembles #5587, but that report covered Hand dependency installation rather than the plugin installation routes.

## Recommended remediation

- Require Owner authorization for plugin source installation, dependency installation, hook testing, and every plugin operation that can execute plugin-controlled code.
- Run dependency installation in an isolated sandbox with a minimal environment.
- Disable package lifecycle scripts where supported unless explicitly required and approved.
- Add authorization integration tests proving Admin receives `403 Forbidden` while Owner remains permitted.

## Acceptance criteria

- Admin cannot invoke any plugin endpoint that installs or executes untrusted plugin code.
- Plugin dependency installers cannot access daemon secrets through inherited environment variables.
- Tests cover the complete plugin install and dependency-execution route set.

Issue #6630, opened 2026-07-29
security: MCP server list exposes inline environment secrets to Viewer users


whatnick:
## Summary

`GET /api/mcp/servers` serializes each configured MCP server’s raw environment entries to callers with the Viewer role.
The supported `KEY=VALUE` representation allows API tokens and other credentials to be returned in plaintext.

## Affected code

- `crates/librefang-api/src/routes/skills/mcp.rs`: the MCP server list response includes the raw `env` field from each server configuration.
- `crates/librefang-runtime-mcp/src/lib.rs`: MCP environment configuration accepts inline `KEY=VALUE` entries.
- `crates/librefang-api/src/middleware.rs`: ordinary GET routes are available to Viewer-role callers, including the MCP server list route.

## Attack path

1. An operator configures an MCP server with an inline value such as `GITHUB_PERSONAL_ACCESS_TOKEN=<token>`.
2. A Viewer-role caller requests `GET /api/mcp/servers`.
3. The response exposes the complete environment entry, including the credential value.

## Impact

A low-privilege Viewer can steal credentials for GitHub or other upstream services and use the permissions granted to those credentials outside LibreFang.

This differs from #3823, which covered environment expansion into child-process arguments rather than API response disclosure.

## Recommended remediation

- Never serialize MCP environment values in list or detail responses.
- Return only variable names and whether each variable is configured.
- Store secret values through the credential vault rather than inline configuration where possible.
- Audit other configuration responses for equivalent `KEY=VALUE` disclosure.

## Acceptance criteria

- Viewer responses contain no MCP environment values.
- Owner responses also redact secret values by default.
- Integration tests configure a recognizable sentinel secret and prove it does not appear anywhere in list or detail response bodies.

Issue #6629, opened 2026-07-29
security: OAuth refresh fallback can return another user’s tokens


whatnick:
## Summary

`POST /api/auth/refresh` is not bound to the authenticated LibreFang user.
When the request omits a refresh token, provider-wide fallback lookup can select another user’s stored refresh token and return newly refreshed upstream tokens to the caller.

## Affected code

- `crates/librefang-api/src/oauth.rs`: stored tokens are indexed by upstream OIDC subject rather than authenticated local-user ownership.
- `find_by_provider` and `find_any_with_refresh` select a matching stored token without verifying that it belongs to the authenticated caller.
- `auth_refresh` performs those fallback lookups and returns the refreshed access token and any rotated refresh token.
- The route is available to Admin through the generic write authorization policy.

## Preconditions

- External authentication is enabled.
- A victim has an unexpired stored refresh token.
- The attacker has valid Admin credentials.

## Attack path

An Admin submits an auth refresh request without a refresh token and identifies the victim’s provider, or relies on the single-provider fallback.
The server chooses an arbitrary matching stored session, refreshes it, and returns the upstream credentials to the Admin caller.

## Impact

An Admin can impersonate another user at the upstream OAuth provider and exercise all scopes granted to that user’s token.

## Recommended remediation

- Bind every stored OAuth token to the authenticated local user in addition to provider and upstream subject.
- Resolve refresh tokens only from records owned by the authenticated caller.
- Remove arbitrary provider-wide and global token fallbacks from authenticated refresh requests.
- Return an explicit error when no caller-owned refresh token is available.

## Acceptance criteria

- A caller can refresh only tokens owned by that caller.
- Omitting `refresh_token` never selects another local user’s token.
- Integration tests create two users and prove that neither can refresh or receive the other user’s upstream credentials.

PR #6638, opened 2026-07-29
feat(deploy): single-replica Kubernetes baseline, readiness contract, and rootless restricted-PSS container

### LINKED ISSUE #6632: container: support Kubernetes restricted Pod Security without a root main process
## Summary

The published image starts its entrypoint as root so it can recursively chown `/data`, rewrite first-boot configuration, and then use `gosu` to launch LibreFang as UID/GID 1001.

This runtime privilege drop protects the daemon process, but Kubernetes clusters enforcing the `restricted` Pod Security Standard commonly require `runAsNonRoot: true` and reject or prohibit a root-starting main container.

## Current behavior

- The final Docker stage intentionally has no `USER` directive.
- `deploy/docker-entrypoint.sh` creates and chowns `/data`, initializes configuration, rewrites listen/model settings, then executes through `gosu librefang`.
- Setting `runAsUser: 1001` directly bypasses the privilege assumptions of the entrypoint and can fail on fresh or differently owned persistent volumes.

#3859 previously tracked the missing `USER` directive, but the current root-entrypoint design still does not provide an image or documented path that satisfies Kubernetes `runAsNonRoot`.

## Proposed scope

Provide one officially supported restricted-pod path without breaking existing Docker volume initialization.
Acceptable designs include:

- A rootless image/entrypoint variant that requires `/data` to be pre-owned by UID/GID 1001.
- A Kubernetes init container that prepares the PVC while the main LibreFang container runs permanently as UID/GID 1001.
- Refactoring initialization so the main image can declare `USER 1001:1001` and operate on a volume prepared through `fsGroup` or equivalent storage semantics.

The supported Kubernetes security context should also drop all Linux capabilities, disable privilege escalation, and use the default seccomp profile.

## Acceptance criteria

- An official configuration boots from a fresh PVC with `runAsNonRoot: true`.
- The LibreFang main container never executes as UID 0.
- The main container runs as UID/GID 1001 with `allowPrivilegeEscalation: false`, all capabilities dropped, and `seccompProfile.type: RuntimeDefault`.
- Existing Docker Compose users retain a working migration path.
- CI boots the restricted configuration and probes the API successfully.
- Documentation explains volume ownership requirements for CSI drivers that do not honor `fsGroup`.

### LINKED ISSUE #6633: api: add a readiness endpoint that fails when required dependencies are degraded
## Summary

`GET /api/health` is suitable as a process liveness endpoint, but it always returns HTTP 200 even when its JSON body reports `status: "degraded"` because the SQLite connectivity check failed.

Kubernetes HTTP probes evaluate the status code and cannot inspect JSON fields, so the current endpoint cannot remove a degraded pod from Service endpoints.

## Current behavior

`crates/librefang-api/src/routes/config/system.rs::health` checks the memory substrate and returns database and embedding status, but wraps every result in a normal JSON response with HTTP 200.

The Docker `HEALTHCHECK` also treats any HTTP 200 response as healthy.
Changing `/api/health` itself to return 503 for dependency failures would conflate liveness and readiness and could cause restart loops during recoverable storage or provider incidents.

## Proposed contract

- Keep `GET /api/health` as the public liveness endpoint that returns 200 while the HTTP process is responsive.
- Add a public minimal `GET /api/ready` endpoint.
- Return 200 only when dependencies required to accept work are ready.
- Return 503 with a minimal, non-sensitive status body when required checks fail.
- Treat optional embedding availability according to the configured memory mode rather than making every warning fatal.
- Document startup, liveness, and readiness probe usage for Docker and Kubernetes.

## Acceptance criteria

- `/api/ready` returns 200 when the required storage substrate is operational.
- `/api/ready` returns 503 when the required database check fails.
- `/api/health` remains a stable liveness signal and does not trigger restart loops for a readiness-only failure.
- The readiness response does not expose detailed diagnostics without authentication.
- Router integration tests cover both status paths and confirm the endpoint remains reachable under configured authentication.
- OpenAPI and deployment documentation describe the distinct contracts.

### LINKED ISSUE #6634: RFC: define the architecture required for active-active multi-replica deployments
## Summary

LibreFang can run reliably as a single stateful Kubernetes pod, but it cannot currently operate as an active-active Deployment with multiple replicas.

The limitation is broader than replacing SQLite: scheduling, cron, trigger dispatch, sessions, budgets, audit-chain state, agent registries, background workers, locks, and several caches assume one authoritative daemon process.

## Current constraints

- `run_daemon` takes an exclusive `daemon.lock` under `LIBREFANG_HOME`, intentionally preventing two daemons from sharing one state directory.
- SQLite WAL provides local concurrency, not distributed coordination.
- Agent and task execution state is substantially process-local.
- Cron and trigger ownership has no distributed leader-election or lease protocol.
- Multiple replicas could duplicate scheduled work, race session mutation, diverge audit-chain tips, and enforce budgets inconsistently.
- Sharing one ReadWriteMany filesystem does not solve those coordination problems and may weaken SQLite/flock guarantees.

## RFC questions

The design should decide:

- Whether the target is active-active execution, active-passive failover, or stateless API replicas around a singleton scheduler/runtime.
- Which durable database replaces or supplements SQLite and the migration path for existing installations.
- How leader election and leases govern cron, triggers, compaction, background agents, and recovery workers.
- How `(agent_id, session_id)` execution ownership is acquired, renewed, fenced, and recovered after pod loss.
- How event delivery and idempotency prevent duplicate tool or channel side effects.
- How budgets, audit-chain ordering, approvals, OAuth state, WebSocket sessions, and caches remain consistent.
- Whether agents are partitioned across workers and how routing discovers the owning replica.
- What compatibility guarantees remain for the default single-node SQLite mode.

## Expected deliverable

Produce an architecture RFC with:

- Explicit consistency and failure-model guarantees.
- Component and data-flow diagrams.
- Storage and coordination technology decisions.
- A phased implementation plan preserving single-node operation.
- Migration, rollback, observability, and disaster-recovery plans.
- Integration and chaos-test scenarios covering duplicate prevention, leader loss, network partitions, and interrupted agent turns.

## Acceptance criteria

- The repository clearly distinguishes supported single-replica deployment from future HA support.
- The RFC identifies every singleton subsystem and assigns an ownership/coordination mechanism.
- The design prevents duplicate externally visible side effects during failover.
- The proposal includes measurable readiness criteria before `replicas > 1` is documented as supported.

### LINKED ISSUE #6635: deploy: provide an official single-replica Kubernetes deployment baseline
## Summary

LibreFang already publishes a multi-architecture `ghcr.io/librefang/librefang` image for `linux/amd64` and `linux/arm64`, but the repository provides Docker Compose only and does not define a supported Kubernetes deployment pattern.

The initial Kubernetes contract should deliberately support one stateful daemon replica rather than imply active-active behavior that the current SQLite and process-local architecture cannot provide.

## Current constraints

- Runtime state lives under `LIBREFANG_HOME=/data`.
- SQLite uses WAL files beside the main database.
- `run_daemon` holds an exclusive `/data/daemon.lock`, so two pods sharing one volume cannot run concurrently.
- The container listens on port 4545 when `LIBREFANG_LISTEN=0.0.0.0:4545`.
- Non-loopback binding requires configured authentication unless the operator explicitly opts out.
- Release images already carry a multi-arch manifest for amd64 and arm64.

## Proposed scope

Add a supported manifest set under `deploy/kubernetes/` using Kustomize or plain YAML with:

- A one-replica `Deployment` using `strategy: Recreate`, or a one-replica `StatefulSet`.
- A `ReadWriteOnce` PVC mounted at `/data`.
- A ClusterIP `Service` exposing port 4545.
- Resource requests and limits with documented starting values.
- Startup, liveness, and readiness probes using the supported health contracts.
- `terminationGracePeriodSeconds` long enough for the daemon shutdown path to flush SQLite WAL state.
- Secret references for dashboard authentication, `LIBREFANG_VAULT_KEY`, provider keys, and optional external-auth state signing.
- `LIBREFANG_LISTEN=0.0.0.0:4545`.
- Documentation that shared NFS-style storage is not supported for the SQLite/WAL workload unless its locking guarantees have been explicitly validated.
- Documentation that `replicas` must remain 1 until a separate multi-replica architecture is implemented.

Do not duplicate the image build pipeline: `.github/workflows/release.yml` already publishes amd64 and arm64 images under one manifest.

## Validation

Add CI validation that:

- Renders the manifests successfully.
- Verifies required selectors, volume mounts, secret references, probes, and `replicas: 1`.
- Boots the manifest in a lightweight Kubernetes test environment and reaches the public health endpoint.
- Exercises a restart and confirms data in `/data` survives pod replacement.

## Acceptance criteria

- A documented deployment works on both amd64 and arm64 Kubernetes nodes using the published image.
- Pod replacement preserves LibreFang state.
- Updates cannot briefly start two pods against the same data volume.
- Authentication and vault keys come from Kubernetes Secrets rather than literal manifest values.
- The documented support boundary clearly states single-replica operation.

houko:
Closes #6635, closes #6633, closes #6632, closes #6634.

## Why these four together

They are one cluster because each one's acceptance criteria references another's deliverable, and splitting them would ship broken cross-references:

- #6635 requires "startup, liveness, and readiness probes using the supported health contracts" — that needs `/api/ready` from #6633 to exist.
- #6632 requires "CI boots the restricted configuration and probes the API successfully" — that needs both the manifests and the probe.
- #6635 must document that "`replicas` must remain 1 until a separate multi-replica architecture is implemented", and #6634's own criterion is that "the repository clearly distinguishes supported single-replica deployment from future HA support". Those are two sides of one sentence, and the manifests' README links the RFC directly.

Landing them separately would mean either a dead link in the README or a probe-less manifest.

## `/api/ready` — readiness as a contract distinct from liveness (#6633)

`GET /api/health` always returns 200 while the HTTP server can answer, even when its own body reports `status: "degraded"` because the SQLite check failed. Kubernetes probes read only the status code, so that endpoint can never remove a degraded pod from Service endpoints. Making `/api/health` return 503 instead would have conflated liveness with readiness and restart-looped pods through recoverable incidents, so the two are now separate endpoints.

- 200 when the dependencies required to accept work are available, 503 otherwise.
- Required means the SQLite substrate always, plus an embedding driver only when the operator pinned a specific `memory.embedding_provider` while leaving vector search on. An unset or `"auto"` provider, or `fts_only = true`, is the documented fallback rather than a degradation and never fails readiness.
- Public, because a kubelet carries no credential and a 401 would pin the pod out of Service endpoints permanently. Registered in the `is_public` allowlist in `middleware.rs`, not by route ordering.
- Payload is minimal by design — check names, coarse status, and whether each check is required. No version, hostname, provider id, model name, or error text; diagnostics stay behind the authenticated `/api/health/detail`.
- Priced at 1 rate-limit token like `/api/health`, for both the versioned and unversioned spellings. A pod runs three probes concurrently and all arrive from the node's kubelet — one source IP, which is what the GCRA limiter keys on. At the 5-token fallback a 1s startup probe alone would drain the 500-token/minute budget and 429 the pod out of its own rollout.
- The requirement is snapshotted at boot in `AppState::readiness_requires_embedding`, not read live. `POST /api/config/reload` swaps the whole `KernelConfig` whenever the plan carries a hot action, and a `[memory]` change is `restart_required` — so reading it live let an edit introduce a requirement against a driver that is never rebuilt, pinning readiness at 503 forever with no automatic recovery. A config-file edit would have become an outage.

**Behaviour change worth calling out for existing Compose users**: the image `HEALTHCHECK` now probes `/api/ready`. Docker does not restart an unhealthy container — the consumer of that check is Compose's `depends_on: condition: service_healthy` gate, which is readiness semantics, and `/api/health` could never fail it. A deployment that pins an embedding provider whose key is absent will now report unhealthy where it previously reported healthy. That is the intended signal, but it is a visible change.

## Rootless container path for restricted Pod Security (#6632)

The image declares no `USER` because its entrypoint starts as root to chown a bind-mounted `/data`, then drops to uid 1001 via `gosu`. Kubernetes `restricted` requires `runAsNonRoot: true` and rejects a root-starting main container — including init containers, so a root chown init container is not a way out either.

`deploy/docker-entrypoint.sh` now asks `id -u` what it already is and adapts, so one image serves both worlds with no second rootless tag to maintain:

- Under root: unchanged. chown, `gosu` for helpers and the final exec. Existing Docker and Compose deployments keep working, which is the migration path the issue requires.
- Under `runAsUser: 1001`: chown and `gosu` are both skipped — neither is possible unprivileged, and dropping privileges requires the privilege to do so. It verifies `/data` is writable up front and exits non-zero naming the `fsGroup` remedy, rather than letting the daemon die later on an opaque `EACCES`.

## Single-replica Kubernetes baseline (#6635)

`deploy/kubernetes/` as Kustomize manifests plus operator documentation. One-replica StatefulSet with a headless governing Service and a ClusterIP Service for clients, a `ReadWriteOnce` PVC at `/data`, resource requests and limits, `terminationGracePeriodSeconds: 60`, and every credential from a `secretKeyRef` — never a literal.

StatefulSet rather than Deployment because it terminates its pod before creating the replacement, so two daemons never contend for the volume. A Deployment would need `strategy: Recreate`; the README says so.

The pod satisfies `restricted` as published: `runAsNonRoot`, uid/gid 1001, `fsGroup: 1001` with `fsGroupChangePolicy: OnRootMismatch`, `allowPrivilegeEscalation: false`, all capabilities dropped, `seccompProfile: RuntimeDefault`. `readOnlyRootFilesystem` is deliberately absent — `restricted` does not require it and the daemon writes outside `/data` (the python venv it provisions for MCP servers, plugin dependency installs, `$TMPDIR` scratch).

`LIBREFANG_API_KEY` is added as an env override because a Secret had no way to supply `api_key`: `config.toml` lives in the daemon's own writable data dir and is rewritten at boot. An empty value is ignored with a warning rather than treated as "clear the key" — a Secret key that exists but holds nothing is indistinguishable from that, and clearing the key disarms bearer auth, which on a non-loopback bind means an open daemon.

Documentation covers the support boundary, volume ownership including CSI drivers that ignore `fsGroup`, why shared NFS-style storage is unsupported for a WAL workload, the probe contracts, and a Compose migration path.

## Multi-replica RFC (#6634)

`docs/architecture/multi-replica-rfc.md`. "Just replace SQLite" is the wrong model for what blocks HA, so the document enumerates the singletons against the tree: the exclusive `daemon.lock` flock, all 24 `spawn_logged` background workers with what each duplicates under a second replica, the in-process `agent_msg_locks` / `session_msg_locks`, the per-process audit chain tip (two appenders fork the chain, and a hash chain has no join), the in-memory `CostReservationLedger`, `MemorySubstrate` holding a `Pool<SqliteConnectionManager>` with no backend seam, and the process-local OAuth token store, dashboard sessions, and session stream hub.

It then decides rather than surveys: stateless API replicas around a leader-elected execution plane instead of active-active, PostgreSQL behind a new backend trait with SQLite remaining the default indefinitely, Postgres advisory locks rather than the Kubernetes Lease API so coordination works identically under systemd and Compose, fencing tokens rather than heartbeats, and idempotency keys on external side effects.

**The storage and coordination decisions are marked as needing maintainer approval before any implementation starts**, since they add a core dependency. Four phases each ship independently and leave single-node SQLite working; `replicas > 1` becomes documented only after seven measurable readiness criteria hold.

## Out-of-cluster fix found while verifying: macOS WASM sandbox aborts the daemon

Not part of the four issues. Found while running the kernel unit tests for the work above and fixed here at the maintainer's request rather than deferred.

`cargo test -p librefang-kernel --lib` died on SIGABRT partway through, at a different test each run. Eight consecutive crash reports all named `machports::handler_thread`, and `--nocapture` surfaced the line libtest's output capture had been swallowing on abort:

```
mach_msg failed with 268451845 (10004005)
```

`0x10004005` is `MACH_RCV_INTERRUPTED`. Wasmtime defaults to Mach exception ports on macOS, parking a handler thread in a blocking `mach_msg` receive flagged `MACH_RCV_INTERRUPT` — which asks the kernel to return that status instead of restarting the call when a signal lands on the thread. Wasmtime treats anything but "port closed" as fatal and calls `libc::abort()`.

This is a production availability bug, not a test artifact. The daemon delivers such signals constantly — SIGCHLD from MCP stdio servers, sidecar channels, exec-tool subprocesses, provider CLI probes — and a process-directed signal may be routed to any thread that has it unblocked. On macOS, any child exiting while the WASM sandbox was live could kill the whole daemon, with no panic, no unwind, and no log beyond that one line.

`make_engine_config` now selects POSIX signal handling: per-process handler, inherited across `fork()`, no thread parked in an interruptible syscall. Guest traps are still caught correctly — both mechanisms are first-class in wasmtime — and Mach ports only buy coexistence with foreign Mach-port-based crash reporters, of which the dependency tree has none.

Attribution was confirmed by reverting the unrelated kernel change and reproducing the abort on a clean base, so it is pre-existing and not introduced here.

CI could not have caught this and would not catch a revert: the macOS lane runs nextest, which gives each test its own process, so no single test accumulates the subprocess churn and thread count needed to land a signal on that one thread. A `cargo test -p librefang-kernel --lib` step is added to the macOS lane as the guard, with the reasoning inline.

An in-crate `#[tokio::test]` guard was written first and then removed: with Mach ports re-enabled it still passed, so it asserted nothing. A regression test that does not fail against the bug is worse than none, because it reads as coverage.

## Verification

Compile and lint:

- `cargo clippy --workspace --all-targets -- -D warnings` — clean, zero warnings.

Tests:

- `cargo test -p librefang-api --lib` — 878 passed.
- `cargo test -p librefang-kernel --lib` — 1339 passed, 0 failed (was SIGABRT before the wasmtime fix).
- `cargo test -p librefang-runtime --lib sandbox` — 120 passed, including the epoch-timeout trap test and a real WASM skill execution.
- `cargo test -p librefang-api --test config_routes_integration --test auth_public_allowlist --test dead_route_audit_test --test openapi_path_coverage_test --test openapi_spec_test --test route_smoke` — 65 passed.

New integration tests in `config_routes_integration.rs`: the public 200 path, the 503 path, the payload-disclosure envelope, `fts_only`, `"auto"`, and the config-reload drift regression. New unit tests in `boot.rs` for the `LIBREFANG_API_KEY` envelope including the empty-value case.

Manifests:

- `kubectl kustomize deploy/kubernetes/base` renders; `kubectl apply --dry-run=server` accepts it; `scripts/check-k8s-manifests.py` passes.
- The validator was **mutation-tested**: ten deliberate regressions (replicas 2, liveness on `/api/ready`, removed `fsGroup`, an inlined literal api-key, `ReadWriteMany`, an unmounted claim, `allowPrivilegeEscalation: true`, `runAsNonRoot: false`, a 5s grace period, a removed memory limit) are each caught, so the checks are not vacuous.
- The CI overlay path was verified after fixing it: kustomize refuses an absolute resource root, so the original `/tmp` overlay pointing at `$GITHUB_WORKSPACE` could never have loaded. It is now in-tree with a relative path, and renders correctly.

Entrypoint, verified inside `debian:bookworm-slim` (same base family as the published image — GNU sed, GNU coreutils, real uid/gid handling) on a Docker **named volume**, because a macOS bind mount does not preserve ownership semantics and silently passes every permission check:

- Root branch: chowns the volume, invokes gosu, lands the daemon at 1001:1001, both config rewrites applied, config.toml owned by librefang.
- Rootless on a 1001-owned volume: skips gosu, runs init unprivileged, applies both rewrites, creates the logs dir, cleans up its write probe.
- Rootless on a root-owned volume: fails fast with the `fsGroup` message, daemon never starts.
- Second rootless boot: does not re-run `librefang init`.

Generated artifacts regenerated and committed: `openapi.json` (+`/api/ready`), the four SDKs, and `xtask/baselines/openapi.sha256`.

## Not verified here

The `kind` lane itself has not run — it is `if: github.event_name != 'pull_request'` (following docker-build.yml's precedent, since the Rust release build dominates its runtime), so it first executes on merge to main. Its constituent parts were verified locally: the overlay renders, the validator passes, and the entrypoint's rootless branch is container-tested. Worth a `workflow_dispatch` run before merge if you want the full boot proven first.

Live LLM verification was not performed — no change here touches an LLM call path.

## Still open from this reporter

@whatnick also filed three security issues that are not addressed here: #6629 (OAuth refresh can return another user's tokens), #6630 (MCP server list exposes inline environment secrets), #6631 (Admin can execute plugin-controlled dependency installers). They are a separate cluster and will get their own branch and PR. Worth noting that #6630 is **worse than reported**: `/api/mcp/servers` is in `PUBLIC_ROUTES_DASHBOARD_READS` (`middleware.rs`), so in open mode an unauthenticated caller reads MCP env values in plaintext, not just a Viewer-role one.


docs/architecture/multi-replica-rfc.md
houko, line 1: CLAUDE.md's prose-wrapping rule ("no column limit; break only at sentence boundaries — one sentence = one line") applies to all new markdown docs, but this file has many lines that pack two-plus sentences together, e.g. line 132: "`librefang start` on a laptop must never require a Postgres. The backend is selected by config…", line 172: "`CostReservationLedger` becomes a `cost_reservations` table. A pre-charge is an insert…", and line 279: "**Phase 0 — document the boundary. Done (#6635, #6632, #6633).**". There are several dozen more instances throughout the doc (table cells are presumably exempt since GFM cells can't contain hard line breaks). Given the volume, I didn't attempt a full mechanical rewrap myself — flagging for the author/maintainer to decide whether to rewrap before merge or accept as a known style debt on this large new document.

---

deploy/kubernetes/base/statefulset.yaml
houko, line 1: Same prose-wrapping issue as the RFC doc: several `#` comment blocks here are hard-wrapped at a column width rather than broken at sentence boundaries, so two sentences land on one physical line, e.g. "Pod Security Standard `restricted` (#6632). The image declares no / USER directive because…" and "SQLite WAL checkpoint on shutdown plus any in-flight agent turn. The / daemon stops accepting new work…". CLAUDE.md's rule explicitly covers "any multi-line free-form prose comment block", so this applies here too. The same pattern recurs in `deploy/kubernetes/secrets.example.yaml`, `crates/librefang-runtime/src/sandbox.rs`, and `crates/librefang-kernel/src/kernel/boot.rs`'s new comments. Leaving this as one comment rather than editing every file — the volume (10+ files) makes a blanket mechanical rewrap risky to do unreviewed; happy to fix on request.

---

houko (commented):
Two prose-wrapping notes per CLAUDE.md's sentence-boundary-only line-break rule; see inline comments. Everything else checked out: changelog fragments are correctly formatted and attributed, the readiness endpoint is wired through the `is_public` allowlist (not route reordering) with a matching `#[tokio::test]` suite in config_routes_integration.rs, no new config-struct fields are missing `#[serde(default)]`, and no stray AI/Claude attribution anywhere in the diff.

---

PR #6641, opened 2026-07-29
feat: auto-detect EveryAPI CLI credentials

houko:
## Summary

- auto-detect a locally installed and logged-in EveryAPI CLI
- resolve short-lived credentials per request and refresh once after HTTP 401
- choose the initial default from the relay key's live chat-capable model directory
- keep explicit keys, provider files, URLs, and user suppression authoritative
- load EveryAPI models dynamically and preserve known model capabilities safely
- report auto-detection state and configuration conflicts in `librefang doctor`

## Cross-repository dependency

- requires everyapi-ai/everyapi#1081 for the versioned credential-process command and optional live model directory
- retains a region-aware legacy credential-cache fallback for older CLI releases

## Write-back guard fix (c7eba26)

The refresh change-detection guard introduced earlier in this branch required `provider.is_custom` before applying a completed refresh to a non-managed EveryAPI entry.
That is a behavior bug, not just a red test: `models connect everyapi` registers the gateway as `providers/everyapi.toml`, and both loaders land that entry with `is_custom = false` — `new_from_dir_with_registry` falls back to it for every provider when `registry/providers/` is missing or unreadable (its documented tri-state), and `From<ProviderCatalogToml>`, the runtime `load_catalog_file` path behind the registry install route, has no better answer either.
`detect_auth` then promotes the entry to `Configured` once the declared key env var is present and leaves the flag alone, so the refresh gate accepted it while the write-back silently refused it: two gateway round trips per TTL expiry, the 60-second retry slot burned, the snapshot discarded, and nothing surfaced to the operator.

`is_custom` was never the invariant the guard needed, and on this path it cannot change mid-fetch anyway — `ensure_managed_everyapi` refuses to touch a custom entry and nothing resets the flag.
The guard now requires that the entry has not been taken over by CLI-managed discovery and still declares the same credential env var; suppression and `base_url` were already checked above it.

`everyapi_refresh_merges_both_gateway_endpoints` asserts exactly that write-back and had been failing on `Test / Ubuntu (shard 2/4)` since the guard landed.
A unit test pins the two predicates the guard depends on against the shape a provider file actually produces.

## Credential provenance is now explicit (6cdf743)

Same root cause, more damaging half. Deciding whether an entry's credentials come from the CLI by reading `!is_custom && matches!(auth_status, AutoDetected | Missing)` misclassifies an explicit configuration in the other direction: installing a provider file at runtime through the registry route, before its key is set, produced a non-custom `Missing` entry that the gate admitted through `(!is_custom && Missing)` and `refresh_now` read as managed, so `ensure_managed_everyapi` rewrote `base_url` to whatever account the EveryAPI CLI happened to be logged into.
A self-hosted or regional gateway was silently repointed until the next daemon restart, where `everyapi_explicit` would have classified it correctly.

Neither flag ever meant provenance: `is_custom` answers "may the dashboard delete this?", and `auth_status` describes availability — an explicit config whose key env var is merely unset sits at `Missing` exactly like an unresolvable CLI login.
`ProviderInfo::cli_managed` now carries it, set only by `ensure_managed_everyapi` and cleared by every explicit source (`ensure_explicit_everyapi`, a user-set base URL, a provider-file merge naming its own `api_key_env`).
The gate's recovery arm, `refresh_now`'s managed/explicit branch, and `is_managed_provider_active` all key on the flag, so an entry with no reachable credential is inert instead of falling through to the CLI.

`detect_auth` also stops touching a CLI-managed entry's status entirely rather than only while it reads `AutoDetected` — no env var describes those credentials, so probing `api_key_env` could only promote the entry from an unrelated key and route its refresh down the explicit-key branch.
`ensure_managed_everyapi` keeps its `is_custom` guard: boot reaches it only after ruling out every explicit source, and runtime callers now arrive through a `cli_managed` gate.

## Verification

- `cargo fmt --check`
- `cargo clippy -p librefang-kernel -p librefang-runtime -p librefang-api -p librefang-cli --all-targets -- -D warnings`
- focused EveryAPI tests: kernel 13, API 31, runtime 4, CLI 10
- full affected-crate suites passed before the final rebase; focused suites passed again after rebase
- `git diff --check origin/main...HEAD`
- after c7eba26: `cargo test -p librefang-api --test providers_routes_test everyapi` (2 passed), `cargo test -p librefang-api --lib everyapi` (32 passed), `cargo clippy -p librefang-api --all-targets -- -D warnings` (clean)
- after 6cdf743: `cargo check --workspace --all-targets` (clean), `cargo fmt --all --check` (clean), `cargo clippy --workspace --all-targets -- -D warnings` (clean), `cargo test -p librefang-runtime --lib model_catalog` (109 passed), `cargo test -p librefang-api --lib everyapi` (33 passed), `cargo test -p librefang-api --test providers_routes_test everyapi` (2 passed), `cargo test -p librefang-types --lib` (891 passed)

## Fallback behavior

- automatic selection only consumes the CLI catalog when the CLI login is the selected credential source
- explicit EveryAPI keys never borrow another CLI account's catalog
- if live discovery fails, startup remains available through the existing built-in fallback

Issue: none. This cross-repository integration was requested directly.


crates/librefang-kernel/src/kernel/boot.rs
```diff
@@ -370,6 +404,67 @@ impl LibreFangKernel {
...
+            || std::env::var("EVERYAPI_API_KEY").is_ok_and(|key| !key.trim().is_empty())
+            || config
+                .home_dir
+                .join("providers")
+                .join("everyapi.toml")
+                .exists();
+        let everyapi_credential = if everyapi_suppressed || everyapi_explicit {
+            None
+        } else {
+            crate::everyapi_credentials::resolve(false).ok()
```
houko, line 435: `everyapi_credentials::resolve(false)` runs synchronously here inside `boot_with_config`, which every production call site invokes from `rt.block_on(...)` (see `commands/daemon.rs`) rather than via `spawn_blocking`. `COMMAND_TIMEOUT` in `everyapi_credentials.rs` is 45s, and the resolver can spawn a real subprocess (network-dependent OAuth refresh) that runs right up to that ceiling. That's a big asymmetry with the other boot-time reachability check a few lines above (`is_ollama_reachable`, 500ms `connect_timeout`), and it blocks a tokio worker thread on the 2-worker daemon runtime for up to 45s if the EveryAPI CLI is installed but slow/unresponsive — worse, a second `resolve(false)` call exists later in this function (around line 1108) and isn't deduped when the first call returned `None` (vs. an `Err`), so a hung CLI could cost up to ~90s of boot-time blocking in that path.

Is this current worst case intentional, or should this resolution go through `spawn_blocking`, get a tighter boot-time timeout than the request-time one, or dedupe more strictly? Not proposing a fix inline since it likely means threading an async path (or a shorter dedicated boot timeout) through `everyapi_credentials::resolve`, which is a design call beyond a mechanical patch.

---

houko (commented):
Automated review pass. Fixed a few prose-wrapping violations in doc comments this PR touched (pushed as a follow-up commit). One architectural question left as an inline comment on the boot-time credential resolution path; everything else — secret handling (Debug redaction, no key ever logged/persisted), suppression/explicit-override precedence, and determinism (sorted `available_models`, no order-sensitive HashMap use) — looked solid.

---

crates/librefang-kernel/src/everyapi_driver.rs
```diff
@@ -0,0 +1,261 @@
...
+            base_config: DriverConfig {
+                provider: "everyapi".to_string(),
+                max_retries,
+                ..DriverConfig::default()
+            },
+            managed_gate: None,
+        }
+    }
+
+    async fn resolve_driver(&self, invalidate: bool) -> Result<Arc<dyn LlmDriver>, LlmError> {
```
houko, line 77: Credential-handling design question, flagging for maintainer judgment rather than fixing myself.

`resolve_driver()` is called from both `complete()` (line ~132) and `stream()` (line ~146) on **every single agent turn**, and unconditionally invokes `self.source.resolve(false)` first — which spawns a fresh `everyapi auth credential --format=json` subprocess via `spawn_blocking` (see `everyapi_credentials.rs`), bounded by a 45-second `COMMAND_TIMEOUT`.

`EveryApiCredential.expires_at` is parsed from the CLI's response but never consulted here to skip re-resolution when the current credential is still fresh. So every LLM completion/stream call pays a subprocess-spawn round trip to the EveryAPI CLI, and if that CLI ever hangs or is slow (network hiccup during its own refresh), the request inherits up to 45s of added latency — worse under concurrent load, since each in-flight call spawns its own subprocess independently (no dedup/coalescing across concurrent calls for the same provider).

This may be an intentional tradeoff (always deferring freshness authority to EveryAPI, per the module doc's stated design), but it's a meaningful latency/reliability cost on the hot path for every agent message, not just credential refresh events. Worth confirming this is the intended behavior, or whether the resolved credential should be cached in-process against `expires_at` (e.g. only re-resolving when near expiry or after a 401) rather than on every call.

---

crates/librefang-api/src/everyapi_catalog.rs
```diff
@@ -528,19 +607,42 @@ async fn refresh_now(kernel: &Arc<dyn KernelApi>) -> Result<usize, String> {
...
     try_claim_refresh_slot(&base_url)?;
 
-    let live = fetch_live_models(&base_url, &api_key).await?;
+    let mut api_key = if managed {
+        let credential = resolve_managed_credential(kernel, false).await?;
+        base_url = credential.base_url;
+        credential.api_key
+    } else {
+        // Treat an empty env var the same as an absent one — `/v1/models`
+        // answers an empty bearer with a 401 either way.
```
houko, line 629: Mechanical nit, left as a comment rather than a pushed fix because this branch is moving fast (I hit two consecutive non-fast-forward rejections against concurrent pushes to `feat/everyapi-auto-detect` while preparing a fix, so I'm stopping after the one retry rather than racing further).

This comment is a single sentence still hard-wrapped across two lines, which is the one spot the repo's "one sentence per line" prose convention (`CLAUDE.md`) still misses in this PR's touched code — everything else in these four files looks correctly reflowed as of the latest push:

```rust
// Treat an empty env var the same as an absent one — `/v1/models`
// answers an empty bearer with a 401 either way.
```

Suggested fix — merge to one line:
```rust
// Treat an empty env var the same as an absent one — `/v1/models` answers an empty bearer with a 401 either way.
```

---

houko (commented):
Automated review pass for the EveryAPI auto-detect credential-handling change. Two inline notes: one architecture question about per-request credential subprocess spawning (flagging for maintainer judgment, not touching credential-handling logic myself), and one small leftover prose-wrap nit. Also see the separate general comment about commit authorship in this branch's history.

Verification performed (scoped, per repo convention): `cargo fmt --check`, `cargo check -p librefang-kernel -p librefang-api -p librefang-runtime -p librefang-cli --lib`, and `cargo test -p librefang-kernel --lib everyapi` (10/10 passing) all green against the current branch tip. I also pushed one small mechanical commit earlier in this session (rewrapping a few doc comments to one-sentence-per-line) but a concurrent push already landed an equivalent fix before mine could land, so nothing further was pushed to avoid churn.

---

houko:
Flagging for maintainer judgment, not something I can fix myself without rewriting already-pushed history (which is outside what this automated pass is authorized to do).

Two commits currently in this branch's history are authored as `Claude <noreply@anthropic.com>`:
- `48619f50` — "docs: add changelog fragment and fix prose wrapping for EveryAPI auto-detect"
- `10f44485` — "docs: rewrap EveryAPI auto-detect comments to one sentence per line"

The commit *messages* themselves are clean (no Claude/Anthropic text, so the `commit-msg` hook's text-based check wouldn't catch this), but the git author identity itself carries the attribution. `CLAUDE.md`'s "No AI / Claude attribution" rule is written in terms of commit message / PR body / comment text, so this may not trip any automated gate, but it seemed worth surfacing explicitly since it's the exact category of thing that rule exists to avoid, and it's now happened twice on this one PR (this appears to be a repo-wide pattern from prior automated runs, not unique to this branch — happy to say more if useful).

Whether to squash/rebase these commits before merge (to fold them into a maintainer-authored commit) is a judgment call I'll leave to you rather than force-pushing this branch myself.

---

PR #6504, opened 2026-07-19
feat(workflows): execution debug UI, live progress, re-run, and fixes

DaBlitzStein:
### Workflow execution UX
- **Live step progress**: current_step_index + total_steps on WorkflowRun, shown as progress bar
- **Re-run with params**: one-click re-run button, auto-populates from last run's inputs
- **Execution timeline**: per-step debug view with {{var}} → value bindings
- **Scrollable detail panels**: right panel scrolls independently, steps auto-expand during execution
- **Execution log console**: streaming log output in run detail view
- **Input params on failed runs**: error detail panel shows what was submitted
- **Empty-state for 0-step runs**: graceful display instead of broken UI

### Fixes
- **FallbackDriver**: strip provider prefix from primary model name in chain
- **agent_send async**: non-blocking by default, resolves default model in streaming path
- **channel_send guard**: exclude system channels (webui/cron/autonomous) from tool suggestion
- **React Hooks**: fix Rules of Hooks violation in WorkflowsPage (useState in map callback)
- **sender_chat_id**: stamp in streaming path for async callback routing
- **rerunWorkflowRun API**: add missing API function used by dashboard

### Verification
- cargo check --workspace --lib: clean
- 14/14 kernel goal tests pass (no regressions)
- Dashboard typecheck: green
- CI: 35 checks on previous submission all green

| Commit | Description |
|--------|------------|
| feat(workflows) | live step progress, re-run with params |
| fix(dashboard) | scrollable right panel, auto-expand steps |
| feat(workflows) | execution timeline debug view |
| fix(dashboard) | input params on failed runs, 0-step empty-state |
| fix | stray variables field in WorkflowRun tests |
| fix(dashboard) | missing rerunWorkflowRun API |
| fix | agent_send async + default model |
| feat(dashboard) | execution log console |
| fix(runtime) | channel_send guard for system channels |
| fix(dashboard) | React Hooks in WorkflowsPage |
| fix(kernel) | sender_chat_id in streaming path |

PR #6644, opened 2026-07-29
fix(api): reject a blank credential on auth refresh, and correct what the callback returns

houko:
Follow-up to #6639. It closes a hole that fix left open, and corrects a factual error in its documentation. Both are mine.

## The hole

`find_by_access_token` compared the caller's value against each stored entry by length and then `subtle::ConstantTimeEq`. `StoredTokens::access_token` is taken verbatim from the provider's token response, so an identity provider that returned an empty `access_token` leaves an entry whose value is `""` — and `""` matches `""` on both counts.

```
POST /api/auth/refresh
{"access_token": ""}
```

would then resolve that session and hand back its refresh token, having proved nothing. Narrow, because it needs a misbehaving IdP to have produced such an entry. But the premise of #6629 is that ownership must be *proven* rather than assumed, and an empty needle proves nothing — the primitive should not depend on every IdP being well-behaved.

Fixed at both layers:

- **Primitive**: `find_by_access_token` fails closed on an empty needle, so a future caller inherits the invariant rather than having to know about it.
- **Boundary**: the handler treats a field that is present but blank as absent. That also stops an empty `refresh_token` from fanning out a doomed request to the provider's token endpoint.

Filtering on the trimmed form while passing the value through untrimmed keeps the comparison an exact match against what the provider issued.

## The error

#6639's documentation claimed the callback "keeps the refresh token server-side, so a client legitimately has no refresh token of its own to present."

It does not. `CallbackResponse` returns the refresh token to the client deliberately, and carries a `SECURITY:` note explaining why. So `access_token` was never the only credential a legitimate caller could present — it is a convenience for a client that kept only that half, and `refresh_token` is the preferred path.

This matters beyond accuracy, because it changes which credential the docs steer callers toward. An access token travels on every request as a `Bearer` header, making it the more exposed of the two, and exchanging one for a refresh token extends a leaked short-lived credential past its own lifetime. The path is still offered — it is bounded by the store's 24 h TTL and does not widen what a holder of that access token can already do while it is live — but the documentation now states the preference and the trade rather than implying there is no choice.

## Also here

- The #6631 completeness guard now covers every **mutating** `/plugins/...` route rather than only `POST`, so a future `PUT` or `DELETE` cannot slip in unclassified.
  It also decides classification by calling `user_role_allows_request` — the real gate — instead of the `plugin_route_executes_plugin_code` predicate directly, because the predicate is consulted only for `POST`: asking it about a path in isolation would report a hypothetical `PUT /plugins/{name}/enable` as gated when the gate would in fact wave an Admin straight through.
- Two `ADMIN_ALLOWED` entries in that guard were stale — `POST /plugins/{name}/lint` and `POST /plugins/{name}/health`, both **GET-only** in the router, so the scan never matched them.
  A stale entry is fail-open: it silently pre-classifies whatever gets registered at that path next as Admin-safe, with no review.
  They are removed, `DELETE /plugins/{name}/state` is classified explicitly (it destroys persisted plugin state, runs no plugin code, and cannot make an unloadable hook loadable), and a new assertion requires every `ADMIN_ALLOWED` entry to be observed in the route table so the list cannot rot again.
- Integration coverage pins the unproven-request response to exactly **400** rather than "some 4xx", and documents which bodies used to succeed at selecting a session.

Both halves of the guard are mutation-tested: dropping an `ADMIN_ALLOWED` entry for a route Admin *can* reach fails the classification assertion, and adding one for a route that does not exist fails the staleness assertion.
Removing the empty-needle guard fails the new `find_by_access_token` test, so it is not vacuous either.

## Verification

- `cargo test -p librefang-api --lib` — 886 passed, including a case that seeds an entry with an empty stored access token and asserts an empty needle still resolves nothing. That test is the direct regression for the hole above.
- `cargo test -p librefang-api --test oauth_routes_test` — 25 passed.
- `openapi.json`, the SDKs, and the schema baseline regenerated: the request-body documentation is part of the published schema.

## Note on provenance

This work was drafted in a parallel session on the #6639 branch and was still uncommitted when that PR merged, so it did not reach `main`. I found it while cleaning up the worktree and verified both claims against merged `main` before rebuilding it here on top of `main`: the empty-string guard is genuinely absent there, and `CallbackResponse` genuinely returns the refresh token.

## Out of scope — flagged for a decision, not deferred silently

While re-reading #6630's breadth scan I found one more value-bearing field on the same route, and it is **not** a mechanical extension of that fix, so it needs a maintainer call rather than a guess.

`serialize_mcp_transport` (`routes/skills/mod.rs`) returns the stdio `command` and `args` verbatim on `GET /api/mcp/servers`, which is in `PUBLIC_ROUTES_DASHBOARD_READS` — the same route and the same unauthenticated-in-open-mode caller as the `env` disclosure #6630 closed.
An operator who wires a server as `npx some-mcp --api-key=sk-live-…` has that credential in `args`, and it is returned in the clear.

The reason this is not just "apply the same redaction": the `env` fix works because entries are `NAME` or `NAME=value`, so a redacted read can be merged back against the stored entry **by name**.
`args` are positional and have no key.
`McpServersPage` hydrates `args` from this response (line 184) and submits them back on save (line 157), so redacting them without a merge story wipes the command line on the next unrelated edit — the exact data-loss failure mode #6630 had to design around.
A positional merge is possible but fragile: an operator reordering or inserting an argument would silently re-associate a redacted slot with the wrong value.

Three shapes, none free:

1. Return `args` only to Owner, keep the name-only list public. Splits the response by role, which the route does not do today.
2. Redact and require the caller to re-supply the full command line on save. Honest, but a UX regression for every edit.
3. Leave it and document that inline secrets belong in `env`, not `args`, now that `env` is redacted.

Say which and I will implement it as its own issue/PR — I am not folding a response-shape decision into this one.

## Correction on this PR's own history

`aae55d5e` was committed from a working tree that was mid-mutation-test: the `POST /plugins/uninstall` entry had been removed to prove the classification assertion is not vacuous, and it did not get put back before the commit.
The first CI run on this branch therefore failed that test.
`76383def` restores the line, and the full suite is green locally as listed above.


crates/librefang-api/src/oauth.rs
```diff
@@ -1424,8 +1436,20 @@ pub async fn auth_refresh(
...
+    let supplied_access = req
+        .access_token
+        .as_deref()
+        .filter(|at| !at.trim().is_empty());
+
+    // Resolve the refresh token from whichever credential the caller proved.
+    let (refresh_token, stored_sub, provider) = if let Some(rt) = supplied_refresh {
         // Client supplied a refresh token explicitly.
         let provider = if let Some(ref pid) = req.provider {
             providers.iter().find(|p| p.id == *pid)
```
houko, line 1455: Unrelated to this PR's fix, but adjacent to code this diff touches (the `supplied_refresh` branch just above): when `req.provider` is `Some(pid)` and `pid` doesn't match any configured provider, `provider` here silently resolves to `None`, and the caller falls through to the generic `"Multiple providers configured; please specify 'provider' in the request"` 400 below — which is misleading when they *did* specify one, just an invalid/misspelled id. Pre-existing behavior, not introduced here, so flagging rather than fixing in this PR: worth a distinct "unknown provider '{pid}'" error message in a follow-up, your call whether that belongs in this PR or a separate one.

---

houko (commented):
Daily automated review pass on this PR: verified the core #6629 fix logic in `find_by_access_token` / `auth_refresh` looks correct (empty-needle guard, trim-without-mutating-the-comparison-value), and fixed a hard-wrapped-prose CLAUDE.md violation across a few of the new comments/doc-comments (rewrapped to one sentence per line, no content change, pushed as a separate commit). One pre-existing, out-of-scope nit left as an inline comment for your judgment call.

---

houko (commented):
Reviewed the auth-refresh fix and the plugin-classification hardening; logic checks out against #6629's threat model (fail-closed on the empty needle at the primitive, blank-as-absent at the handler, no new path that resolves an empty/whitespace credential). `openapi.json` and the sha256 baseline are consistent with the current source doc-comments.

One opinion on the out-of-scope item flagged in the PR description (`args` disclosure on `GET /api/mcp/servers`, in `crates/librefang-api/src/routes/skills/mod.rs`'s `serialize_mcp_transport` — not part of this diff, so I can't anchor an inline comment to it): Option 1 (Owner-only `args`, name-only list stays public) seems the best fit. It reuses the role-split pattern already used elsewhere in this file (`is_owner_only_write`, the privileged-GET gate), avoids inventing the positional-merge story option 2 needs (which the PR body already flags as fragile), and doesn't rely on operator discipline the way option 3 does. Not blocking this PR — recording a preference since one was asked for.

---

crates/librefang-api/src/oauth.rs
houko, line 1: Flagging for a maintainer call, not fixing directly (fixing it would mean rewriting the tip commit's metadata, which needs a force-push — outside what an automated pass should do unilaterally).

Commit `59f7bc61` ("docs(api): rewrap #6629/#6644 comments to sentence-per-line") — the one this file's prose-wrap cleanup landed in — has its **git author identity** set to `Claude <noreply@anthropic.com>`:

```
$ git log -1 --format='%an <%ae>' 59f7bc61
Claude <noreply@anthropic.com>
```

CLAUDE.md's rule is "No AI / Claude attribution in commit messages, PR bodies, or comments … because attribution pollutes `git log` and signals provenance the project does not want to imply." The commit *message* text here is clean, but the author field is a more permanent and more visible form of the same thing — it shows up in `git log`, `git blame`, and GitHub's commit-author UI regardless of message content.

It also slipped past both defense-in-depth layers: `.claude/hooks/guard-bash-safety.sh` only inspects the `-m` flag value, and `scripts/hooks/commit-msg` (`grep -qiE '(Co-Authored-By:...|Generated with .{0,40}Claude|🤖.*Claude...)'`) only scans message text — neither greps `%an`/`%ae`. An automated session committing with `GIT_AUTHOR_*` (or `--author`) set to a Claude identity bypasses both.

Two independent things worth a maintainer decision:
1. This PR's tip commit: amend the author to a human identity (or squash into the previous commit) before merge — your call, since rewriting it needs a force-push I'm not doing unilaterally.
2. The hooks: `commit-msg` (and/or `guard-bash-safety.sh`) arguably should also reject `%an`/`%ae` matching a Claude/Anthropic pattern, so this class of leak is caught at commit time in future sessions, not found after the fact in review.

---

houko:
Daily automated review pass — scope: CLAUDE.md compliance only. Otherwise clean: the fail-closed empty-needle guard in `find_by_access_token` and the `RefreshRequest` blank-field handling are correctly implemented, the `every_mutating_plugin_route_is_explicitly_classified` rework (method-qualified keys, staleness assertion) is sound, the changelog fragment format is correct (`(#6629) (#6639) (#6644) (@houko)`), and prose wrapping follows the one-sentence-per-line rule throughout.

One finding: the tip commit, `59f7bc61` ("docs(api): rewrap #6629/#6644 comments to sentence-per-line"), carries `Claude <noreply@anthropic.com>` as the git commit **author** identity, not just in message text. CLAUDE.md's "no AI attribution" rule and the repo's `commit-msg` hook both scan commit message text, so neither catches attribution living in the author field instead — the same class of gap flagged on #6504.

Not fixing this myself: unlike the interleaved-history case on #6504, this is the single tip commit with nothing stacked after it, so amending its author is mechanically trivial — but doing so changes the SHA and needs a force-push, which this review pass's own rules forbid regardless of how safe the rewrite would be. Maintainer call: amend-and-push (safe here since it's the last commit), or leave it — a squash-merge would discard individual commit authorship anyway.

---

PR #6661, opened 2026-07-30
fix(registry): let an operator stop the daemon overwriting the registry checkout

houko:
Refs #6636 (observation (a)). First of two follow-ups from that issue's observation list; (c), (d) and (f) are dashboard-side and land separately.

## The bug

`~/.librefang/registry/` is a git clone the sync fast-forwards with `git reset --hard origin/main` (`registry_sync.rs`, `git_clone_fallback`), so every local modification under it is destroyed.

That is not only a hand-edited file. `PUT /api/hands/{id}/manifest` writes a registry-shipped hand's edit back into `registry/hands/<id>/HAND.toml` — deliberately, because `scan_hands_dir` reads `registry/hands/` before `workspaces/` and drops duplicate ids, so an edit written anywhere else would lose to the registry copy on reload. The supported way to customise a built-in hand is therefore self-erasing.

**The clobbering does not come from boot.** `boot_with_config` passes the configured `cache_ttl_secs` (86400 by default) and normally skips the fetch. The background catalog task in `server.rs` passes a hard-coded `0` into `refresh_registry_checkout`, and `should_refresh` compares `age.as_secs() > cache_ttl_secs`, so any marker older than a second forces a refresh: the reset runs on the task's first tick — which is at daemon start, matching the report — and again every 24 h. Gating boot alone would have changed nothing. Both paths are gated, and the task's call site now records that it is the load-bearing one.

## Changes

- **`[registry] auto_sync`** (default `true`, so nothing changes unless set). Gates the two *automatic* refresh paths: boot's `sync_registry` pass and the periodic catalog task.
- **Explicit actions keep fetching.** `librefang init`, `POST /api/catalog/update` and `POST /api/hands/reload` are operator-initiated, so freezing the registry does not strand someone who wants an update. `POST /api/catalog/update` passes `true` rather than the configured value, stated at the call site.
- **The fetch is gated, the catalog rebuild is not.** `sync_catalog_to` takes `refresh_from_upstream` and skips only the network half, still reading `registry/providers/` off disk. Gating the rebuild too would leave an operator who froze the registry with an empty model catalog — a worse outcome than the clobbering they were avoiding.
- **Config surface.** `registry.` is already a writable section prefix, so the field is settable through `POST /api/config/set` with no allowlist change, and it is added to the `GET /api/config` read block: an operator who cannot read the freeze back has no way to confirm it took, which is the gap #6605 / #6618 closed elsewhere. The `registry` section stays classified `R` in `docs/operations/config-reload.md`; the row now records that the 24 h task re-reads config each tick, so flipping the knob off stops the next automatic refresh without a restart even though the section as a whole needs one.

Overlaps the existing `LIBREFANG_REGISTRY_OFFLINE`, which is scoped to every fetch in the process tree and lives in whatever launched the daemon rather than in `config.toml`. The new knob is narrower and durable; both remain.

## Verification

- `cargo test -p librefang-types --lib` — 891 passed.
- `cargo test -p librefang-runtime --lib catalog_sync` — 10 passed. New: `frozen_registry_still_rebuilds_the_catalog_from_disk`, which also proves the gated call makes no network attempt — the temp dir is not a git repo and has no sync marker, so an ungated run would try a git clone and then an HTTP download of the real registry.
- `cargo test -p librefang-api --lib` — 892 passed.
- `cargo test -p librefang-api --test config_routes_integration` — 44 passed. New: `registry_auto_sync_round_trips_through_get_and_set`, which sets the non-default so a hardcoded literal cannot pass, then writes through `POST /api/config/set` and reads it back.
- `cargo test -p librefang-kernel --lib` — 1357 passed.

## Not fixed here, found while doing it

**`kernel_config_schema.golden.json` is already stale on `main`.** `kernel_config_schema_matches_golden_fixture` self-ignores with "golden fixture out of date — see header comment to regen", and it does so on an unmodified checkout of `main`, not because of this PR. Regenerating it here produces a 7,278-line whole-file reshuffle of schemars output that nobody in this PR authored, and would flip a disabled guard back on carrying that unreviewed diff — so the fixture is left untouched. Worth its own issue: while the guard is off, any change to the derived `KernelConfig` schema ships without the reviewable diff the fixture exists to force.

**The real fix for hand customisation is an overlay, not a knob.** `auto_sync = false` stops the loss but leaves the underlying shape wrong: user edits live inside a checkout whose contract is "mirror of upstream". The durable fix is to write them to a user-owned location that wins on load — which means flipping `scan_hands_dir` precedence to `workspaces/`-first and answering what happens to an install that already has both files. That is a behaviour change for every existing install with a silent failure mode (a stale `workspaces/<id>/HAND.toml` from a past marketplace install suddenly winning), and `update_manifest_persisted` runs a supply-chain audit before writing whose coverage would need re-checking against the new target. Deliberately not bundled here.


crates/librefang-types/src/config/types.rs
```diff
@@ -5566,6 +5566,16 @@ pub struct RegistryConfig {
     /// Set to a full base URL such as `"https://codeberg.org"` to consume a Codeberg-hosted registry; the tarball, git-clone, and tarball-prefix values are all derived from it.
     #[serde(default)]
     pub registry_host: Option<String>,
+    /// Whether the daemon refreshes `~/.librefang/registry/` from upstream on its own (default: `true`).
+    ///
+    /// That checkout is a git clone the sync fast-forwards with `git reset --hard origin/main`, so every local modification under it is destroyed — including the ones `PUT /api/hands/{id}/manifest` writes, which land in `registry/hands/<id>/HAND.toml` for a hand that shipped with the registry.
+    /// Set to `false` to freeze the checkout: boot skips its `sync_registry` pass and the periodic catalog task skips the network refresh while still rebuilding the model catalog from what is already on disk.
+    ///
+    /// Only *automatic* refreshes are gated. `librefang init`, `POST /api/catalog/update`, and `POST /api/hands/reload` are explicit operator actions and still sync, so freezing the registry does not strand an operator who wants an update.
```
houko: `POST /api/hands/reload` doesn't actually fetch from upstream, so grouping it with `librefang init` / `POST /api/catalog/update` as an "explicit action that still syncs" reads as inaccurate.

`reload_hands` → `hand_registry.reload_from_disk(home_dir)` → `scan_hands_dir(home_dir)` is a pure local filesystem re-scan (`crates/librefang-hands/src/registry.rs:622-641`); it never touches the registry git checkout or does any network I/O, before or after this PR. So the "still sync" framing here (and the identical phrasing repeated in the changelog fragment, the PR body, and the `boot.rs`/comment) overstates what that endpoint does — it doesn't need `auto_sync` gating not because it's an "explicit fetch" but because it never fetches at all.

Not asking for a revert — the config field/gating logic itself is correct — just flagging that the doc comment (and the changelog line/PR body making the same claim) describes `/api/hands/reload` incorrectly and may mislead an operator who reads it as "this endpoint will pull the latest registry for me."

---

houko (commented):
Automated review pass against CLAUDE.md conventions. The `[registry] auto_sync` config field is wired correctly (struct field, `#[serde(default = "default_true")]`, `Default` impl entry, Serialize/Deserialize derives all present; `docs/operations/config-reload.md` classification matches `build_reload_plan`'s `R` for the `registry` section, with an accurate footnote on the per-tick live-read exception; changelog fragment is correctly located/formatted and references the right PR number). Pushed one small fix (comment formatting only, no logic change) and left one inline note below for a documentation-accuracy judgment call.

---

houko (commented):
Automated review pass — one thing flagged for human judgment, plus a status note.

**Commit-author attribution (needs your action, not mine):** the third commit on this branch, `f04b17600f657986fdcdce3e6769aad4e36e9a3b` ("docs(registry): rewrap new auto_sync comments to one sentence per line"), has its git author identity set to `Claude <noreply@anthropic.com>`:

```
$ git log --format='%H %an %ae %s' -1 f04b176
f04b17600f657986fdcdce3e6769aad4e36e9a3b Claude noreply@anthropic.com docs(registry): rewrap new auto_sync comments to one sentence per line
```

CLAUDE.md's "No Claude / Anthropic / AI attribution" rule covers commit messages/PR bodies/comments; the `commit-msg` hook checks message *text* (and clearly passed here, since the body text itself contains no attribution string — it only references the `CLAUDE.md` filename), but doesn't inspect the author field, so this slipped through the hook. I didn't rewrite it myself: it's HEAD of an already-reviewed branch, and CLAUDE.md's own guidance is to prefer fixup commits over rewriting history once a reviewer has loaded the diff (`houko`'s review thread on `types.rs:5574` predates this). Recommend `git commit --amend --author="<your name/email>"` (or equivalent) and a force-push to your own branch to correct the author identity — that action is yours to take, not mine.

**Everything else checked out clean** against CLAUDE.md's codified rules: `[registry] auto_sync` has struct field + `#[serde(default = "default_true")]` + `Default` impl entry + the derive quartet, all present together; the `config-reload.md` row edit matches `build_reload_plan`'s `restart_if_changed(... "registry")` classification and the field-name token the drift-guard test (`doc_reload_table_matches_classified_reload_fields`) parses is unaffected by the added prose; `registry.` as a writable section prefix carries no credential-bearing fields (`cache_ttl_secs`, `registry_mirror`, `registry_host`, `auto_sync` — none are secrets); the changelog fragment format, all five test-scoping commands (`-p <crate>` throughout), and conventional-commit format on all three commits are all correct. I also pushed one mechanical fix (`c002566`): a new test doc comment in `catalog_sync.rs` (`frozen_registry_still_rebuilds_the_catalog_from_disk`) was column-wrapped mid-sentence — the same class of issue the third commit already fixed elsewhere in this file, just missed there. Rewrapped to one sentence per line.

Not re-raising `houko`'s own open thread about the `/api/hands/reload` phrasing (accurate as filed, still unresolved) or the stale `kernel_config_schema.golden.json` (explicitly out-of-scope here per the PR body, covered by #6664).

---

changelog.d/fixed/6661-registry-auto-sync.md
```diff
@@ -0,0 +1,6 @@
...
+  `~/.librefang/registry/` is a git clone the sync fast-forwards with `git reset --hard origin/main`, so every local modification under it is destroyed — including the ones `PUT /api/hands/{id}/manifest` writes, which land in `registry/hands/<id>/HAND.toml` whenever the hand shipped with the registry, making the supported way to customise a built-in hand self-erasing.
+  Reported as "local edits are overwritten on daemon start, even with the file set read-only, and no config option disables the registry sync".
+  The culprit was not boot: `boot_with_config` honours `cache_ttl_secs` (86400 by default) and normally skips, while the background catalog task passes a hard-coded TTL of `0` into `refresh_registry_checkout`, which makes `should_refresh` true for any marker older than a second and forces the reset on its first tick and every 24 h after — so gating boot alone would have changed nothing.
+  Setting `auto_sync = false` freezes the checkout on both automatic paths and leaves the explicit ones (`librefang init`, `POST /api/catalog/update`, `POST /api/hands/reload`) fetching as before, so freezing does not strand an operator who wants an update.
```
houko: `POST /api/hands/reload` doesn't actually sync from upstream.
`reload_hands()` calls `hand_registry.reload_from_disk()` (crates/librefang-hands/src/registry.rs:622), which only re-parses whatever `HAND.toml` files are already on disk via `scan_hands_dir`.
It never touches `registry_sync`/`sync_catalog_to`, and its own route doc agrees: "Reload hand definitions from disk" (crates/librefang-api/src/routes/skills/hands.rs:1248).
So grouping it with `librefang init` and `POST /api/catalog/update` under "explicit operator actions ... still sync" is misleading: an operator who froze the registry (`auto_sync = false`) and then calls hand reload gets the same frozen files re-parsed, not fresh content from upstream.
The same claim also appears in crates/librefang-types/src/config/types.rs:5574 and crates/librefang-kernel/src/kernel/boot.rs:1038 and 1049 — the boot log even tells the operator to run it as a way to "refresh on demand."
Worth rewording all three occurrences, since this fragment ships verbatim into the GitHub release notes.

---

docs/operations/config-reload.md
```diff
@@ -184,7 +184,7 @@ classified differently — the row note spells out which is which.
 | `a2a` | H | Agent-to-Agent protocol config. |
 | `skills` | H | Skills config (bundled + user-installed) — reloads registry. |
 | `plugins` | R | Plugin registry config. |
-| `registry` | R | Registry sync config (cache TTL, …). |
+| `registry` | R | Registry sync config (cache TTL, mirror, host, `auto_sync`). `auto_sync` is the exception the R classification understates: the 24 h catalog task re-reads the config each tick, so flipping it off stops the next automatic refresh without a restart — but boot's own sync pass has already run by then, so a restart is still what makes the whole section's behaviour match the file. |
```
houko, line 187: This note claims flipping `auto_sync` off "stops the next automatic refresh without a restart," but that doesn't match `should_store_config` (crates/librefang-kernel/src/config_reload.rs:1147-1154): it only stores the reloaded config when `plan.hot_actions` or `plan.noop_changes` is non-empty.
`registry` as a whole is only ever pushed onto `plan.restart_required` (config_reload.rs:703, via the `restart_if_changed` closure defined at line 625) — no hot action or noop entry is registered for it.
So if `auto_sync` is the only field an operator changes and they call `POST /api/config/reload`, `should_store_config` returns `false`, `self.config` is never swapped, and the periodic catalog task's `kernel.config_snapshot()` (server.rs:2211, inside the loop) keeps returning the stale value indefinitely — the opposite of what this note says.
The claim only holds if the same reload also touches an unrelated hot-reloadable/noop field in the same edit.
Worth re-verifying against `should_store_config`'s actual gating (maybe with a reload-plan unit test) before merging this note.

---

crates/librefang-kernel/src/kernel/boot.rs
```diff
@@ -1034,12 +1034,21 @@ impl LibreFangKernel {
...
         // Skips download if cache is fresh; skips copy if files already exist.
-        librefang_runtime::registry_sync::sync_registry(
-            &config.home_dir,
-            config.registry.cache_ttl_secs,
-            &config.registry.registry_mirror,
-            config.registry.registry_host.as_deref(),
-        );
+        // `[registry] auto_sync = false` freezes `~/.librefang/registry/`: the sync fast-forwards that checkout with `git reset --hard origin/main`, which destroys every local modification under it — including the ones `PUT /api/hands/{id}/manifest` writes for a registry-shipped hand.
+        // Explicit operator actions (`librefang init`, `POST /api/catalog/update`, `POST /api/hands/reload`) still sync.
+        if config.registry.auto_sync {
```
houko, line 1039: The headline behavior this PR fixes — the daemon no longer running `git reset --hard origin/main` over local registry edits at boot when `auto_sync = false` — doesn't have a test exercising this boot path directly.
The two new tests (`frozen_registry_still_rebuilds_the_catalog_from_disk` in catalog_sync.rs and `registry_auto_sync_round_trips_through_get_and_set` in config_routes_integration.rs) cover the periodic-task fetch gate and the config get/set surface, but neither calls `LibreFangKernel::boot_with_config` to confirm `sync_registry` itself is skipped here — which is the exact scenario from the bug report ("overwritten on daemon start").
Worth adding a boot-level regression test (e.g. a temp home dir with a stale/absent registry checkout, asserting `sync_registry`'s clobber doesn't run when `auto_sync = false`) before this ships as the fix for that report.

---

houko (commented):
Automated review pass against CLAUDE.md conventions (config-field rules, hot-reload classification, changelog fragment rules, integration-test coverage). The mechanics of this PR are solid — `#[serde(default)]` + Default impl entry present, changelog fragment correctly formatted, and the config get/set round-trip is well tested — but three of the PR's own explanatory claims don't hold up against the code they describe. Left as line comments since each needs a judgment call on the right wording/fix rather than a mechanical correction.

---

crates/librefang-kernel/src/kernel/boot.rs
```diff
@@ -1034,12 +1034,21 @@ impl LibreFangKernel {
         // Auto-sync registry content on first boot or after upgrade when
         // Sync registry: downloads if cache is stale, pre-installs providers/agents/integrations.
         // Skips download if cache is fresh; skips copy if files already exist.
-        librefang_runtime::registry_sync::sync_registry(
-            &config.home_dir,
-            config.registry.cache_ttl_secs,
-            &config.registry.registry_mirror,
-            config.registry.registry_host.as_deref(),
-        );
+        // `[registry] auto_sync = false` freezes `~/.librefang/registry/`: the sync fast-forwards that checkout with `git reset --hard origin/main`, which destroys every local modification under it — including the ones `PUT /api/hands/{id}/manifest` writes for a registry-shipped hand.
```
houko, line 1037: Not about this line's content — about the commit that last touched it. Two commits on this branch, `f04b176` ("docs(registry): rewrap new auto_sync comments…") and `c002566` ("docs(runtime): rewrap frozen_registry test doc comment…"), are authored/committed as `Claude <noreply@anthropic.com>` and show up on GitHub as authored by the `claude` bot account (id 81847). CLAUDE.md's "No AI / Claude attribution" rule isn't scoped to message text only — a commit whose author identity is literally `Claude` carries the same attribution the rule exists to prevent, visible in `git log`/blame and the PR's commit list regardless of what the message says.

Fixing this cleanly requires re-authoring those two commits (interactive rebase / `git commit --amend --author`), which changes their SHAs and needs a force-push to this already-pushed branch — outside what an automated reviewer should do unilaterally on someone else's PR. Flagging so @houko (or whoever drives this branch) can rebase with `-c user.name=... -c user.email=...` (or squash them into the preceding commit) before merge.

---

houko (commented):
Reviewed against CLAUDE.md. Config wiring, changelog fragment, prose-wrap rule, and call-site plumbing for `[registry] auto_sync` all check out; pushed one small mechanical fix (see below) and flagged one thing that needs the PR author's own git surgery.

---

houko:
Reviewed the `auto_sync` change. The config field is wired correctly (struct field + `#[serde(default = "default_true")]` + `Default` impl entry per CLAUDE.md's config-field ritual), the default (`true`) preserves today's behavior, and `sync_catalog_to`'s fetch/rebuild split is sound — confirmed by `frozen_registry_still_rebuilds_the_catalog_from_disk` and `registry_auto_sync_round_trips_through_get_and_set`.

One inaccuracy found and fixed in `91fcc9d`: the doc comment on `RegistryConfig::auto_sync`, the `boot.rs` gating comment, the boot-time `info!` log, and the changelog fragment all listed `POST /api/hands/reload` alongside `librefang init` / `POST /api/catalog/update` as an action that "still syncs" when `auto_sync = false`. It doesn't — `reload_hands` (`crates/librefang-kernel/src/kernel/hands_lifecycle.rs:736`) only calls `HandRegistry::reload_from_disk`, which rescans the local `registry/hands/` directory into memory; it never fetches from upstream, `auto_sync` or not. An operator who froze the registry and then hit "reload" expecting a refresh would get nothing and no indication why. Corrected the wording in:
- `crates/librefang-types/src/config/types.rs` (doc comment)
- `crates/librefang-kernel/src/kernel/boot.rs` (gating comment + the `info!` log's suggested command)
- `changelog.d/fixed/6661-registry-auto-sync.md`

Also flagging as a separate review comment (not fixed, needs a judgment call): there's no test that exercises the actual claim in the PR title — that the registry checkout survives a daemon *restart* with `auto_sync = false`. The existing tests cover the config round-trip (API) and the catalog-rebuild-without-fetch unit test (`catalog_sync.rs`), but nothing exercises `boot.rs`'s own `if config.registry.auto_sync { sync_registry(...) }` branch against a real git checkout across two boots. `boot()` has no existing test scaffolding to hang that off of, so I left it as a comment rather than bolting on a new kernel-boot test harness unilaterally.

---

crates/librefang-kernel/src/kernel/boot.rs
houko, line 1: No test exercises this branch directly (the PR's stated goal — "let an operator stop the daemon overwriting the registry checkout" — is a restart-survival claim, and this `if config.registry.auto_sync { sync_registry(...) }` gate is where boot actually honors it).

What's covered: the config round-trips through GET/POST `/api/config` (`registry_auto_sync_round_trips_through_get_and_set`), and `sync_catalog_to`'s fetch-vs-rebuild split is unit-tested (`frozen_registry_still_rebuilds_the_catalog_from_disk`). Neither touches this `boot.rs` branch or a real git checkout across two boot cycles, so nothing currently proves that setting `auto_sync = false`, restarting the daemon, and finding a local edit under `~/.librefang/registry/` still intact actually works end-to-end — only that the *pieces* it's built from behave correctly in isolation.

`boot()` itself has no existing test scaffolding (no `#[test]`/`#[tokio::test]` in this file), so closing this gap means either extracting the gated call into a small unit-testable helper, or standing up a first kernel-boot-level test fixture — a bigger design decision than a comment should make unilaterally. Flagging for a maintainer call: is a boot.rs-level regression test warranted here, or is the existing catalog_sync + config-route coverage considered sufficient given `sync_registry`'s git-subprocess dependency?

---

houko, line 1: The comment here (and the matching claim in the PR description) says `POST /api/hands/reload` is one of the "explicit operator actions" that keeps fetching when `auto_sync = false`.
Tracing it: `reload_hands` → `HandRegistry::reload_from_disk` → `scan_hands_dir` only rereads `registry/hands/` and `workspaces/` off local disk (`crates/librefang-hands/src/registry.rs:622`) — it never touches the network.
So an operator who freezes the registry and then calls `/api/hands/reload` expecting a refresh will just re-read the still-frozen local checkout, not pull anything new; only `librefang init` and `POST /api/catalog/update` actually invoke `refresh_registry_checkout`/`sync_registry`.
Worth either correcting the comment/PR narrative (drop `/api/hands/reload` from the "still sync" list) or, if the intent really was for hand-reload to also refresh, wiring that in — as written it may lead an operator to believe hitting that endpoint gets them upstream changes while the registry is frozen.

---

houko, line 1: Nothing in this PR's test additions exercises this new `if config.registry.auto_sync { sync_registry(...) } else { ... }` branch itself.
`registry_auto_sync_round_trips_through_get_and_set` covers the config surface and `frozen_registry_still_rebuilds_the_catalog_from_disk` covers `sync_catalog_to`'s fetch gate, but neither touches `boot_with_config`'s own gate added here.
The PR body itself argues boot's `cache_ttl_secs` usually makes this a no-op already, which is exactly the kind of reasoning that benefits from a regression test (e.g. asserting `sync_registry` is/isn't invoked — or that no git operation runs against a fixture home dir — for `auto_sync = false` vs `true`) rather than resting on that argument alone.

---

houko (commented):
Mechanically, the diff is solid: `auto_sync` has struct field + Default impl entry, a compliant changelog.d/fixed/ fragment, and integration test coverage for both the config round-trip and the catalog_sync fetch gate. Two smaller correctness/coverage notes left inline — nothing that looks build-breaking or contradicts the per-agent config-file rule (this PR is process-config, config.toml, not the HAND.toml per-agent surface).

---

houko (commented):
Reviewed the current diff against `origin/main` and against CLAUDE.md's config-field ritual, changelog-fragment format, prose-wrapping rule, and hot-reload classification guidance. The `[registry] auto_sync` field itself is wired correctly (struct field + `#[serde(default = "default_true")]` + `Default` impl entry), the changelog fragment at `changelog.d/fixed/6661-registry-auto-sync.md` matches the required format (no leading `- `, sentence-per-line, ends `(#6661) (@houko)`), and the new/edited doc comments in `types.rs`, `boot.rs`, and `catalog_sync.rs` are already sentence-per-line.

One remaining inconsistency not covered by the existing review threads: the PR description's "Changes" section still states `POST /api/hands/reload` is one of the "explicit actions" that "keeps fetching" when `auto_sync = false` ("`librefang init`, `POST /api/catalog/update` and `POST /api/hands/reload` are operator-initiated, so freezing the registry does not strand someone who wants an update"). That claim was identified as false and corrected in the code itself by commit `91fcc9d` (which updated `crates/librefang-types/src/config/types.rs`, `crates/librefang-kernel/src/kernel/boot.rs`, and the changelog fragment to drop `/api/hands/reload` from that list, since `reload_hands` only calls `HandRegistry::reload_from_disk` / `scan_hands_dir` — a local rescan with no network I/O). The PR body itself was not updated to match, so it still tells a reviewer the opposite of what the shipped code and changelog now say. Worth a quick edit to the description before merge so it doesn't contradict its own fix commit.

Everything else substantive I found (the boot-level restart-survival test gap, the `docs/operations/config-reload.md` hot-reload-classification claim, and the AI-authorship on `91fcc9d`/`4336377`/`c002566`/`f04b176`) is already raised in open threads on this PR, so not repeating those here.

---

Issue #6654, opened 2026-07-29
fix(goals): list_goals silently returns empty array on storage failure


DaBlitzStein:
## Bug
`list_goals` swallows storage errors and returns HTTP 200 with \[\] instead of 500. Operator sees empty goal list when data loss occurred.

**File:** `goals.rs:61-63`
**Fix:** Return 500 on storage failure.
**Found by:** Goal system audit 2026-07-29

houko:
Fixed in 6eaa4511 (#6662).

`list_goals` now returns a scrubbed 500 on a substrate read failure instead of folding the error into the same empty array it returns for "nothing created yet". An absent or non-array key still answers `200 {"items": [], "total": 0}`, since that is the genuine not-yet-created shape.

Two more instances of the same class were found in the same file and fixed with it: `POST /api/goals/{id}/start` reported an unreadable store as `404 Goal not found`, and its `in_progress` flip discarded its `Result` with `let _ =` (now logged — that one is deliberately non-fatal, since the run starts regardless and its live phase comes from the runner's own state).

The dashboard half matters too: `GoalsPage` had no `isError` branch, so returning a 500 alone would have moved the swallow one layer up — the query yields no goals and the page draws its template picker over data it failed to read. It now renders `ErrorState` with a retry.

Regression test: `goals_list_returns_500_on_storage_failure_6654`.

This issue stayed open because the PR body wrote `Closes #6650, #6652, #6653, #6654` — GitHub only honours the first reference in that comma-joined form.

Issue #6653, opened 2026-07-29
fix(goals): get_goal_children returns 200 with leaked error on storage failure


DaBlitzStein:
## Bug
`get_goal_children` returns HTTP 200 with `{"error": format!("{e}")` on storage failure, leaking internal implementation details.

**File:** `goals.rs:122-123`
**Fix:** Use `ApiErrorResponse::internal_scrub(e)` returning 500.
**Found by:** Goal system audit 2026-07-29

houko:
Fixed in 6eaa4511 (#6662).

`get_goal_children` was wrong in both directions at once — a 200 carrying an empty `children` array *and* a raw `format!("{e}")`, so a client checking the status code saw success while the body handed out the SQLite path and error chain. It now returns the scrubbed 500 envelope that `get_goal`, directly between it and `list_goals`, already used.

Regression test: `goals_children_returns_scrubbed_500_on_storage_failure_6653`, which also scans the whole serialized body for leaked internals (`serialization`, `sqlite`, `kv_store`, the corrupt bytes) rather than asserting on one field.

This issue stayed open because the PR body wrote `Closes #6650, #6652, #6653, #6654` — GitHub only honours the first reference in that comma-joined form.

Issue #6652, opened 2026-07-29
fix(goals): cross-session dedupe shares async task registry entries


DaBlitzStein:
## Bug
Async task registry deduplicates by `(target_agent, prompt_hash)` across ALL sessions. Two sessions sending the same message to the same agent share one registry entry; only the first registrant gets the callback.

**File:** `task_registry.rs:183-207`
**Fix:** Include session_id in the dedupe key.
**Found by:** Goal system audit 2026-07-29

houko:
Fixed in 6eaa4511 (#6662), but on a different axis than this issue proposes — worth recording why.

The suggested fix, adding `session_id` to the registry dedupe key, would reverse a deliberate #5033 decision that `register_async_task`'s own docstring records across two paragraphs, and would break its pinned test `register_dedupe_is_cross_session_for_delegation_kind`.

That same docstring already prescribes the correct axis: *"Callers that need per-session isolation must salt their `prompt_hash` (e.g. include the calling session id in the hash input) — the kernel does not silently distinguish them."* The kernel's only production caller was not following its own contract — `send_to_agent_async_tracked` hashed `message` alone, making the hash a pure function of the prompt.

So the salt moved to the caller (`delegation_prompt_hash`), covering the caller agent, the caller session **and** the conversation key. The first two are a superset of the registry's `(agent_id, session_id)` delivery key, so two distinct delivery targets can never share an entry; `conversation_key` is included because it selects a different callee session (`SessionId::for_channel(agent, "agent_send:{key}")`) and therefore denotes a genuinely different operation. What still dedupes is exactly the idempotency case #5033 wanted: one caller re-sending the same message on the same conversation while the first is in flight.

The kind-only match key and its pinned test are untouched.

Regression tests: `delegation_hash_separates_callers_sending_the_same_message`, `delegation_hash_separates_conversations`, `delegation_hash_is_stable_for_the_same_caller_and_conversation`.

This issue stayed open because the PR body wrote `Closes #6650, #6652, #6653, #6654` — GitHub only honours the first reference in that comma-joined form.

Issue #6651, opened 2026-07-29
fix(goals): completion events silently dropped when kernel self_handle unavailable


DaBlitzStein:
## Bug
When `spawn_wake_idle_turn` tries to upgrade the kernel self_handle and fails (shutdown, boot race), the completion event is silently dropped. The registry entry was already consumed (delete-on-delivery), so there is no replay.

**File:** `task_registry.rs:447-457`
**Impact:** Lost async callbacks during daemon shutdown.
**Found by:** Goal system audit 2026-07-29

houko:
Checked this against `main` (`469634ec`) and it does not reproduce — recommending closure as not-a-bug, with the evidence below. Line numbers are from that commit and I re-checked each quote against it.

**"Silently" is not accurate.** The branch at `crates/librefang-kernel/src/kernel/task_registry.rs:447-457` already logs at `WARN` and names the consequence:

```
tracing::warn!(
    agent_id = %agent_id,
    session_id = %session_id,
    "Async task wake-idle: kernel self-handle not yet initialised; dropping completion"
);
```

The behaviour is also the documented contract, at `task_registry.rs:420-425`: *"`false` if the kernel self-handle has not been initialised yet (boot-time race; the entry has already been consumed from the registry per the delete-on-delivery contract, so the event is dropped)."*

**Neither stated trigger is reachable.**

- *Shutdown*: all three production callers hold a **strong** `Arc` across the `.await` (`workflow_runner.rs:450`, `handles/agent_control.rs:265`, `librefang-runtime/src/tool_runner/process.rs:34`). `Weak::upgrade()` returns `None` only once every strong reference is gone, so "kernel dropped" and "a caller is inside `complete_async_task`" are mutually exclusive.
- *Boot race*: both registration paths propagate a failed upgrade with `?` before any completion is scheduled (`workflow_runner.rs:334-342`, `handles/agent_control.rs:200-208`), and `self_handle` is a `OnceLock` that never reverts. An entry that reaches the wake-idle branch is one whose registration already ran past that gate.

[...]

PR #6662, opened 2026-07-30
fix(goals,kernel): surface goals storage failures, and stop two async delegations from being silently mislabelled

### LINKED ISSUE #6650: fix(goals): async agent_send silently falls back to blocking when caller_session_id missing
## Bug
When `caller_session_id` is `None` or unparseable, `send_to_agent_async_tracked` silently falls back to a blocking send. No task is registered, no callback will ever arrive. The caller receives the response inline but thinks it was async.

**File:** `agent_control.rs:156-175`
**Impact:** For channel-based agents (Telegram/Slack), async agent_send degrades to blocking without warning.
**Found by:** Goal system audit 2026-07-29

houko:
Closes #6650, #6652, #6653, #6654.

Triages the seven open `bug` issues, three of which turned out to be wrong about the code on `main`. Each was verified against the source before any change, then the verdict was independently re-checked by a second pass that tried to refute it. Two issues get no code change and a recommendation instead; the verdicts and their evidence are below.

## Fixed

**#6654 / #6653 — goals storage failures were reported as success** (`crates/librefang-api/src/routes/goals.rs`)

`GET /api/goals` folded a substrate read error into the same empty array it returns when nothing has been created yet. A corrupt blob or an unreadable SQLite file reached the operator as `200 {"items": [], "total": 0}` — indistinguishable on the wire from a daemon with no goals.

`GET /api/goals/{id}/children` was wrong in both directions at once: a 200 carrying an empty list *and* a raw `format!("{e}")`, so a client checking the status code saw success while the body handed out the SQLite path and the error chain.

Both now return the scrubbed 500 envelope that `get_goal`, directly between them, already used. `Ok(_)` — key absent, or present but not an array — stays a 200 empty result, because that is the genuine not-yet-created shape.

Two more instances of the same class in the same file, found while fixing those two and fixed with them:

- `POST /api/goals/{id}/start` had the identical swallow on a write path. Its catch-all `_ => Vec::new()` turned an unreadable store into `404 Goal '<id>' not found`, sending the operator to re-create a goal that exists. Now a 500.
- The `in_progress` flip in the same handler discarded its `Result` with `let _ =`. That one is deliberately non-fatal and now says so: the run starts regardless and its live phase is served from the runner's own state by `GET /api/goals/{id}/run`, so a failure costs a stale `status` field, not a lost run. Failing the request would report "could not start" for a run that did start. It is logged rather than discarded.

`create_goal` / `update_goal_by_id` / `delete_goal` were checked in the same pass and already propagate correctly — no change.

**The dashboard half.** `GoalsPage` had no `isError` branch: `isLoading ? skeleton : goals.length === 0 ? template-picker : list`. Returning a 500 alone would have moved the swallow one layer up — the query would yield no goals and the page would draw "pick a template" over data it had failed to read. It now renders `ErrorState` with a retry, matching `ApprovalsPage`. New `goals.loadError` key in all four locales.

**#6652 — cross-session delegation dedupe** (`crates/librefang-kernel/src/kernel/handles/agent_control.rs`)

`register_async_task` dedupes delegations on `(target agent, prompt_hash)` and deliberately ignores the caller — a #5033 decision its docstring records across two paragraphs, along with the instruction that *"callers that need per-session isolation must salt their `prompt_hash`"*. The kernel's own only production caller did not: it hashed the message text alone.

So two independent agents asking the same agent the same question collided on one registry entry. The second received the first's handle, the completion event was delivered only to the first caller's `(agent, session)`, and the second — already told "delegation started asynchronously; do not wait" — waited for a reply that would never arrive.

Fixed caller-side, which is the axis the contract prescribes, **not** by adding `session_id` to the registry match key as the issue suggests. That would reverse the #5033 decision and break its pinned test `register_dedupe_is_cross_session_for_delegation_kind`, which registers directly against the registry and is untouched here.

The salt (`delegation_prompt_hash`) covers the caller agent, the caller session and the conversation key as well as the message: a superset of the registry's `(agent, session)` delivery key, so two distinct delivery targets can never share an entry, plus the field that selects a different callee session (`SessionId::for_channel(agent, "agent_send:{key}")`). What still dedupes is exactly the intended idempotency case — one caller re-sending the same message on the same conversation while the first is in flight.

**#6650 — untrackable `agent_send` reported as a delegation** (`librefang-kernel-handle`, `librefang-kernel`, `librefang-runtime`)

`send_to_agent_async_tracked` has two legitimate outcomes that mean opposite things — a task id on the tracked path, the callee's whole response body when no parseable caller session made tracking possible — and both arrived through one undiscriminated `Result<String, _>`. The tool that consumes it could not tell them apart, so it labelled the response body `task_id`, set `status: "delegated"`, and told the model to end its turn and wait for a completion event that no registered task would ever produce. The answer was in hand and thrown away.

The fallback itself is correct and stays: with nowhere to route a completion event, an inline answer beats an orphaned delegation. Only the conflation was the bug. It now returns `AsyncSendOutcome::{Tracked, Inline}`, and `tool_agent_send` branches on it — a tracked delegation still renders its task id and note verbatim, an inline one returns the reply exactly as the blocking path does.

Two things that came with it:

- The `agent_send` tool schema advertised *"returns immediately with a task_id"* unconditionally, which is false on the `Inline` path. Amended to say tracking needs a caller session and to read what you receive.
- The fallback logged at `debug!`. Now `warn!` with the caller fields attached — an operator whose agents silently lose async delegation should not have to raise the log level to find out.

Note the issue's stated impact is wrong and a reviewer should not try to reconcile against it. Channel agents are **not** affected: the in-process agent loop always supplies a real session (`ctx.session.id` is a `SessionId`, not an `Option`), and channel-originated turns route through it like any other. The live path is the `claude-code` subprocess driver via the MCP HTTP bridge (`routes/network.rs:1546` passes `session_id: None` by construction, and no session header exists to forward). The REST `/api/tools/{name}` bridge reaches it too but is opt-in and off by default.

## Not a bug — no code change

**#6651 — "completion events silently dropped"**: does not reproduce. The branch already logs at `WARN` and the behaviour is its documented contract. Neither stated trigger is reachable: all three production callers hold a strong `Arc` across the `.await`, so `Weak::upgrade()` cannot fail during shutdown, and both registration paths propagate a failed upgrade with `?` before any completion is scheduled. The ordering observation is correct — the entry is consumed ~115 lines before the delivery attempt — but the proposed fix would break the idempotency-of-completion contract, open a double-delivery window across the injection `.await` under a `parking_lot::Mutex`, and still not guarantee delivery, since `spawn_wake_idle_turn` is fire-and-forget. Three existing tests pin the current behaviour. Detail posted on the issue; recommending closure rather than closing someone else's issue.

**#6649 — "auto-spawned agents not killed on goal completion"**: not on `main`. `grep -rn 'spawn_sub_agent' crates/` returns nothing, and the cited `goal_runner.rs:861-898` falls inside `#[cfg(test)] mod tests` (opens at line 694). The machinery exists only on the unmerged PR #6505. The leak is real *there* and larger than the issue states — `stop_locked` calls `task.abort()` so the tail of `run_loop` never runs on an operator stop, and there is a second un-reaped spawner in `routes/goals.rs`. Both written up on the issue for that PR's author.

## Already handled

**#6636 — hand settings never reach the system prompt**: fixed by #6637, which merged while this branch was in progress. This branch is rebased on top of it; not touched here.

## Verification

Run locally against this branch, rebased on `469634ec`:

- `cargo clippy --workspace --all-targets -- -D warnings` — exit 0, no warnings. `--all-targets` is what compiles the new `mod tests` in `handles/agent_control.rs`, the `rusqlite` dev-dependency, and every `AgentControl` impl in a `tests/` tree, so it is the check that covers the trait signature change end to end.
- `cargo test -p librefang-api --test goals_routes_integration` — exit 0, 39 passed / 0 failed, including all three new storage-failure cases.
- `cargo test -p librefang-kernel --lib` — exit 0, 1360 passed / 0 failed / 1 ignored, including the three new `delegation_hash_*` cases.
- `cargo test -p librefang-runtime --lib` — exit 0, 2148 passed / 0 failed / 2 ignored. `agent_send_async_routes_to_tracked_path_and_returns_task_id` still passes unchanged, which is the check that the `Tracked` branch renders the identical JSON it did before.
- Dashboard `pnpm vitest run` — 939 tests / 91 files pass (exit 0), plus `pnpm typecheck` and `pnpm lint` clean.
- Negative control on the two new dashboard tests: removing the `isError` branch from `GoalsPage` makes both fail, and only those two.

New tests:

- `goals_list_returns_500_on_storage_failure_6654`, `goals_children_returns_scrubbed_500_on_storage_failure_6653`, `goal_run_start_returns_500_not_404_on_storage_failure` — a `corrupt_goals_blob` helper writes a non-JSON blob straight through the connection pool, which is the only way to make `structured_get` fail from outside `librefang-memory`. The children test also scans the whole body for leaked internals.
- `agent_send_async_without_session_returns_the_reply_not_a_fake_task_id_6650`. The `DispatchCapture` stub now mirrors the kernel's own discrimination — a caller session yields `Tracked`, its absence yields `Inline` — so the test drives the real branch rather than a hard-coded return.
- Three unit tests on `delegation_prompt_hash` in `handles/agent_control.rs`: the hash separates callers, separates conversations, and stays stable for the same caller on the same conversation, which is the #5033 idempotency case the salt must not break.
- `GoalsPage.test.tsx`: renders the error state rather than the template picker, and retries from it.

`rusqlite` is added as a **dev-dependency** of `librefang-api` for the storage-failure helper — same workspace version the memory crate already compiles, test-only, no new third-party surface.

The `i18n-parity` check fails on this branch, and fails identically with the locale changes stashed — it is pre-existing drift in `network.*`, `prompts.*` and `workflows.*` plural keys, unrelated to this PR. The four `goals.loadError` keys are at parity.


crates/librefang-api/src/routes/goals.rs
```diff
@@ -59,7 +63,7 @@ pub async fn list_goals(State(state): State<Arc<AppState>>) -> impl IntoResponse
         Ok(_) => Vec::new(),
         Err(e) => {
             tracing::warn!("Failed to load goals: {e}");
```
houko, line 65: Minor, non-blocking: each new `Err(e)` arm (here in `list_goals`, and again in `get_goal_children` / `start_goal_run`) does `tracing::warn!("Failed to load goals: {e}")` immediately before `ApiErrorResponse::internal_scrub(e)`, which itself logs the same error at `tracing::error!` (`internal_scrub` in `types.rs`). That's the same failure logged twice, at two different levels, on every one of these paths. It matches the pre-existing `get_goal` arm right above (explicitly called "mirrors `get_goal`" in the new doc comment), so it's not a regression this PR introduced — but since three new call sites now repeat it, it might be worth dropping the redundant `warn!` here (and eventually in `get_goal`) and relying on `internal_scrub`'s `error!` alone, in a follow-up. Not fixing this myself since it touches the pre-existing `get_goal` pattern this PR deliberately mirrors, which is outside this PR's stated diff.

---

changelog.d/fixed/6654-goals-storage-failure-is-a-500.md
houko, line 1: Judgment call, not fixing myself: the PR closes #6650, #6652, #6653, and #6654, and has a changelog fragment named for each of #6650, #6652, and #6654 — but #6653 (the `GET /api/goals/{id}/children` leak, fixed in the same handler in `routes/goals.rs`) doesn't get its own `changelog.d/fixed/6653-*.md` file. It is described in this fragment's body ("`GET /api/goals/{id}/children` was worse in both directions at once..."), so the content isn't missing, and the PR body itself groups #6654/#6653 together under one heading — so this may well be intentional grouping rather than an oversight. Flagging in case the convention (`changelog.d/README.md`: "named after the PR or issue number so fragments sort usefully") was meant to produce one fragment per closed issue; if so, worth either renaming this file to reference both (e.g. `6653-6654-goals-storage-failure-is-a-500.md`) or splitting it, your call.

---

crates/librefang-kernel/src/kernel/handles/agent_control.rs
```diff
@@ -12,6 +12,32 @@ use super::super::{manifest_to_capabilities, LibreFangKernel};
 /// receiving a truncated `[ASYNC_RESULT]` preview.
 const SPILL_DELEGATION_RESULT_ABOVE: usize = 500;
 
+/// Opaque, deterministic dedupe hash for a `TaskKind::Delegation` (#6652).
```
houko, line 15: Not about this line's content, which is fine — flagging a process finding on this PR's commit history instead, since there is no other place to attach it.

Commit `27f529d7` ("docs: reflow this PR's new comments to one sentence per line", the commit that last touched this docstring) is authored as `Claude <noreply@anthropic.com>` — visible in the PR's Commits tab and in `git log` on this branch. CLAUDE.md's "No AI / Claude attribution in commit messages, PR bodies, or comments" rule and the repo's `commit-msg` hook both target the message *text*; neither checks the author identity field, so this slipped through even though the message body itself contains no attribution string. Worth a squash or `git commit --amend --author` before merge so the permanent record doesn't carry an AI author on a human-owned PR.

No action taken here — this is now baked into the merged history (squash-merged as `6eaa451`, whose squash-commit author is correctly `Evan Hu`), so there's nothing left to fix mechanically; leaving this for awareness / to close the gap in the attribution check for next time.

---

PR #6663, opened 2026-07-30
fix(dashboard): stop offering config edits the write endpoint refuses, and key field descriptions by section

houko:
Refs #6636. Second of two follow-ups from #6636's observation list, covering (d), (f) and (c).

**(d) Read-only fields were rendered editable.** `approval.require_approval` is deliberately absent from `WRITABLE_EXACT_PATHS` so a leaked Owner-role API key cannot relax approval policy over HTTP. The dashboard did not know that, offered the edit, and the save came back as a bare 403 — which reads as "saved but not applied", the reported symptom.

`GET /api/config/schema` now carries `x-non-writable`: the resolved list of paths `POST /api/config/set` rejects. The server sends the verdict rather than the two allowlists because `is_writable_config_path` is not a lookup — it layers an exact-path list, section prefixes, a depth-2-only rule for some of those prefixes, and a suffix scrub for secret-bearing and privilege-bearing key names. Re-implementing that in TypeScript would make the SPA a third place to keep in sync with the two Rust lists, and it would drift silently, greying out fields the write path would have accepted. The oracle is the derived schema rather than a serialized config, matching `every_writable_allowlist_entry_has_a_backing_config_field`: `config/types.rs` carries dozens of `#[serde(skip_serializing_if = …)]` attributes, so a value walk cannot see a field whose predicate holds for its default. A path absent from the set is treated as writable, so an enumeration gap degrades to the previous behaviour rather than to a field nobody can edit.

The fields stay visible — an operator must be able to read a security setting back — but the control subtree is `inert` and carries a line pointing at `config.toml`. `inert` rather than a `disabled` prop threaded through every `ConfigFieldInput` branch: it blocks pointer and keyboard interaction for the whole subtree, including the composite editors that render several controls.

**(f) Every `mode` field showed the kernel's description.** `t(\`config.desc_${fieldKey}\`)` keyed on the leaf name alone, so `exec_policy.mode`, `reload.mode`, `docker.mode`, `privacy.mode` and `sanitize.mode` all rendered "Kernel operating mode — stable, default, or dev", which is true of none of them. Lookup is now section-first with the bare leaf as fallback, so the five ambiguous fields get their own strings while genuinely section-neutral names (`enabled`, `timeout_secs`, `model`, `provider`) keep sharing one. Labels follow the same rule. Six new strings in all four locales, since `locale-parity` requires matching key sets.

**(c) does not reproduce as described, but an adjacent edit-loss path does.** The report was that the Agent tab's prompt textarea does not refresh when a different agent is selected. It does: `HandAgentEditor` is keyed by role and `HandDetailPanel` by hand id, so both switching agent and switching hand remount the editor and reset every draft. That has been true since #6166 (2026-06-18), before the issue was filed.

What is real in the same component is losing a draft to an async arrival. `currentPrompt` starts as the HAND.toml baseline and switches to the live value the moment `useAgentDetail` resolves — which for a hand agent always differs, since the live prompt carries the rendered tails — and a bare `useEffect(() => setPrompt(currentPrompt), [currentPrompt])` wiped whatever had been typed in between. The effect now compares against the last value it applied, which distinguishes "the server moved" from "the user typed", and only the former wins. Same for the tool list, whose open-coded comparison moved into a shared `sameTools` helper the dirty check already needed.

Verification: `cargo clippy --workspace --all-targets -- -D warnings` clean. `cargo test -p librefang-api --lib` 892 passed. `cargo test -p librefang-api --test config_routes_integration` 44 passed, including the new `config_schema_reports_the_paths_the_write_endpoint_refuses`, which carries a sanity floor on the set size and cross-checks one non-writable and one writable path against the write endpoint so the emitted set cannot drift from the rule it mirrors. `cargo test -p librefang-api --test config_schema_overlay` 4 passed. Dashboard: `pnpm typecheck` clean, `pnpm lint` 0 errors, `pnpm test --run` 937 passed, `pnpm build` succeeds.

Refs #6636


houko (commented):
Reviewed against CLAUDE.md. Code, tests, and locale changes look solid — verified locally in a fresh worktree:
- `cargo check -p librefang-api --lib`: clean
- `pnpm typecheck` / `pnpm lint` (dashboard): clean, 0 errors
- `pnpm build` (dashboard): succeeds
- `pnpm test --run` (dashboard): 936/937 passed (the one failure was a `testTimeout` on an unrelated locale-coverage scan under heavy host load, not an assertion failure — not a regression from this PR)

`non_writable_schema_paths` walks exactly one level below each top-level `KernelConfig` field, which matches how `ConfigPage.tsx` actually renders paths (nested structs render as a single `object`-typed field under `section.field`, never flattened further), so the depth is correct and not just coincidentally matching the two test cases. Good catch on `is_writable_config_path` being the oracle rather than re-deriving the rule client-side.

Two process nits, not blocking, no code changes needed:

1. The extended body of the `fix(dashboard): …` commit (d0af4f0) and the current PR description both pack multiple sentences per paragraph onto a single line, rather than one sentence per line as CLAUDE.md's prose-wrapping rule requires (that section explicitly lists "commit message bodies" and "PR bodies"). Two follow-up commits on this branch already reflowed the *code* doc-comments this same PR added (in `manage.rs`, `config_routes_integration.rs`, `ConfigPage.tsx`, `HandsPage.tsx`) to one-sentence-per-line, but the commit message and PR body text itself were not. Fixing the already-pushed commit's message would need an amend + force-push, which is out of scope for this review pass — flagging for awareness rather than action.

2. Those two follow-up commits (`e129ae1`, `5a43a59`) are authored as `Claude <noreply@anthropic.com>`. The message text itself carries no attribution boilerplate (so the `commit-msg` hook's regex correctly let them through), but the author identity itself is a literal AI identity in `git log`/`git blame`, which is the kind of provenance signal the repo's no-AI-attribution policy is aimed at. Worth a squash/re-author before merge if that matters to this repo's history hygiene — a maintainer call, not something to fix unilaterally from a review pass.

No functional, security, or auth-adjacent issues found. Not touching auth middleware, config hot-reload classification, or route registration, so those checklist items don't apply here.

---

crates/librefang-api/src/routes/config/manage.rs
houko, line 1: `non_writable_schema_paths` also pushes the bare top-level section name (e.g. `web`, `channels`, `approval`, `exec_policy`) whenever `is_writable_config_path(section)` is false, which it always is for a struct-typed section — `is_writable_config_path` only returns true for an exact scalar leaf or a `prefix.` match, and a bare section name like `"web"` matches neither (it doesn't start with `"web."`).

The dashboard only ever checks `nonWritablePaths.has(path)` for `path = "<section>.<field>"`, so these bare-section entries are currently inert noise rather than a rendering bug. But they do inflate the emitted array with paths the write endpoint was never going to receive in the first place, and the integration test's sanity floor (`non_writable.len() > 20`) can't distinguish "real per-field non-writable paths" from "one spurious entry per struct-typed section" — so the floor is weaker evidence than it looks.

Worth a maintainer call: skip pushing `section` itself when the property is a struct (has nested `properties`), and reserve the bare-name entry for genuinely scalar root leaves only?

---

houko (commented):
Pushed one mechanical fix: rewrapped the remaining new prose that was still hard-wrapped at a fixed column (the `x-non-writable` JSDoc in `api.ts`, plus three inline comments in the new `config_schema_reports_the_paths_the_write_endpoint_refuses` test) to one sentence per line, per CLAUDE.md's prose-wrapping rule. The two earlier commits on this branch had already reflowed `ConfigPage.tsx`, `HandsPage.tsx`, and the doc comments in `manage.rs` / the same test file, but missed these four spots.

One more thing worth a maintainer look before merge, not fixed here: commits `e129ae1` and `5a43a59` on this branch are authored `Claude <noreply@anthropic.com>`. CLAUDE.md's "No Claude / Anthropic / AI attribution in commit messages, PR bodies, or comments" rule is enforced for message text by the `commit-msg` hook, but the author identity on those two commits carries the same attribution the rule is meant to keep out of `git log`/`blame`. Re-authoring them needs a rebase + force-push to this branch, which is outside what this review pass will do unilaterally — flagging so a maintainer can decide (rebase with `--author` fixed, or accept as-is).

One design question left inline on `manage.rs`: `non_writable_schema_paths` also enumerates bare top-level section names, which are currently harmless noise but weaken the integration test's sanity-floor assertion.

---

houko (commented):
CLAUDE.md compliance pass (automated, mechanical check only — not a design review).

**Finding: AI attribution in commit metadata (Git Conventions / Commit & PR hygiene — "No Claude / Anthropic / AI attribution in commit messages, PR bodies, or comments").**

Two commits on this branch carry the git *author* identity `Claude <noreply@anthropic.com>` rather than a human author:

- `e65710229e0195846041e2f70198d329081f7fbc` — "docs(api): reflow new config-schema doc comments to one sentence per line"
- `ac751cab943d6dd4864e4d352efd7338f86643fb` — "docs(dashboard): reflow new prose comments to one sentence per line"

The commit-msg bodies themselves don't contain a `Co-Authored-By: Claude` footer or the usual `🤖 Generated with` string that the `commit-msg` hook and `guard-bash-safety.sh` scan for, so neither caught it — but the author field is a more direct form of the same attribution the rule bans, and it's visible in `git log`, `git blame`, and the GitHub commit view for anyone who clicks through.

I'm flagging rather than fixing because the correction (`git commit --amend --author=...` on non-tip commits, or an interactive rebase, followed by a force-push) is exactly the class of operation this automation is barred from performing unilaterally — force-push and history rewrites on someone else's branch need the maintainer's explicit go-ahead. @houko: rebasing those two commits onto `Evan Hu <e-hu@unext.jp>` (or squashing them into the adjacent commits you authored) and force-pushing your own branch would clear this.

Everything else checked out: `changelog.d/fixed/6663-dashboard-config-papercuts.md` follows the fragment format (sentence-per-line body, `(#6663) (@houko)` trailer, correct section directory); the dashboard changes only read `x-non-writable` off the existing `useConfigSchema()` result (no inline `fetch`/`api.*`, no new query-key factories, no mutation changes needing new `invalidateQueries`); locale keys are present in all four locale files; and the new prose in doc-comments/JSDoc/inline comments is already one-sentence-per-line (a prior commit in this same branch reflowed it). The already-open thread about `non_writable_schema_paths` emitting bare struct-section names is a separate (already-reported, unresolved) logic question, not re-flagged here.

---

changelog.d/fixed/6663-dashboard-config-papercuts.md
houko, line 1: Per CLAUDE.md's "One PR ↔ one issue (or one tight cluster)" and the changelog convention of "one file, one bullet": this fragment (and the PR) bundles three fixes across two unrelated dashboard surfaces — the Config page's non-writable-field graying, the Config page's per-section field-description keying, and a `HandAgentEditor` (Hands page) edit-loss bug on `prompt`/`tools` drafts. The first two are a tight cluster (same page, same root cause area), but the hand-agent-tab draft-loss fix touches a different component/page and a functionally unrelated bug (state sync vs. an editable-vs-actually-writable UI question). Was this intentionally scoped as one PR, or should the `HandAgentEditor` fix be its own PR/fragment? If it's staying bundled, no action needed — just flagging for a maintainer call since it isn't a mechanical fix.

---

crates/librefang-api/dashboard/src/pages/HandsPage.tsx
```diff
@@ -528,8 +533,24 @@ function HandAgentEditor({
...
+  const appliedPrompt = useRef(currentPrompt);
+  useEffect(() => {
+    setPrompt((draft) => (draft === appliedPrompt.current ? currentPrompt : draft));
+    appliedPrompt.current = currentPrompt;
+  }, [currentPrompt]);
+
+  const appliedTools = useRef(currentTools);
+  useEffect(() => {
+    setTools((draft) => (sameTools(draft, appliedTools.current) ? currentTools : draft));
+    appliedTools.current = currentTools;
```
houko, line 552: This is a real, well-explained fix for the edit-loss bug (comparing against `appliedPrompt`/`appliedTools` refs rather than blindly re-adopting `currentPrompt`/`currentTools`), but `HandsPage.test.tsx` (which already mocks `useAgentDetail`/`useAgentTools` via `useHandDetail` etc. and has 20 passing cases) doesn't appear to gain a regression test for it. The repo's own convention for API route changes is "add an integration test at the injection site, not just the implementation site" — the same principle applies here: a test that (1) renders with a draft the user has typed, (2) resolves/changes `currentPrompt`/`currentTools` via the mocked query, and (3) asserts the draft survives would pin this behavior and catch a future regression to the old `useEffect(() => setPrompt(currentPrompt), [currentPrompt])` pattern. Worth adding, or is there a reason it's out of scope for this PR?

---

houko (commented):
Mechanical prose-wrap fix pushed (see commit below). Two remaining points are judgment calls for a maintainer, left as inline comments rather than fixed directly.

---

crates/librefang-api/dashboard/src/pages/HandsPage.tsx
houko, line 552: This is a real async-race fix (item (c)) — comparing against the last-applied value rather than blindly overwriting the draft on every `currentPrompt`/`currentTools` change. It's exactly the kind of subtle timing bug that regresses silently if someone later "simplifies" this back to a bare `useEffect(() => setPrompt(currentPrompt), [currentPrompt])`.

`HandsPage.test.tsx` has no case exercising `HandAgentEditor`'s draft-preservation behavior (no test title mentions prompt/tools/draft). Given the PR already has a dedicated integration test for the backend half of this PR (`config_schema_reports_the_paths_the_write_endpoint_refuses`), would a small RTL test asserting "typing in the prompt survives a `currentPrompt` change that differs from the last-applied value, but a currentPrompt change matching the draft still lands" be worth adding here too? Not blocking since CLAUDE.md's mandatory-test rule is scoped to backend routes, but this fix has no regression coverage as-is.

---

crates/librefang-api/src/routes/config/manage.rs
```diff
@@ -1137,18 +1137,76 @@ pub async fn config_schema(State(state): State<Arc<AppState>>) -> impl IntoRespo
...
+
+    let definitions = root.get("definitions").and_then(|v| v.as_object());
+    let Some(properties) = root.get("properties").and_then(|v| v.as_object()) else {
+        return Vec::new();
+    };
+
+    let mut paths = Vec::new();
+    for (section, prop) in properties {
+        if !super::is_writable_config_path(section) {
+            paths.push(section.clone());
```
houko, line 1190: Minor: this pushes the bare top-level section name (e.g. `web`, `budget`, `approval`, `channels`, `exec_policy`...) into `x-non-writable` whenever `is_writable_config_path(section)` is false for the section name alone — which it almost always is, since `is_writable_config_path` only matches a bare name via an exact-path entry (used for root-level scalars and collection-typed sections like `provider_urls`) or a `<prefix>.` match, never a bare struct-section name. So most struct sections end up contributing a harmless but meaningless entry to the array (the dashboard only ever looks up `sKey` alone for `isCollection` fields, where the bare name legitimately matters and is correctly exact-listed when writable).

Not a correctness bug — `nonWritablePaths.has(path)` on the dashboard side never matches these extra entries for non-collection sections, and the sanity-floor test (`> 20`) still holds either way — but it does inflate the payload and could confuse someone reading raw `x-non-writable` output later trying to reason about "which paths does this include and why." Worth either skipping the section-bare-name push when the section isn't a collection-typed field, or a one-line comment noting the bare-name entries are intentionally over-inclusive/harmless.

---

PR #6664, opened 2026-07-30
fix(config): re-enable the KernelConfig schema golden guard and regenerate its fixture

houko:
Found while working on #6636's follow-ups: the `KernelConfig` schema golden guard has been off, and real schema changes have been landing without the review it exists to force.

## What happened

`kernel_config_schema_matches_golden_fixture` carries a hardcoded `#[ignore]` whose own comment says it is temporary:

```
// TEMPORARILY IGNORED — golden fixture drifted by 490 bytes …
// then drop this `#[ignore]`.  Tracked separately so #4084 can land.
```

The fixture was never regenerated and the attribute never came off. Because the test reports `ignored` rather than failing, nothing surfaced — every subsequent change to the derived schema passed CI silently, which is the exact failure mode the fixture was added to prevent (#3055 review P2: "the schema is auto-derived, so a rename can silently reshape the wire format").

## What slipped through

Regenerating and diffing against the committed fixture *order-insensitively* — the raw diff is dominated by key-ordering churn — leaves four semantic changes:

| Change | Kind |
| --- | --- |
| `properties.providers` + `definitions.ProvidersConfig` | new section — the org-wide LLM provider allowlist (#6484) |
| `ExecPolicy.full_mode_skips_approval` | new field |
| `ApprovalPolicy.auto_approve` | description rewritten |
| `properties.exec_policy` | gained a `description`; `default` reshaped |

None looks wrong. The point is that all four reached `GET /api/config/schema` without anyone reviewing the schema diff.

## Changes

- Regenerate `kernel_config_schema.golden.json`.
- Remove the `#[ignore]` and replace the stale comment with one that says what to do when the guard fires.

The diff is 7,278 lines for those four changes because schemars' output ordering also moved. That is precisely why this lands on its own instead of inside a feature PR — a reviewer can take the summary above on faith for the reshuffle and read the four real entries, which is impossible once it is mixed with unrelated work.

## Verification

`cargo test -p librefang-api --test config_schema_golden` — `kernel_config_schema_matches_golden_fixture` now runs and passes (was `ignored`). `regenerate_golden` stays `#[ignore]`d; it is the manual regeneration entry point.

## Ordering note for reviewers

Two open PRs of mine add fields to `KernelConfig`, so whichever of the three lands second needs a one-command fixture bump:

- #6661 adds `registry.auto_sync`.
- #6663 adds no schema field (it only extends the `x-` overlay, which this fixture deliberately excludes).

That bump is the guard working as intended — a small, readable delta attached to the change that caused it.


houko (commented):
Reviewed the three changed files against the base at 469634e (current `origin/main` tip — this branch is not stale; no `KernelConfig`-touching commits have landed on `main` since, so #6661/#6663's "whichever lands second needs a fixture bump" note doesn't apply yet).

Mechanical checks:
- `kernel_config_schema_matches_golden_fixture` has the `#[ignore]` attribute genuinely removed (not just reworded), and `regenerate_golden` correctly stays `#[ignore = "run manually with --ignored to regenerate the golden fixture"]` as the manual entry point. Matches the PR description.
- Determinism: no non-determinism risk found. `librefang-types/Cargo.toml` does not enable schemars' own `preserve_order` feature, so `schemars::Map` resolves to `BTreeMap` for `properties`/`definitions`/`extensions` regardless of the workspace's `serde_json` `preserve_order` (which pulls in `indexmap` per `Cargo.lock`, but only governs `serde_json::Value`'s own map type, not the source ordering fed into it by schemars' `BTreeMap`-backed structs). The new `ProvidersConfig.allowed` field is a `Vec<String>`, not a keyed map, so it carries no ordering risk either. The manual `JsonSchema` impls (`PromptCacheStrategy`, `ApprovalDecision`) build via `serde_json::json!` but deserialize back into schemars' `BTreeMap`-typed `SchemaObject`, so the intermediate map type doesn't leak into final ordering.
- `changelog.d/fixed/6664-config-schema-golden.md` is present, in the correct section (`fixed/`, since this re-enables a broken guard), and ends `(#6664) (@houko)` in the expected format.
- Test scoping in the PR body (`cargo test -p librefang-api --test config_schema_golden`) is correctly scoped per repo convention.

One nit worth fixing before merge: both the commit message body and the PR description pack multiple sentences onto single lines in several paragraphs (e.g. the commit's first paragraph runs two full sentences together: "...said to drop the attribute once the fixture was regenerated. The fixture was never regenerated and the attribute never came off, so..."; same pattern in the "What happened" section of the PR body). CLAUDE.md's prose-wrapping rule ("one sentence = one line, regardless of length") applies explicitly to commit message bodies and PR bodies. Not blocking, and not something I'm rewriting for you since amending the commit would need a force-push — just flagging so it can be split at the sentence boundaries if you get a chance.

No fixture regeneration needed from my side — the golden file itself looks internally consistent with the described four semantic changes and the base-SHA check above rules out the "already stale" scenario this run was asked to check for.

---

crates/librefang-api/tests/config_schema_golden.rs
houko, line 1: Verified this genuinely re-enables the guard (the `#[ignore]` attribute is removed, not just its message changed) and cross-checked the regenerated fixture against current source: an order-independent structural diff of the old vs. new `kernel_config_schema.golden.json` shows exactly 11 added leaf values, 0 removed, 1 changed — landing precisely on the four clusters this PR's description calls out (`providers`/`ProvidersConfig`, `ExecPolicy.full_mode_skips_approval`, the `ApprovalPolicy.auto_approve` doc-comment reword, `exec_policy`'s new description/reshaped default), each matching the current struct definitions in `crates/librefang-types/src/config/types.rs` and `crates/librefang-types/src/approval.rs` verbatim. No stale/hand-edited content found.

One coordination note, not something I'm resolving here: **this overlaps with #6666** (`fix/mcp-transport-deny-unknown-fields`). #6666 adds `#[serde(deny_unknown_fields)]` to `McpTransportEntry`, `HttpCompatHeaderConfig`, and `HttpCompatToolConfig`, all reachable from `KernelConfig` via `mcp_servers: Vec<McpServerConfigEntry>`. `schemars` reflects `deny_unknown_fields` as `additionalProperties: false` in the generated schema, so that PR changes the same golden schema this PR guards. #6666's own description says the fixture "is not regenerated here" and "stays ignored either way," reasoning from the guard's current `#[ignore]`d state — but once this PR merges, that assumption is stale: `kernel_config_schema_matches_golden_fixture` will actually run and fail against #6666's diff until it regenerates the fixture too. Worth flagging to the author/maintainer so #6666 gets a fixture bump rather than landing red, but I'm not touching #6666 to resolve it.

Minor, non-blocking: the PR body and the changelog fragment both say the fixture diff is "7,278 lines," but `git diff --stat` (and the GitHub files API) report 6,872 insertions / 6,844 deletions = 13,716 changed lines for this file. Cosmetic only — doesn't affect correctness — but the changelog fragment ships to the release notes verbatim, so it might be worth a one-line correction before merge.

---

houko (commented):
Reviewed the golden-guard re-enable and the regenerated fixture. Summary in the inline comment on `config_schema_golden.rs`: the four claimed semantic changes check out exactly against current source (structural diff confirms nothing else moved), and there's a genuine sequencing overlap with #6666 worth a maintainer's attention.

---

crates/librefang-api/tests/config_schema_golden.rs
```diff
@@ -87,15 +87,13 @@ fn normalize_path_like(value: &mut serde_json::Value) {
...
-// compile-break window.  Regenerate with:
-//     cargo test --test config_schema_golden -p librefang-api -- \
-//         --ignored regenerate_golden --nocapture
-// then drop this `#[ignore]`.  Tracked separately so #4084 can land.
+// Was `#[ignore]`d as a temporary measure while a 490-byte drift went
+// unresolved, with the attribute meant to come off once the fixture was
+// regenerated. It did not, and the guard stayed off long enough for real schema
+// changes to land unreviewed — see the PR that re-enabled it for the inventory.
+// If this fails, review the diff before regenerating: an unexplained change here
+// is a wire-format change to `GET /api/config/schema`.
```
houko: Minor CLAUDE.md prose-wrap nit: this replacement comment (lines 90-95) is hard-wrapped at ~80 columns, breaking mid-sentence (e.g. \"...a 490-byte drift went\" / \"unresolved, with the attribute...\"). Per CLAUDE.md's prose rule, new prose should break only at sentence boundaries — one sentence per line. Since this is new prose being introduced by this PR (replacing the old `TEMPORARILY IGNORED` comment), the rule applies. Suggested rewrap:

```rust
// Was `#[ignore]`d as a temporary measure while a 490-byte drift went unresolved, with the attribute meant to come off once the fixture was regenerated.
// It did not, and the guard stayed off long enough for real schema changes to land unreviewed — see the PR that re-enabled it for the inventory.
// If this fails, review the diff before regenerating: an unexplained change here is a wire-format change to `GET /api/config/schema`.
```

---

changelog.d/fixed/6664-config-schema-golden.md
```diff
@@ -0,0 +1,4 @@
+Re-enable the `KernelConfig` JSON Schema golden guard, which had been `#[ignore]`d long enough for real schema changes to land unreviewed.
+  The attribute was added as a temporary measure over a 490-byte drift, with a comment saying to drop it once the fixture was regenerated; the fixture was never regenerated, so the guard reported "ignored" rather than failing and three subsequent schema changes reached `GET /api/config/schema` without the reviewable diff the fixture exists to force.
```
houko, line 2: The change count here looks internally inconsistent, and I think it's an overcount by one. Line 2 says "three subsequent schema changes reached `GET /api/config/schema`", but line 3 lists four items, and the PR body/commit message both say "four semantic changes."

I diffed the golden fixture order-insensitively against `origin/main` (flattened every leaf path, not just line-diffed) and only found 3 independent additions/changes:
1. `providers` property + `ProvidersConfig` definition (7 leaf entries)
2. `ExecPolicy.full_mode_skips_approval` (4 leaf entries — 3 in the `ExecPolicy` definition, 1 in the `exec_policy` default value embedded at the top level)
3. `ApprovalPolicy.auto_approve` description reworded (1 leaf entry)

The 4th item listed here — "a `description` plus reshaped `default` on `exec_policy`" — doesn't hold up: I compared `old['properties']['exec_policy']['description']` vs `new[...]` directly and they're byte-identical (`"Shell/exec security policy."` both times). It didn't "gain" a description — that field already existed and just moved position in the JSON object (key-ordering churn, same as everything else the PR correctly writes off). The only real change in that neighborhood is the `default` value picking up `full_mode_skips_approval`, which is the *same* field addition as item 2, not a separate 4th change.

Given the whole point of this PR is making the schema diff reviewable and correctly enumerated, worth tightening this bullet (and the matching PR-body table row / commit message) to 3 changes, folding the `exec_policy` default reshape into the `full_mode_skips_approval` bullet rather than listing it separately. Low confidence this needs a code change — flagging for author judgment rather than fixing directly since it's a narrative correction across the PR body/commit message/changelog, not a code file.

---

PR #6665, opened 2026-07-30
fix(api): make the plugin route-classification guard line-ending agnostic (fixes red main)

houko:
Fixes `main`, which is red on `Test / Windows (shard 2/2)`. The failure is mine, from the #6631 completeness guard.

## What broke

```
thread 'middleware::tests::every_mutating_plugin_route_is_explicitly_classified'
panicked at crates\librefang-api\src\middleware.rs:2031:14
```

The guard reflects the route table out of `routes/plugins.rs` via `include_str!`, then locates the end of `router()` with:

```rust
let body_end = body
    .find("\n}\n")
    .expect("router() body must terminate at a column-0 brace");
```

`include_str!` yields the file's bytes verbatim, and a Windows checkout stores CRLF. So the terminator is `\r\n}\r\n`, `find` returns `None`, and `.expect` fires. Linux and macOS store LF, which is why every lane I ran locally and every lane on #6639 was green — the Windows shard was the only one that could see it.

## The fix

Extraction normalises CRLF once, and is factored into `extract_mutating_plugin_routes` so the property is testable rather than assumed.

It now returns an empty vec instead of panicking when the shape is unrecognisable. The caller already asserts on the route count, so an extraction failure produces a message naming what it *did* parse rather than an opaque `expect` — a better failure mode for the next person, whatever the cause.

## The regression test is the point

`plugin_route_extraction_is_line_ending_agnostic` builds the CRLF form itself and asserts both inputs yield identical routes.

That is deliberate. A checkout on a Linux or macOS runner stores LF, so a test that relied on the host's line endings would pass on every lane that already passed and prove nothing about the one that failed. Same class as #5716 — platform-divergent behaviour needs a platform-independent reproduction, not a platform-specific runner.

**Mutation-tested**: removing the normalisation fails the new test with the CRLF-vs-LF mismatch, so it is not vacuous.

## Verification

- `cargo test -p librefang-api --lib` — 893 passed, 0 failed.

The Windows shard itself only runs on CI, so that lane on this PR is the confirmation that matters.

### DIFF
--- a/changelog.d/fixed/6665-plugin-guard-crlf.md
+++ b/changelog.d/fixed/6665-plugin-guard-crlf.md
@@ -1,0 +1,2 @@
+Fixed the `#6631` plugin route-classification guard panicking on a Windows checkout, where `include_str!` reflects `routes/plugins.rs` back with CRLF line endings and the guard's `\n}\n` terminator search matched nothing.
+Route extraction now normalises CRLF to LF before parsing and is factored into a shared helper covered by its own regression test that builds the CRLF fixture directly, so the platform-divergent case is provable without a Windows runner (#6665) (@houko)
--- a/crates/librefang-api/src/middleware.rs
+++ b/crates/librefang-api/src/middleware.rs
@@ -1991,18 +1991,12 @@ mod tests {
 
     /// Extract every mutating route registered in `routes::plugins::router()`.
     ///
-    /// Shared by the classification guard and by
-    /// `plugin_route_extraction_is_line_ending_agnostic`, which is what makes
-    /// the CRLF case provable without a Windows runner.
+    /// Shared by the classification guard and by `plugin_route_extraction_is_line_ending_agnostic`, which is what makes the CRLF case provable without a Windows runner.
     ///
-    /// Line endings are normalised first. `include_str!` yields the file's
-    /// bytes verbatim and a Windows checkout stores CRLF, so the `"\n}\n"`
-    /// body terminator matched nothing there — the guard panicked on `.expect`
-    /// and the Windows shard was the only lane that went red.
+    /// Line endings are normalised first.
+    /// `include_str!` yields the file's bytes verbatim and a Windows checkout stores CRLF, so the `"\n}\n"` body terminator matched nothing there — the guard panicked on `.expect` and the Windows shard was the only lane that went red.
     ///
-    /// Returns an empty vec rather than panicking when the shape is
-    /// unrecognisable: the caller asserts on the count, which produces a
-    /// message naming what it did parse instead of an opaque `expect`.
+    /// Returns an empty vec rather than panicking when the shape is unrecognisable: the caller asserts on the count, which produces a message naming what it did parse instead of an opaque `expect`.
     fn extract_mutating_plugin_routes(src: &str) -> Vec<(axum::http::Method, String)> {
         let src = src.replace("\r\n", "\n");
         let Some(body_start) = src.find("pub fn router()") else {
@@ -2016,8 +2010,7 @@ mod tests {
 
         let mut out = Vec::new();
         // Each `.route(` chunk holds one path literal and its method calls.
-        // Formatting splits these across lines, so operate per chunk rather
-        // than per line.
+        // Formatting splits these across lines, so operate per chunk rather than per line.
         for chunk in body.split(".route(").skip(1) {
             let Some(path_start) = chunk.find('"') else {
                 continue;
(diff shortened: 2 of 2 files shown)


crates/librefang-api/src/middleware.rs
houko, line 1: Verified the fix logic itself: `extract_mutating_plugin_routes` normalises `\r\n` → `\n` once at the top, and every downstream search (`"pub fn router()"`, `"\n}\n"`, `.route(` splitting) runs against that already-normalised string.
Both call sites (`plugin_route_extraction_is_line_ending_agnostic` and `every_mutating_plugin_route_is_explicitly_classified`) now go through the shared function, so there's no leftover copy of the old CRLF-unsafe `.find("\n}\n").expect(...)` anywhere in this file — confirmed by searching the whole file for `find("\n}\n")` and `ADMIN_ALLOWED`.
`plugin_route_extraction_is_line_ending_agnostic` builds a real CRLF fixture (`lf_src.replace('\n', "\r\n")`) and asserts the LF/CRLF results are non-empty and equal, so it's not a vacuous "both sides do nothing" pass.
I also checked the other `include_str!`-based source-introspection tests in the repo (`kernel/tests.rs`, `kernel/llm_drivers.rs`) for the same bug class — none of them embed a literal `\n` inside a `.find("...")` pattern, so they aren't exposed to this failure mode. No sibling fix needed.

One thing worth flagging before merge: the PR description says *"The Windows shard itself only runs on CI, so that lane on this PR is the confirmation that matters."*
That isn't what happened on this PR's own check-run (run [30507717257](https://github.com/librefang/librefang/actions/runs/30507717257)): `Test / Windows (shard ${{ matrix.shard }}/2)` shows `conclusion: skipped`, matrix unexpanded — it never ran.
That's because `.github/workflows/ci.yml` gates `test-windows` on `github.event_name == 'push' || github.event_name == 'merge_group'` (see the comment block right above `test-windows:`, lines ~917-934) — Windows/macOS intentionally don't run on `pull_request` events at all, only after landing on `main`.
I did independently confirm the underlying claim ("main is red") — [run 30501264939](https://github.com/librefang/librefang/actions/runs/30501264939) at `469634ec` (this PR's base sha) shows `Test / Windows (shard 2/2)` failing with exactly `middleware::tests::every_mutating_plugin_route_is_explicitly_classified` panicking, matching the PR body's repro.
So the bug report is accurate, but the "this PR's CI is the confirmation" framing isn't — the real confirmation only happens on the post-merge push to `main`. Worth softening that line so a merger doesn't read a green PR check list as proof the Windows lane passed.

---

houko (commented):
Reviewed the CRLF fix and cross-checked its "fixes red main" claim against CI history. Summary in the inline comment.

---

changelog.d/fixed/6665-plugin-guard-crlf.md
houko, line 1: The changelog-fragment commit (9378344) has its git author identity set to `Claude <noreply@anthropic.com>`.
That's AI attribution embedded directly in the commit metadata — visible in `git log`, `git blame`, and the GitHub commit view — even though the commit message text itself is clean.
CLAUDE.md's "No AI / Claude attribution" rule and the `commit-msg` hook both target message content, so this slipped through the automated checks, but it's the same category of provenance-in-history the rule exists to prevent.

Fixing this requires rewriting the author of an already-pushed commit (rebase + force-push), which isn't something I'll do unilaterally on someone else's commit — flagging for @houko to rewrite locally (`git commit --amend --author="..."` then force-push, since review hasn't loaded this yet) or squash it away.

---

houko (commented):
Reviewed the CRLF fix against CLAUDE.md's checklist (route-classification logic, LF/CRLF/lone-CR coverage, prose-wrap rule, AI attribution, changelog format).

Pushed one small fix: rewrapped the new doc-comments/inline-comments in middleware.rs that were hard-wrapped mid-sentence instead of breaking at sentence boundaries per CLAUDE.md's prose rule (no behavior change).

Everything else checked out: the CRLF normalization is correct and centralized in one shared helper (no leftover un-fixed duplicate parsing), the extraction fails closed via the `seen.len() >= 14` count guard rather than silently passing on an empty/degenerate parse (covers the lone-CR edge case even without an explicit test for it), the new regression test genuinely exercises an embedded `\r\n` fixture, the commit message is conventional-commit formatted with no AI attribution, and the changelog fragment (added in a follow-up commit) is correctly placed/formatted. `cargo test -p librefang-api --lib` (70 middleware tests) and `cargo clippy -p librefang-api --all-targets -- -D warnings` both pass on the current head.

One finding left as an inline comment: the changelog-fragment commit's git author identity is literally `Claude <noreply@anthropic.com>`, which is AI attribution in commit metadata rather than message text, so it slipped past the message-scanning hooks.

---

Issue #5130, opened 2026-05-16
config: most sections + repeated tables lack deny_unknown_fields, typos silently dropped


houko:
## Problem
Only `PromptCacheConfig` uses `#[serde(deny_unknown_fields)]` (added in #4970). For every other nested section, `KernelConfig::detect_unknown_nested_fields` catches typos only in single-table dotted paths — it does NOT descend into elements of repeated tables (`[[channels.telegram]]`, `[[mcp_servers]]`, `[[hands]]`, …). A typo inside such an element is silently dropped and the field reverts to its struct `Default`.

## Evidence
- The lone struct using the guard: `crates/librefang-types/src/config/types.rs:4894`.
- Nested validator: `crates/librefang-types/src/config/validation.rs` (`detect_unknown_nested_fields` walks single-table paths; `Vec<T>` element keys are not iterated against the inner struct's JsonSchema).
- Example shape: `[[channels.telegram]] bot_tken_env = "FOO"` deserializes successfully → `TelegramChannelConfig::default()` fills in `bot_token_env = ""` → adapter logs `Telegram configured but is not set` and refuses to start.

## Impact
Common operator complaint: "I set `bot_token_env` and the bot still doesn't start." The dashboard shows the channel as configured; the daemon logs say "not set". The mismatch wastes hours of diagnosis. The same shape applies to `[[mcp_servers]] commnd`, `[[hands]] allowed_env` typos, per-agent override blocks in `agent.toml`, etc.

## Suggested fix
1. Add `#[serde(deny_unknown_fields)]` to every nested config struct touched by repeated tables (`TelegramChannelConfig`, `DiscordChannelConfig`, `SlackChannelConfig`, `MattermostChannelConfig`, `WhatsAppChannelConfig`, `McpServerConfig`, `HandConfig`, agent overrides …).
2. Or extend `detect_unknown_nested_fields` to descend into `Vec<Table>` elements and validate per-element keys against the inner struct's schemars schema.
3. Add a regression test: a typo inside `[[channels.telegram]]` must produce a startup error.

PR #6666, opened 2026-07-30
fix(config): reject unknown keys inside `[mcp_servers.transport]`, and make the MCP transport round-trip lossless

### LINKED ISSUE #6612: McpTransportEntry lacks deny_unknown_fields that its parent has — a whole [mcp_servers.transport.env] table is silently discarded
`McpServerConfigEntry` carries `#[serde(deny_unknown_fields)]`, with a comment stating exactly why:

```rust
// repeated-table elements (#5130), so the only way to surface a typo in,
// say, `[[mcp_servers]] timout_secs = 30` is for serde itself to reject it.
#[derive(Debug, Clone, Serialize, Deserialize, schemars::JsonSchema)]
#[serde(deny_unknown_fields)]
pub struct McpServerConfigEntry {
```

`McpTransportEntry`, one level down inside the same `[[mcp_servers]]` block, has no such guard:

```rust
#[derive(Debug, Clone, Serialize, Deserialize, schemars::JsonSchema)]
#[serde(tag = "type", rename_all = "snake_case")]
pub enum McpTransportEntry {
```

So a misspelled scalar on the parent (`timout_secs`) fails the load loudly, as intended, while an entire unsupported table nested one level deeper is accepted and dropped without a word.

## What that cost us

Our config had this for a local stdio MCP server:

```toml
[mcp_servers.transport]
command = "..."
args = ["..."]
type = "stdio"

[mcp_servers.transport.env]
MY_API_KEY = "..."
MY_API_URL = "..."
```

`McpTransportEntry::Stdio` has only `command` and `args`, so the `env` table was discarded at load. The env-passing mechanism that does exist is a different field — the top-level `McpServerConfigEntry.env: Vec<String>` name list, resolved per `crates/librefang-runtime-mcp/src/lib.rs` (`env_clear()`, then `SAFE_ENV_VARS`, then each declared entry either as `KEY=value` or as a bare name looked up from the parent environment).

Verified on the running daemon: the subprocess has neither variable in its environment, and `GET /api/mcp/servers/{name}` reports `"env": []`. The config looks configured and is not.

Nothing broke, for a reason that is worse than breaking: the server's own script has hardcoded fallback defaults that happen to equal the values in the dead table, so it has been running on the defaults the whole time. The failure mode this sets up is a credential rotation — change the secret, update the (inert) block, and the subprocess silently keeps using the stale hardcoded default with no error anywhere.

Our other six MCP servers use the correct top-level `env = [...]` form, so this was a hand-authored mistake. The point is that the schema had every opportunity to say so at load time and stayed quiet, in a config file whose parent struct is explicitly guarded against exactly this class of error.

## Suggested fix

Reject unknown keys inside the transport table so the mistake surfaces at load, the same way the parent's guard does.

Note `deny_unknown_fields` interacts awkwardly with serde's internally-tagged enums (`#[serde(tag = "type")]`), so a straight attribute addition may not compile or may not behave as expected here. If that turns out to be the case, an explicit validation pass over the raw transport table at load time would achieve the same result — the goal is a loud failure, not a specific serde attribute.

A lower-effort variant that would also have caught this: warn (not fail) on unknown keys under `[[mcp_servers]]` subtables, listing the dropped key paths.

## Environment

- Deployed `v2026.7.21`; both structs unchanged on current `main` (`cb7608d9`).

Happy to test a patch.

houko:
`[mcp_servers.transport]` accepted and silently discarded any key it did not recognise. The reporter wrote a `[mcp_servers.transport.env]` table — plausible, but not the mechanism, which is the parent entry's `env: Vec<String>` name list — and got a server that looked configured, ran on its script's hardcoded fallbacks, and would have failed at the next credential rotation.

`McpServerConfigEntry` has carried `deny_unknown_fields` since #5130 for exactly this reason: `detect_unknown_nested_fields` bails on array-of-table paths, so serde is the only layer that can see a typo inside a `[[mcp_servers]]` element at all. The guard stopped one level above where operators hand-write the most config.

Closes #6612

### `librefang-types` — the guard

- `#[serde(deny_unknown_fields)]` on `McpTransportEntry`, `HttpCompatHeaderConfig`, and `HttpCompatToolConfig`.
- On an internally-tagged enum serde applies the attribute per variant against the buffered content, so the discriminating `type` key is still accepted. That is not obvious from the attribute alone and is *not* true of adjacently- or untagged containers, so it is pinned by test rather than assumed from the fact that it compiles.
- Both `HttpCompat` structs sit under `[[mcp_servers.transport.headers]]` / `[[mcp_servers.transport.tools]]` — arrays of tables nested inside an array of tables — and every field except `name` / `path` has a `serde(default)`, so a misspelled `responce_mode` left the tool wired to the default JSON response mode.
- `input_schema` is a free-form `serde_json::Value`; the guard does not reach into it, and a test asserts operator-authored schema content passes through untouched.

### `librefang-api` — the round trip the guard makes load-bearing

Guarding the transport turns `serialize_mcp_transport` from a display summary into a contract: any key the read route synthesises into the `transport` object that is not a field of the variant is now a `400` on the way back in. Two such keys existed, and serde short-circuits at the first unknown key, so both had to go.

- Dropped the derived `source` discriminator (`"env"` / `"static"` / `"unset"`) from `http_compat_header_summary` and the derived `tools_count` from the `HttpCompat` arm. Neither has a consumer: `skills_routes_test.rs:180-181` asserts `tools_count` / `source` on the *skills* list route (`routes/skills/skill.rs:76`), the dashboard reads `connected[].tools_count` (a genuine count of *discovered* tools, inserted at `routes/skills/mcp.rs:348`), and the dashboard's `McpTransport` type (`dashboard/src/api.ts:4448`) has no `http_compat` variant at all. `tools_count` was also just `tools.len()` — derivable from the array beside it.
- Renaming rather than dropping was not available: `deny_unknown_fields` rejects every name outside `{name, value, value_env}`, so there is no non-colliding name for a synthesised header key.
- Added `input_schema` to `http_compat_tool_summary`. It has `#[serde(default = "default_http_compat_input_schema")]`, so omitting it made a `GET` → `PUT` overwrite a hand-authored JSON Schema with `{"type":"object"}` — a lossy round trip that no status code reports. Pre-existing, but it sits in the lines the fix above had to edit, and leaving it would have made the round trip still lossy.
- Added `merge_http_compat_secrets` in `routes/skills/mcp.rs`, wired into `update_mcp_server` beside the existing `env` merge. A static `http_compat` header `value` stays redacted on read because it is a credential and `/api/mcp/servers` is in `PUBLIC_ROUTES_DASHBOARD_READS` (#6630), so the write side restores it from the stored entry — semantics transliterated from `merge_env_preserving_inline_values` rather than invented: submitted `value` wins, a header with neither `value` nor `value_env` restores the stored value, an absent header is an explicit removal, and a submitted `value_env` is left alone because it names a variable rather than holding a secret. The merge applies only when both the submitted and stored transports are `http_compat`, so switching a server's transport type does not resurrect old credentials.

### `librefang-types` — a second, separate corruption found while testing the above

Writing the round-trip test surfaced a bug on a path the test harness did *not* cover, and it is worse than the one the issue reports.

`boot()` in `mcp_env_disclosure_test.rs` uses `mcp_runtime_store = Db`, which stores JSON directly. The **default** store is `File`, which goes `serde_json::to_value` → `json_to_toml_value` → `toml::to_string_pretty` → disk → serde. TOML has no null, and that converter maps `Value::Null` to an **empty string** rather than dropping the key (`routes/config/mod.rs:942`). This is exactly why `McpServerConfigEntry::template_id` and `oauth` already carry `skip_serializing_if = "Option::is_none"` with comments calling it load-bearing.

`HttpCompatHeaderConfig.value` / `.value_env` are `Option<String>` on that same path with no such attribute. Observed on-disk output before the fix:

```toml
[[mcp_servers.transport.headers]]
name = "Authorization"
value = ""
value_env = "COMPAT_TOKEN"
```

That reloads as `value: Some("")`, and `apply_http_compat_headers` (`librefang-runtime-mcp/src/lib.rs:3363`) tests `value` **before** `value_env` — so the transport sends an empty header and never resolves the variable. Silent credential failure, nothing logged, and it fires on the default store for any `http_compat` server written through the API. Both fields now carry `#[serde(default, skip_serializing_if = "Option::is_none")]`.

This does not disturb the redaction design: the read route builds its JSON by hand in `http_compat_header_summary`, so the attribute only affects the storage serializer.

### `librefang-api` — a third defect, in the merge this PR itself adds

Mutation-probing `merge_http_compat_secrets` surfaced a case its own fixtures never reached: a header carrying *both* a static `value` and a `value_env`.

That combination is meaningful rather than redundant. `apply_http_compat_headers` (`librefang-runtime-mcp/src/lib.rs:3363`) resolves `value` first, so the static value is the live credential and the variable is a fallback that only takes effect once the static value is removed.

The first version of the merge skipped on `header.value.is_some() || header.value_env.is_some()`. The read route redacts `value` and emits `value_env`, so on submit-back such a header arrives carrying only its `value_env` — and that `value_env.is_some()` branch suppressed the restore. Observed directly:

```
stored BEFORE: value: Some("sk_h3ader…"), value_env: Some("COMPAT_TOKEN")
GET returned:  { "name": "X-Api-Key", "value_env": "COMPAT_TOKEN" }
stored AFTER:  value: None,             value_env: Some("COMPAT_TOKEN")
```

`200`, nothing logged, and the transport silently stops sending the operator's static credential and starts resolving the variable instead — the same class of silent credential swap the rest of this PR exists to prevent, introduced by the fix for it.

The condition now keys on `value` alone. A purely env-sourced header is unaffected either way, because the stored-value lookup is built with `filter_map(|h| h.value.as_ref()...)` and so has no entry for a name that never carried a `value` — the existing assertion that an env-only `Authorization` header comes back with `value: None` still holds.

## BEHAVIOUR CHANGE

**A stray key under `[mcp_servers.transport]` now stops the daemon from booting.** Not "that entry is skipped" — nothing starts.

The path is `try_into::<KernelConfig>()` Err → `crates/librefang-kernel/src/config.rs:237-243` → `crates/librefang-cli/src/commands/daemon.rs:334` `std::process::exit(1)`. `cmd_start` short-circuits on the load error deliberately, so the diagnostic naming the offending key reaches stderr instead of being swallowed by a tolerant `KernelConfig::default()`. Because the rejection happens inside serde it applies whether or not `strict_config` is set — the same way the parent entry has already behaved for a misspelled scalar such as `timout_secs`.

The reporter's own v2026.7.21 deployment has `[mcp_servers.transport.env]` in its `config.toml` and hits this on upgrade. That is the intended outcome (the alternative is the silent misconfiguration this issue is about), but it is a hard stop and the changelog fragment says so in those words.

**The hard stop is boot-only.** `POST /api/config/reload` maps the same error to a `400` and leaves the live config in place (`routes/config/manage.rs:990-993`), so a running daemon surfaces the typo without dropping its MCP connections. Worth stating because the asymmetry is actionable: on a running host, reload tells you about the mistake at no cost, while a restart on the same file will not come back up.

**`GET /api/mcp/servers` and `GET /api/mcp/servers/{name}` no longer emit `transport.headers[].source` or `transport.tools_count` for `http_compat` servers, and now emit `transport.tools[].input_schema`.** No consumer in the repo reads the two removed keys (evidence above). The addition is what makes read-modify-write non-destructive.

**`config.toml` written by the API no longer contains `value = ""` / `value_env = ""` lines under `[[mcp_servers.transport.headers]]`.** Reading an existing file that has them is unchanged — `""` still deserializes to `Some("")` — so this is not a breaking read change, but an operator with such a file on disk (written by a previous version) still has the broken empty-header behaviour until that server is rewritten through the API or the line is deleted by hand. Worth a look if an `http_compat` server has been sending unauthenticated requests.

## Verification

Every cargo invocation ran through the repo's sanctioned Linux dev image, scoped per crate.

Test-first on the regression: the new round-trip test was written before the fix and reproduced the reviewer's exact failure —

```
test every_transport_variant_survives_a_get_then_put_round_trip ... FAILED
PUT of the transport object GET returned for rt-http-compat was rejected: 400
  "unknown field `source`, expected one of `name`, `value`, `value_env`"
  left: 400  right: 200
```

- `cargo check -p librefang-types` — clean.
- `cargo check -p librefang-api` — clean.
- `cargo clippy -p librefang-types -p librefang-api --all-targets -- -D warnings` — exit 0, zero warnings. (Workspace-wide clippy is CI's lane.)
- `cargo fmt --check -p librefang-types -p librefang-api` — clean. The `let`-else tuple pattern in `merge_http_compat_secrets` needed a rustfmt-shaped rewrite, which the `pre-commit` hook would have caught.
- `cargo test -p librefang-types` — 896 + 5 + 49 + 4 + 2 passed, 0 failed.
- `cargo test -p librefang-api --test mcp_env_disclosure_test` — 6 passed, 0 failed.
- `cargo test -p librefang-api --lib upsert_mcp_server` — 4 passed, 0 failed (the `File`-store serialization path, including the new header test).
- `cargo test -p librefang-api --test skills_routes_test --test mcp_runtime_store_db_test` — 21 + 2 passed. These are the suites that assert the neighbouring `tools_count` / `source` keys (on the skills route, which is untouched) and the DB-backed store path the new tests write through.
- `cargo test -p librefang-runtime-mcp --lib` — 181 passed, 0 failed. This is the crate that consumes `HttpCompatHeaderConfig` at request time, so it covers the `skip_serializing_if` change's consumer side.
- `cargo test -p librefang-extensions --test docs_mcp_servers_examples --test everyapi_catalog_entry` — 1 + 4 passed. This parses every `[[mcp_servers]]` stanza in the en/zh docs tree through the now-guarded types, so a documented example the guard would reject fails here.

The full-target run for `librefang-api` (every test binary in the crate) is **not** claimed as passing: it died in the link step with `collect2: fatal error: ld terminated with signal 9 [Killed]`, an OOM from concurrent builds on a shared 16 GB Docker VM rather than anything in this diff. The per-target suites above cover every target this PR touches; the full lane is CI's.

Mutation-checked the new merge rather than trusting a green run. Removing the `header.value.is_some()` guard in `merge_http_compat_secrets` — the change that makes a stored value win over an explicitly submitted one, i.e. makes rotation impossible — produces:

```
test http_compat_header_values_rotate_remove_and_do_not_cross_a_transport_switch ... FAILED
assertion `left == right` failed: an explicitly submitted value must replace the stored one, or
a static header value could never be rotated:
  [HttpCompatHeaderConfig { name: "X-Api-Key", value: Some("sk_h3ader…"), value_env: None }]
```

The sabotage was reverted and the suite re-run to 6 passed.

### Tests added

In `crates/librefang-types/src/config/mod.rs`:

- `mcp_transport_rejects_unknown_key_in_transport_table_6612` — the reporter's exact TOML. Asserts on the serde diagnostic (`unknown field \`env\``) rather than the bare key name, because `toml`'s `Display` echoes the offending input span, so `contains("env")` would be satisfied by the source text and would keep passing with the guard removed.
- `mcp_transport_rejects_unknown_scalar_on_url_variants_6612` — `sse` and `http`, since the attribute binds per variant.
- `http_compat_transport_rejects_keys_outside_its_own_field_set_6612` — **new in this round**; the `HttpCompat` variant's own field set had no coverage. Uses `command` (a real field of the *stdio* variant, so an operator switching `type` used to leave an inert line behind) and a misspelled required `base_url`, where the error must name the typo rather than only reporting `base_url` missing.
- `mcp_transport_rejects_unknown_key_in_http_compat_headers_and_tools_6612` — the two nested structs.
- `every_valid_mcp_transport_variant_still_round_trips_6612` — drift sentinel; asserts the nested `headers` / `tools` arrays actually carry their entries, since `transport.is_some()` would pass even if both fell back to their `serde(default)` empty vectors.

In `crates/librefang-api/tests/mcp_env_disclosure_test.rs`:

- `every_transport_variant_survives_a_get_then_put_round_trip` — **the test that holds the fix.** The existing `round_tripping_the_redacted_env_preserves_the_stored_inline_value` hardcodes `{"type":"stdio",...}`, which is why it did not catch this. The new test covers all four variants.
  - Status alone does not prove losslessness — a serializer can be accepted-but-lossy, which is exactly what the omitted `input_schema` and header `value` were. So the fixture is seeded with a *non-default* `input_schema` (properties + `required`) and a static-`value` header, and the assertions read the persisted entry back via a new `stored_entry` helper rather than trusting the `200`.
  - Also asserts the static header value never appears in any read body, so the merge can never be "fixed" by disclosing the credential instead.
- `http_compat_header_values_rotate_remove_and_do_not_cross_a_transport_switch` — the three cases that would make the new merge a data trap rather than a safety net, which the round-trip test does not reach: an explicitly submitted value must win (or a static header value could never be rotated), a header dropped from the submission must be removed rather than merged back, and a credential must not survive a switch to `stdio` and reappear when the operator switches back to `http_compat`.

- `a_header_carrying_both_value_and_value_env_keeps_its_static_value_across_a_round_trip` — the case neither of the two tests above can reach, because both set exactly one of the two fields per header. Seeds a header with both, round-trips it, and asserts the static `value` survives while the `value_env` beside it is untouched. Also asserts the static value never appears in the `GET` body, so the fix cannot be satisfied by disclosing the credential.

In `crates/librefang-api/src/routes/skills/mod.rs` (unit, `File` store):

- `upsert_mcp_server_preserves_an_env_sourced_http_compat_header_6612` — the `serde_json` → TOML → disk → serde path the integration harness does not exercise, since it runs in `Db` mode. Asserts the written file contains no `value = ""` and that both header shapes reload with the absent `Option` still `None`. It fails on the pre-fix code with the on-disk TOML in the panic message, which is how the corruption above was found.

## Needs a maintainer call: the same bug is self-healable in `librefang-runtime-mcp`

The corruption fixed above has a second half that lives in a different crate, so I have not touched it. `librefang-runtime-mcp/src/lib.rs` disagrees with itself about what "this header has a value" means:

- `validate_http_compat_config:3164` — presence means **non-empty**: `header.value.as_ref().is_some_and(|v| !v.trim().is_empty())`
- `apply_http_compat_headers:3363` — presence means **`is_some()`**: `if let Some(value) = &header.value { value.clone() }`

So `value = Some("")` with `value_env = Some("TOKEN")` passes validation and then sends an empty header, never reading the variable. That gap is the mechanism of the bug this PR fixes at the serialization layer, and it stays live for any `config.toml` already on disk carrying a `value = ""` line written by an earlier version.

The fix is one line — `.filter(|v| !v.trim().is_empty())` on the apply path, aligning it with the validator's own definition of presence — and it would make every such existing file self-heal instead of requiring the operator to rewrite the server or hand-edit the line. Shipping it here would close the bug end to end rather than only at the serialization layer, but it lands in a third crate that this issue does not otherwise touch, so it is flagged rather than decided. **It can be folded into this PR on request; otherwise it needs a follow-up issue.**

## Pre-existing issues surfaced, not fixed

Three envelope-level omissions in `update_mcp_server`, all at the level *above* the transport and all pre-existing. Feeding a whole `GET` body into `PUT` was already a `400` before this PR, because `McpServerConfigEntry` has carried `deny_unknown_fields` since #5130 while the read routes add `connected` (`routes/skills/mcp.rs:326`), `auth_state`, and `tools` to the envelope. Related, `entry.headers` and `oauth` are not merged the way `env` now is.

Fixing any of them means changing a contract this issue does not touch, and both escape routes are worse than the disease: dropping `connected` / `auth_state` from the read routes breaks dashboard consumers, and a strip-list on `PUT` invents a new API rule. Flagging for a maintainer decision rather than deciding unilaterally.

Separately, `crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json` is now stale in a fourth way — `deny_unknown_fields` adds `additionalProperties: false` to the three types. It is not regenerated here: `kernel_config_schema_matches_golden_fixture` is already `#[ignore]`d for unrelated pre-existing 490-byte drift, so regenerating would fold that unrelated drift into this diff. It stays ignored either way.


houko:
Automated review pass. Reviewed the diff against `origin/main` and cross-checked via the GitHub API.

The `deny_unknown_fields` placement is correct: it lands on `McpTransportEntry` (per-variant, verified by `mcp_transport_rejects_unknown_scalar_on_url_variants_6612`) and on the two nested `http_compat` structs, not on anything broader that would break forward-compat elsewhere. The "lossless round-trip" claim is backed by real tests (`every_transport_variant_survives_a_get_then_put_round_trip`, `upsert_mcp_server_preserves_an_env_sourced_http_compat_header_6612`, `a_header_carrying_both_value_and_value_env_keeps_its_static_value_across_a_round_trip`) that assert on the persisted entry, not just on HTTP status, so the "accepted-but-lossy" failure mode is actually covered.

**Fix pushed:** `fc21bbe`
- `changelog.d/fixed/6612-mcp-transport-deny-unknown-fields.md:18` — the trailing tag was `(#6612) (@nevgenov)`: #6612 is the *issue* this PR closes (filed by @nevgenov, the reporter), not this PR's own number, and the release tooling matches the generated `- <PR title> (#N) (@author)` line against the last `(#N)` group on the bullet's last line to suppress it. With the issue number there instead of `#6666`, this PR would get a duplicate generated line in the release notes on top of the curated bullet. Corrected to `(#6666) (@houko)`.

**Left as a review comment (ambiguous, needs a maintainer call):** the `kernel_config_schema.golden.json` interaction with sibling PR #6664, which un-ignores `kernel_config_schema_matches_golden_fixture`. Not something I fixed given the merge-order dependency involved.


---

crates/librefang-types/src/config/types.rs
```diff
@@ -6236,8 +6247,18 @@ pub struct HttpCompatToolConfig {
...
+//
+// The attribute applies per variant on an internally-tagged enum: serde buffers the content and each variant rejects fields outside its own set, so the discriminating `type` key itself is still accepted.
+// That behaviour is not obvious from the attribute alone — `deny_unknown_fields` is documented as unsupported on *adjacently*- and *untagged*-enum containers — so it is pinned by `mcp_transport_rejects_unknown_key_in_transport_table_6612` rather than left to the reader to assume.
+//
+// Guarding the enum makes `GET /api/mcp/servers/{name}` → `PUT` load-bearing rather than merely conventional: any key the read route synthesises into the `transport` object that is not a real field now fails the write.
+// `serialize_mcp_transport` in `librefang-api` is therefore a faithful representation of the stored variant, not a display summary — see the comment on that function.
 #[derive(Debug, Clone, Serialize, Deserialize, schemars::JsonSchema)]
-#[serde(tag = "type", rename_all = "snake_case")]
+#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
```
houko, line 6261: Ambiguous / needs a maintainer call, not something to fix here.

Per the PR description, `crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json` is knowingly not regenerated for the `additionalProperties: false` this `deny_unknown_fields` adds to `McpTransportEntry`, `HttpCompatHeaderConfig`, and `HttpCompatToolConfig`, on the reasoning that `kernel_config_schema_matches_golden_fixture` is already `#[ignore]`d for unrelated drift, so it "stays ignored either way." That's correct against current `main`.

But sibling PR #6664 (same author, also open, also based on current `main`) removes that exact `#[ignore]` unconditionally. Whichever of #6664 / #6666 merges second will need a follow-up fixture regeneration to avoid breaking `kernel_config_schema_matches_golden_fixture` once it's live — right now neither PR's diff shows that being handled, and #6664's own body doesn't list #6666 among the "ordering note for reviewers" PRs it enumerates as needing a bump (it only names #6661 and #6663). Worth a maintainer decision on merge order / who owns the follow-up bump before both land.

---

changelog.d/fixed/6612-mcp-transport-deny-unknown-fields.md
```diff
@@ -0,0 +1,18 @@
...
+The runtime checks `value` before `value_env`, so it then sent an empty header and never resolved the variable — a silent credential failure with nothing logged.
+`value` and `value_env` now carry `skip_serializing_if`, the same fix already documented as load-bearing on the entry's own `template_id` and `oauth` fields for the identical reason.
+A header that carries *both* a static `value` and a `value_env` lost the static one on every read-modify-write, which the merge above initially reproduced rather than fixed: the read route redacts `value` and emits `value_env`, and the merge treated that returned `value_env` as "nothing to restore here".
+Since the runtime resolves `value` first, the header silently stopped sending the operator's static credential and started resolving the variable instead — a `200` with nothing logged and a different request on the wire.
+The merge now keys the decision on `value` alone, so the presence of a `value_env` beside it no longer suppresses the restore (#6666) (@houko)
```
houko, line 18: The content fix here (correcting the trailing tag from `(#6612) (@nevgenov)` to `(#6666) (@houko)`) is right, but the commit that made it — fc21bbe — has `Author: Claude <noreply@anthropic.com>` and `Committer: Claude <noreply@anthropic.com>`, not a human identity. That's the exact class of AI attribution CLAUDE.md bans from "commit messages, PR bodies, or comments" — the commit-msg hook only greps message text, so it didn't catch it, but the author/committer identity fields land in `git log` and, worse, in a GitHub "Squash and merge" of this PR: GitHub auto-generates `Co-authored-by:` trailers for every distinct author across the squashed commits, so merging as-is would inject `Co-authored-by: Claude <noreply@anthropic.com>` into the squash commit message on `main` regardless of what either commit's own message says.

Recommend amending fc21bbe with a real human `--author` before merge (or squashing locally with the author overridden) so the identity doesn't ride into the merge commit.

---

houko (commented):
Reviewed the deny_unknown_fields addition against CLAUDE.md's config/changelog/attribution rules. Code, tests, changelog fragment, and commit message all check out — one finding left as an inline comment (commit author identity, not diff content) since fixing it would require rewriting already-pushed history.

---

Issue #3572, opened 2026-04-27
security: default config encourages 0.0.0.0:4545 without enforced auth


houko:
## Summary
Init template comments suggest flipping the bind to `0.0.0.0:4545` while runtime only **warns** when the bind is non-loopback and `api_key` is empty. On hosts that share loopback (multi-tenant containers, Codespaces / Devcontainers exposing `127.0.0.1` to the host, SSH local-forwards from another user), this combination yields full unauthenticated admin access — including the RCE path in the hands install endpoint.

## Evidence
- `crates/librefang-cli/templates/init_default_config.toml:7`
  ```toml
  api_listen = "127.0.0.1:4545"  # Local-only; use "0.0.0.0:4545" to expose on LAN/remote
  ```
- Test fixtures bind `0.0.0.0:4545`: `crates/librefang-kernel/src/config.rs:359, 408, 610`
- `crates/librefang-api/src/server.rs:917-940` — only `warn!`s, doesn't refuse to start
- `middleware.rs:483-485` — loopback callers get anonymous-trusted access

## Suggested fix
Refuse to start (instead of warning) when the resolved bind is non-loopback AND `api_key` / `user_api_keys` / `dashboard_auth` are all empty AND `LIBREFANG_ALLOW_NO_AUTH` is unset. Remove the "use 0.0.0.0:4545" suggestion from the default template comment; require an explicit opt-in flag.

---
_Reported by automated multi-agent code review (security / concurrency / architecture / performance / testing). Evidence is grep-verified file:line; please re-confirm before fixing._

houko:
Fixed by #4203 (merged). Closing.

Issue #1034, opened 2026-03-17
Feature: Implement Provider Capability System


houko:
---
name: Feature request
about: Suggest an idea for this project
title: 'Feature: Implement Provider Capability System'
labels: 'enhancement'
assignees: ''
---

**Is your feature request related to a problem? Please describe.**
A clear description of what the problem is. Ex. I'm always frustrated when [...]

**Describe the solution you'd like**
A clear description of what you want to happen.

**Describe alternatives you've considered**
A clear description of alternative solutions or features you've considered.

**Additional context**

---

**Feature Request: Provider Capability System**

## Summary
Implement a provider capability system for defining API compatibility modes.

## Motivation
OpenClaw has provider capabilities for different API modes. Librefang should implement similar capability system.

### Phase 1: Capability Types
- ProviderCapabilities type
- anthropicToolSchemaMode
- anthropicToolChoiceMode
- providerFamily

### Phase 2: Registry
- Provider registry
- Capability lookup
- Default capabilities

### Phase 3: Features
- Model-specific hints
- Tool call modes
- Thinking signatures

## Acceptance Criteria
- [ ] ProviderCapabilities type
- [ ] Provider registry
- [ ] Capability lookup
- [ ] Model hints

## References
- OpenClaw: `src/agents/provider-capabilities.ts` - Provider capabilities
- OpenClaw: `src/providers/` - Provider implementations

PR #6667, opened 2026-07-30
feat(api): give the master api_key env/vault indirection and a hashed form, and close the hash-only WS/terminal auth bypass

### LINKED ISSUE #6613: The daemon's own api_key is the only credential class with neither env indirection nor hashing — plaintext in config.toml is the only option
`KernelConfig.api_key` is the master credential for the whole API — with it, every endpoint outside `/api/health` is reachable. It is also the one credential in the config that has no alternative to being stored in cleartext on disk.

```rust
/// API authentication key. When set, all API endpoints (except /api/health)
/// require a `Authorization: Bearer <key>` header.
/// If empty, the API is unauthenticated (local development only).
pub api_key: String,
```

Searched current `main` (`cb7608d9`) for an escape hatch and found none: no `LIBREFANG_API_KEY` read anywhere in `crates/*/src`, no vault resolution for this field at boot, no `*_env` sibling. The middleware reads the configured value directly. So the only way to run an authenticated daemon is to have the working secret sitting in `config.toml`.

## What makes this stand out is that both mechanisms already exist here

**Env indirection** — `api_key_env` appears nine times across the config types (default model, provider entries, credential pools, TTS, embeddings, …). Every third-party credential the daemon consumes can be sourced from the environment, and therefore from the vault via the existing loader. The daemon's own key is the exception.

**Hashing at rest** — two other credential classes are already hashed:

- `UserConfig.api_key_hash: Option<String>` — per-user API keys are stored hashed.
- `dashboard_pass_hash` — and `crates/librefang-api/src/password_hash.rs` documents a complete migration path that was already executed for exactly this shape of problem:

  > Supports transparent migration from legacy plaintext passwords:
  > - If `dashboard_pass_hash` is set (Argon2id PHC string), verify against it.
  > - If only `dashboard_pass` is set (plaintext/vault), fall back to constant-time plaintext comparison and return the Argon2id hash for transparent upgrade.

So the project has already decided, twice, that a verifiable credential should not sit in cleartext — and has shipped the transparent-upgrade pattern to get there without breaking existing installs. The master key just never got the same treatment.

## Why it matters in practice

Verification is a comparison against a presented bearer token, so nothing about the daemon's use of this value requires it to be recoverable — `api_key_hash` would work exactly like `dashboard_pass_hash`.

The client side is a separate need: local MCP servers and scripts have to *send* the key, so they need the plaintext from somewhere. That is what `api_key_env` would solve — the operator puts the value in the vault and references it by name, the same way every provider key already works.

Concretely, on our host the value appears in cleartext in `config.toml`, in an MCP server's script as a hardcoded fallback default, and in that script's header documentation — and the config file shipped with mode `0644` until we noticed and tightened it. None of that is the daemon's fault, but none of it has a supported alternative either.

## Suggested fix

Either half is useful on its own; together they close it:

1. **`api_key_hash`**, following `password_hash.rs` verbatim — including the transparent upgrade, so an existing plaintext `api_key` keeps working and is replaced by its hash on first successful auth.
2. **`api_key_env`**, following the nine existing sites, so the working value can live in the vault and be referenced by name for the clients that must transmit it.

If there is a reason the master key was deliberately left out of both — some bootstrap ordering constraint where auth must be live before the vault is unlocked, say — that would be worth a comment on the field, since the current shape reads as an oversight rather than a decision.

## Environment

- Deployed `v2026.7.21`; field and middleware unchanged on current `main` (`cb7608d9`).

Happy to test a patch.

houko:
`KernelConfig.api_key` was the one credential class in the config with neither env indirection nor a hashed form, so the only way to run an authenticated daemon was to keep the working secret in cleartext in `config.toml` — a file the daemon itself rewrites, so it can never be a read-only Kubernetes Secret mount. Both mechanisms already existed for other credentials (`api_key_env` on nine provider fields, `dashboard_pass_hash` and `users[].api_key_hash` for hashing at rest, including the transparent-upgrade pattern). This adds both to the master key.

Closes #6613

### `api_key` env / vault indirection

- `api_key` now resolves through the same three steps the dashboard credentials use: `LIBREFANG_API_KEY`, then a `vault:NAME` prefix, then the literal value. `resolve_dashboard_credential` is renamed `resolve_credential` since the mechanism was never dashboard-specific.
- Resolution happens per auth snapshot rather than once at boot, so `POST /api/config/reload` no longer clobbers an env-sourced or vault-sourced key by re-reading `config.toml` from disk.
- The outbound leg is fixed too: `build_mcp_bridge_cfg` (the bearer CLI drivers send back to the daemon's own `/mcp`) now applies the same env / `vault:` resolution. It previously handed the bridge the literal string `vault:name`, which no daemon would accept, and after a reload handed it an empty key on an env-only deployment.

### `api_key_hash`

- New `KernelConfig.api_key_hash`, verified after the constant-time comparison against `api_key` misses. Accepts `$sha256$…` and `$argon2id$…`, dispatched by prefix exactly as `users[].api_key_hash` already is.
- Redacted in the `Debug` impl. The hash is the verifier: anyone who reads it can paste it into their own config and authenticate.
- Transparent upgrade: the first request that authenticates against a plaintext-only `api_key` writes a hash of it to a mode-0600 `$LIBREFANG_HOME/api-key-hash.upgrade-hint` and logs a pointer to the file. Gated once per process, and gated on the presented token actually being the master key — a dashboard session token also rides in the `api_key_lock` composite, and hashing one into an `api_key_hash` hint would tell the operator to configure a credential they never chose and that expires.
- New `librefang hash-api-key` produces the value. `--generate` mints a fresh 256-bit key and prints it beside its hash; without the flag it hashes a key you already have. Recommending a format that no tool in the repo could produce would not have been advice.

### SHA-256 for the master key, Argon2id for the dashboard password

The two look inconsistent and are not, so the reasoning is recorded in a doc comment on `middleware::MasterKeyState` and on the `api_key_hash` field rather than left implicit:

- `dashboard_pass` is human-chosen and low-entropy. A memory-hard KDF is what makes an offline dictionary attack uneconomic, so Argon2id is correct there and stays.
- The master `api_key` is a machine-generated bearer token. There is no dictionary to enumerate, so the KDF buys nothing against an offline attacker and charges its cost to every request instead. That cost is ~50-100 ms of CPU per verify by construction, and with `api_key` unset *every* presented token reaches the verify — including every wrong one, from an unauthenticated caller, on paths with no login-attempt limiter. A fast constant-time hash over a high-entropy secret is the standard bearer construction, and it is already what this codebase reached for: `hash_device_token` hashes paired-device bearers with `$sha256$` and argues the same point.

Applied coherently: the upgrade hint writes `$sha256$`, `hash-api-key` produces `$sha256$`, the field docs and the ops table recommend `$sha256$`, and `$argon2id$` stays accepted for a hand-written value or a deliberately short human-memorable key.

- `master_hash_matches` is now `async` and dispatches on the stored hash via the new `password_hash::is_cheap_to_verify`: `$sha256$` verifies inline, `$argon2id$` goes to `spawn_blocking`. So whichever format an operator picks, the verify never stalls a tokio worker — which also closes the remote CPU-exhaustion shape, since a garbage bearer against a `$sha256$` deployment now costs one SHA-256.

### WebSocket / terminal auth bypass (the reason this needed a second pass)

Both upgrade paths derived "is auth configured?" as `!valid_api_tokens(..).is_empty()`. That was a sound proxy only while `api_key` was the single way to configure a master key: a daemon whose key exists solely as `api_key_hash` lists no plaintext token, so the expression reported it as unauthenticated. The terminal path then drove `decide_auth` past its reject branch into `LocalBypass` — unauthenticated shell access on a daemon the operator believes is bearer-gated — and the WebSocket path skipped its whole auth block, re-opening the openfang #1034 B2 branch for the new config shape. A caller presenting the *correct* key was rejected at the same time, because there was nothing to compare against.

- Both sites now share one `server::master_auth_required` derivation and fall back to verifying the presented token against `api_key_hash`.
- That derivation takes the **live auth handles** (`api_key_lock` + `MasterKeyState::hash`) — the same two the HTTP middleware reads — instead of re-resolving from a config snapshot. Re-resolving per connection meant a `vault:NAME` master key cost a `CredentialVault` construction, an OS keyring read, and an AEAD file decrypt on every WS and terminal upgrade, twice each (once directly, once inside `valid_api_tokens`), on paths reachable before any credential has been presented. The handles are rewritten from one snapshot by `refresh_master_credential`, so they track the same reload generation without the per-connection cost.
- The `\n`-joined-composite compare is now one function, `server::matches_master_token`, used by all three surfaces instead of being spelled out inline in each.

### Reload reaches the middleware

`api_key_lock` was written only at boot and on a dashboard credential change, so an `api_key` edited in `config.toml` kept authenticating with the old value until restart — even though `build_reload_plan` classified the field as effective-immediately. `refresh_master_credential` now runs on `POST /api/config/reload`, on the config-file watcher tick, and on a dashboard credential change, writing both handles from one snapshot so a reader can never see a new plaintext beside a stale hash.

### Callers that read `api_key` directly and were wrong for the new shapes

Each of these reported "no auth" for a daemon that is in fact bearer-gated:

- `server.rs::any_auth_configured` (the #3572 boot-time bind guard) — would have refused to start, or started open, on a hash-only or env-only deployment.
- `routes/config/security.rs::security_status` — reported `localhost_only`.
- `server.rs::dashboard_auth_check` — reported mode `none` / `credentials`, hiding the API-key entry field the operator needs. It also open-coded the dashboard-credential test and read `config_ref()` rather than `auth_snapshot()`, observing a different hot-reload generation than the middleware enforces; both halves now go through the shared helpers. This endpoint is on the unauthenticated `is_public` allowlist, so it takes the master credential from the live handle (`MasterKeyState::is_configured`) rather than re-resolving — otherwise an anonymous caller could drive a keyring read plus an AEAD vault decrypt per request. Same for `security_status`.
- `librefang-cli status --verbose` — told an operator who had migrated to `api_key_hash` that their daemon had no auth at all.
- `evaluate_bind_auth_safety`'s refusal message listed only `api_key`, dashboard credentials, and `[[users]].api_key_hash`, while `any_auth_configured` directly above it also accepts a master `api_key_hash`, an env-sourced key, and paired devices. It now names every accepted form, with a test that fails if the two drift again.

## Verification

Run locally through the repo's sanctioned Docker dev image (`Dockerfile.rust-dev`, per-worktree target volume), scoped per CLAUDE.md — no unscoped workspace runs:

```
cargo check -p librefang-api --lib                     # Finished
cargo check -p librefang-types --lib                   # Finished
cargo check -p librefang-cli --bins                    # Finished
cargo check -p librefang-api -p librefang-testing --tests   # Finished
cargo test  -p librefang-api --lib                     # 916 passed; 0 failed
```

Remaining lanes (`librefang-kernel`, the integration binaries including the new `hash_only_master_key_ws_auth`, `librefang-cli`'s `i18n_checks`, workspace clippy) are left to CI. `cargo check --tests` compiles all of them, so this is a run-time rather than a compile-time gap. The shared 16 GB Docker VM carried up to twelve concurrent cargo jobs from parallel sessions during this work and one link step was already OOM-killed (`collect2: fatal error: ld terminated with signal 9`); serialising more link-heavy runs locally buys nothing CI does not do better.

New tests, by what they pin:

`middleware.rs::tests` — the auth middleware against a hash-only snapshot:
- `hash_only_master_key_authenticates_the_matching_bearer`
- `hash_only_master_key_rejects_a_wrong_bearer`
- `hash_only_master_key_rejects_a_missing_bearer_from_loopback`
- `hash_only_master_key_rejects_an_empty_bearer`

`middleware.rs::tests` — the transparent-upgrade machinery, which had no coverage at all:
- `upgrade_hint_names_both_config_fields_and_verifies_the_key`
- `upgrade_hint_writes_the_cheap_to_verify_hash_format`
- `upgrade_hint_file_is_owner_only` (unix)
- `upgrade_hint_is_written_at_most_once_per_process`
- `upgrade_hint_is_skipped_without_a_daemon_home`
- `a_dashboard_session_token_is_not_the_master_plaintext_key`
- `hash_only_daemon_never_reports_a_master_plaintext_match`
- `authenticating_with_a_plaintext_master_key_leaves_an_upgrade_hint`
- `a_hash_configured_daemon_writes_no_upgrade_hint`

`routes/terminal.rs::hash_only_terminal_auth_tests` — `authorize_terminal_request` itself, not just the pure `decide_auth` it calls, because the bypass was in the argument it passed:
- `hash_only_master_key_authenticates_the_matching_bearer`
- `hash_only_master_key_rejects_a_wrong_bearer`
- `hash_only_master_key_rejects_a_missing_bearer_from_loopback` (pre-fix this returned `Ok(LocalBypass)`)
- `hash_only_master_key_rejects_an_empty_bearer`

`tests/hash_only_master_key_ws_auth.rs` — `agent_ws` over a real socket, mounted with no auth middleware in front so every status is the handler's own verdict:
- `hash_only_master_key_authenticates_a_websocket_upgrade`
- `hash_only_master_key_rejects_a_wrong_bearer_on_a_websocket_upgrade`
- `hash_only_master_key_rejects_a_missing_bearer_from_loopback` (pre-fix this reached the agent lookup anonymously)
- `hash_only_master_key_rejects_an_empty_bearer_on_a_websocket_upgrade`

`server.rs::master_credential_tests`:
- `hash_only_master_key_lists_no_plaintext_token` (pins the trap the module guards)
- `hash_only_master_key_counts_as_configured`
- `no_credential_at_all_is_not_configured`
- `master_hash_matches_only_the_right_token`
- `empty_hash_never_matches`
- `argon2id_master_hash_still_verifies_off_the_async_worker` (multi-thread flavour is load-bearing: it exercises the `spawn_blocking` branch)
- `matches_master_token_handles_the_composite_and_rejects_empty`
- `master_auth_required_agrees_with_master_credential_is_configured`
- `refresh_master_credential_replaces_both_handles`
- `plaintext_master_key_still_lists_its_token`

`server.rs::bind_auth_safety_tests`:
- `refusal_names_every_form_of_auth_that_would_satisfy_it`

`password_hash.rs::tests`:
- `generated_bearer_token_is_256_bits_of_hex_and_hashes_cheaply`
- `is_cheap_to_verify_tracks_the_prefix_verify_password_dispatches_on`
- `is_cheap_to_verify_treats_garbage_as_expensive`

Two test-harness notes worth recording, because both are ways a test can pass for the wrong reason. Every hash-only fixture sets **both** `KernelConfig.api_key_hash` (what `auth_snapshot()` reports) and the live `master_key` handle (what the upgrade paths read) — production keeps them in step via `refresh_master_credential`, and setting one leaves the handler seeing an unconfigured daemon. And every fixture presents a **loopback** peer, because loopback is the origin the no-auth branch treats as trusted and therefore the only place the bypass was reachable; with a remote peer the request is rejected on the remote-policy rules regardless, and the test would still pass with the fix reverted. `TestAppState::with_api_key_hash` — added in the first pass and called nowhere — is now used by the WS test, with its doc comment stating that it writes the live handles only.

`TestAppState::build_state` now seeds `api_key_lock` and `master_key` from the test's `KernelConfig`, mirroring what `refresh_master_credential` does at boot. Without it a fixture that sets `cfg.api_key` through `MockKernelBuilder::with_config` alone left both handles empty, so every surface that reads them saw an unconfigured daemon while `auth_snapshot()` reported a configured one — reproducing inside the harness the exact split this PR is about. Env / `vault:` resolution is deliberately *not* applied there: a fixture's credential should be what its config literally says, not what the developer's environment and on-disk vault happen to hold.

## BEHAVIOUR CHANGE

- **The transparent upgrade writes `$sha256$`, not Argon2id.** An operator following the hint file lands on the cheap verifier. Anyone who already pasted an `$argon2id$` value into `api_key_hash` keeps working unchanged — the format stays accepted, and is now verified on a blocking thread rather than inline.
- **Rotating a `vault:NAME` master key takes effect on `POST /api/config/reload` rather than on the next WebSocket / terminal connection.** `librefang vault set` writes `vault.enc`, and the config watcher polls `config.toml`'s mtime, so nothing notices the rotation on its own. This is the posture the HTTP middleware always had (`api_key_lock` was resolved once and swapped on reload); the two upgrade paths now agree with it instead of re-resolving per connection. A reload suffices — no restart. Documented as a gotcha section in `docs/operations/config-reload.md`, alongside the stronger case that `LIBREFANG_API_KEY` needs a restart because a process cannot observe an edit to its own environment.
- **The non-loopback bind refusal message changed text.** It now names every form of auth `any_auth_configured` accepts (env var, `vault:`, `api_key_hash`, paired devices) rather than only three of them. Anything scraping the old string will not match.
- **`GET /api/auth/dashboard-check` and `GET /api/config/security-status` now report a hash-only or env-sourced deployment as authenticated.** They previously reported `none` / `credentials` and `localhost_only` respectively. That was the bug, but it is an observable response change.
- **A daemon that binds non-loopback with only `api_key_hash` or `LIBREFANG_API_KEY` configured now starts.** It previously hit the #3572 refusal, since the guard read `api_key` directly.

## Not regenerated: `kernel_config_schema.golden.json`

Adding `api_key_hash` to `KernelConfig` changes the schemars-derived schema, so the golden fixture is now stale by that field. It is **not** regenerated here, on purpose: `kernel_config_schema_matches_golden_fixture` is already `#[ignore]`d on `main` for unrelated pre-existing drift ("drifted by 490 bytes", per the comment above it), so regenerating would fold someone else's un-reviewed diff into this PR under cover of my one-field change — the opposite of what the fixture exists for. The assertion does not run in CI while the `#[ignore]` stands, so nothing breaks. `config_schema_overlay`, which *does* run, is updated with the new field.

## Deferred

Nothing from this review. Two adjacent observations, both genuinely outside the crates touched here and neither introduced by this change:

- There is no login-attempt rate limiting on non-`/api/auth/*` paths. The KDF change removes the amplification that made it acute for `api_key_hash` (a wrong bearer now costs one SHA-256), and `users[].api_key_hash` values are Argon2id on that same unlimited path — but that is the per-user auth surface in `middleware.rs`'s user-key loop with its own migration story, not the master credential this issue is about. Worth a separate issue; naming it here rather than silently leaving it.
- `dashboard_pass_hash` has no `hash-api-key`-style generator that also mints the secret; `hash-password` prompts for one you invented. That is correct for a password and wrong for a key, which is why `hash-api-key` is a sibling command rather than a flag.


changelog.d/added/6613-api-key-hash-and-env.md
houko, line 1: Minor, non-security nit: both fragments in this PR (`added/6613-api-key-hash-and-env.md` and `security/6613-hash-only-ws-terminal-auth.md`) end with `(@nevgenov)`, but this PR was opened by `houko`.
Per `changelog.d/README.md`, the trailing tag should be `(@your-github-login)` — the author of *this* PR.
Since I can't independently confirm which login is correct here, I'm flagging rather than editing it myself; worth a quick check before merge so the release-notes attribution lands on the right person.

---

houko (commented):
Automated review of #6667. Focused on the secret-handling surface: `password_hash.rs`, `middleware.rs::MasterKeyState`, `server.rs` (resolve_credential / master_hash_matches / matches_master_token / refresh_master_credential / write_upgrade_hint), `ws.rs`, and `routes/terminal.rs`.

Security assessment: sound. SHA-256-for-bearer vs Argon2id-for-password choice is justified and documented; `verify_sha256_token` / `matches_master_token` use `subtle::ConstantTimeEq`; the upgrade-hint file is written via atomic temp+rename at mode 0600 with no TOCTOU window; the hash/plaintext are never logged (only vault key *names* and error strings appear in `tracing::warn!`); Argon2id verification is moved to `spawn_blocking` so a hash-only deployment can't be used to stall tokio workers. The WS/terminal auth-bypass fix (`master_auth_required` / `matches_master_token` / `master_hash_matches` shared across middleware, ws.rs, and terminal.rs) is correct and has dedicated regression tests (`hash_only_master_key_ws_auth.rs`, `hash_only_terminal_auth_tests`) that would have caught the pre-fix `LocalBypass`/skip-auth-block bug. `Default` impl and `Debug` redaction for the new `api_key_hash` config field are both present; `config_reload.rs` classifies it; changelog fragments (`added/` and `security/`) exist.

One nit left as an inline comment: the changelog fragment attribution reads `(@nevgenov)` but the PR was opened by `houko` — worth confirming before merge.

No other findings. No commits pushed by this review pass.

---

crates/librefang-kernel/src/kernel/mod.rs
```diff
@@ -327,14 +334,54 @@ fn build_mcp_bridge_cfg(cfg: &KernelConfig) -> librefang_llm_driver::McpBridgeCo
...
+    // Same precedence as the API layer's `resolve_credential`: env, then
+    // `vault:`, then the literal.
+    let api_key = match std::env::var(API_KEY_ENV) {
+        Ok(key) if !key.trim().is_empty() => key.trim().to_string(),
+        _ => resolve_vault_prefixed(cfg.api_key.trim(), &cfg.home_dir),
     };
     librefang_llm_driver::McpBridgeConfig {
         base_url: base,
-        api_key,
+        api_key: (!api_key.is_empty()).then_some(api_key),
```
houko, line 356: Architectural question rather than a bug I'm confident is wrong, so flagging for a maintainer call rather than pushing a fix.

`build_mcp_bridge_cfg` resolves the outbound bearer the same way `resolve_credential` resolves the inbound one — env, then `vault:`, then the literal `cfg.api_key` — but it never looks at `cfg.api_key_hash`. A hash is a verifier, not a secret, so there's nothing here it *could* recover from a hash-only config; that's a hard limitation, not an oversight in this function specifically.

The question is what happens next when an operator follows this PR's own recommended path for `api_key_hash`: "Put the hash here, give the key to your clients, and leave `api_key` unset" (per the new field doc), or completes the transparent upgrade hint by "remove `api_key`" after copying the hash in. On a daemon left in that state — no `api_key`, no `LIBREFANG_API_KEY`, no `vault:` reference, only `api_key_hash` — `api_key` here resolves to `""`, `McpBridgeConfig.api_key` becomes `None`, and every CLI-based driver (Claude Code) self-call to the daemon's own `/mcp` endpoint goes out with no `Authorization` header. Since the whole point of this PR is that the auth middleware now correctly treats hash-only as configured, that self-call gets a 401 from the daemon's own middleware.

Two things make this worth a second look rather than assuming it's fine:
- `resolve_vault_prefixed` only `warn!`s when a `vault:NAME` reference fails to resolve (lines 376-384). The "no plaintext configured at all" case — the exact one the recommended hash-only posture produces — falls through silently to `String::new()` with no log line anywhere in this path.
- I don't see a test exercising `build_mcp_bridge_cfg` (or an integration test) against a hash-only `KernelConfig`, so this wouldn't currently be caught by CI either.

If this is a known, accepted limitation (i.e., "hash-only is for the HTTP/WS/terminal surfaces; keep `api_key` or a `vault:` reference too if you use CLI-based drivers"), it might be worth a line in the `api_key_hash` field doc and/or the CLI's `hash-api-key` output, since right now both explicitly tell the operator to drop `api_key` entirely. If it's not accepted, a `warn!` when the bridge ends up keyless while `api_key_hash` is non-empty would at least surface it instead of failing silently on every driver tool call.

---

houko (commented):
Reviewed the auth/security-sensitive core of this PR closely: `middleware.rs` (`MasterKeyState`, the transparent upgrade, the auth-configured derivation), `password_hash.rs` (SHA-256 vs Argon2id dispatch), `server.rs` (`master_credential`, `matches_master_token`, `master_hash_matches`, `refresh_master_credential`, `evaluate_bind_auth_safety`), `ws.rs` and `routes/terminal.rs` (the WS/terminal bypass fix itself), the new `KernelConfig.api_key_hash` field (struct + `#[serde(default)]` + `Default` impl + `Debug` redaction, all present), and the config-reload wiring.

The fix itself looks solid: `matches_master_token` and `verify_sha256_token` both use `subtle::ConstantTimeEq`, empty candidates are filtered out of the composite so an empty bearer can't match an empty stored token, the Argon2id path is moved to `spawn_blocking` so a hash-only deployment can't stall the async runtime, and — most importantly for the stated bug — both the WS upgrade (`ws.rs::agent_ws`) and the terminal upgrade (`routes/terminal.rs::authorize_terminal_request`) now derive `auth_required`/`auth_configured` via the same `server::master_auth_required` the HTTP middleware effectively uses, instead of the old `!valid_api_tokens(..).is_empty()` proxy that reported a hash-only daemon as unauthenticated. `tests/hash_only_master_key_ws_auth.rs` and `routes/terminal.rs::hash_only_terminal_auth_tests` both drive the real handlers (not just the pure `decide_auth`) over a loopback peer with a hash-only credential and assert the pre-fix `LocalBypass`/unauthenticated-upgrade case now 401s — that's the right regression test for this class of bug.

One left one inline comment (ambiguous/architectural, not a confirmed bug) about `build_mcp_bridge_cfg` silently producing a keyless outbound bridge config when an operator follows this PR's own recommended hash-only posture (`api_key_hash` set, `api_key` removed) — worth a maintainer call on whether that's an accepted limitation that should be documented, or a gap worth a warning log.

No mechanical fixes pushed — nothing met the bar of a small, ≥90%-confident, local bug in this diff.

---

crates/librefang-api/src/server.rs
```diff
@@ -168,32 +217,134 @@ pub(crate) fn dashboard_session_token(snap: &ApiAuthSnapshot) -> Option<String>
...
+///
+/// Call after anything that can change the effective master credential:
+/// `POST /api/config/reload`, the config-file watcher, and the dashboard
+/// credential-change endpoint (which alters the derived session token that
+/// rides in the same list).
+///
+/// This is the *only* place the env / `vault:` indirection is re-run, which makes it the boundary for one operational fact worth stating plainly: rotating a `vault:NAME` master key with `librefang vault set` writes `vault.enc`, not `config.toml`, so nothing here notices until the operator calls `POST /api/config/reload`.
+/// That matches the posture the HTTP middleware has always had — `api_key_lock` was likewise resolved once and swapped on reload — and since #6613 the WS and terminal upgrade paths agree with it instead of re-resolving per connection.
+/// A reload suffices; a daemon restart is not needed.
+pub(crate) async fn refresh_master_credential(
```
houko, line 333: Architectural question rather than a confirmed bug, so flagging for a maintainer call rather than touching the credential-refresh wiring myself.

`refresh_master_credential` is only called after `state.kernel.reload_config()` at two sites: the dedicated `POST /api/config/reload` handler (`routes/config/manage.rs`) and `change_password` (`server.rs`), plus the 30s background file-watcher (`run_daemon`) as a catch-all.

`reload_config()` itself is called from several other places that don't refresh the two live handles afterward:
- `routes/config/manage.rs` (the `PUT /api/config` raw-write handler, a second call site further down the file from the one that was patched)
- `routes/budget.rs` (`persist_budget`)
- `routes/users.rs` (rotate-key / import-users)
- `routes/skills/mcp.rs` (two call sites, MCP server add/remove)
- `routes/skills/extensions.rs` (two call sites, enable/disable)
- `routes/config/system.rs`
- `routes/channels.rs` (two call sites)
- `routes/memory.rs`

Any of these writing `config.toml` with an edited `api_key` / `api_key_hash` (e.g. an operator using the dashboard's raw config editor, which goes through `PUT /api/config`) will get picked up by `reload_config()` into the kernel's live `KernelConfig`, but `api_key_lock` / `master_key` stay stale until the 30s file-watcher poll notices the mtime change and does its own reload — because the watcher gates only on mtime (`current != last_modified`), not on whether the write already went through `reload_config()`.

So there's a bounded (~30s) window, not a permanent bypass, but it's the same class of staleness #6613 itself was filed to fix for the dedicated reload endpoint, and `docs/operations/config-reload.md`'s new `api_key_hash` / `api_key` rows say "effective immediately via config swap" without qualifying that this only holds for the two call sites that got the explicit refresh.

Worth a maintainer decision: is the ~30s window across these other endpoints acceptable, or should the refresh move inside `reload_config()` itself (kernel-side) so no API-crate caller can forget it? Given `MasterKeyState` lives in `librefang-api`, that would need the trait-injection pattern this repo already uses for kernel↔runtime boundaries.

---

crates/librefang-api/src/routes/config/security.rs
```diff
@@ -17,9 +17,13 @@ use super::*;
...
-    let scfg = state.kernel.config_ref();
-    let api_key_empty = scfg.api_key.is_empty();
-    drop(scfg);
+    // Ask the live master-credential handle — the same one the auth middleware
+    // verifies against (#6613). Reading `api_key` alone reported
+    // `localhost_only` for a hash-only or env-sourced deployment that is in
+    // fact fully bearer-gated; re-resolving the field from a config snapshot
+    // would report correctly but unlock the vault per request for a
+    // `vault:NAME` value, to answer a yes/no question the handle already holds.
+    let api_key_empty = !state.master_key.is_configured().await;
```
houko, line 26: Minor test-coverage gap, flagging rather than adding the test myself since it's on the auth-reporting surface.

This handler (and `dashboard_auth_check` in `server.rs`, same pattern) now derives its answer from `state.master_key.is_configured()` instead of the raw `api_key` field, specifically to fix mis-reporting for a hash-only or env/vault-sourced deployment. The underlying primitive (`MasterKeyState::is_configured`) is unit-tested, and the auth *enforcement* path (middleware / WS / terminal) all get dedicated hash-only `#[tokio::test]`s in this PR — but I don't see an equivalent `TestServer`-based integration test hitting `GET /api/config/security-status` (or the dashboard auth-check endpoint) against a hash-only `KernelConfig` to confirm the reported `auth_mode` flips from `localhost_only` to bearer-gated. Per CLAUDE.md's integration-testing mandate this is a wiring change to an existing endpoint, so it'd be good to pin the response shape end-to-end the same way `hash_only_master_key_ws_auth.rs` does for the WS path.

---

PR #6646, opened 2026-07-29
feat: add EveryAPI partner surfaces

houko:
## Summary
- add official EveryAPI partner links to the website, README, dashboard, and documentation navigation
- add English and Chinese partner documentation for managed credential discovery, live models and pricing, and MCP account tools
- correct the provider guide to use `everyapi auth login` and the current credential-process flow
- point EveryAPI CLI references at the public `everyapi-ai/everyapi-ai` repository
- eliminate Dashboard lint and production chunk warnings found while verifying the partner UI
- fix unstable React hook dependencies, deterministic test storage, and pnpm workspace configuration at their sources

## Verification
- website lint, 39 tests, and production build passed
- dashboard ESLint passed with zero warnings
- dashboard TypeScript typecheck passed
- dashboard full suite passed: 91 files, 936 tests
- dashboard production build passed with no chunk-size warning
- dashboard diff whitespace check passed
- docs production build generated 144 static pages, including both partner routes
- all public CTA destinations returned HTTP 200

`cargo check --workspace --lib` was attempted, but its container exited without diagnostics. It is not reported as passing. No Rust source changed.

## Out of scope
- deployment of the new documentation route
- EveryAPI private-repository changes, covered by its separate PR
- dependency-security upgrades that require coordinated parent-package and Node-runtime changes

The Dashboard warning cleanup is intentionally included at the maintainer request made during this PR. It does not add lint suppressions or raise warning thresholds.

Issue #4732, opened 2026-05-07
fix(kernel): SSRF validation gaps in cron delivery webhook URLs


leszek3737:
## Problem
`is_blocked_webhook_host` in `crates/librefang-types/src/scheduler.rs` blocks only a subset of local/loopback targets. Missing:

- `0.0.0.0` (unspecified)
- Private ranges: `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`
- Hex IP: `0x7f000001`
- Decimal IP: `2130706433`
- IPv4-mapped IPv6: `::ffff:127.0.0.1`, `::ffff:7f000001`
- Octal IP: `0177.0.0.1`

Additionally:
- `CronDelivery::Webhook` `validate_delivery` only checks scheme + length (`scheduler.rs:668-678`)
- `update_job` / `set_delivery_targets` skip `validate_delivery_targets` entirely (`cron.rs:296-305`, `cron.rs:318-326`)
- `webhook_http_ok` test accepts `http://localhost:8080/hook` (`scheduler.rs:1262-1268`)

## Impact
Any cron job with `CronDeliveryTarget::Webhook` can target loopback, private, or cloud metadata endpoints.

## Suggested fix
1. Use real URL/IP parser, normalize host
2. Block: loopback, unspecified, link-local, private, metadata IPs in all forms (hex, decimal, octal, IPv4-mapped IPv6)
3. Validate on both create AND update/set paths
4. Centralize validation for `CronDeliveryTarget::Webhook` and `CronDelivery::Webhook`
5. Fix tests to reject `localhost`

## References
- Review of `refaktor/dashboard4` PR identified this as blocker for UI blocklist port
- UI blocklist ported as UX-only feedback (not security boundary)

PR #6668, opened 2026-07-30
fix(dashboard): restore clipboard copy on plain HTTP, and close four read/write gaps behind the dashboard's config, schedule, and session views

### LINKED ISSUE #6611: Dashboard papercuts: copy buttons silently no-op over plain HTTP, plus three read-surface inconsistencies
Four small, independent issues found while using and re-auditing the dashboard. Bundled because each is individually minor; the first is the one that actually bites daily.

## 1. Copy buttons do nothing when the dashboard is served over plain HTTP on a LAN/VPN address

Clicking the copy icon next to a config field produces no clipboard content and no "copied" feedback.

The repo already contains the fix for exactly this, in `crates/librefang-api/dashboard/src/lib/clipboard.ts`, whose own doc comment describes the situation precisely:

> `navigator.clipboard` is only defined in secure contexts (HTTPS or localhost). When the dashboard is served over plain HTTP on a LAN IP (`http://192.168.x.x:4545`) the clipboard API is `undefined`, so a bare `navigator.clipboard.writeText(...)` call throws with no feedback to the user. Fall back to the legacy `document.execCommand('copy')` path via a detached textarea so LAN-exposed dashboards still work.

But almost nothing uses that helper. `ConfigPage.tsx`'s `CopyPathButton` calls the raw API directly:

```tsx
const handleCopy = useCallback(() => {
  navigator.clipboard.writeText(path).then(() => {
    setCopied(true);
    ...
```

With `navigator.clipboard` undefined, the property access throws before `.then()` is reached, so the `copied` state never flips — no clipboard write, no error, no visual feedback.

Call sites that bypass the helper and hit `navigator.clipboard.writeText` directly:

| File | Uses shared helper? |
|---|---|
| `pages/ConfigPage.tsx` | no |
| `pages/AuditPage.tsx` | no — defines its own local `copyToClipboard` that calls the raw API |
| `pages/AgentsPage.tsx` | no |
| `pages/UsersPage.tsx` | no |
| `components/SkillHubBar.tsx` | no |
| `components/TomlViewer.tsx` | no |
| `pages/ChatPage.tsx` | **yes** — the only importer of `lib/clipboard` |

**Repro:** serve the dashboard on a non-localhost address over plain HTTP (a LAN or VPN IP, e.g. `http://<host-ip>:4545`) and click any copy button outside the chat page. Ours is reached over a VPN IP, which is a normal way to run a self-hosted daemon.

**Fix:** route the six remaining call sites through `copyToClipboard` from `lib/clipboard`. `AuditPage`'s local function can be deleted in favour of the import. Worth considering a lint rule banning bare `navigator.clipboard` in this tree, since the helper's own comment shows this was already fixed once and regressed elsewhere.

## 2. `GET /api/sessions/{id}` omits metrics that `GET /api/sessions` provides

The list endpoint computes `cost_usd`, `total_tokens` and `duration_ms` via a SQL join; the detail endpoint hand-builds a fixed object and includes none of them.

Verified live on the same session id — list row carries all three, detail returns only `active`, `agent_id`, `context_window_tokens`, `created_at`, `label`, `message_count`, `messages`, `model_override`, `session_id`.

Currently latent: no shipped view reads those fields off the detail endpoint, and the frontend types mark them optional there. Same root cause as #6596 (hand-built partial JSON diverging from the fuller query), so worth aligning before something depends on it.

## 3. `/api/schedules` omits `peer_id`, which `/api/cron/jobs` exposes for the same underlying job

The two endpoints are intentional alternate views over the same `CronJob` store, but the schedules view has fallen behind. Verified live on this host's 5 cron jobs:

- `/api/cron/jobs` → `action, agent_id, created_at, delivery, delivery_targets, enabled, id, last_run, name, next_run, peer_id, schedule`
- `/api/schedules` → `agent_id, created_at, cron, delivery_targets, enabled, id, last_run, message, name, next_run, tz, workflow_id`

`peer_id` (cross-session memory routing) is visible on one page and not the other for the identical job. The audit that surfaced this also flagged `session_mode`; I could not confirm that one — it is absent from both responses on our data, likely skipped when unset, so I am only claiming `peer_id`.

## 4. `approval.trusted_senders` has no dashboard surface at all

`trusted_senders` is an approval-bypass list: a sender on it skips prompts for non-high-risk tools. It appears nowhere in the dashboard — grep across both the built bundle and the TS source returns zero matches. By contrast `channel_rules` at least surfaces on the permission simulator.

So an operator cannot see, audit, or review who is on the approval-bypass list from any UI, only by reading `config.toml`. Related to #6605 (that one is about the field being absent from `GET /api/config`); this is the UI half — even with the API fixed, nothing renders it.

## Environment

- Deployed `v2026.7.21`; items 1–3 verified against current `main` (`cb7608d9`, post-#6602/#6603/#6604). Items 2 and 3 verified live against the running daemon.

Happy to test a patch, and happy to split any of these out if you'd rather track them separately.

houko:
Four papercuts from #6611, plus the write halves and error-status fixes that fall out of them.

## Clipboard copy on a plain-HTTP dashboard

`navigator.clipboard` is only defined in a secure context, so on a daemon reached at `http://<lan-ip>:4545` the property is `undefined` and a bare `navigator.clipboard.writeText(...)` throws before the promise exists — the button produced no clipboard content, no error, and no visual feedback.

- `lib/clipboard.ts` has fixed this since it was written (it falls back to `document.execCommand('copy')` through a detached textarea), but only the chat page imported it. Routed the config page, audit detail, agents page (two sites), users page, skill install command, and TOML viewer through the helper.
- The helper reports failure by resolving to `false` rather than throwing, so each call site branches on the result instead of on a `catch` that could never fire. Two sites showed a "Copied" toast unconditionally; the users page gated the close button of a one-time-visible rotated API key on a copy it never verified; `SkillInstallCommand` skipped the write entirely when the API was missing and then fired `onCopied` anyway.
- `ConfigPage`'s `CopyPathButton` now surfaces a `common.copy_failed` toast on a total failure. Returning early without a signal reproduced the reporter's exact symptom on that one call site and left it inconsistent with the other five.
- Added an eslint guard so the regression cannot recur silently. `no-restricted-properties` bans `navigator.clipboard`, and a `no-restricted-syntax` selector adds the `window.navigator.clipboard` / `globalThis.navigator.clipboard` spelling that `no-restricted-properties` structurally cannot see (its `object` option matches a bare identifier only). `src/lib/clipboard.ts` disables the former alone, since it reads the bare form.
- The rule comment states measured coverage rather than assumed coverage: caught are `navigator.clipboard`, `navigator?.clipboard?.`, `const { clipboard } = navigator`, and the `<obj>.navigator.clipboard` prefixed forms; not caught are an aliased receiver (`const n = navigator; n.clipboard`) and `const { clipboard } = window.navigator`, both of which need alias tracking these syntactic rules do not have. A previous revision of this comment claimed the destructured form "slips past", which is the opposite of what the config does.

## `/api/schedules` routing fields — read and write

`/api/schedules` and `/api/cron/jobs` are deliberate alternate views over one `CronJob` store: the cron view serializes the struct whole, the schedules view renders a flattened presentation. The flattened one had fallen behind on the three fields that decide how a fire behaves.

- Read: `cron_job_to_schedule_json` now emits `peer_id`, `session_mode`, and `delivery`. Both optionals are explicit nulls when unset, matching how the view already renders `tz` / `last_run` / `next_run` — `CronJob::session_mode` carries `skip_serializing_if = "Option::is_none"`, so an unset value is *absent* from the cron view and indistinguishable from a field the server does not know about.
- Write, `POST /api/schedules`: all three are now settable. `peer_id` was hardcoded to `None`, `delivery` was forced to `CronDelivery::None`, and `session_mode` was parsed through an `.ok()` that turned a misspelling into "use the agent's default" behind a `201`. Malformed values now return a 400 that names the field; an empty `peer_id` is rejected at the boundary because `peer_scoped_key` refuses it as ambiguous with global scope and would otherwise fail at fire time.
- Omitting the fields keeps the historical defaults (no peer, no override, fire-and-forget delivery). Deliberately *not* the `LastChannel` default that the kernel-side `cron_create` applies: an existing client that posts no `delivery` must not start delivering output it never asked for.
- Write, `PUT /api/schedules/{id}`: `delivery` is now patched. The kernel's `update_job` has always supported it; this route simply never put it in the `updates` map. Shape validation happens in the route so a malformed body is a 400 — the kernel reports a serde failure as `LibreFangError::Internal`, which this route's fall-through arm renders as "Schedule not found", pointing the caller at a missing id instead of at their payload.

## `GET /api/sessions/{id}` metric parity

The list endpoint derives `cost_usd` / `total_tokens` from a `usage_events` join, `duration_ms` from the message timestamps, and a `label` snippet from the first user message when the column is empty. The detail handler hand-built a fixed object carrying none of it, so the same session answered differently depending on which route you asked.

- Both views now share one helper per value (`derive_session_label`, `session_duration_ms`, `SessionStore::session_usage_totals`) rather than a second copy of the derivation — a copy is what let them diverge in the first place.
- Shapes match the list exactly: numeric zeros for an unmetered session rather than null, null duration below two stamped messages, explicit label still wins over the snippet.
- A failed usage aggregate returns 500 rather than a coerced zero, matching the session load immediately above it. Reporting no spend for a session that spent money is the same silent-wrong-value failure this issue is about.

## `approval.trusted_senders` on `GET /api/config`

A sender on that list skips the approval prompt for every tool the risk classifier does not rank high, and it appeared on no surface at all — auditing who holds the waiver required shell access to read `config.toml`.

- Added to the redacted config response, and rendered as a read-only card on the Approvals page above the pending queue (it explains the requests that never arrive). An empty list is presented as the reassuring state it is.
- It stays out of `WRITABLE_EXACT_PATHS` for the reason it needed exposing: adding yourself to an approval-bypass list over HTTP is precisely the escalation the approval gate exists to prevent. The integration test pins non-writability alongside readability.

## BEHAVIOUR CHANGE

- `POST /api/schedules` now returns **400 instead of 500** for a payload that `CronJob::validate_with_home` rejects. `add_job` funnels every such rejection through `LibreFangError::InvalidInput`, and the route previously sent all errors through `internal_scrub`. Making `delivery` settable makes the SSRF host checks reachable from this route, so a refused webhook host would otherwise have been reported as a server fault with the reason scrubbed away. Mirrors the split already present in `update_schedule` (#4732). Any client treating a 500 here as retryable will now see a non-retryable 400 — which is the correct classification.
- `POST /api/schedules` now returns **400 instead of 201** for a malformed `session_mode`. Previously a typo was silently coerced to "no override", so a caller asking for per-fire isolation got the shared persistent session and a success status.
- `POST /api/schedules` now **honours** `peer_id` and `delivery`, which it previously ignored. A client that has been sending either field (expecting it to be dropped) will now get a job that resolves peer-scoped memory and delivers its output.
- `PUT /api/schedules/{id}` now returns **400 instead of 200** when a request tries to *change* `peer_id` or `session_mode`. **This affects existing callers**: before this change the route accepted either field, returned 200, and silently ignored it, so any script that sends a `peer_id` differing from the stored one will now get a 400 where it previously got a success. That is the point — the old 200 asserted a change that never happened — but it is a compatibility break, not just a design note. Rationale for choosing refusal over silent-ignore is in the escalation section below.
  - Sending the currently-stored value is a 200 no-op, so a GET-mutate-PUT round trip is unaffected.
  - `session_mode` is compared by *effective* mode, not by raw `Option`: a stored `None` and a requested `"persistent"` are the same thing (`cron_fire_session_override` only diverges on `New`), so writing the default out longhand is not treated as a change. Without this, the common case — stored `None`, which is what `skip_serializing_if` on the field makes typical — would have falsely rejected a caller who spelled out `"persistent"`.
  - Checked the blast radius in-repo: both dashboard call sites (`SchedulerPage.tsx:295`, `HandsPage.tsx:1078`) send a narrow `{ enabled }` patch through `useUpdateSchedule`, not a spread of the whole `ScheduleItem`, so the dashboard itself is unaffected.
- `GET /api/sessions/{id}` now returns **500 instead of 200** when the `usage_events` aggregate query fails. Previously that path could not fail because it did not exist.

## Escalation: `peer_id` / `session_mode` on PUT need a kernel change

The write half of #6611 is complete inside `librefang-api` except for patching these two fields on update, which cannot be done from this crate.

`CronScheduler::update_job` (`crates/librefang-kernel/src/cron.rs:383`) has no branch for either field, and every branch it does have tests `!updates["<field>"].is_null()` — an omitted-or-null-means-untouched convention that structurally cannot express "clear this optional field". Supporting the two needs presence-based semantics there: `updates.get("peer_id")` present and `null` ⇒ clear, present and a string ⇒ set, absent ⇒ leave alone. Same for `session_mode`.

Read-modify-write from the API layer is not a workaround. `remove_job` + `add_job` would discard the `JobMeta` that lives beside the `CronJob` (`one_shot`, `last_status`, `consecutive_errors`, `auto_disabled`, `consecutive_fallbacks`) and is not atomic against a concurrent `due_jobs` tick.

Trade-off for a reviewer to weigh, which is why this is surfaced rather than decided here: `PUT /api/cron/jobs/{id}` forwards its raw request body to the same `update_job`, so switching that function to presence-based semantics changes both routes at once. Both would gain the ability together, and both would gain the "explicit null clears the field" contract together.

Meanwhile, rather than accept a field that has no effect, `update_schedule` refuses a *change* with a 400 that says a recreate is needed. Silently returning 200 for a patch that never applied is the same class of defect as the read gap this issue reported, in the opposite direction.

## Verification

`cargo check -p librefang-api` — clean. `cargo clippy -p librefang-api --all-targets -- -D warnings` — clean.

Integration binaries, run individually:

- `cargo test -p librefang-api --test workflows_routes_integration` — 66 passed
- `cargo test -p librefang-api --test system_tools_sessions_test` — 23 passed
- `cargo test -p librefang-api --test config_routes_integration` — 43 passed
- `cargo test -p librefang-api --test openapi_path_coverage_test --test dead_route_audit_test` — 3 passed (no route shape change, so no spec regeneration)
- `cargo test -p librefang-memory --lib session` — 56 passed

Dashboard: `npm run typecheck` clean, `npm run lint` 0 errors (85 pre-existing warnings, unchanged), `npm run test` 938 passed across 89 files.

`npm run test:i18n-parity` reports `zh` and `ko` at full parity and `uk` failing with 8 extra keys. That failure is pre-existing and untouched: the same 8 keys are extra on `origin/main`, and all four new `approvals.trustedSenders*` keys were added to all four locales (verified by flattening each locale against `en.json` before and after this branch's changes — the extra-key set is byte-identical). The 8 are Ukrainian `_few` / `_many` plural forms that English has no counterpart for, so the parity script counts them as drift; fixing that is a change to the script's plural handling in a different domain and is not bundled here.

New tests:

- `schedules_view_carries_the_same_routing_fields_as_cron_jobs`
- `schedule_detail_carries_the_same_routing_fields_as_the_list_row`
- `schedules_view_renders_unset_routing_fields_as_null`
- `schedule_create_persists_peer_id_session_mode_and_delivery`
- `schedule_create_without_routing_fields_keeps_historical_defaults`
- `schedule_create_rejects_malformed_routing_fields`
- `schedule_create_rejects_ssrf_delivery_webhook_with_400`
- `schedule_update_patches_delivery`
- `schedule_update_rejects_malformed_delivery_with_400`
- `schedule_update_without_delivery_preserves_it`
- `schedule_update_refuses_to_change_peer_id_or_session_mode`
- `schedule_update_treats_explicit_persistent_as_no_change_against_unset`
- `schedule_update_unknown_id_with_peer_id_returns_404_not_400`
- `schedule_update_accepts_unchanged_peer_id_and_session_mode`
- `session_detail_reports_same_metrics_as_the_list_row`
- `session_detail_metrics_are_present_on_an_unmetered_session`
- `session_detail_label_falls_back_to_the_snippet_then_prefers_an_explicit_value`
- `get_config_exposes_approval_trusted_senders`

### Mutation control

Two earlier tests were green but proved nothing — reverting the fix left them passing. Both were rewritten and each rewrite was re-verified against a hand-reverted source hunk, with the fix restored afterwards.

`schedule_detail_carries_the_same_routing_fields_as_the_list_row` previously asserted only `detail == list row`. Both sides render through `cron_job_to_schedule_json`, so removing the fields removed them from both sides equally and the equality still held. It now pins the values directly, and with the three keys removed from the helper it fails:

```
thread 'schedule_detail_carries_the_same_routing_fields_as_the_list_row' panicked at
crates/librefang-api/tests/workflows_routes_integration.rs:1505:5:
  left: None
 right: Some("peer-42")
```

`session_detail_label_prefers_the_explicit_value_over_the_derived_snippet` seeded an explicit label and asserted the response carried it — which the pre-fix handler already did, since it emitted `session.label` verbatim. The fallback was never exercised. Rewritten as `session_detail_label_falls_back_to_the_snippet_then_prefers_an_explicit_value`, testing the unlabeled case first; with the fallback reverted to `"label": session.label` it fails:

```
thread 'session_detail_label_falls_back_to_the_snippet_then_prefers_an_explicit_value' panicked at
crates/librefang-api/tests/system_tools_sessions_test.rs:849:5:
assertion `left == right` failed: an unlabeled session must report the derived snippet, not null
  left: None
 right: Some("rotate the staging credentials")
```

The write-half tests were controlled the same way. Re-hardcoding `peer_id: None` / `delivery: CronDelivery::None` in `create_schedule` and dropping the `delivery` insert from `update_schedule` fails three of them:

```
schedule_create_persists_peer_id_session_mode_and_delivery ... FAILED   left: None      right: Some("peer-42")
schedule_create_rejects_ssrf_delivery_webhook_with_400     ... FAILED   left: 201       right: 400
schedule_update_patches_delivery                           ... FAILED   left: Some("none") right: Some("channel")
```

Removing the PUT refusal blocks fails the fourth:

```
schedule_update_refuses_to_change_peer_id_or_session_mode ... FAILED
assertion `left == right` failed: changing `peer_id` must not report success:
Object {"status": String("updated"), "schedule_id": String("94d1ec82-...")}
  left: 200
 right: 400
```

Reviewing that refusal code found a bug in it before it shipped: the first draft compared `current.as_ref().map(...) != Some(requested)`, which for an unknown schedule id made `current` `None` and returned a field-level 400 instead of the route's 404 — sending the caller to inspect their payload rather than the id. The comparison now runs only when a stored job exists, and `schedule_update_unknown_id_with_peer_id_returns_404_not_400` pins it. Reinstating the buggy comparison fails it:

```
schedule_update_unknown_id_with_peer_id_returns_404_not_400 ... FAILED
  left: 400
 right: 404
```

## Collision with open PR #6637

#6637 is still **open** (not merged as of `origin/main` at `b0b83149`), so nothing was rebased onto it. It adds seven approval fields to the same `set!("approval", …)` block in `routes/config/manage.rs`, including `trusted_senders`, and inserts a test at the same place in `config_routes_integration.rs`.

Resolution when one of the two merges first: keep a single `trusted_senders` line in the `set!` block (the two are identical), and keep **both** tests — `get_config_exposes_approval_trusted_senders` here additionally asserts the field stays non-writable via a `POST /api/config/set` that must return 403, which #6637's test does not cover.

## Base

This branch is based on `3680db29`; `origin/main` has since advanced to `b0b83149`. Overlapping files are `dashboard/src/api.ts`, the four locale files, `AgentsPage.tsx`, `ApprovalsPage.tsx`, and `ConfigPage.tsx` — all additive in both directions except the `ConfigPage.tsx` `eslint-disable-next-line no-console` directive, which #6646 removes upstream and this branch leaves alone.

Closes #6611


houko (commented):
Automated review pass (diffed against the merge-base `3680db29`, not against a post-conflict state).

**Merge conflicts (informational, not addressed here):** this PR self-reports `has-conflicts` against open PR #6637 — both touch the same `set!("approval", …)` block in `routes/config/manage.rs` and the same spot in `config_routes_integration.rs`. Per the PR body, resolution is to keep one `trusted_senders` line and keep both tests when one of the two merges first. Leaving that resolution to the author/maintainer.

**What I checked:**
- Clipboard fix: `lib/clipboard.ts`'s `execCommand` fallback is unchanged and correct; all six call sites (ConfigPage, AuditPage, AgentsPage ×2, UsersPage, SkillHubBar, TomlViewer) now route through `copyToClipboard` and branch on its boolean result rather than a `catch` that could never fire. New eslint `no-restricted-properties` / `no-restricted-syntax` rules parse correctly (verified with `npx eslint .` — 0 errors, 85 pre-existing warnings, matching the PR description exactly).
- Dashboard data-layer rule: no inline `fetch()`/`api.*` calls were added; `ApprovalsPage`'s new `TrustedSendersCard` goes through the existing `useFullConfig()` query hook. `npx tsc --noEmit` is clean.
- Backend gaps (`/api/schedules` routing fields, `GET /api/sessions/{id}` metric parity, `approval.trusted_senders` on `GET /api/config`) match the PR description; `cargo fmt --check` is clean on the touched crates. Spot-checked the `update_schedule` peer_id/session_mode refusal logic and the 404-vs-400 ordering bug fix described in the PR body — both look correct, and the named test files/functions are all present in the diff.

**One mechanical fix pushed** (commit `b9587e0`, `docs(changelog): reference this PR, not the issue it closes, in fragments`): all four `changelog.d/fixed/6611-*.md` fragments ended their bullet with `(#6611)` — the issue number — instead of `(#6668)` — this PR's own number. Per `changelog.d/README.md`, the release tooling only suppresses the auto-generated `- <PR title> (#N) (@author)` line for a PR whose own number appears in a curated bullet's trailing group, so as written all four entries would have doubled up with a generic line in the release notes. Fixed to `(#6668)`; re-ran `python3 scripts/check-changelog-attribution.py --all-unreleased` afterward — clean.

No other findings; the rest of the diff (session/substrate helpers, cron routing-field plumbing, integration + dashboard tests) matches the PR's own description closely and looks solid on read-through.

---

crates/librefang-api/src/routes/config/manage.rs
houko, line 1: The PR description's "Collision with open PR #6637" section and its test-coverage claims look stale against current `origin/main`.

`#6637` is already merged (`469634e`, "report the whole approval policy on config read", is an ancestor of `origin/main` at `820d30f`) — it isn't open. Its own test `get_config_exposes_non_writable_approval_fields` (`config_routes_integration.rs:339`) already asserts both that `approval.trusted_senders` round-trips on `GET /api/config` **and** that a `POST /api/config/set` against it returns 403. That's exactly the coverage this PR's description says only its own new test (`get_config_exposes_approval_trusted_senders`) provides — but that test doesn't exist anywhere in this diff, and the "43 passed" `config_routes_integration` count in the Verification section doesn't add a new case for it either.

Functionally this looks fine (the dashboard card only reads an already-covered field, `manage.rs` here only gained explanatory comments — no new field), but the PR body is describing a merge-conflict resolution and a test that never landed. Worth updating the description before merge so reviewers aren't chasing a test that isn't there, and dropping the now-resolved collision note.

---

houko (commented):
Reviewed against CLAUDE.md's dashboard data-layer/query-key/mutation rules, changelog fragment format, and the prose-wrapping rule. Pushed mechanical rewrap fixes for a few comment blocks the PR's own two rewrap passes missed; left one inline comment about a stale part of the PR description.

---

houko:
Automated review pass on this PR's own diff (base `820d30f`..head `01faa54`, plus the two follow-up commits pushed alongside this comment).

**Fixed and pushed** (mechanical, verified):
- `crates/librefang-api/dashboard/src/api.ts`: `ScheduleItem` (the read type) picked up `peer_id` / `session_mode` / `delivery`, but the `createSchedule` / `updateSchedule` request payload types were not widened to match, so the dashboard's typed HTTP layer had no way to use the write capability this PR adds to `POST`/`PUT /api/schedules`. Added the three fields to both payload types, mirroring the shape already established for `/api/cron/jobs`'s `CreateCronJobPayload`/`UpdateCronJobPayload`. Verified with `pnpm typecheck`, `pnpm lint`, `pnpm test --run` (all green, including the 5 new trusted-senders tests).
- `crates/librefang-api/tests/workflows_routes_integration.rs`: added `schedule_update_rejects_ssrf_delivery_webhook_with_400`, mirroring the existing `schedule_create_rejects_ssrf_delivery_webhook_with_400`. The existing malformed-shape update test only exercises the 400 path that's rejected at the API boundary before `update_job` is ever called; this one pins that the kernel's SSRF check *inside* `update_job` (already unit-tested at `update_job_rejects_ssrf_webhook_in_delivery`) also maps to 400 at the route layer on the update side, not just create. Verified with `rustfmt --check`; a full `cargo check -p librefang-api` did not finish locally within a reasonable window due to a cold, contended build environment, so please let CI have the final word on compilation.

**Worth a look, not changed** (design judgment, not mechanical):
- The Scheduler dashboard page (`crates/librefang-api/dashboard/src/pages/SchedulerPage.tsx`) still has no form inputs for `peer_id`, `session_mode`, or the primary `delivery` on schedule create/edit — only `delivery_targets` (fan-out) is wired into the UI. After this PR (plus the api.ts fix above), the three fields are readable, settable via typed API calls, and round-trippable, but an operator still cannot set any of them without hand-crafting an HTTP request; the dashboard form is API-only-reachable for this feature. Whether to add UI for this now or as a separate follow-up seems like a product call (representing `CronDeliverySpec`'s four variants and a session-mode toggle in the create form isn't a one-line change), so flagging rather than deciding it unilaterally.

No comment on the `CronScheduler::update_job` presence-based-semantics escalation or the #6637 field-overlap note in the PR body — both read as correctly-scoped deferrals already.

---

crates/librefang-api/src/routes/config/manage.rs
```diff
@@ -398,6 +398,8 @@ fn redacted_config_json(
     // `cache_approvals_per_session` was the visible one: it defaults to `true`, so an operator who left it alone still saw the box unchecked and had no way to tell whether per-session caching was on (#6636 observation (e)).
     // None of the seven is writable, which is why `every_writable_config_leaf_is_readable` never saw them; `approval_policy_fields_are_all_readable` is the guard for this direction.
     // Nothing added here is secret-bearing: `trusted_senders`, `channel_rules`, and `routing` hold operator-chosen user ids, channel names, tool globs, and notification recipients — the same class as the `external_auth` fields exposed in #6605, and credentials never live in `ApprovalPolicy`.
+    // Of those seven, `trusted_senders` is the one with teeth: it is an approval-*bypass* list, so a sender on it skips the prompt for every tool `classify_risk` ranks below `High` (see `ApprovalManager::requires_approval_with_context_for`), which is why the Approvals page renders it as its own card rather than leaving it to the generic config form (#6611).
```
houko, line 401: The code here looks correct (single `trusted_senders` line, no duplication), but the PR description's "Collision with open PR #6637" section reads as if #6637 is still open and the merge is yet to happen. It has already merged into `main` as `469634ece` ("fix(hands): ... report the whole approval policy on config read (#6637)"), which already added `trusted_senders` to this `set!` block and added a test (`get_config_exposes_non_writable_approval_fields`) that asserts both the read and the 403-on-write behavior for it. This branch appears to have already been rebased on top of that (see the "docs(api): stop the approval comment contradicting the block above it" commit), so the actual collision is resolved in code — only the description's "Collision" section and the Verification section's claimed new test `get_config_exposes_approval_trusted_senders` are now stale/inaccurate relative to current `main`. Worth updating the PR description before merge so reviewers don't go looking for a test that isn't part of this diff, or a conflict that no longer exists.

---

PR #6670, opened 2026-07-30
fix(test): regenerate the KernelConfig schema golden fixture main had drifted past

houko:
`main` is red on `kernel_config_schema_matches_golden_fixture`. This regenerates the fixture; no schema is touched.

## What happened

#6664 re-enabled the guard against a fixture generated from its own base. Three PRs then merged schema-visible changes without regenerating it, because on each branch the guard was still matching either the pre-#6664 state or the pre-merge schema:

- #6661 — `[registry] auto_sync` (new field, plus its entry in the `RegistryConfig` default block).
- #6666 — `additionalProperties: false` on `McpTransportEntry`, `HttpCompatHeader` and `HttpCompatBody` from the `deny_unknown_fields` addition, and the two `"default": null` entries schemars stops emitting under it. The #6666 review flagged exactly this as a sequencing risk and left the call to a maintainer.
- #6667 — the rewritten `api_key` doc comment and the new `api_key_hash` field.

Every one of those PRs was green on its own branch. `main` went red when the last one landed.

## Verification

- `cargo test -p librefang-api --test config_schema_golden` — passes on this branch.
- The same command on `origin/main` fails: `actual: 7963 lines / 334503 bytes` against `expected: 7948 lines / 329443 bytes`.
- The fixture is the `regenerate_golden` entry point's output, so the diff is exactly the three merged changes above and nothing else.

## Worth a maintainer's decision (not changed here)

The guard cannot catch this class on its own: a PR whose base predates a schema change passes its own CI and breaks `main` on merge. Requiring branches to be up to date with `main` before merge would close it at the source, which is a repo setting rather than a code change — say the word and I'll open it as an issue instead of leaving it in a PR body.

### DIFF
--- a/changelog.d/fixed/6670-golden-fixture-catchup.md
+++ b/changelog.d/fixed/6670-golden-fixture-catchup.md
@@ -1,0 +1,3 @@
+Regenerate the `KernelConfig` schema golden fixture, which `main` had drifted past.
+#6664 re-enabled `kernel_config_schema_matches_golden_fixture` against a fixture generated from its own base, and #6661, #6666 and #6667 each merged a schema-visible change afterwards — `[registry] auto_sync`, `additionalProperties: false` on the MCP transport structs, and the `api_key` / `api_key_hash` documentation — so the guard went red on `main` the moment the last of them landed, while every one of those PRs was green on its own branch.
+The fixture is generated from the code, so the only change here is the fixture catching up; no schema was touched (#6670) (@houko)
--- a/crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json
+++ b/crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json
@@ -33,7 +33,12 @@
       }
     },
     "api_key": {
-      "description": "API authentication key. When set, all API endpoints (except /api/health) require a `Authorization: Bearer <key>` header. If empty, the API is unauthenticated (local development only).",
+      "description": "API authentication key. When set, all API endpoints (except /api/health) require a `Authorization: Bearer <key>` header. If empty (and no `api_key_hash` is set), the API is unauthenticated (local development only).\n\nThe value is resolved by the API layer in three steps, the same order the dashboard credentials use: 1. The `LIBREFANG_API_KEY` environment variable, when set to a non-empty value, wins over whatever this field holds. 2. A `vault:KEY_NAME` prefix reads `KEY_NAME` out of the encrypted vault (`$LIBREFANG_HOME/vault.enc`). Example: `api_key = \"vault:master_api_key\"`, then run `librefang vault set master_api_key`. 3. Anything else is the literal bearer token.\n\nSteps 1 and 2 keep the working secret out of `config.toml` entirely, which matters because the daemon rewrites that file and cannot read it from a read-only Kubernetes Secret mount. Resolution happens per auth snapshot rather than once at boot, so an env-sourced or vault-sourced key survives `POST /api/config/reload` — a reload re-reads `config.toml` from disk and would otherwise clobber the boot-time override.\n\nPrefer [`KernelConfig::api_key_hash`] when the daemon only needs to *verify* the key rather than transmit it — run `librefang hash-api-key` to produce the value. The two can be combined, and a plaintext value here is verified with a constant-time comparison.",
+      "default": "",
+      "type": "string"
+    },
+    "api_key_hash": {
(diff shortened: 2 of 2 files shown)


crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json
houko, line 1: Verified the regen is correct and complete: after `rustup update stable` (1.94.1 → 1.97.1, needed for `sysinfo` in this sandbox) and installing `libgtk-3-dev`/`libdbus-1-dev`, `cargo check -p librefang-types -p librefang-api --lib` compiles clean and `cargo test -p librefang-api --test config_schema_golden` passes `kernel_config_schema_matches_golden_fixture` against this exact fixture — no further drift. Traced each hunk back to source (`RegistryConfig::auto_sync` default `true` in `types.rs`, `deny_unknown_fields` on `McpTransportEntry`/`HttpCompatHeaderConfig`/`HttpCompatToolConfig`, the `api_key`/`api_key_hash` doc comments) and all match current code, so this is a genuine mechanical catch-up, not a hand-edit that could re-drift.

One thing worth a look outside this PR's diff: neither #6661 nor #6666 touched this fixture file (only #6664 and this PR do, per `git log` on the path), even though both added schema-visible fields/attributes. Per the changelog fragment's own account, they were green on their own (pre-#6664 or stale) base and only went red once stacked on `main` — so the golden guard didn't actually catch the drift at merge time for either PR, it just detected the accumulated result afterward. That's consistent with each PR being validated against a base that predated `main`'s eventual state rather than tip-of-main at merge time (the "re-created worktree → fetch and diff against origin before editing" drift this repo's CLAUDE.md already calls out, just showing up here as a CI-order problem instead of a worktree one). Not asking for anything in this PR — the fixture fix itself is correct — but might be worth a merge-queue / "rebase and re-run schema golden test against tip-of-main before merge" follow-up so future schema PRs get caught individually instead of needing a catch-up commit like this one.

---

houko (commented):
Verified the golden fixture regen against current code (compiles + `config_schema_golden` test passes exactly). Left one non-blocking observation on the fixture file about the upstream process gap that caused the drift in the first place.

---

changelog.d/fixed/6670-golden-fixture-catchup.md
```diff
@@ -0,0 +1,3 @@
+Regenerate the `KernelConfig` schema golden fixture, which `main` had drifted past.
+#6664 re-enabled `kernel_config_schema_matches_golden_fixture` against a fixture generated from its own base, and #6661, #6666 and #6667 each merged a schema-visible change afterwards — `[registry] auto_sync`, `additionalProperties: false` on the MCP transport structs, and the `api_key` / `api_key_hash` documentation — so the guard went red on `main` the moment the last of them landed, while every one of those PRs was green on its own branch.
```
houko, line 2: Since this fragment ends up verbatim in the release notes, worth double-checking the merge-order claim here. Actual commit timestamps on `main` are:

- `fd74faa` (#6667) — 13:03:46
- `ca3e94e` (#6661) — 13:04:25
- `705862d` (#6664) — 13:05:53
- `820d30f` (#6666) — 13:06:31

So #6667 and #6661 actually landed on `main` *before* #6664, not "afterwards" as this line states — only #6666 genuinely merged after #6664. The PR body's own "What happened" section gets the mechanism right (each PR's fixture matched whatever base/pre-merge schema its own branch had, independent of literal merge order onto `main`), but this changelog line's "afterwards" reads as a merge-order claim that the timestamps don't support. Worth rewording to match the branch-basis explanation rather than implying merge chronology, since a reader will otherwise reasonably interpret it as merge order.

---

changelog.d/fixed/6670-golden-fixture-catchup.md
houko, line 1: Nit, not blocking: the changelog fragment itself follows the CLAUDE.md prose-wrapping rule (one sentence per line, terminal `(#6670) (@houko)` attribution) — no issue there.

But the commit message body (8e08453) and the PR description both pack multiple sentences onto a single physical line in a few paragraphs, e.g. `"#6664 re-enabled the golden guard against a fixture generated from its own base. #6661, #6666 and #6667 then each merged a schema-visible change — … — none of which regenerated the fixture, because …"` — two sentences run together on one line. CLAUDE.md's "Prose wrapping" section explicitly extends the one-sentence-per-line rule to "Commit message bodies" and to "PR titles, PR bodies". Worth a squash/reword at merge time if a maintainer wants strict compliance; not re-flowing here since fixing it cleanly would mean rewriting the only commit on this branch (force-push territory), which is out of scope for an automated pass.

---

houko (commented):
Automated daily review pass. Checked against CLAUDE.md: changelog fragment format/section/attribution, no direct `## [Unreleased]` edit, conventional commit type, worktree/git hygiene, schema/registry/config conventions, no AI attribution. `python3 scripts/check-changelog-attribution.py --staged` and `--all-unreleased` both pass. Ran `cargo test -p librefang-api --test config_schema_golden` to confirm the regenerated fixture actually matches the derived schema. One minor prose-wrapping nit left inline (not blocking, not mechanically fixable without rewriting the branch's only commit). No functional issues found — this is a pure fixture regeneration with no schema drift beyond what #6661/#6666/#6667 already introduced.

---

changelog.d/fixed/6670-golden-fixture-catchup.md
houko, line 1: This overlaps with #6676 ("fix(config): regenerate the schema golden fixture #6664 landed one merge behind"), also open, also by @houko, also targeting `crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json`.
Diffing the two PRs' fixture patches, the resulting JSON is identical between them (same `auto_sync`, `api_key_hash`, and `additionalProperties: false` hunks) — both regenerate the fixture to the same end state, just from slightly different base commits (this PR bases on 820d30f2, #6676 bases on 198bdf25, one commit ahead but CI-only in between) and with different changelog narratives / fragment filenames (`6670-golden-fixture-catchup.md` vs `6676-regen-golden-after-merge-order.md`).
Whichever merges first will make the other a same-content no-op that still needs a maintainer decision: close the second as superseded, or let it merge as an empty diff and drop its now-redundant changelog fragment.
Flagging rather than resolving — this is a call for whoever merges, not something an automated pass should decide by closing either PR.

---

houko (commented):
Automated daily review pass. One new finding not already covered by the existing review threads: this PR overlaps with #6676, which regenerates the same golden fixture to the same resulting content.

---

houko:
Superseded by #6676, which landed the same regeneration in 79036957.

That is my duplicate, not a disagreement: this PR was open first and reached the same diagnosis independently — #6661, #6666 and #6667 each merged a schema-visible change while the guard on their own branch was still matching a pre-merge fixture, so `main` went red when the last one landed. I opened #6676 without checking for an existing PR on the same failure and merged it, which made this one redundant rather than wrong.

`main` is green on `kernel_config_schema_matches_golden_fixture` as of 79036957.

PR #6671, opened 2026-07-30
fix(config): make flipping `[registry] auto_sync` survive the reload it claimed to survive, and cover the boot gate

houko:
Follow-up to #6661, which merged before these two review findings on it were addressed. Refs #6636 (observation (a)).

## The claim did not hold

#6661's doc says that flipping `[registry] auto_sync` off "stops the next automatic refresh without a restart". It did not.

`registry` was classified restart-required as a whole section, so a reload that changed only that section produced neither a hot action nor a live-read change — and `should_store_config` swaps the config only when one of those is non-empty (`config_reload.rs`). The reload therefore reported restart-required and then discarded the new value, leaving the 24 h catalog task reading the old `auto_sync` out of the previous snapshot until the daemon actually restarted.

That is the trap version of the bug #6661 set out to fix: an operator whose local hand edits are being clobbered edits `config.toml`, calls `POST /api/config/reload`, is told a restart is needed, and the next tick clobbers the checkout anyway.

## The fix

`auto_sync` is now classified as a live-read change on its own; `cache_ttl_secs` / `registry_mirror` / `registry_host` stay restart-required. The split follows how each is consumed: the catalog task calls `config_snapshot()` at the top of every tick and passes `auto_sync` straight into `sync_catalog_to` (`server.rs`), while the other three are read when the checkout is set up.

`docs/operations/config-reload.md` records the section as `R/N` with the split spelled out, in the same shape the `external_auth` row already uses for its `H/N` split. That table is drift-guarded by `doc_reload_table_matches_classified_reload_fields`, which passes.

## The boot gate had no test

The second finding: nothing exercised `boot_with_config`'s `if config.registry.auto_sync` gate, which is where the restart-survival half of the claim is actually implemented. The config round-trip and the catalog-task path were covered; the gate was not.

`boot_with_auto_sync_disabled_leaves_the_registry_checkout_alone` seeds a hand into the checkout, leaves `.sync_marker` absent so a refresh is due, boots with `auto_sync = false`, and asserts the checkout is untouched and the hand still loads.

It also asserts that a probe provider file was **not** fanned out into the home dir. That is the assertion that makes the test worth having: `sync_registry` always ends in `fanout_registry_content`, which copies `registry/providers/*.toml` into `<home>/providers/` and needs no network — so on an offline runner, where the fetch fails and leaves the checkout alone by accident, the checkout-content assertions would pass vacuously while the probe still catches the regression.

Verified by mutation: replacing the gate with `if true` fails the test both with connectivity (the sync wipes the checkout) and without it (the fan-out copies the probe).

## Verification

- `cargo test -p librefang-kernel --lib` — 1363 passed, run on this branch. New: `registry_auto_sync_is_read_live_and_reaches_the_config_store`, `registry_fields_other_than_auto_sync_still_require_restart`, `boot_with_auto_sync_disabled_leaves_the_registry_checkout_alone`.


crates/librefang-kernel/src/config_reload.rs
```diff
@@ -705,7 +705,14 @@ pub fn build_reload_plan_with_caps(
         );
         restart_if_changed(field_changed(&old.heartbeat, &new.heartbeat), "heartbeat");
         restart_if_changed(field_changed(&old.plugins, &new.plugins), "plugins");
-        restart_if_changed(field_changed(&old.registry, &new.registry), "registry");
+        // `registry` minus `auto_sync`: the mirror / host / TTL are read once when the checkout is set up, but `auto_sync` is re-read from the config snapshot on every tick of the 24 h catalog task, so it belongs in the NOOP block below.
```
houko, line 708: Not about the content of this line — it's the last commit that touched it (`3691761`, "fix(kernel): correct a doc-comment misattachment and reflow hard-wrapped prose") that's worth a look before merge.

That commit's git author identity is `Claude <noreply@anthropic.com>`, which will show up on the commit list, in `git log`, and in blame for this file. CLAUDE.md's "No Claude / Anthropic / AI attribution" rule is aimed at exactly this kind of provenance signal in the repo history, even though the wording calls out messages/bodies/comments rather than the author field specifically — the effect (an AI identity permanently attached to a commit in `main`'s history once this merges) is the same thing the rule exists to prevent, and neither `commit-msg` nor the `guard-bash-safety.sh` Bash-hook pattern-match would have caught it, since both scan message text, not author identity.

The other two commits on the branch are attributed to human authors, so this looks like an isolated slip rather than a pattern. Fixing it means rewriting this commit's author (`git commit --amend --author` under an interactive rebase) and force-pushing the branch — force-push is fine on your own pre-review branch per CLAUDE.md, but it's a history rewrite so I'm flagging it for you to do rather than doing it myself.

---

houko (commented):
Reviewed against CLAUDE.md (changelog fragment format, config hot-reload classification docs, prose-wrapping rule, commit/PR hygiene, no-AI-attribution). The reload-classification fix itself checks out: `RegistryConfig` has exactly the four fields the diff accounts for, the doc table's `R/N` split mirrors the existing `external_auth` `H/N` pattern, and the new tests match the classification. One item left as a line comment for a human call: the commit-author identity on `3691761`.

---

crates/librefang-kernel/src/kernel/tests.rs
houko, line 1: Commit 3691761 ("fix(kernel): correct a doc-comment misattachment and reflow hard-wrapped prose") is authored and committed by the `claude` GitHub account (`Claude <noreply@anthropic.com>`, confirmed via the commits API — GitHub resolves it to https://github.com/claude, id 81847).

CLAUDE.md is explicit that commit author identity resolving to Claude/Anthropic is rejected, independent of message content: "a commit whose author identity (`git var GIT_AUTHOR_IDENT` …) resolves to Claude / Anthropic even when the message itself is clean" is what the `commit-msg` hook exists to catch, and "No Claude / Anthropic / AI attribution in commit messages, PR bodies, or issue comments" is called out again under Commit & PR hygiene. This commit is already on `origin/fix/registry-auto-sync-reload`, so it must have bypassed that hook (not installed for whatever session produced it, or pushed through a path that skips it).

The content of the commit (relocating the new test's doc comment, reflowing prose to one sentence per line) looks correct and doesn't need to be redone — only the author/committer identity needs to change. That's an amend + rebase of the tip commit, which requires a force-push; since only the branch owner should rewrite this PR's history, recommend @houko run something like:

```
git commit --amend --author="<real name> <real email>" 
# or, if this isn't the very tip by the time you see this:
git rebase -i <base> # reword the author on 3691761
git push --force-with-lease origin fix/registry-auto-sync-reload
```

Flagging rather than fixing directly since rewriting another contributor's branch history is out of scope for this review pass.

---

crates/librefang-kernel/src/config_reload.rs
houko, line 1: Minor, separate from the comment on tests.rs: commit 398c09f's body paragraphs (e.g. the one starting "The doc claimed that flipping `auto_sync` off …") each pack several sentences onto a single line. CLAUDE.md's prose-wrapping rule applies to commit message bodies explicitly ("One sentence = one line, regardless of length" / "Commit message bodies" is listed under "This rule applies universally"). The in-diff comments and the changelog fragment in this PR do follow one-sentence-per-line correctly — it's only this commit message body that doesn't. Not blocking, and fixing it would mean rewriting the same commit's history as the authorship issue above, so bundling the note here rather than asking for a separate rebase pass.

---

houko (commented):
Reviewed against CLAUDE.md's config hot-reload classification note and general engineering rules. The config_reload.rs classification split, the new kernel tests, and the docs/operations/config-reload.md row update all check out functionally (verified `RegistryConfig.auto_sync`, the `boot_with_config` gate, and `sync_registry`/`fanout_registry_content` against the actual source referenced in the test doc comments; the doc-table drift guard test still passes since `registry` stays a single classified field name). The changelog fragment in `changelog.d/fixed/` is correctly formatted and passes `scripts/check-changelog-attribution.py --all-unreleased`. Two findings left as comments below — both require a history rewrite (author-identity amend / rebase) that only the branch owner should do, so left as review comments rather than pushed fixes.

---

Issue #5664, opened 2026-05-23
[High] Git hooks & CI hygiene — rustfmt paths, openapi sha256, missing pre-push, unconsumed `.secrets.baseline`


houko:
**Severity:** High · **Domain:** CI / hooks

Merges 4 git-hook / CI hygiene findings under one tracking item.

## Sub-findings rollup

| # | Description | Location |
|---|---|---|
| 1 | `pre-commit` uses unquoted `$STAGED_RS` + `grep`; files with spaces or glob characters get word-split | `scripts/hooks/pre-commit:25-30` |
| 2 | The openapi-sha check in `pre-commit` only calls `shasum`; no fallback on `sha256sum`-only environments (typical Linux dev boxes) | `scripts/hooks/pre-commit` (openapi-sha section) |
| 3 | `CLAUDE.md` and `scripts/hooks/` describe a `pre-push` hook, but no such file exists in the repo — docs and reality disagree | `CLAUDE.md`, `scripts/hooks/` |
| 4 | `.secrets.baseline` exists but CI never calls `detect-secrets`; only `pre-commit` soft-warns | `.github/workflows/`, `.secrets.baseline` |

## Why merged

All four are git-hook / CI hygiene items concentrated in `scripts/hooks/` + `.github/workflows/`; a single sweep is more efficient than 4 separate PRs.

## Combined fix plan

1. **NUL-delimited xargs** for the rustfmt staged-files loop:
   ```bash
   git diff --cached --name-only --diff-filter=ACM -z \
     | grep -z '\.rs$' \
     | xargs -0 -r rustfmt --check --edition 2021
   ```
2. **shasum / sha256sum interop**:
   ```bash
   sha256() { command -v sha256sum >/dev/null && sha256sum "$@" || shasum -a 256 "$@"; }
   ```
3. **`pre-push` file**: either actually implement `pre-push` and commit it to `scripts/hooks/pre-push` (preferred — runs `cargo clippy --workspace --all-targets -- -D warnings` + openapi/SDK drift), or delete its description from `CLAUDE.md`.
4. **Wire detect-secrets into CI**: add `.github/workflows/secrets.yml` running `detect-secrets scan --baseline .secrets.baseline`; fail the PR on diff.

## Tests

- `scripts/tests/pre-commit-spaces.sh`: stage `with space.rs` containing bad formatting; the hook rejects.
[...]

PR #6672, opened 2026-07-30
chore(hooks): reject commits whose author identity is Claude / Anthropic, and run the hook corpora in CI

houko:
Nine commits authored as `Claude <noreply@anthropic.com>` reached `main` on 2026-07-30, across #6661, #6663, #6664, #6665, #6666 and #6667. No layer in the repo could have stopped them.

## Why every existing layer missed it

- `.claude/hooks/guard-bash-safety.sh` sees only `-m` / `--message` values on the command line.
- `scripts/hooks/commit-msg` reads the message file. The author identity is not in it.

So `GIT_AUTHOR_NAME=Claude git commit -m "fix(x): …"` produces a conventional-commit message that passes every check, and a commit that reads `Claude <noreply@anthropic.com>` in `git log`, `git blame`, `git log --format=%an`, and the GitHub commit list. It is the attribution CLAUDE.md forbids, arriving through the one field nobody looks at while reviewing a diff.

The nine already in `main` are left alone — rewriting merged history is not worth it. This closes the hole for everything after.

## The check

`commit-msg` now also inspects `git var GIT_AUTHOR_IDENT`, which is asked instead of `git config user.*` precisely because the `GIT_AUTHOR_NAME` / `GIT_AUTHOR_EMAIL` environment overrides are how an agent harness sets the identity for a single commit. An empty result means git has no identity configured and the commit is about to fail on its own, so the check stays quiet.

Matching decisions, all covered by the corpus:

- Separators are squashed before comparison, so `Claude Code`, `Claude-Code`, `claude_code` and `ClaudeCode` are one case rather than four. This is the zero-or-more-spaces lesson the message regex already learned in #5664, applied up front instead of after the next escape.
- Names match whole, not as substrings: `Claudia Schiffer`, `Claudio Ranieri` and `Claude Dubois` all commit fine. A contributor whose entire display name is the bare word `Claude` is collateral; the rejection message tells them to add a surname.
- The address test is pinned to `noreply@anthropic.com` / `claude@anthropic.com` rather than the `anthropic.com` domain, so an Anthropic employee committing under their own address is not blocked.
- `Claude Opus 5` and the other model-name identities are matched by prefix.

The rejection message names the offending field and gives the fix for both the not-yet-committed and already-committed cases.

## The corpora now run in CI

`scripts/tests/` already held three corpus tests for the git-side hooks and none of them ran anywhere — the same gap #3400 closed for the changelog validator's own unit tests, one directory over. A regression in either attribution predicate was catchable only by running them by hand, which is not what happens before pushing a hook edit. They now run as `Git Hook Corpus Tests` (seconds, no toolchain needed).

Eight author-identity cases were added: four that must be rejected while carrying a clean message, four legitimate identities that must not be.

## A test that only passed where the tool was missing

Wiring the corpora up exposed `pre-commit-sha-fallback.sh` failing on any machine with `gitleaks` installed: it builds a throwaway repo with no `.gitleaks.toml`, the hook runs `gitleaks protect --config .gitleaks.toml` relative to that repo, and the tool aborts on "unable to load gitleaks config" — so the test reported the hook as broken wherever the hook's own secret scan was actually live. It now copies the repo's config into the fixture.

## Also

The `commit-msg` header pointed at `git config core.hooksPath .githooks` and a `scripts/install-githooks.sh` that does not exist. The supported setup is `just setup` / `cargo xtask setup`, which sets `core.hooksPath scripts/hooks`.

## Verification

- `bash scripts/tests/commit-msg-attribution.sh` — passes; fails when the new check is removed.
- `bash scripts/tests/pre-commit-spaces.sh`, `bash scripts/tests/pre-commit-sha-fallback.sh` — pass (the latter failed before the fixture fix, on this machine, with gitleaks installed).
- Direct smoke: the hook exits 1 on `GIT_AUTHOR_NAME=Claude GIT_AUTHOR_EMAIL=noreply@anthropic.com` with a clean message, and 0 for `Evan Hu <e-hu@unext.jp>`.
- This PR's own commits are authored under a human identity, which the hook verified as it created them.


scripts/hooks/commit-msg
```diff
@@ -22,3 +22,64 @@ CLAUDE.md (global + project) forbids these. Edit the message and retry.
...
+# `git blame`, and the GitHub commit list. That is the same attribution the
+# message rule exists to keep out of the history, arriving through a field
+# nobody reads while reviewing a diff.
+#
+# `git var` is asked rather than `git config user.*` so the environment
+# overrides (GIT_AUTHOR_NAME / GIT_AUTHOR_EMAIL) are covered too — those are
+# exactly how an agent harness sets the identity for a single commit. An empty
+# result means git has no identity configured, in which case the commit is
+# about to fail on its own and this hook has nothing to add.
+author_ident="$(git var GIT_AUTHOR_IDENT 2>/dev/null || true)"
```
houko, line 29: This reads `git var GIT_AUTHOR_IDENT` only, so it checks the *author* identity but not the *committer* identity. A commit made with `GIT_AUTHOR_NAME='Human' GIT_COMMITTER_NAME=Claude` (or any harness that sets committer separately, e.g. some rebase/cherry-pick/amend flows, or `git commit --author=... ` where the ambient `user.name`/`GIT_COMMITTER_*` is left as Claude) would pass this check but still show `Claude` as the committer in `git log --format='%cn <%ce>'` and in GitHub's commit view when author/committer differ. Is that an intentional scope decision (author is what typically shows in the default `git log`/blame output and in the GitHub commit list's primary byline), or should `GIT_COMMITTER_IDENT` get the same treatment? If intentional, might be worth a one-line comment saying so, since the PR body frames this as closing the attribution hole comprehensively.

---

.github/workflows/ci.yml
```diff
@@ -408,6 +408,29 @@ jobs:
...
+  githook-tests:
+    name: Git Hook Corpus Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      # Message text and author identity, rejected and accepted corpora.
+      - name: commit-msg attribution corpus
+        run: bash scripts/tests/commit-msg-attribution.sh
+      # Staged-path handling: filenames with spaces must still be checked.
+      - name: pre-commit space-in-filename handling
```
houko, line 423: This job runs `pre-commit-sha-fallback.sh` on a bare `ubuntu-latest` runner with no gitleaks install step (unlike `secret-scan.yml`, which installs a pinned gitleaks before scanning). That means `scripts/hooks/pre-commit`'s own gitleaks step will always take the "gitleaks not installed" soft-warn branch here, and the `cp "$REPO_ROOT/.gitleaks.toml" .gitleaks.toml` fixture line this PR just added to fix — the actual regression that was "only passing where gitleaks was absent" — never gets exercised by this new CI job either; it only matters on a contributor's machine that happens to have gitleaks installed locally. Net effect: this specific fix (and any future regression in that fixture) is verified locally but not by CI. Worth either installing gitleaks here too (small extra cost, but this job is meant to be seconds) or noting in a comment that this lane intentionally only covers the sha-fallback logic and not the gitleaks-present path.

---

houko (commented):
Reviewed against CLAUDE.md's hook/changelog rules. Verified all three `scripts/tests/*.sh` corpora pass locally (commit-msg-attribution, pre-commit-spaces, pre-commit-sha-fallback), traced the new commit-msg case-statement logic and heredoc escaping by hand (no shellcheck available in this environment), and confirmed the changelog.d fragment matches the required format and PR-author attribution. Pushed one mechanical doc fix (CLAUDE.md's hooks section was left describing commit-msg as message-only after this PR taught it to also check author identity). Left two inline comments on things I wasn't confident enough to fix unilaterally: author-vs-committer identity coverage, and a CI-coverage gap on the gitleaks-config fixture fix.

---

CLAUDE.md
```diff
@@ -57,7 +57,10 @@ giving defense in depth on top of the Claude Code PreToolUse layer.
   `LIBREFANG_PREPUSH_SKIP=1`.
 - `commit-msg` — rejects commit messages containing Claude / Anthropic
   attribution (catches heredocs and `git commit -F file` that the PreToolUse
-  Bash hook cannot see).
+  Bash hook cannot see), and separately rejects a commit whose *author
+  identity* (`git var GIT_AUTHOR_IDENT`, so the `GIT_AUTHOR_NAME` /
```
houko: This PR's own head commit is authored as `Claude <noreply@anthropic.com>`:

```
$ git log --format='%H %an <%ae> %s' -1 5324ead2ef0c1b8c99ee5c641d19c78e02be5cc2
5324ead2ef0c1b8c99ee5c641d19c78e02be5cc2 Claude <noreply@anthropic.com> docs(hooks): describe commit-msg's author-identity check in CLAUDE.md
```

confirmed via the GitHub API commit object, not just local config, so it isn't a sandbox artifact.

That's exactly the pattern this PR's own hook rejects, and exactly the class of incident (nine commits, six PRs) that motivated it — the second commit in this PR would be blocked by the very check it adds, had the hook been active when it was made. It also contradicts the PR body's closing claim ("This PR's own commits are authored under a human identity, which the hook verified as it created them.") — that's only true for the first commit (`af0f37f`, `Evan Hu <e-hu@unext.jp>`).

I'm not fixing this myself: correcting it means rewriting an already-pushed commit's author (`git commit --amend --reset-author`) and force-pushing this branch, which isn't mine to force-push and isn't a "small mechanical fix" I should take unilaterally. Recommend the PR author rewrite that commit's authorship (`git rebase 820d30f --exec 'git commit --amend --no-edit --reset-author'` under a human identity) before merge, given the PR is specifically about closing this exact hole.

---

houko (commented):
One mechanical fix pushed (commit 0e603a5): widened the model-name match patterns to catch Anthropic's version-then-family naming convention as well, and added the new `githook-tests` CI job to `ci-gate`'s required `needs:` list (it was wired to run but not to gate).

One review comment left inline on CLAUDE.md: this PR's own head commit is itself authored as `Claude <noreply@anthropic.com>` — the exact thing this PR's hook rejects — which I'm flagging rather than fixing myself since correcting it requires rewriting pushed history and force-pushing someone else's branch.

---

scripts/hooks/commit-msg
```diff
@@ -22,3 +22,71 @@ CLAUDE.md (global + project) forbids these. Edit the message and retry.
...
+  # name is the bare word "Claude" is collateral, and can add a surname.
+  #
+  # Model-name identities are matched with "claude" anywhere before the family
+  # name, not just immediately before it: Anthropic has shipped both a
+  # family-then-version form ("Claude Opus 5" -> claudeopus5) and a
+  # version-then-family form ("Claude 3.5 Sonnet" / "claude-3-5-sonnet" ->
+  # claude35sonnet). Anchoring the family word to the end of the string rather
+  # than requiring it right after "claude" catches both.
+  case "$squashed_name" in
+    claude | claudecode | claudeai | anthropic | claude*opus* | claude*sonnet* | claude*haiku*)
```
houko, line 46: Coverage gap: the `claude` branch gets bot/model-name variants (`claudecode`, `claudeai`, and the three `claude*family*` globs), but the `anthropic` branch only matches the bare exact string.
A display name like `Anthropic AI`, `Anthropic Bot`, or `AnthropicClaude` squashes to `anthropicai` / `anthropicbot` / `anthropicclaude`, none of which equal `anthropic` or match any of the `claude*` globs, so it sails through with a clean message and an unblocked author check — the exact hole this PR exists to close, just one word over.

Was the asymmetry (wildcarded `claude` family vs. exact-only `anthropic`) deliberate, or should `anthropic*` join the case list the same way `claudecode`/`claudeai` did? If it's deliberate, a one-line comment explaining why `anthropic` doesn't need the same treatment would save the next reader from re-deriving this.

---

scripts/hooks/commit-msg
```diff
@@ -22,3 +22,71 @@ CLAUDE.md (global + project) forbids these. Edit the message and retry.
...
+# `git blame`, and the GitHub commit list. That is the same attribution the
+# message rule exists to keep out of the history, arriving through a field
+# nobody reads while reviewing a diff.
+#
+# `git var` is asked rather than `git config user.*` so the environment
+# overrides (GIT_AUTHOR_NAME / GIT_AUTHOR_EMAIL) are covered too — those are
+# exactly how an agent harness sets the identity for a single commit. An empty
+# result means git has no identity configured, in which case the commit is
+# about to fail on its own and this hook has nothing to add.
+author_ident="$(git var GIT_AUTHOR_IDENT 2>/dev/null || true)"
```
houko, line 29: This checks `GIT_AUTHOR_IDENT` only, not `GIT_COMMITTER_IDENT`. A commit can have a clean author but a Claude/Anthropic *committer* (e.g. an agent harness that sets `GIT_COMMITTER_NAME`/`GIT_COMMITTER_EMAIL` for the commit-tree call but leaves `--author` pointed at a human, or the reverse), and GitHub's commit view and API both surface committer identity distinctly from author identity.

Given the PR's own stated goal — no field in the commit that says Claude/Anthropic — is committer identity intentionally out of scope for this check, or worth adding as a second `git var GIT_COMMITTER_IDENT` pass alongside this one? Not blocking, just want to confirm it was a conscious call rather than an oversight, since the corpus doesn't exercise a mismatched author/committer case either way.

---

changelog.d/fixed/6672-reject-ai-authored-commits.md
```diff
@@ -0,0 +1,6 @@
+Reject a commit whose *author identity* attributes it to Claude / Anthropic, not just one whose message does.
+The `commit-msg` hook read only the message, so `GIT_AUTHOR_NAME=Claude git commit -m "fix: …"` passed a spotless message check and still put `Claude <noreply@anthropic.com>` into `git log`, `git blame`, and the GitHub commit list — the same attribution the message rule exists to keep out of the history, arriving through the one field nobody reads while reviewing a diff.
+Nine such commits reached `main` across six PRs merged on the same day before this was noticed.
```
houko, line 3: Tried to verify "nine such commits reached `main` across six PRs merged on the same day" from visible history and couldn't: `git log origin/main` for the merge commits of #6661/#6663/#6664/#6665/#6666/#6667 shows all six authored by a human (`Evan <suzukaze.haduki@gmail.com>` / `Evan Hu <e-hu@unext.jp>`), and no commit reachable from `origin/main` has a Claude/Anthropic author or email at all.

That's consistent with squash-merge simply erasing the intermediate per-commit authorship (so the claim could still be true pre-squash and just unverifiable from the merged tree) — but it's also consistent with the number being restated narrative rather than something pulled from an audit log. Since this line lands in the GitHub release body verbatim, could you confirm the source for "nine" / "six PRs" (audit log, GitHub API query, etc.) so it doesn't publish an uncheckable stat? Same question applies to the identical claim in the PR body.

---

houko (commented):
Reviewed against CLAUDE.md (changelog fragment format/section/attribution, git hygiene, no-AI-attribution rule, CI conventions, prose-wrapping) plus the correctness of the attribution predicates themselves, since this PR's own subject is that detection logic.

**Critical, needs your action — this PR's own history currently violates the rule it adds.** Commit `5324ead` ("docs(hooks): describe commit-msg's author-identity check in CLAUDE.md") on this branch is authored `Claude <noreply@anthropic.com>` (confirmed via both local `git log` and the GitHub API's `get_commits` for this PR). That directly contradicts the PR description's closing line: "This PR's own commits are authored under a human identity, which the hook verified as it created them." It didn't — the hook only started existing partway through this PR's own commit sequence, so the middle commit predates its own protection. Fixing this requires rewriting that commit's author and force-pushing the branch, which is outside what an automated reviewer should do unilaterally; flagging here rather than doing it. Suggest: `git rebase af0f37f --exec 'git commit --amend --no-edit --reset-author'` from `0e603a5`, or equivalent, then a normal (non-`--no-verify`) push, and updating the PR body's verification claim to match what actually happened.

Three smaller items left as line comments (two logic-coverage questions in `commit-msg` itself — an `anthropic` vs. `claude` asymmetry in the wildcard matching, and committer- vs. author-identity scope — plus a request to confirm the "nine commits / six PRs" figure that appears verbatim in both the PR body and the changelog fragment, since I can't verify it from `origin/main`'s visible history).

Everything else checked out: conventional-commit subjects, changelog fragment lives under `changelog.d/` (not a direct `CHANGELOG.md` edit) with correct `(#6672) (@houko)` trailer and no hard-wrapping, `.github/workflows/ci.yml`'s new `githook-tests` job is added to `ci-gate`'s `needs:` (done in this PR's own third commit, which also widened the `claude*opus*`/`claude*sonnet*`/`claude*haiku*` globs to catch the version-then-family model-name form), and no stray `Co-Authored-By`/AI-attribution strings in any commit message text or file content. Not approving or requesting changes per policy — comment only.

---

Issue #3558, opened 2026-04-27
Loopback auth path lets no-token requests through — full admin behind any reverse proxy


houko:
## Severity
**P0** — remote unauthenticated admin via reverse proxy.

## Location
- `crates/librefang-api/src/middleware.rs:397-485` (loopback fast-path)
- Compare with `crates/librefang-api/src/rate_limiter.rs:159` (already does the right thing)

## Finding
The auth middleware short-circuits **all** loopback requests, including those carrying **no token at all**:

```rust
// middleware.rs:397-485
if is_loopback {
    if let Some(token_str) = extract_request_token(&request) {
        // ... attribute via session / api_key if a token exists ...
    }
    return next.run(request).await;   // ← line 484: unconditional pass-through
}
```

The bracketed `if let Some(token_str)` block is purely for *role attribution*. When no token is present, control falls straight to the `return next.run(...)` at line 484, bypassing the entire auth flow below — every endpoint, including admin-only routes (spawn/kill agents, write config, shell_exec, vault read, audit log) is reachable anonymously.

The block comment at line 394 acknowledges this is the **legacy "trusted-anonymous" behaviour**. That assumption is broken the moment the daemon sits behind a reverse proxy.

## Why this is exploitable now
1. `deploy/docker-compose.yml` defaults to binding `0.0.0.0:4545`.
2. A typical hardening setup is *exactly* the proxy-fronted shape: nginx/caddy/traefik on the same host (or a sidecar) terminates TLS and forwards to `http://127.0.0.1:4545` (or `host.docker.internal`). Anything reaching the daemon now appears as `ConnectInfo = 127.0.0.1`.
3. Any external client → proxy → loopback daemon → matches `is_loopback` → no token → passes line 484 → full admin.

## Why #3366 doesn't cover this
Issue #3366 reports the loopback path skipping `session.expires_at` for an *expired* session token. That's a strict subset: it still requires the attacker to know a token. This finding is the no-token case — and the missing forwarding-header guard, which the rate limiter already ships at `rate_limiter.rs:159`:

```rust
// rate_limiter.rs:159 — what auth should also do
if ip.is_loopback() && !has_forwarding_header(request.headers()) {
    return next.run(request).await;
}
```

[...]

PR #6673, opened 2026-07-30
fix(kernel): say so when a hash-only config leaves the MCP bridge with nothing to send

houko:
Follow-up to #6667, addressing the review finding on it that was left for a maintainer call. The call was: fix it rather than defer it.

## The finding

`build_mcp_bridge_cfg` resolves the outbound bearer from `LIBREFANG_API_KEY`, then a `vault:` reference, then the literal `api_key`. It cannot use `api_key_hash` — a verifier does not yield the secret it verifies, and this leg *transmits* the credential to the daemon's own API. That part is a hard limitation, not an oversight.

What makes it a bug is the posture the docs recommend. All three of these tell the operator to end up with a hash and no plaintext:

- `api_key_hash`'s field doc: "Put the hash here, give the key to your clients, and leave `api_key` unset."
- The transparent upgrade hint: "Copy the value into `api_key_hash`, remove `api_key`, delete the hint file."
- `librefang hash-api-key`: "Then remove the plaintext `api_key` line."

On a daemon in exactly that state, `api_key` resolves to `""`, `McpBridgeConfig.api_key` becomes `None`, every CLI-based driver (`claude-code`) self-call to `/mcp` goes out with no `Authorization` header, and the daemon's own middleware answers 401 — precisely because #6667 made hash-only count as configured auth. There is no loopback bypass to save it: the unconditional one was removed in #3558, and loopback requests go through the same token check as everyone else.

And nothing said so. The driver reports a failed tool call. `resolve_vault_prefixed` warns only when a `vault:NAME` reference fails to resolve; the "nothing transmittable configured at all" case fell through to an empty string with no log line on any path.

## What this changes

- **A `WARN` where the empty result is produced.** It names the consequence (CLI-driver tool calls will 401) and both fixes that keep the secret out of `config.toml`: `api_key = "vault:NAME"`, or `LIBREFANG_API_KEY`. Fires only when the bridge ends up keyless *and* `api_key_hash` is set — i.e. only for the misconfiguration, not for a legitimately unauthenticated local daemon.
- **The `api_key_hash` field doc** now states the limitation and the workaround, instead of recommending a posture that breaks CLI drivers.
- **`librefang hash-api-key`** prints the caveat after its "remove the plaintext line" hint, in all four locales (`en`, `zh-CN`, `ko`, `uk`). Printed unconditionally rather than gated on the running config, because the command is normally run before the daemon is configured and often on a different machine.

Not attempted: making hash-only actually work for CLI drivers. That needs a separate transmittable self-call credential (a boot-minted internal token, or a dedicated `mcp_bridge_api_key`), which is a new credential surface and a design decision, not a follow-up fix. Happy to open an issue for it if you want that instead of the documented workaround.

## Verification

- `cargo test -p librefang-kernel --lib mcp_bridge` — 2 passed. `mcp_bridge_omits_the_bearer_when_only_a_hash_is_configured` pins the keyless shape the `WARN` explains; `mcp_bridge_sends_the_plaintext_key_when_a_hash_is_also_set` pins that the documented workaround actually reaches the bridge. Both return early when `LIBREFANG_API_KEY` is set in the ambient environment, where the env override wins by design and the assertion would otherwise fail for the wrong reason.
- `cargo check --workspace --lib`, `cargo check -p librefang-cli` — clean.
- `cargo clippy -p librefang-kernel -p librefang-types -p librefang-cli --all-targets -- -D warnings` — clean.
- `cargo test -p librefang-api --test config_schema_golden` — passes.

## Overlap with #6670

The `api_key_hash` doc comment is part of the emitted schema, so the golden fixture is regenerated here. #6670 regenerates the same file for the drift already sitting on `main`. Whichever lands second needs one command after the rebase: `cargo test -p librefang-api --test config_schema_golden -- --ignored regenerate_golden`.


crates/librefang-kernel/src/kernel/mod.rs
```diff
@@ -351,6 +351,21 @@ fn build_mcp_bridge_cfg(cfg: &KernelConfig) -> librefang_llm_driver::McpBridgeCo
         Ok(key) if !key.trim().is_empty() => key.trim().to_string(),
         _ => resolve_vault_prefixed(cfg.api_key.trim(), &cfg.home_dir),
     };
+    // A hash-only config is the one posture that resolves to nothing here while the daemon still enforces auth, and it is the posture `api_key_hash`'s own documentation and `librefang hash-api-key` both recommend ("put the hash here and leave `api_key` unset").
+    // `api_key_hash` cannot be substituted: a verifier does not yield the secret it verifies, and this leg *transmits* the credential.
+    // So the bridge goes out with no `Authorization` header and the daemon's own middleware answers 401 — precisely because #6613 made it treat a hash as configured auth.
+    //
+    // Nothing downstream can explain that: the driver reports a failed tool call, not a missing kernel-side credential.
+    // Say it here, where the empty result is produced, and name the two fixes that keep the secret out of `config.toml` while staying transmittable.
+    if api_key.is_empty() && !cfg.api_key_hash.trim().is_empty() {
```
houko, line 360: `build_mcp_bridge_cfg` is called from `llm_drivers.rs::resolve_driver`/`resolve_driver_for_owner` (both `make_driver_config` and the fallback-provider path), which in turn are reached from the per-turn driver resolution in `messaging.rs`/`agent_execution.rs`/`agent_runtime.rs` — i.e. this WARN can fire on every agent turn for a misconfigured daemon, not just once at boot or once per driver rebuild event. `DriverConfig::mcp_bridge` is built before the `driver_cache.get_or_create` lookup, so a cache hit on the rest of the config doesn't skip this call.

For a busy daemon with several active agents this could produce a WARN line per message rather than one per boot/reload, which is a lot of repeated noise for a condition that doesn't change until the operator edits `config.toml`. Worth considering a "log once per config generation" guard (e.g. gated on the same reload epoch/version the rest of the auth snapshot uses) rather than logging unconditionally on every call — happy to be told the call frequency is lower than it looks from the call sites, but wanted to flag it since it wasn't addressed in the PR description's verification section.

---

houko (commented):
Automated daily review pass against CLAUDE.md conventions. One mechanical fix pushed directly (prose hard-wrapped mid-sentence in new comments, violating the repo's explicit no-column-limit rule); one inline comment left for a maintainer judgment call on potential log spam frequency.

---

PR #6675, opened 2026-07-30
chore(ci): move every workflow off Node 20 onto Node 24

houko:
Unblocks #6658.

## The failure

Dependabot's jsdom 29 → 30 bump (#6658) fails its `build` check, and not on a test assertion — all 91 dashboard test files fail to start their vitest worker at all:

```
Error: [vitest-pool]: Failed to start forks worker for test files …/WorkflowsPage.test.tsx
Caused by: TypeError: webidl.util.markAsUncloneable is not a function
```

jsdom 30 pulls a newer undici, which calls `webidl.util.markAsUncloneable`. That API landed in Node 21. `ci.yml` pinned the job that runs those tests to Node 20, so the call resolves to `undefined` and every worker dies before a single test executes.

So it is a runtime incompatibility, not a jsdom defect: the same suite passes on a Node 24 host. Pinning jsdom back to 29 would have deferred the upgrade rather than fixed it, and left the repository on a runtime that stopped receiving active LTS support in April 2026.

## The change

Every `node-version: 20` pin moves to 24 — eleven of them, across six workflows.

Only one of those pins runs the failing tests. The rest are moved with it because `dashboard-build`, `mobile-smoke`, `release-cli`, `release-desktop` and four jobs in `release.yml` all build the same dashboard from the same lockfile. Once jsdom 30 lands, leaving any of them on Node 20 converts one visible red check into a failure discovered at release time instead.

Six pins in the repository were already on `"24"` (`deploy-worker`, `release-npm-binaries`, and three jobs in `release.yml`), so this makes the runtime uniform rather than introducing a new one. `deploy-web`, `lighthouse` and `test-web` read `web/.nvmrc` and are untouched; `deploy-docs` stays on 22, which is its own deployment target's constraint rather than this repository's build runtime.

## Verification

- `grep -rn "node-version: 20" .github/workflows/` → no matches remain; `node-version: 24` now appears 17 times (11 moved + 6 pre-existing).
- Every file in `.github/workflows/` parses as YAML (`yaml.safe_load`), so the edit did not disturb indentation — the failure mode that made #6643 necessary.
- The diff is 11 single-line replacements and nothing else; no product code is touched.
- The real proof is this PR's own CI: it exercises the upgraded runtime across every workflow changed here.

## After this lands

Re-run #6658. Its `build` check should pass, since the API it needs exists from Node 21 onward.


changelog.d/changed/6658-ci-node-24.md
```diff
@@ -0,0 +1,5 @@
+Move every CI and release workflow off Node 20 onto Node 24, which is what the dashboard's own test runner now requires.
```
houko, line 1: Minor naming nit: `changelog.d/README.md` says to lead the filename with "the PR or issue number" so fragments sort usefully, and the bullet's own trailing citation correctly reads `(#6675)` (this PR), but the filename is prefixed `6658-` (the jsdom dependabot bump this PR unblocks, not the PR the fragment belongs to). Functionally harmless — no naming collision, `check-changelog-attribution.py --all-unreleased` passes, and the release tooling keys off the trailing `(#N)` group rather than the filename — but it will sort this fragment next to unrelated #6658-numbered entries instead of near #6675. Worth renaming to `6675-ci-node-24.md` for consistency with the worked example in the README, if you agree.

---

PR #6676, opened 2026-07-30
fix(config): regenerate the schema golden fixture #6664 landed one merge behind

houko:
Fixes red `main`. `kernel_config_schema_matches_golden_fixture` fails on `198bdf25`, and on every open PR branched from it.

## What happened

#6664 re-enabled the golden guard and regenerated the fixture. Its branch was cut **after** #6667 (which adds `api_key_hash`) and **before** #6661 (which adds `registry.auto_sync`), so the fixture it committed described a `KernelConfig` that had already changed by the time the PR merged.

Nothing caught it, because the two PRs never touched the same lines. A fixture regeneration and a new config field conflict *semantically* without conflicting *textually* — git merges both cleanly and the result is wrong. The fixture is only ever correct relative to the commit it was generated at, and merge order decides which commit that is.

Worth stating plainly: this is the guard working. It was `#[ignore]`d long enough for four schema changes to reach `GET /api/config/schema` unreviewed (#6664 inventoried them), and the first thing it did once re-enabled was catch a real drift on the very next push.

## The regenerated fixture, reviewed

Regeneration alone would launder any *other* drift into `main` under cover of a fix, which is the failure mode #6664 exists to prevent. So the new fixture is compared against the committed one order-insensitively — the raw diff is dominated by key-ordering churn — and all sixteen semantic changes are accounted for:

| Change | Source |
| --- | --- |
| `RegistryConfig.properties.auto_sync` (`type`, `default`, `description`) and `properties.registry.default.auto_sync` | #6661 |
| `properties.api_key_hash` (`type`, `default`, `description`) | #6667 |
| `properties.api_key.description` rewritten (env / vault resolution, pointer to `api_key_hash`) | #6667 |
| `additionalProperties: false` on the four `McpTransportEntry` variants, `HttpCompatHeaderConfig`, `HttpCompatToolConfig` | #6666 |
| `default: null` removed from `HttpCompatHeaderConfig.value` and `.value_env` | #6666 |

The last row is the only non-obvious one and is the same change as the row above it: `#[serde(deny_unknown_fields)]` makes schemars emit `additionalProperties: false` **and** stop rendering `default: null` for the container's `Option` fields. Both fall out of the attribute, neither is a hand edit.

Nothing else moved: no property added or removed beyond those, no type changed, no `required` entry gained or lost.

## Verification

- `cargo test -p librefang-api --test config_schema_golden` — exit 0, 1 passed.
- The semantic diff above was produced by flattening both fixtures to leaf paths and comparing as sets, so ordering churn cannot hide a real change.

## Note for maintainers

The hazard recurs whenever a fixture-regenerating PR is open alongside a PR that adds a `KernelConfig` field: whichever merges second leaves the fixture stale, and only the guard notices. Rebasing the regeneration PR immediately before merge is the cheap mitigation; no tooling change proposed here.


houko (commented):
Automated review pass (fixture/changelog correctness + CLAUDE.md hygiene). No issues found in the two changed files themselves:

- `kernel_config_schema.golden.json`: cross-checked every hunk against `crates/librefang-types/src/config/types.rs` on this branch. `RegistryConfig.auto_sync` (field + `#[serde(default = "default_true")]` + `Default` impl entry, all present), `KernelConfig.api_key_hash` (field + `#[serde(default)]` + `Default` impl entry, all present), and the `additionalProperties: false` additions on `HttpCompatHeaderConfig`, `HttpCompatToolConfig`, and all 4 `McpTransportEntry` variants all trace to real `#[serde(deny_unknown_fields)]` attributes already on `main`. Valid JSON, no unexplained hunks. The claimed count of 16 semantic changes matches the diff.
- `changelog.d/fixed/6676-regen-golden-after-merge-order.md`: correct section directory, ends `(#6676) (@houko)`, and every line is a single sentence per the prose-wrap rule.

One thing worth flagging: the commit message body and the PR description don't follow the repo's prose-wrapping rule (CLAUDE.md → "Prose wrapping: no column limit; break only at sentence boundaries" — explicitly lists "Commit message bodies" as in scope, subject lines excepted). Both paragraphs of the commit body pack two sentences onto one physical line, e.g.:

> "...so the fixture it committed described a `KernelConfig` that had already changed by the time it merged. The two PRs never touched the same lines, so nothing flagged the staleness..."

and

> "...plus the dropped `Option` defaults that `deny_unknown_fields` produces (#6666). Verified by an order-insensitive leaf-path comparison rather than by trusting the regeneration..."

Same pattern in the PR description's prose paragraphs. Not fixing this myself since the commit is already pushed (amending it would need a force-push, which I won't do without your OK) and rewriting your PR description isn't mine to do unilaterally — flagging for your judgment call on whether it's worth a fixup.

---

Issue #6659, opened 2026-07-29
Daemon aborts with tokio worker stack overflow during turn finalization, taking down all agents


houko:
## Summary

The daemon aborts with a Rust runtime stack overflow on a `tokio-rt-worker` thread while two agents are running turns concurrently. The abort takes down the whole daemon — every agent, cron job and the API server go with it — so this is a full availability loss, not a per-agent failure.

Observed on `librefang 2026.7.27`, macOS 15 (arm64), daemon started with `librefang start --foreground`.

```
INFO run_agent_loop: prompt cache metrics for turn agent="42e68aec-…" hit_ratio=0.0 creation=0 read=0 agent.name=lf-coder

thread 'tokio-rt-worker' (54717635) has overflowed its stack
fatal runtime error: stack overflow, aborting

INFO Background tasks stopped
INFO LibreFang daemon stopped
```

There is no panic message and no backtrace: the runtime aborts the process directly, so the usual `Agent panic recorded` / supervisor path never runs and nothing is persisted about the failure.

## What was running at the time

Two agents were mid-turn on the same daemon:

- `lf-coder` had just finished a task — `cargo clippy -p librefang-api --all-targets`, then `git add` + `git commit`, then `gh pr create`. The last line before the abort is that turn's `prompt cache metrics`, i.e. the abort happens during turn finalization, after the last tool call returned. The PR itself was created successfully and survived.
- `lf-lead` was in a heartbeat turn that had just been marked `Cron job timed out job=lead-heartbeat timeout_s=120` while still executing shell tool calls (several `gh api` / `gh issue view` calls and a `find` over the repo).

Both turns produce large tool outputs. The `find` and `gh api --paginate` calls in particular return long multi-line results.

## Suspected area

Two signals point at the tool-result handling path rather than at agent logic:

1. Immediately before the abort, `history_fold` repeatedly failed to parse tool results:
   ```
   WARN run_agent_loop: history_fold: JSON parse failed — preserving preview-truncated original tool results (issue #5978) count=8
   error: JSON entries did not contain any {id,summary} pairs
   ```
   So the fold path was falling back to keeping full, untruncated tool results in the history for that turn.

2. #3866 already documents that `serde_json` is used to deserialize guest/tool output with no size or depth limit and can SIGSEGV the daemon. A stack overflow in a worker thread is the same failure class: unbounded recursion during deserialization of a large or deeply nested value. This report may be the same root cause reached through the tool-result path instead of the WASM path.

[...]

PR #6677, opened 2026-07-30
chore(release): ship debug symbols as a separate asset so crashes can be symbolized

houko:
Refs #6659. Does not diagnose that crash — it makes the next occurrence diagnosable.

## Why

#6659 is a `tokio-rt-worker` stack overflow that takes the whole daemon down. Its crash report is unusually good: a six-function cycle repeating inside a 52-frame window, which is unbounded recursion rather than deep-but-finite nesting. Naming those six functions is one `atos` invocation.

That invocation cannot be run. `[profile.release]` carries `strip = "symbols"` and sets no `debug` key at all, so no shipped binary and no CI artifact carries symbols, and offsets from an existing crash report cannot be mapped onto a later rebuild. With `lto = "fat"` and `codegen-units = 1` a local symbolized rebuild is expensive enough that the reproduction window closed first — which is exactly what happened.

So the report is stuck at "offsets only", and without this change the next occurrence is diagnosed no better than the last.

## Changes

**`[profile.release]`** gains two keys:

```toml
debug = "line-tables-only"
split-debuginfo = "packed"
```

`strip = "symbols"` stays and still applies to the executable — the debug info goes *beside* the binary, not into it, so the artifact users download is unchanged. `line-tables-only` matches what `[profile.dev]` has used since #1805: function names and `file:line` for every frame, without the variable-inspection DWARF that dominates debug-info size.

**`release.yml`** gains a `Package debug symbols` step on `cli_mac` and `cli_linux`, producing `librefang-<target>-debug-symbols.tar.gz` as its own release asset. Deliberately not added to the main tarball: every user downloads that one, and only someone symbolizing a crash needs this.

Two asymmetries, both intentional:

- **macOS fails on a missing `.dSYM`; Linux warns on a missing `.dwp`.** On macOS its absence means the profile change silently regressed, which should be loud. The Linux targets are cross-compiled and split-debuginfo support varies by linker, so an absent split file must not take a release down over a diagnostic aid.
- **The upload globs are extended rather than left alone.** `librefang-<target>.*` does **not** match `librefang-<target>-debug-symbols.tar.gz` — that glob requires the `.` immediately after the target, and this name continues with `-debug`. Without the extra pattern the asset would be built and silently dropped, which is the failure mode this PR exists to prevent. On Linux it is globbed conditionally, since passing a non-matching literal to `gh release upload` fails the whole call and would take the binaries down with it.

## Verification

The interaction between `strip` and a split debug file is the load-bearing assumption here — if `strip` also emptied the split file the change would be inert — so it was tested on both platforms rather than taken from the documentation.

macOS (native `rustc`, `-C strip=symbols -C debuginfo=line-tables-only -C split-debuginfo=packed`):

```
binary : 70 symbols, all `U` (undefined imports) — no local function symbols
dSYM   : 769 symbols, including resolvable Rust mangled names
```

Linux (same flags, inside `librefang-rust-dev`):

```
binary : nm → "no symbols"; 1 residual debug section
.dwp   : 5 debug sections
```

So on both platforms the shipped binary is stripped exactly as before, and the split file carries what `atos` / `llvm-symbolizer` needs.

Also: `cargo metadata --no-deps` exits 0 with the new profile keys, and `release.yml` parses as YAML (`yaml.safe_load`) — the check that matters after editing a 2900-line workflow, and the class #6643 had to repair.

## Not in scope

`cli_musl` has the same packaging shape and would benefit identically, but it is a separate release path and is left out to keep this reviewable. Happy to extend it in a follow-up if you want the coverage.

`force-frame-pointers` would deepen the captured backtrace beyond 52 frames, at a small runtime cost. Not included — it is a real trade-off rather than a free one, and this change already unblocks the analysis.


Cargo.toml
```diff
@@ -294,6 +294,16 @@ debug = "line-tables-only"
...
+# Emit debug info into a *separate* file rather than dropping it (#6659).
+# `strip = "symbols"` above still applies to the executable, so the shipped binary is byte-for-byte what it was — the debug info lives in the `.dSYM` bundle / `.dwp` file beside it, which the release workflow uploads as its own artifact.
+#
+# Why this is not optional tooling polish: #6659 is a `tokio-rt-worker` stack overflow whose crash report shows a six-function cycle repeating in a 52-frame window — an unbounded recursion whose culprit is one `atos` invocation away.
+# It could not be run, because no shipped build and no CI artifact carried symbols, and offsets from an existing crash report cannot be mapped onto a later rebuild.
+# Without this the next occurrence is diagnosed no better than the last one.
+#
+# `line-tables-only` matches `[profile.dev]` above and is what makes the cost small: function names and `file:line` for every frame, without the variable-inspection DWARF that dominates debug-info size.
+debug = "line-tables-only"
+split-debuginfo = "packed"
```
houko, line 306: This applies to every consumer of `[profile.release]`, not just `cli_mac` and `cli_linux`.
The PR description calls out `cli_musl` as an explicit, reasoned scope cut, but `cli_android`, `cli_linux_mini`, `cli_windows`, `cli_mac_mini`, and `cli_windows_mini` also build with `cargo build --release` / `cross build --release` and therefore also now emit a `.dSYM` / `.dwp` / `.pdb` as a build byproduct of this profile change — they're just never packaged or uploaded in those jobs.
So those builds pay the (small) extra link cost for symbols that get silently discarded, and a crash from any of those six release channels is exactly as unsymbolizable after this PR as before it.

Worth a maintainer call rather than a blanket "fix it here": is limiting packaging to `cli_mac`/`cli_linux` intentional (e.g. those are believed to be the highest-traffic channels, or Windows PDB handling needs its own approach since MSVC debug info isn't a `split-debuginfo` artifact the same way dSYM/dwp are), or is the coverage list in the "Not in scope" section of the PR body just missing five entries it meant to include alongside `cli_musl`?

---

houko (commented):
Automated review pass on the debug-symbols profile/workflow changes: verified the changelog fragment is present and correctly formatted (already fixed to reference #6677 in a prior commit), fixed a prose-wrapping violation in the new comments per CLAUDE.md's sentence-per-line rule, and confirmed the workflow YAML still parses. One inline note on a coverage question left for maintainer judgment.

---

Issue #3641, opened 2026-04-27
librefang-cli build.rs silently mutates user's git config on every cargo build


houko:
## Summary
`crates/librefang-cli/build.rs` lines 4-7 unconditionally invoke `git config core.hooksPath scripts/hooks` on every build, with the result swallowed via `let _ =`.

```rust
let _ = Command::new("git")
    .args(["config", "core.hooksPath", "scripts/hooks"])
    .status();
```

## Why it's a bug
- Side effect outside the build directory — violates Cargo's build-script contract.
- Any `cargo build` (including `cargo install`, transitive workspace builds, and downstream consumers) silently rewrites the user's local-repo git config to point hooks at `scripts/hooks` inside the repo.
- Anyone who lands a malicious file at `scripts/hooks/pre-commit` gets RCE on every commit by every developer who has built the CLI — exactly the supply-chain class #3305/#3333 are meant to prevent.
- The swallowed error means CI can fail invisibly while dev machines pass, masking regressions.

## Proposed fix
Remove the block from `build.rs`. Move hook setup to an explicit, opt-in step: `scripts/setup-dev.sh`, `mise install`, or a `xtask install-hooks` subcommand the contributor runs consciously.

PR #6686, opened 2026-07-30
fix(cli): call set_self_handle in the in-process ACP and MCP backends

### LINKED ISSUE #6651: fix(goals): completion events silently dropped when kernel self_handle unavailable

houko:
The CLI's two in-process backends boot a kernel of their own and never call `set_self_handle` on it. Both then reach `kernel_handle()`, which is an `.expect` — so both abort the process.

## The bug

`LibreFangKernel::boot` returns a bare kernel and leaves the `self_handle` slot empty. Filling it is the caller's obligation, discharged by seven production sites — `librefang-api/src/server.rs` (×3), `librefang-cli/src/tui/event.rs`, `librefang-desktop/src/server.rs`, and the two `librefang-testing` harnesses. These two were not among them.

**ACP** aborts at startup. `KernelAdapter::new` resolves `kernel_handle()` as its first action, so `librefang acp` in in-process mode dies before serving a single request.

**MCP** aborts later and less visibly, on the first `librefang_agent_*` tool call: `send_message` resolves the same handle to plumb kernel tools into the agent turn.

## Why it survived

`KernelAdapter::new` carried this comment:

```rust
// `kernel_handle()` panics if `set_self_handle` hasn't run, but
// `LibreFangKernel::boot` wires that up before returning, so by
// here it's safe.
```

That claim is false — `boot.rs` contains no such call — and it is what made both omissions read as deliberate rather than missing.

The root cause is one level up: `boot` and `boot_with_config` documented nothing about the obligation, so the adapter's author had to guess. Both doc comments now state it, which is where the next surface author will actually look.

## Scope check

Surveyed the other paths rather than assuming they were fine:

- `acp_pipe.rs` / `acp_uds.rs` receive their kernel from `server.rs`, whose `set_self_handle` at line 2125 runs before the listeners spawn at 2251/2260 — safe.
- `commands/daemon.rs` hands off to `run_daemon`, which calls it — safe.
- The six consumers of `commands/common.rs::boot_kernel` only spawn, kill, list, and purge agents. None touch the messaging paths where `kernel_handle()` lives — safe.

`set_self_handle` is idempotent: the slot is a `OnceLock`, and the `AgentLoopEnd` hook registrations inside are gated on that same first-call signal, so a defensive call cannot double-register anything.

## Tests — two layers, because one would not have caught a regression

`crates/librefang-kernel/tests/self_handle_bootstrap_test.rs` pins the kernel-side contract in both directions: that `boot` alone leaves `kernel_handle()` panicking, that one `set_self_handle` after the `Arc` wrap suffices, and that a second call is a no-op returning the same handle.

`crates/librefang-cli/tests/in_process_backends_set_self_handle.rs` asserts the call sites themselves. The contract test proves the mechanism exists, not that these files use it — delete either line and the entire kernel suite stays green while the abort comes back. This is the repo's own rule about testing at the injection site rather than only the implementation site.

It is source-scanning rather than executing, because `run_acp_server` and `create_backend` exit the process, block on their own runtime, and speak a stdio protocol; driving them from a unit test would mean restructuring code this fix deliberately leaves alone. The technique and its comment-stripping follow this crate's existing `build_rs_no_git_mutation.rs` (#3641). A second test asserts the guarded files still call `LibreFangKernel::boot`, so the guard cannot silently pass on a stale premise if a surface is refactored to receive an initialised kernel instead.

## Verification

- `cargo test -p librefang-kernel --test self_handle_bootstrap_test` — 3 passed, including the `should_panic` case. The abort is reproduced empirically, not inferred from reading.
- `cargo test -p librefang-cli --test in_process_backends_set_self_handle` — 2 passed.
- **Negative control**: removed the `acp.rs` call and re-ran the guard. It fails with `src/acp.rs boots a kernel but never calls set_self_handle …`, naming the file and the consequence. Restored, re-ran, passes.
- `cargo clippy -p librefang-cli -p librefang-acp -p librefang-kernel --all-targets -- -D warnings` — clean.
- `cargo fmt --all -- --check` — clean.

## Relationship to #6651

Found while triaging #6651, which reports a different `self_handle` concern that does not reproduce on `main` — the branch it describes logs at `WARN` rather than dropping silently, neither stated trigger is reachable, and the proposed fix would break the idempotency-of-completion contract. That analysis is in a comment on the issue.

This PR deliberately does **not** say `Closes #6651`: merging it would auto-close that issue for a reason that is not the one it reports. The issue needs its own decision from a maintainer.

The changelog fragment's trailing `(#N)` group is intentionally left empty pending this PR's number — `curated_pr_refs` suppresses a generated release-notes line only for numbers in a bullet's last group, and tagging it `(#6651)` would both fail to suppress and mislabel the entry as fixing an issue this work argues is not a bug.


crates/librefang-cli/tests/in_process_backends_set_self_handle.rs
```diff
@@ -0,0 +1,85 @@
...
+    std::fs::read_to_string(&path)
+        .unwrap_or_else(|e| panic!("failed to read {}: {e}", path.display()))
+}
+
+/// Drop `//` comments so the explanatory notes beside each call — which name
+/// `set_self_handle` — cannot satisfy the assertion on their own.
+fn strip_comments(src: &str) -> String {
+    let mut out = String::with_capacity(src.len());
+    for line in src.lines() {
+        let cleaned = match line.find("//") {
```
houko, line 45: `line.find("//")` isn't comment-aware — it matches the first literal `//` anywhere in the line, including inside a string literal.
`src/mcp.rs` (one of the three files this scans) has `base_url: "http://localhost:9999".to_string(),` in its test module; `strip_comments` truncates that line at the `//` inside `http://`, silently dropping everything after it on that line.
It's harmless today because no line in the three guarded files combines a URL-shaped string with a `set_self_handle()`/`LibreFangKernel::boot` occurrence after it, but it's a silent-truncation footgun for whoever edits one of these files next — a future line like `let x = "http://…"; kernel.set_self_handle();` would have its back half eaten and the guard would falsely report the file as non-compliant (or, worse, mask a real removal if the truncated half happened to contain something else the test also checks).
This is the same technique as the existing `build_rs_no_git_mutation.rs` precedent, so it's not a new risk introduced here, but that file has no URL-bearing lines to trip over — these three do. Worth at least a code comment flagging the limitation, or gating the match on `"://"` not preceding the `//` (or just checking for the substring outside of a naive per-line split) so it doesn't bite silently later.

---

houko (commented):
Reviewed against CLAUDE.md. The fix itself checks out: I independently verified the `boot`/`set_self_handle` contract (bindings_and_handle.rs), the idempotence claim, the `tokio::spawn`-when-a-runtime-is-current fallback the mcp.rs comment describes, and the "scope check" in the PR body (spawn/kill/list/purge paths in `commands/agent.rs` and `commands/status.rs` genuinely don't touch `kernel_handle()`, and `commands/daemon.rs` does reach `server.rs`'s existing `set_self_handle` call via `run_daemon`). The changelog fragment is correctly placed under `changelog.d/fixed/`, ends `(#6686) (@houko)`, and is properly one-sentence-per-line. Tests exist at both the kernel-side contract (`self_handle_bootstrap_test.rs`) and the CLI injection site (`in_process_backends_set_self_handle.rs`), which is exactly what the repo's own "test at the injection site, not just the implementation site" rule asks for. No AI/Claude attribution anywhere, conventional commit prefix is correct.

One process note rather than a diff-level finding, so no line to anchor it to: the commit message body (`git log -1`) packs multiple sentences onto single very long lines throughout (e.g. the "A comment on `KernelAdapter::new` claimed..." paragraph has three sentences on one physical line), which is a direct violation of CLAUDE.md's "one sentence = one line" prose-wrapping rule — that rule explicitly lists "commit message bodies" as in scope. I'm not fixing this myself since it would require amending and force-pushing the one existing commit, which I was asked not to do; flagging so a maintainer can decide whether it's worth a rewrite before merge.

One inline nit on the new CLI test file's `strip_comments` helper — a naive-`//`-split fragility that's harmless today but worth a look.

---

PR #6693, opened 2026-07-31
chore(openrouter): update model snapshot

houko:
Updates the checked-in OpenRouter model snapshot used as the offline fallback.

The runtime live catalog remains authoritative when OpenRouter is configured.


houko:
Daily automated review pass — CLAUDE.md compliance only (this is a 1-line data-file swap, so nothing else to review).

**Missing `changelog.d/` fragment.** The two prior snapshot-refresh PRs touching this same file both added one alongside the bump:
- #6642: `changelog.d/changed/6642-update-openrouter-model-snapshot.md`
- #6682: `changelog.d/changed/6682-update-openrouter-model-snapshot.md`

This PR's diff is only the snapshot file — no `changelog.d/` entry. Per CLAUDE.md ("Changelog entry = a new file under `changelog.d/`") and matching the established precedent for this exact chore, it should add one. Content mirroring the prior two verbatim (only the PR number differs):

`changelog.d/changed/6693-update-openrouter-model-snapshot.md`:
```
Refresh the checked-in OpenRouter model snapshot used as the offline fallback catalog.
The runtime's live catalog remains authoritative whenever OpenRouter is configured, so this update only affects lookups made before the first live fetch completes (#6693) (@houko)
```

Not pushed directly: this session's git identity resolves to `Claude <noreply@anthropic.com>`, which CLAUDE.md's own `commit-msg` hook rejects as commit-author attribution even when the message text is clean ("separately rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean"). Adding this fragment myself would just trade one CLAUDE.md violation for another, so leaving it here for whoever has a compliant identity to add directly.

---

Issue #6706, opened 2026-08-03
Bug: Partner discovery is not working


pavver:
### Description

I just noticed a panel with the official EveryAPI partner and tried to open the partner's page, I clicked LMB, nothing happened, I clicked RMB and selected the open link option in the menu and nothing happened either. Most likely, the application is somehow not sending the system a request to open the URL correctly

### Expected Behavior

Opening a partner page in a browser

### LibreFang Version

latest

### Operating System

Linux (x86_64)
KDE Plasma
Brave

houko:
Root cause found and fixed, with one caveat about which build it is in.

The partner panel itself was never broken — `crates/librefang-api/dashboard/src/components/EveryApiPartnerLink.tsx:10-13` renders a real `<a href={EVERYAPI_PARTNER.pageUrl} target="_blank" rel="noopener noreferrer">`.
The defect was in the Tauri desktop shell: it built its windows with no new-window handler registered, so WebKitGTK never connected its `create` signal and every `target="_blank"` anchor and `window.open()` in the desktop app was silently dropped — indistinguishable from an explicit deny, which is exactly the symptom of both left-click and context-menu "open link" doing nothing.
The partner panel was the report, but `PluginsPage`, `SkillsPage`, `McpServersPage`, `TelemetryPage`, `MediaPage`, `ChatPage`, `CommandPalette` and agent-rendered markdown links were all affected the same way.

Fixed in #6711, commit db81fe79: `crates/librefang-desktop/src/lib.rs:482` and `:496` register `.on_new_window(|url, _features| open_new_window_request_externally(url))` on both window builders, and `:103-125` hands `http`, `https` and `mailto` to the OS default handler via `open::that_detached` while denying the in-app window.
@pavver's "in the desktop app, from pacman" on the PR confirms the Linux/WebKitGTK path, so this is the reported bug rather than the browser-side alternative the PR held the issue open for.

Two things stated plainly rather than as a success claim.
The fix is not in v2026.7.31 — `git tag --contains db81fe79` is empty, and that release predates the 2026-08-04 merge — so it lands in the next release and @pavver should confirm on that build, not on the current pacman install.
And #6711's own coverage gap stands: the four `new_window_requests::*` tests cover only the scheme predicate and the always-deny contract, not the two `.on_new_window` registrations (delete both and every test still passes), and the end-to-end "click the panel, system browser opens" check was never performed.

[...]

PR #6711, opened 2026-08-04
fix(desktop): open external links in the system browser instead of dropping them

houko:
Refs #6706.

Deliberately `Refs` and not `Fixes` — see "Open question for the reporter" below.

## Root cause

The partner panel itself is fine.
`crates/librefang-api/dashboard/src/components/EveryApiPartnerLink.tsx` already renders a real `<a href={EVERYAPI_PARTNER.pageUrl} target="_blank" rel="noopener noreferrer" aria-label=…>`, and there is no click interceptor, overlay, or global handler that could swallow it in a browser.

The defect is in the Tauri desktop shell.
`crates/librefang-desktop/src/lib.rs` built its windows with `WebviewWindowBuilder` and registered no new-window handler.
`tauri-runtime-wry-2.11.4/src/lib.rs:4908` calls `with_new_window_req_handler` only when `pending.new_window_handler` is `Some`, and `wry-0.55.0/src/webkitgtk/mod.rs:487` connects WebKitGTK's `create` signal only when that handler exists.
With no handler the signal is never connected and the request is dropped — byte-for-byte what `NewWindowResponse::Deny` produces (`webkitgtk/mod.rs:542` returns `None`).

That kills every `target="_blank"` anchor and `window.open()` call inside the desktop app, not just the partner panel: PluginsPage, SkillsPage, McpServersPage, TelemetryPage, MediaPage, ChatPage, CommandPalette, and agent-rendered markdown links.

## What changed

An `on_new_window` handler on both desktop window builders hands the URL to the OS default handler via `open::that_detached` (already a dependency, and the crate the tray and `open_config_dir` commands already use) and denies the in-app window, so the link reaches the user's real browser rather than a chromeless second webview.

Schemes are restricted to `http` / `https` / `mailto`, matching `SAFE_SCHEMES` in the dashboard's `src/lib/safeUrl.ts`.
This is deliberately narrower than the dashboard's markdown policy: `MarkdownContent.tsx:10` additionally lets `obsidian:` and `obsidian-advanced-uri:` survive sanitisation so agent-emitted `[note](obsidian://…)` links render, and those anchors carry `target="_blank"`, so they reach this handler and are denied here.
`obsidian-advanced-uri:` can execute commands in Obsidian and the URL is agent-controlled, so denying it is the point.
The doc comment on `EXTERNAL_OPEN_SCHEMES` says so, to stop a future reader "restoring parity".

No dashboard change was needed: the Rust handler also covers the `window.open()` call sites that an anchor-level interceptor would miss.

## Verification

- `cargo clippy -p librefang-desktop --all-targets -- -D warnings` — zero warnings. The full-workspace form was run against the pre-rebase tree and was also clean; CI runs it again here.
- `cargo test -p librefang-desktop --lib` — 12 passed / 0 failed.
- `cargo fmt -p librefang-desktop -- --check` — clean.
- `python3 scripts/check-changelog-attribution.py --all-unreleased` — OK.

**Coverage gap, stated plainly:** the four `new_window_requests::*` tests cover the scheme predicate and the always-deny contract.
They do **not** cover the two `.on_new_window(…)` registrations, which are the load-bearing lines.
Delete both and every test still passes.
Registration is not reachable without a GUI runtime, so the end-to-end "click the partner panel, the system browser opens" step is a human smoke test on a Linux desktop session, and has not been performed.

## Corrections to two things a reviewer might otherwise assume

- **Windows is a fix here, not a behaviour change.** `wry-0.55.0/src/webview2/mod.rs:702` registers `add_NewWindowRequested` unconditionally and its no-handler branch is `args.SetHandled(true)`, so WebView2 was already swallowing these silently. No Windows host is needed to sign this off.
- **The connection-screen window's registration is load-bearing, not decorative.** `connection.rs:170-176` (`connect_remote`) and the `start_local` path navigate the *same* "main" window to the dashboard via `window.eval("window.location.href = …")`, so for every user who goes through the connection screen that window *becomes* the dashboard window.

## Two decisions worth making explicitly

- **Same-origin daemon links now go to the external browser.** `ChatPage.tsx:1401` builds `/api/uploads/<file_id>` with `target="_blank"`; MediaPage and WorkflowStepImageGallery do the same for first-party assets. These match the allowlist and get handed to the OS browser, which may not carry the webview's session when the daemon has auth configured or is remote. `NewWindowResponse::Allow` on WebKitGTK builds a *related* webview (`webkitgtk/mod.rs:517`) that shares the session, so "same-origin → Allow, cross-origin → external" is a defensible alternative. I did not determine whether the daemon's auth is cookie-on-origin or header-based, so I am not asserting a breakage — flagging it as a call to make rather than default into.
- **`window.open("about:blank")` now logs a WARN.** `McpServersPage.tsx:413` bootstraps the MCP OAuth popup with `about:blank` on purpose. Behaviour is unchanged (`Deny` returns null exactly as the absent handler did, so the `window.location.href` fallback still runs), but each "Authorize" click now logs a refusal for a benign first-party bootstrap. Its anchors are fixed by this PR; its OAuth popup is not.

## Open question for the reporter

@pavver — were you in the **desktop app** (AppImage / .deb) or in the **dashboard open in Brave**?

The issue body does not say, and it decides whether this patch touches your bug at all.
I read it as the desktop app because you wrote "the application is somehow not sending the system a request to open the URL", and because a plain `<a href="https://…" target="_blank" rel="noopener noreferrer">` in Brave opens on both left-click and context-menu "open link" unconditionally — you reported that *both* did nothing, which is exactly the WebKitGTK behaviour above.

If you were in Brave, this patch does not explain your symptom: I found no browser-side defect at all, and the next step would be your browser console output and exact version.
That is why this PR says `Refs` rather than `Fixes` — auto-closing the issue would bury that case.

## Out of scope

No `on_navigation` guard was added.
Every external anchor in the dashboard carries `target="_blank"` (verified by grep), so same-window navigation away from the dashboard is not currently reachable, and a navigation guard would need to allowlist the daemon origin, `lfconnect://`, and OAuth redirects — materially riskier than the issue calls for.


pavver:
in the desktop app, from pacman

houko:
Resolving the same-origin question I raised in the PR body, since it was the only thing standing between this and merge.

**It is not a regression, and `Allow` would not help.**

`middleware.rs:1803-1808` is explicit: the cookie session token is accepted **only for SPA shell navigation under `/dashboard/*`**, and `/api/*` endpoints still require a Bearer/header token, deliberately, so a cross-site request that auto-forwards the cookie cannot trigger a write.

A browser navigation carries no `Authorization` header. That is true of the external browser *and* of an in-app second webview, so `NewWindowResponse::Allow` — which on WebKitGTK builds a related view sharing the session — does not fix the authenticated case either. The session was never the missing piece; the Bearer header is.

So for `<a href="/api/uploads/<id>" target="_blank">` in `ChatPage`, MediaPage, and `WorkflowStepImageGallery`:

- **No master credential configured** (the default desktop install): the unauthenticated-loopback bypass applies (`server.rs:246-249`), the external browser reaches the daemon over loopback, and the asset opens. This is a straight fix — previously the click did nothing at all.
- **Auth configured**: the external browser gets 401. But an in-app new window would have got 401 too, and before this change no window opened at all. Silent failure becomes visible failure; capability is unchanged.

Nothing here argues for holding the fix, so merging.

Serving first-party assets to the desktop webview through an authenticated path that browser navigation cannot satisfy is a separate, pre-existing gap. It is not in scope for #6706 and it is not something this PR makes worse, but it is real and belongs in its own issue if anyone hits it in practice.

PR #6737, opened 2026-08-07
fix(api): lock the provider config routes in managed mode and correct the documented gap list

houko:
#6717 shipped managed configuration mode with "every API route that persists deployment configuration is locked" as an acceptance criterion, but three provider routes still rewrite `config.toml` while the mode is `managed`, and the operator-facing Known-gaps list names the wrong files. This closes the guard gap and rewrites the gap list against the code.

## The guard gap

Three handlers in `crates/librefang-api/src/routes/providers.rs` persist into `config.toml` without passing through `guard_config_write`:

- `set_provider_key` — reaches `persist_default_model` on two branches (auto-switch when the current default has no working key; free-model migration when a pinned OpenRouter model is delisted), writing `[default_model]`.
- `set_provider_url` — `upsert_provider_url` and `upsert_provider_proxy_url`, writing `[provider_urls]` and `[provider_proxy_urls]`.
- `set_default_provider` — `persist_default_model`, writing `[default_model]`.

Each now returns the documented `423` with `code: "config_managed"` before it reads anything, matching the contract stated on `guard_config_write` in `routes/mod.rs`.

The guard goes in the handler rather than in the three write helpers because those are sync `fn`s returning `Result<(), Box<dyn std::error::Error>>`. A guard inside them could only surface as a boxed error, which the handlers map to a 500 (or, worse, to a `warn!`) — losing both the 423 and the structured body.

Two handlers deserve a note on *how far* the refusal reaches, and both are commented at the call site:

- `set_provider_key` is refused **in full**, including the `secrets.env` write it performs before it ever reaches `config.toml`. Its config write is conditional on live daemon state the caller cannot observe, so a guard at the write would accept or refuse the identical request depending on timing, and in the refusing case would already have rewritten `secrets.env` and mutated the process environment. Refusing up front keeps the request atomic. `docs/operations/managed-config.md` now has a dedicated subsection explaining this, because it is wider than the rest of the mode, plus the deliberate asymmetry that `DELETE /api/providers/{name}/key` is *not* refused (it never touches `config.toml`).
- `set_default_provider` is refused before the OpenRouter / EveryAPI catalog refreshes, so a refused request costs no outbound HTTP. This also matters for correctness: its persist failure is only a `warn!`, so without the guard a managed deployment would answer 200 and hot-switch the live default while the manifest kept saying otherwise.

`PUT /api/providers/{name}/discovery` is deliberately **not** guarded. Its `atomic_write` targets `providers/{name}.toml` under `providers_dir`, a per-provider catalog fragment, not the managed config file. It is now classified explicitly in the doc rather than left unmentioned.

## The documented gap list was wrong — corrected against the code

The plan for this change asserted that `routes/memory.rs`, `src/server.rs` and `routes/skills/mod.rs` do not write `config.toml`, and that the Known-gaps section naming them should be deleted. **That assertion is false.** It came from grepping only for `crate::atomic_write`; all three write the config file through other calls:

- `routes/memory.rs` — `std::fs::write(&config_path, &new_content)` in `memory_config_patch`, where `config_path` is `home_dir/config.toml`.
- `routes/skills/mod.rs` — `upsert_mcp_server_config` / `remove_mcp_server_config` both end in `std::fs::write(config_path, toml_string)`, reached from `routes/skills/mcp.rs` and `routes/skills/extensions.rs`.
- `src/server.rs` — `change_password` reads `home_dir/config.toml`, inserts `dashboard_user` / `dashboard_pass_hash`, and writes it back.

So the section stays, but it is now precise instead of approximate. It gained a table naming each route with the exact call that performs the write, so the list can be checked rather than trusted, and it picked up two facts the original did not have:

- The MCP routes only write `config.toml` under `mcp_runtime_store = "file"` — which is the default; `"db"` moves that persistence into SQLite.
- `quick_init` (`POST /api/init`) writes the whole file, but only when it does not exist, so a ConfigMap-mounted deployment hits its `already_initialized` early return.

It also fixes an inaccuracy in the original text: the claim that the gap routes "do not take `config_write_lock` either" is not true of the sidecar-channel routes, which acquire it around the rewrite (`routes/channels.rs`).

## Why the other gaps are not closed here

Locking `change_password`, the MCP/extension routes, or `quick_init` is a product decision, not a mechanical extension of the guard: refusing `change-password` means a managed deployment can only rotate the dashboard password through a rollout, and refusing the MCP and extension routes removes the dashboard's whole install surface. #6717 listed them as gaps "and honestly so", which reads as deliberate. The three provider routes are unambiguous by contrast — `[default_model]`, `[provider_urls]`, `[provider_proxy_urls]` are deployment configuration under any reading. The trade-offs on the rest are stated in the doc for a maintainer to rule on.

## Maintainer question: `config_write_lock` on the provider routes

Not changed in this PR, stated for the record. `grep config_write_lock crates/librefang-api/src/routes/providers.rs` returns nothing, while the guarded siblings take it: `budget.rs:591-596` and `config/manage.rs:1356-1359` both do guard-then-lock. So the three provider handlers can race a concurrent `POST /api/config/set` in a read-modify-write on `config.toml` — a real pre-existing bug, independent of managed mode. Adding the lock changes concurrency semantics on paths this PR otherwise only adds an early return to, so it is raised rather than silently fixed.

No `423` was added to the `utoipa` `responses(...)` annotations, matching #6717: `grep -n 423` across `config/manage.rs`, `budget.rs` and `users.rs` finds no status annotation, and `openapi.json` contains zero `"423"`. Annotating only the provider routes would be inconsistent in the other direction, and it would move `openapi.json` for a status the rest of managed mode leaves undocumented. Worth doing uniformly in a maintainer-directed pass.

## Verification

Ran in the worktree with native cargo on macOS (arm64):

- `cargo check --workspace --lib` — finished clean in 43m 37s, no errors.
- `cargo clippy --workspace --all-targets -- -D warnings` — `Finished` with **0 warnings** and exit code 0 (`grep -c '^warning\|^error'` over the captured log returns 0).
- `cargo test -p librefang-api --test config_managed_mode_test` — **6 passed; 0 failed**, exit code 0. The three new cases plus the three #6717 cases.
- `cargo test -p librefang-api` — lib suite **925 passed; 0 failed**. The integration-test binaries were still executing when this PR was opened; at last check **990 tests across 8 of ~90 binaries had reported, 0 failed** — see the note below.

Three new `#[tokio::test]` cases in `crates/librefang-api/tests/config_managed_mode_test.rs`, each mirroring the `config_set_is_locked_in_managed_mode_and_leaves_the_file_untouched` pattern (seed the file, assert 423 + `code == "config_managed"` + non-empty `source`, then assert the bytes are unchanged):

- `set_provider_key_is_locked_in_managed_mode` — also seeds and asserts `secrets.env` is byte-identical, which is the widened refusal described above.
- `set_provider_url_is_locked_in_managed_mode` — also asserts the live model catalog's `base_url` did not move, since a refused request must not drift the running daemon.
- `set_default_provider_is_locked_in_managed_mode` — also asserts the effective default model is still `test-model`, so the in-memory override did not hot-switch.

All three run under the existing `ManagedModeGuard` / `managed_mode_lock()` mechanism from #6717 rather than a new one. `LIBREFANG_CONFIG_MODE` is process-global and `config_mode()` reads it on every call, so a second serialization mechanism in the same binary would produce intermittent failures.

Two things I did **not** finish, stated plainly rather than glossed:

- The full `cargo test -p librefang-api` run was still working through its ~90 integration-test binaries when this PR was opened, and I did not wait for it to finish. The lib suite is green (925/925) and every binary that had reported was green (990 tests, 0 failures), but I am **not** claiming the whole crate suite passed. This machine took 43m for `cargo check` alone under heavy load, and CI runs the same suite sharded across 4 runners, so the run there is the authoritative one. The binary that actually covers this change, `config_managed_mode_test`, is green in full.
- `cargo xtask codegen --openapi` was not run. No `utoipa` annotation changed (see above), so `openapi.json` should not move — but that is reasoning, not a verified diff, and the pre-commit hook / `openapi-drift` CI job will settle it.

Nothing in this change is platform-dependent, but there is no Windows machine in this environment, so only the macOS (arm64) host results above are real.

Refs #6695 #6717


crates/librefang-api/src/routes/providers.rs
```diff
@@ -1454,6 +1454,14 @@ pub async fn set_provider_key(
     Path(name): Path<String>,
     Json(body): Json<serde_json::Value>,
 ) -> impl IntoResponse {
+    // Managed mode (#6695) refuses the whole request, not just the config write.
+    // This handler writes `secrets.env` at the top and only *conditionally* reaches `persist_default_model` further down — the auto-switch branches depend on whether the current default already has a working key, which the caller cannot predict.
+    // A guard at the config write would therefore refuse a request that had already rewritten `secrets.env` and mutated the process environment, and would return 200 or 423 for the identical request depending on daemon state.
+    // Refusing first keeps the request atomic and the contract documentable: in a managed deployment provider credentials come from the environment or secret manifest, not from this route.
+    if let Some(locked) = crate::routes::guard_config_write() {
```
houko, line 1461: Re the "Maintainer question" in the PR body: per `CLAUDE.md`'s "fix what you found" rule, the bar for deferring is "would fixing it require touching a different crate or domain than the one you're already in?"
`config_write_lock` is not that here — `set_provider_key`, `set_provider_url` (line 2423), and `set_default_provider` (line 2612) are the exact three functions this PR is already editing, in the file it is already editing, right next to the `guard_config_write()` call each one adds.
The sibling handlers this PR's own body cites, `budget.rs:591-596` and `config/manage.rs:1356-1359`, both do guard-then-lock in that order:
```rust
if crate::routes::guard_config_write().is_some() { return Err(PersistBudgetError::Managed); }
let _guard = state.config_write_lock.lock().await;
```
So by CLAUDE.md's own bar this reads as in-scope for this PR rather than a deferred item, since it's the same file/domain already open.

That said, I'm not making this fix myself: `state.config_write_lock` is a real `tokio::sync::Mutex` guarding a read-modify-write against `config.toml`, and adding it here changes concurrency semantics on a hot path (three provider routes), not a mechanical/textual change — exactly the kind of change CLAUDE.md says needs a maintainer's call rather than a unilateral edit. Concretely the gap: without the lock, a concurrent `POST /api/config/set` and (e.g.) `set_default_provider` can race a read-modify-write on `config.toml`, same class of bug `config_write_lock` exists to close in `budget.rs`/`config/manage.rs`.

Flagging so a maintainer can decide whether to fold the lock into this PR (same file, same handlers, small diff) or explicitly accept it as an out-of-scope follow-up with a linked issue — CLAUDE.md treats "different crate/domain" as the only clean bar for deferring, and this doesn't clear it.

---

houko:
Automated CLAUDE.md convention pass. Changelog fragment, prose wrapping, guard placement, and integration-test coverage all checked out clean (the three new `config_managed_mode_test.rs` cases genuinely exercise the routes through a real router, not just the guard function directly). One finding, left as a comment rather than fixed since it requires rewriting this branch's history:

Commit `4c87eb0` ("fix(api): rewrap new test comments to one sentence per line") has git author identity `Claude <noreply@anthropic.com>`. CLAUDE.md's `commit-msg` hook section states this is rejected even when the commit message text itself is clean — please re-author (`git commit --amend --author`) before merge.

---

crates/librefang-api/tests/config_managed_mode_test.rs
```diff
@@ -206,24 +235,151 @@ async fn config_set_is_locked_in_managed_mode_and_leaves_the_file_untouched() {
...
+#[tokio::test(flavor = "multi_thread")]
+async fn set_provider_key_is_locked_in_managed_mode() {
+    let _guard = ManagedModeGuard::set().await;
+
+    let h = boot_router_with_api_key(API_KEY).await;
+    let config_path = h.home.join("config.toml");
+    let config_seed = format!("api_key = \"{API_KEY}\"\n");
+    std::fs::write(&config_path, &config_seed).expect("seed config.toml");
+
+    // Seed `secrets.env` too.
```
houko, line 266: Commit `4c87eb017a8c08f19ab411fdb749f2e32ca1f9e8` ("fix(api): rewrap new test comments to one sentence per line"), which touches this file, is authored as `Claude <noreply@anthropic.com>` (`git log --format='%an <%ae>'` on this branch confirms it).

CLAUDE.md is explicit that this is not allowed: "No AI / Claude attribution in commit messages, PR bodies, or comments," and separately describes the `commit-msg` hook as rejecting "a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean." This commit's author identity is exactly that case — it reached `origin/fix/managed-config-provider-guard` regardless, so whatever normally blocks this locally didn't fire here.

Separately, and a little ironically given what the commit does: its own message body hard-wraps mid-sentence —
```
CLAUDE.md's prose-wrapping rule requires new prose to break only at
sentence boundaries, not at a fixed column.
```
— which is itself a violation of the "Prose wrapping: no column limit; break only at sentence boundaries" rule ("Commit message bodies" are explicitly listed as covered).

I'm not able to fix this myself: removing the offending commit means rewriting history on this branch and force-pushing, which is off-limits for me. Flagging for a maintainer to decide — e.g. rebase it out / re-author before merge, or squash on merge if GitHub's squash-merge is used (that would attribute the squashed commit to the merger, not to `Claude`, but the intermediate history and the PR's commit list would still show it).

---

crates/librefang-api/src/routes/providers.rs
```diff
@@ -2592,6 +2606,13 @@ pub async fn set_default_provider(
     Path(name): Path<String>,
     body: Option<axum::Json<serde_json::Value>>,
 ) -> impl IntoResponse {
+    // Managed mode (#6695): the whole point of this route is to persist `[default_model]` into config.toml (`persist_default_model` below), which a managed deployment owns.
+    // The guard runs before the catalog refreshes because those issue outbound HTTP to OpenRouter / EveryAPI — a refused request should cost nothing, and refreshing a catalog we are about to refuse to act on is pure waste.
+    // Refusing here also keeps the in-memory `default_model_override` aligned with the file: the persist failure below is only a `warn!`, so without the guard a managed deployment would answer 200 and hot-switch the live default while the manifest kept saying otherwise — the exact silent drift managed mode exists to prevent.
+    if let Some(locked) = crate::routes::guard_config_write() {
```
houko, line 2612: The PR body raises, and explicitly declines to fix, that `set_provider_key` / `set_provider_url` / `set_default_provider` (this guard and its two siblings at lines 1461 and 2423) don't take `config_write_lock`, unlike the sibling guarded routes it names (`budget.rs:591-596`, `config/manage.rs:1356-1359`), so they can race a concurrent `POST /api/config/set` in a read-modify-write on `config.toml`.

CLAUDE.md's "Fix what you found" rule sets the bar for deferring at "would fixing it require touching a different crate or domain than the one you're already in?" — here the answer is no: it's the same file, the same three handlers this PR is already editing, and the pattern to follow (guard-then-lock) already exists twice elsewhere in the same crate. Per that rule, this reads like it should be fixed in this PR rather than raised as a "maintainer question," unless there's a reason the guard-then-lock pattern doesn't transplant cleanly onto these three handlers specifically.

I'm leaving this as a comment rather than fixing it myself: the PR body notes "adding the lock changes concurrency semantics on paths this PR otherwise only adds an early return to," and I haven't verified that acquiring the lock around `persist_default_model` / `upsert_provider_url` / `upsert_provider_proxy_url` is deadlock-free or behaviorally equivalent to the `budget.rs` / `config/manage.rs` call sites — that's a correctness claim I'd want a maintainer's sign-off on rather than asserting from a diff read.

---

houko (commented):
Reviewed the guard placement, test coverage, and doc changes against CLAUDE.md. The core fix (guard_config_write on the three provider routes, using the existing #6717 mechanism rather than anything ad hoc) is correct and the coverage is complete for config.toml writes in routes/providers.rs — checked the other handlers in that file (aliases, custom models, model overrides, discovery, enable/delete-key) and none of them write config.toml, so the guard placement isn't missing a sibling route. Tests and docs are solid. Two things worth a maintainer's attention before merge, left as inline comments rather than fixed directly since both involve a git-history / concurrency-semantics judgment call I shouldn't make unilaterally.

---

crates/librefang-api/tests/config_managed_mode_test.rs
```diff
@@ -206,24 +235,151 @@ async fn config_set_is_locked_in_managed_mode_and_leaves_the_file_untouched() {
...
-    assert_eq!(v["code"], "config_managed");
-    assert_eq!(v["ok"], false);
-    assert!(
-        v["source"].as_str().is_some_and(|s| !s.is_empty()),
-        "the refusal must tell the operator which file is managed; got {v}"
+// ---------------------------------------------------------------------------
+// Provider routes that persist deployment configuration
+// ---------------------------------------------------------------------------
+//
+// `set_provider_key` persists `[default_model]` (auto-switch / free-model migration), `set_provider_url` persists `[provider_urls]` and `[provider_proxy_urls]`, and `set_default_provider` persists `[default_model]` outright.
```
houko, line 251: Not a comment on this line's content — flagging a commit-history issue that touches this file.

Commit `4c87eb0` ("fix(api): rewrap new test comments to one sentence per line", which rewrapped the comment block starting here) has author identity `Claude <noreply@anthropic.com>`.

CLAUDE.md's git conventions say "No AI / Claude attribution in commit messages, PR bodies, or comments," and the repo's `commit-msg` hook is specifically described as rejecting "a commit whose author identity (`git var GIT_AUTHOR_IDENT`...) resolves to Claude / Anthropic even when the message itself is clean." That's exactly this commit's situation — clean message, AI-attributed author.

Suggested fix: rebase this branch with `git commit --amend --reset-author` (or `git rebase -i` + `--author`) on commit `4c87eb0` to set a real human author, then force-push. I'm not doing that myself since it requires rewriting already-pushed history, which is outside what I should do unilaterally here — leaving it for a maintainer to decide (squash-merge would also sidestep it, if that's the intended merge strategy).

---

crates/librefang-api/src/routes/providers.rs
houko, line 1461: Flagging for a maintainer decision rather than blocking on it — the PR body already raises this explicitly under "Maintainer question," so this is just pinning it to the code.

None of the three newly-guarded handlers (`set_provider_key` here, `set_provider_url`, `set_default_provider`) take `state.config_write_lock` around their `config.toml` writes, while the sibling guarded routes do (`budget.rs:591-596`, `config/manage.rs:1356-1359` both guard-then-lock). That means a concurrent `POST /api/config/set` can still race a read-modify-write here.

Per CLAUDE.md's "fix what you found" rule, the bar for deferring is whether the fix needs a different crate/domain — this is the same file the PR is already editing, which leans toward "fix it here" rather than defer. That said, I agree with the PR author that adding the lock changes concurrency semantics beyond a mechanical guard addition (unlike the other deferred gaps in memory.rs/skills/mod.rs/server.rs, which really are a different domain), so this reads as a legitimate case for explicit maintainer sign-off rather than something to fix unilaterally in this pass. Recommend the maintainer either approve adding `config_write_lock` here before merge, or explicitly accept the pre-existing race as out of scope so it doesn't quietly slip through.

---

crates/librefang-api/tests/config_managed_mode_test.rs
houko, line 1: Commit `4c87eb017a8c08f19ab411fdb749f2e32ca1f9e8` ("fix(api): rewrap new test comments to one sentence per line"), which touches this file, is authored as `Claude <noreply@anthropic.com>`.

CLAUDE.md's `commit-msg` hook description is explicit that author *identity* resolving to Claude/Anthropic is rejected independently of message content ("rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean"), and the "No AI / Claude attribution" rule under Git Conventions / Commit & PR hygiene applies to the whole commit, not just its message text.

This commit is already on `fix/managed-config-provider-guard` and its message body is otherwise clean, so the violation is the author identity alone. Flagging rather than fixing directly: correcting it means rewriting an already-pushed commit (amend + force-push, or squash), which the review policy for this task does not allow me to do unilaterally. Recommend the author re-commit this change under their own identity (e.g. `git commit --amend --reset-author` or an interactive rebase) before merge.

---

docs/operations/managed-config.md
houko, line 1: Commit `d2bc8c6c353203145bd6530bbc9051d1f0b93cf2` ("docs(managed-config): clarify mcp_runtime_store escape does not cover extension routes"), which touches this file, hard-wraps its message body at ~72 columns mid-sentence:

```
The known-gaps table groups the extension install/uninstall routes with
the direct MCP-server routes under the same mcp_runtime_store qualifier,
but routes/skills/extensions.rs calls upsert_mcp_server_config and
remove_mcp_server_config unconditionally, with no mcp_runtime_store
check, so setting mcp_runtime_store = "db" does not stop
install_extension/uninstall_extension from writing config.toml.
```

CLAUDE.md's prose-wrapping rule explicitly lists "Commit message bodies" as governed by "one sentence = one line, regardless of length" (only the subject line gets git's own ~72-char truncation convention as an exception). This is a single sentence split across six lines at a fixed column, which is exactly the pattern the rule forbids — and notably, an earlier commit on this same branch (`4c87eb01`) exists specifically to fix this same violation in code comments, so the convention was known within the PR but not applied here.

Flagging rather than fixing directly: this commit is already pushed to `fix/managed-config-provider-guard`, so correcting the message requires amend + force-push, which I'm not authorized to do unilaterally. Recommend squashing or amending before merge.

---

houko (commented):
Reviewed against CLAUDE.md. The functional change itself (three `guard_config_write()` calls in `crates/librefang-api/src/routes/providers.rs`, plus new `#[tokio::test]` coverage in `config_managed_mode_test.rs` for all three routes and an updated `docs/operations/managed-config.md`) is mechanically sound: `guard_config_write` is a pre-existing helper, the placement rationale for each of the three guards checks out against the code paths it guards, the new tests exercise `TestServer` per the MANDATORY Integration Testing section and assert both file and in-memory-catalog state, and the accessor methods the tests call (`config_ref`, `default_model_override_ref`, `model_catalog_ref`) exist on the kernel trait. No `is_public`/auth-allowlist change was needed (no new public routes), no config-struct fields were added (Default-impl rule n/a), and no HashMap/HashSet was introduced in a prompt-ordering-sensitive path.

Two findings, both in already-pushed commit metadata rather than in file content, so left as comments rather than fixed in-place (fixing either requires amending an already-pushed commit / force-push, which I'm not authorized to do unilaterally):

1. Commit `4c87eb017a8c08f19ab411fdb749f2e32ca1f9e8` is authored as `Claude <noreply@anthropic.com>` — violates the "No AI / Claude attribution" rule at the author-identity level, per CLAUDE.md's own description of what the `commit-msg` hook checks.
2. Commit `d2bc8c6c353203145bd6530bbc9051d1f0b93cf2`'s message body hard-wraps a single sentence across six ~72-column lines, violating the prose-wrapping rule's explicit inclusion of commit message bodies.

Everything else checked out: commit conventions (type prefixes, no Co-Authored-By), changelog fragment format and placement (`changelog.d/fixed/6737-managed-config-provider-routes.md`), and prose-wrap compliance in the actual file diffs (code comments, doc comments, and the `docs/operations/managed-config.md` prose additions are all one-sentence-per-line).

---

changelog.d/fixed/6737-managed-config-provider-routes.md
houko, line 1: Not a finding about this file's content (the fragment itself is correctly formatted: right directory, no leading `- `, ends `(#6737) (@houko)`) — flagging a repo-hygiene issue elsewhere in this PR's history that's easiest to anchor here.

Commit `4c87eb0` ("fix(api): rewrap new test comments to one sentence per line") in this branch has author identity `Claude <noreply@anthropic.com>`. CLAUDE.md bans AI/Claude attribution in commits, and per its description of `scripts/hooks/commit-msg`, that hook specifically rejects a commit whose *author identity* resolves to Claude/Anthropic even when the message text is clean — which is this commit's exact shape. It isn't the branch HEAD (that's `d2bc8c6`, authored by `houko`), but it's still in the history that would land on `main`. Flagging for the author/maintainer to squash or re-author before merge, since fixing it requires rewriting this branch's history rather than something I can push as a new commit.

---

houko (commented):
Automated daily review pass. Checked the guard placement against `guard_config_write` call sites, and cross-checked every write-site claim in the rewritten Known-gaps table (`routes/memory.rs::memory_config_patch`, `routes/skills/mod.rs::upsert_mcp_server_config`/`remove_mcp_server_config` reached from both `mcp.rs` and `extensions.rs`, `routes/sidecar_toml.rs`, `server.rs::change_password`, `routes/config/system.rs::quick_init`, plus the `mcp_runtime_store` and `config_write_lock` claims) against the actual code — all accurate. No auth-allowlist / route-reordering concerns since this doesn't touch `middleware.rs` or unauthenticated-route registration, and it doesn't touch config hot-reload classification so `docs/operations/config-reload.md` isn't implicated. Changelog fragment is correctly placed and formatted. No functional findings on the code itself.

One repo-hygiene finding, noted inline: a non-HEAD commit in this branch (`4c87eb0`) has AI/Claude author attribution, which CLAUDE.md's commit-msg hook is specifically designed to reject even when the message text is clean.

---

crates/librefang-api/tests/config_managed_mode_test.rs
houko, line 1: Commit `4c87eb0` ("fix(api): rewrap new test comments to one sentence per line") in this branch has its author *and* committer identity set to `Claude <noreply@anthropic.com>`.
That is exactly the pattern `scripts/hooks/commit-msg` is written to reject — it checks `git var GIT_AUTHOR_IDENT` specifically so a clean commit message can't carry AI attribution through the author field instead of the text — and CLAUDE.md's "Commit & PR hygiene" section states the same rule for the message body.
The commit's content is fine (it's a legitimate prose-wrap fix per CLAUDE.md's own rule), so this is purely an authorship-metadata issue, not a content one.

Fixing it means rewriting history on this branch (`commit --amend --reset-author` or an interactive rebase) and force-pushing, which isn't something to do unilaterally on someone else's PR — leaving this as a comment for @houko or a maintainer to resolve (e.g. amend-and-force-push, or squash-merge so the merge commit's authorship isn't at issue).

---

houko (commented):
Reviewed the diff against `origin/main`. The three managed-mode guard additions (`set_provider_key`, `set_provider_url`, `set_default_provider`) reuse the existing `guard_config_write()` helper already proven at three other call sites, placed correctly before any mutation (secrets.env write, in-memory catalog update, outbound catalog refresh), and match its `(StatusCode, Json<Value>)` return shape exactly. Mode detection (`LIBREFANG_CONFIG_MODE`) is read server-side from the process environment only — not client-controllable via header/param. The "documented gap list" correction in `docs/operations/managed-config.md` is prose-only; I checked every named route/write-site claim (memory.rs, skills/mod.rs + extensions.rs, sidecar_toml.rs, server.rs::change_password, config/system.rs::quick_init, the discovery route, and the sidecar `config_write_lock` usage) against the actual code and all check out — no route registration, OpenAPI, or dead-route-audit drift. New tests are thorough and assert both the file and in-memory-state invariants. Changelog fragment is present and correctly formatted.

One finding, left as a comment rather than fixed: commit `4c87eb0` has both author and committer identity set to `Claude <noreply@anthropic.com>`, which is the exact pattern the repo's `commit-msg` hook exists to reject even when the message text is clean. Left for the branch owner to amend, since fixing it means rewriting and force-pushing someone else's history.

---

PR #6505, opened 2026-07-19
feat(goals): autonomous goal runner with loop engineering, /goal CLI, TUI, and channels

DaBlitzStein:
Complete autonomous goal execution system with verifier loop, self-evolution memory, disposable worker provisioning, /goal CLI command, TUI screen, dashboard agent selector, and channel slash command.

## Goal Runner (kernel)
- Autonomous loop: agent → parse GOAL_PROGRESS/GOAL_DONE → iterate
- Verifier loop: separate agent judges PASS/FAIL/NEEDS_REWORK
- Auto-spawn reviewer after 2 verifier rejections
- Evaluator model: separate configurable model for completion judgment
- GOAL_LEARNED markers → accumulated across iterations → injected into future prompts
- Auto-skills on goal completion via create_skill()
- Auto-resume stale runs on daemon restart
- loop_engineering toggle: off = upstream behavior, on = all of the above

## API
- Auto-spawn worker agent when no agent_id assigned
- Auto-spawn verifier agent when loop_engineering on and no reviewer assigned
- evaluator_model field: per-goal configurable model for completion judgment

## /goal CLI
librefang goal "Fix the login bug"
librefang goal "Research" --loop-engineering --watch

## TUI Goals Screen
- List/create/edit/start/stop/delete goals
- 6-step create form with human labels and info hints
- Agent/Auto-review/Reviewer/Goal Judge fields
- Search filter, detail view

## Dashboard
- Agent selector in create/edit forms
- Free-form create visible with 0 goals
- Human-friendly i18n labels

## Channels
- /goal command in Telegram/Slack

## Tests
14 kernel + 12 types + 26 channels = 52 goal tests pass. Dashboard typecheck green.

PR #5033, opened 2026-05-14
feat(types,kernel,runtime): async task tracker — kernel registry + event injection + wake-idle (#4983)

houko:
Closes #4983

## Summary

Implements #4983 in full — kernel-level async task tracker that lets agents launch `workflow_start` (and, later, `agent_send_async` / external-webhook) operations without blocking their conversation loop, and that delivers the eventual result back into the originating session via synthetic event injection (mid-turn or wake-idle).

Replaces the production failure narrative from #4983 (Telegram-assistant `ltdata` agent blocked for the full duration of a multi-minute `workflow_run`, user follow-ups ignored, no `run_id` surfaced on timeout, $2/day orchestrator-polling workaround) with: `workflow_start` returns immediately; agent keeps serving the user; kernel injects the result whenever the workflow finishes; agent picks it up on its next turn or wakes from idle.

### Types (originally step 1)

- New module `crates/librefang-types/src/task.rs`:
  - `TaskId(pub Uuid)` — kernel-assigned identifier (`new()` / `Default` / `Display` / `FromStr`).
  - `WorkflowRunId(pub Uuid)` — colocated here since `librefang-types` is at the bottom of the dep DAG. Kernel's previous `workflow::WorkflowRunId` is now `pub use librefang_types::task::WorkflowRunId`; one canonical type.
  - `TaskKind` — `#[serde(tag = "kind", rename_all = "snake_case")]` with `Workflow { run_id }` and `Delegation { agent_id, prompt_hash }`. Externally tagged so additive variants (`ExternalWebhook`, `LongRunningTool`, …) do not break wire-compat with already-registered handles.
  - `TaskHandle { id, kind, started_at }` — typed handle returned synchronously to spawning agents, holdable across turns.
  - `TaskStatus` — `#[serde(tag = "status", content = "value")]` with `Pending` / `Running` / `Completed(serde_json::Value)` / `Failed(String)` / `Cancelled`.
  - `TaskCompletionEvent { handle, status, completed_at }` — wire payload injected into the originating session.

### Kernel substrate (originally step 2)

- New file `crates/librefang-kernel/src/kernel/task_registry.rs` — `register_async_task(agent_id, session_id, kind) -> TaskHandle` and `complete_async_task(task_id, status) -> KernelResult<bool>` on `LibreFangKernel`. `complete_async_task` is **idempotent** — second call for the same `TaskId` no-ops and returns `Ok(false)`, guarding against retry races between the workflow engine and any future supervisor.
[...]

PR #6740, opened 2026-08-07
feat(runtime): make agent_send non-blocking by default when a caller is known

houko:
## Why

`agent_send` defaulted to blocking, which made every unannotated delegation a timeout risk. The model had to predict *in advance* that the callee would be slow and opt in to `"async": true`, and a wrong guess spent the whole turn waiting until `tool_timeout_secs` fired. The asymmetry runs one way: an unnecessary `task_id` costs one extra turn to collect, while an unnecessary block can lose the turn entirely.

This was requested as its own PR on 2026-07-19. It is also the shared root of the red `Test / Unit (lib+bin)` lane on both #6504 and #6505 — both carry the same one-line flip against four untouched test expectations. Landing it here lets both rebase and drop it, taking #6504 from four red checks to two.

## The flip is conditional, deliberately

The obvious change is `unwrap_or(false)` → `unwrap_or(true)` at `crates/librefang-runtime/src/tool_runner/agent.rs:60`. That is wrong, and it is what both contributor branches do.

The async path requires a known caller — `caller_agent_id.ok_or(ToolError::InvalidParameter { name: "async", … })` at `agent.rs:110-113` — because the tracker has nowhere to route the completion without one. The blocking path, by contrast, has explicit arms for callerless sends: `(None, Some(key))` and `(None, None)` at `agent.rs:154-160`, with a comment stating that system-initiated calls fall back to the legacy path.

So an unconditional `true` converts **every callerless `agent_send` into an `InvalidParameter` error** — including the kernel's own system-initiated dispatches. The existing test `agent_send_async_without_caller_is_invalid_parameter` already pins that error for the explicit case, which is direct evidence the unconditional form breaks the callerless one.

The default is therefore `unwrap_or_else(|| caller_agent_id.is_some())`:

| `"async"` | caller | behaviour | changed? |
| --- | --- | --- | --- |
| omitted | known | non-blocking, returns `task_id` | **yes — this is the fix** |
| omitted | none | blocking | no |
| `false` | any | blocking | no |
| `true` | known | non-blocking | no |
| `true` | none | `InvalidParameter` | no (pre-existing contract) |

## Changes

**`crates/librefang-runtime/src/tool_runner/agent.rs`** — the conditional default, with a comment recording why it is not unconditional so the next reader does not "simplify" it.

**`crates/librefang-runtime/src/tool_runner/definitions.rs`** — the tool description said "By default this BLOCKS" and the `async` property said "Defaults to false (blocking)". Both now describe the non-blocking default and tell the model to pass `"async": false` for a quick sub-question it needs answered inside the turn. This is what the LLM actually reads, so leaving it stale would be the whole bug in a different place.

**`crates/librefang-runtime/src/tool_runner/tests/mod.rs`**

Three tests that cover caller-aware and keyed *blocking* routing now pass `"async": false` explicitly, rather than having their expectations bumped to the tracker path:

- `agent_send_no_key_with_caller_routes_to_send_to_agent_as`
- `agent_send_same_key_routes_to_as_with_key_both_calls`
- `agent_send_distinct_keys_produce_isolated_dispatch`

Bumping them instead would have deleted the only coverage of `send_to_agent_as` / `send_to_agent_as_with_key` — dispatch arms that still exist and are still reachable via `"async": false`. Moving a test's expectation to the new behaviour is not the same as keeping the old behaviour tested.

`agent_send_no_key_no_caller_routes_to_send_to_agent` is untouched: with no caller the default is still blocking, so its expectation is correct as written.

Two new tests:

- `agent_send_defaults_to_async_when_caller_is_known` — no `"async"` key at all (what a model emits when it does not think about blocking) with a caller and session must reach `async_tracked` and return a `task_id`.
- `agent_send_without_caller_still_defaults_to_blocking` — the guard against collapsing the conditional into `unwrap_or(true)`. A callerless send must stay on `send_to_agent` and must not be rejected.

**`docs/src/app/agent/tools/page.mdx`** and its `zh/` mirror — the tools table said `agent_send` "and wait for the reply" / 「并等回复」, which is no longer true. Both now state the non-blocking default and the `async: false` opt-out.

## Verification

Run natively on macOS aarch64 in the worktree, based on `origin/main` at c403439a.

- **`cargo test -p librefang-runtime --lib tool_runner::tests::agent_send`** — 10 passed, 0 failed. Includes both new tests and the three modified ones, plus the pre-existing `agent_send_async_without_caller_is_invalid_parameter` and `agent_send_async_without_session_returns_the_reply_not_a_fake_task_id_6650`, which still pass unchanged.
- **`cargo check --workspace --lib`** — clean, `Finished` in 8m 34s.
- **`cargo clippy --workspace --all-targets -- -D warnings`** — exit code **0**, captured explicitly rather than inferred from the `Finished` line.
- **`cargo test -p librefang-runtime --lib`** (full crate) — 2175 passed, 0 failed, 2 ignored. See the flake note below.
- **Breadth scan** for other dependencies on the blocking default: `grep` for `tool_agent_send` across `crates/` finds only the definition, the `dispatch.rs:964` call site (which threads `caller_agent_id` through unchanged) and comment references in `librefang-kernel`. No copy of the old `"Defaults to false (blocking)"` string survives anywhere in the tree. No JSON schema fixture or SDK snapshot carries the `agent_send` tool description.

### Unrelated flake observed, not fixed here

Across four full-crate runs, one test failed twice:

```
---- plugin_runtime::tests::overflow_kills_child_without_waiting_for_timeout stdout ----
panicked at crates/librefang-runtime/src/plugin_runtime/tests.rs:412:5:
kill took 5.670414042s, expected <5s (timeout was 30s) — child likely deadlocked
```

It is a wall-clock assertion with a hard 5s threshold, and it failed on the runs that followed a full `clippy --all-targets` compile — i.e. under machine load, not because the child deadlocked. Runs 1 and 2 of a back-to-back set of three passed; run 3 failed. It is in `plugin_runtime`, touches nothing in this diff, and is a pre-existing condition on `main`.

Not fixed here because it is a different module from the one this PR changes and it deserves its own change (condition polling instead of a load-sensitive deadline) rather than riding along inside a semantics change. Flagging it rather than staying quiet: it will randomly red other PRs.

### Behaviour change reviewers should weigh

Every agent that calls `agent_send` without specifying `async` now gets a `task_id` instead of the callee's reply, and must collect the reply from its session on a later turn instead of using it inside the same turn. Prompts or skills that assume the reply arrives inline will need `"async": false`. The tool description now says so, so a model reading the current schema will adapt; anything hardcoding the old assumption will not.

Refs #6504 #6505

### DIFF
--- a/crates/librefang-runtime/src/tool_runner/agent.rs
+++ b/crates/librefang-runtime/src/tool_runner/agent.rs
@@ -58,7 +58,7 @@ pub(super) async fn tool_agent_send(
     // blocking this agent's loop until the callee replies (which otherwise
     // trips `tool_timeout_secs` for any long delegation).
     //
-    // Non-blocking is the DEFAULT when a caller agent is known, because the blocking default made every unannotated delegation a timeout risk: the model had to predict that the callee would be slow and opt out in advance, and a wrong guess burned the turn on `tool_timeout_secs`.
+    // Non-blocking is the DEFAULT when a caller agent is known, because the blocking default made every unannotated delegation a timeout risk: the model had to predict that the callee would be slow and opt in to async in advance, and a wrong guess burned the turn on `tool_timeout_secs`.
     //
     // The default is deliberately conditional on `caller_agent_id` rather than an unconditional `true`.
     // The async path below requires a known caller so the tracker can route the completion back, and rejects the call outright without one; the blocking path has explicit `(None, _)` arms for system-initiated sends.
--- a/crates/librefang-runtime/src/tool_runner/agent.rs
+++ b/crates/librefang-runtime/src/tool_runner/agent.rs
@@ -57,7 +57,24 @@ pub(super) async fn tool_agent_send(
     // kernel async-task tracker and return a task id immediately instead of
     // blocking this agent's loop until the callee replies (which otherwise
     // trips `tool_timeout_secs` for any long delegation).
-    let async_mode = input["async"].as_bool().unwrap_or(false);
+    //
+    // Non-blocking is the DEFAULT when a caller agent is known, because the
+    // blocking default made every unannotated delegation a timeout risk: the
+    // model had to predict that the callee would be slow and opt out in
+    // advance, and a wrong guess burned the turn on `tool_timeout_secs`.
+    //
+    // The default is deliberately conditional on `caller_agent_id` rather than
+    // an unconditional `true`. The async path below requires a known caller so
+    // the tracker can route the completion back, and rejects the call outright
+    // without one; the blocking path has explicit `(None, _)` arms for
+    // system-initiated sends. An unconditional default would therefore turn
+    // every callerless send — the kernel's own system-initiated dispatches —
+    // into an `InvalidParameter` error. An explicit `"async": true` still
+    // errors without a caller, which is the pre-existing contract and is left
+    // untouched.
+    let async_mode = input["async"]
+        .as_bool()
+        .unwrap_or_else(|| caller_agent_id.is_some());
 
     if let Some(caller) = caller_agent_id {
         if caller == agent_id {
--- a/crates/librefang-runtime/src/tool_runner/agent.rs
+++ b/crates/librefang-runtime/src/tool_runner/agent.rs
@@ -58,20 +58,12 @@ pub(super) async fn tool_agent_send(
     // blocking this agent's loop until the callee replies (which otherwise
     // trips `tool_timeout_secs` for any long delegation).
     //
-    // Non-blocking is the DEFAULT when a caller agent is known, because the
-    // blocking default made every unannotated delegation a timeout risk: the
-    // model had to predict that the callee would be slow and opt out in
-    // advance, and a wrong guess burned the turn on `tool_timeout_secs`.
+    // Non-blocking is the DEFAULT when a caller agent is known, because the blocking default made every unannotated delegation a timeout risk: the model had to predict that the callee would be slow and opt out in advance, and a wrong guess burned the turn on `tool_timeout_secs`.
     //
-    // The default is deliberately conditional on `caller_agent_id` rather than
-    // an unconditional `true`. The async path below requires a known caller so
-    // the tracker can route the completion back, and rejects the call outright
-    // without one; the blocking path has explicit `(None, _)` arms for
-    // system-initiated sends. An unconditional default would therefore turn
-    // every callerless send — the kernel's own system-initiated dispatches —
-    // into an `InvalidParameter` error. An explicit `"async": true` still
-    // errors without a caller, which is the pre-existing contract and is left
-    // untouched.
+    // The default is deliberately conditional on `caller_agent_id` rather than an unconditional `true`.
+    // The async path below requires a known caller so the tracker can route the completion back, and rejects the call outright without one; the blocking path has explicit `(None, _)` arms for system-initiated sends.
+    // An unconditional default would therefore turn every callerless send — the kernel's own system-initiated dispatches — into an `InvalidParameter` error.
+    // An explicit `"async": true` still errors without a caller, which is the pre-existing contract and is left untouched.
     let async_mode = input["async"]
         .as_bool()
         .unwrap_or_else(|| caller_agent_id.is_some());
--- a/crates/librefang-runtime/src/tool_runner/definitions.rs
+++ b/crates/librefang-runtime/src/tool_runner/definitions.rs
@@ -320,7 +320,7 @@ use instead of web_fetch + file_write (which round-trips the entire body through
             },
             ToolDefinition {
                 name: tool_name::AGENT_SEND.to_string(),
-                description: "Send a message to another agent. By default this BLOCKS until the agent replies and returns their response — only use the blocking mode for quick sub-questions whose answer you need within this turn. For any delegation that may take a while (research, multi-step work), set \"async\": true so you are not blocked and don't hit the tool timeout. Accepts UUID or agent name. Use agent_find first to discover agents.".to_string(),
+                description: "Send a message to another agent. By default this is NON-BLOCKING: you get back a task_id immediately and the target's reply is delivered to your session when it finishes, so a long delegation cannot burn your turn on the tool timeout. Set \"async\": false for a quick sub-question whose answer you need within this same turn — that blocks until the agent replies and returns their response directly. Accepts UUID or agent name. Use agent_find first to discover agents.".to_string(),
                 input_schema: serde_json::json!({
                     "type": "object",
                     "properties": {
@@ -332,7 +332,7 @@ use instead of web_fetch + file_write (which round-trips the entire body through
                         },
                         "async": {
                             "type": "boolean",
-                            "description": "When true, requests non-blocking delegation. On success you get back a task_id and the target agent's response is delivered to your session automatically when it finishes, so you can continue or end your turn. Async tracking needs a live caller session: on surfaces that have none (the MCP HTTP bridge, the REST tool endpoint) the delegation runs inline and you get the target's reply itself rather than a task_id — read what you receive rather than assuming a task is pending. Use this for any delegation that might take longer than a few seconds (it avoids the tool-execution timeout). Defaults to false (blocking)."
+                            "description": "Controls blocking. Omit it to get the default non-blocking delegation: you receive a task_id and the target agent's response is delivered to your session automatically when it finishes, so you can continue or end your turn. Set it to false to block until the target replies and get their response directly — appropriate only for a quick sub-question you need answered inside this turn. Async tracking needs a live caller session: on surfaces that have none (the MCP HTTP bridge, the REST tool endpoint) the delegation runs inline and you get the target's reply itself rather than a task_id — read what you receive rather than assuming a task is pending."
                         }
                     },
                     "required": ["agent_id", "message"]
--- a/changelog.d/changed/6740-agent-send-async-default.md
+++ b/changelog.d/changed/6740-agent-send-async-default.md
@@ -2,4 +2,4 @@
   The previous blocking default required the model to predict in advance that a delegation would be slow and opt in to `"async": true`; a wrong guess spent the entire turn waiting for `tool_timeout_secs`.
   An unnecessary `task_id` costs one extra turn to collect, whereas an unnecessary block can lose the turn outright.
   Pass `"async": false` for a quick sub-question whose answer is needed within the same turn.
-  Callerless system-initiated sends keep dispatching synchronously, because the async tracker has no session to route a completion back to. (#6740) (@houko)
+  Callerless system-initiated sends keep dispatching synchronously, because the async tracker requires a known caller agent to route a completion back to. (#6740) (@houko)
--- a/changelog.d/changed/6740-agent-send-async-default.md
+++ b/changelog.d/changed/6740-agent-send-async-default.md
@@ -1,0 +1,5 @@
+`agent_send` now delegates non-blockingly by default when the calling agent is known, returning a `task_id` whose reply is delivered to the caller's session on completion.
+  The previous blocking default required the model to predict in advance that a delegation would be slow and opt in to `"async": true`; a wrong guess spent the entire turn waiting for `tool_timeout_secs`.
+  An unnecessary `task_id` costs one extra turn to collect, whereas an unnecessary block can lose the turn outright.
+  Pass `"async": false` for a quick sub-question whose answer is needed within the same turn.
+  Callerless system-initiated sends keep dispatching synchronously, because the async tracker has no session to route a completion back to. (#6740) (@houko)
--- a/crates/librefang-runtime/src/tool_runner/tests/mod.rs
+++ b/crates/librefang-runtime/src/tool_runner/tests/mod.rs
@@ -1161,10 +1161,8 @@ async fn agent_send_no_key_no_caller_routes_to_send_to_agent() {
 async fn agent_send_no_key_with_caller_routes_to_send_to_agent_as() {
     let cap = Arc::new(DispatchCapture::default());
     let kernel: Arc<dyn KernelHandle> = cap.clone();
-    // `"async": false` is explicit because non-blocking is now the default when
-    // a caller is known. This test covers the blocking path's caller-aware
-    // routing, which still exists and is still reachable — see
-    // `agent_send_defaults_to_async_when_caller_is_known` for the new default.
+    // `"async": false` is explicit because non-blocking is now the default when a caller is known.
+    // This test covers the blocking path's caller-aware routing, which still exists and is still reachable — see `agent_send_defaults_to_async_when_caller_is_known` for the new default.
     let input = serde_json::json!({ "agent_id": "target", "message": "hi", "async": false });
 
     let result =
@@ -1183,8 +1181,7 @@ async fn agent_send_no_key_with_caller_routes_to_send_to_agent_as() {
 async fn agent_send_same_key_routes_to_as_with_key_both_calls() {
     let cap = Arc::new(DispatchCapture::default());
     let kernel: Arc<dyn KernelHandle> = cap.clone();
-    // `"async": false` throughout: this test pins the keyed BLOCKING dispatch,
-    // which non-blocking delegation does not go through.
+    // `"async": false` throughout: this test pins the keyed BLOCKING dispatch, which non-blocking delegation does not go through.
     let input = serde_json::json!({
         "agent_id": "target",
         "message": "turn one",
@@ -1224,8 +1221,7 @@ async fn agent_send_distinct_keys_produce_isolated_dispatch() {
     let call = |key: &'static str| {
         let kernel = kernel.clone();
         async move {
-            // `"async": false`: distinct-key isolation is a property of the
-            // blocking keyed dispatch.
+            // `"async": false`: distinct-key isolation is a property of the blocking keyed dispatch.
             let input = serde_json::json!({
                 "agent_id": "target",
                 "message": "msg",
@@ -1330,7 +1326,8 @@ async fn agent_send_defaults_to_async_when_caller_is_known() {
 /// Omitting `async` WITHOUT a caller agent still dispatches blocking, because the async path cannot serve a callerless send at all.
 ///
 /// `tool_agent_send` rejects `async` with `InvalidParameter` when `caller_agent_id` is `None` — the tracker has nowhere to route the completion — while the blocking path has explicit `(None, Some(key))` and `(None, None)` arms for system-initiated sends.
-/// So an unconditional non-blocking default would convert every kernel-initiated `agent_send` into an error. This test is the guard against "simplifying" the conditional default into `unwrap_or(true)`.
+/// So an unconditional non-blocking default would convert every kernel-initiated `agent_send` into an error.
+/// This test is the guard against "simplifying" the conditional default into `unwrap_or(true)`.
 #[tokio::test]
 async fn agent_send_without_caller_still_defaults_to_blocking() {
     let cap = Arc::new(DispatchCapture::default());
--- a/crates/librefang-runtime/src/tool_runner/tests/mod.rs
+++ b/crates/librefang-runtime/src/tool_runner/tests/mod.rs
@@ -1161,7 +1161,11 @@ async fn agent_send_no_key_no_caller_routes_to_send_to_agent() {
 async fn agent_send_no_key_with_caller_routes_to_send_to_agent_as() {
     let cap = Arc::new(DispatchCapture::default());
     let kernel: Arc<dyn KernelHandle> = cap.clone();
-    let input = serde_json::json!({ "agent_id": "target", "message": "hi" });
+    // `"async": false` is explicit because non-blocking is now the default when
+    // a caller is known. This test covers the blocking path's caller-aware
+    // routing, which still exists and is still reachable — see
+    // `agent_send_defaults_to_async_when_caller_is_known` for the new default.
+    let input = serde_json::json!({ "agent_id": "target", "message": "hi", "async": false });
 
     let result =
         super::agent::tool_agent_send(&input, Some(&kernel), Some("parent-agent"), None, None)
@@ -1179,10 +1183,13 @@ async fn agent_send_no_key_with_caller_routes_to_send_to_agent_as() {
 async fn agent_send_same_key_routes_to_as_with_key_both_calls() {
     let cap = Arc::new(DispatchCapture::default());
     let kernel: Arc<dyn KernelHandle> = cap.clone();
+    // `"async": false` throughout: this test pins the keyed BLOCKING dispatch,
+    // which non-blocking delegation does not go through.
     let input = serde_json::json!({
         "agent_id": "target",
         "message": "turn one",
         "conversation_key": "thread-abc",
+        "async": false,
     });
 
     super::agent::tool_agent_send(&input, Some(&kernel), Some("parent-agent"), None, None)
@@ -1193,6 +1200,7 @@ async fn agent_send_same_key_routes_to_as_with_key_both_calls() {
         "agent_id": "target",
         "message": "turn two",
         "conversation_key": "thread-abc",
+        "async": false,
     });
     super::agent::tool_agent_send(&input2, Some(&kernel), Some("parent-agent"), None, None)
         .await
@@ -1216,10 +1224,13 @@ async fn agent_send_distinct_keys_produce_isolated_dispatch() {
     let call = |key: &'static str| {
         let kernel = kernel.clone();
         async move {
+            // `"async": false`: distinct-key isolation is a property of the
+            // blocking keyed dispatch.
             let input = serde_json::json!({
                 "agent_id": "target",
                 "message": "msg",
                 "conversation_key": key,
+                "async": false,
             });
             super::agent::tool_agent_send(&input, Some(&kernel), Some("parent-agent"), None, None)
                 .await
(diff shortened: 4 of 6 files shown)


houko:
Automated CLAUDE.md convention pass. Changelog fragment, docs, and tool-description wording all checked out clean. Two commit-level findings, left as comments rather than fixed since remediation means rewriting this branch's history, which isn't mine to do unilaterally:

1. Commit `56371d6c` ("docs(runtime): reflow new agent_send comments to one sentence per line") has git author identity `Claude <noreply@anthropic.com>`. CLAUDE.md's `commit-msg` hook section states this is rejected even when the message text itself is clean — please re-author (`git commit --amend --author`) before merge.
2. The `feat` commit (`90dd4546`, "make agent_send non-blocking by default…") has body paragraphs with multiple sentences run together on one line, rather than one-sentence-per-line per CLAUDE.md's prose-wrapping rule, which explicitly covers commit message bodies.

---

crates/librefang-runtime/src/tool_runner/agent.rs
```diff
@@ -57,7 +57,16 @@ pub(super) async fn tool_agent_send(
...
+    //
+    // Non-blocking is the DEFAULT when a caller agent is known, because the blocking default made every unannotated delegation a timeout risk: the model had to predict that the callee would be slow and opt out in advance, and a wrong guess burned the turn on `tool_timeout_secs`.
+    //
+    // The default is deliberately conditional on `caller_agent_id` rather than an unconditional `true`.
+    // The async path below requires a known caller so the tracker can route the completion back, and rejects the call outright without one; the blocking path has explicit `(None, _)` arms for system-initiated sends.
+    // An unconditional default would therefore turn every callerless send — the kernel's own system-initiated dispatches — into an `InvalidParameter` error.
+    // An explicit `"async": true` still errors without a caller, which is the pre-existing contract and is left untouched.
+    let async_mode = input["async"]
+        .as_bool()
+        .unwrap_or_else(|| caller_agent_id.is_some());
```
houko, line 69: Flagging this for maintainer judgment rather than fixing it myself, per CLAUDE.md's guidance that concurrency/session-semantics changes need a human call.

Making non-blocking the default here removes a throttle that was implicit in the old blocking default: previously a single calling agent's own turn serialized its delegations, because each `agent_send` tool call blocked until the callee replied before the model could make another one. With `async_mode` now `true` whenever `caller_agent_id.is_some()`, a caller can fire off many delegations in quick succession within one turn without ever waiting on one.

Two things in the surrounding architecture make that fan-out effectively uncapped, not just faster:

1. The comment a few lines below this diff (`"the synchronous AGENT_CALL_DEPTH scope is intentionally NOT carried across the async boundary"`) confirms the depth guard that limits A→B→C chains does not apply to the spawned delegation.
2. `docs/architecture/trigger-dispatch-concurrency.md` states explicitly: *"`agent_send`, channel bridges, and cron all serialize on the existing `agent_msg_locks` / `session_msg_locks` inside `send_message_full` and are not throttled by the per-agent cap"* — i.e. `agent_send` is deliberately outside all three of the layered caps (global lane semaphore, per-agent semaphore, per-session mutex-as-throttle) that bound trigger-dispatcher fan-out. And `register_async_task` (`crates/librefang-kernel/src/kernel/task_registry.rs`) only dedupes on `(agent_id, prompt_hash)` — it has no concurrency limit of its own.

None of this is new in this PR — the async path and its lack of a fan-out cap both predate it (#6043). What changes here is exposure: the behavior that used to require an explicit `"async": true` opt-in per call is now what every model-authored `agent_send` gets unless it explicitly opts out. A chatty or compromised agent (or one simply following the new tool description's advice literally) can now spawn an unbounded number of concurrent detached delegations from a single turn where before it could only have one in flight at a time.

Worth a maintainer decision: is the per-session mutex inside `send_message_full` considered sufficient here because it only needs to prevent *races*, not bound *throughput*? Or does raising this to the default warrant adding a cap (e.g. a per-caller in-flight-delegation limit) alongside it, the way trigger dispatch has one?

---

houko (commented):
Automated review against CLAUDE.md. One inline comment below on the concurrency/fan-out implications of the new default — left for maintainer judgment rather than fixed, since it's exactly the kind of session/concurrency-semantics call CLAUDE.md says needs a human.

Separately, flagging without a code comment because it isn't tied to a diff line: the current head commit (56371d6c, "docs(runtime): reflow new agent_send comments to one sentence per line") is authored as `Claude <noreply@anthropic.com>`. CLAUDE.md's Git Conventions / Commit & PR hygiene section is explicit that commit messages, PR bodies, and comments must carry no Claude/Anthropic/AI attribution, and the repo's `commit-msg` hook is specifically designed to reject this at the author-identity level (`git var GIT_AUTHOR_IDENT`), independent of message content — this commit's message text itself is clean, only the author identity trips the rule. Fixing it correctly requires rewriting that commit (amend or rebase), which in turn requires a force-push to update the already-pushed branch tip; force-pushing was out of scope for this automated pass, so it's left for @houko to amend locally (`git commit --amend --author="..."`) and force-push, or to let a maintainer do the same.

Everything else checked out: the `changelog.d/changed/6740-agent-send-async-default.md` fragment is correctly formed (right section, one-sentence-per-line, ends `(#6740) (@houko)`, passes `scripts/check-changelog-attribution.py --all-unreleased`); this PR touches no HTTP routes so the mandatory-integration-test rule doesn't apply; the two human-authored commits are conventional-commit-formatted with no AI attribution in their text; and the tool description / docs updates in `definitions.rs` and the `page.mdx` files (plus their `zh/` mirror) are consistent with the new default.

---

crates/librefang-runtime/src/tool_runner/agent.rs
houko, line 69: CLAUDE.md's "Trigger dispatch concurrency" section documents that `agent_send` is deliberately excluded from the `Lane::Trigger` and per-agent (`max_concurrent_invocations`) semaphores; it's meant to rely only on the per-session mutex inside `send_message_full`.
That was a safe assumption while non-blocking mode was opt-in (`"async": true` was the rare case).

Flipping the default here means every caller-known `agent_send` now takes the `send_to_agent_async_tracked` path.
The dispatch spawn inside that function (`crates/librefang-kernel/src/kernel/handles/agent_control.rs`, the `tokio::spawn` block starting around line 238) fires the callee's turn via `send_message_as` / `send_message_as_with_key` with no `agent_concurrency_for` or `Lane::Trigger` acquisition at all.
Compare that to the *completion*-delivery wake-idle path in `crates/librefang-kernel/src/kernel/task_registry.rs::spawn_wake_idle_turn` (lines ~427-504), which explicitly takes both semaphores and documents why in its "Concurrency (#5033 review)" doc comment: "Without these two acquisitions a batch agent that fires N ... calls would have N ... turns spawn in parallel on completion, silently bypassing operator-set fan-out caps."

Net effect after this PR: a single agent that calls `agent_send` in a loop, or that gets several `tool_use` blocks fanned out in one turn, can now spawn an unbounded number of concurrent target-agent turns with zero rate limiting at dispatch time.
Previously the blocking default throttled this naturally, because the caller's own loop waited on each call before it could issue the next.
The per-session mutex (`session_msg_locks`) still protects one `(agent, session)` pair from concurrent writers, but it does nothing to cap how many *distinct* target-agent turns one caller fans out concurrently.

This is a behavior-semantics question rather than a small fix, so raising it for a design call rather than editing: should the dispatch spawn in `send_to_agent_async_tracked` also acquire the target agent's `agent_concurrency_for` semaphore (mirroring the wake-idle path), or is unbounded dispatch intentional here because delegation targets are assumed to have their own caps enforced elsewhere?

---

crates/librefang-runtime/src/tool_runner/agent.rs
houko, line 1: Commit 56371d6 on this branch ("docs(runtime): reflow new agent_send comments to one sentence per line") has its git author identity set to `Claude <noreply@anthropic.com>`.

CLAUDE.md is explicit that this is checked independently of the commit message text: "the `commit-msg` hook ... separately rejects a commit whose *author identity* (`git var GIT_AUTHOR_IDENT` ...) resolves to Claude / Anthropic even when the message itself is clean." The repo-wide "No AI / Claude attribution" rule under Git Conventions applies the same way.

The commit is already pushed to `origin/fix/agent-send-async-default`, so fixing the author identity means rewriting that commit (amend/rebase) and force-pushing the branch. That's out of scope for this pass, so flagging it here rather than doing it: a maintainer should re-author (or squash) the commit and force-push before merge.

---

houko (commented):
Automated review pass against CLAUDE.md.
Two findings, both design/process-level rather than mechanical, left inline: a concurrency-cap gap introduced by making this the default path, and a commit with an AI git-author identity already on the branch.
No mechanical issues found (changelog fragment format, prose wrapping, and config/test conventions all check out for this diff).

---

crates/librefang-runtime/src/tool_runner/agent.rs
houko, line 69: Design question for reviewers, not a bug: this flips a widely-documented default (`agent_send` is explicitly called out in the repo's session_mode/concurrency docs as a primitive other subsystems reason about) for every existing agent/skill that never set `"async"`, with no opt-out at the manifest or config level — only per-call `"async": false`.

The PR body's own table is convincing that an *unconditional* `true` is wrong, but it doesn't make the case for why this shouldn't instead be an `agent.toml` knob (the repo already has this shape of precedent: `session_mode` is per-agent/per-trigger/per-job with a documented resolution order, not a global flip). A prompt/skill that assumes an inline reply from `agent_send` silently starts getting a `task_id` instead — same failure mode `session_mode` was designed to make explicit and opt-in per agent, not implicit and global.

Given CLAUDE.md's "don't reach for a flag when you can just change the code" guidance, a straight default change is the right instinct *if* the blast radius is acceptable — but that's a product/back-compat judgment call for a maintainer, not something this review should wave through silently. Worth an explicit maintainer sign-off on the breaking-default trade-off called out in the PR's own "Behaviour change reviewers should weigh" section before merge.

---

crates/librefang-runtime/src/tool_runner/agent.rs
houko, line 1: Process finding, not a code issue in this file: commit `56371d6c18a796c93ad10511bda98ddf17a7c588` ("docs(runtime): reflow new agent_send comments to one sentence per line"), which touches this exact file, has author identity `Claude <noreply@anthropic.com>` (per `git log`/the GitHub commits API — the commit message text itself is clean).

CLAUDE.md's commit-msg hook is documented to reject "a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean," and the repo-wide rule is "No AI / Claude attribution in commit messages, PR bodies, or comments." This one slipped through — the branch presumably wasn't set up with `core.hooksPath scripts/hooks` (`just setup`), so the hook never ran.

Not fixing this myself: correcting author identity needs a history rewrite (amend + force-push) on a branch that already has review activity, and force-push here requires explicit maintainer sign-off per the repo's hard rules. Flagging so a maintainer can decide whether to have the author rebase with a corrected identity before merge, or accept it as-is.

---

houko, line 1: Re: lines 110-118 (unchanged by this diff, but their meaning changes with it) — this comment (correctly) documents that `AGENT_CALL_DEPTH` is a task-local (`tokio::task_local!`, see `tool_runner/mod.rs:127`) and does not survive the `tokio::spawn` in `send_to_agent_async_tracked` (`crates/librefang-kernel/src/kernel/handles/agent_control.rs:238`), so the callee starts its own depth chain at 0.

That behavior predates this PR (#6043), but it was previously reachable only when the caller explicitly opted in with `"async": true`. Making non-blocking the *default* means the A->B->C depth guard just above (lines 101-108) — the fork-bomb / runaway-delegation protection `max_agent_call_depth` exists for — is now bypassed by default for the common case (any agent that doesn't think about blocking, per this PR's own test naming), not just for callers who deliberately asked for detached execution.

Please confirm this is an accepted trade-off rather than an oversight: does anything else now bound the total number of live detached delegation chains a single agent can fan out (depth-unlimited, since each hop resets to 0), or does `max_agent_call_depth` effectively stop being the backstop it was for the majority of `agent_send` traffic once this ships?

---

houko, line 1: Re: the `send_to_agent_async_tracked` call at lines 125-133 (unchanged by this diff, but now reachable by default) — concurrency-cap gap worth checking before this becomes the default path.

Per repo docs (`CLAUDE.md` "Trigger dispatch concurrency"), `agent_send` is explicitly *outside* the trigger dispatcher's layered caps (global `Lane::Trigger` semaphore, per-agent semaphore, per-session mutex) — it "still serialize\[s\] at the existing per-agent / per-session locks inside `send_message_full`."

That serialization was real in practice under the old default because a blocking `agent_send` occupies the caller's own tool-call slot until the reply lands — a caller could only ever have one outstanding delegation at a time, which was free backpressure. `send_to_agent_async_tracked` (`crates/librefang-kernel/src/kernel/handles/agent_control.rs:238`) detaches via a bare `tokio::spawn` with no semaphore/permit acquisition, so once non-blocking is the default, a single turn that calls `agent_send` N times now fans out N concurrently-running callee loops with nothing capping N. The per-session mutex inside `send_message_full` only serializes writes *within* one target session; it doesn't limit how many distinct sessions/agents one caller can fan out to in parallel.

Is an unbounded fan-out from a single caller turn (bounded only by `max_agent_call_depth` per chain, not by breadth) an intended consequence of making this the default, or does it need a cap analogous to the trigger lane's?

---

houko (commented):
Automated daily review. No mechanical/safe fixes applied (changelog fragment, commit conventional-commit format, and PR description are all already correct). Four judgment-level findings left inline for a maintainer: (1) whether this backwards-compatible-breaking default flip should instead be an opt-in `agent.toml` knob given the `session_mode` precedent, (2) the default flip silently bypasses the `max_agent_call_depth` chain guard for the common case since the async task-local doesn't survive the `tokio::spawn`, (3) no concurrency cap on the number of concurrent detached delegations a single caller can fan out now that blocking backpressure is gone, and (4) commit `56371d6` in this branch has AI (Claude/Anthropic) author identity, which the repo's own `commit-msg` hook is designed to reject — not fixed here because correcting it needs a history rewrite/force-push on a branch already under review.

---

crates/librefang-runtime/src/tool_runner/agent.rs
houko, line 69: Backward-compat / safety concern per CLAUDE.md's guidance on default-behavior changes to widely used primitives (this flip touches every `agent_send` call site that omits `"async"` and has a known caller — i.e. most agent-to-agent delegation).

Further down (the "Non-blocking path (#6043)" block, unchanged by this diff) `AGENT_CALL_DEPTH` is intentionally **not** carried across the async boundary — the callee starts its own depth chain. Before this PR that reset only happened when a caller explicitly opted in to `"async": true`. After this PR it happens on the *default* path for every caller-known `agent_send`, so the `max_agent_call_depth` runaway-recursion guard (the same class of protection CLAUDE.md calls out for `agent_spawn`/`agent_kill` fork-bomb prevention) is now bypassed by default rather than by explicit opt-in.

`agent_send_depth_exceeded_is_permission_denied` still passes because the depth check runs before the async/blocking branch on the *initiating* call, but there's no test exercising the multi-hop case (A → B → C through the new async-default path) to confirm the depth reset's blast radius is acceptable now that it is the common case rather than the opt-in exception. Worth an explicit regression test and a callout in the PR description (which currently discusses the reply-timing behavior change but not this depth-guard interaction) so reviewers can weigh it consciously.

---

houko (commented):
Automated CLAUDE.md-conformance pass. Two findings, both left as review comments (ambiguous / not mine to fix directly):

1. **Commit authorship violates the no-AI-attribution rule.** The current PR head, `56371d6` ("docs(runtime): reflow new agent_send comments to one sentence per line"), has author identity `Claude <noreply@anthropic.com>`. CLAUDE.md's Git Conventions section is explicit that the `commit-msg` hook "rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean" — this is exactly that case; the message itself is clean, but `GIT_AUTHOR_IDENT` is not. Fixing this requires rewriting the commit (reauthor/squash) and force-pushing, which is outside what an automated review pass is permitted to do here (no force-push). Recommend the branch owner amend the author on that commit (or squash it into `90dd454`, which is already correctly attributed to a human) before merge.
2. **Depth-guard interaction with the new async default** — inline comment on `crates/librefang-runtime/src/tool_runner/agent.rs` line 69. Making non-blocking the default means the pre-existing `AGENT_CALL_DEPTH` reset-on-async-boundary behavior (previously only reachable via explicit `"async": true`) now applies to the common, unannotated case. No test covers the multi-hop depth-reset interaction under the new default, and the PR description's "Behaviour change reviewers should weigh" section covers reply-timing but not this.

Everything else checked out: the prose-wrapping rule (one sentence per line) is already satisfied by the branch's own `56371d6` reflow commit, the changelog fragment in `changelog.d/changed/` matches the required format and section, no config fields were touched (so the Default-impl rule doesn't apply), and no ordering-sensitive collections were introduced (deterministic-prompt-ordering rule doesn't apply). No mechanical fixes were made to this PR's files.

---

houko:
Correction to my own prior comment on this PR: commit `68531cc` (the changelog-wording fix I pushed earlier in this run) carries author identity `Claude <noreply@anthropic.com>` — the same AI-attribution problem already flagged on `56371d6` in this PR.
Root cause: the automation container's global git config defaulted to that identity, and my commit didn't override it.
I've fixed the global git config for the remainder of this run so it won't recur, but `68531cc` itself is already pushed and fixing its author metadata now requires an amend + force-push, which this routine's hard rules don't let it do.
Recommend folding it into the same authorship cleanup already needed for `56371d6` when a maintainer rebases this branch — sorry for adding a second instance of the exact issue I was reviewing for.

---

crates/librefang-runtime/src/tool_runner/agent.rs
```diff
@@ -57,7 +57,16 @@ pub(super) async fn tool_agent_send(
...
```
houko, line 70: Flagging this as a behavior-changing default on a widely-used primitive, per the concurrency notes in CLAUDE.md around `agent_send` / `session_mode`.

The conditional default itself (`caller_agent_id.is_some()`) is sound, and the async-tracked path it now routes through by default (`send_to_agent_async_tracked` → `complete_async_task`) is pre-existing machinery already exercised by `async: true` and by the async-workflow path, so this isn't new plumbing. Two things I'd still want confirmed before this becomes the default for every unannotated `agent_send`:

1. **Mid-turn injection into the caller's own session.** `complete_async_task` injects the callee's reply into the caller's session "mid-turn or wake-idle" per the doc comment on `send_to_agent_async_tracked`. That's an existing capability, but today it's opt-in (`async: true`) and rare. Flipping the default means *every* unannotated delegation now does this, including on `session_mode = "persistent"` sessions that other triggers/cron fires may also be writing to concurrently. Is there a test that pins "the async completion write and an in-progress turn on the same session don't stomp each other," or is that entirely delegated to whatever `complete_async_task` already does for workflows? If it's the latter, a one-line pointer to that coverage in the PR description would help reviewers who don't already know the workflow-async history.
2. **Same-turn fan-out to the same target.** `agent_send` isn't in `tool_classifier.rs`'s explicit list, so it falls to `Unknown → WriteShared`, which `execute_tool_group` allows to run concurrently with other `WriteShared` calls in the same batch. That's pre-existing (a model could already emit two `agent_send { async: true }` calls to the same target in one turn), but making non-blocking the *default* makes this the common case instead of an edge case — worth a test showing two concurrent `agent_send`s (same turn, same target) don't race past whatever locking `send_message_full` relies on, if that isn't already covered elsewhere.

Neither is a request to change the approach — just asking for the concurrency story to be explicit given how central `agent_send` is, per CLAUDE.md's own callout that this exact kind of change deserves scrutiny rather than a mechanical pass.

---

houko (commented):
Automated daily review pass. One inline concurrency question on the default-flip itself (see below).

Separately, flagging a repo-hygiene issue independent of the code: two commits in this PR's history have author identity `Claude <noreply@anthropic.com>` — `56371d6` ("docs(runtime): reflow new agent_send comments...") and the current branch HEAD, `68531cc` ("docs(changelog): attribute callerless agent_send fallback..."). CLAUDE.md's commit hygiene rule bans AI/Claude attribution in commits, and per the CLAUDE.md description of `scripts/hooks/commit-msg`, that hook rejects a commit whose *author identity* resolves to Claude/Anthropic even when the message text itself is clean — which is exactly the shape of these two. Since this isn't something I can fix by pushing a new commit (it needs history rewritten and re-authored, which means a rebase + force-push to someone else's branch), flagging here for the author/maintainer to squash or re-author before merge.

---

crates/librefang-runtime/src/tool_runner/agent.rs
houko, line 69: Behavioral question worth a maintainer look before merge: flipping the default to non-blocking removes the implicit throttle the old blocking-by-default behavior gave for free. Previously, an agent loop that called `agent_send` in a fan-out pattern (broadcast to N agents, retry-without-backoff, etc.) had each call serialize on the callee's reply before the next could fire, since the model had to wait for the tool result either way. With non-blocking now the default, the same loop spawns N detached `tokio::spawn` tasks (see `send_to_agent_async_tracked` in `crates/librefang-kernel/src/kernel/handles/agent_control.rs`) with no cap — `AsyncTasksConfig` only has `default_timeout_secs` / `notify_on_timeout`, and the tracker's dedupe only matches on `(target agent, prompt_hash)`, so distinct message bodies never collapse.

Was a per-agent in-flight-delegation cap considered (something in the spirit of the existing `Lane::Trigger` semaphore documented for trigger dispatch), or is unbounded fan-out accepted as within the async-task-tracker's existing design envelope? Not blocking if the latter, but it's a new-by-default surface that previously required an explicit opt-in (`"async": true`).

---

crates/librefang-runtime/src/tool_runner/definitions.rs
```diff
@@ -320,7 +320,7 @@ use instead of web_fetch + file_write (which round-trips the entire body through
             },
             ToolDefinition {
                 name: tool_name::AGENT_SEND.to_string(),
-                description: "Send a message to another agent. By default this BLOCKS until the agent replies and returns their response — only use the blocking mode for quick sub-questions whose answer you need within this turn. For any delegation that may take a while (research, multi-step work), set \"async\": true so you are not blocked and don't hit the tool timeout. Accepts UUID or agent name. Use agent_find first to discover agents.".to_string(),
+                description: "Send a message to another agent. By default this is NON-BLOCKING: you get back a task_id immediately and the target's reply is delivered to your session when it finishes, so a long delegation cannot burn your turn on the tool timeout. Set \"async\": false for a quick sub-question whose answer you need within this same turn — that blocks until the agent replies and returns their response directly. Accepts UUID or agent name. Use agent_find first to discover agents.".to_string(),
```
houko, line 323: The changelog fragment does the "why" well, but the only signal an *existing* agent gets about this default flip is this description text, which the model only reads freshly on its next prompt build — an agent whose system prompt or few-shot examples hard-code "call agent_send and read the reply text" (a reasonable pattern under the old blocking default) will now silently receive a `{"task_id": ..., "status": "delegated", ...}` JSON blob instead of the expected response, with no error and no crash — just a wrong-shaped result the agent's own prompt wasn't written to handle.

That's a real behavior change for pre-existing agents that isn't observable via a test failure, only via the agent behaving oddly in production. Was there a shorter-term escape hatch considered (e.g. an org-level config to keep the old default while operators audit/update prompts), or is "the tool description already says the new default" the intended mitigation? Asking because CLAUDE.md flags this exact class of default-flip as needing extra scrutiny beyond the changelog entry.

---

houko (commented):
Reviewed the default-flip diff (agent.rs, definitions.rs, tests, changelog fragment, docs). Mechanics check out: `caller_agent_id` is the same identity primitive already used pervasively for tool-call scoping elsewhere, so "caller is known" detection isn't ambiguous. The callerless (system-initiated) path is explicitly guarded and covered by a new test (`agent_send_without_caller_still_defaults_to_blocking`), and the new default path has its own test (`agent_send_defaults_to_async_when_caller_is_known`); existing blocking-path tests were updated with explicit `"async": false` rather than silently changing behavior. Error surfacing on the async path (`TaskStatus::Failed` → `TaskCompletionEvent` injected into the session) is pre-existing #6043/#6650 infra, unmodified here, so no new drop-on-the-floor risk. The two session-less bridges (MCP HTTP, REST tool endpoint) already fall back to `AsyncSendOutcome::Inline`, so they see no behavior change despite the new default. Changelog fragment is present under `changed/`, explains the why, and ends `(#6740) (@houko)` per CLAUDE.md. Docs (EN+ZH) updated consistently; no stale references to the old default found elsewhere.

Left two comments for maintainer judgment, not fixes — this being a default-behavior flip, per instructions I'm not touching either without a human call: (1) whether unbounded fan-out of detached delegation tasks under the new default warrants a concurrency cap, and (2) whether existing agents whose prompts assume the old blocking reply shape need more signal than the updated tool description.

---

crates/librefang-runtime/src/tool_runner/agent.rs
```diff
@@ -57,7 +57,16 @@ pub(super) async fn tool_agent_send(
...
+    //
+    // Non-blocking is the DEFAULT when a caller agent is known, because the blocking default made every unannotated delegation a timeout risk: the model had to predict that the callee would be slow and opt in to async in advance, and a wrong guess burned the turn on `tool_timeout_secs`.
+    //
+    // The default is deliberately conditional on `caller_agent_id` rather than an unconditional `true`.
+    // The async path below requires a known caller so the tracker can route the completion back, and rejects the call outright without one; the blocking path has explicit `(None, _)` arms for system-initiated sends.
+    // An unconditional default would therefore turn every callerless send — the kernel's own system-initiated dispatches — into an `InvalidParameter` error.
+    // An explicit `"async": true` still errors without a caller, which is the pre-existing contract and is left untouched.
+    let async_mode = input["async"]
+        .as_bool()
+        .unwrap_or_else(|| caller_agent_id.is_some());
```
houko, line 69: Design question for a human to weigh, not a mechanical issue: making non-blocking the default when a caller is known removes the implicit backpressure that blocking used to provide. Previously a caller naturally serialized its delegations to a given callee one-at-a-time (each `agent_send` blocked until the reply came back). Now the model can fire several `agent_send` calls to the same callee agent in quick succession within one turn, each returning immediately with a `task_id`, before any of them complete.

Per CLAUDE.md's "Trigger dispatch concurrency" section, `agent_send` is documented as serializing only via the existing per-agent/per-session locks inside `send_message_full` — it is explicitly *outside* the `Lane::Trigger` semaphore and `max_concurrent_invocations` cap that scope to the trigger dispatcher only. That's pre-existing behavior (already reachable via explicit `"async": true"`), so this diff doesn't introduce a new code path — but flipping the default is likely to increase how often callees see multiple concurrent async deliveries fired by one caller, since the model no longer needs to wait for a reply before issuing the next delegation.

Worth confirming before merge: is a callee agent's own concurrency still bounded in this scenario (e.g. by its persistent-session-implies-cap-1 auto-clamp, or per-agent semaphore), or can one chatty caller now fan out an unbounded number of concurrent in-flight delegations to the same target that blocking used to naturally throttle?

---

houko (commented):
Automated daily review. The PR is well-documented and already carries a thoughtful changelog fragment, conditional-default rationale, and both new-default and old-behavior test coverage — no mechanical CLAUDE.md violations found there (changelog format, prose wrapping, session_mode resolution order, no AI attribution, no stray HashMap, no missing tests).

Pushed one small mechanical fix (commit 5360d59): a comment explaining the new default said the model had to "opt out in advance" to avoid blocking, but the old default was blocking and the escape hatch was opting IN to async — the wording was inverted relative to the PR's own description of the problem. Fixed to match.

Left one inline comment flagging a design question for a maintainer: whether removing blocking-as-default backpressure could let one caller fan out more concurrent in-flight delegations to the same callee than before, given `agent_send` is documented as sitting outside the trigger-lane concurrency caps.

---

PR #6743, opened 2026-08-07
fix(kernel,cli): bound nested workflow agent turns and size the daemon worker stack

houko:
Refs #6659

## Diagnostic honesty first

Read this section before the changes.

**The depth cap fixes a confirmed defect that stands on its own merit.** `AGENT_CALL_DEPTH` was incremented in exactly one place — `tool_runner/agent.rs` — and the `workflow_run` tool bypassed it entirely. `tool_workflow_run` calls `kh.run_workflow(...)`, which runs `LibreFangKernel::run_workflow` inline, whose per-step `send_message` closure awaits `self.send_message_full(...)` inline, which is a complete nested agent turn whose own `workflow_run` repeats the whole thing. `WorkflowRun` carries no parent-run, depth or lineage field, so the only bound in production was the wall-clock `triggers.max_workflow_secs`. That is unguarded recursion regardless of what caused #6659.

**The worker stack sizing is defence in depth, not a fix.** It buys headroom so a legal-depth chain of large frames does not abort before the cap can refuse it.

**Neither change is proven to be the cause of the reported SIGABRT.** #6659's mechanism is not confirmed. Its 52 frames over a 2 MiB stack works out to roughly 40 KB per frame, which reads as bounded depth with very large frames rather than deep recursion — and the original crash report is unrecoverable (`~/Library/Logs/DiagnosticReports` and its `Retired/` are both empty), so the frame chain cannot be re-read. Both readings are consistent with both fixes being correct, which is why they are here; that is not the same as having identified the root cause.

**What would settle it.** Either a symbolized crash report showing the repeated frame chain and its actual depth, or the next occurrence on the now-8 MiB worker stack: an abort that still happens implies genuinely unbounded recursion somewhere, and aborts that stop imply bounded depth with fat frames. `Refs`, not `Closes`, for exactly that reason — the issue should stay open until one of those arrives.

One number that supports the fat-frame reading, measured in this PR rather than asserted: `run_agent_loop`'s future is **56 816 bytes** on macOS aarch64 debug. That is one function in the per-level chain, and it is the one that gets stacked again on every nested turn.

## Changes

- **`LibreFangKernel::run_workflow` refuses a run entered from `max_agent_call_depth` already-stacked agent turns** (`triggers_and_workflow.rs`). The refusal is `LibreFangError::CapabilityDenied`, matching the precedent and reasoning in `tool_agent_send`: a self-imposed kernel-policy quota should not surface as a 5xx that retry logic reads as a downstream crash. The check sits ahead of `create_run`, so a refused nesting leaves no orphan `Pending` run in the engine.
- **Two downstream hops were erasing that policy shape, and both are fixed.** This was found while checking the status-code claim in an earlier draft of this body, and it made the first version of the refusal useless in practice. `WorkflowRunner::run_workflow` — the trait impl `workflow_run` actually goes through — stringified *every* `KernelError` into `KernelOpError::Internal`, and `tool_workflow_run` then mapped everything through `ToolError::upstream`. So an agent nesting too deep received an opaque 5xx-class upstream failure: exactly the shape `tool_agent_send`'s comment argues against. The trait impl now passes `CapabilityDenied` through verbatim (`KernelOpError` *is* `LibreFangError`, so the variant survives) and leaves every other error on the historical stringified `Internal` path; `tool_workflow_run` maps that one variant to `ToolError::PermissionDenied`. Beyond the status class this decides turn survival: `PermissionDenied` classifies as `ToolExecutionStatus::Denied`, a **soft** failure, so an agent that keeps hitting the cap does not burn through `MAX_CONSECUTIVE_ALL_FAILED` and lose the turn to an abort.
- **The workflow step dispatch now enters the depth scope**, wrapping `send_message_full` in `tool_runner::with_agent_call_depth`. Without this half, every nesting level would still read depth 0 and the check above would never fire.
- **Workflow nesting is charged to the existing inter-agent budget** (`config.toml: max_agent_call_depth`, default 5) rather than a second counter, because `A --agent_send--> B --workflow--> C` stacks turns exactly the way `A -> B -> C` does and one operator knob should cap both.
- **New public helper `librefang_runtime::tool_runner::with_agent_call_depth`.** The task-local is module-private, so the kernel cannot enter its scope directly. The helper boxes the wrapped future for the same documented reason `held_agent_locks::scope` does — the thing being wrapped is an entire agent turn, so inlining it stacks another full copy per nesting level, adding stack cost at exactly the boundary the helper exists to bound.
- **`tool_agent_send` now routes through the same helper.** It previously inlined its async block directly into `AGENT_CALL_DEPTH.scope`, so this removes one inlined copy of the callee turn from the *confirmed* `A -> B -> C` path, and keeps a single definition of "one level deeper" shared with the workflow path. Its depth read also moves from a duplicated `try_with` to the existing `current_agent_depth()`.
- **Daemon tokio worker stack: 2 MiB (tokio's default) → 8 MiB** (`librefang-cli/src/commands/daemon.rs`). With `worker_threads(2)`, one overflowing turn takes both workers, the HTTP API and every cron job with it.
- **Desktop embedded server: same 8 MiB, on both threads that matter** (`librefang-desktop/src/server.rs`). It boots the same kernel as the daemon and was not mentioned in the original plan. `Builder::thread_stack_size` covers the runtime's worker and blocking threads only — the outermost turn there is polled by `rt.block_on` on the `std::thread::Builder`-spawned `librefang-server` thread, whose stack that setting does not touch, so that builder gets a matching `.stack_size(...)`.
- **A comment on the API-side workflow driver** (`routes/workflows/workflow.rs`) recording why it does *not* enter the depth scope and under what condition that stops being safe. See "Deviations".

## Verification

Native cargo on macOS aarch64. Every number below is from a run in this worktree.

- `cargo clippy --workspace --all-targets -- -D warnings` — **exit code 0**, captured explicitly with `cmd >log 2>&1; echo $?`, not inferred from a `Finished` line. The first run exited **101** on a `clippy::single_match` in one of my new tests; that is fixed and the re-run is the 0.
- `cargo fmt --check` — exit code 0.
- `cargo test -p librefang-kernel` — exit 0. 1367 lib tests passed, 0 failed, 1 ignored, plus 23 integration binaries all passing (17 + 20 + 27 + 9 + 9 + 8 + 8 + 6 + 6 + 5 + 4 + 4 + 4 + 4 + 4 + 3 + 3 + 2 + 2 + 1 + 0 + 0 + 0).
- `cargo test -p librefang-runtime` — exit 0. 2176 lib tests passed, 0 failed, 2 ignored, plus all integration binaries. The known load-sensitive `plugin_runtime::tests::overflow_kills_child_without_waiting_for_timeout` passed on this run.
- `cargo test -p librefang-cli` — exit 0. 345 + 3 + 3 + 2 + 2 passed, 0 failed.
- `cargo test -p librefang-desktop` — exit 0. 11 passed, 0 failed.
- `cargo test -p librefang-api --lib` — exit 0, 925 passed, 0 failed. Run because the change touches that crate, though the change there is a comment only.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` (exit 0) and the targeted kernel + runtime tests were all re-run after the error-shape fix and after a rebase onto a comment-rewrap commit that landed on this branch from another session; the two pre-existing source-shape sentinels on the workflow `send_message` closure pass in every run.

New tests:

- `kernel::tests::nested_workflow_run_past_max_agent_call_depth_is_capability_denied` — behavioral, and the load-bearing one. Boots a kernel with `max_agent_call_depth = 2`, registers a one-step workflow, then asserts the pairing: at depth 0 the run is *not* refused by the quota (it fails later, on the unregistered step agent) and a run is created; nested to depth 2 it returns `CapabilityDenied` naming the configured cap, and the run count does not increase. The pairing is what proves the refusal comes from the quota rather than from the workflow being unrunnable, and the run-count assertion is what proves the check precedes `create_run`.
- `kernel::tests::workflow_step_dispatch_enters_agent_call_depth_scope` — comment-stripped source-shape sentinel, in the established style of the two that already guard this same closure. It pins the *other* half: that the step dispatch calls `send_message_full` through `with_agent_call_depth`, and that the cap is read from `cfg.max_agent_call_depth`. It is a sentinel and not a behavioral test because observing the depth inside `send_message_full` needs a real LLM turn, and `librefang-kernel` has no driver-injection seam — there is no mock provider and no `set_driver` hook. Inventing one would be a larger change than the fix. Both existing sentinels on this closure still pass.
- `tool_runner::tests::with_agent_call_depth_adds_exactly_one_level_per_nesting` — each level adds exactly one, an inner level does not leak into the frame that resumes after it, and sibling scopes at the same level do not accumulate.
- `tool_runner::tests::agent_send_quota_sees_depth_established_by_with_agent_call_depth` — seeds depth 9 directly, lets the helper supply the tenth level, and asserts `agent_send`'s guard fires. This is what proves the helper writes to the same counter the guard reads, i.e. that a workflow step's agent cannot regain a full delegation budget by hopping through the workflow path.
- The trait-hop leg inside `nested_workflow_run_past_max_agent_call_depth_is_capability_denied` — drives `WorkflowRunner::run_workflow` (the path `workflow_run` takes, not the inherent method) and fails if the variant is folded back into `Internal`. Without this the same stringification could be reintroduced silently.
- `tool_runner::tests::workflow_run_depth_refusal_is_permission_denied_not_upstream` — drives `tool_workflow_run` against a stub that refuses, asserts the mapped variant *and* its `execution_status() == Denied`, and asserts that a non-policy kernel failure still arrives as `Upstream` so the mapping stays a narrow special case rather than a blanket reclassification.
- `agent_loop::tests::integration::run_agent_loop_future_size_stays_within_its_order_of_magnitude` — the 56 816-byte measurement above. Explicitly labelled in its own doc comment as a proxy for fat poll frames and *not* a test that would have caught this defect. It constructs the future without polling it, prints the size so anyone can re-read the figure with `--nocapture`, and asserts a deliberately generous 512 KiB bound so ordinary feature work does not have to re-baseline it.

Not verified:

- **No Windows or Linux run.** There is no Windows machine here. Every test above is macOS aarch64 only; CI covers the other two. Nothing in the change is platform-conditional and no test touches a host path.
- **No live daemon or real-LLM run.** Building a release binary and starting a daemon on port 4545 is blocked for AI sessions by `.claude/hooks/`, and the stack-size change cannot be exercised any other way — its effect is only observable on a real overflow.
- **The 8 MiB stack is not shown to prevent the reported crash**, because the reported crash cannot be reproduced. See the diagnostic section.
- **Per-future sizes for `send_message_full_inner` and `execute_llm_agent` were not measured.** They are private to `librefang-kernel` and measuring them the way `run_agent_loop` was measured needs a booted kernel and a constructible turn. Only the `run_agent_loop` figure is real.

## Deviations from the plan

- **The cap is read from `cfg.max_agent_call_depth`, not the `MAX_AGENT_CALL_DEPTH` const the plan cited.** That const is `#[allow(dead_code)]` and load-bearing nowhere; the live value is the config knob, via `kh.max_agent_call_depth()` on the tool path and `cfg.max_agent_call_depth` on the new kernel path. Reading the const would have made the workflow cap ignore an operator's `config.toml`. The dead const is left alone — removing it is unrelated to this change.
- **`clippy::large_futures` was deliberately not enabled**, per the plan. `clippy.toml` is empty and there is no `[workspace.lints]`, so that lint has never run, and CI runs clippy with `-D warnings` — enabling it in a codebase whose central problem is enormous futures would very likely fail the build on dozens of pre-existing hits and block this PR. `-Zprint-type-sizes` was also not used (it needs nightly). The `size_of_val` test above is the measurement instead, and `clippy.toml` is untouched.
- **Boxing further await sites is not attempted here**, and the reason is concrete rather than deferral. `AgentControl` is `#[async_trait]`, so `kh.send_to_agent_as` already returns a `Pin<Box<dyn Future>>` and the nested turn's state machine is already heap-allocated at the `agent_send` hop. An extra box there adds a `Box<dyn Future>::poll` indirection frame rather than shrinking `run_agent_loop`'s own poll frame. If the 8 MiB stack still aborts, the 56 816-byte figure is the starting point for deciding what to shrink.
- **The desktop embedded server was added to the stack-sizing change** (the plan listed only the daemon). It boots the same kernel, runs the same background agents and cron sweeps, and so stacks the same agent-turn chain on 2 MiB workers. Leaving it out would have fixed one of two identical sites.
- **The API-side workflow driver was left un-wrapped**, and this is the one place I would most like a reviewer's opinion. `routes/workflows/workflow.rs::spawn_background_run` drives `execute_run` with its own `send_message` closure — a second copy of the step-dispatch semantics — and it does *not* enter the depth scope. I wrapped it, found that `librefang-runtime` is only a **dev-dependency** of `librefang-api`, and reverted rather than promote it to a real dependency without being asked. It is safe today because that driver always starts from a fresh `tokio::spawn`, so the depth task-local is unset and its steps run at depth 0, and a step target that nests another workflow still goes through `LibreFangKernel::run_workflow`, which does check and increment. The only cost is that steps on that path get a budget of 5 instead of 4. The revert left an explanatory comment at the site stating the invariant it relies on. If a maintainer would rather have the accounting uniform, promoting the dev-dependency is a one-line change and I will make it.
- **The TUI runtime (`librefang-cli/src/tui/event.rs`) was deliberately not changed**, per the plan, and the reasoning checks out: `block_on_tui` polls on the calling thread, which for the TUI is the process main thread with the OS-provided ~8 MiB, so the deepest nesting there is already on a large stack.
- **The error-shape fix was not in the plan.** It was found while verifying the plan's own requirement that the refusal be a policy error rather than an opaque upstream one. Without it the cap still bounded the recursion but reported it as a 5xx-class crash, so the refusal was misclassified on the only surface an agent sees. Fixing it stayed inside the two crates already being edited.
- **Rebased onto a `docs: rewrap new PR comments to one-sentence-per-line` commit** that another session pushed onto this branch while the error-shape fix was being verified. Their rewrap is preserved untouched, and the newer comments were rewrapped the same way in a follow-up so the branch is internally consistent.

## Behaviour changes a reviewer must weigh

- **A workflow chain nested past `max_agent_call_depth` now fails where it previously recursed.** On the surface that can actually reach it — an agent calling `workflow_run` — the failure is `ToolError::PermissionDenied`, a soft `Denied` whose text names the configured cap and points at `workflow_start`. Any deployment that legitimately nests workflows more than 5 agent turns deep will start seeing refusals and needs to raise `max_agent_call_depth`. Note that "the previous behaviour" for such a chain was to keep recursing until the wall-clock timeout or the stack ran out, so nothing that was reliably working before starts failing.
- **On the HTTP surface the refusal is unreachable, so no status-code change was made.** An earlier draft of this body claimed 403; that was wrong and is corrected here. `POST /api/workflows/{id}/run?wait=true` blanket-maps every `run_workflow_typed` error to **422** `{"error":"workflow_failed"}` — pre-existing, untouched by this PR. It does not matter in practice, and this is checkable rather than assumed: after this change `AGENT_CALL_DEPTH.scope` is entered in exactly one production place, inside `with_agent_call_depth` (`grep -rn AGENT_CALL_DEPTH crates/` returns the task-local declaration, the reader in `dispatch.rs`, that one `.scope`, two doc-comment mentions, and two test sites — nothing else). Its only callers are `tool_agent_send` and the kernel's workflow step dispatch, so no HTTP request path can carry a non-zero depth and the cap cannot fire there. The default async path additionally spawns a fresh task and never calls the checked entry point at all. I did not add a 403 arm, because doing so would change an existing endpoint's error contract (and require updating its `#[utoipa::path]`, which currently documents only 200/202) for a case that cannot occur. Flagging it rather than deciding silently.
- **Workflow steps now consume the same depth budget as `agent_send` hops.** An agent one `agent_send` hop deep that runs a workflow gets 4 remaining levels, not 5. This is the intended semantics — both stack real agent turns — but it is a tightening, not a no-op.
- **Two more threads' worth of 8 MiB stack reservation per runtime.** Thread stacks are lazily committed on Linux and macOS, so this is address-space reservation rather than resident memory, but it is visible in VSZ.

---

## Update — review findings addressed

**Half-change fixed.** `triggers_and_workflow.rs` has two step-dispatch closures, not one: `run_workflow::send_message` and the twin inside `KernelOperatorResumeDriver::drive_operator_timeout`. Both acquire the same per-agent semaphore and both run a whole agent turn inline, but only the first entered the depth scope. An operator-timeout resume drives the run's remaining steps through the second, so a timed-out HITL workflow ran its step agents at depth 0 — a budget of one extra stacked turn compared to every other path, and without the boxing that keeps each level to a pointer.

**The sentinel could not have caught it.** It was a single `contains` for `with_agent_call_depth(self.send_message_full(`, and the twin calls through `k.`, so it would have passed even against the fix. It now counts occurrences — the way the neighbouring semaphore sentinel at `tests.rs:6448` already does, and for the same reason — and separately asserts that exactly one `send_message_full` call stays outside a depth scope: the trigger dispatch path, which *starts* a chain on its own task rather than nesting inside one, so depth 0 is correct there. A newly added inline dispatch now fails the second assertion rather than passing unnoticed.

Mutation check: unwrapping the resume twin again fails the sentinel with `expected BOTH step-dispatch closures … but found 1`. Restored and re-confirmed green.

### Behaviour at `max_agent_call_depth = 0`, previously undeclared

The guard is `current_depth >= max_depth` with no lower clamp, so at the legal value `0` a nested `workflow_run` is refused outright. That is deliberate and now documented at the site: `tool_agent_send` uses the identical comparison (`tool_runner/agent.rs`), so `0` has always meant "permit no nesting" for `agent_send`. Before this PR `workflow_run` ignored the knob entirely, so an operator running `0` could nest workflows while `agent_send` was blocked — that inconsistency is what goes away. Clamping here would have re-created it in the other direction and defeated the point of sharing one knob.

### Verification of this update

- `cargo test -p librefang-kernel --lib agent_call_depth` — 2 passed (`workflow_step_dispatch_enters_agent_call_depth_scope`, `nested_workflow_run_past_max_agent_call_depth_is_capability_denied`).
- `cargo test -p librefang-kernel --lib workflow_send_message` — 2 passed.
- `cargo clippy --workspace --all-targets -- -D warnings` — exit 0.
- `cargo fmt --check` — exit 0.


crates/librefang-runtime/src/tool_runner/mod.rs
```diff
@@ -132,6 +132,29 @@ tokio::task_local! {
...
+///
+/// Before this existed, `agent_send` was the only path that accounted for nesting; the workflow-step dispatch in the kernel nested turns with no accounting at all, so an agent whose workflow step targets an agent that re-runs the same workflow recursed with no depth bound whatsoever (refs #6659).
+/// Callers outside `tool_runner` — the kernel's workflow drivers — need this because `AGENT_CALL_DEPTH` is module-private.
+///
+/// The check itself stays with the caller: each entry point already knows how to shape its own refusal (`ToolError::PermissionDenied` for a tool, `LibreFangError::CapabilityDenied` for a kernel op), and both read the same `max_agent_call_depth` knob.
+///
+/// `fut` is boxed for the same reason [`crate::held_agent_locks::scope`] boxes its argument: the future being wrapped here is an entire agent turn, an enormous state machine.
+/// Inlining it would stack another full copy of it per nesting level, adding stack cost at exactly the boundary this helper exists to bound.
+/// Boxing makes each level cost a pointer.
+pub async fn with_agent_call_depth<F>(fut: F) -> F::Output
```
houko, line 147: Commit `9c432e6` on this branch ("docs: rewrap new PR comments to one-sentence-per-line") has author identity `Claude <noreply@anthropic.com>`.

CLAUDE.md's "No AI / Claude attribution" rule and the repo's `commit-msg` hook both cover this: the hook rejects a commit whose *author identity* resolves to Claude/Anthropic, not just message text containing it. This commit's message body is clean, but the author field itself is the violation.

The PR body already acknowledges this ("Rebased onto a `docs: rewrap...` commit that another session pushed onto this branch"), so it's a known artifact of an AI session pushing directly rather than through the PR author's own tooling.

Not fixing this myself: removing it needs a history rewrite (rebase + force-push) on `houko`'s branch, which CLAUDE.md/AGENTS.md reserve for the branch owner ("don't force-push to someone else's branch"). Flagging so the author can squash/rebase it out (e.g. `git rebase -i` dropping/reauthoring that commit) before merge.

---

crates/librefang-kernel/src/kernel/tests.rs
houko, line 1: CLAUDE.md, "No AI / Claude attribution": two commits on this branch carry the git author identity `Claude <noreply@anthropic.com>`, not just AI-attributed text in a message:

- `9c432e6fa6f7036549ab9ba88d9b6a75a4e2f427` — "docs: rewrap new PR comments to one-sentence-per-line" (touches this file plus `triggers_and_workflow.rs`, `workflow.rs` (api + tool_runner), `daemon.rs`, `server.rs`, `agent_loop/tests/integration.rs`, `tool_runner/agent.rs`, `tool_runner/mod.rs`, `tool_runner/tests/mod.rs`)
- `19f130f2a19da19631cd14e5d1017521a910391c` — "style: rewrap remaining two-sentence comment lines to one-sentence-per-line" (touches this file and `triggers_and_workflow.rs`)

CLAUDE.md is explicit that this is checked at the commit level, not just the message text: "`commit-msg` — ... separately rejects a commit whose *author identity* (`git var GIT_AUTHOR_IDENT`, so the `GIT_AUTHOR_NAME` / `GIT_AUTHOR_EMAIL` env overrides are covered) resolves to Claude / Anthropic even when the message itself is clean." Both commit messages here are clean of any AI phrasing, but the author identity itself is `Claude <noreply@anthropic.com>` — the exact shape the hook is meant to catch, and the same failure mode called out for PR #6751 (container-wide git author default landing an automated commit under that identity). Since `core.hooksPath` enforcement only runs where it's configured, this slipped through on whatever path pushed these two commits.

Rewriting history on a branch already up for review is its own tradeoff, so this isn't a suggestion block — flagging for the maintainer to decide: rebase with the author identity corrected (`git commit --amend --reset-author` / `git rebase -i` with `exec git commit --amend --reset-author`, force-push before further review lands), or squash-merge so the final `main` history only carries the human-authored commit identity regardless of these two.

---

houko (commented):
Automated daily review against CLAUDE.md's stated rules only (changelog fragment format, AI/Claude attribution, BTreeMap/HashMap prompt-determinism, config-field completeness, workspace-wide cargo commands, prose-wrapping). One finding posted inline. Everything else checked clean — see summary in the review comment.

---

crates/librefang-kernel/src/kernel/triggers_and_workflow.rs
```diff
@@ -744,6 +744,30 @@ impl LibreFangKernel {
...
+        //
+        // `CapabilityDenied` rather than `Internal`, matching the precedent in `tool_agent_send`: this is a self-imposed kernel-policy quota, so a capped chain records a policy refusal (HTTP 403) on the step instead of an opaque 5xx that would read as a downstream crash to retry logic.
+        // Checked before `create_run` so a refusal leaves no orphan `Pending` run behind.
+        // `>=` with no lower clamp, matching `tool_agent_send` byte for byte (`tool_runner/agent.rs`).
+        // At the legal value `max_agent_call_depth = 0` that refuses a nested `workflow_run` outright, which is the same thing `agent_send` has always done at 0 — the knob means "permit no nesting", and honouring it on only one of the two paths it now governs would defeat the point of sharing it.
+        // Before this change `workflow_run` ignored the knob entirely, so an operator running 0 could nest workflows while `agent_send` was blocked; that inconsistency is what goes away here.
+        let max_depth = cfg.max_agent_call_depth;
+        let current_depth = librefang_runtime::tool_runner::current_agent_depth();
```
houko, line 1253: Four commits on this branch carry an AI-attributed author identity, `Claude <noreply@anthropic.com>`, including the current branch head:
- `017e902` — "style: rewrap two remaining two-sentence comment lines" (this file + `kernel/tests.rs`)
- `dabbfe0` — "fix(cli): size the TUI's in-process runtime and stream thread stacks"
- `19f130f` — "style: rewrap remaining two-sentence comment lines to one-sentence-per-line" (this file + `kernel/tests.rs`)
- `9c432e6` — "docs: rewrap new PR comments to one-sentence-per-line" (touches this file plus 8 others in the diff)

CLAUDE.md is explicit that this is forbidden, and says the repo's own `commit-msg` hook is supposed to catch it server-side: "rejects a commit whose author identity (`git var GIT_AUTHOR_IDENT` …) resolves to Claude / Anthropic even when the message itself is clean." None of these four commit messages mention Claude/Anthropic — only the author identity does — so this is exactly the gap that hook exists to close, and it's landed on the PR branch anyway.

None of the commit *messages* are otherwise a problem (conventional-commit subjects, clean bodies), and the resulting file content is fine — these commits' own job was fixing prose-wrap violations, and the current diff is correctly one-sentence-per-line as a result. So the fix here is narrowly about authorship, not content: the four commits need their author identity rewritten to the actual human maintainer (interactive rebase + `--reset-author` or equivalent) and force-pushed onto this branch.

Flagging rather than fixing: rewriting already-pushed history and force-pushing is outside what I can do here (this review pass never force-pushes, even to fix something this concrete). A maintainer needs to do the rebase and force-push themselves.

---

houko (commented):
Reviewed the diff against CLAUDE.md's contributor rules (changelog fragment format/attribution, config-field Default pairing, conventional-commit format, one-sentence-per-line prose wrap in touched comments/docs, and the relevant architecture Gotchas). The code changes themselves are compliant: the changelog fragment in `changelog.d/fixed/` is correctly formatted and attributed, `max_agent_call_depth` is a pre-existing config field (no Default-impl pairing needed), and every comment/doc-comment block added by this PR is already wrapped one sentence per line. One finding, left inline: four commits on this branch (including the current head) carry an AI-attributed git author identity, which CLAUDE.md explicitly forbids and which its `commit-msg` hook is meant to reject. No code fix applied for that finding since correcting it requires rewriting pushed history and force-pushing, which is outside what this review pass does.

---

crates/librefang-cli/src/commands/daemon.rs
```diff
@@ -459,6 +459,11 @@ pub(crate) fn cmd_start(config: Option<PathBuf>, tail: bool, spawned: bool, fore
 
     let rt = tokio::runtime::Builder::new_multi_thread()
         .worker_threads(2)
+        // Raise the worker/blocking thread stack from tokio's 2 MiB default (refs #6659).
+        // An agent turn is a deep chain of very large futures — `send_message_full_inner` -> `execute_llm_agent` -> `run_agent_loop` — and a tool call that dispatches to another agent (`agent_send`, a workflow step) runs the callee's entire turn inline on the same worker, stacking that chain again.
+        // Overflowing it SIGABRTs the whole process: with only 2 workers a single deep turn takes the HTTP API and every cron job down with it.
+        // The depth cap on the workflow path is the real bound; this is headroom so a legal-depth chain of fat frames does not abort before the cap can refuse it.
+        .thread_stack_size(8 * 1024 * 1024)
```
houko, line 466: Minor, not blocking: `8 * 1024 * 1024` is repeated as a bare literal in five places across four files (here, `librefang-desktop/src/server.rs` x2, `librefang-cli/src/tui/event.rs` x2), all justified by the same rationale comment copy-pasted at each site. Worth a single shared `pub const AGENT_TURN_STACK_SIZE: usize = 8 * 1024 * 1024;` (e.g. next to `with_agent_call_depth` in `librefang-runtime::tool_runner`, which all four already reference in their comments) so a future re-tuning of the headroom is a one-line change instead of a grep-and-replace across crates, and so the five comments can point at one definition instead of restating the same reasoning. Leaving as a suggestion rather than editing: picking the right home for the constant crosses `librefang-cli`/`librefang-desktop`/`librefang-runtime` crate boundaries, and per CLAUDE.md's own review rules that's a call for the maintainer rather than something to unilaterally restructure in an automated pass. Separately: this is unrelated to the value 8 MiB itself (which the PR body already treats as an open question pending the next #6659 occurrence), only to the duplication of the literal.

---

houko (commented):
Reviewed against CLAUDE.md's stated rules (changelog fragment format, config-field/Default pairing, AI attribution, recursion-bound soundness, regression-test coverage). The depth-cap fix and its policy-error plumbing hold up: `max_agent_call_depth` is a pre-existing, already-defaulted config field reused (not a new hardcoded magic number), the workflow-nesting refusal surfaces as `CapabilityDenied` → `PermissionDenied` rather than a panic or opaque 5xx, and `nested_workflow_run_past_max_agent_call_depth_is_capability_denied` plus the source-shape sentinel actually drive both step-dispatch closures past the old unbounded-recursion point. Changelog fragment in `changelog.d/fixed/` is present and correctly formatted/attributed. No mechanical fix applied — nothing met the ≥90%-confidence bar for a direct edit.

Two things flagged, neither actioned directly:
1. One new nit (inline, non-blocking): the `8 * 1024 * 1024` stack-size literal is duplicated across 5 sites in 4 files with the rationale copy-pasted at each; suggest a single shared named constant. Crosses `librefang-cli`/`librefang-desktop`/`librefang-runtime` boundaries, so left as a suggestion rather than restructured unilaterally.
2. Not re-flagging, only noting it's still open: three earlier unresolved review threads on this exact head commit (`017e902b`) already cover the AI-attributed git author identity (`Claude <noreply@anthropic.com>`) on 4 commits in this branch's history, which CLAUDE.md forbids. Nothing has changed on the branch since those were posted, so repeating that finding here would be redundant — it still needs a maintainer-driven rebase + force-push, which is outside what this pass does.

---

crates/librefang-cli/src/commands/daemon.rs
houko, line 466: Stack-sizing scope check: `Builder::thread_stack_size` on a `new_multi_thread()` runtime sizes both the named worker threads *and* the `spawn_blocking` pool — confirmed in the vendored tokio 1.53.1 source (`runtime/blocking/pool.rs` reads `builder.thread_stack_size` into the blocking-pool's own `stack_size` config, the same field this call sets). The comment above only reasons about the 2 named workers ("with only 2 workers a single deep turn takes the HTTP API and every cron job down with it"), but `spawn_blocking` defaults to up to 512 threads (`max_blocking_threads`), so this also quadruples the reserved stack size for every blocking-pool thread (file I/O, sync DB calls, etc.) — up to ~4 GiB of reserved (not necessarily committed, but reserved) address space at the pool's ceiling, vs. ~1 GiB before.

That's very likely fine on 64-bit address space and probably intentional/harmless, but it doesn't appear to have been considered in the comment or the changelog fragment, and the same pattern repeats in `librefang-desktop/src/server.rs` and the TUI runtime in `librefang-cli/src/tui/event.rs`. Worth a one-line acknowledgment (or a deliberate call to scope this more narrowly, e.g. by building blocking-pool-sensitive workloads differently) so a future reader doesn't have to rediscover that this knob is not worker-thread-only.

---

crates/librefang-kernel/src/kernel/triggers_and_workflow.rs
```diff
@@ -804,7 +828,10 @@ impl LibreFangKernel {
                         ))
                     }
                 };
-                self.send_message_full(
+                // Account for the nesting (refs #6659).
+                // A workflow step runs a whole agent turn inline on this task, so it belongs one level deeper in the same chain `agent_send` tracks — that is what makes the `max_agent_call_depth` check at the top of `run_workflow` bound a workflow that re-runs itself through a step target's `workflow_run`.
+                // The helper also boxes the turn's future, so each nesting level costs a pointer rather than another inlined copy of the agent-loop state machine.
+                librefang_runtime::tool_runner::with_agent_call_depth(self.send_message_full(
```
houko, line 1327: Concurrency-interaction question: a few lines above this (`self.agent_concurrency_for(agent_id)`), the step dispatch still acquires the per-agent `tokio::sync::Semaphore` — capped at 1 for any `Persistent`-mode or default-configured agent, see `accessors.rs::agent_concurrency_for` — *before* this `with_agent_call_depth`-wrapped `send_message_full` call, exactly as it did before this PR.

For a workflow whose step targets the *same* agent that is already running the outer turn (directly self-referential, or indirectly via `A`'s workflow having a step that targets `A` again), the second-level dispatch tries to acquire that same semaphore while the first-level dispatch is still holding its permit (blocked awaiting the second level's completion) — a classic self-deadlock. That happens at the very first re-entry (depth 0→1), well before `current_agent_depth() >= max_agent_call_depth` (default 5, checked earlier in this function) is ever reached, so the new depth cap added in this PR does not actually close the "unbounded recursion" hole for this case — it only helps once the recursion routes through *different* agents each hop, each of whose semaphores has a free permit.

Worth confirming whether this is a known, accepted pre-existing limitation of `agent_concurrency_for` (it predates this PR, per the "audit fix for `triggers_and_workflow.rs:334-336`" comment above), or whether it should be called out / guarded here — as written, the fix note and changelog fragment read as if `max_agent_call_depth` now fully bounds workflow self-recursion, but a same-agent cycle turns what used to be a stack-overflow crash into a silent permanent hang instead, which is arguably worse operationally (no crash to restart from, no log line, just a stuck run).

---

houko (commented):
Automated daily review of #6743. Checked against CLAUDE.md: changelog fragment, config-field completeness, regression test coverage, commit format, and AI attribution are all clean — no mechanical fixes were needed, so nothing was pushed. Left two inline comments on concurrency/stack-sizing points that need a maintainer judgment call rather than a mechanical fix.

---

crates/librefang-kernel/src/kernel/triggers_and_workflow.rs
```diff
@@ -744,6 +744,30 @@ impl LibreFangKernel {
...
+        // At the legal value `max_agent_call_depth = 0` that refuses a nested `workflow_run` outright, which is the same thing `agent_send` has always done at 0 — the knob means "permit no nesting", and honouring it on only one of the two paths it now governs would defeat the point of sharing it.
+        // Before this change `workflow_run` ignored the knob entirely, so an operator running 0 could nest workflows while `agent_send` was blocked; that inconsistency is what goes away here.
+        let max_depth = cfg.max_agent_call_depth;
+        let current_depth = librefang_runtime::tool_runner::current_agent_depth();
+        if current_depth >= max_depth {
+            return Err(KernelError::LibreFang(LibreFangError::CapabilityDenied(
+                format!(
+                    "Nested workflow run depth exceeded (max {max_depth}); this run is already \
+                     {current_depth} agent turns deep. Start the workflow asynchronously \
+                     (workflow_start) instead of nesting it inside the current turn."
```
houko, line 1259: Coverage gap: `workflow_start` (the async/tracked path) never goes through this check at all, and the refusal message above literally tells a capped caller to switch to it.

`tool_workflow_start` calls `kh.start_workflow_async_tracked(...)`, which lives in `crates/librefang-kernel/src/kernel/handles/workflow_runner.rs` (`start_workflow_async_tracked`, the `tokio::spawn` block around line 353). That function drives the run via `kernel_arc.workflows.engine.execute_run(run_id, resolver, send_message)` directly — it does not call through `LibreFangKernel::run_workflow`, so the depth check added in this PR never executes on that path. Separately, because the whole thing runs inside a fresh `tokio::spawn`, `current_agent_depth()` would read 0 there even if the check were reachable, since task-locals don't cross a spawn boundary — so depth would have to be captured from the calling task *before* spawning and threaded through explicitly, the way `spawn_background_run` in `crates/librefang-api/src/routes/workflows/workflow.rs` reasons about (and documents) for its own depth-0 case.

That's likely fine for the specific stack-overflow crash this PR targets (a fresh task can't restack the caller's future), but it does mean the "one operator knob caps both" framing in the changelog isn't quite accurate: an agent whose workflow step targets an agent that calls `workflow_start` on the same workflow again can still recurse with no depth bound at all, just without a crash — unbounded async task fan-out instead of a stack overflow. And the refusal text this PR adds (`"Start the workflow asynchronously (workflow_start) instead of nesting it inside the current turn."`) is exactly the escape hatch that reintroduces that unbounded recursion for a capped caller.

Worth a decision from a maintainer rather than a drive-by fix here: either (a) accept the async path as intentionally exempt because it can't crash the way the sync path did, and adjust the changelog/refusal wording to stop implying it's bounded too, or (b) thread the caller's depth into `start_workflow_async_tracked` and enforce the same cap there (harder to report back to the caller cleanly, since `workflow_start` already returns a `run_id` before the run executes).

---

houko (commented):
Automated daily review pass. Checked the diff against CLAUDE.md (config-field checklist n/a — no new config field; changelog.d fragment format compliant; prose one-sentence-per-line compliant) and for correctness independent of CLAUDE.md (depth-cap enforcement across recursive/nested-spawn paths, graceful-failure shape, cross-platform stack sizing). One new finding left inline. Also noted but not re-flagged: this branch has four commits (including the current head) with git author identity `Claude <noreply@anthropic.com>`, already reported in three separate review threads on this PR by earlier runs of this same routine — not repeating that here, and not force-pushing to fix it myself per policy.

---

houko:
Disclosure: this automated review run's own fix commit `52f1441c` (doc update) was mistakenly authored under this review environment's default git identity, `Claude <noreply@anthropic.com>` — a violation of this repo's own no-AI-attribution policy, the same class of issue already flagged multiple times elsewhere on this thread (for other, pre-existing commits). Root cause: the environment's global git config defaulted to that identity; this has been corrected mid-run (to the authenticated `houko` identity) so further commits from this routine don't repeat it. Recommend squashing/amending `52f1441c` under a real author before merge, along with the other previously-flagged commits on this branch.

---

houko (commented):
Automated review pass (content + CLAUDE.md compliance check).

**Content: solid, no correctness issues found.** The depth-cap fix is well-tested — `nested_workflow_run_past_max_agent_call_depth_is_capability_denied` proves the refusal fires at the configured cap, doesn't create an orphan `Pending` run, and survives the `WorkflowRunner` trait hop as `CapabilityDenied` rather than getting flattened to `Internal`; the `workflow_step_dispatch_enters_agent_call_depth_scope` source-shape sentinel pins that *both* step-dispatch closures (including the operator-resume twin) are wrapped, not just one; and `with_agent_call_depth_adds_exactly_one_level_per_nesting` / `agent_send_quota_sees_depth_established_by_with_agent_call_depth` confirm the shared counter semantics. That's a real behavioral test of the bound, not just documentation of intent.

Also checked: the per-agent semaphore in `triggers_and_workflow.rs`'s `send_message` closure is acquired *before* the new `with_agent_call_depth` wrap and held across it (permit drops at the end of the same async block), so this doesn't bypass or double-apply the trigger-dispatch concurrency caps described in CLAUDE.md. `.response` (not `.response_text`) is used correctly throughout. Changelog fragment is present and correctly formatted. `cargo fmt --check` is clean on every touched crate (kernel, runtime, cli, desktop, api).

**Process flag, not a code issue:** several commits on this branch carry the git *author identity* `Claude <noreply@anthropic.com>` (e.g. `b3d3986`, `bed1478`, `254c1a4`, `bd4f108`, `1d03414`, `dabbfe0`, `52f1441`), even though the commit message bodies themselves are clean of any AI mention. CLAUDE.md's `commit-msg` hook is explicitly designed to catch this ("rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean"), so these commits would not pass that hook as configured — suggesting they were pushed from an environment where `core.hooksPath` wasn't set to `scripts/hooks`, or hooks were bypassed. I'm not rewriting history on an open PR unilaterally (that needs a force-push and risks clobbering review state), so flagging for a maintainer to rebase with a corrected author identity (or squash-merge) before landing.

---

crates/librefang-cli/src/commands/daemon.rs
houko, line 466: The 8 MiB figure is justified as "headroom" in the comment, but I don't see it tied back to an actual number: `max_agent_call_depth` defaults to 5, and `run_agent_loop_future_size_stays_within_its_order_of_magnitude` (added in this PR, `crates/librefang-runtime/src/agent_loop/tests/integration.rs`) pins the future at under 512 KiB with a note that the #6659 crash worked out to ~40 KB/frame.
Neither of those numbers is multiplied out anywhere against the new stack size — 8 MiB just reads as "4x the tokio default," a round number rather than a derived bound.

Given the new future-size regression test already tracks the per-level cost, it would strengthen the fix to either:
- add a comment here (or in that test) doing the arithmetic: `max_agent_call_depth (5) × <per-level worst case> + other stack usage ≪ 8 MiB`, so a future change to the depth default or the future-size bound has something concrete to check against, or
- explicitly state that 8 MiB was chosen empirically/by margin rather than computed, so reviewers don't need to reverse-engineer that.

Not blocking — the change is a strict improvement over the 2 MiB default either way — but as written, "not an arbitrary round number" isn't demonstrated for this specific value.

---

crates/librefang-api/src/routes/workflows/workflow.rs
```diff
@@ -529,6 +529,9 @@ fn spawn_background_run(state: Arc<AppState>, run_id: WorkflowRunId) {
                       session_mode_override: Option<librefang_types::agent::SessionMode>| {
                     let sc = state_for_sender.clone();
                     async move {
+                        // NOTE (refs #6659): unlike the kernel's own workflow driver, this one does not enter `tool_runner::with_agent_call_depth` around the step's agent turn — `librefang-runtime` is only a dev-dependency of this crate.
+                        // It is safe today because `spawn_background_run` always starts from a fresh `tokio::spawn`, so the depth task-local is unset and every step here runs at depth 0; a step target that nests another workflow still goes through `LibreFangKernel::run_workflow`, which does check and increment.
```
houko, line 533: The safety argument here (`spawn_background_run` always starts from a fresh `tokio::spawn`, so the depth task-local is unset) is correct today, but it's an invariant about *how this function is called*, not something the function itself enforces — and unlike the two step-dispatch closures in `triggers_and_workflow.rs`, which get a counted sentinel test (`workflow_step_dispatch_enters_agent_call_depth_scope`) that fails if a new unwrapped `send_message_full` call shows up, this path has only the comment.

If a future change ever routes this driver through a caller that's already inside an agent turn (the comment even flags this as the risk), nothing here catches the regression — the depth counter would just silently read 0 and the nesting cap this PR adds would stop applying to that path.

Worth considering a lightweight guard, e.g. a debug_assert!(current_agent_depth() == 0, ...) at the top of the spawned closure, or at least a similar source-shape sentinel test asserting this file has no `with_agent_call_depth` wrapping (so the day someone adds one, they're prompted to reconsider why). Not blocking given `librefang-runtime` is only a dev-dependency of this crate, which is the real reason no compile-time enforcement is available here — but the current safety net for this specific gap is weaker than the two gaps this PR closed in the kernel.

---

houko (commented):
Automated review pass on the nested-workflow-depth + worker-stack-size fix.

Verified: the depth bound is enforced at every same-task recursive entry point I could find — `agent_send` (pre-existing) and now `workflow_run`'s step dispatch, via the new shared `with_agent_call_depth` helper, checked in `LibreFangKernel::run_workflow` before `create_run` (no orphan `Pending` runs on refusal). Both the operator-resume twin dispatch and the primary one are wrapped (confirmed by the counted sentinel test, which is a good defense against exactly the kind of "missed one call site" bug this class of fix invites). The refusal is a clean `CapabilityDenied` → `PermissionDenied` → soft `Denied` all the way to the tool layer, not a panic or a 5xx that would read as a downstream crash — confirmed against the existing `agent_send` precedent and the new `workflow_run_depth_refusal_is_permission_denied_not_upstream` test. The changelog fragment is present, correctly located, and ends `(#6743) (@houko)`.

Two non-blocking findings, both left as comments per review policy (design/judgment calls, not clear bugs):
1. `daemon.rs` — the 8 MiB worker stack size is described as "not an arbitrary round number" but the PR never multiplies out `max_agent_call_depth × per-level frame cost` against it; the new future-size regression test has the numbers to do that arithmetic explicitly.
2. `workflow.rs::spawn_background_run` — intentionally skips the depth scope based on a "always called via fresh `tokio::spawn`" invariant that has no test/assertion enforcing it, unlike the two kernel dispatch sites this PR did add a sentinel for.

No fixes applied — nothing met the ≥90%-confidence mechanical bar, and both findings above are control-flow/design judgment calls. Build verification via `cargo check -p librefang-runtime --lib` was started but did not complete in a reasonable window due to heavy concurrent `cargo` contention on the shared host; the diff was cross-checked manually against call sites, trait/type definitions (`KernelOpError` = `LibreFangError` alias, `ToolExecutionStatus::Denied` classification, `held_agent_locks::scope` boxing precedent) for signature and behavior consistency.

---

crates/librefang-cli/src/commands/daemon.rs
houko, line 466: Two things worth a maintainer decision rather than a mechanical fix:

1. **Sufficiency of the bound.** The PR's own measurement (`run_agent_loop_future_size_stays_within_its_order_of_magnitude`) only sizes one function in the per-level chain — `send_message_full_inner` and `execute_llm_agent` are called out in the PR body as not measured because they're private to `librefang-kernel` and need a booted kernel to construct. 8 MiB / `max_agent_call_depth` default 5 gives roughly 1.6 MiB per level of headroom, but the actual per-level cost is the full chain plus whatever LLM-driver and tool-execution frames sit on top of `run_agent_loop`, not just the 56 816 bytes measured. Worth asking whether that arithmetic was actually checked against a real nested-turn stack profile, or is being treated as "big enough headroom" on the strength of one partial data point.
2. **Magic number duplicated 5x with no shared constant** (here, `librefang-desktop/src/server.rs` x2, `librefang-cli/src/tui/event.rs` x2) and not exposed as a config knob, unlike most other size-related caps in `KernelConfig` (`max_upload_size_bytes`, `max_request_body_bytes`, etc. — see `docs/operations/config-reload.md`). If a future incident needs a different value, this requires a code change + rebuild across four sites instead of a config edit. A `pub const AGENT_TURN_WORKER_STACK_BYTES: usize = 8 * 1024 * 1024;` shared between the four call sites would at least remove the duplication; whether it should also be a `KernelConfig` field is a bigger design question worth raising explicitly rather than deciding silently.

---

crates/librefang-api/src/routes/workflows/workflow.rs
```diff
@@ -529,6 +529,9 @@ fn spawn_background_run(state: Arc<AppState>, run_id: WorkflowRunId) {
...
+                        // If this driver ever gains a caller that is already inside an agent turn, its steps must be wrapped the same way the kernel's are.
```
houko, line 534: Flagging this for maintainer sign-off since the PR body itself says this is "the one place I would most like a reviewer's opinion" — this needs a human decision, not an automated fix.

The invariant this comment documents ("every step here runs at depth 0 because `spawn_background_run` always starts from a fresh `tokio::spawn`") is correct today, but it's an invariant enforced only by convention across two crates, not by a type or a runtime assertion. If a future change gives this driver a caller that is already inside an agent turn (e.g. a workflow step that itself triggers `POST /api/workflows/{id}/run` on another workflow via an HTTP tool, or some future in-process shortcut that skips the `tool_workflow_run`/`kh.run_workflow` path), the depth accounting silently under-counts and the whole point of this PR — bounding nested-turn stack growth — has a gap on this one entry point. Since `librefang-runtime` is only a dev-dependency of `librefang-api`, promoting it to a real dependency to close this (as the PR body offers to do) is a one-line change; worth deciding now whether accounting uniformity is worth that dependency promotion, versus leaving it as a documented invariant that the next person touching this function has to notice and preserve.

---

houko (commented):
Automated daily review pass. Checked against CLAUDE.md's condensed rule set (config-field defaults, changelog fragments, prose wrapping, determinism/BTreeMap, test coverage, AI attribution). No mechanical, high-confidence violations found — the changelog fragment (`changelog.d/fixed/6743-nested-workflow-turn-depth.md`) is present and correctly formatted, the reused `max_agent_call_depth` config field needed no new Default-impl wiring since it's not a new field, its `docs/operations/config-reload.md` classification (`N`/live-read) is unaffected since neither read site changed, and the new behavior has solid regression coverage (behavioral + source-shape sentinel + unit tests on the shared depth helper). Left two inline comments on points that need a maintainer's judgment call rather than a bot fix: the stack-size bound's sufficiency/duplication, and the intentionally-unwrapped API-side workflow driver the PR body already flagged for review.

---

Issue #6624, opened 2026-07-28
browser_read_page: 50k cap reached on ordinary pages, and markdown link syntax inflates link-dense pages ~3.5x


nevgenov:
Two measurements about `EXTRACT_CONTENT_JS`, the script behind `browser_navigate` / `browser_read_page`. Neither is a bug — both are design trade-offs — but the numbers surprised me enough that they seem worth a maintainer's opinion before anyone patches either.

All figures below are from live runs against Chrome for Testing 148.0.7778.96 and Lightpanda `c1fc2b13` on macOS 15 / aarch64.

## 1. The 50,000-character cap is reached on ordinary pages

| page | `document.body.innerText` | extracted | outcome |
|---|---|---|---|
| `en.wikipedia.org/wiki/Rust_(programming_language)` | 73,447 | **50,016** | truncated |
| `news.ycombinator.com` | 4,267 | 14,782 | not truncated |
| `example.com` | 167 | 167 | not truncated |

A mainstream Wikipedia article overruns the cap, so a single `browser_navigate` can spend roughly 12–15k tokens of context *and still lose the tail of the page*. The agent gets no signal about what it is missing beyond a trailing `... (truncated)`.

The cap is currently a fixed compile-time value, not operator-tunable. Whether it should be configurable, or whether truncation should be smarter than a hard cut at N characters, is a product call rather than something to guess at.

(#6623 makes the constant actually control the cap — today the literal is hard-coded in the JS and `MAX_CONTENT_CHARS` is dead code — but does not change the value or the policy.)

## 2. Markdown link syntax inflates link-dense pages ~3.5×

`walk()` emits every anchor as `[text](href)`. On pages that are mostly links, the URLs dominate the payload:

| page | visible text | extracted | ratio |
|---|---|---|---|
| `news.ycombinator.com` | 4,267 | 14,782 | **346%** |
| `rust-lang.org` | ~3.1k | 3,126 | ~100% |
| `example.com` | 167 | 167 | 100% |

The HN case is the interesting one: the output is 3.5× the text a human sees, and the excess is entirely URLs — many of them long tracking-ish query strings that the model is unlikely to act on directly. Search-result pages, link aggregators, and index pages all have this shape, and they are exactly the pages an agent tends to land on first.

The URLs are obviously useful — that is how the agent knows where to navigate next. The question is whether every anchor needs its full `href` inline, or whether something like eliding long URLs, or dropping hrefs already reachable via `browser_click`'s text matching, would be a better default.

[...]

PR #6708, opened 2026-08-03
fix(api): deliver manually run schedule outputs

Kvitral:
## Type

- [ ] Agent template (TOML)
- [ ] Skill (Python/JS/Prompt)
- [ ] Channel adapter
- [ ] LLM provider
- [ ] Built-in tool
- [x] Bug fix
- [ ] Feature (Rust)
- [ ] Documentation / Translation
- [ ] Refactor / Performance
- [ ] CI / Tooling
- [ ] Other

## Summary

Fixes #6707.

Manual schedule runs completed their agent turn or workflow but did not deliver
the resulting output through the schedule's `delivery` or `delivery_targets`.

The manual API route now uses the same output-delivery operation as timed cron
fires.

## Changes

- Add a shared kernel operation for primary and fan-out cron output delivery.
- Use the shared operation from timed agent and workflow cron execution.
- Deliver successful manual agent and workflow output to configured targets.
- Preserve silent-response behavior for agent turns.
- Add an integration test using a recording Telegram adapter without external services.

## Attribution

- [x] This PR preserves author attribution for any adapted prior work (`Co-authored-by`, commit preservation, or explicit credit in the PR body)

No prior work was adapted.

## Testing

- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [x] `cargo test --workspace` passes
- [x] Live integration tested (if applicable)

Additional verification:

- `cargo test -p librefang-api --test workflows_routes_integration -- --test-threads=1`
  - 67 passed
- `cargo test -p librefang-kernel cron_bridge::tests -- --test-threads=1`
  - 2 passed
- `cargo check --workspace --lib`
- `cargo fmt --all -- --check`
- Reproduced through the dashboard using a local Ollama agent and Telegram target.

The workspace-wide test command was not run. Scoped tests were used to avoid
target-directory contention and excessive file-descriptor usage.

## Security

- [x] No new unsafe code
- [x] No secrets or API keys in diff
- [x] User input validated at boundaries

PR #6745, opened 2026-08-07
fix(browser): keep links inside list items, and cover feed pages instead of one card

nevgenov:
Parts 1 and 2 of the revised step 3 agreed in #6624, sent as one correctness PR per https://github.com/librefang/librefang/issues/6624#issuecomment-5213268857.
Both are independent of any size budget, and neither needs a contract decision.

## 1. `walk()` drops the href of every link inside a list item

The `li` branch pushed `node.textContent` and returned before the child loop, so nested anchors never reached the anchor branch.
The text survived, the destination did not.

| page | anchors in the extraction root | lost their href to `li` |
|---|---|---|
| Wikipedia (Rust) | 1,723 | **1,100** |
| DuckDuckGo | 11 | **11** |
| Hacker News | 197 | 0 |
| Rust book | 23 | 0 |

The model received `- Multi-paradigm: concurrent` where the page has `Multi-paradigm` → `/wiki/Comparison_of_multi-paradigm_programming_languages` and `concurrent` → `/wiki/Concurrent_computing`.
HN keeps its links only because it is built out of tables.

Clicking by text is not a fallback for those: of the 1,100 Wikipedia links that lost their href, **408 do not resolve to themselves** under `cmd_click`'s substring matcher, so there was neither a URL nor a working text handle.

The branch now recurses and folds its children back onto one line, so a bullet still reads as a bullet and the links inside it keep their identity.

## 2. Root selection takes the first `article`, which on a results page is one card

`querySelector('main, article, [role="main"], .content, #content')` returns the first match:

| page | chosen root | root text | pruned page text | coverage |
|---|---|---|---|---|
| DuckDuckGo | `` | 1,109 | 8,101 | **13.7%** |
| Wikipedia | `` | 74,323 | 74,404 | 99.9% |

So `browser_read_page` on a search-results page returned one result and discarded the rest.
That is why DuckDuckGo was the only page whose extracted output was smaller than its visible text, and truncation is not what did it.

Selection now climbs to the ancestor holding repeated sibling `article` elements.
Readability resolves the same shape by climbing to the common ancestor when ≥3 of its top-5 candidates score within 75% of the leader; this narrows that to the case actually hit rather than porting the scorer.
A page carrying a `main` or `[role="main"]` landmark is unaffected, and a single `article` still selects itself.

## Notes

- Both tests are mutation-checked: restoring the flattening `li` branch or the first-match root query turns exactly one of them red.
  Verified locally on this branch.
- Asserted against `EXTRACT_CONTENT_JS_TEMPLATE` rather than against extracted output, because the traversal needs a live DOM — same shape as the existing `test_extract_content_js_varies_with_cap`.
- `2,177` tests pass in `librefang-runtime`, `clippy --all-targets` and `fmt` clean.
- Measured against Chrome for Testing 148.0.7778.96, macOS 26.5.2 / aarch64.
- The link-representation change (markers plus a deduplicated table, as a separate field) follows as a third PR, so this retention win is not held up by the contract change.

Refs #6624

## Review follow-ups

Both code findings were real; thanks.

**Nested lists.** Confirmed by porting `walk()` and tracing `<ul><li>Fruits<ul><li>Apple</li><li>Banana</li></ul></li></ul>` — it produced `- Fruits - Apple - Banana`, exactly as you described.
I went a different way than "prefix only at the outermost `li`", because that gives `- Fruits Apple Banana` and gives up the sub-list structure the recursion just recovered.
A list item now folds only its *own* text and leaves a sub-list's bullets on their own lines with a level of indent, so the example renders as a real nested list:

```
- Fruits
  - Apple
  - Banana
```

Three levels nest correctly too. Whether a produced line came from a nested `li` is tracked as a flag alongside the line rather than sniffed back out of the string, since a text node may legitimately begin with `- `.

**`article` separator.** Added to the block list beside `p`, `div`, `section`, `tr`, for the reason you gave — an entry with no heading of its own emitted no separator, so cards ran together.

**Prose wrapping.** Reflowed to one sentence per line, in the source comments and in this description.

**Changelog fragment.** Now closes `(#6624, #6745)`, so the generated line is suppressed and the issue reference survives.

One note on this PR description: the `&#34;` / `&#39;` entities that appeared in it came from the API edit that removed the footer — they were sitting inside code spans, where GitHub renders them literally rather than decoding them.
Fixed in this update.


changelog.d/fixed/6624-browser-list-item-links.md
```diff
@@ -0,0 +1,7 @@
...
+The extraction script's `li` branch flattened the item to `textContent` and returned before descending, so a nested anchor reached the model as bare text with no URL — 1,100 of 1,723 anchors on the Rust Wikipedia article and 11 of 11 on a DuckDuckGo results page.
+Clicking the text was not a fallback for those: 408 of the 1,100 do not resolve to themselves under `browser_click`'s substring matcher, so there was neither a URL nor a working text handle.
+The branch now recurses and folds its children back onto one line, so a bullet still renders as a bullet and the links inside it keep their identity.
+Root selection used `querySelector('main, article, [role="main"], .content, #content')`, which returns the *first* match — on a page built from sibling `<article>` cards that is one card, measured at 13.7% of a DuckDuckGo results page.
+Selection now climbs to the ancestor holding repeated sibling `article` elements, the way Readability resolves the same shape by walking to the common ancestor of its close-scoring candidates.
+A page with a `main` or `[role="main"]` landmark is unaffected, and a page with a single `article` still selects it (#6624) (@nevgenov)
```
houko: Per `changelog.d/README.md`, the trailing `(#N)` group is what tells the release automation which generated line your prose replaces, and that match is keyed on the PR number, not the tracking issue.
This fragment ends with `(#6624)`, which is the tracking issue, while the PR itself is #6745, so `cargo xtask release` will still emit a generated `- fix(browser): ... (#6745) (@nevgenov)` line alongside this fragment and print a duplicate-bullet warning.
Suggested fix (mirrors the multi-number convention already used in `changelog.d/fixed/6706-desktop-external-links.md`): end the line with `(#6624, #6745) (@nevgenov)` instead of `(#6624) (@nevgenov)`.

I attempted to push this exact one-line fix to `fix/browser-list-links-and-feed-root` but the session's git proxy isn't authorized against `nevgenov/librefang`, so leaving it here instead.

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1562,6 +1585,38 @@ mod tests {
...
+        );
+        assert!(
+            EXTRACT_CONTENT_JS_TEMPLATE.contains("if (tag === 'li')"),
+            "the list-item branch must still exist so bullets keep rendering as bullets"
+        );
+    }
+
+    /// Root selection must not stop at the first `article` when a page has several.
+    ///
+    /// `querySelector('main, article, …')` returned the first match, which on a results page is one card — 13.7% of the page. Readability resolves the same case by climbing to the common ancestor of its close-scoring candidates.
```
houko: CLAUDE.md's prose-wrapping rule ("one sentence = one line, regardless of length") applies to source doc-comments, and this doc-comment line carries two sentences: `... on a results page is one card — 13.7% of the page. Readability resolves the same case by climbing to the common ancestor of its close-scoring candidates.`

Suggested fix — split at the sentence boundary:
```rust
    /// `querySelector('main, article, …')` returned the first match, which on a results page is one card — 13.7% of the page.
    /// Readability resolves the same case by climbing to the common ancestor of its close-scoring candidates.
```

Separately (lower confidence, flagging rather than fixing): the multi-line `//` comments added inside `EXTRACT_CONTENT_JS_TEMPLATE` at the `querySelector`/`li` changes are hard-wrapped at ~90 columns and break mid-sentence (e.g. "...is one / card rather than the list of them — measured at 13.7% of the page..."). The prose rule's "any multi-line free-form prose comment block" clause arguably covers these too, but they're comments embedded in an inlined JS string literal rather than a Rust doc-comment, so it's a judgment call whether the convention was meant to reach that far — worth a maintainer decision either way.

I attempted to push the doc-comment split to `fix/browser-list-links-and-feed-root` but the session's git proxy isn't authorized against `nevgenov/librefang`, so leaving it here instead.

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1312,7 +1312,23 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
     const remove = ['script','style','nav','footer','header','aside','iframe','noscript','svg','canvas'];
     remove.forEach(tag => clone.querySelectorAll(tag).forEach(el => el.remove()));
 
-    let root = clone.querySelector('main, article, [role="main"], .content, #content');
+    // `querySelector` returns the *first* match, which on a feed or a results page is one
+    // card rather than the list of them — measured at 13.7% of the page on a DuckDuckGo
+    // results page. Repeated sibling `article` elements are the signature of that shape, so
+    // climb to the ancestor that holds them, the way Readability resolves the same case by
+    // walking to the common ancestor of its close-scoring candidates.
```
houko: CLAUDE.md's "Prose wrapping: no column limit; break only at sentence boundaries" rule applies to "any multi-line free-form prose comment block", which covers this `//` block. This comment is hard-wrapped at ~90 columns instead of one sentence per line (e.g. "one" / "card" and "DuckDuckGo" / "results page" are split mid-sentence). Please reflow to one sentence per line.

```suggestion
    // `querySelector` returns the *first* match, which on a feed or a results page is one card rather than the list of them — measured at 13.7% of the page on a DuckDuckGo results page.
    // Repeated sibling `article` elements are the signature of that shape, so climb to the ancestor that holds them, the way Readability resolves the same case by walking to the common ancestor of its close-scoring candidates.
```

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1334,7 +1350,14 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
             return;
         }
         if (tag === 'li') {
-            lines.push('- ' + node.textContent.trim());
+            // Recurse rather than flattening to `textContent`: returning here dropped the
+            // destination of every link inside a list item, which on a Wikipedia article is
+            // 1100 of 1723 links and on a results page is all of them. Children are folded
+            // back onto one line so a bullet still reads as a bullet.
```
houko: Same "Prose wrapping" rule violation as the block above — this `//` comment is hard-wrapped at a fixed column instead of one sentence per line ("the" / "destination", "is" / "1100 of 1723 links"). Reflow to one sentence per line per CLAUDE.md.

```suggestion
            // Recurse rather than flattening to `textContent`: returning here dropped the destination of every link inside a list item, which on a Wikipedia article is 1100 of 1723 links and on a results page is all of them.
            // Children are folded back onto one line so a bullet still reads as a bullet.
```

---

crates/librefang-runtime/src/browser.rs
houko: CLAUDE.md: "One sentence = one line, regardless of length." This doc-comment line packs two sentences onto one line ("...— 13.7% of the page." followed by "Readability resolves...close-scoring candidates."). Split at the sentence boundary.

```suggestion
    /// `querySelector('main, article, …')` returned the first match, which on a results page is one card — 13.7% of the page.
    /// Readability resolves the same case by climbing to the common ancestor of its close-scoring candidates.
```

---

changelog.d/fixed/6624-browser-list-item-links.md
houko: Per `changelog.d/README.md` ("Format"), the fragment must end `(#PR) (@your-github-login)`, and that trailing `(#N)` group is "how the release flow knows which generated line your prose replaces" — every other fragment in this directory closes with its own PR number matching the filename (e.g. `6689-release-body-limits.md` ends `(#6689)`, `6708-manual-schedule-delivery.md` ends `(#6708)`). This fragment closes with `(#6624)`, which is the tracking issue, not PR #6745. As written, `cargo xtask release` will keep the generated `- <PR title> (#6745) (@nevgenov)` line in addition to this bullet, so the fix is described twice in the release notes. Please reference the PR number in the closing citation.

```suggestion
A page with a `main` or `[role="main"]` landmark is unaffected, and a page with a single `article` still selects it (#6745) (@nevgenov)
```

---

houko (commented):
Automated CLAUDE.md compliance check: 4 findings (prose-wrapping rule x3, changelog PR-citation mismatch x1). See inline comments.

---

changelog.d/fixed/6624-browser-list-item-links.md
houko: Per `changelog.d/README.md`, the trailing `(#N)` group on a fragment is how the release automation matches a curated bullet to the PR it's replacing the generated line for: "Only the last `(#N)` group ... counts, so a mid-bullet cross-reference to some other PR is never mistaken for yours." and "Without it your PR keeps its generated line, so it appears twice in the release body."

This fragment ends `(#6624) (@nevgenov)`, but `#6624` is the tracking issue ("Refs #6624" in the PR body) — the PR that will actually merge is **#6745**. Since 6745 doesn't appear in this bullet's trailing group, `cargo xtask release` will emit *both* this curated bullet *and* a generated `- fix(browser): keep links inside list items, and cover feed pages instead of one card (#6745) (@nevgenov)` line, duplicating the entry in the release notes / Dev.to article / social post.

Suggested fix: end the bullet with the merged PR number, e.g. `(#6745) (@nevgenov)` (or `(#6624, #6745)` if you want to keep the issue reference too — per the README, either form works as long as `6745` is present in the last group on the last non-empty line).

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1334,7 +1350,14 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
...
+            const start = lines.length;
+            for (const child of node.childNodes) walk(child);
+            const inner = lines.splice(start).join(' ').replace(/\s+/g, ' ').trim();
+            if (inner) lines.push('- ' + inner);
```
houko: Edge case: nested `<ul>`/`<ol>` inside an `<li>` now produces confusing embedded bullet markers instead of just losing structure.

Trace through:
```html
<ul>
  <li>Fruits
    <ul><li>Apple</li><li>Banana</li></ul>
  </li>
</ul>
```
Walking the inner `<li>Apple</li>` and `<li>Banana</li>` each runs this same branch first (since `walk` recurses depth-first), so each pushes its own `'- ' + inner` entry into the shared `lines` array *before* the outer `<li>` reaches its own `lines.splice(start)`. The outer li's `splice(start).join(' ')` then re-flattens those already-prefixed inner bullets into itself, producing one line:

```
- Fruits - Apple - Banana
```

instead of either a real nested list or (as before this PR) the old plain-text flatten `- Fruits Apple Banana` (no PII loss, but at least it didn't look like a list). The extra `- ` tokens embedded mid-line read as spurious list markers / dashes to a model consuming this as markdown (worse if the sub-item text itself starts with a number, e.g. `- Price - 5 - 10`, which looks like a numeric range).

Since links inside these nested items still keep their href (the primary goal of this PR), this is a formatting regression rather than a data-loss regression, but given nested lists are explicitly one of the shapes worth checking here, worth a look. One option: track list nesting depth and only add the `'- '` prefix at the outermost `li` frame, joining inner li text without re-prefixing.

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1312,7 +1312,23 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
...
+        if (articles.length >= 3) {
+            let anc = articles[0].parentElement;
+            while (anc && anc !== clone && articles.filter(a => anc.contains(a)).length < 3) {
+                anc = anc.parentElement;
+            }
+            root = anc;
+        }
+    }
+    if (!root) root = clone.querySelector('article, .content, #content');
     if (!root) root = clone;
```
houko, line 1347: Now that root selection here can resolve to an ancestor holding several sibling `<article>` elements, each of those `<article>` children gets walked through the generic fallthrough path in `walk()` (further down, around the `if (['p','div','section','tr'].includes(tag)) lines.push('');` lines) — but `'article'` isn't in that blank-line tag list, so no separator is emitted before/after each entry.

Concretely: for a feed where entries don't each start with a heading (`h1`-`h6` do push a leading `'\n'`, so those self-separate), consecutive `<article>` cards get concatenated in `lines` with no blank line between them. The extraction now *covers* every feed entry (traced the recursion — it doesn't stop early, `for (const child of node.childNodes) walk(child)` visits every sibling regardless of what a previous call did), but without headings the entries can run together into one undifferentiated block rather than reading as distinct feed items, which works against the stated goal of "cover feed pages instead of one card."

Suggest adding `'article'` to that blank-line tag list (both the pre- and post-recursion occurrences) so feed entries stay visually distinct in the extracted markdown.

---

houko:
Note: this PR's description contains "🤖 Generated with [Claude Code]", which conflicts with this repo's no-AI-attribution policy for commit messages, PR bodies, and comments (see CLAUDE.md, "Git Conventions" / "Commit & PR hygiene"). Could you edit the PR description to remove it before merge?

---

changelog.d/fixed/6624-browser-list-item-links.md
houko: This fragment's closing tag references `(#6624)`, the tracking issue, not this PR's own number (`#6745`).
Per `changelog.d/README.md`: "end the bullet with its PR reference, `(#1234)`" — "That group is how the release flow knows which generated line your prose replaces."
`cargo xtask release` generates a `- <PR title> (#6745) (@nevgenov)` line for every merged PR unless that PR's number appears in a curated bullet's trailing `(#N)` group.
Since this fragment's group is `(#6624)` instead of `(#6745)`, the generated line for #6745 won't be suppressed and this entry will appear twice in the release notes.

Suggested fix (mechanical, one-character-class change):
```
A page with a `main` or `[role="main"]` landmark is unaffected, and a page with a single `article` still selects it (#6745) (@nevgenov)
```

I attempted to push this exact fix but lack write access to `nevgenov/librefang` from this session, so leaving it here instead.

---

crates/librefang-runtime/src/browser.rs
houko: This comment is hard-wrapped at a fixed column, which CLAUDE.md's prose rule forbids repo-wide ("no column-width rule for prose... the only legitimate line break inside a paragraph is at a complete sentence boundary"), and it explicitly lists "any multi-line free-form prose comment block."
This crate's own AGENTS.md reinforces it further: "Telegraph style. Short sentences. One idea per line."
It currently spans two sentences hard-wrapped across five lines instead of one line per sentence.

Suggested reflow:
```
    // `querySelector` returns the *first* match, which on a feed or a results page is one card rather than the list of them — measured at 13.7% of the page on a DuckDuckGo results page.
    // Repeated sibling `article` elements are the signature of that shape, so climb to the ancestor that holds them, the way Readability resolves the same case by walking to the common ancestor of its close-scoring candidates.
```

I attempted to push this exact fix but lack write access to `nevgenov/librefang` from this session, so leaving it here instead.

---

crates/librefang-runtime/src/browser.rs
houko: Same prose-wrap issue as the root-selection comment above: this is hard-wrapped at a fixed column across four lines for two sentences, instead of one sentence per line.

Suggested reflow:
```
            // Recurse rather than flattening to `textContent`: returning here dropped the destination of every link inside a list item, which on a Wikipedia article is 1100 of 1723 links and on a results page is all of them.
            // Children are folded back onto one line so a bullet still reads as a bullet.
```

I attempted to push this exact fix but lack write access to `nevgenov/librefang` from this session, so leaving it here instead.

---

crates/librefang-runtime/src/browser.rs
houko: This doc-comment line runs two sentences together ("...one card — 13.7% of the page. Readability resolves the same case...") instead of one sentence per line, per CLAUDE.md's prose-wrap rule which explicitly covers doc-comments (`///`).

Suggested fix:
```
    /// `querySelector('main, article, …')` returned the first match, which on a results page is one card — 13.7% of the page.
    /// Readability resolves the same case by climbing to the common ancestor of its close-scoring candidates.
```

I attempted to push this exact fix but lack write access to `nevgenov/librefang` from this session, so leaving it here instead.

---

houko (commented):
Reviewed the diff against CLAUDE.md's contributor rules (prose-wrap and changelog-fragment format). No AI-attribution footer was present in commit messages, and the actual git commit author (Nikita Evgenov) is clean — but the PR description itself carried a "🤖 Generated with [Claude Code]" footer, which I removed directly via the GitHub API since that's PR metadata, not a file in the diff.

I prepared mechanical fixes for the four findings below (comment/doc-comment prose-wrap reflow, and a changelog-fragment PR-reference mismatch) and committed them locally, but this session's credentials don't have write access to `nevgenov/librefang`, so the push was rejected — the fixes are left as inline suggestions instead.

---

changelog.d/fixed/6624-browser-list-item-links.md
houko: Per `changelog.d/README.md`: "Every merged PR in the range gets a `- <PR title> (#N) (@author)` line **unless** its number appears in the trailing `(#N)` group of a curated bullet." This fragment ends `(#6624) (@nevgenov)`, but #6624 is the tracking issue, not this PR's number (#6745). Since the merged PR number doesn't appear in the trailing group, `cargo xtask release` will still add a generated `- fix(browser): keep links inside list items... (#6745) (@nevgenov)` line alongside this curated one, duplicating the entry in the release notes (README: "Without it your PR keeps its generated line, so it appears twice in the release body, and `cargo xtask release` prints a warning naming the bullet."). Suggest ending with `(#6745) (@nevgenov)`, or `(#6624, #6745)` if both references are wanted.

---

crates/librefang-runtime/src/browser.rs
houko: CLAUDE.md's prose-wrapping rule ("Prose wrapping: no column limit; break only at sentence boundaries") explicitly covers "any multi-line free-form prose comment block," not just `///`/`//!` doc comments. This `//` block is hard-wrapped at a fixed column with line breaks mid-sentence (e.g. "...which on a feed or a results page is one" / "card rather than the list of them..."). The test doc-comments added later in this same PR (e.g. the `///` blocks above `test_list_items_recurse_instead_of_flattening_to_text`) correctly use one-sentence-per-line with no column limit — this block should follow the same convention.

---

crates/librefang-runtime/src/browser.rs
houko: Same prose-wrapping issue as the comment above `root` selection (lines 1315-1319): this `//` block is hard-wrapped at a fixed column with breaks mid-sentence ("...returning here dropped the" / "destination of every link..."). CLAUDE.md requires one-sentence-per-line with no column limit for any multi-line prose comment.

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1334,7 +1350,14 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
...
         }
         if (tag === 'li') {
-            lines.push('- ' + node.textContent.trim());
+            // Recurse rather than flattening to `textContent`: returning here dropped the
+            // destination of every link inside a list item, which on a Wikipedia article is
+            // 1100 of 1723 links and on a results page is all of them. Children are folded
+            // back onto one line so a bullet still reads as a bullet.
+            const start = lines.length;
+            for (const child of node.childNodes) walk(child);
+            const inner = lines.splice(start).join(' ').replace(/\s+/g, ' ').trim();
```
houko: Nested lists collapse into one inline-dash-joined line rather than preserving structure. For `<li>Outer<ul><li>Inner 1</li><li>Inner 2</li></ul></li>`, the inner `<li>` calls each push their own `'- Inner N'` string into the shared `lines` array (via the same branch, recursively), and then the outer `li`'s `lines.splice(start).join(' ')` folds those already-dashed strings back in as plain text — producing `- Outer - Inner 1 - Inner 2` as a single bullet, with the nested items' own `- ` markers now sitting mid-sentence as stray hyphens. Links inside the nested items keep their `[text](url)` markdown (so the core "lost href" bug this PR fixes doesn't recur here), but the result contradicts the stated goal that "a bullet still renders as a bullet" for any page with nested lists (TOCs, navboxes, threaded/nested content are common on real pages, e.g. Wikipedia). Worth either a comment noting this is a known/accepted limitation, or a check to avoid double-prefixing when `inner` already contains embedded `- ` markers from a nested list. No test exercises the nested case (both new tests only assert against substrings of `EXTRACT_CONTENT_JS_TEMPLATE`).

---

houko (commented):
Automated daily review. The core fix (recursing into `li` instead of flattening to `textContent`, and climbing to the common ancestor of repeated sibling `article` elements) looks sound and is well-tested for the cases it targets. Four items worth a look, none blocking:

1. `changelog.d/fixed/6624-browser-list-item-links.md` ends with the tracking-issue number `(#6624)` rather than this PR's number `(#6745)`, so per `changelog.d/README.md` the release tooling will still emit a duplicate generated bullet for #6745.
2. Two new `//` comment blocks in `browser.rs` (around the `main`/`article` root-selection logic and the `li` recursion) are hard-wrapped at a fixed column mid-sentence, which CLAUDE.md's prose-wrapping rule explicitly disallows for any multi-line prose comment block — the new `///` test doc-comments in the same PR correctly follow one-sentence-per-line instead.
3. Nested `<li>` structures collapse into a single line with stray inline `- ` markers from the inner items (links keep their href, but the bullet nesting/readability the PR aims for doesn't hold for this common real-world shape), and it's untested.

No approval/changes requested per policy — leaving these for the maintainer to weigh.

---

nevgenov:
Both code findings were real and are fixed; the mechanical ones too.

**Nested lists.**
Confirmed before changing anything — I ported `walk()` and traced `<ul><li>Fruits<ul><li>Apple</li><li>Banana</li></ul></li></ul>`, which produced `- Fruits - Apple - Banana` exactly as you described.
I did not take the suggested fix of prefixing only at the outermost `li`, because that yields `- Fruits Apple Banana` and gives up the sub-list structure the recursion had just recovered.
A list item now folds only its *own* text and leaves a sub-list's bullets on their own lines with a level of indent, so the example renders as a real nested list, and three levels nest correctly.
Whether a produced line came from a nested `li` is tracked as a flag alongside the line rather than sniffed back out of the string, since a text node may legitimately begin with `- `.

**`article` separator.**
Added beside `p`, `div`, `section`, `tr`, for the reason you gave.

**Prose wrapping.**
Reflowed to one sentence per line in the source comments and in the PR description.
I left the pre-existing paragraphs in these files alone, per CLAUDE.md's "do not rewrap an untouched paragraph just to enforce the rule."

**Changelog fragment.**
Now closes `(#6624, #6745)` — the PR number so the generated line is suppressed, the issue kept since it is the tracking issue.

**AI footer.**
Removed from #6746's description as well.
One thing worth flagging: the API edit that stripped the footer from this PR's description HTML-escaped the quotes in it, leaving `&#34;` and `&#39;` inside code spans, where GitHub renders them literally rather than decoding them — so `[role="main"]` was displaying as `[role=&#34;main&#34;]`. Decoded in this update.

Two new tests: the nested-list fold ported to Rust, and the `article` separator. `2,177` tests pass in `librefang-runtime`, clippy `--all-targets` and fmt clean.

changelog.d/fixed/6624-browser-list-item-links.md
```diff
@@ -0,0 +1,8 @@
...
+Clicking the text was not a fallback for those: 408 of the 1,100 do not resolve to themselves under `browser_click`'s substring matcher, so there was neither a URL nor a working text handle.
+The branch now recurses and folds its children back onto one line, so a bullet still renders as a bullet and the links inside it keep their identity.
+Root selection used `querySelector('main, article, [role="main"], .content, #content')`, which returns the *first* match — on a page built from sibling `<article>` cards that is one card, measured at 13.7% of a DuckDuckGo results page.
+Selection now climbs to the ancestor holding repeated sibling `article` elements, the way Readability resolves the same shape by walking to the common ancestor of its close-scoring candidates.
+A page with a `main` or `[role="main"]` landmark is unaffected, and a page with a single `article` still selects it (#6624, #6745) (@nevgenov)
+A sub-list nested inside a list item keeps its own bullets, indented, rather than being folded into its parent's line, and an `article` now earns a blank line the way `section` and `div` already do so feed entries without headings stay distinct (#6624, #6745) (@nevgenov)
```
houko: changelog.d format nit (per `changelog.d/README.md` / CLAUDE.md's changelog-fragment rules): "One file, one bullet," ending with a single `(#PR) (@login)` stamp.

This file now has two complete sentences that each end in their own `(#6624, #6745) (@nevgenov)` stamp (line 7 and line 8). `render_fragment_bullet` in `xtask/src/changelog.rs` doesn't split a fragment on an embedded stamp — it folds the *entire* file into one `- ` bullet with two-space-indented continuation lines. So after `cargo xtask collect-fragments` this becomes a single CHANGELOG bullet with the `(@nevgenov)` attribution appearing twice, once mid-paragraph and once at the true end, e.g.:

```
- ...
  A page with a `main` or `[role="main"]` landmark is unaffected, and a page with a single `article` still selects it (#6624, #6745) (@nevgenov)
  A sub-list nested inside a list item keeps its own bullets, indented, ... (#6624, #6745) (@nevgenov)
```

I checked `CHANGELOG.md` for precedent and found zero existing bullets with more than one `(#N) (@user)` group — every multi-PR bullet closes with exactly one trailing group covering all the PRs it documents (e.g. `(#6594, #6595) (@houko)`).

Suggest either merging line 8's content into the existing narrative so there's a single trailing `(#6624, #6745) (@nevgenov)`, or splitting it into its own fragment file (e.g. `changelog.d/fixed/6745-nested-list-bullets.md`) if it's meant to read as a second, separate bullet.

---

houko (commented):
Reviewed against CLAUDE.md's engineering rules (worktree hygiene N/A for a review pass, changelog-fragment format, deterministic-ordering, config/Default-impl and Option&lt;Arc&lt;dyn Trait&gt;&gt; patterns, commit conventions, no-AI-attribution, and the sentence-per-line prose rule).

The code change itself (crates/librefang-runtime/src/browser.rs) is in good shape: the follow-up commit already reflowed all new comments and doc-comments to one sentence per line per CLAUDE.md, commit messages are conventional and carry no AI/Anthropic attribution, and the diff doesn't touch anything implicated by the determinism/BTreeMap, config-Default, or Option&lt;Arc&lt;dyn Trait&gt;&gt; rules (no HashMap/HashSet, no new config fields, no new dyn-trait fields).

One changelog-fragment format issue left as an inline comment: the fragment file bundles two bullets (each ending in its own `(#PR) (@login)` stamp) into one file, which the assembly tool folds into a single bullet with the attribution stamp appearing twice.

Note: this PR's head repo is a fork (nevgenov/librefang), so I don't have push access to fix anything directly — everything above is left as review comments rather than commits.

---

nevgenov:
Both fixed.

**Fragment with two stamps.**
Confirmed against `render_fragment_bullet` in `xtask/src/changelog.rs:889` — it folds the whole file into one `- ` bullet with indented continuations and never splits on an embedded stamp, so the attribution would indeed have landed twice.
I also checked `CHANGELOG.md`: zero bullets carry more than one `(#N) (@user)` group, so there is no precedent either way.
Split into its own file rather than merged, since the sub-list fold and the `article` separator are a separate change from the two the first fragment documents: `changelog.d/fixed/6745-nested-list-bullets.md`.

**Commit bodies.**
You are right that CLAUDE.md lists commit message bodies in scope, and I had applied the rule everywhere except there.
All three commits reworded to one sentence per line and force-pushed.
The trees are byte-identical to what you reviewed apart from the changelog split above — `git diff` between the old and new tips is only those two fragment files.

crates/librefang-runtime/src/browser.rs
```diff
@@ -1312,35 +1312,64 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
...
 
-    let root = clone.querySelector('main, article, [role="main"], .content, #content');
+    // `querySelector` returns the *first* match, which on a feed or a results page is one card rather than the list of them — measured at 13.7% of the page on a DuckDuckGo results page.
+    // Repeated sibling `article` elements are the signature of that shape, so climb to the ancestor that holds them, the way Readability resolves the same case by walking to the common ancestor of its close-scoring candidates.
+    let root = clone.querySelector('main, [role="main"]');
+    if (!root) {
+        const articles = Array.from(clone.querySelectorAll('article'));
+        if (articles.length >= 3) {
+            let anc = articles[0].parentElement;
+            while (anc && anc !== clone && articles.filter(a => anc.contains(a)).length < 3) {
```
houko: The climb condition `articles.filter(a => anc.contains(a)).length < 3` counts any 3 `<article>` elements the ancestor transitively contains, not 3 elements that are actual siblings in a repeated-card layout. On an ordinary blog page — one `<article>` for the main post plus a "related posts" widget in the sidebar/footer whose items are also `<article>` tags (a very common pattern) — this loop keeps climbing past the intended main-content container. The only stop condition besides finding 3 contained articles is `anc !== clone`, i.e. it can climb all the way to the top of the cloned document, pulling in sidebar/footer/related-post content instead of the focused article extraction the old code gave for a single-article page.

`test_root_selection_climbs_past_repeated_articles` only asserts the string `"articles.length >= 3"` is present; it doesn't exercise the actual DOM-climbing behavior, so this over-reach case isn't caught. Consider requiring the 3 articles to share a common *direct* parent (true siblings) rather than merely being contained somewhere under the climbing ancestor.

---

nevgenov:
Confirmed and fixed — this one was a real over-reach, not just a missing test.

I built the case before changing anything: an ordinary post page with no `main`, one `article` for the post, and a related-posts widget of two more.
The old condition resolved the root to `body`.
Worth noting the blast radius is a little smaller than it looks, since `nav`, `header`, `footer` and `aside` are stripped before selection runs — a widget inside `<aside>` was already gone.
A plain `<div class="related">` was not, and that is common enough.

I did not take the suggested rule of requiring the three articles to share a direct parent, because it misses a feed whose cards each sit in a wrapper — `<div class="card"><article/></div>` repeated — which is at least as common as bare siblings.
The condition is now on the ancestor's *direct children*: at least three of them must each carry an article of their own.
That reads the same shape without the string-matching on layout:

| page | old rule | new rule |
|---|---|---|
| results page, 11 sibling cards | `div.results` | `div.results` |
| feed, 5 cards each in a wrapper | `div.feed` | `div.feed` |
| post + related-posts widget | **`body`** | falls through to the post |
| single-article page | falls through | falls through |

On the blog row the new rule counts two article-carrying children, one for the post and one for the widget, so selection falls through to `querySelector('article, …')` and the page keeps the focused extraction it always had.

You are also right that the existing test could not have caught it.
`test_root_climb_stops_at_a_container_of_repeated_cards` ports the rule over a small tree and pins all four rows above by behaviour rather than by asserting a string is present in the template.

crates/librefang-runtime/src/browser.rs
```diff
@@ -1312,35 +1312,70 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
...
+    // The test is on the ancestor's *direct children*: at least three of them must each carry an article of their own.
+    // Merely containing three articles somewhere below is not the same shape and climbs too far — an ordinary post page whose "related posts" widget is built out of `article` clears that bar, and the climb then runs to the top of the document and pulls the widget in beside the post.
+    // Counting article-carrying children instead sees two there, one for the post and one for the widget, so selection falls through to the single-article path that page always had.
+    // Counting the articles themselves as direct children would be the narrower rule and would miss a feed whose cards each sit in a wrapper, which is the common shape.
+    let root = clone.querySelector('main, [role="main"]');
+    if (!root) {
+        const articles = Array.from(clone.querySelectorAll('article'));
+        if (articles.length >= 3) {
+            let anc = articles[0].parentElement;
```
houko: Correctness: the climb anchors on `articles[0]` — the very first `&lt;article&gt;` in document order — and only walks *that element's* ancestor chain.
If the page has an early, unrelated `&lt;article&gt;` that sits *before* the real feed container in the DOM (a very common layout: a single "featured post" card above a grid of `&lt;article&gt;` cards, e.g. typical WordPress/Ghost blog homepages), the climb never visits the feed's branch of the tree at all — it only walks up from the featured card.
None of the featured card's ancestors will have ≥3 qualifying direct children (the feed is a sibling subtree, not an ancestor), so the climb falls through to `clone.querySelector('article, .content, #content')` on line 1334 — which is the *same first-match query* this PR set out to fix, so it just re-selects that one featured card and misses the whole feed. That reproduces exactly the "13.7% coverage / one card from a feed" bug this PR fixes, just triggered by a "featured card before the grid" shape instead of a flat results list.

Concretely:
```html
&lt;div id="lone"&gt;&lt;article&gt;Featured post&lt;/article&gt;&lt;/div&gt;
&lt;div id="feed"&gt;&lt;article/&gt;&lt;article/&gt;&lt;article/&gt;&lt;/div&gt;
```
`articles[0]` is inside `#lone`. Its ancestors (`#lone`, then `body`) never see 3 carriers (`body`'s two children are `#lone` and `#feed`, both "carriers" by the `c.querySelector('article')` branch, so `carriers.length` is 2, not 3) — climb fails, falls through, selects only the featured post. `#feed` (the actual 3-article group) is never examined.

This also means `test_root_climb_stops_at_a_container_of_repeated_cards` (added in this PR) doesn't actually pin the real algorithm: its `chosen()` helper does a full-tree DFS for *any* node with ≥3 qualifying children, not a single climb from one starting article, and its own doc-comment's equivalence claim ("walking down and taking the deepest ancestor that qualifies reaches the same node, since an ancestor of a qualifying node cannot qualify with fewer carriers") doesn't hold in the counterexample above — `#feed` is not an ancestor of `articles[0]`, so the real JS climb can't reach it even though the Rust model would find it.

Suggested fix: don't anchor the climb on `articles[0]` specifically. Either (a) try climbing from every article's ancestor chain (or from the *last* article, or the one deepest in the tree) and accept the first ancestor that qualifies from any of them, or (b) do the full-tree search the Rust test already models, so the JS and its own regression test describe the same algorithm. Worth adding a "featured card + separate feed" case to the test either way.

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1312,35 +1312,70 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
...
+    // Repeated sibling `article` elements are the signature of that shape, so climb to the ancestor that holds them, the way Readability resolves the same case by walking to the common ancestor of its close-scoring candidates.
+    // The test is on the ancestor's *direct children*: at least three of them must each carry an article of their own.
+    // Merely containing three articles somewhere below is not the same shape and climbs too far — an ordinary post page whose "related posts" widget is built out of `article` clears that bar, and the climb then runs to the top of the document and pulls the widget in beside the post.
+    // Counting article-carrying children instead sees two there, one for the post and one for the widget, so selection falls through to the single-article path that page always had.
+    // Counting the articles themselves as direct children would be the narrower rule and would miss a feed whose cards each sit in a wrapper, which is the common shape.
+    let root = clone.querySelector('main, [role="main"]');
```
houko, line 1327: Sequencing note (not a defect in this PR): PR #6746, from the same author, touches this exact code path — it changes the `tag === 'a'` branch to emit a `⟨n⟩` link marker instead of `[text](href)`, and adds the marker/table budgeting logic right after this `root` selection block. #6746's diff already contains this PR's full final state (the direct-children `carriers` climb, the `isBullet`/nested-list fold, `article` in `BLOCK`), so it looks rebased on top of this PR's tip rather than on `main`. No functional conflict as things stand, but #6746 should merge after this one (and be re-verified against whatever lands here, including the `articles[0]` climb-anchor issue above, since #6746 inherits that same code).

---

houko (commented):
Reviewed the current head (0181adc). All findings from the prior automated review round (prose wrapping, changelog PR-number citation, nested-list bullets, article block separator, related-posts over-reach) are already fixed in this commit — no need to repeat those. Two new items below: one correctness bug in the feed-root climb logic, one cross-PR sequencing note re #6746.

---

crates/librefang-runtime/src/browser.rs
houko: Independent-review confirmation (corroborates the still-unresolved, non-outdated review comment already on this line from 2026-08-08T11:42): the climb starts at `articles[0].parentElement` and only walks that one article's ancestor chain, so it can only ever select a container that is an ancestor of the very first `<article>` in document order.

On a "featured card above a grid" layout — one standalone `<article>` followed by a sibling container of ≥3 more `<article>` cards, e.g.:
```html
<div id="lone"><article>Featured post</article></div>
<div id="feed"><article/><article/><article/></div>
```
`articles[0]` is the featured card. Walking up from it, `body`'s direct children are `#lone` and `#feed` — both count as "carriers" via `c.querySelector('article')`, so `carriers.length` is 2, never 3. The climb never reaches or examines `#feed` (it isn't an ancestor of `articles[0]`), falls through to `clone.querySelector('article, .content, #content')` at the next line, and re-selects only the featured card — reproducing the exact "one card out of a feed" bug this PR sets out to fix, just via a different DOM shape than the flat results-list case the new test covers.

`test_root_climb_stops_at_a_container_of_repeated_cards`'s `chosen()` helper does a full-tree search for any qualifying ancestor rather than a single climb anchored on one starting node, so it doesn't pin the actual algorithm here and passes despite this gap.

Since the climb must be anchored on *some* article to walk upward at all, and the true container isn't necessarily an ancestor of the first one in document order, this needs either trying the climb from each article (or the last one) and taking the first ancestor that qualifies, or replacing the single-anchor climb with the full-tree search the Rust test already models.

---

houko (commented):
Scheduled automated review. Most of the substantive issues raised in this thread's earlier review rounds are already resolved on the current tip (0181adcb): the changelog fragments each carry a single trailing (#6624, #6745) stamp, the added JS comments and doc-comments are reflowed to one-sentence-per-line, the nested-sub-list-folding bug is fixed via the isBullet flag, article now gets a block separator, and the "related posts widget" over-reach case is fixed by requiring direct-child carriers. One correctness gap from the most recent review round remains open and unaddressed on the current tip; left as a single inline comment rather than restating everything already tracked in this thread.

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1312,35 +1312,70 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
...
+    if (!root) {
+        const articles = Array.from(clone.querySelectorAll('article'));
+        if (articles.length >= 3) {
+            let anc = articles[0].parentElement;
+            while (anc) {
+                const carriers = Array.from(anc.children).filter(c => c.tagName === 'ARTICLE' || c.querySelector('article'));
```
houko: Question on the climb bound, since it's tested against the "one legit feed + one unrelated widget" shape but not against a page where the 3-carrier threshold is met by *unrelated* sections rather than one repeated layout.

`c.querySelector('article')` looks arbitrarily deep below each direct child of `anc`, so a child can count as a "carrier" for having a single incidentally-nested `<article>` anywhere under it (e.g. an embedded article widget, a "featured post" teaser in a sidebar, a syndicated card in a totally separate section) rather than being part of the repeated grid the heuristic is trying to find.

Concretely: a page whose `<body>` has three direct sections — a real 3+ card feed, a "trending" sidebar with one `<article>` teaser, and a "related" widget with one more — has 2 real feed articles plus 2 unrelated ones. `articles.length >= 3` is satisfied, and the climb from `articles[0]`'s parent keeps walking up (each level's carrier count stays below 3 until the shared ancestor, which could be `body`/`clone` itself) — extraction then returns the whole page (nav/footer already stripped, but sidebar + feed + related widgets all mixed in), which is a regression from "one card" to "everything", not from "one card" to "the feed".

`test_root_climb_stops_at_a_container_of_repeated_cards` covers the post+widget case (2 carriers, correctly stays put) but there's no case where the 3-carrier bar is cleared by *scattered, unrelated* single-article containers rather than one repeated layout. Is that shape considered acceptable fallback behavior (worst case: back to whole-page extraction, no worse than the pre-#6624 baseline for non-`main` pages), or would it be worth tightening the carrier test to also require size/shape similarity between the candidate carriers (closer to what Readability's "close-scoring candidates" comparison actually checks)?

---

houko (commented):
One question on the new feed-root climb heuristic, inline below. Everything else — the `li` recursion/nested-bullet fold, the `article` block separator, changelog fragments, commit messages — looks correct and well-tested for the cases exercised.

---

crates/librefang-runtime/src/browser.rs
```diff
@@ -1562,6 +1597,214 @@ mod tests {
...
+            "repeated sibling articles are the signal to climb to their container"
+        );
+    }
+
+    /// The climb stops at a container of repeated cards, and does not run past it on a page that merely has articles in more than one place.
+    ///
+    /// Ported from the JS the script runs, because the climb needs a live DOM.
+    /// Counting *contained* articles rather than article-carrying children climbs to the top of the document on an ordinary post page whose "related posts" widget is built out of `article`, taking the widget along with the post.
+    #[test]
+    fn test_root_climb_stops_at_a_container_of_repeated_cards() {
```
houko: This is a synthetic mock tree (`Node { tag, name, kids }` built by hand with placeholder `article("card")` leaves), and `chosen()` is a *reimplementation* of the climb — it walks top-down via DFS and returns the first "deepest" qualifying node, whereas the real JS (`extract_content_js`) starts at `articles[0].parentElement` and climbs upward. The doc comment asserts these are equivalent ("walking down and taking the deepest ancestor that qualifies reaches the same node, since an ancestor of a qualifying node cannot qualify with fewer carriers") but that's not obviously true in general — e.g. a page with two independent sibling containers that each independently clear the `>=3` bar (two separate result lists under one wrapper) would have `deepest()` return whichever branch is explored first in child order, not necessarily the ancestor chain of `articles[0]` that the real JS would actually climb to. None of the four cases here exercise that.

More importantly for what the task description asks for: this doesn't satisfy "a test with a real feed-page fixture proving multiple items are extracted, not just one card." All of `test_root_selection_climbs_past_repeated_articles` / `test_article_earns_a_block_separator` / this test assert either substring presence in the JS template string or a synthetic node-count decision — none of them run the actual `walk()`/`emit()` extraction against realistic markup (e.g. an embedded DuckDuckGo-results-shaped or Wikipedia-list-shaped HTML string) and assert on the resulting *text output* that N distinct items/links actually made it through. Given this crate has no live-Chrome test harness (per `librefang-runtime/CLAUDE.md`: "historically had ZERO integration tests"), the closest achievable version would be parsing a real HTML fixture with a Rust HTML crate (e.g. `scraper`/`html5ever`) and porting enough of `walk()` to run against it — the same "port to Rust because the JS needs a live DOM" pattern already used here, just applied to the full pipeline output instead of only the root-selection decision.

---

houko (commented):
One finding on test coverage for the root-selection climb; changelog fragments, formatting (`cargo fmt --check` clean), and no-AI-attribution all check out fine.

---

nevgenov:
The climb-anchor finding is right, and it was the same bug class this PR exists to fix. Fixed.

I reproduced it before changing anything, on the shape you gave — one `article` in its own wrapper, then a sibling container of three:

| page | old | new |
|---|---|---|
| featured card above a grid | **one card** | the grid |
| feed flanked by trending(1) + related(1) | the feed | the feed |
| post + related-posts widget | the post | the post |
| results page, 11 sibling cards | the container | the container |
| feed, 5 cards each in a wrapper | the container | the container |
| single-article page | the article | the article |

Selection now searches the tree — children before the node itself — and takes the deepest node whose direct children carry at least three articles between them. The carrier test is unchanged.

That also answers your second question, the one about the 3-carrier bar being cleared by scattered unrelated containers rather than one repeated layout. Taking the *deepest* match means the feed qualifies below the ancestor that happens to hold the feed and both widgets, so selection lands on the feed rather than widening to the page. Row 2 above is exactly that shape and it is now pinned.

You were also right that the old test could not have caught this: its helper searched the whole tree while the script climbed one chain, so it described a different algorithm and passed anyway, and its doc-comment asserted an equivalence that does not hold precisely when the qualifying node is not an ancestor of the first article. The helper now matches the search the script performs, and I checked parity by lifting the JS straight out of `browser.rs` and running both over all six shapes above rather than trusting that they agree.

The remaining string assertion is narrowed to one thing — that the first-match query does not come back — and renamed to say so. Which container gets chosen is behaviour and is tested as behaviour.

## On the real-HTML-fixture test

Not doing this one, and I would rather say why than quietly skip it.

It needs an HTML parser in `librefang-runtime`, which has none — `html5ever` is in the lock file only transitively, through something else's tree — so it is a new dev-dependency. That part is cheap. The expensive part is what it implies: to assert on extracted *text* you have to port `walk()`, `emit()`, the `li` fold, the block-separator list and the cap budgeting into Rust. That is a second implementation of the extraction, and unlike the small rule-ports here it is large enough to drift from the JS while still passing — which is the failure mode I would be buying the coverage with.

The ports in this PR are deliberately small enough to eyeball against the source, and I check them against the JS rather than assuming: the parity run above is what convinced me the new search is right, not the Rust test on its own.

If a live-Chrome harness ever lands, that is where this coverage belongs, and I would be glad to write it there.

crates/librefang-runtime/src/browser.rs
```diff
@@ -1312,35 +1312,73 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
...
+    // Searched over the whole tree rather than climbed from the first article's ancestors, because the container is not always an ancestor of the article that happens to come first in the document.
+    // A featured card above a grid — `<div><article/></div>` followed by a sibling `<div>` of three more — puts the first article in a branch the grid is not under, so a climb never examines the grid at all and falls through to the very first-match query this exists to replace.
+    // The deepest qualifying node wins, so a page whose feed sits beside unrelated single-article widgets selects the feed rather than their common ancestor.
+    let root = clone.querySelector('main, [role="main"]');
+    if (!root) {
+        const carriesArticle = el => el.tagName === 'ARTICLE' || el.querySelector('article');
+        function cardContainer(node) {
+            for (const child of node.children) {
+                const found = cardContainer(child);
+                if (found) return found;
```
houko: `cardContainer` returns on the *first* child branch (in document order) whose subtree contains any qualifying node, before ever looking at that node's siblings — and it recurses to the bottom of a branch before checking that branch's own top-level qualification. That means a small, deeply-nested cluster that happens to sit in an earlier sibling can win over a larger, shallower cluster that is the actual feed.

Concretely:
```html
<body>
  <div id="sidebarWrap">
    <div id="smallWidgetWrapper">
      <div id="smallWidget"><article/><article/><article/></div>  <!-- 3 articles, deep, first sibling -->
    </div>
  </div>
  <div id="feed"><article/><article/>...<article/></div>  <!-- 20 articles, shallow, second sibling: the real feed -->
</body>
```
`cardContainer(body)` recurses into `sidebarWrap` first, finds `smallWidget` qualifies (3 direct children each carrying an article) at depth 3, and returns it immediately — `#feed` (depth 1, the actual feed this PR is trying to cover) is never even visited, because the loop returns on the first non-null child result before moving to the next sibling.

This is a different shape than the "flanked" / "featured card" cases already tested in `test_root_selection_finds_a_container_of_repeated_cards` — those have exactly one legitimately-qualifying cluster. Here there are two independent qualifying clusters at different depths in different sibling branches, and the "deepest wins" rule (which is otherwise the right call for the flanked case) picks the smaller/incidental one purely because it's nested deeper, not because it's the better match. A "trending"/"related" widget with 3+ small cards placed before the main feed in the DOM (not wrapped in `<aside>`, so not stripped by the earlier `remove` list) seems like a plausible real-world trigger.

Not proposing a specific fix here since it's a heuristic trade-off (e.g. preferring the candidate with the most carrier children, or the widest/shallowest match, would need its own test matrix) — flagging so it's a conscious call rather than an unexamined gap. Worth a test case alongside the existing "featured card" / "flanked" ones either way.

Lower-confidence secondary note: `cardContainer` calls `carriesArticle` (which does `el.querySelector('article')`, an O(subtree) scan) for every direct child of every node it visits, and it visits every node in the tree when no qualifying container exists (e.g. the "blog" fallthrough case). That's roughly O(n·depth) rather than O(n) in the worst case. Probably fine at the DOM sizes discussed in the PR description, but worth a thought if this ever needs to run on much larger/deeper pages.

---

houko (commented):
Independent pass over the current diff (head 39daca6). Most of what I found had already been raised and fixed in this thread's earlier rounds — the `li` nested-list fold, the `article` block separator, and the climb-anchor-on-`articles[0]` issue all look correctly resolved in the current code, and I traced the recursion/splice logic in the `li` branch by hand and didn't find a link-loss or double-count case. Left one new finding below on the full-tree `cardContainer` search: a first-branch-wins-in-document-order DFS can pick a small, deeply-nested qualifying cluster over a larger, shallower one (the actual feed) sitting in a later sibling — a different shape than the "flanked"/"featured card" cases already tested. Also included a lower-confidence note on `cardContainer`'s worst-case time complexity.

Separately, not a defect in this PR: this touches the same extraction code as #6746 from the same author, which per an earlier comment on this thread is already rebased on this PR's tip — so it should merge after this one and be re-checked against whatever lands here.

---

nevgenov:
Confirmed and fixed. Depth was doing work it could not justify.

I built your shape before touching anything — a three-card widget nested three levels deep in the first sibling, a twenty-card feed one level down in the second — and the search returned `smallWidget`, exactly as you traced. `#feed` was never visited.

The rule now has two parts:

- A candidate that **contains** another loses to it. That is what keeps the flanked case from widening to the ancestor holding the feed and both widgets, which is what deepest-wins was really buying.
- Between candidates that contain neither one another, the one with the **most article-carrying children** wins, first-in-document-order on a tie. They are independent clusters, and as you say, depth says nothing about which is the page's subject.

| page | before | after |
|---|---|---|
| deep 3-card widget before a 20-card feed | **the widget** | the feed |
| feed flanked by trending(1) + related(1) | the feed | the feed |
| featured card above a grid | the grid | the grid |
| results page, 11 sibling cards | the container | the container |
| feed, 5 cards each in a wrapper | the container | the container |
| post + related-posts widget | the post | the post |
| two equal feeds | first | first |

The last two rows are new in the test: your shape, and an equal-size tie so the choice is pinned as stable rather than an artefact of traversal order.

I checked parity by lifting the JS out of `browser.rs` and running it against the Rust model over all eight shapes, rather than trusting they agree — that is also what caught a subtlety worth mentioning: Rust's `max_by_key` returns the *last* of equal maxima where the script's `reduce` keeps the first, so the model is a fold rather than a `max_by_key`.

**On the complexity note** — taken, and it was the easier half. `carriesArticle` did `querySelector('article')` per child, rescanning each subtree once per sibling, which is the O(n·depth) you describe and worst exactly in the fallthrough case where nothing qualifies. It is now one post-order pass that counts articles per subtree and collects candidates from those counts, so each node is visited once. The candidate set is small enough that the containment filter over it does not matter.

PR #6750, opened 2026-08-08
fix(media): stop the media tests reaching real providers, and drop response_format for gpt-image

houko:
## Why

Found while running `cargo test -p librefang-api` during unrelated work: the media integration tests **made a real, billable OpenAI call**.

`media_speech_no_configured_provider_returns_missing_key` asserts 422 for the missing-key path. With `OPENAI_API_KEY` present in the shell it returned **200**, having performed an actual TTS request and written the resulting mp3 to `/api/uploads/7e386717-d565-41ae-982d-98ced8d837b1` with `duration_ms: 500`. `media_image_no_configured_provider_returns_missing_key` hit the live image endpoint the same way.

The module doc claimed these run "with no media-provider API keys configured", and the comment at what is now line 337 states the assumption outright. It was an assumption about the developer's shell, never something the harness enforced. A test whose name asserts *no configured provider* should be structurally incapable of reaching one.

## Changes

**`crates/librefang-api/tests/media_routes_integration.rs`** — `boot()` clears the 13 environment variables any media driver reads for credentials (`MEDIA_PROVIDER_KEY_VARS`, derived by grepping `crates/librefang-runtime-media/src/` for `_API_KEY`). Every driver reads its key through `std::env::var` at call time (`openai.rs:42`, `:281`, and the equivalents in each sibling module), so clearing the process environment before the first request is what makes the assertion true rather than merely likely.

`Once` rather than a per-`boot()` loop: all 18 tests reach a provider only through a request issued after their own `boot()` returns, so a single clear completing before the first harness is handed out covers all of them. Verified all 18 `#[tokio::test]` functions call `boot()`.

A new test, `boot_clears_provider_credentials_from_the_environment`, asserts the environment really is clean afterwards. This is the guard for the guard — the other assertions are 4xx checks that a bogus key could also satisfy via a provider-side 401, so without this one they could keep passing for the wrong reason.

**`crates/librefang-runtime-media/src/media_understanding.rs`** — `transcribe_audio_accepts_video_type` had the same environment dependency from the opposite side, and **fails on `main` today for anyone with an STT key exported**.

`transcribe_audio` resolves the provider (`config.audio_provider`, else `detect_audio_provider()` reading `OPENAI_API_KEY` / `GROQ_API_KEY` / …) at `media_understanding.rs:208-216`, *before* reading the file at `:231-237`. On a clean machine resolution fails and the call stops there, which is the message the test pinned. With a key present resolution succeeds and the call proceeds to the file read, so the test failed on an assertion about the developer's shell rather than about the type guard it exists to cover.

It now accepts either error — both prove the call cleared the type guard — and its path is a constructed absent one under `temp_dir()` rather than the bare relative `"test.mp4"`. That second part is load-bearing, not cosmetic: provider resolution succeeding is not hypothetical, and if the read then succeeded the next step would be a real billable STT request. A guaranteed-absent path makes that unreachable. Constructed rather than borrowed from the host filesystem, per #5716.

**`crates/librefang-runtime-media/src/openai.rs`** — the billable call that exposed all this returned `400 Unknown parameter: 'response_format'`, which is a live defect independent of the tests.

Both image-generation bodies (`:91` and `:329`) sent `"response_format": "b64_json"` unconditionally. The DALL·E models accept it; the `gpt-image-*` family rejects it outright and always returns base64, so **image generation fails completely against those models today**. The parameter is now gated on `image_model_accepts_response_format(model)`, which excludes only that family — unrecognised names keep sending it, so third-party OpenAI-compatible endpoints are unaffected.

The response parser at `:127-140` already reads both `b64_json` and `url`, so omitting the parameter needs no parsing change. The `response_format` at `:187` is left alone: that is the TTS body, where it is a valid parameter selecting the audio container.

## Verification

macOS aarch64, native cargo, based on `origin/main` at c403439a.

- **`cargo test -p librefang-api --test media_routes_integration`** with `OPENAI_API_KEY`, `GEMINI_API_KEY`, `ELEVENLABS_API_KEY` and `GROQ_API_KEY` all exported — **19 passed, 0 failed**. This is the configuration that previously made live calls.
- **Mutation check on the new guard test.** Disabling the body of `clear_media_provider_env` makes `boot_clears_provider_credentials_from_the_environment` fail with `OPENAI_API_KEY is still set after boot() — a test asserting the missing-key path could reach a real provider and bill the developer`. Restored and re-confirmed green. The test discriminates; it is not green by construction.
- **`cargo test -p librefang-runtime-media --lib`** in both environments: **93 passed, 0 failed** with keys present (the case that is red on `main`), and **93 passed, 0 failed** under `env -u` for all 13 variables, which is what CI sees.
- **`cargo clippy --workspace --all-targets -- -D warnings`** — exit code **0**, captured explicitly.
- **`cargo fmt --check`** — exit code 0.

### Not verified

No live provider call was made to confirm that `gpt-image-1` now succeeds — doing so would be the billable request this PR exists to prevent. The fix is derived from the observed `400 Unknown parameter: 'response_format'` returned by the live endpoint and from the documented parameter set; the unit tests pin the model-name predicate, not the provider's acceptance of the resulting body.

## Scope note

The `transcribe_audio_accepts_video_type` fix is a pre-existing failure rather than something this PR introduced. It is included because it is the same defect class in the same crate — a media test whose correctness depends on the developer having no credentials, with a real billable call sitting one successful file read away.

### DIFF
--- a/crates/librefang-api/tests/media_routes_integration.rs
+++ b/crates/librefang-api/tests/media_routes_integration.rs
@@ -65,12 +65,8 @@ const MEDIA_PROVIDER_KEY_VARS: &[&str] = &[
 fn clear_media_provider_env() {
     static ONCE: std::sync::Once = std::sync::Once::new();
     ONCE.call_once(|| {
-        // SAFETY: runs inside `Once::call_once`, which guarantees this body
-        // executes exactly once and completes before any other caller of
-        // `clear_media_provider_env` proceeds — including the first `boot()`,
-        // which is the only place any test in this file spawns work that
-        // reads the environment. No other thread reads or writes these vars
-        // concurrently with this block.
+        // SAFETY: runs inside `Once::call_once`, which guarantees this body executes exactly once and completes before any other caller of `clear_media_provider_env` proceeds — including the first `boot()`, which is the only place any test in this file spawns work that reads the environment.
+        // No other thread reads or writes these vars concurrently with this block.
         unsafe {
             for var in MEDIA_PROVIDER_KEY_VARS {
                 std::env::remove_var(var);
--- a/crates/librefang-api/tests/media_routes_integration.rs
+++ b/crates/librefang-api/tests/media_routes_integration.rs
@@ -1,9 +1,6 @@
 //! Integration tests for the `/media/*` HTTP surface.
 //!
-//! These exercise the real `media` router against a freshly-booted kernel
-//! with no media-provider API keys configured — `boot()` strips them from the
-//! process environment rather than assuming the developer's shell has none, so
-//! a test asserting the missing-key path cannot silently make a billable call.
+//! These exercise the real `media` router against a freshly-booted kernel with no media-provider API keys configured — `boot()` strips them from the process environment rather than assuming the developer's shell has none, so a test asserting the missing-key path cannot silently make a billable call.
 //! We focus on:
 //!
 //! 1. Validation paths in each `MediaError`-producing handler — the cheap,
@@ -38,10 +35,7 @@ struct Harness {
 
 /// Every environment variable any media driver consults for credentials.
 ///
-/// Kept as one list rather than scattered per-test removals so that adding a
-/// provider to `librefang-runtime-media` and forgetting this array is the only
-/// way to regress — grep for `_API_KEY` under `crates/librefang-runtime-media/src/`
-/// to re-derive it.
+/// Kept as one list rather than scattered per-test removals so that adding a provider to `librefang-runtime-media` and forgetting this array is the only way to regress — grep for `_API_KEY` under `crates/librefang-runtime-media/src/` to re-derive it.
 const MEDIA_PROVIDER_KEY_VARS: &[&str] = &[
     "ANTHROPIC_API_KEY",
     "ELEVENLABS_API_KEY",
@@ -60,25 +54,14 @@ const MEDIA_PROVIDER_KEY_VARS: &[&str] = &[
 
 /// Strip provider credentials from this test binary's environment.
 ///
-/// The module doc above says these tests run "with no media-provider API keys
-/// configured", but that was an assumption about the developer's shell rather
-/// than something the harness enforced. With `OPENAI_API_KEY` exported, tests
-/// whose names assert the *absence* of a provider stopped testing the
-/// missing-key path and instead made real, billable calls to OpenAI —
-/// `media_speech_no_configured_provider_returns_missing_key` asserted 422 and
-/// got 200 back with a generated mp3. A test that asserts "no configured
-/// provider" must be structurally incapable of reaching one.
+/// The module doc above says these tests run "with no media-provider API keys configured", but that was an assumption about the developer's shell rather than something the harness enforced.
+/// With `OPENAI_API_KEY` exported, tests whose names assert the *absence* of a provider stopped testing the missing-key path and instead made real, billable calls to OpenAI — `media_speech_no_configured_provider_returns_missing_key` asserted 422 and got 200 back with a generated mp3.
+/// A test that asserts "no configured provider" must be structurally incapable of reaching one.
 ///
-/// Every driver reads its key through `std::env::var` at call time
-/// (`openai.rs:42`, `:281`, and the equivalents in each sibling module), so
-/// clearing the process environment before the first request is what makes the
-/// assertion true rather than merely likely. This is an integration test, so
-/// the process is this test binary alone and nothing outside it is affected.
+/// Every driver reads its key through `std::env::var` at call time (`openai.rs:42`, `:281`, and the equivalents in each sibling module), so clearing the process environment before the first request is what makes the assertion true rather than merely likely.
+/// This is an integration test, so the process is this test binary alone and nothing outside it is affected.
 ///
-/// `Once` rather than a per-`boot()` loop because every test reaches its
-/// provider through a request issued after its own `boot()` returns, so a
-/// single clear that completes before the first harness is handed out covers
-/// all of them.
+/// `Once` rather than a per-`boot()` loop because every test reaches its provider through a request issued after its own `boot()` returns, so a single clear that completes before the first harness is handed out covers all of them.
 fn clear_media_provider_env() {
     static ONCE: std::sync::Once = std::sync::Once::new();
     ONCE.call_once(|| {
@@ -90,11 +73,9 @@ fn clear_media_provider_env() {
 
 /// The guard for the guard: `boot()` must leave the environment credential-free.
 ///
-/// Every other test in this file asserts a 4xx that is only meaningful when no
-/// provider is reachable. Those assertions can pass for the wrong reason — a
-/// bogus key set in the developer's shell produces a provider-side 401, which
-/// several of them would also accept. This one fails outright if the clearing
-/// stops happening, which is what makes the rest trustworthy.
+/// Every other test in this file asserts a 4xx that is only meaningful when no provider is reachable.
+/// Those assertions can pass for the wrong reason — a bogus key set in the developer's shell produces a provider-side 401, which several of them would also accept.
+/// This one fails outright if the clearing stops happening, which is what makes the rest trustworthy.
 #[tokio::test(flavor = "multi_thread")]
 async fn boot_clears_provider_credentials_from_the_environment() {
     let _h = boot().await;
--- a/crates/librefang-api/tests/media_routes_integration.rs
+++ b/crates/librefang-api/tests/media_routes_integration.rs
@@ -1,7 +1,10 @@
 //! Integration tests for the `/media/*` HTTP surface.
 //!
 //! These exercise the real `media` router against a freshly-booted kernel
-//! with no media-provider API keys configured. We focus on:
+//! with no media-provider API keys configured — `boot()` strips them from the
+//! process environment rather than assuming the developer's shell has none, so
+//! a test asserting the missing-key path cannot silently make a billable call.
+//! We focus on:
 //!
 //! 1. Validation paths in each `MediaError`-producing handler — the cheap,
 //!    deterministic 4xx slice. No live network, no real provider keys.
@@ -33,7 +36,79 @@ struct Harness {
     _test: TestAppState,
 }
 
+/// Every environment variable any media driver consults for credentials.
+///
+/// Kept as one list rather than scattered per-test removals so that adding a
+/// provider to `librefang-runtime-media` and forgetting this array is the only
+/// way to regress — grep for `_API_KEY` under `crates/librefang-runtime-media/src/`
+/// to re-derive it.
+const MEDIA_PROVIDER_KEY_VARS: &[&str] = &[
+    "ANTHROPIC_API_KEY",
+    "ELEVENLABS_API_KEY",
+    "FIREWORKS_API_KEY",
+    "GEMINI_API_KEY",
+    "GOOGLE_API_KEY",
+    "GOOGLE_CLOUD_API_KEY",
+    "GROQ_API_KEY",
+    "LOCAL_WHISPER_KEY",
+    "MINIMAX_API_KEY",
+    "MINIMAX_CN_API_KEY",
+    "OPENAI_API_KEY",
+    "SILICONFLOW_API_KEY",
+    "TOGETHER_API_KEY",
+];
+
+/// Strip provider credentials from this test binary's environment.
+///
+/// The module doc above says these tests run "with no media-provider API keys
+/// configured", but that was an assumption about the developer's shell rather
+/// than something the harness enforced. With `OPENAI_API_KEY` exported, tests
+/// whose names assert the *absence* of a provider stopped testing the
+/// missing-key path and instead made real, billable calls to OpenAI —
+/// `media_speech_no_configured_provider_returns_missing_key` asserted 422 and
+/// got 200 back with a generated mp3. A test that asserts "no configured
+/// provider" must be structurally incapable of reaching one.
+///
+/// Every driver reads its key through `std::env::var` at call time
+/// (`openai.rs:42`, `:281`, and the equivalents in each sibling module), so
+/// clearing the process environment before the first request is what makes the
+/// assertion true rather than merely likely. This is an integration test, so
+/// the process is this test binary alone and nothing outside it is affected.
+///
+/// `Once` rather than a per-`boot()` loop because every test reaches its
+/// provider through a request issued after its own `boot()` returns, so a
+/// single clear that completes before the first harness is handed out covers
+/// all of them.
+fn clear_media_provider_env() {
+    static ONCE: std::sync::Once = std::sync::Once::new();
+    ONCE.call_once(|| {
+        for var in MEDIA_PROVIDER_KEY_VARS {
+            std::env::remove_var(var);
+        }
+    });
+}
+
+/// The guard for the guard: `boot()` must leave the environment credential-free.
+///
+/// Every other test in this file asserts a 4xx that is only meaningful when no
+/// provider is reachable. Those assertions can pass for the wrong reason — a
+/// bogus key set in the developer's shell produces a provider-side 401, which
+/// several of them would also accept. This one fails outright if the clearing
+/// stops happening, which is what makes the rest trustworthy.
+#[tokio::test(flavor = "multi_thread")]
+async fn boot_clears_provider_credentials_from_the_environment() {
+    let _h = boot().await;
+    for var in MEDIA_PROVIDER_KEY_VARS {
+        assert!(
+            std::env::var(var).is_err(),
+            "{var} is still set after boot() — a test asserting the missing-key path \
+             could reach a real provider and bill the developer"
+        );
+    }
+}
+
 async fn boot() -> Harness {
+    clear_media_provider_env();
     let test = TestAppState::with_builder(MockKernelBuilder::new().with_config(|cfg| {
         cfg.default_model = librefang_types::config::DefaultModelConfig {
             provider: "ollama".to_string(),
--- a/crates/librefang-api/tests/media_routes_integration.rs
+++ b/crates/librefang-api/tests/media_routes_integration.rs
@@ -65,8 +65,16 @@ const MEDIA_PROVIDER_KEY_VARS: &[&str] = &[
 fn clear_media_provider_env() {
     static ONCE: std::sync::Once = std::sync::Once::new();
     ONCE.call_once(|| {
-        for var in MEDIA_PROVIDER_KEY_VARS {
-            std::env::remove_var(var);
+        // SAFETY: runs inside `Once::call_once`, which guarantees this body
+        // executes exactly once and completes before any other caller of
+        // `clear_media_provider_env` proceeds — including the first `boot()`,
+        // which is the only place any test in this file spawns work that
+        // reads the environment. No other thread reads or writes these vars
+        // concurrently with this block.
+        unsafe {
+            for var in MEDIA_PROVIDER_KEY_VARS {
+                std::env::remove_var(var);
+            }
         }
     });
 }
--- a/crates/librefang-runtime-media/src/media_understanding.rs
+++ b/crates/librefang-runtime-media/src/media_understanding.rs
@@ -1808,15 +1808,21 @@ mod tests {
         assert_eq!(default_vision_model("unknown"), "unknown");
     }
 
+    /// Same latent dependency as `test_transcribe_audio_no_provider` above: `describe_image` resolves the provider (`config.image_provider`, else `detect_vision_provider()`, which reads vision-capable keys from the process) BEFORE it reads the file.
+    /// With no explicit `image_provider` set, a machine with a vision-capable key exported resolves a real provider here; only a guaranteed-absent input path keeps the call from proceeding to a real, billable description request.
     #[tokio::test]
     async fn test_describe_image_no_provider_configured() {
-        // With no API keys set and no explicit provider, should fail with provider error.
+        // With no API keys set, should fail with provider error.
+        // With a key set, provider resolution succeeds instead and the call must still stop at the guaranteed-absent file read, never reaching a real description request.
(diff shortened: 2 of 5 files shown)


crates/librefang-runtime-media/src/openai.rs
```diff
@@ -48,6 +48,16 @@ impl OpenAIMediaDriver {
     }
 }
 
+/// Whether `/images/generations` accepts a `response_format` parameter for this model.
+///
+/// The DALL·E models take `response_format` as `"url"` or `"b64_json"`.
```
houko, line 53: Not about this line specifically, but about the commit that reformatted it: `30a7d246e35294e7e3674c66657b2d3fd9e9cd19` (\"docs: rewrap new media doc-comments to one sentence per line\") has author **and** committer identity `Claude <noreply@anthropic.com>`:

```
$ git log -1 --format='%an <%ae> / %cn <%ce>' 30a7d24
Claude <noreply@anthropic.com> / Claude <noreply@anthropic.com>
```

CLAUDE.md's git conventions are explicit that this is disallowed project-wide (\"No AI / Claude / Anthropic attribution in commit messages, PR bodies, or comments\"), and the repo's own `commit-msg` hook is specifically designed to reject exactly this case: \"rejects a commit whose *author identity* (`git var GIT_AUTHOR_IDENT` ...) resolves to Claude / Anthropic even when the message itself is clean.\" The message body here is clean, but the identity itself trips the rule — this commit must have been produced through a path that bypassed the hook (e.g. `core.hooksPath` not configured, or a non-git commit path).

Since fixing this requires rewriting this commit (interactive rebase to amend the author/committer identity) and then force-pushing `fix/media-test-isolation`, and force-push here is out of scope for an automated review pass, this needs a maintainer action rather than an automated fix: `git rebase -i --root` (or `git commit --amend --author=...` if it's the tip) with `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` (and committer equivalents) set to your real identity, then force-push before this lands. Given the branch has no reviews yet, force-pushing your own branch here is safe per CLAUDE.md's \"force-push only to your own branches, only before review\" rule.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1659,18 +1659,28 @@ mod tests {
         assert!(result.unwrap_err().contains("Expected image"));
     }
 
-    /// #6679: `transcribe_audio` must accept `MediaType::Video` (it used to
-    /// reject every non-`Audio` type outright) — asserted by checking the
-    /// call gets PAST the type guard and fails on provider resolution
-    /// instead, which needs no ffmpeg and no configured provider.
+    /// #6679: `transcribe_audio` must accept `MediaType::Video` (it used to reject every non-`Audio` type outright) — asserted by checking the call gets PAST the type guard, which needs no ffmpeg and reaches no network.
```
houko, line 1662: Commit `30a7d246e35294e7e3674c66657b2d3fd9e9cd19` ("docs: rewrap new media doc-comments to one sentence per line"), which touches this file, is authored and committed as `Claude <noreply@anthropic.com>`.

CLAUDE.md is explicit that this is not allowed: "No AI / Claude / Anthropic attribution in commit messages, PR bodies, or issue comments," and separately that the `commit-msg` hook "rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean." This commit is already on `origin/fix/media-test-isolation`, so fixing the author identity requires rewriting this commit (changing its hash) and force-pushing the branch — both of which are outside what an automated reviewer should do unilaterally on someone else's branch.

Recommend a maintainer reauthor it locally and force-push once, e.g.:
```
git rebase -i <base> # or: git commit --amend --author="houko <12625278+houko@users.noreply.github.com>"
git push --force-with-lease origin fix/media-test-isolation
```

Separately, and lower-priority: that same commit's own message body is hard-wrapped at ~65-70 columns ("CLAUDE.md's prose-wrapping rule has no column limit and requires\nline breaks only at sentence boundaries..."), which is itself a violation of the no-column-wrap / one-sentence-per-line rule the commit's subject claims to be enforcing.

---

crates/librefang-api/tests/media_routes_integration.rs
```diff
@@ -33,7 +33,63 @@ struct Harness {
     _test: TestAppState,
 }
 
+/// Every environment variable any media driver consults for credentials.
```
houko, line 36: The doc comment claims `MEDIA_PROVIDER_KEY_VARS` covers "every environment variable any media driver consults for credentials," rederivable by grepping `_API_KEY` under `crates/librefang-runtime-media/src/`. That grep misses one credential path: `GenericOpenAICompatMediaDriver` (openai.rs) derives its key's env var name *dynamically* per configured custom provider — `let api_key_env = format!("{}_API_KEY", provider.to_uppercase().replace('-', "_"))` (openai.rs:285, e.g. `VOLCENGINE_API_KEY` per the comment at openai.rs:273) — so there is no fixed literal for a static grep or a static list to catch.

Currently harmless: this test file's `boot()` only configures `provider: "ollama"` for `default_model` and no custom OpenAI-compatible media provider, so `GenericOpenAICompatMediaDriver` is never exercised here and no dynamically-named var is ever read. But the guard test's assertion ("Every environment variable any media driver consults") is not actually exhaustive against that driver, and a future test that does configure a custom compat provider (or a config default that adds one) could silently reintroduce exactly the live-billing risk this PR is fixing, without `boot_clears_provider_credentials_from_the_environment` catching it. Worth either a comment caveat here noting the dynamic-key exception, or a follow-up that also clears any `*_API_KEY` var matching a configured custom provider name.

---

houko (commented):
Reviewed against CLAUDE.md's engineering rules.
The code changes look correct and compliant: changelog fragments in `changelog.d/fixed/` carry proper `(#6750) (@houko)` attribution and pass `check-changelog-attribution.py --all-unreleased`, the doc-comments added in this PR are already reflowed to one-sentence-per-line, and the fixes are wired into the actual call sites they claim to fix.
On that last point: I checked every remaining hardcoded `"test.png"` / `"test.mp4"` / etc. path left in `media_understanding.rs` (not just the three switched to `temp_dir()`-based absent paths) and confirmed each one fails `MediaAttachment::validate()` or the media-type guard before provider resolution is ever reached, so none of them is a leftover real-provider gap.
`image_model_accepts_response_format` is applied at both `OpenAIMediaDriver` and `GenericOpenAICompatMediaDriver` call sites using the same `model` variable that goes into the request body, so the gpt-image fix isn't partial either.

One thing I'm leaving for a maintainer rather than fixing: three of the five commit message bodies (99b64aad, 167fd6c038e, 30a7d246e) are hard-wrapped at roughly 70-80 columns with multiple sentences per line and some sentences split mid-line, which is exactly the "commit message bodies" case CLAUDE.md's prose-wrapping rule calls out by name.
Commit 30a7d246e is itself titled "rewrap new media doc-comments to one sentence per line" and its body explains this same rule, so its own message not following it stands out.
I didn't rewrite these — fixing already-pushed commit text means rewriting history, and force-push is out of scope for this review — so flagging it here for you to squash/reword if you want the history clean before merge.

---

crates/librefang-api/tests/media_routes_integration.rs
```diff
@@ -33,7 +33,71 @@ struct Harness {
...
+fn clear_media_provider_env() {
+    static ONCE: std::sync::Once = std::sync::Once::new();
+    ONCE.call_once(|| {
+        // SAFETY: runs inside `Once::call_once`, which guarantees this body
+        // executes exactly once and completes before any other caller of
+        // `clear_media_provider_env` proceeds — including the first `boot()`,
+        // which is the only place any test in this file spawns work that
+        // reads the environment. No other thread reads or writes these vars
+        // concurrently with this block.
+        unsafe {
```
houko, line 70: Two commits landed after the previous review pass on this PR repeat both issues already flagged on `30a7d24` (see the two earlier threads on `openai.rs` and `media_understanding.rs`):

```
$ git log --format='%h %an <%ae> / %cn <%ce>' -1 167fd6c
167fd6c Claude <noreply@anthropic.com> / Claude <noreply@anthropic.com>
$ git log --format='%h %an <%ae> / %cn <%ce>' -1 99b64aa
99b64aa Claude <noreply@anthropic.com> / Claude <noreply@anthropic.com>
```

1. **Author/committer identity is `Claude <noreply@anthropic.com>`** on both — the same violation of CLAUDE.md's "no AI/Claude/Anthropic attribution" rule and the same thing the repo's `commit-msg` hook is designed to reject via `GIT_AUTHOR_IDENT`. This unsafe block itself is introduced by `99b64aa`, one of the two.
2. **Both commit message bodies are hard-wrapped at ~70-80 columns** with multiple sentences per line (e.g. `167fd6c`'s body runs one sentence across 5 wrapped lines), which is the same one-sentence-per-line violation called out on `30a7d24`'s message.

Net: this branch now has three commits (`30a7d24`, `167fd6c`, `99b64aa`) needing the same reauthor. Fixing requires rewriting history and force-pushing this branch, which is out of scope for an automated pass — flagging so a single rebase covers all three rather than trickling in one at a time:
```
git rebase -i <base>   # reword/reauthor all three, or squash them
git push --force-with-lease origin fix/media-test-isolation
```
Since the branch has no approving reviews yet, a force-push to your own branch is within CLAUDE.md's "force-push only to your own branches, only before review" allowance.

---

houko (commented):
Reviewed the current diff plus the code paths it touches (gpt-image response_format handling in both OpenAI drivers, the env-clearing test harness, and the three fixed unit tests in media_understanding.rs). Ran `cargo check -p librefang-runtime-media --lib` clean. Cross-checked `MEDIA_PROVIDER_KEY_VARS` against every literal `_API_KEY`/`_KEY` read in the crate — the static list is complete for literal env-var names (the one gap, the dynamically-named key for `GenericOpenAICompatMediaDriver`, was already flagged in an earlier pass on this PR). Verified `image_model_accepts_response_format` is scoped to the `gpt-image` prefix only and does not change behavior for DALL-E or unrecognised third-party models. Checked the other unit tests in this crate that assert "no provider configured" or resolve providers pre-network (elevenlabs.rs, gemini.rs, google_tts.rs) — they already guard correctly (either fail before any env read, or deliberately set a bogus key and assert failure before the network call), so no further real-provider-reachability gaps found beyond what this PR already closes.

One new finding below, left as a review comment since fixing it requires a force-push this automated pass won't perform.

---

crates/librefang-runtime-media/src/openai.rs
```diff
@@ -48,6 +48,16 @@ impl OpenAIMediaDriver {
...
+/// The `gpt-image-*` family rejects the parameter outright with `400 Unknown parameter: 'response_format'` and always returns base64, so sending it makes every image request fail against those models.
+///
+/// Defaulting to `true` for unrecognised names keeps third-party OpenAI-compatible endpoints on the behaviour they have today; only the family known to reject it is special-cased.
+fn image_model_accepts_response_format(model: &str) -> bool {
```
houko, line 57: Scope question, not a code objection: this PR's title and diff bundle two independently-shippable fixes — (1) hardening the media integration/unit tests so they can't reach real providers, and (2) this `response_format` gating for `gpt-image-*`. CLAUDE.md's "One PR ↔ one issue (or one tight cluster)" convention allows a tight cluster, and the PR body's "Scope note" does connect them causally (the billable-call investigation that motivated the test hardening is what surfaced the `400 Unknown parameter` response_format defect). That's a reasonable justification, but the two changes are still independently mergeable/revertable — the test-isolation fix stands on its own with zero production code touched, and the response_format fix is a plain provider-behavior bug fix that doesn't depend on the test changes at all. Flagging for maintainer judgment: is the shared discovery story enough to keep these as one PR, or would splitting the `openai.rs` production fix into its own PR make the history/revert story cleaner? Not asking for a split myself — just surfacing the ambiguity per repo convention.

---

houko (commented):
Reviewed the current diff against CLAUDE.md's conventions (changelog fragments, prose-wrapping, config-field rules, gitleaks-adjacent secret check, and the "one PR ↔ one issue" bundling rule).

Pushed one mechanical fix (commit 7a015b5, human-authored — see note below): the `Once::call_once` `SAFETY` comment in `media_routes_integration.rs` was still hard-wrapped across two sentences at ~80 columns, missed by the earlier pass (`30a7d24`) that rewrapped this file's doc comments to CLAUDE.md's one-sentence-per-line rule. No other prose-wrap violations found in the touched doc-comments — the rest are already correctly one-sentence-per-line.

Other checks: changelog fragments in `changelog.d/fixed/` are correctly new files (not edits to CHANGELOG.md), end with `(#6750) (@houko)`, and read as prose explaining *why*. No config-struct fields were added, so the `#[serde(default)]`/Default-impl rule doesn't apply here. No hardcoded API keys or secret-shaped strings found in the touched files. Left one ambiguous scope comment inline (bundling of the two fixes) per the task's instruction to flag rather than split it myself.

Not re-flagging here: the pre-existing, already-repeatedly-flagged finding that commits `30a7d246`, `167fd6c0`, and `99b64aad` carry `Claude <noreply@anthropic.com>` author/committer identity (violates CLAUDE.md's no-AI-attribution rule and the repo's own `commit-msg` hook, which wasn't active for whichever pass produced those commits since `core.hooksPath` wasn't configured). That's already documented on four separate threads on this PR and needs a maintainer-driven rebase + force-push, which is outside what an automated pass should do. Enabled `core.hooksPath` in my own clone before committing, and used an explicit human git identity for my commit specifically to avoid reproducing that same issue a fourth time.

---

crates/librefang-api/tests/media_routes_integration.rs
```diff
@@ -33,7 +33,67 @@ struct Harness {
     _test: TestAppState,
 }
 
+/// Every environment variable any media driver consults for credentials.
+///
+/// Kept as one list rather than scattered per-test removals so that adding a provider to `librefang-runtime-media` and forgetting this array is the only way to regress — grep for `_API_KEY` under `crates/librefang-runtime-media/src/` to re-derive it.
+const MEDIA_PROVIDER_KEY_VARS: &[&str] = &[
```
houko, line 39: Nice fix for the immediate problem, but I think the safety net has one gap worth a second look.

`MEDIA_PROVIDER_KEY_VARS` is derived by grepping for hardcoded `"..._API_KEY"` literals, which covers every *built-in* driver (openai.rs, gemini.rs, elevenlabs.rs, etc.) — I checked, and it's a complete and accurate enumeration of those (12 literals + `LOCAL_WHISPER_KEY`).

But `GenericOpenAICompatMediaDriver::new()` (openai.rs) derives its credential env var **dynamically** from a `provider_urls` config entry: `format!("{}_API_KEY", provider.to_uppercase()...)`. A test that registers a custom `provider_urls` entry (e.g. `volcengine`) and exercises it would read `VOLCENGINE_API_KEY` from the environment — a name this static list can never contain, no matter how carefully it's kept in sync with `grep _API_KEY`.

Today this isn't exploitable: `boot()`'s `MockKernelBuilder` config in this file doesn't set any `provider_urls`, so no test actually instantiates that driver. But the doc comment above ("adding a provider ... and forgetting this array is the only way to regress") overstates the guarantee — a future test exercising a custom provider wouldn't "forget the array," it would hit a gap the array is structurally unable to cover.

Worth either a comment noting this class of gap explicitly, or a small addition (e.g. `clear_media_provider_env` also strips any env var matching `_API_KEY$`/`_KEY$`, or the harness refuses to construct `GenericOpenAICompatMediaDriver` in tests without going through a helper that registers its var here too). Not blocking this PR, since it fixes a real live bug — just flagging so it doesn't quietly become the next version of the same problem.

---

houko (commented):
Reviewed the diff against CLAUDE.md. The core fixes look correct — traced `image_model_accepts_response_format` through both call sites (types check out, TTS's separate `response_format` at :187 is untouched and is a different, valid parameter), and independently re-derived `MEDIA_PROVIDER_KEY_VARS` from `grep _API_KEY` across `librefang-runtime-media/src/` — it's a complete and accurate list for every built-in driver, and `is_configured()`/`select_driver` in `lib.rs:279` already gate driver selection on it, so clearing these before `boot()` genuinely prevents the route handlers from reaching a network call rather than just swapping a URL. Left one inline comment on a narrower gap in that safety net (custom `provider_urls` drivers derive their env var name dynamically, so the static list can't cover them — not exploitable today since no test in this file registers one). Changelog fragments and their format check out.

One non-blocking, non-code note: three commit bodies in this branch (`99b64aa`, `167fd6c`, `30a7d24`) hard-wrap prose at ~80 columns with line breaks mid-sentence, which is what CLAUDE.md's prose rule (no column limit, break only at sentence boundaries, explicitly including "commit message bodies") asks not to do — a bit ironic since two of those very commits are dedicated to fixing that exact issue in the code's doc-comments. I'm not raising this as something to fix via history rewrite; just flagging in case a squash-merge commit message is going to be hand-written anyway.

---

crates/librefang-api/tests/media_routes_integration.rs
houko, line 1: Three commits on this branch have author identity `Claude <noreply@anthropic.com>`: 30a7d24 ("docs: rewrap new media doc-comments to one sentence per line"), 167fd6c ("fix(media): close two more test/real-provider gaps in media_understanding"), and 99b64aa ("fix(media): wrap env::remove_var in unsafe in clear_media_provider_env").
None of the messages themselves contain a banned phrase, but CLAUDE.md's `commit-msg` hook is specifically written to reject this independent of message text: "rejects a commit whose *author identity* … resolves to Claude / Anthropic even when the message itself is clean."
These evidently landed without `core.hooksPath` enforcing that locally.
Worth rewriting the author on those three commits (interactive rebase + `--reset-author`, or a maintainer squash-merge under their own identity) before merge — one of them (167fd6c) is also a substantive fix (the two guaranteed-absent-path tests), not just a docs pass, so squashing needs to preserve that content rather than drop it.

---

crates/librefang-runtime-media/src/openai.rs
houko, line 1: Nit on the new harness's doc comment (`media_routes_integration.rs`, on `MEDIA_PROVIDER_KEY_VARS`): it says the array covers "every environment variable any media driver consults for credentials" and can be "re-derived" by grepping this crate for `_API_KEY`.
That's not quite true for `GenericOpenAICompatMediaDriver` here — `new()` (line 285) builds its credential var name at runtime from the operator-configured `provider` string (`format!("{}_API_KEY", provider.to_uppercase()...)`), so the real set of names is open-ended and config-dependent, not something a literal grep enumerates.

It doesn't affect this PR: `boot()` never configures a custom OpenAI-compatible provider, so `image_provider`/`audio_provider` resolution in these tests only ever walks the fixed `detect_*_provider()` chain that the array does fully cover.
But the comment's "grep to re-derive" instruction would give a future maintainer false confidence if they add a test that does exercise a custom provider config — worth a caveat noting the generic-compat driver is exempt from that methodology, so nobody trusts the array to protect that path too.

---

houko (commented):
Automated daily review. Code itself checks out: the env-var clearing list matches every literal `env::var("*_API_KEY")` call site in the crate, the two guaranteed-absent-path test fixes are correct and don't reintroduce host-filesystem coupling, the `response_format` gating is complete (both `OpenAIMediaDriver` and `GenericOpenAICompatMediaDriver` call sites fixed, no other call site in the workspace sends it for an image-gen request), the changelog fragments are correctly formatted, and `cargo fmt --check` / `cargo check -p librefang-runtime-media --lib` are clean on this branch. Two things flagged inline: three AI-authored commits in the branch history, and a doc-comment overstatement about what the credential-clearing array can guarantee.

---

crates/librefang-runtime-media/src/openai.rs
```diff
@@ -326,8 +339,10 @@ impl MediaDriver for GenericOpenAICompatMediaDriver {
             "prompt": request.prompt,
             "n": request.count,
             "size": size,
-            "response_format": "b64_json",
         });
+        if image_model_accepts_response_format(model) {
```
houko, line 343: Design question, not a blocker: `image_model_accepts_response_format` is shared between the real OpenAI driver and `GenericOpenAICompatMediaDriver`, so any user-configured third-party endpoint (via `provider_urls`) whose model name happens to start with `gpt-image` also loses `response_format` here — even if that particular endpoint's `gpt-image*`-named model actually behaves like DALL-E and expects it. That's presumably fine if third-party providers proxying a `gpt-image*` model always inherit the same quirk, but it's an assumption worth confirming rather than inferring from the OpenAI-specific doc comment on the function. If it turns out to be OpenAI-only behavior, this predicate may need a provider-scoped variant instead of a model-name-only one.

---

crates/librefang-api/tests/media_routes_integration.rs
```diff
@@ -33,7 +33,67 @@ struct Harness {
...
+    "ELEVENLABS_API_KEY",
+    "FIREWORKS_API_KEY",
+    "GEMINI_API_KEY",
+    "GOOGLE_API_KEY",
+    "GOOGLE_CLOUD_API_KEY",
+    "GROQ_API_KEY",
+    "LOCAL_WHISPER_KEY",
+    "MINIMAX_API_KEY",
+    "MINIMAX_CN_API_KEY",
+    "OPENAI_API_KEY",
```
houko, line 50: `MEDIA_PROVIDER_KEY_VARS` is a complete list of every statically-named `_API_KEY` env var the drivers read (verified by grep) — good.
One gap this list can't close: `GenericOpenAICompatMediaDriver` derives its key var name at runtime from `provider_urls` config (`format!("{}_API_KEY", provider.to_uppercase()...)` in `openai.rs`), so a hypothetical future test that configures a custom `provider_urls` entry (e.g. `VOLCENGINE_API_KEY`) would not have that variable cleared by this harness.
No current test in this file exercises that path, so it's not a live bug today, but the doc comment's claim that "forgetting this array is the only way to regress" is slightly overstated — a dynamically-named provider key is a second way.
Worth a one-line caveat in the doc comment, or leaving as-is if the custom-provider path is considered out of scope for this harness.

---

Issue #6679, opened 2026-07-30
media_transcribe rejects video containers by extension, though the audio track transcribes fine


nevgenov:
## Summary

`media_transcribe` and `speech_to_text` reject video containers on the strength of the filename extension alone, even though the audio track behind them transcribes fine and the type system already models video with its own size budget.

```
Error: Invalid parameter 'path': Unsupported audio format: .mp4
```

`crates/librefang-runtime/src/tool_runner/media.rs`:

```rust
pub(super) fn audio_mime_from_ext(ext: &str) -> Option<&'static str> {
    match ext {
        "mp3" => Some("audio/mpeg"),
        "wav" => Some("audio/wav"),
        "ogg" => Some("audio/ogg"),
        "oga" => Some("audio/oga"),
        "flac" => Some("audio/flac"),
        "m4a" => Some("audio/mp4"),
        "webm" => Some("audio/webm"),
        _ => None,
    }
}
```

## The filter reads the filename, not the file

Three calls, all on files that contain a video track:

| Input | Result |
|---|---|
| `clip.mp4` (H.264 + AAC) | `Unsupported audio format: .mp4` |
| the same bytes copied to `clip.m4a` | transcribed correctly and in full |
| a genuine `.webm` (VP9 + Opus) | **transcribed correctly** |

The `.webm` row is the interesting one.
`webm` is on the allowlist because it is also an audio container, so video-in-webm already works today and has presumably been working all along.
What fails is only the set of extensions that are *exclusively* video — `mp4`, `mov`, `mkv`, `avi`.

So the behaviour is not "video transcription is unimplemented".
It is one allowlist that happens to admit video through a side door and turn it away at the front.
Renaming the file is enough to get a correct transcript, which is a good sign that the gate is the only obstacle.

## The pieces are already in the tree

- `MediaType::Video` exists, with its own `MAX_VIDEO_BYTES` (50 MB) alongside `MAX_AUDIO_BYTES` (20 MB) in `crates/librefang-types/src/media.rs`.
- `transcode_oga_to_ogg_opus()` in `media_understanding.rs` already spawns `ffmpeg`, pipes bytes through it, caps the run at 30 s, degrades with a clear "install ffmpeg" message when the binary is missing, and has a test that skips itself via `ffmpeg_available()` on CI images without it.

Adding `-vn` to that existing invocation drops the video stream and leaves exactly the audio the transcriber wants.

## Proposed change

[...]

PR #6751, opened 2026-08-08
fix(media): stage video input to a file so mp4/mov audio extraction can seek

### LINKED ISSUE #6747: media_transcribe silently returns an empty audio track for .mp4 / .mov
## Symptom

`media_transcribe` / `speech_to_text` on an ordinary `.mp4` or `.mov` produces no usable transcript.
Depending on the STT provider the call either fails with an opaque downstream error or returns an empty string — in neither case does the error name the actual cause.
`.mkv` and `.avi`, added by the same PR, work correctly. `.webm` also works (it never enters the extraction path).

So the four extensions #6679 / #6683 enabled split cleanly in half, and the two that fail are the two that dominate real recordings.

## Root cause (read from current `main`, c403439a)

`extract_video_audio_track` feeds the container bytes to ffmpeg over a **non-seekable pipe**:

```rust
// crates/librefang-runtime-media/src/media_understanding.rs:1153
async fn extract_video_audio_track(input_bytes: &[u8]) -> Result<Vec<u8>, String> {
    run_ffmpeg_pipe(
        &[..., "-i", "pipe:0", "-vn", "-c:a", "libopus", ..., "-f", "ogg", "pipe:1"],
```

An ISO-BMFF file (`mp4`, `mov`) stores its index in the `moov` atom.
When `moov` sits *after* `mdat` — which is what ffmpeg's own muxer, phone cameras, screen recorders and meeting-recording exporters all produce by default — the demuxer must seek backwards to read it.
A pipe cannot seek, so the demux fails.

`mkv` and `avi` are streamable and are unaffected.
`+faststart` mp4 (index relocated to the front, an opt-in flag used mainly for web delivery) is also unaffected.

**The failure is silent because it passes both guards in `run_ffmpeg_pipe`.**
On ffmpeg 8.1 the process writes `partial file` / `Error during demuxing: Invalid data found when processing input` to stderr but **exits 0**, and emits a 261-byte Ogg — headers, zero audio packets:

```rust
// crates/librefang-runtime-media/src/media_understanding.rs:1106-1115
if !status.success() {          // <- exit code is 0, guard does not fire
    return Err(...);
}
if out.is_empty() {             // <- 261 bytes is not empty, guard does not fire
    return Err("ffmpeg produced an empty output stream".to_string());
}
Ok(out)
```

Those 261 bytes are then uploaded to the transcription provider, and whatever the provider says about a soundless file is what the operator sees — arbitrarily far from the real cause.

(Exit-code behaviour may differ across ffmpeg builds. If another version returns non-zero, the outcome is a hard error instead of an empty transcript — the demux failure itself is inherent to a non-seekable input either way.)

## Why the test suite is green

`extract_video_audio_track_smoke` builds its fixture with:

```rust
// crates/librefang-runtime-media/src/media_understanding.rs:1476
"-movflags",
"frag_keyframe+empty_moov",
```

That is a **fragmented** MP4 — the one MP4 flavour that *is* streamable from a pipe.
The test therefore exercises the only shape that works.

The assertions would not catch the bug even with an ordinary fixture: a 261-byte Ogg is non-empty, starts with `OggS`, and is smaller than the source mp4, so all three pass.

## Measurements

Verbatim `extract_video_audio_track` arguments, ffmpeg 8.1, synthetic files generated with `-f lavfi`:

| Input | In | Ogg out | Audio duration in output |
|---|---|---|---|
| fragmented mp4 (the test's own fixture shape) | 8 234 B | 2 641 B | 0.72 s |
| ordinary mp4, 3 min | 10 559 234 B | **261 B** | **none** |
| ordinary mov, 3 min | 10 559 285 B | **259 B** | **none** |
| mkv, 3 min | 10 560 492 B | 1 127 667 B | 180.03 s |
| avi, 3 min | 6 273 204 B | 1 083 467 B | 180.04 s |
| ordinary mp4, 15 min / 50.3 MB | 52 705 182 B | **261 B** | **none** |
| same bytes, `+faststart` | 52 705 182 B | 5 497 289 B | 900.01 s |

## Proposed fix

**Minimal:** give ffmpeg a seekable input instead of a pipe.

`MediaEngine::transcribe_audio` already accepts `MediaSource::FilePath` and reads it itself (`media_understanding.rs:232`), and the tool layer has the resolved, sandbox-checked path in hand at the point where it currently reads the bytes.
Building the video attachment as `FilePath` rather than `Base64` — and passing that path to ffmpeg as `-i <path>` — makes the demuxer seekable and fixes every affected container.

That change also removes an incidental cost on the same path: today the file is read into a `Vec`, base64-encoded (`tool_runner/media.rs:179-184`), decoded back (`media_understanding.rs:238-246`), then copied a third time for the ffmpeg stdin writer (`media_understanding.rs:1063`) — roughly 3.3x the file size resident at peak, for bytes that only ever travel in-process.

On a 50 MB input the file-path form took **2.1 s**, identical to the faststart-pipe form, so there is no throughput trade-off.

**Defensive (independent of the above):** `run_ffmpeg_pipe` currently treats "exit 0 with a plausible-looking but soundless stream" as success.
A check that the extracted stream actually carries audio — or simply surfacing ffmpeg's stderr when it is non-empty on a zero exit — would turn this class of failure into a named error instead of an empty transcript.

**Test:** the smoke test's fixture flags are what hide this. Dropping `-movflags frag_keyframe+empty_moov` (or adding a second case without it) reproduces the bug directly.

## Setup (for reference)

- Reproduced against `main` @ `c403439a` and tag `v2026.7.31`; the three relevant files are byte-identical between them.
- Self-hosted OpenAI-compatible Whisper endpoint via `[media.custom_stt]`; ffmpeg 8.1 on Apple Silicon.
- Source of the recordings in our setup is a meeting-recording exporter, so `.mp4` / `.webm` are the two containers we actually see; `.webm` masked this for us at first because it bypasses extraction entirely.
- Workaround on our side: an out-of-tree MCP tool that shells out to ffmpeg with a file path.

Happy to test a patch.

Refs: #6679 (video containers), #6683 (the PR that added the extraction path)

houko:
Fixes #6747. @nevgenov's diagnosis was right end to end and is what this implements; the one thing it turned out to need on top was a size threshold, described below, without which the regression test is green either way.

## The bug

`extract_video_audio_track` fed the container to ffmpeg over `pipe:0`. An `mp4` / `mov` keeps its index in the `moov` atom, and when `moov` sits after `mdat` — the default output of ffmpeg's own muxer, phone cameras, screen recorders and meeting exporters — the demuxer must seek backwards to read it. A pipe cannot seek, so the demux fails.

It fails *silently*. ffmpeg writes `stream 0, offset 0x30: partial file` to stderr and still exits 0, emitting a 261-byte Ogg containing headers and zero audio packets. Both guards in `run_ffmpeg_pipe` let it through — the exit code is success, and 261 bytes is not empty. Those bytes were uploaded to the transcription provider, so the operator saw whatever a provider says about a soundless file, arbitrarily far from the cause.

`mkv` and `avi` are streamable and were never affected, which is why the four containers enabled by #6679 / #6683 split exactly in half.

## Changes

**Stage the input to a file.** `ScopedTempFile` writes the bytes to a uniquely-named scratch file (pid + a process-wide counter, so concurrent extractions cannot collide) and removes it on drop. Written by hand rather than adding `tempfile` to this crate for one call site.

**`run_ffmpeg_pipe` now takes `Option<&[u8]>`.** `Some` feeds stdin as before — which is correct for `transcode_oga_to_ogg_opus`, since Ogg is streamable — and `None` means the args already name a path. The stdin handle is dropped in that case so ffmpeg does not wait on input that will never arrive.

**A headers-only guard.** `ogg_contains_audio` counts `OggS` page captures: an Opus stream opens with `OpusHead` and `OpusTags`, so audio begins on the third page. This is defence in depth, not the fix — but it is necessary, because the exit-code behaviour that made this silent varies across ffmpeg builds and cannot be relied on. On a build that returns non-zero the existing guard catches it; on 8.1.1 nothing did.

## The size threshold, which is not in the issue

My first regression fixture was the smoke test's 0.5 s clip with the fragmentation flag removed. It passed **with and without the fix** — the mutation check is what caught that, not review.

ffmpeg buffers the head of an unseekable input (32 KB by default), so a file small enough to fit in that buffer can still be seeked backwards and demuxes fine over a pipe. Measured against ffmpeg 8.1.1, feeding each file to `ffmpeg -i pipe:0 -vn -c:a libopus -f ogg pipe:1`:

| Clip | mp4 size | Ogg out | stderr |
| --- | --- | --- | --- |
| 0.5 s | 11,572 B | 2,500 B | — |
| 5 s | 68,977 B | **261 B** | `partial file` |
| 30 s | 388,664 B | **261 B** | `partial file` |
| 120 s | 1,550,323 B | **261 B** | `partial file` |

So the existing smoke test could not have caught this for **two** independent reasons: its fixture is fragmented (the one mp4 flavour that streams from a pipe), *and* at ~8 KB it is under the buffer threshold. Fixing only the first would have produced a test that still passed against the broken code.

The new fixture is 10 s at 320x240, around 130 KB. Its atom order was confirmed to be `ftyp / free / mdat / moov` — index last, the layout the issue describes.

## Verification

macOS aarch64, ffmpeg 8.1.1, native cargo, based on `origin/main` at c403439a.

- **Mutation check.** Reverting the fix (input back on `pipe:0`) makes `extract_video_audio_track_handles_non_fragmented_mp4_6747` fail with `ffmpeg produced an Ogg stream with no audio packets (261 bytes)` — the same 261 bytes reported in the issue, surfaced by the new guard instead of being uploaded. Restored and re-confirmed green. Note `extract_video_audio_track_smoke` stays green under that mutation, which is the issue's fragmented-fixture observation reproduced.
- **`cargo test -p librefang-runtime-media --lib`** — 92 passed, 0 failed, in an environment with all 13 provider credential variables cleared, which is what CI sees.
- **`cargo check --workspace --lib`** — exit 0.
- **`cargo clippy --workspace --all-targets -- -D warnings`** — exit 0, captured explicitly.
- **`cargo fmt --check`** — exit 0.
- `ogg_contains_audio` is pinned by its own unit test so the regression test cannot be satisfied by a guard that always returns true.

### Not verified

Only ffmpeg 8.1.1 on macOS was exercised. The issue notes exit-code behaviour may differ on other builds; the guard exists precisely so the outcome is a clear error rather than an empty transcript either way, but that path was not executed against a build that returns non-zero.

No real transcription was performed — the extracted Ogg is asserted to contain audio packets, not to transcribe correctly.

## Note for local runs

`transcribe_audio_accepts_video_type` fails on this branch for anyone with an STT key exported. That is a pre-existing condition on `main` and is fixed in #6750; it does not reproduce in a clean environment, so it does not affect CI here.

### DIFF
--- a/crates/librefang-runtime-media/Cargo.toml
+++ b/crates/librefang-runtime-media/Cargo.toml
@@ -18,6 +18,7 @@ tracing = { workspace = true }
 base64 = { workspace = true }
 hex = { workspace = true }
 metrics = { workspace = true }
+tempfile = { workspace = true }
 
 [dev-dependencies]
 # Serialises the env-mutating ElevenLabs voice_id validator tests so
--- a/crates/librefang-runtime-media/src/media_understanding.rs
+++ b/crates/librefang-runtime-media/src/media_understanding.rs
@@ -1609,9 +1609,12 @@ mod tests {
 
         // Written to a file rather than `pipe:1`: a non-fragmented mp4 cannot be *muxed* to a pipe either, since the muxer rewinds to write moov.
         // Generating the fixture the same way the failing input arises is the point of the test.
-        let dir = std::env::temp_dir().join(format!("librefang-6747-{}", std::process::id()));
-        std::fs::create_dir_all(&dir).expect("temp dir");
-        let fixture = dir.join("plain.mp4");
+        // `TempDir` cleans up on drop even if an `assert!` below panics — a hand-rolled path plus `remove_dir_all` after the assertions would leak the directory on that path.
+        let dir = tempfile::Builder::new()
+            .prefix("librefang-6747-")
+            .tempdir()
+            .expect("temp dir");
+        let fixture = dir.path().join("plain.mp4");
 
         let gen = tokio::process::Command::new("ffmpeg")
             .args([
@@ -1647,7 +1650,6 @@ mod tests {
         );
 
         let mp4_bytes = std::fs::read(&fixture).expect("fixture readable");
-        let _ = std::fs::remove_dir_all(&dir);
         assert!(!mp4_bytes.is_empty());
 
         let out = extract_video_audio_track(&mp4_bytes)
--- a/crates/librefang-runtime-media/src/media_understanding.rs
+++ b/crates/librefang-runtime-media/src/media_understanding.rs
@@ -1031,9 +1031,8 @@ fn mime_to_ext(mime: &str) -> Option<String> {
 
 /// A scratch file that deletes itself when it goes out of scope.
 ///
-/// Used only where ffmpeg genuinely cannot work from a pipe — see
-/// `extract_video_audio_track`. Written by hand rather than pulling `tempfile`
-/// into this crate's dependencies for one call site.
+/// Used only where ffmpeg genuinely cannot work from a pipe — see `extract_video_audio_track`.
+/// Written by hand rather than pulling `tempfile` into this crate's dependencies for one call site.
 struct ScopedTempFile {
     path: std::path::PathBuf,
 }
@@ -1041,9 +1040,7 @@ struct ScopedTempFile {
 impl ScopedTempFile {
     /// Create a uniquely-named scratch file and write `bytes` into it.
     ///
-    /// Uniqueness comes from pid plus a process-wide counter, so two
-    /// concurrent extractions in the same process cannot collide and neither
-    /// can two daemons sharing a temp dir.
+    /// Uniqueness comes from pid plus a process-wide counter, so two concurrent extractions in the same process cannot collide and neither can two daemons sharing a temp dir.
     async fn write(bytes: &[u8], extension: &str) -> Result<Self, String> {
         use std::sync::atomic::{AtomicU64, Ordering};
         static COUNTER: AtomicU64 = AtomicU64::new(0);
@@ -1070,13 +1067,10 @@ impl Drop for ScopedTempFile {
 
 /// Run `ffmpeg` with the given arguments and collect stdout.
 ///
-/// `input_bytes` is `Some` when the input is fed on stdin (the args then name
-/// `pipe:0`), and `None` when the args already point at a path on disk. The
-/// distinction matters: a pipe cannot seek, and some containers require a
-/// backward seek to be demuxed at all — see `extract_video_audio_track`.
+/// `input_bytes` is `Some` when the input is fed on stdin (the args then name `pipe:0`), and `None` when the args already point at a path on disk.
+/// The distinction matters: a pipe cannot seek, and some containers require a backward seek to be demuxed at all — see `extract_video_audio_track`.
 ///
-/// Shared by every ffmpeg-based transcode in this module so the spawn / pipe /
-/// timeout / kill-on-timeout plumbing exists once.
+/// Shared by every ffmpeg-based transcode in this module so the spawn / pipe / timeout / kill-on-timeout plumbing exists once.
 ///
 /// `install_hint` names the feature that needs ffmpeg, so the "not on PATH"
 /// error tells the operator what stopped working rather than just that a
@@ -1103,8 +1097,7 @@ async fn run_ffmpeg_pipe(
 
     // Feed stdin concurrently; hanging the write inside the main task
     // would deadlock once ffmpeg's stdout pipe buffer fills.
-    // With a file input there is nothing to write — dropping the handle
-    // closes it so ffmpeg does not wait on a stdin that will never arrive.
+    // With a file input there is nothing to write — dropping the handle closes it so ffmpeg does not wait on a stdin that will never arrive.
     match (child.stdin.take(), input_bytes) {
         (Some(mut stdin), Some(bytes)) => {
             let bytes = bytes.to_vec();
@@ -1201,20 +1194,14 @@ async fn transcode_oga_to_ogg_opus(input_bytes: &[u8]) -> Result<Vec<u8>, String
 /// which is what makes this work across `mp4`/`mov`/`mkv`/`avi` uniformly
 /// rather than needing a format hint per extension.
 ///
-/// The input is staged to a scratch file rather than piped, because a pipe
-/// cannot seek and ISO-BMFF cannot always be demuxed without seeking (#6747).
+/// The input is staged to a scratch file rather than piped, because a pipe cannot seek and ISO-BMFF cannot always be demuxed without seeking (#6747).
 ///
-/// An `mp4` / `mov` stores its index in the `moov` atom. When `moov` sits after
-/// `mdat` — the default output of ffmpeg's own muxer, phone cameras, screen
-/// recorders and meeting exporters — the demuxer has to seek backwards to read
-/// it. Over `pipe:0` that fails, and it fails *quietly*: ffmpeg writes
-/// `Error during demuxing` to stderr but can still exit 0, emitting a
-/// header-only Ogg with zero audio packets. Those bytes then went to the
-/// transcription provider, and the operator saw whatever a provider says about
-/// a soundless file — arbitrarily far from the real cause.
+/// An `mp4` / `mov` stores its index in the `moov` atom.
+/// When `moov` sits after `mdat` — the default output of ffmpeg's own muxer, phone cameras, screen recorders and meeting exporters — the demuxer has to seek backwards to read it.
+/// Over `pipe:0` that fails, and it fails *quietly*: ffmpeg writes `Error during demuxing` to stderr but can still exit 0, emitting a header-only Ogg with zero audio packets.
+/// Those bytes then went to the transcription provider, and the operator saw whatever a provider says about a soundless file — arbitrarily far from the real cause.
 ///
-/// `mkv` and `avi` are streamable and were never affected, which is why the
-/// four container types enabled by #6679 / #6683 split exactly in half.
+/// `mkv` and `avi` are streamable and were never affected, which is why the four container types enabled by #6679 / #6683 split exactly in half.
 async fn extract_video_audio_track(input_bytes: &[u8]) -> Result<Vec<u8>, String> {
     let staged = ScopedTempFile::write(input_bytes, "media").await?;
     let input_path = staged.path.to_string_lossy().into_owned();
@@ -1244,11 +1231,9 @@ async fn extract_video_audio_track(input_bytes: &[u8]) -> Result<Vec<u8>, String
     )
     .await?;
 
-    // Defence in depth for the silent half of #6747. The staged file fixes the
-    // cause, but neither guard in `run_ffmpeg_pipe` can catch a demux that
-    // fails while the process still exits 0 and still writes a container: the
-    // exit code is success and the output is not empty. Exit-code behaviour
-    // also varies across ffmpeg builds, so it cannot be the only line.
+    // Defence in depth for the silent half of #6747.
+    // The staged file fixes the cause, but neither guard in `run_ffmpeg_pipe` can catch a demux that fails while the process still exits 0 and still writes a container: the exit code is success and the output is not empty.
+    // Exit-code behaviour also varies across ffmpeg builds, so it cannot be the only line.
     if !ogg_contains_audio(&out) {
         return Err(format!(
             "ffmpeg produced an Ogg stream with no audio packets ({} bytes) — \
@@ -1262,10 +1247,8 @@ async fn extract_video_audio_track(input_bytes: &[u8]) -> Result<Vec<u8>, String
 
 /// Whether an Ogg stream carries audio, rather than only its headers.
 ///
-/// An Opus stream opens with two header pages, `OpusHead` and `OpusTags`; audio
-/// data begins on the third. Counting `OggS` page captures is enough to tell a
-/// real stream from the 261-byte headers-only artefact a failed demux produces,
-/// and does not require parsing the container.
+/// An Opus stream opens with two header pages, `OpusHead` and `OpusTags`; audio data begins on the third.
+/// Counting `OggS` page captures is enough to tell a real stream from the 261-byte headers-only artefact a failed demux produces, and does not require parsing the container.
 fn ogg_contains_audio(bytes: &[u8]) -> bool {
     const OGG_PAGE_MAGIC: &[u8; 4] = b"OggS";
     bytes
@@ -1603,30 +1586,19 @@ mod tests {
         );
     }
 
-    /// #6747: a plain, non-fragmented mp4 — `moov` written after `mdat`, which
-    /// is what every phone camera, screen recorder and meeting exporter
-    /// produces — must yield a stream with actual audio in it.
+    /// #6747: a plain, non-fragmented mp4 — `moov` written after `mdat`, which is what every phone camera, screen recorder and meeting exporter produces — must yield a stream with actual audio in it.
     ///
-    /// The existing smoke test cannot catch this, for two independent reasons,
-    /// and both had to be fixed here for the test to discriminate.
+    /// The existing smoke test cannot catch this, for two independent reasons, and both had to be fixed here for the test to discriminate.
     ///
-    /// Its fixture passes `-movflags frag_keyframe+empty_moov`, producing a
-    /// *fragmented* mp4 — the one mp4 flavour that demuxes from a non-seekable
-    /// pipe. It exercised precisely the shape that worked.
+    /// Its fixture passes `-movflags frag_keyframe+empty_moov`, producing a *fragmented* mp4 — the one mp4 flavour that demuxes from a non-seekable pipe.
+    /// It exercised precisely the shape that worked.
     ///
-    /// It is also only ~8 KB, and that alone is disqualifying. ffmpeg buffers
-    /// the head of an unseekable input (32 KB by default), so a file small
-    /// enough to fit entirely in that buffer can still be "seeked" backwards
-    /// and demuxes fine over a pipe. Measured against ffmpeg 8.1.1: a 0.5 s /
-    /// 11 KB clip yields 2,500 bytes of Ogg and succeeds, while 5 s / 69 KB,
-    /// 30 s / 389 KB and 120 s / 1.5 MB all yield exactly 261 bytes with
-    /// `stream 0, offset 0x30: partial file` on stderr. That is why the
-    /// fixture below is 10 s rather than the smoke test's 0.5 s — a shorter
-    /// one passes with or without the fix.
+    /// It is also only ~8 KB, and that alone is disqualifying.
+    /// ffmpeg buffers the head of an unseekable input (32 KB by default), so a file small enough to fit entirely in that buffer can still be "seeked" backwards and demuxes fine over a pipe.
+    /// Measured against ffmpeg 8.1.1: a 0.5 s / 11 KB clip yields 2,500 bytes of Ogg and succeeds, while 5 s / 69 KB, 30 s / 389 KB and 120 s / 1.5 MB all yield exactly 261 bytes with `stream 0, offset 0x30: partial file` on stderr.
+    /// That is why the fixture below is 10 s rather than the smoke test's 0.5 s — a shorter one passes with or without the fix.
     ///
-    /// The assertions are the ones the old code failed: a header-only Ogg is
-    /// non-empty, starts with `OggS`, and is smaller than the source mp4, so
-    /// every assertion in the smoke test passed on a stream carrying no audio.
+    /// The assertions are the ones the old code failed: a header-only Ogg is non-empty, starts with `OggS`, and is smaller than the source mp4, so every assertion in the smoke test passed on a stream carrying no audio.
     #[tokio::test]
     async fn extract_video_audio_track_handles_non_fragmented_mp4_6747() {
         if !ffmpeg_available() {
@@ -1634,10 +1606,8 @@ mod tests {
             return;
         }
 
-        // Written to a file rather than `pipe:1`: a non-fragmented mp4 cannot
-        // be *muxed* to a pipe either, since the muxer rewinds to write moov.
-        // Generating the fixture the same way the failing input arises is the
-        // point of the test.
+        // Written to a file rather than `pipe:1`: a non-fragmented mp4 cannot be *muxed* to a pipe either, since the muxer rewinds to write moov.
+        // Generating the fixture the same way the failing input arises is the point of the test.
         let dir = std::env::temp_dir().join(format!("librefang-6747-{}", std::process::id()));
         std::fs::create_dir_all(&dir).expect("temp dir");
         let fixture = dir.join("plain.mp4");
@@ -1656,9 +1626,8 @@ mod tests {
                 "lavfi",
                 "-i",
                 "sine=frequency=440",
-                // 10 s at this size lands around 130 KB, comfortably past the
-                // 32 KB read-ahead buffer that lets a small unseekable input
-                // demux anyway. See the doc comment for the measured cutoff.
+                // 10 s at this size lands around 130 KB, comfortably past the 32 KB read-ahead buffer that lets a small unseekable input demux anyway.
+                // See the doc comment for the measured cutoff.
                 "-t",
                 "10",
                 "-c:v",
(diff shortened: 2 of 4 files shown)


crates/librefang-runtime-media/src/media_understanding.rs
houko, line 1: Process/hygiene issue, not a code issue: commit `a76ec6e` ("docs: reflow media_understanding.rs comments to one-sentence-per-line") on this branch has `GIT_AUTHOR_IDENT` = `Claude <noreply@anthropic.com>`.

CLAUDE.md is explicit that this must never land: the `commit-msg` hook "separately rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean," independent of message content. This commit's message text is clean, but its author identity is exactly the case the hook exists to catch — meaning the hook was either not installed (`core.hooksPath` unset) or bypassed for whoever pushed it.

The reflow content itself looks correct (it brings the new doc comments in line with the project's one-sentence-per-line prose rule), so this isn't a request to revert the change — only to fix its provenance before merge, e.g.:
- `git rebase -i` back to `a618bd1`, squash or re-author `a76ec6e` (`git commit --amend --reset-author` with the correct identity), then force-push `fix/mp4-seekable-extraction`, or
- squash-merge the PR (collapses to one commit under the merger's identity), if that's an acceptable final history for this repo.

I can't do either myself: rewriting `a76ec6e` requires a force-push to update the already-pushed branch, and this review run is under a hard "never force-push" constraint. Flagging for maintainer action instead.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1029,18 +1029,56 @@ fn mime_to_ext(mime: &str) -> Option<String> {
...
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+
+        let path = std::env::temp_dir().join(format!(
+            "librefang-media-{}-{}.{extension}",
+            std::process::id(),
+            COUNTER.fetch_add(1, Ordering::Relaxed)
+        ));
+        tokio::fs::write(&path, bytes)
+            .await
+            .map_err(|e| format!("failed to stage media for extraction at {path:?}: {e}"))?;
```
houko: If `tokio::fs::write` fails partway through — e.g. ENOSPC after some bytes are already flushed — this returns `Err` before `Self { path }` is ever constructed.
`ScopedTempFile`'s `Drop` therefore never runs on this path, and the partial file is left behind in the temp dir.
Worth a best-effort `remove_file` on this error path too, mirroring the best-effort language already used in the `Drop` impl below.

Not blocking — just flagging the gap in the cleanup guarantee for a human to weigh in on.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1029,18 +1029,56 @@ fn mime_to_ext(mime: &str) -> Option<String> {
...
+    /// Uniqueness comes from pid plus a process-wide counter, so two concurrent extractions in the same process cannot collide and neither can two daemons sharing a temp dir.
+    async fn write(bytes: &[u8], extension: &str) -> Result<Self, String> {
+        use std::sync::atomic::{AtomicU64, Ordering};
+        static COUNTER: AtomicU64 = AtomicU64::new(0);
+
+        let path = std::env::temp_dir().join(format!(
+            "librefang-media-{}-{}.{extension}",
+            std::process::id(),
+            COUNTER.fetch_add(1, Ordering::Relaxed)
+        ));
```
houko: Minor security note, not blocking: the filename is fully predictable (pid plus a monotonic counter), and the file lands in the shared OS temp dir with default create permissions.
On a multi-user host another local account could potentially read the staged bytes while they're on disk — this is a voice note or video recording, not throwaway data.
Might be worth restricting permissions (e.g. `0600` via `OpenOptions` on Unix) if that's a threat model this deployment needs to care about.

Leaving this for a human call rather than changing it myself, since it's a deliberate lifetime/security trade-off rather than a mechanical fix.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1029,18 +1029,56 @@ fn mime_to_ext(mime: &str) -> Option<String> {
...
+        tokio::fs::write(&path, bytes)
```
houko: `ScopedTempFile::write` stages the raw input — potentially a user's voice/video recording — under `std::env::temp_dir()` with whatever mode `tokio::fs::write` (i.e. `File::create`) defaults to, typically `0644` under a permissive umask, and no `O_EXCL`.

On a shared/multi-user host that makes the recording briefly world-readable, and the path (`librefang-media-<pid>-<counter>.media`) is guessable enough that a co-resident user racing the write could pre-place a symlink at it before this call runs, so the write would follow that symlink rather than land on a fresh file — the pid+counter uniqueness protects against two extractions in this process colliding with *each other*, not against an adversary guessing ahead of time.

Not flagging this as a required fix since I don't know this project's threading model for the temp dir (single-tenant host assumption may make it moot, and `crates/librefang-kernel-handle/src/tool_policy.rs:85` already uses a bare `std::env::temp_dir()` without extra hardening for uploads), but restricting the mode (e.g. `OpenOptions::new().mode(0o600)` on Unix) and/or opening with `create_new(true)` would close it cheaply if it matters here.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1029,18 +1029,56 @@ fn mime_to_ext(mime: &str) -> Option<String> {
...
+            .await
+            .map_err(|e| format!("failed to stage media for extraction at {path:?}: {e}"))?;
+        Ok(Self { path })
+    }
```
houko, line 1062: Temp-file security: predictable path + no `O_EXCL` + default permissions.

`path` is built from `pid` + a process-wide `AtomicU64` counter that starts at `0` — both are guessable (pid via `/proc`, counter deterministically from process start), and `tokio::fs::write` (via `File::create`) opens with `O_CREAT | O_TRUNC`, not `O_EXCL`. On a shared/multi-user host, another local user can pre-create a symlink at the predicted path (e.g. the very first call after daemon start, `librefang-media-<pid>-0.media`) pointing at a file the daemon has write access to; the write then follows the symlink (CWE-377 / classic insecure-temp-file TOCTOU). Separately, no explicit mode is set, so the file lands at the process umask (commonly `0644`) — the staged bytes are raw user-supplied audio/video, which can be world-readable to other local users for the (short but nonzero) window before `Drop` removes it.

The doc comment's stated tradeoff ("written by hand rather than pulling `tempfile` into this crate's dependencies for one call site") is exactly what the well-known `tempfile` crate (`NamedTempFile` / `tempfile()`) exists to avoid — it creates with `O_EXCL` and restrictive permissions specifically to close this class of bug. Given this is a security-relevant design choice (dependency footprint vs. a real local-attacker vector), leaving for maintainer judgment rather than changing it directly.

---

houko (commented):
Automated daily review. One inline finding on temp-file staging security (see below), plus a process finding that isn't tied to a diff line:

**Commit authorship: `a76ec6e86c37669786364551008f4a43bef4008a` ("docs: reflow media_understanding.rs comments to one-sentence-per-line") is authored by `Claude <noreply@anthropic.com>`.**
Per this repo's own CLAUDE.md ("No Claude / Anthropic / AI attribution in commit messages, PR bodies, or issue comments" — the `commit-msg` hook is documented to reject a commit whose author identity resolves to Claude/Anthropic even when the message text is clean), this commit shouldn't have landed as-is. I'm not rewriting it myself: fixing author identity requires amending/rebasing history on this branch and force-pushing, and this review's operating rules forbid force-push regardless of whose branch it is. Flagging for the maintainer to re-author (`git commit --amend --author=...` + a normal push, or squash) before merge.

Everything else checked out: the `changelog.d/fixed/6751-mp4-mov-audio-extraction.md` fragment is present and correctly formatted (ends `(#6751) (@houko)`), the fix commit uses conventional-commit format, and `extract_video_audio_track_handles_non_fragmented_mp4_6747` is a real regression test against a generated fixture (no dependence on host paths), which is exactly the coverage this class of bug needed. The `ScopedTempFile` `Drop` guard itself does correctly clean up on every exit path (including the `?` early-return through `run_ffmpeg_pipe`'s failure) — my one finding is about how the file is created, not whether it's cleaned up.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1029,18 +1029,56 @@ fn mime_to_ext(mime: &str) -> Option<String> {
...
+/// Written by hand rather than pulling `tempfile` into this crate's dependencies for one call site.
+struct ScopedTempFile {
+    path: std::path::PathBuf,
+}
+
+impl ScopedTempFile {
+    /// Create a uniquely-named scratch file and write `bytes` into it.
+    ///
+    /// Uniqueness comes from pid plus a process-wide counter, so two concurrent extractions in the same process cannot collide and neither can two daemons sharing a temp dir.
+    async fn write(bytes: &[u8], extension: &str) -> Result<Self, String> {
```
houko, line 1045: `ScopedTempFile::write` builds its path from `std::env::temp_dir()` plus the PID and a monotonic counter, then calls `tokio::fs::write`, which creates-or-truncates without `O_EXCL`.
The path is fully predictable from the outside (PID is visible, the counter starts at zero).
On a shared, world-writable `/tmp` (the common case on multi-user Linux hosts) another local user could pre-create a symlink at the predicted name, and this write would follow it rather than fail.
The doc comment explains the deliberate choice to hand-roll this instead of pulling in the `tempfile` crate for one call site, which is a reasonable trade-off — but `tempfile`'s value here isn't just avoiding a dependency, it's `create_new`/`O_EXCL`-style exclusive creation that closes exactly this class of race.
Flagging for a judgment call rather than fixing: is this an accepted risk given the daemon's threat model (single-tenant host), or worth the extra dependency / an explicit `O_EXCL` open here?

---

houko (commented):
Reviewed against CLAUDE.md's engineering rules (worktree/build workflow doesn't apply here; checked changelog fragment format, prose-wrapping, deterministic-ordering, config-Default, `Option<Arc<dyn Trait>>` serde-skip, and commit/attribution conventions), plus the tempfile-lifecycle concern called out for this change.

Changelog fragment (`changelog.d/fixed/6751-mp4-mov-audio-extraction.md`), doc-comment prose (reflowed to one-sentence-per-line in a76ec6e), and the new `ScopedTempFile` `Drop` cleanup all check out — no mechanical fixes needed there.
No AI/Claude/Anthropic attribution anywhere in the diff or commit messages.
Test isolation looks fine: the new regression test's scratch dir (`librefang-6747-<pid>`) and `ScopedTempFile`'s own naming (`librefang-media-<pid>-<counter>.<ext>`) use disjoint patterns, so they can't collide with each other or with other tests in this file.

One inline finding below (predictable temp-file path / no exclusive-create) — a judgment call, not fixed in place.

Also flagging, without a good way to anchor it to a diff line: two commit message bodies here don't follow CLAUDE.md's "one sentence per line" prose rule for commit message bodies.
`a618bd1` ("fix(media): stage video input...") packs multiple sentences per paragraph line instead of one sentence per line.
`a76ec6e` ("docs: reflow media_understanding.rs comments...") — the commit whose stated purpose is fixing exactly this class of wrapping in the source comments — itself hard-wraps its own message body at a fixed column, breaking mid-sentence ("added for the mp4/mov seek fix" / "were hard-wrapped at a fixed column, breaking mid-sentence" split across two lines).
Not fixing via amend + force-push per this review's ground rules; noting for awareness since a future rebase/squash is a low-cost place to tidy it.

---

houko (commented):
No fixes pushed here — the fix and tests (`a618bd1`) are solid: `run_ffmpeg_pipe`'s `Option<&[u8]>` split, `ScopedTempFile`'s Drop-based cleanup, and the `ogg_contains_audio` defence-in-depth guard all check out under a full read of the changed functions, `cargo check -p librefang-runtime-media --lib` and `cargo clippy -p librefang-runtime-media --all-targets -- -D warnings` are both clean, and the `MEDIA_PROVIDER_KEY_VARS`-style completeness argument doesn't apply here but the credential/env-var surface this touches (`std::env::temp_dir()`) has no analogous gap I could find.

One process finding worth flagging before merge: commit `a76ec6e` ("docs: reflow media_understanding.rs comments to one-sentence-per-line") has author identity `Claude <noreply@anthropic.com>` (`git log` / the commits API both confirm — not just the message body, the actual `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL`). That's exactly what CLAUDE.md's commit-msg hook is supposed to reject ("rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean"), so this either bypassed the hook or was pushed from a clone that never ran `just setup`. I'm not rewriting this myself — fixing it needs an amend/rebase + force-push, and force-push is off the table for me here regardless of whose branch it is. Worth a `git commit --amend --author` (or a rebase to drop/reauthor it) before this merges, or a squash-merge that discards individual commit authorship if that's an acceptable substitute for this repo's policy.

---

crates/librefang-runtime-media/src/media_understanding.rs
houko, line 1045: `ScopedTempFile` writes to a predictable path (`temp_dir()/librefang-media-<pid>-<counter>.media`) via `tokio::fs::write`, which opens with plain create+truncate (no `O_EXCL`) and whatever permissions the process umask leaves (typically world/group-readable).
On a shared multi-user host that is two separate weaknesses: a local user who can read `/proc` to learn the daemon's pid can pre-create a symlink at the exact next path and have this call follow it (predictable-tempfile / symlink race, CWE-377), and while the file exists the uploaded video's raw bytes are readable by other local accounts.

The doc comment justifies hand-rolling this "rather than pulling `tempfile` into this crate's dependencies for one call site" — but `tempfile = "3"` is already a **workspace** dependency (root `Cargo.toml:272`) used by 17+ other crates, so adding `tempfile = { workspace = true }` to this crate's `Cargo.toml` does not add a new dependency to audit.
`tempfile::NamedTempFile` / `Builder` creates with `O_EXCL` and a securely-random name, and defaults to `0600` permissions on Unix, closing both gaps for free while keeping the same "deletes itself on drop" behavior this code already relies on.

Not blocking on its own, since the input is already capped at 50 MB by `MediaAttachment::validate()` before it reaches here, but worth a second look given this stages user-uploaded media into a shared temp directory.

---

houko (commented):
Reviewed against CLAUDE.md and for correctness (temp-file cleanup, disk/size bounds, path traversal, seek regression test).

**Blocking:** the branch's HEAD commit (`a76ec6e`) is authored as `Claude <noreply@anthropic.com>`.
That is exactly what the repo's own `commit-msg` hook is written to reject — CLAUDE.md's git-side-hooks section: "separately rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean."
The commit message text itself is fine; only the author identity is the problem.
Please rebase and re-author that commit with your own identity before merge (`git commit --amend --author=...` or an interactive rebase), then push — force-pushing your own branch before review is fine per CLAUDE.md ("Force-push only to your own branches, only before review").

Everything else checked out:
- Temp-file cleanup is RAII (`ScopedTempFile::drop`), so it runs on success, on an early `?` return, and on panic-unwind.
- Disk/size exposure is already bounded upstream by `MAX_VIDEO_BYTES` (50 MB, `librefang-types/src/media.rs:200`) plus the extraction semaphore, so this change does not introduce an unbounded-disk vector.
- No path traversal: the on-disk extension is the hardcoded literal `"media"`, never derived from attachment/user input.
- `extract_video_audio_track_handles_non_fragmented_mp4_6747` genuinely exercises a real backward seek (a 10 s non-fragmented mp4 well past ffmpeg's 32 KB read-ahead buffer) rather than just checking that a file got staged, and it is paired with a unit test that pins `ogg_contains_audio` so the regression test cannot be satisfied by an always-true stub.
- Changelog fragment and doc-comment prose follow the one-sentence-per-line convention and the `changelog.d/fixed/` format.

One inline note on the temp-file approach below — not blocking on its own.

---

crates/librefang-runtime-media/src/media_understanding.rs
houko, line 1: Commit a76ec6e ("docs: reflow media_understanding.rs comments to one-sentence-per-line") on this branch has author identity `Claude <noreply@anthropic.com>`.
The commit message text itself carries no banned phrase, but CLAUDE.md's `commit-msg` hook is explicitly designed to reject this independent of the message: "rejects a commit whose *author identity* (`git var GIT_AUTHOR_IDENT` …) resolves to Claude / Anthropic even when the message itself is clean."
It looks like this went in without `core.hooksPath` enforcing that locally.
Worth rewriting the author on that commit (e.g. `git rebase -i` with `--reset-author`, or a human maintainer squash-merging with their own identity) before this lands — otherwise the merged history carries exactly the attribution the project's hooks exist to keep out.

---

crates/librefang-runtime-media/Cargo.toml
houko, line 1: Minor: the PR description says "Written by hand rather than adding `tempfile` to this crate for one call site," but the branch's actual head (commit de2a4ad6, "stage extracted audio to a securely-created temp file") superseded that — it replaced the hand-rolled pid+counter path with `tempfile::NamedTempFile` and added this `tempfile` dependency, for good reason (O_EXCL + 0600 perms, guaranteed cleanup on a failed write, matching the pattern already used elsewhere in the workspace).
The code itself looks correct; the top-of-PR summary is just stale relative to the final diff and could mislead a reviewer skimming it into thinking the current code hand-rolls the temp path.

---

houko (commented):
Automated daily review. Code itself checks out: `ScopedTempFile` cleans up correctly on every path (including a failed write), the `.oga` streaming path is untouched, the mp4 regression fixture and `ogg_contains_audio` guard are well-targeted, the changelog fragment is correctly formatted, and `cargo fmt --check` / `cargo check -p librefang-runtime-media --lib` are both clean on this branch. Two things flagged inline: an AI-authored commit in the branch history, and a stale line in the PR description.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1029,18 +1029,57 @@ fn mime_to_ext(mime: &str) -> Option<String> {
...
 }
 
-/// Run `ffmpeg` with the given arguments, feeding `input_bytes` on stdin and
-/// collecting stdout — no scratch files on disk. Shared by every ffmpeg-based
-/// transcode in this module so the spawn / pipe / timeout / kill-on-timeout
-/// plumbing exists once.
+/// A scratch file that deletes itself when it goes out of scope.
+///
+/// Used only where ffmpeg genuinely cannot work from a pipe — see `extract_video_audio_track`.
+/// Backed by `tempfile::NamedTempFile` (already a workspace dependency used by `librefang-runtime` and others) rather than a hand-rolled pid+counter path.
```
houko, line 1035: Minor doc/PR-description drift, not a code issue.
The PR description says "Written by hand rather than adding `tempfile` to this crate for one call site," describing the original pid+counter approach from `a618bd1`.
This doc comment (added in `de2a4ad`) correctly describes what actually shipped — `tempfile::NamedTempFile` — but the PR body above was never updated to match, so it now asserts the opposite of what the diff does.
Worth a one-line edit to the PR description before merge so the record (and the eventual changelog/release-notes trail) matches the code.

Reviewed the three specific concerns this class of change usually raises, and all three look correctly handled in the current revision:
- Cleanup on all exit paths: `ScopedTempFile` wraps `tempfile::NamedTempFile`, whose own `Drop` removes the file — success, error (`?` early-return), and panic/unwind all go through the same guard, not a manual cleanup call that an early return could skip.
- Predictable filename / symlink race: `tempfile::Builder::tempfile()` creates with `O_EXCL` and mode `0600` on Unix, so there's no race window and no world-readable window for what the comment correctly notes is "a user's raw audio/video."
- Size bound: `extract_video_audio_track` is only reachable via `MediaUnderstanding::transcribe_audio`, which calls `attachment.validate()` (line ~203) *before* the container bytes are read into memory — that rejects anything over `MAX_VIDEO_BYTES` (50 MiB, `librefang-types/src/media.rs`) up front, and every call site populates `size_bytes` from the actual decoded byte length rather than a client-supplied claim, so the 50 MiB cap can't be bypassed by lying about size. The scratch file this writes is bounded by the same cap.

The regression test's own temp-dir handling (`bc16e77`, using `tempfile::Builder::tempdir()`) has the same panic-safe cleanup property, good pairing with the production code.

---

houko (commented):
Reviewed the temp-file staging change (guard-based cleanup, filename/symlink safety, size bound) plus changelog/prose-wrapping/attribution conventions from CLAUDE.md. No mechanical fixes needed — the specific concerns this class of change usually raises are already handled correctly across the PR's commit history (`de2a4ad`, `bc16e77`). One inline note on a stale line in the PR description.

---

PR #6752, opened 2026-08-08
fix(api): redact hand requirement environment values

houko:
## Type

- [ ] Agent template (TOML)
- [ ] Skill (Python/JS/Prompt)
- [ ] Channel adapter
- [ ] LLM provider
- [ ] Built-in tool
- [x] Bug fix
- [ ] Feature (Rust)
- [ ] Documentation / Translation
- [ ] Refactor / Performance
- [ ] CI / Tooling
- [ ] Other

## Summary

Prevent `GET /api/hands` from returning plaintext values for satisfied environment-variable requirements. The response keeps its existing requirement-key contract and continues to expose only the boolean satisfaction state needed by the Dashboard.

## Changes

- Remove `current_value` from serialized hand requirement status.
- Add a real-router regression that proves the resolved process environment value is absent from the complete response.
- Add a GET → POST secret → GET regression covering Dashboard save compatibility, satisfaction refresh, and canonical `secrets.env` persistence.

## Attribution

No prior work was adapted.

## Testing

- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [ ] `cargo test --workspace` passes

Executed:

- `cargo test -p librefang-api --test hands_routes_integration` (41 passed)
- `cargo clippy -p librefang-api --test hands_routes_integration --no-deps -- -D warnings -A clippy::nonminimal-bool`
- `git diff --check`

The workspace-wide Clippy gate remains blocked by an existing Rust 1.96 `clippy::nonminimal-bool` warning in `crates/librefang-api/src/routes/config/system.rs`, outside this diff.

## Security

- [x] No new unsafe code

Independent final review reported no Critical, Important, or Minor findings.

### DIFF
--- a/changelog.d/security/6752-hand-requirement-secret-redaction.md
+++ b/changelog.d/security/6752-hand-requirement-secret-redaction.md
@@ -1,0 +1,2 @@
+Stop `GET /api/hands` from returning plaintext values for satisfied environment-variable requirements, which exposed host credentials and other sensitive process configuration to any caller allowed to list Hands.
+  Requirement status now reports only whether each variable is present while preserving the existing Dashboard save contract (#6752) (@houko)
--- a/crates/librefang-api/src/routes/skills/hands.rs
+++ b/crates/librefang-api/src/routes/skills/hands.rs
@@ -1072,12 +1072,8 @@ pub async fn set_hand_secret(
         }
     };
 
-    // Resolve the stable requirement key (`r.key`, as exposed by the single-hand
-    // detail endpoint) to the actual environment-variable name (`r.check_value`,
-    // as exposed by `list_hands`'s "key" field). Accepting either alias as input
-    // but persisting under whatever string the client sent would silently write
-    // the secret under a name `check_requirement` never reads, so a save could
-    // return 200 OK while leaving the requirement permanently unsatisfied.
+    // Resolve the stable requirement key (`r.key`, as exposed by the single-hand detail endpoint) to the actual environment-variable name (`r.check_value`, as exposed by `list_hands`'s "key" field).
+    // Accepting either alias as input but persisting under whatever string the client sent would silently write the secret under a name `check_requirement` never reads, so a save could return 200 OK while leaving the requirement permanently unsatisfied.
     let env_key = {
         let defs = state.kernel.hands().list_definitions();
         defs.iter().find(|d| d.id == hand_id).and_then(|def| {
--- a/crates/librefang-api/src/routes/skills/hands.rs
+++ b/crates/librefang-api/src/routes/skills/hands.rs
@@ -67,18 +67,13 @@ pub async fn list_hands(
                 "degraded": degraded,
                 "is_custom": is_custom,
                 "requirements": reqs.iter().map(|(r, ok)| {
-                    let mut req = serde_json::json!({
-                        "key": r.check_value,
+                    serde_json::json!({
+                        "key": r.key,
+                        "check_value": r.check_value,
                         "label": r.label,
                         "satisfied": ok,
                         "optional": r.optional,
-                    });
-                    if *ok {
-                        if let Ok(val) = std::env::var(&r.check_value) {
-                            req["current_value"] = serde_json::json!(val);
-                        }
-                    }
-                    req
+                    })
                 }).collect::<Vec<_>>(),
                 "dashboard_metrics": d.dashboard.metrics.len(),
                 "has_settings": !d.settings.is_empty(),
--- a/crates/librefang-api/src/routes/skills/hands.rs
+++ b/crates/librefang-api/src/routes/skills/hands.rs
@@ -68,8 +68,7 @@ pub async fn list_hands(
                 "is_custom": is_custom,
                 "requirements": reqs.iter().map(|(r, ok)| {
                     serde_json::json!({
-                        "key": r.key,
-                        "check_value": r.check_value,
+                        "key": r.check_value,
                         "label": r.label,
                         "satisfied": ok,
                         "optional": r.optional,
@@ -1058,7 +1057,7 @@ pub async fn set_hand_secret(
     Path(hand_id): Path<String>,
     Json(body): Json<serde_json::Value>,
 ) -> impl IntoResponse {
-    let requested_key = match body["key"].as_str() {
+    let env_key = match body["key"].as_str() {
         Some(k) if !k.trim().is_empty() => k.trim().to_string(),
         _ => {
             return ApiErrorResponse::bad_request("Missing 'key' field (env var name)")
@@ -1073,26 +1072,26 @@ pub async fn set_hand_secret(
         }
     };
 
-    // Resolve the stable requirement key exposed by `list_hands` to the
-    // actual environment-variable name. Keep accepting `check_value`
-    // directly for compatibility with older clients.
-    let env_key = {
+    // Verify this key belongs to a requirement of the specified hand
+    let valid = {
         let defs = state.kernel.hands().list_definitions();
-        defs.iter().find(|d| d.id == hand_id).and_then(|def| {
-            def.requires
-                .iter()
-                .find(|r| r.check_value == requested_key || r.key == requested_key)
-                .map(|r| r.check_value.clone())
-        })
+        defs.iter()
+            .find(|d| d.id == hand_id)
+            .map(|def| {
+                def.requires
+                    .iter()
+                    .any(|r| r.check_value == env_key || r.key == env_key)
+            })
+            .unwrap_or(false)
     };
 
-    let Some(env_key) = env_key else {
+    if !valid {
         return ApiErrorResponse::bad_request(format!(
             "'{}' is not a requirement of hand '{}'",
-            requested_key, hand_id
+            env_key, hand_id
         ))
         .into_json_tuple();
-    };
+    }
 
     // Write to secrets.env
     let secrets_path = state.kernel.home_dir().join("secrets.env");
--- a/crates/librefang-api/src/routes/skills/hands.rs
+++ b/crates/librefang-api/src/routes/skills/hands.rs
@@ -1058,7 +1058,7 @@ pub async fn set_hand_secret(
     Path(hand_id): Path<String>,
     Json(body): Json<serde_json::Value>,
 ) -> impl IntoResponse {
-    let env_key = match body["key"].as_str() {
+    let requested_key = match body["key"].as_str() {
         Some(k) if !k.trim().is_empty() => k.trim().to_string(),
         _ => {
             return ApiErrorResponse::bad_request("Missing 'key' field (env var name)")
@@ -1073,26 +1073,26 @@ pub async fn set_hand_secret(
         }
     };
 
-    // Verify this key belongs to a requirement of the specified hand
-    let valid = {
+    // Resolve the stable requirement key exposed by `list_hands` to the
+    // actual environment-variable name. Keep accepting `check_value`
+    // directly for compatibility with older clients.
+    let env_key = {
         let defs = state.kernel.hands().list_definitions();
-        defs.iter()
-            .find(|d| d.id == hand_id)
-            .map(|def| {
-                def.requires
-                    .iter()
-                    .any(|r| r.check_value == env_key || r.key == env_key)
-            })
-            .unwrap_or(false)
+        defs.iter().find(|d| d.id == hand_id).and_then(|def| {
+            def.requires
+                .iter()
+                .find(|r| r.check_value == requested_key || r.key == requested_key)
+                .map(|r| r.check_value.clone())
+        })
     };
 
-    if !valid {
+    let Some(env_key) = env_key else {
         return ApiErrorResponse::bad_request(format!(
             "'{}' is not a requirement of hand '{}'",
-            env_key, hand_id
+            requested_key, hand_id
         ))
         .into_json_tuple();
-    }
+    };
 
     // Write to secrets.env
     let secrets_path = state.kernel.home_dir().join("secrets.env");
--- a/crates/librefang-api/src/routes/skills/hands.rs
+++ b/crates/librefang-api/src/routes/skills/hands.rs
@@ -1057,7 +1057,7 @@ pub async fn set_hand_secret(
     Path(hand_id): Path<String>,
     Json(body): Json<serde_json::Value>,
 ) -> impl IntoResponse {
-    let env_key = match body["key"].as_str() {
+    let requested_key = match body["key"].as_str() {
         Some(k) if !k.trim().is_empty() => k.trim().to_string(),
         _ => {
             return ApiErrorResponse::bad_request("Missing 'key' field (env var name)")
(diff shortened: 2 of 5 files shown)


houko (commented):
Reviewed the redaction fix itself: `list_hands` in `crates/librefang-api/src/routes/skills/hands.rs` no longer reads `std::env::var(&r.check_value)` into the response, the new regression tests (`list_hands_does_not_expose_satisfied_environment_variable_values`, `set_hand_secret_accepts_the_key_returned_by_list_hands`) cover both the redaction and the save-contract compatibility, and the `changelog.d/security/...` fragment is correctly formatted. I checked the other requirement-serializing sites in this file (`get_hand`, `check_hand_deps`, the post-install response) and none of them ever surfaced `current_value`, so this was the one leak site and it's fully closed. `cargo check -p librefang-api --lib` passes on this branch.

One thing outside this diff, flagged here because GitHub won't let me anchor a line comment to a file the diff doesn't touch: `crates/librefang-api/dashboard/src/pages/HandsPage.tsx:356` still does `if (r.key && r.current_value) init[r.key] = r.current_value;`, and `HandRequirementItem.current_value?: string` is still declared in `crates/librefang-api/dashboard/src/api.ts:1025`. `hand.requirements` there is populated from the same `GET /api/hands` list response this PR fixes, so after this change `r.current_value` is always `undefined` — the code doesn't leak anything (the field simply never arrives now), but it's dead code documenting a contract the backend no longer honors, and it silently drops the "prefill the input with the already-configured value" UX with no replacement (no "already configured" indicator using `r.satisfied` instead). Worth a small follow-up to remove the dead branch/field and decide the replacement UX, partly so a future contributor doesn't "fix" the dashboard by re-adding `current_value` to the API response and reopening this exact leak.

---

crates/librefang-api/src/routes/skills/hands.rs
```diff
@@ -67,18 +67,12 @@ pub async fn list_hands(
...
                         "optional": r.optional,
-                    });
-                    if *ok {
-                        if let Ok(val) = std::env::var(&r.check_value) {
-                            req["current_value"] = serde_json::json!(val);
-                        }
-                    }
-                    req
+                    })
                 }).collect::<Vec<_>>(),
```
houko, line 76: Scope question, not a blocker for this PR: this closes the `current_value` leak for `env_var` **requirements**, and I confirmed the sibling `GET /api/hands/{hand_id}` handler (`get_hand`, ~line 229) never had the equivalent bug — it only ever serialized `check_value` (the env var name), never the resolved value.

However `GET /api/hands/{hand_id}/settings` (`get_hand_settings`, `crates/librefang-api/src/routes/skills/hands.rs` ~line 1176) returns `"current_values": instance_config` verbatim — the full stored settings map for a hand instance, unredacted. If any hand's `settings` schema is used to hold a credential-shaped value (an API token entered via the Settings UI rather than declared as an `env_var` requirement), that endpoint returns it in plaintext to any caller allowed to read hand settings, same class of exposure this PR is fixing for requirements.

Is that intentionally out of scope (e.g. `settings` is documented as never holding secrets, secrets always go through the `requirements`/`secrets_env` path), or is it a second instance of the same class of bug that should get its own fix? Flagging rather than touching it since it's a different endpoint/mechanism than this diff and the answer depends on a security-scope decision, not a mechanical read of the code.

---

houko (commented):
Reviewed the diff against CLAUDE.md contributor rules (changelog fragment format/attribution, no-AI-attribution, conventional commits, secret hygiene, prose wrapping). No mechanical violations found — the core fix, changelog fragment, and new integration tests all comply. One scope question left inline about whether the redaction should also cover hand `settings` values.

---

crates/librefang-api/src/routes/skills/hands.rs
```diff
@@ -67,18 +67,12 @@ pub async fn list_hands(
...
                         "satisfied": ok,
                         "optional": r.optional,
-                    });
-                    if *ok {
-                        if let Ok(val) = std::env::var(&r.check_value) {
-                            req["current_value"] = serde_json::json!(val);
-                        }
-                    }
-                    req
+                    })
```
houko, line 75: Dropping `current_value` here is the right fix for the leak on this endpoint, but the dashboard still has live code built around this field, and this PR doesn't touch it:

- `crates/librefang-api/dashboard/src/api.ts:1025` — `HandRequirementItem.current_value?: string` type is still declared.
- `crates/librefang-api/dashboard/src/pages/HandsPage.tsx:356` — `RequirementsForm` seeds its secret-input state from `r.current_value` on mount: `if (r.key && r.current_value) init[r.key] = r.current_value;`. `RequirementsForm` is fed `hand.requirements` (line 856), and `hand` comes from `useHands()` → `listHands()` → this same `GET /api/hands` endpoint (`crates/librefang-api/dashboard/src/lib/queries/hands.ts`).

So the "already configured" pre-fill in the requirement edit form silently stops working after this change — `r.current_value` is now always `undefined`, the field always renders empty even for a satisfied requirement whose secret was previously saved. That's arguably the *correct* security behavior (no plaintext secret should round-trip into a browser input either), but as written it's an unannounced regression: the type still advertises the field, the changelog only mentions the API/save-contract side and says nothing about the UI losing its pre-fill, and there's no dashboard test guarding either the old or the new behavior.

Given this is a security fix, I'd rather flag than silently patch: please have a maintainer confirm intent — remove the dead `current_value` references from the dashboard (and adjust `RequirementsForm` to show a "configured" indicator instead of relying on the plaintext value), or explicitly document the UX change — rather than leaving stale frontend code pointed at a field the backend can no longer legally send.

---

houko (commented):
Reviewed against CLAUDE.md engineering rules (worktree hygiene, changelog fragment format, no-AI-attribution, prose wrapping, config/serde rules) plus redaction-completeness for the secret leak this PR fixes.

No mechanical rule violations found: the changelog fragment (`changelog.d/security/6752-hand-requirement-secret-redaction.md`) is correctly placed, correctly formatted (sentence-per-line, 2-space continuation indent, `(#6752) (@houko)` attribution), the `hands.rs` fix and its new integration tests (`list_hands_does_not_expose_satisfied_environment_variable_values`, `set_hand_secret_accepts_the_key_returned_by_list_hands`) are sound, and I found no stray AI/Claude/Anthropic attribution anywhere in the diff or commit history.

I traced the redaction fix for completeness and confirmed the sibling `GET /api/hands/{hand_id}` and `check_hand_deps` handlers only ever echo `check_value` (the env-var *name*), never the resolved value — no other backend leak of the actual secret exists.

One completeness concern left as a single inline comment (not code I felt safe silently patching, since it's a UX/product-intent call on the dashboard, not a mechanical fix): removing `current_value` from `GET /api/hands` also silently breaks the dashboard's "pre-fill previously-saved secret in the edit form" behavior (`HandsPage.tsx`'s `RequirementsForm`, typed via `api.ts`'s `HandRequirementItem.current_value`), which isn't mentioned in the changelog or cleaned up in this PR.

---

houko (commented):
Pushed a fix (734d7fc) for a regression this branch's own history reintroduced: `66f8892` ("preserve hand requirement key contract") correctly restored `list_hands`'s `"key": r.check_value` shape, but in the same commit it also reverted `ac2f9c9`'s normalization in `set_hand_secret`, going back to persisting the secret under whatever string the client sent instead of resolving it to `r.check_value`. Since `GET /api/hands/{hand_id}` (single-hand detail) still exposes both `"key": r.key` (stable id) and `"check_value": r.check_value` (actual env var name) as separate fields, and `set_hand_secret` accepts either as valid input, submitting the stable key would validate successfully but persist the secret under a name `check_requirement` never reads — a 200 OK save that silently never satisfies the requirement. Restored the resolve-to-`check_value` behavior and added a regression test (`set_hand_secret_resolves_stable_requirement_key_to_env_var_name`) that exercises the stable-key path directly, which the existing test only covers indirectly since it always submits the `list_hands`-shaped key.

One low-confidence, non-blocking note I'm leaving as a comment rather than fixing (outside this PR's diff, frontend): `crates/librefang-api/dashboard/src/pages/HandsPage.tsx` (~line 356) still reads `r.current_value` to prefill the secret input, which was exactly the field this PR removes from the `list_hands` response. It's harmless today (always `undefined`, so nothing prefills — arguably the right outcome for a secret-redaction fix), but it's now a dead reference to a field the backend will never send again, and `HandRequirementItem.current_value` in `api.ts` (~line 1025) is equally stale. Worth a follow-up cleanup to drop both.

---

houko:
Two findings from review, neither anchorable as an inline diff comment since both sit outside the changed lines.

**1. Commit `734d7fce` (`fix(api): resolve stable requirement key to env var name in set_hand_secret`) carries AI author/committer identity.**
`git log --format="%an <%ae> %cn <%ce>"` shows author and committer both as `Claude <noreply@anthropic.com>`.
CLAUDE.md is explicit that no commit may carry Claude/Anthropic attribution, and the repo's `commit-msg` hook specifically checks `GIT_AUTHOR_IDENT`/`GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` for exactly this case — this commit should have been rejected by it.
The same commit's body is also hard-wrapped at ~72 columns mid-sentence (e.g. "A prior commit in this branch reverted the requirement-key normalization" / "in set_hand_secret while restoring list_hands' \"key\" field contract,"), which conflicts with the repo's sentence-boundary-only prose wrapping rule.
Since rewriting/amending an existing commit is off-limits for an automated reviewer, flagging both so the identity can be corrected (re-author + force-push your own branch, or squash on merge) before this lands on `main` with that authorship intact.

**2. `HandRequirementItem.current_value` (`crates/librefang-api/dashboard/src/api.ts:1025`) is now permanently unset, and `RequirementsForm` in `crates/librefang-api/dashboard/src/pages/HandsPage.tsx:356` still reads it:**
```ts
if (r.key && r.current_value) init[r.key] = r.current_value;
```
Since `GET /api/hands` no longer serializes `current_value` (the point of this PR), this branch can never fire — dead code, and a stale field in the wire type. Functionally harmless now (the input just starts empty instead of prefilled with the secret, which if anything is the more correct behavior), but worth an intentional cleanup pass — dropping the dead field/branch, or replacing it with an explicit "already set" affordance — rather than leaving a type field that lies about what the backend sends.

Everything else in the diff (redaction removal in `list_hands`, the `check_value`/`key` alias resolution in `set_hand_secret`, and the new regression tests in `hands_routes_integration.rs`) looks correct: the new `list_hands_does_not_expose_satisfied_environment_variable_values` and `set_hand_secret_resolves_stable_requirement_key_to_env_var_name` tests do assert the value is absent from the response body and that the secret persists under `check_value` rather than the stable `key`, and the changelog fragment is correctly formatted under `changelog.d/security/`.

---

crates/librefang-api/src/routes/skills/hands.rs
houko, line 1: The `list_hands` fix itself is good and well tested — `list_hands_does_not_expose_satisfied_environment_variable_values` asserts the resolved value never appears anywhere in the serialized response body, not just that `current_value` is absent by key.

I think there's a related, more directly exploitable leak of the same class that this PR doesn't touch: `get_hand_settings` (this file, `GET /api/hands/{hand_id}/settings`) returns `"current_values": instance_config` — the raw stored settings map — completely unredacted. `HandSetting` (crates/librefang-hands/src/lib.rs) has a `Text`-type setting variant with an `env_var: Option<String>`, and its own doc comment gives `ELEVENLABS_API_KEY` as the worked example — i.e. hands can and do store operator-entered secrets directly as a "setting" (not a "requirement"), and this endpoint hands the plaintext value back.

Worse: `/api/hands/` is prefix-matched into the public allowlist in `middleware.rs` (`PublicRoute::prefix_get("/api/hands/")`), so `GET /api/hands/{id}/settings` requires **no authentication at all** — any network-reachable, unauthenticated caller can read back any hand's configured API keys this way. That's a strictly worse exposure than the `list_hands` requirement-value leak this PR fixes (which at least required "allowed to list Hands" per the changelog wording).

Given the PR's whole premise is closing hand-related secret disclosure, I'd expect this in scope rather than a follow-up — but redacting it needs a design call (e.g. an explicit `sensitive`/`secret` flag on `HandSetting`, since `env_var`-presence alone doesn't obviously distinguish "API key" from a non-secret env-exposed setting), so flagging for a maintainer decision rather than pushing a guess myself.

---

changelog.d/security/6752-hand-requirement-secret-redaction.md
houko, line 1: Changelog fragment format is correct (right directory, ends `(#6752) (@houko)`, no leading `- `). Separate process note: commit `734d7fc` in this branch is authored as `Claude <noreply@anthropic.com>`, which CLAUDE.md's attribution ban and the `commit-msg` hook's author-identity check both reject — worth re-authoring before merge.

---

houko (commented):
Automated daily review: the `list_hands` requirement-value redaction is correct and tested against the actual serialized JSON, per CLAUDE.md's bar. Flagging inline a related, unauthenticated secret-disclosure surface in `get_hand_settings` that looks like the same vulnerability class and appears in-scope for this PR's stated purpose, plus an AI-attributed commit in the branch history.

---

crates/librefang-api/src/routes/skills/hands.rs
houko, line 75: The backend-side redaction here looks complete — I checked every other `check_requirements` call site in this file (`get_hand`, `check_hand_deps`, `install_hand_deps`) and the `install_hand`/`get_hand_manifest` handlers, and none of them resolve a live env var value; they only echo the static `check_value`/`key` config strings, which is fine.

One consumer this PR doesn't touch: `crates/librefang-api/dashboard/src/pages/HandsPage.tsx` (`RequirementsForm`, around line 356) still does `if (r.key && r.current_value) init[r.key] = r.current_value;` to pre-fill the secret input with the previously-saved plaintext value, and `dashboard/src/api.ts` still types `current_value?: string` on the requirement shape.

Since `GET /api/hands` will never send `current_value` again, that pre-fill is now permanently dead code (harmless — the input will just always start blank instead of showing the plaintext secret, which if anything is the more secure behavior) but the type and the read are stale.

Is a dashboard follow-up in scope here to drop the dead field/type and decide whether the "satisfied" input should instead show a neutral "configured, click to replace" affordance instead of an empty box? Flagging rather than fixing myself since that's a UX call, not a mechanical redaction fix.

---

houko (commented):
Automated daily review: backend redaction fix looks complete and well-tested (checked every other `check_requirements` call site in `hands.rs` for the same shape of leak — none found). Changelog fragment and prose wrapping are already compliant, so no mechanical fixes were needed this pass. Left one inline comment about a dashboard-side consumer that reads a field this PR just stopped sending.

---

changelog.d/security/6752-hand-requirement-secret-redaction.md
```diff
@@ -0,0 +1,2 @@
+Stop `GET /api/hands` from returning plaintext values for satisfied environment-variable requirements, which exposed host credentials and other sensitive process configuration to any caller allowed to list Hands.
+  Requirement status now reports only whether each variable is present while preserving the existing Dashboard save contract (#6752) (@houko)
```
houko, line 2: This fragment only documents the redaction change in `list_hands`. The diff also fixes a distinct behavior bug in `set_hand_secret`: previously, submitting the stable requirement key (`r.key`, the value the single-hand detail endpoint exposes) instead of the actual env var name (`r.check_value`) would pass validation but persist the secret under the wrong name, so the save would return 200 OK while the requirement stayed permanently unsatisfied. That's a user-visible fix worth its own line (or an added sentence here) since `cargo xtask release` will otherwise ship this PR's release-notes entry describing only the redaction half of the change.

---

houko (commented):
Automated review pass. Verified redaction coverage across list_hands, the single-hand detail endpoint, check-deps, request logging, secret write path, and error scrubbing — found no other surface leaking the raw env var value. One dead-code follow-up fixed directly and pushed; one changelog-completeness note left inline.

---

PR #6758, opened 2026-08-08
fix(approval): fail closed for blocking hand RBAC

houko:
## Summary

- remove the context-free blocking Hand auto-approval carve-out
- always route blocking Hand requests through the real human approval queue because the trait carries neither sender identity nor `force_human`
- retain non-blocking Hand auto-approval where deferred context exists and `force_human` is false
- cover both RBAC-enabled and no-user modes with real-kernel pending-queue tests

## Testing

- RED on prior implementation: blocking request returned `Ok(Approved)`
- `cargo test -p librefang-kernel --test rbac_m3_evaluate_tool_call blocking_hand_approval_ -- --nocapture` (2 passed)
- `cargo test -p librefang-kernel --test rbac_m3_evaluate_tool_call` (11 passed)
- `cargo test -p librefang-kernel-handle --test defaults_approval` (3 passed)
- `cargo clippy -p librefang-kernel --lib --test rbac_m3_evaluate_tool_call --no-deps -- -D warnings`
- `git diff --check`

## Review

### DIFF
--- a/crates/librefang-kernel/src/kernel/handles/approval_gate.rs
+++ b/crates/librefang-kernel/src/kernel/handles/approval_gate.rs
@@ -1,9 +1,6 @@
-//! [`kernel_handle::ApprovalGate`] — tool-approval policy + RBAC gate. Holds
-//! the synchronous "does this tool require approval?" predicates and the
-//! async request/submit/resolve flow used by the agent loop. Hand-tagged
-//! agents (curated trusted packages) auto-approve on the context-carrying
-//! non-blocking path unless per-user policy demanded human approval
-//! (RBAC M3, #3054).
+//! [`kernel_handle::ApprovalGate`] — tool-approval policy + RBAC gate.
+//! Holds the synchronous "does this tool require approval?" predicates and the async request/submit/resolve flow used by the agent loop.
+//! Hand-tagged agents (curated trusted packages) auto-approve on the context-carrying non-blocking path unless per-user policy demanded human approval (RBAC M3, #3054).
 
 use std::sync::Arc;
 
@@ -104,9 +101,8 @@ impl kernel_handle::ApprovalGate for LibreFangKernel {
         use librefang_types::approval::ApprovalRequest as TypedRequest;
 
         // The blocking trait carries neither sender identity nor `force_human`.
-        // It therefore cannot prove that per-user policy allowed a hand-agent
-        // carve-out. Always queue a real approval on this context-free path;
-        // context-carrying hand execution uses `submit_tool_approval` below.
+        // It therefore cannot prove that per-user policy allowed a hand-agent carve-out.
+        // Always queue a real approval on this context-free path; context-carrying hand execution uses `submit_tool_approval` below.
 
         let policy = self.governance.approval_manager.policy();
         let risk_level = crate::approval::ApprovalManager::classify_risk(tool_name);
--- a/crates/librefang-kernel/src/kernel/handles/approval_gate.rs
+++ b/crates/librefang-kernel/src/kernel/handles/approval_gate.rs
@@ -1,8 +1,9 @@
 //! [`kernel_handle::ApprovalGate`] — tool-approval policy + RBAC gate. Holds
 //! the synchronous "does this tool require approval?" predicates and the
 //! async request/submit/resolve flow used by the agent loop. Hand-tagged
-//! agents (curated trusted packages) auto-approve unless the per-user
-//! policy demanded human approval (RBAC M3, #3054).
+//! agents (curated trusted packages) auto-approve on the context-carrying
+//! non-blocking path unless per-user policy demanded human approval
+//! (RBAC M3, #3054).
 
 use std::sync::Arc;
 
@@ -100,18 +101,12 @@ impl kernel_handle::ApprovalGate for LibreFangKernel {
         action_summary: &str,
         session_id: Option<&str>,
     ) -> Result<librefang_types::approval::ApprovalDecision, kernel_handle::KernelOpError> {
-        use librefang_types::approval::{ApprovalDecision, ApprovalRequest as TypedRequest};
+        use librefang_types::approval::ApprovalRequest as TypedRequest;
 
-        // Hand agents are curated trusted packages — auto-approve tool execution.
-        // Check if this agent has a "hand:" tag indicating it was spawned by activate_hand().
-        if let Ok(aid) = agent_id.parse::<AgentId>() {
-            if let Some(entry) = self.agents.registry.get(aid) {
-                if entry.tags.iter().any(|t| t.starts_with("hand:")) {
-                    info!(agent_id, tool_name, "Auto-approved for hand agent");
-                    return Ok(ApprovalDecision::Approved);
-                }
-            }
-        }
+        // The blocking trait carries neither sender identity nor `force_human`.
+        // It therefore cannot prove that per-user policy allowed a hand-agent
+        // carve-out. Always queue a real approval on this context-free path;
+        // context-carrying hand execution uses `submit_tool_approval` below.
 
         let policy = self.governance.approval_manager.policy();
         let risk_level = crate::approval::ApprovalManager::classify_risk(tool_name);
--- a/changelog.d/security/6758-blocking-hand-approval-rbac.md
+++ b/changelog.d/security/6758-blocking-hand-approval-rbac.md
@@ -1,0 +1,1 @@
+Require context-free blocking Hand tool requests to enter the human approval queue, preventing curated Hand auto-approval from bypassing per-user RBAC when sender and `force_human` context are unavailable (#6758) (@houko)
--- a/crates/librefang-kernel/tests/rbac_m3_evaluate_tool_call.rs
+++ b/crates/librefang-kernel/tests/rbac_m3_evaluate_tool_call.rs
@@ -18,6 +18,7 @@
 //!   (d) user policy says NeedsApproval (User role + no allow-list) →
 //!       NeedsApproval
 
+use librefang_kernel::GovernanceSubsystemApi;
 use librefang_kernel::LibreFangKernel;
 use librefang_kernel::SecuritySubsystemApi;
 use librefang_runtime::kernel_handle::prelude::*;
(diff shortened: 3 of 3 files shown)


crates/librefang-kernel/src/kernel/handles/approval_gate.rs
```diff
@@ -100,18 +98,11 @@ impl kernel_handle::ApprovalGate for LibreFangKernel {
...
-            if let Some(entry) = self.agents.registry.get(aid) {
-                if entry.tags.iter().any(|t| t.starts_with("hand:")) {
-                    info!(agent_id, tool_name, "Auto-approved for hand agent");
-                    return Ok(ApprovalDecision::Approved);
-                }
-            }
-        }
+        // The blocking trait carries neither sender identity nor `force_human`.
+        // It therefore cannot prove that per-user policy allowed a hand-agent carve-out.
+        // Always queue a real approval on this context-free path; context-carrying hand execution uses `submit_tool_approval` below.
```
houko, line 105: Fix logic itself looks correct and the new tests genuinely go RED on the prior implementation (confirmed by reading the removed carve-out). One thing worth confirming before merge, though: I searched the whole workspace (`crates/`, `xtask/`) for callers of `kernel_handle::ApprovalGate::request_approval` (the blocking trait method this PR patches) outside of tests, and found none — production tool dispatch (`tool_runner/dispatch.rs:1766`) always goes through `submit_tool_approval` (the non-blocking path), which already had the `force_human` guard before this PR. The dashboard "manual approval" route (`routes/approvals.rs:308`) and the CLI/channel `/api approve` text flow both call `ApprovalManager::request_approval` directly (bypassing the `kernel_handle::ApprovalGate` wrapper entirely), so they were never affected by the bug either.

If that's accurate, the vulnerability this PR closes isn't reachable from any live code path in the current tree — which doesn't make the fix wrong (closing a fail-open trap before something starts calling it is good hygiene, and the trait's own default impl still auto-approves for stub kernels), but it does mean the `security/` changelog severity/framing ("preventing curated Hand auto-approval from bypassing per-user RBAC") may overstate current-day exploitability. Could you confirm whether there's a caller I missed (e.g. in a WASM host-function boundary, MCP tool exec, or something not yet merged), or whether this is intentionally defensive/forward-looking? Worth a one-line note either way for whoever reads the changelog later.

---

houko (commented):
Automated daily review: logic change is correct and well-tested (verified the removed carve-out and confirmed the two new tests go RED on the prior code). CLAUDE.md conventions (changelog fragment format/location, prose one-sentence-per-line, commit hygiene, no AI attribution) all check out. One inline note left below re: whether the fixed code path currently has any production caller — a question for the author, not a blocker.

---

crates/librefang-kernel/src/kernel/handles/approval_gate.rs
houko, line 1: Flagging a commit-hygiene issue that isn't visible from the diff content, only from `git log`: the current HEAD commit on this branch, `e70d205b` ("docs(approval-gate): rewrap PR-touched comments to one sentence per line"), has author identity `Claude <noreply@anthropic.com>`.

CLAUDE.md's git-side `commit-msg` hook is explicitly written to reject exactly this ("...rejects a commit whose *author identity* ... resolves to Claude / Anthropic even when the message itself is clean"), and the top-level "No AI / Claude attribution" rule under Git Conventions covers commit *and* author metadata, not just message text. The commit message itself is clean, so the substance of that rewrap (matching the prose-wrapping convention) is fine — it's only the `Author:` field that's the problem. Worth noting this also means the previous review pass on this PR (the "commit hygiene, no AI attribution ... all check out" summary on `e70d205b`) missed this, since it apparently only checked message text, not author identity.

I can't fix this myself: correcting it requires rewriting the commit (reauthor + force-push to `fix/hand-blocking-approval-rbac`), and force-push is outside what I'm allowed to do autonomously. Suggest squashing/reauthoring that commit (`git commit --amend --author` or an interactive rebase) before merge, and separately checking whatever produced it isn't defaulting `user.name`/`user.email` to the Claude identity for future automated commits.

Separate from the above: I independently re-verified the open question in the still-unresolved thread on line 105 (whether `kernel_handle::ApprovalGate::request_approval` — the blocking path this PR patches — has any production caller). A workspace-wide search for `.request_approval(` and `ApprovalGate for` turns up only test doubles/mocks and the `ApprovalManager::request_approval` (a different method, called from `submit_tool_approval`/`resolve_tool_approval`/`routes/approvals.rs`) — no production call site reaches the trait's blocking `request_approval`. This corroborates the prior reviewer's finding; not asking for anything new here, just confirming it independently since it bears on the changelog's exploitability framing.

Everything else checks out: the fail-closed logic is correct on both tested paths (RBAC-enabled and no-RBAC-context both now queue a real approval instead of auto-approving), the changelog fragment is correctly placed/formatted/attributed, and the new tests genuinely exercise the removed carve-out.

---

houko (commented):
Automated daily review: the security fix itself is sound and fails closed on both tested paths (RBAC-enabled and no-RBAC-context). Changelog fragment, test coverage, and prose-wrap conventions all check out. One commit-hygiene issue found (AI author identity on the HEAD commit, not fixable by me without a force-push) — see inline note. Also independently corroborated the already-open question about whether the patched path has a live production caller; no new action needed there, just confirming.

---

crates/librefang-kernel/src/kernel/handles/approval_gate.rs
```diff
@@ -1,8 +1,6 @@
-//! [`kernel_handle::ApprovalGate`] — tool-approval policy + RBAC gate. Holds
-//! the synchronous "does this tool require approval?" predicates and the
-//! async request/submit/resolve flow used by the agent loop. Hand-tagged
-//! agents (curated trusted packages) auto-approve unless the per-user
-//! policy demanded human approval (RBAC M3, #3054).
+//! [`kernel_handle::ApprovalGate`] — tool-approval policy + RBAC gate.
```
houko, line 1: Policy issue, not a code-correctness one: the tip commit on this branch, `e70d205` ("docs(approval-gate): rewrap PR-touched comments to one sentence per line"), has both author and committer identity `Claude <noreply@anthropic.com>` (the commit's own message text is clean — it's the identity, not the text). CLAUDE.md's commit-msg hook is specifically designed to reject this exact case ("rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean"), so this commit landing on the branch means the hook wasn't active in whatever environment produced it. Content-wise the commit is a pure cosmetic rewrap (verified via `git diff 8a42205..e70d205` — only comment line-breaks changed, no logic touched), so there's no functional concern. Recommend squashing/reauthoring this commit under your own identity before merge so it doesn't ship into `main` history. Flagging rather than fixing myself since correcting authorship requires rewriting an existing commit, which is outside what an automated reviewer should do unilaterally.

---

crates/librefang-kernel/src/kernel/handles/approval_gate.rs
houko, line 105: The fail-closed logic itself looks correct: this now falls straight through to a real `approval_manager.request_approval(req)` call, which blocks on a oneshot channel and resolves to `Denied`/`TimedOut`/`Skipped` on timeout (see `timeout_decision` in `approval.rs`) rather than defaulting to `Approved`, so there's no fail-open path left here that I can find.

One thing worth confirming before merge, though: I grepped for production callers of this specific trait method (`ApprovalGate::request_approval`, the blocking one) across `librefang-runtime`, `librefang-acp`, `librefang-hands`, `librefang-extensions`, `librefang-runtime-mcp`, `librefang-runtime-sandbox-docker`, `librefang-desktop`, and `librefang-cli`, and the only hits are the trait definition/impl and test mocks — I couldn't find a real call site that would exercise this code path today (only a comment reference to it in `librefang-acp/src/permission.rs:121`). If that's accurate, this is still the right fix to land (dead code shouldn't carry a security hole either), but it's worth noting in the PR description whether this path is currently reachable in production, and if so from where — since the new tests assert it now requires a human even in single-user/no-RBAC mode, which is a real behavior change for whatever caller does exercise it (previously Hand-tagged agents on this path auto-approved unconditionally, including in trusted single-user setups).

---

houko (commented):
Automated review pass. The core fix (removing the unconditional Hand-tag auto-approve carve-out on the blocking `request_approval` trait path) looks correct and fail-closed on inspection — no timeout/error branch in `ApprovalManager::request_approval` resolves to `Approved`. Two inline notes: one policy issue (an AI-attributed commit already on the branch, not fixable by me without rewriting history) and one open question for the author about where this blocking path is actually called from in production. Everything else (changelog fragment format/section, conventional commit messages otherwise, no `HashMap` introduced, no config fields touched, scoped test commands in the PR description) checked out clean.

---

crates/librefang-kernel/src/kernel/handles/approval_gate.rs
houko, line 1: The fail-closed logic itself looks correct: the blocking `request_approval` path no longer has a hand-tag carve-out, the non-blocking `submit_tool_approval` path keeps its `force_human` exception (RBAC M3), and the two new tests (`blocking_hand_approval_does_not_bypass_enabled_rbac`, `blocking_hand_approval_without_rbac_context_still_requires_human`) exercise both the RBAC-enabled and no-user-configured cases against a real kernel, which is exactly the "prove the fail-closed default" coverage CLAUDE.md asks for.

Flagging one thing the diff doesn't show: the PR's tip commit, `e70d205b208b3df65112a01f9f7a916c31241ba5` ("docs(approval-gate): rewrap PR-touched comments to one sentence per line"), is authored as `Claude <noreply@anthropic.com>`. Repo CLAUDE.md is explicit that commit *author identity* resolving to Claude/Anthropic must be rejected even when the message text itself is clean — that's exactly this case, since the message has no attribution string but `git log --format='%an %ae'` shows the AI identity. The in-repo `commit-msg` hook is supposed to catch this at commit time; it apparently didn't run for this commit (hooks not installed via `core.hooksPath`, or bypassed). Content-wise the commit is a no-op prose rewrap, but landing it as-is would put an Anthropic-attributed author on `main`'s history. Suggest re-authoring/squashing that commit under the human contributor's identity before merge.

---

houko (commented):
Logic and test coverage for the fail-closed fix look solid. One flag: the PR's tip commit is authored by `Claude <noreply@anthropic.com>` — see inline comment for detail and the CLAUDE.md rule it runs into.

---

crates/librefang-kernel/src/kernel/handles/approval_gate.rs
houko, line 105: I grepped the whole workspace for callers of `KernelHandle::request_approval` (the blocking method this fix hardens) and found none outside tests: the two new tests here, `crates/librefang-kernel-handle/tests/defaults_approval.rs`, and a mock override in `crates/librefang-runtime/src/tool_runner/tests/mod.rs`. Production dispatch (`crates/librefang-runtime/src/tool_runner/dispatch.rs`) appears to go exclusively through `submit_tool_approval` (the deferred/non-blocking path), which already had the correct `force_human` guard before this PR.

If that's accurate, no in-tree caller currently reaches this method for a hand-tagged agent today, so this change hardens a path that isn't wired into any live tool-execution flow yet rather than closing a bypass a real request can currently hit. Could you confirm there's a caller I'm missing (an ACP driver, WASM sandbox tool, or out-of-tree SDK), or otherwise note in the PR description that this closes a latent/defense-in-depth gap rather than a reachable one? Either way it seems worth doing, just want the security framing to be accurate.

---

crates/librefang-kernel/src/kernel/handles/approval_gate.rs
```diff
@@ -100,18 +98,11 @@ impl kernel_handle::ApprovalGate for LibreFangKernel {
         action_summary: &str,
         session_id: Option<&str>,
     ) -> Result<librefang_types::approval::ApprovalDecision, kernel_handle::KernelOpError> {
-        use librefang_types::approval::{ApprovalDecision, ApprovalRequest as TypedRequest};
+        use librefang_types::approval::ApprovalRequest as TypedRequest;
```
houko, line 102: Residual fail-open concern, not in this diff's files but directly implicated by it: `ApprovalGate::request_approval`'s *default trait impl* in `crates/librefang-kernel-handle/src/approval_gate.rs` still returns `Ok(ApprovalDecision::Approved)` unconditionally for any `KernelHandle` implementor that doesn't override it — the trait boundary itself fails open. `LibreFangKernel` (this file) correctly overrides it, and this PR fixes the hand-agent gap in that override, but any future lighter-weight implementor (embedded caller, plugin harness, alternate kernel) that forgets to override `request_approval` silently auto-approves every blocking tool call, including hand agents.

Given the PR's explicit goal is "fail closed for blocking hand RBAC," should the default flip to a denial (or an `Err`) so fail-closed is the contract at the trait level, not just in this one override? Not asking for a fix here since it's a different crate/domain — flagging so a maintainer can decide whether it's in scope.

---

houko (commented):
Automated review pass: the fail-closed fix itself is correct (verified `TimeoutFallback::Deny` is the default so a timeout on the newly-queued approval denies rather than approves, and the per-agent pending cap also denies on overflow). CLAUDE.md conventions (changelog fragment format/section, conventional commit messages, one-sentence-per-line doc comments, no AI attribution) all check out. Two design/scope questions left as inline comments — nothing blocking.

---

PR #6759, opened 2026-08-08
fix(api): reject traversal in Skillhub hand IDs

houko:
## Summary
- validate optional Skillhub `hand` identifiers before any filesystem path join
- preserve the canonical 128-character Hand ID limit while keeping skill names capped at 64
- add network-free HTTP regressions for traversal and the Hand ID compatibility boundary

## Security impact
Prevents `POST /api/skillhub/install` from escaping `workspaces/hands/` through values such as `../../x`, and avoids exposing filesystem existence through the prior 404 response.

## Verification
- `cargo test -p librefang-api --test skills_routes_test` (23 passed)
- `cargo test -p librefang-api skill_identifier_validation --lib` (13 passed)
- `cargo clippy -p librefang-api --lib --test skills_routes_test --no-deps -- -D warnings -A clippy::nonminimal-bool`
- `git diff --check`

### DIFF
--- a/crates/librefang-api/src/routes/skills/mod.rs
+++ b/crates/librefang-api/src/routes/skills/mod.rs
@@ -301,8 +301,7 @@ pub struct PendingListQuery {
 /// `workspaces/hands/`).
 ///
 /// Contract:
-/// - non-empty, ≤ 64 chars for skill names; ≤ 128 for hand ids to
-///   match `librefang_hands`' canonical `MAX_HAND_ID_LEN`
+/// - non-empty, ≤ 64 chars for skill names; ≤ 128 for hand ids to match `librefang_hands`' canonical `MAX_HAND_ID_LEN`
 /// - characters limited to `[A-Za-z0-9_-]` — the strictest project
 ///   convention; cannot contain `..`, `/`, `\`, or any platform
 ///   path separator
--- a/crates/librefang-api/src/routes/skills/mod.rs
+++ b/crates/librefang-api/src/routes/skills/mod.rs
@@ -301,8 +301,8 @@ pub struct PendingListQuery {
 /// `workspaces/hands/`).
 ///
 /// Contract:
-/// - non-empty, ≤ 64 chars (caps log noise and matches the project
-///   pattern from `agent_templates.rs::validate_template_name`)
+/// - non-empty, ≤ 64 chars for skill names; ≤ 128 for hand ids to
+///   match `librefang_hands`' canonical `MAX_HAND_ID_LEN`
 /// - characters limited to `[A-Za-z0-9_-]` — the strictest project
 ///   convention; cannot contain `..`, `/`, `\`, or any platform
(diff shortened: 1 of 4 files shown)


crates/librefang-api/src/routes/skills/mod.rs
```diff
@@ -313,9 +312,10 @@ pub struct PendingListQuery {
 /// `field` is "name" or "hand" — used to scope the rejection
 /// message so the client knows which input was bad.
 fn validate_skill_identifier(value: &str, field: &str) -> Result<(), String> {
-    if value.is_empty() || value.len() > 64 {
+    let max_len = if field == "hand" { 128 } else { 64 };
+    if value.is_empty() || value.len() > max_len {
```
houko, line 316: Nit, not blocking: `max_len` is picked by string-equality on `field` (`"hand"` vs everything else defaulting to 64). If a future call site passes a typo'd literal (`"Hand"`, `"hand_id"`, …) it silently falls back to the 64-char skill-name limit instead of erroring, which is a quiet way to reintroduce exactly the kind of length-mismatch this PR is fixing. Verified all four current call sites (`skill.rs`, `clawhub.rs` x2, `skillhub.rs`) pass the literal correctly, so no live bug — just flagging the shape as fragile for the next caller. An enum (`SkillIdentifierField::{Name,Hand}`) would make the mismatch a compile error instead of a silent semantic change.

---

crates/librefang-api/src/routes/skills/mod.rs
```diff
@@ -301,8 +301,7 @@ pub struct PendingListQuery {
 /// `workspaces/hands/`).
 ///
 /// Contract:
-/// - non-empty, ≤ 64 chars (caps log noise and matches the project
-///   pattern from `agent_templates.rs::validate_template_name`)
+/// - non-empty, ≤ 64 chars for skill names; ≤ 128 for hand ids to match `librefang_hands`' canonical `MAX_HAND_ID_LEN`
 /// - characters limited to `[A-Za-z0-9_-]` — the strictest project
```
houko, line 305: Worth a follow-up thought (not asking for it in this PR): the 128 here is a hand-copied literal of `librefang_hands::MAX_HAND_ID_LEN`, which is itself private to that crate, and there's a *third* independent copy of the same alphanumeric/`-`/`_`/first-char-alnum validation logic in `hands_hub::validate_hand_id`. This exact drift (64 vs. 128) between two unsynced copies is what caused the bug this PR fixes. Exporting the constant (and ideally the validator) from `librefang_hands` for reuse here would close the loophole structurally instead of relying on the doc comment to keep both in sync by convention. Cross-crate change though, so leaving as a suggestion rather than asking for it in a security-fix-sized PR.

---

houko (commented):
Automated daily review. Security fix logic is sound — see inline notes for two non-blocking nits (both left as suggestions, not asked for in this PR).

---

PR #6760, opened 2026-08-08
fix(types): block traversal segments in capability path globs

houko:
## Summary
- reject literal `.` and `..` segments before every wildcard matcher dispatch
- close traversal through path-scoped `*`/`**` and separator-free prefix/suffix patterns
- cover Unix, Windows, FileRead, and FileWrite cases while preserving explicit bare `*`, exact grants, and dotfiles

## Security impact
Prevents scoped wildcard capabilities from matching values that escape through parent/current-directory segments. Bare `*` remains the explicit universal grant.

## Verification
- `cargo test -p librefang-types` (903 unit tests; all integration and doc tests passed)
- `cargo clippy -p librefang-types --all-targets --no-deps -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/librefang-types/src/capability.rs
+++ b/crates/librefang-types/src/capability.rs
@@ -224,10 +224,8 @@ pub fn glob_matches(pattern: &str, value: &str) -> bool {
     if pattern == value {
         return true;
     }
-    // Wildcard grants must never manufacture access through filesystem
-    // traversal components, even when the pattern itself has no separator
-    // and would otherwise take the legacy matcher path. Bare `*` above is
-    // the documented universal grant; an exact grant above remains explicit.
+    // Wildcard grants must never manufacture access through filesystem traversal components, even when the pattern itself has no separator and would otherwise take the legacy matcher path.
+    // Bare `*` above is the documented universal grant; an exact grant above remains explicit.
     if has_path_traversal_segment(value) {
         return false;
     }
--- a/crates/librefang-types/src/capability.rs
+++ b/crates/librefang-types/src/capability.rs
@@ -195,7 +195,8 @@ pub fn validate_capability_inheritance(
 /// `data/*` matches `data/file.txt` but NOT `data/../../etc/passwd`.
 ///
 /// **Double-segment wildcard `**`** — matches any characters including `/`,
-/// so `data/**` matches `data/a/b/c/file.txt`.
+/// so `data/**` matches `data/a/b/c/file.txt`. Path values containing literal
+/// `.` or `..` segments never match a path glob.
 ///
 /// **Bare `*`** (the entire pattern is just `"*"`) — matches anything, for
 /// backward compatibility with the universal wildcard grant.
@@ -306,6 +307,12 @@ fn glob_matches_path(pattern: &str, value: &str) -> bool {
     const SEPS: &[char] = &['/', '\\'];
     let pat_segs: Vec<&str> = pattern.split(SEPS).collect();
     let val_segs: Vec<&str> = value.split(SEPS).collect();
+    if val_segs
+        .iter()
+        .any(|segment| matches!(*segment, "." | ".."))
+    {
+        return false;
+    }
     glob_match_segments(&pat_segs, &val_segs)
 }
 
@@ -620,6 +627,27 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_path_globs_reject_dot_and_parent_segments() {
+        for (pattern, value) in [
+            ("data/*", "data/.."),
+            ("data/*", "data/."),
+            ("data/**", "data/../../etc/passwd"),
+            ("data/**", "data/./file.txt"),
+            ("C:\\data\\**", "C:\\data\\..\\Windows\\secret"),
+        ] {
+            assert!(
+                !glob_matches(pattern, value),
+                "path glob {pattern:?} must reject traversal value {value:?}"
+            );
+        }
+
+        assert!(
+            glob_matches("data/**", "data/.env"),
+            "ordinary dot-prefixed names are not traversal segments"
+        );
+    }
+
     /// URL capability patterns: `*` in the host portion must NOT match an
     /// entirely different domain. The host string is already extracted as
     /// `hostname:port` before this function is called, so the separator to
@@ -676,6 +704,14 @@ mod tests {
         assert!(!capability_matches(
             &Capability::FileRead("/data/*".to_string()),
             &Capability::FileRead("/data/../../etc/passwd".to_string()),
+        ));
+        assert!(!capability_matches(
+            &Capability::FileRead("/data/**".to_string()),
+            &Capability::FileRead("/data/../../etc/passwd".to_string()),
+        ));
+        assert!(!capability_matches(
+            &Capability::FileWrite("/data/*".to_string()),
+            &Capability::FileWrite("/data/..".to_string()),
         ));
     }
 
--- a/crates/librefang-types/src/capability.rs
+++ b/crates/librefang-types/src/capability.rs
@@ -222,6 +222,13 @@ pub fn glob_matches(pattern: &str, value: &str) -> bool {
     if pattern == value {
         return true;
     }
+    // Wildcard grants must never manufacture access through filesystem
+    // traversal components, even when the pattern itself has no separator
+    // and would otherwise take the legacy matcher path. Bare `*` above is
+    // the documented universal grant; an exact grant above remains explicit.
+    if has_path_traversal_segment(value) {
+        return false;
+    }
 
     // If the pattern contains a path separator we apply segment-aware
     // matching so that a single `*` cannot cross a separator.  Both `/`
@@ -307,13 +314,14 @@ fn glob_matches_path(pattern: &str, value: &str) -> bool {
     const SEPS: &[char] = &['/', '\\'];
     let pat_segs: Vec<&str> = pattern.split(SEPS).collect();
     let val_segs: Vec<&str> = value.split(SEPS).collect();
-    if val_segs
-        .iter()
-        .any(|segment| matches!(*segment, "." | ".."))
-    {
-        return false;
-    }
     glob_match_segments(&pat_segs, &val_segs)
+}
+
+fn has_path_traversal_segment(value: &str) -> bool {
+    const SEPS: &[char] = &['/', '\\'];
+    value
+        .split(SEPS)
+        .any(|segment| matches!(segment, "." | ".."))
 }
 
 /// Recursive segment-by-segment matcher.
@@ -646,6 +654,22 @@ mod tests {
             glob_matches("data/**", "data/.env"),
             "ordinary dot-prefixed names are not traversal segments"
         );
+    }
+
+    #[test]
+    fn test_separator_free_globs_reject_path_traversal_values() {
+        for (pattern, value) in [
+            ("**", ".."),
+            ("**", "data/../../etc/passwd"),
+            ("data*", "data/../secret"),
+            ("*passwd", "data/../etc/passwd"),
+            ("C:*", "C:\\data\\..\\Windows\\secret"),
+        ] {
+            assert!(
+                !glob_matches(pattern, value),
+                "separator-free glob {pattern:?} must reject traversal value {value:?}"
+            );
+        }
     }
 
     /// URL capability patterns: `*` in the host portion must NOT match an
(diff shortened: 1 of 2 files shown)


crates/librefang-types/src/capability.rs
```diff
@@ -309,6 +317,13 @@ fn glob_matches_path(pattern: &str, value: &str) -> bool {
     glob_match_segments(&pat_segs, &val_segs)
 }
 
+fn has_path_traversal_segment(value: &str) -> bool {
+    const SEPS: &[char] = &['/', '\\'];
+    value
+        .split(SEPS)
+        .any(|segment| matches!(segment, "." | ".."))
+}
```
houko, line 325: `has_path_traversal_segment` only flags a segment that is *exactly* `.` or `..` after splitting on `/`/`\`.
On Windows, a drive-relative segment like `C:..` (no separator between the drive letter and the `..`) is a distinct segment string and slips past this check untouched, even though Windows resolves it as "parent of the current directory on drive C" — a real traversal primitive.

In practice this doesn't currently escalate through `glob_matches_path` (line ~370, `segment_matches`) because a granted first segment of `"C:"` won't literally equal a value segment of `"C:.."`, so path-style patterns like `C:\data\**` stay safe.
But a separator-free grant such as `FileRead("C:**")` — plausible as an (arguably malformed but not unreasonable) "whole C: drive" grant, and not caught by the `pattern.contains('/') || pattern.contains('\\')` branch since it has no separator at all — falls through to the legacy `glob_matches_simple`, where the existing `strip_prefix`/`find('*')` logic treats `"C:**"` as "starts_with(\"C:\")", and `"C:..\Windows\secret"` starts with `"C:"`.
`has_path_traversal_segment` won't reject that value first because its lone offending segment is `"C:.."`, not `".."`.

This is a narrow, Windows-specific, separator-free-pattern edge case — I'm not confident it's worth a fix in this PR (would need either routing any pattern containing a drive-letter-like prefix through the path matcher, or extending the segment check to also flag a segment matching `^[A-Za-z]:\.\.?$`), but flagging since this PR's own stated goal is closing traversal bypasses and the new tests only cover the `C:\data\..\Windows\secret` (separator present) shape, not the drive-relative-without-separator shape.

---

houko (commented):
Reviewed the traversal-guard change against CLAUDE.md's rules (changelog fragment format, prose wrapping, config/determinism conventions) and sanity-checked the path-traversal logic itself.

Fixed one mechanical nit directly (pushed): the new inline comment above `has_path_traversal_segment`'s call site used the old hard-wrap convention for brand-new prose, which CLAUDE.md's prose-wrapping section explicitly disallows for new text — rewrapped to one sentence per line.

Left one inline comment on a narrow Windows drive-relative-path edge case (`C:..` without a separator) that I'm not confident warrants a fix in this PR — see the thread on `capability.rs`.

Everything else checked out: `changelog.d/security/6760-capability-glob-traversal.md` fragment is correctly formatted (right directory, no leading `-`, ends `(#6760) (@houko)`), no HashMap/BTreeMap or config-field concerns apply here, no route/wiring touched so no integration test gap, and `cargo test -p librefang-types --lib capability::` (31 tests) plus `cargo clippy -p librefang-types --lib -- -D warnings` and `cargo fmt -p librefang-types -- --check` are all clean after my fix.

---

crates/librefang-types/src/capability.rs
```diff
@@ -221,6 +224,11 @@ pub fn glob_matches(pattern: &str, value: &str) -> bool {
     if pattern == value {
         return true;
     }
+    // Wildcard grants must never manufacture access through filesystem traversal components, even when the pattern itself has no separator and would otherwise take the legacy matcher path.
+    // Bare `*` above is the documented universal grant; an exact grant above remains explicit.
+    if has_path_traversal_segment(value) {
```
houko, line 229: `has_path_traversal_segment` runs for every `Capability` variant that funnels through `glob_matches` — not just `FileRead`/`FileWrite`. `capability_matches` dispatches the same pattern matcher for `ShellExec`, `EnvRead`, `MemoryRead`/`MemoryWrite`, `AgentMessage`/`AgentKill`, `LlmQuery`, `NetConnect`, `ToolInvoke`, `OfpConnect`, and `EconTransfer`.

For those non-path domains, a value that legitimately contains a bare `.` or `..` segment (e.g. a `ShellExec` command like `./run.sh`, or a hierarchical `MemoryWrite` scope segment that happens to be `..`) will now be silently denied, even under a narrow grant that was written to match it exactly (only the bare `*` universal grant and an exact `pattern == value` grant are exempt — both checked *before* this guard).

Is the intent for this guard to be file/path-scoped only, or deliberately universal as defense-in-depth across every capability domain? If universal is intended, it'd be worth a doc note on `glob_matches` making that explicit (right now the surrounding doc comment and every added test talk only about path/Windows-path cases), plus at least one regression test for a non-path domain (e.g. `ShellExec` or `MemoryWrite`) to lock in the chosen behavior.

---

crates/librefang-types/src/capability.rs
```diff
@@ -309,6 +317,13 @@ fn glob_matches_path(pattern: &str, value: &str) -> bool {
     glob_match_segments(&pat_segs, &val_segs)
 }
 
+fn has_path_traversal_segment(value: &str) -> bool {
```
houko, line 320: This rejects a bare `.` segment identically to `..`, but `.` alone (current-directory reference) is not a traversal — `data/./file.txt` and `data/file.txt` name the same resource. The added test asserts `data/**` must reject `data/./file.txt` and `data/*` must reject `data/.`, so this looks intentional, but it means a narrow grant like `FileRead("data/**")` now denies a perfectly benign path just because some component of the request normalized with a leading `./` or an internal `/./`, and there's no bare-`*` escape hatch for that case since the pattern isn't `*`.

Is folding `.` into the same bucket as `..` a deliberate simplification (reject any special dot-segment, don't try to distinguish benign self-reference from parent-escape), or should `.` segments be stripped/normalized rather than treated as a hard reject? Worth confirming since it changes what previously-working narrow-glob grants continue to work after this lands.

---

crates/librefang-types/src/capability.rs
houko, line 1: Not about this file's content, but flagging since it's about a commit that touches only this file: commit ef98a27 ("docs(types): rewrap new traversal-guard comment to one sentence per line") rewraps the doc comment to comply with CLAUDE.md's "one sentence per line" prose rule — but the commit's own message body is itself hard-wrapped mid-sentence:

```
The comment added alongside the traversal-segment guard used the old
hard-wrap convention.
CLAUDE.md's prose-wrapping rule requires new prose to break only at
sentence boundaries.
```

"...used the old / hard-wrap convention." and "...break only at / sentence boundaries." are each one sentence split across two lines — the exact violation the commit is fixing in code. Since CLAUDE.md's prose rule explicitly covers "commit message bodies" too, and existing commits can't be amended/rewritten per this run's constraints, flagging for the author to fix in a follow-up commit (or just note for future commit messages) rather than doing it myself.

---

crates/librefang-types/src/capability.rs
houko, line 320: Nice centralized fix — since every `capability_matches` arm and every other caller (`user_policy.rs`, `tool_policy.rs`, `approval.rs`, `approval_gate.rs`, `dispatch.rs`, `tools_and_skills.rs`, `loader.rs`, `config/types.rs`, `agents/config.rs`) all funnel through this one `glob_matches`, the guard automatically covers every consumer — confirmed via `search_code` for other `glob_matches`/duplicate glob implementations, found none besides the intentional TS mirror for tool names (which never sees `/`/`.`, so it's exempt).

One completeness question on the guard itself, not a request to block this PR: `has_path_traversal_segment` does an exact string match on `"."`/`".."` after splitting on the literal separator characters. That's solid for the raw-path cases in the tests (start/mid/end, mixed `/`/`\`), but this same function is also used for URL-shaped patterns (see `test_glob_url_star_does_not_cross_scheme_host_boundary` a few lines below, and `NetConnect`/`OfpConnect`). If any caller ever passes a not-yet-decoded value — e.g. a percent-encoded path segment like `%2e%2e` in a URL, or a doubly-encoded `%252e%252e` — this check won't catch it, since it only recognizes the literal two-character `..`/`.` after splitting on `/`/`\`. Do we know all call sites always pass already-decoded/canonicalized values into `glob_matches`? If not, worth either documenting that precondition loudly on `glob_matches`/`has_path_traversal_segment`, or adding a decode-then-check step (or a test asserting the current behavior is intentional/safe because decoding always happens upstream).

---

houko (commented):
Automated daily review: the traversal guard looks solid and correctly centralized (verified no duplicate glob-matching implementation elsewhere in the repo bypasses it), changelog fragment and test coverage (start/mid/end positions, Windows separators, legitimate dot-prefixed names, separator-free patterns) all meet the repo's documented conventions. Left one inline completeness question about pre-decode/encoded traversal values; no mechanical fixes needed.

---

crates/librefang-types/src/capability.rs
houko, line 229: This guard lives inside the shared `glob_matches`, so it now changes matching semantics for every `Capability` variant that routes through it, not just `FileRead`/`FileWrite`: `NetConnect`, `MemoryRead`/`MemoryWrite`, `ShellExec`, `AgentMessage`/`AgentKill`, `OfpConnect`, `EconTransfer`, `LlmQuery`.
All the new tests exercise file-path patterns only.
If any of those non-path capability kinds legitimately use `/`-hierarchical values where a segment can be exactly `.` or `..` (e.g. a memory scope, a shell-command argument, an OFP peer path), this silently starts rejecting matches that previously worked.
Worth either scoping the check to the `FileRead`/`FileWrite` dispatch path explicitly, or adding a regression test per non-path variant confirming no behavior change is intended/acceptable there.

---

crates/librefang-types/src/capability.rs
houko, line 320: `has_path_traversal_segment` only catches a literal `.` or `..` string segment after splitting on `/`/`\`.
It does not, and by design at this layer cannot, catch percent-encoded traversal (`%2e%2e`), Windows trailing-dot/space component quirks, or symlink-based escapes — those need canonicalization/OS-level resolution, not string matching.
That's fine for the WASM `host_fs_*` call sites in `host_functions.rs`, which check the capability against the already-canonicalized path (confirmed at lines ~307, ~343, ~427 there).
But `validate_capability_inheritance` (this file) and `CapabilityManager::check` (`librefang-kernel/src/capabilities.rs`) compare raw, non-canonicalized strings — e.g. a subagent-spawn request's requested `Capability::FileRead(..)` string is checked here directly, before any filesystem resolution happens.
For those callers this function *is* the only traversal defense, so it's worth being explicit (in the doc comment) that this is a string-level literal-segment check, not a canonicalizing one, so a future caller doesn't assume it covers encoded/symlink variants too.
Not asking for a fix to those encodings here since they aren't real OS-level traversal primitives, just flagging so it's a documented assumption rather than an implicit one.

---

houko (commented):
Reviewed the traversal fix for correctness against `..`, `%2e%2e`, Windows backslash, absolute-path, and symlink escape vectors, and against CLAUDE.md's rules.

The core fix is correct and needed: prior to this PR, `data/*` matched `data/..` (since `*` matches any single segment, including `..`), and `data/**` matched any depth of `../..` since `**` unconditionally absorbs segments — a real bypass for any `**` grant. The new `has_path_traversal_segment` check closes both, is applied consistently for both `/` and `\` separators, and is well covered by the new tests (Unix, Windows, mixed-separator, dotfile-preservation cases).

CLAUDE.md compliance checked: conventional-commit title, changelog fragment format/section/attribution, prose-wrapping (one sentence per line) in the new doc comments, no AI attribution — all fine. No workspace-wide `HashMap`/prompt-ordering concerns apply here.

Left two inline comments on residual concerns I'm not fully confident are non-issues (not blocking, no REQUEST_CHANGES):
1. The guard sits in the shared `glob_matches`, so it changes semantics for every capability kind that routes through it (`MemoryRead`/`Write`, `ShellExec`, `NetConnect`, etc.), not just `FileRead`/`FileWrite` — untested for those variants.
2. It's a literal-segment string check with no canonicalization; that's safe for the WASM `host_fs_*` sites (which check post-canonicalize) but is the *only* defense for `validate_capability_inheritance` and `CapabilityManager::check`, which compare raw strings — worth documenting that assumption explicitly.

No mechanical fixes applied — nothing I found rose to the confidence level of "known-correct commit," and this is a security-sensitive matcher where I'd rather flag than guess.

---

PR #6761, opened 2026-08-08
fix(runtime): harden legacy web-fetch SSRF transport

houko:
## Summary
- resolve and validate legacy web-fetch targets asynchronously
- use a direct, no-proxy client pinned to every SSRF-validated socket address
- disable automatic redirects so no unvalidated target can be followed
- preserve multi-address fallback with `resolve_to_addrs`

## Security impact
Closes DNS-rebinding through both reqwest re-resolution and proxy-side resolution in the fallback web-fetch path. Redirect targets cannot bypass the single-target validation.

## Verification
- `cargo test -p librefang-http` (5 passed)
- `cargo test -p librefang-runtime tool_runner::web_legacy::tests --lib` (4 passed)
- `cargo test -p librefang-runtime web_fetch::tests --lib` (43 passed)
- `cargo clippy -p librefang-http -p librefang-runtime --lib --no-deps -- -D warnings`
- `git diff --check`

### DIFF
--- a/changelog.d/security/6761-legacy-web-fetch-ssrf.md
+++ b/changelog.d/security/6761-legacy-web-fetch-ssrf.md
@@ -1,0 +1,1 @@
+Pin legacy web-fetch connections to SSRF-validated DNS results and reject automatic redirects (#6761) (@houko)
--- a/crates/librefang-http/src/lib.rs
+++ b/crates/librefang-http/src/lib.rs
@@ -165,6 +165,15 @@ pub fn proxied_client_builder() -> reqwest::ClientBuilder {
     build_http_client(&active_proxy())
 }
 
+/// Build an HTTP client that never uses configured or environment proxies.
+///
+/// SSRF-sensitive callers that pin a hostname to locally validated socket
+/// addresses must connect directly. A proxy can resolve the original hostname
+/// itself, bypassing reqwest's DNS overrides and reopening DNS rebinding.
+pub fn direct_client_builder() -> reqwest::ClientBuilder {
+    build_http_client(&ProxyConfig::default()).no_proxy()
+}
+
 /// Convenience: build a ready-to-use proxy-aware [`reqwest::Client`].
 pub fn proxied_client() -> reqwest::Client {
     proxied_client_builder()
@@ -303,6 +312,12 @@ mod tests {
     fn test_empty_proxy_config_builds_client() {
         let proxy = ProxyConfig::default();
         let client = build_http_client(&proxy).build().unwrap();
+        drop(client);
+    }
+
+    #[test]
+    fn test_direct_client_builder_builds_client() {
+        let client = direct_client_builder().build().unwrap();
         drop(client);
     }
 
--- a/crates/librefang-http/src/lib.rs
+++ b/crates/librefang-http/src/lib.rs
@@ -167,9 +167,8 @@ pub fn proxied_client_builder() -> reqwest::ClientBuilder {
 
 /// Build an HTTP client that never uses configured or environment proxies.
 ///
-/// SSRF-sensitive callers that pin a hostname to locally validated socket
-/// addresses must connect directly. A proxy can resolve the original hostname
-/// itself, bypassing reqwest's DNS overrides and reopening DNS rebinding.
+/// SSRF-sensitive callers that pin a hostname to locally validated socket addresses must connect directly.
+/// A proxy can resolve the original hostname itself, bypassing reqwest's DNS overrides and reopening DNS rebinding.
 pub fn direct_client_builder() -> reqwest::ClientBuilder {
     build_http_client(&ProxyConfig::default()).no_proxy()
 }
--- a/crates/librefang-runtime/src/host_functions.rs
+++ b/crates/librefang-runtime/src/host_functions.rs
@@ -546,13 +546,8 @@ fn host_net_fetch(state: &GuestState, params: &serde_json::Value) -> serde_json:
                     return e;
                 }
 
-                // DNS-pinned, non-redirecting, direct (no-proxy) client: connect
-                // to the exact IPs we just validated for this hop. A configured
-                // proxy would re-resolve `hostname` itself, bypassing the pin
-                // below and reopening DNS rebinding (#6761); `resolve_to_addrs`
-                // (rather than repeated `resolve()` calls, which overwrite the
-                // mapping instead of accumulating it) keeps every validated
-                // address available as a fallback.
+                // DNS-pinned, non-redirecting, direct (no-proxy) client: connect to the exact IPs we just validated for this hop.
+                // A configured proxy would re-resolve `hostname` itself, bypassing the pin below and reopening DNS rebinding (#6761); `resolve_to_addrs` (rather than repeated `resolve()` calls, which overwrite the mapping instead of accumulating it) keeps every validated address available as a fallback.
                 let builder = librefang_http::direct_client_builder()
                     .redirect(reqwest::redirect::Policy::none())
                     .resolve_to_addrs(&ssrf_result.hostname, &ssrf_result.resolved);
--- a/crates/librefang-runtime/src/host_functions.rs
+++ b/crates/librefang-runtime/src/host_functions.rs
@@ -546,13 +546,16 @@ fn host_net_fetch(state: &GuestState, params: &serde_json::Value) -> serde_json:
                     return e;
                 }
 
-                // DNS-pinned, non-redirecting client: connect to the exact IPs
-                // we just validated for this hop.
-                let mut builder = librefang_http::proxied_client_builder()
-                    .redirect(reqwest::redirect::Policy::none());
-                for addr in &ssrf_result.resolved {
-                    builder = builder.resolve(&ssrf_result.hostname, *addr);
-                }
+                // DNS-pinned, non-redirecting, direct (no-proxy) client: connect
+                // to the exact IPs we just validated for this hop. A configured
+                // proxy would re-resolve `hostname` itself, bypassing the pin
+                // below and reopening DNS rebinding (#6761); `resolve_to_addrs`
+                // (rather than repeated `resolve()` calls, which overwrite the
+                // mapping instead of accumulating it) keeps every validated
+                // address available as a fallback.
+                let builder = librefang_http::direct_client_builder()
+                    .redirect(reqwest::redirect::Policy::none())
+                    .resolve_to_addrs(&ssrf_result.hostname, &ssrf_result.resolved);
                 let client = builder.build().expect("HTTP client build");
                 let request = match current_method.as_str() {
                     "POST" => client.post(&current_url).body(current_body.clone()),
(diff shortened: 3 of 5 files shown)


crates/librefang-runtime/src/tool_runner/web_legacy.rs
```diff
@@ -33,6 +33,18 @@ where
     }
 }
 
+fn build_legacy_fetch_client(
+    resolution: crate::web_fetch::SsrfResolution,
+) -> Result<reqwest::Client, ToolError> {
+    let builder = crate::http_client::direct_client_builder()
+        .timeout(std::time::Duration::from_secs(30))
+        .redirect(reqwest::redirect::Policy::none());
```
houko, line 41: Disabling redirects here is the right call, but with `Policy::none()` the caller now gets back a bare `HTTP 3xx` with whatever (often empty) body the redirecting host returned — the `Location` header is never read or surfaced, since `read_body_limited` only looks at the body.
Before this PR the legacy client silently followed the chain, so this is a behavior change agents will hit on any ordinary redirect (link shorteners, `http://`→`https://` canonicalization, `example.com`→`www.example.com`), and the response gives no hint of where the request was going, just a status code and no target to retry.
Worth considering surfacing the `Location` value in the returned text (e.g. `HTTP 302 -> <location>`) so a caller can still act on it (re-invoke `web_fetch` with the new URL, which gets its own fresh SSRF check) instead of hitting a dead end.
Not a security issue — just flagging as a UX/completeness gap versus the full `WebFetchEngine` path, which re-validates and follows redirects manually instead of dropping them.

---

crates/librefang-http/src/lib.rs
```diff
@@ -165,6 +165,14 @@ pub fn proxied_client_builder() -> reqwest::ClientBuilder {
     build_http_client(&active_proxy())
 }
 
+/// Build an HTTP client that never uses configured or environment proxies.
+///
+/// SSRF-sensitive callers that pin a hostname to locally validated socket addresses must connect directly.
+/// A proxy can resolve the original hostname itself, bypassing reqwest's DNS overrides and reopening DNS rebinding.
+pub fn direct_client_builder() -> reqwest::ClientBuilder {
```
houko, line 172: `direct_client_builder()` correctly closes the proxy-based DNS-rebinding bypass this comment describes (a proxy can re-resolve the original hostname itself, defeating reqwest's `.resolve()`/`resolve_to_addrs()` override).

But it's only wired into the legacy fetch path (`tool_runner/web_legacy.rs::build_legacy_fetch_client`). Two sibling call sites pin DNS after an SSRF check the exact same way and still build their client from `proxied_client_builder()`, which does not disable proxies:
- `WebFetchEngine::pinned_client` in `crates/librefang-runtime/src/web_fetch.rs:48-58` — the primary/enhanced (non-legacy) fetch path used by `fetch_with_options` / `send_with_pinned_redirects`.
- `host_net_fetch` in `crates/librefang-runtime/src/host_functions.rs:551-556` — the WASM guest's `host_net_fetch` host function.

If an HTTP(S) proxy is configured (`config.toml [proxy]` or `HTTP_PROXY`/`HTTPS_PROXY` env vars), both of those paths remain vulnerable to the same proxy-side-resolution SSRF bypass this PR fixes for the legacy fallback only. Is narrowing the fix to the legacy path intentional (with a tracked follow-up for the other two), or should `direct_client_builder()` be applied there too in this PR?

---

crates/librefang-runtime/src/web_fetch.rs
```diff
@@ -354,11 +354,8 @@ pub struct SsrfResolution {
 impl SsrfResolution {
     /// Apply the pinned DNS resolution to a [`reqwest::ClientBuilder`] so
     /// the actual HTTP request connects to the IPs we already validated.
-    pub fn pin_dns(self, mut builder: reqwest::ClientBuilder) -> reqwest::ClientBuilder {
-        for addr in &self.resolved {
-            builder = builder.resolve(&self.hostname, *addr);
-        }
-        builder
+    pub fn pin_dns(self, builder: reqwest::ClientBuilder) -> reqwest::ClientBuilder {
```
houko, line 359: Good catch on the `.resolve()`-in-a-loop bug — `reqwest::ClientBuilder::resolve(domain, addr)` overwrites the previous mapping for the same hostname on repeated calls, so the old loop silently kept only the last SSRF-validated address; `resolve_to_addrs` fixes that (and the new `legacy_client_uses_all_validated_addresses` test in `web_legacy.rs` guards it well).

The identical buggy pattern is still present, untouched, in `host_functions.rs:551-556` (`host_net_fetch`'s per-hop pinned client for WASM guest fetches):
```rust
let mut builder = librefang_http::proxied_client_builder()
    .redirect(reqwest::redirect::Policy::none());
for addr in &ssrf_result.resolved {
    builder = builder.resolve(&ssrf_result.hostname, *addr);
}
```
This will drop all but the last validated address for any WASM guest fetch, same failure mode as the bug fixed here. Worth a follow-up (or folding into this PR) since it's the same root cause and the same file family this PR is already hardening.

---

houko (commented):
Reviewed the SSRF hardening. The legacy-path fix itself (async SSRF check, DNS pinning via `resolve_to_addrs`, no-proxy direct client, disabled auto-redirects) looks correct and well-tested. Left two comments flagging that the same vulnerability classes (proxy-side DNS re-resolution defeating the pin; `.resolve()`-in-a-loop dropping all but the last validated address) remain open in two sibling call sites this PR doesn't touch. Convention-wise (changelog fragment, commit format, no AI attribution, prose wrapping) everything checks out.

---

crates/librefang-runtime/src/host_functions.rs
```diff
@@ -546,13 +546,11 @@ fn host_net_fetch(state: &GuestState, params: &serde_json::Value) -> serde_json:
...
-                let mut builder = librefang_http::proxied_client_builder()
-                    .redirect(reqwest::redirect::Policy::none());
-                for addr in &ssrf_result.resolved {
-                    builder = builder.resolve(&ssrf_result.hostname, *addr);
-                }
+                // DNS-pinned, non-redirecting, direct (no-proxy) client: connect to the exact IPs we just validated for this hop.
+                // A configured proxy would re-resolve `hostname` itself, bypassing the pin below and reopening DNS rebinding (#6761); `resolve_to_addrs` (rather than repeated `resolve()` calls, which overwrite the mapping instead of accumulating it) keeps every validated address available as a fallback.
+                let builder = librefang_http::direct_client_builder()
+                    .redirect(reqwest::redirect::Policy::none())
+                    .resolve_to_addrs(&ssrf_result.hostname, &ssrf_result.resolved);
```
houko, line 553: This hop's client now gets the same no-proxy + `resolve_to_addrs` hardening as `tool_web_fetch_legacy`, but unlike that path it has no dedicated regression test proving the fix. `tool_web_fetch_legacy` got two new wiremock tests (`legacy_client_does_not_follow_redirects`, `legacy_client_uses_all_validated_addresses` in `tool_runner/web_legacy.rs`); `host_net_fetch` got none, and the PR's own verification list only cites `librefang-http`, `web_legacy::tests`, and `web_fetch::tests`.

Worth noting for the record: `is_ssrf_target` (this file) has no `allowed_hosts` parameter the way `web_fetch::check_ssrf_async` does, so there's no way to point it at a loopback wiremock server for a true end-to-end test without a small follow-on change — that's presumably why coverage stopped short here. Flagging so it's a conscious gap rather than an overlooked one; not blocking given the fix mirrors the already-tested `SsrfResolution::pin_dns` code path line-for-line.

---

crates/librefang-runtime/src/tool_runner/web_legacy.rs
```diff
@@ -173,12 +184,98 @@ pub(super) async fn tool_web_search_legacy(input: &serde_json::Value) -> ToolRes
...
+        )
+        .await;
+        assert!(
+            matches!(r, Err(ToolError::InvalidParameter { name: "url", .. })),
+            "expected SSRF block, got: {r:?}"
+        );
+    }
+
     #[tokio::test]
     async fn web_search_legacy_missing_query_is_missing_parameter() {
```
houko, line 280: Minor/cosmetic: the PR description's verification section says `cargo test -p librefang-runtime tool_runner::web_legacy::tests --lib (4 passed)`, but this file now has 5 `#[tokio::test]`s in the `tests` module (the two new ones plus the pre-existing three, including `web_fetch_legacy_blocks_private_ip` added in the last commit). Looks like the count in the PR body just wasn't updated after the final commit — worth a quick edit so reviewers don't second-guess the test count against the actual file.

---

changelog.d/security/6761-legacy-web-fetch-ssrf.md
houko, line 1: Informational, not a blocker on this PR: #6762 and #6764 are both stacked on an earlier commit of this branch (75f7410, "bypass proxies for pinned SSRF fetches") rather than the current tip (8dc5498). Since then this branch added 7a7346c ("bypass proxies and fix DNS pin in remaining SSRF paths"), which already applies `direct_client_builder` to `WebFetchEngine::pinned_client` (crates/librefang-runtime/src/web_fetch.rs) and to the WASM `host_net_fetch` path (crates/librefang-runtime/src/host_functions.rs) — the same two targets #6762's description says it covers ("enhanced WebFetch and A2A pinned clients", "WASM host fetch"). Once #6762 rebases onto this branch's current tip, its WebFetch/host-fetch hunks should collapse to no-ops (or need re-diffing against what's already here); only its A2A-client change looks genuinely additive. Flagging so the author can rebase #6762 before merge rather than hit a confusing diff.

---

houko (commented):
Reviewed the SSRF hardening for correctness against the four security criteria: private/loopback/link-local/cloud-metadata IP blocking (pre-existing `validate_resolved`/`is_cloud_metadata_ip`, unaffected here and already well covered), resolved-IP-not-hostname checking (this PR's core fix — confirmed `reqwest::ClientBuilder::resolve()` calls `resolve_to_addrs` with a single-element slice and *overwrites* the same domain key on each call rather than accumulating, so the old per-address loop in `host_functions.rs`/`web_fetch.rs::pin_dns` silently kept only the last resolved address; switching to `resolve_to_addrs` fixes that), no-proxy pinning (confirmed `direct_client_builder` correctly calls `.no_proxy()`, which is required because `reqwest::Client::builder()` reads `HTTP_PROXY`/`HTTPS_PROXY` env vars by default even with no explicit `Proxy` configured — without it a configured proxy could still re-resolve the pinned hostname itself, reopening the rebinding window), and redirect handling (`tool_web_fetch_legacy` now disables redirects entirely via `Policy::none()` with no manual follow, which is safe-by-default since there's nothing to re-validate; the fuller `WebFetchEngine`/`host_net_fetch` paths already re-validate + re-pin every hop, pre-existing and untouched in policy here). No blocking issues found. Left three non-blocking comments: a test-coverage gap on the `host_net_fetch` hardening (mirrors the `web_legacy.rs` fix but lacks its own wiremock regression test), a stale test count in the PR body, and an informational note about #6762/#6764 being stacked on an older commit of this branch and likely overlapping once rebased.

---

PR #6762, opened 2026-08-08
fix(runtime): enforce direct transport for pinned SSRF clients

houko:
## Summary
- use direct/no-proxy transport for enhanced WebFetch and A2A pinned clients
- use direct transport plus all validated addresses for WASM host fetch
- document that proxy-side DNS must not bypass local SSRF resolution

## Security impact
Extends the no-proxy DNS-pin invariant from #6761 to the remaining runtime clients that claim check-to-connect pinning. This PR is stacked on #6761 and should merge after it.

## Verification
- `cargo test -p librefang-runtime web_fetch::tests --lib` (43 passed)
- `cargo test -p librefang-runtime a2a::tests --lib` (21 passed)
- `cargo test -p librefang-runtime host_functions::tests --lib` (44 passed)
- `git diff --check`

### DIFF
--- a/changelog.d/security/6762-pinned-runtime-clients-no-proxy.md
+++ b/changelog.d/security/6762-pinned-runtime-clients-no-proxy.md
@@ -1,0 +1,1 @@
+Require direct transport for DNS-pinned WebFetch, A2A, and WASM host requests so proxies cannot re-resolve validated targets (#6762) (@houko)
--- a/crates/librefang-runtime/src/a2a.rs
+++ b/crates/librefang-runtime/src/a2a.rs
@@ -928,9 +928,9 @@ impl A2aClient {
     /// [`build_client_for_url`], which:
     /// 1. Runs `web_fetch::check_ssrf` on the URL to resolve DNS once,
     ///    validate the addresses, and obtain a [`SsrfResolution`].
-    /// 2. Pins those exact addresses via `ClientBuilder::resolve` so the
-    ///    HTTP stack cannot re-resolve and connect to a different IP — this
-    ///    closes the DNS-rebinding TOCTOU window (#3563).
+    /// 2. Pins those exact addresses on a direct (no-proxy) client so neither
+    ///    the HTTP stack nor a proxy can re-resolve and connect to a different
+    ///    IP — this closes the DNS-rebinding TOCTOU window (#3563).
     /// 3. Installs a custom redirect policy that re-runs `check_ssrf` against
     ///    every redirect target (#3782), since the DNS pin only protects the
     ///    original hostname.
@@ -953,7 +953,7 @@ impl A2aClient {
         // already short-circuits on `response.status().is_redirection()`.
         let redirect_policy = reqwest::redirect::Policy::none();
 
-        let builder = crate::http_client::proxied_client_builder()
+        let builder = crate::http_client::direct_client_builder()
             .timeout(std::time::Duration::from_secs(30))
             .redirect(redirect_policy)
             // Bug #3785: disable transport-layer decompression. Otherwise
--- a/crates/librefang-runtime/src/a2a.rs
+++ b/crates/librefang-runtime/src/a2a.rs
@@ -928,9 +928,7 @@ impl A2aClient {
     /// [`build_client_for_url`], which:
     /// 1. Runs `web_fetch::check_ssrf` on the URL to resolve DNS once,
     ///    validate the addresses, and obtain a [`SsrfResolution`].
-    /// 2. Pins those exact addresses on a direct (no-proxy) client so neither
-    ///    the HTTP stack nor a proxy can re-resolve and connect to a different
-    ///    IP — this closes the DNS-rebinding TOCTOU window (#3563).
+    /// 2. Pins those exact addresses on a direct (no-proxy) client so neither the HTTP stack nor a proxy can re-resolve and connect to a different IP — this closes the DNS-rebinding TOCTOU window (#3563).
     /// 3. Installs a custom redirect policy that re-runs `check_ssrf` against
     ///    every redirect target (#3782), since the DNS pin only protects the
     ///    original hostname.
--- a/crates/librefang-runtime/src/host_functions.rs
+++ b/crates/librefang-runtime/src/host_functions.rs
@@ -546,9 +546,8 @@ fn host_net_fetch(state: &GuestState, params: &serde_json::Value) -> serde_json:
                     return e;
                 }
 
-                // Direct, DNS-pinned, non-redirecting client: connect to the
-                // exact IPs validated for this hop. Proxies are disabled
-                // because proxy-side DNS would bypass reqwest's overrides.
+                // Direct, DNS-pinned, non-redirecting client: connect to the exact IPs validated for this hop.
+                // Proxies are disabled because proxy-side DNS would bypass reqwest's overrides.
                 let builder = librefang_http::direct_client_builder()
                     .redirect(reqwest::redirect::Policy::none())
                     .resolve_to_addrs(&ssrf_result.hostname, &ssrf_result.resolved);
--- a/crates/librefang-runtime/src/host_functions.rs
+++ b/crates/librefang-runtime/src/host_functions.rs
@@ -546,13 +546,12 @@ fn host_net_fetch(state: &GuestState, params: &serde_json::Value) -> serde_json:
                     return e;
                 }
 
-                // DNS-pinned, non-redirecting client: connect to the exact IPs
-                // we just validated for this hop.
-                let mut builder = librefang_http::proxied_client_builder()
-                    .redirect(reqwest::redirect::Policy::none());
-                for addr in &ssrf_result.resolved {
-                    builder = builder.resolve(&ssrf_result.hostname, *addr);
-                }
+                // Direct, DNS-pinned, non-redirecting client: connect to the
+                // exact IPs validated for this hop. Proxies are disabled
+                // because proxy-side DNS would bypass reqwest's overrides.
+                let builder = librefang_http::direct_client_builder()
+                    .redirect(reqwest::redirect::Policy::none())
+                    .resolve_to_addrs(&ssrf_result.hostname, &ssrf_result.resolved);
                 let client = builder.build().expect("HTTP client build");
                 let request = match current_method.as_str() {
                     "POST" => client.post(&current_url).body(current_body.clone()),
--- a/crates/librefang-runtime/src/web_fetch.rs
+++ b/crates/librefang-runtime/src/web_fetch.rs
@@ -33,9 +33,7 @@ impl WebFetchEngine {
 
     /// Build a per-request reqwest client pinned to the SSRF-validated IPs.
     ///
-    /// Uses the resolved addresses from [`check_ssrf`] to configure DNS
-    /// pinning on a direct (no-proxy) builder, preventing both local and
-    /// proxy-side DNS rebinding TOCTOU attacks.
+    /// Uses the resolved addresses from [`check_ssrf`] to configure DNS pinning on a direct (no-proxy) builder, preventing both local and proxy-side DNS rebinding TOCTOU attacks.
     ///
     /// Auto-redirects are **disabled** (`Policy::none`). A `Policy::custom`
     /// that re-runs `check_ssrf` on each hop is not enough: reqwest
--- a/crates/librefang-runtime/src/web_fetch.rs
+++ b/crates/librefang-runtime/src/web_fetch.rs
@@ -34,7 +34,8 @@ impl WebFetchEngine {
     /// Build a per-request reqwest client pinned to the SSRF-validated IPs.
     ///
     /// Uses the resolved addresses from [`check_ssrf`] to configure DNS
-    /// pinning on the builder, preventing DNS-rebinding TOCTOU attacks.
+    /// pinning on a direct (no-proxy) builder, preventing both local and
+    /// proxy-side DNS rebinding TOCTOU attacks.
     ///
     /// Auto-redirects are **disabled** (`Policy::none`). A `Policy::custom`
     /// that re-runs `check_ssrf` on each hop is not enough: reqwest
@@ -46,7 +47,7 @@ impl WebFetchEngine {
     /// [`Self::send_with_pinned_redirects`], which re-validates AND re-pins
     /// every hop so the IP we checked is the IP we connect to.
     pub(crate) fn pinned_client(&self, resolution: SsrfResolution) -> reqwest::Client {
-        let builder = crate::http_client::proxied_client_builder()
+        let builder = crate::http_client::direct_client_builder()
             .timeout(std::time::Duration::from_secs(self.config.timeout_secs))
             .redirect(reqwest::redirect::Policy::none())
             .gzip(true)


crates/librefang-runtime/src/web_fetch.rs
```diff
@@ -34,7 +34,8 @@ impl WebFetchEngine {
     /// Build a per-request reqwest client pinned to the SSRF-validated IPs.
     ///
     /// Uses the resolved addresses from [`check_ssrf`] to configure DNS
-    /// pinning on the builder, preventing DNS-rebinding TOCTOU attacks.
+    /// pinning on a direct (no-proxy) builder, preventing both local and
+    /// proxy-side DNS rebinding TOCTOU attacks.
```
houko: The three call-site swaps to `direct_client_builder` (here, `a2a.rs::build_client_for_url`, `host_functions.rs::host_net_fetch`) look correct and complete — I checked for other pinned-client construction sites in these files and didn't find any that were missed, and `web_fetch_to_file.rs` inherits the fix transitively via `pinned_client()`.
One gap: none of the existing tests actually exercise the property this PR adds (that a configured proxy is bypassed for these pinned clients).
The existing `test_direct_client_builder_builds_client` in `librefang-http` only asserts the builder builds, not that it ignores proxy config.
Worth a regression test — e.g. set `HTTP_PROXY`/`HTTPS_PROXY` to an unreachable address before calling `pinned_client`/`build_client_for_url`/`host_net_fetch`'s builder and assert the request still succeeds (would hang/fail if actually routed through the bogus proxy) — so a future refactor that re-introduces `proxied_client_builder` on one of these paths gets caught by CI instead of silently reopening the proxy-side DNS-rebind bypass.
Not blocking this PR; flagging since it's the kind of regression a later "harmless" cleanup could easily reintroduce.

---

changelog.d/security/6762-pinned-runtime-clients-no-proxy.md
```diff
@@ -0,0 +1 @@
+Require direct transport for DNS-pinned WebFetch, A2A, and WASM host requests so proxies cannot re-resolve validated targets (#6762) (@houko)
```
houko, line 1: Checked this against #6764 ("fix(api): enforce direct transport for pinned requests"), which looked like it might be duplicate/conflicting work at a glance.
On closer inspection they don't overlap: this PR touches `librefang-runtime` (`a2a.rs`, `host_functions.rs`, `web_fetch.rs`), #6764 touches `librefang-api` (`attachments.rs`, `webhooks.rs`), and both call the same pre-existing `direct_client_builder()` helper (added earlier in the `fix/legacy-web-fetch-ssrf` stack, not introduced by either PR).
So this reads as one fix intentionally split across the two crates it applies to (same author, no competing implementation), not a parallel-duplicate situation CLAUDE.md's "one PR per issue" rule is meant to catch.
No action needed from my side — just documenting the cross-check per the review routine's request, in case a maintainer wants to land them together / reference each other in the merge commit.

---

houko (commented):
Automated daily review pass. Diff (incremental over base fix/legacy-web-fetch-ssrf) reviewed for correctness, SSRF/proxy-bypass logic, changelog convention, and overlap with #6764. Found the change correct and complete; left two non-blocking notes inline (test-coverage gap, and a cross-check against #6764 confirming it's complementary, not a duplicate).

---

crates/librefang-runtime/src/web_fetch.rs
```diff
@@ -46,7 +45,7 @@ impl WebFetchEngine {
     /// [`Self::send_with_pinned_redirects`], which re-validates AND re-pins
     /// every hop so the IP we checked is the IP we connect to.
     pub(crate) fn pinned_client(&self, resolution: SsrfResolution) -> reqwest::Client {
-        let builder = crate::http_client::proxied_client_builder()
+        let builder = crate::http_client::direct_client_builder()
             .timeout(std::time::Duration::from_secs(self.config.timeout_secs))
             .redirect(reqwest::redirect::Policy::none())
```
houko, line 50: The functional swap here (and in `a2a.rs` / `host_functions.rs`) looks correct — `direct_client_builder()` already exists and is unit-tested in `librefang-http`, and `resolve_to_addrs` is a real `reqwest` 0.13 API whose signature matches `SsrfResolution::resolved: Vec<SocketAddr>`.

One gap: none of the three call sites (nor `librefang-http`) has a regression test that actually asserts the pinned client ignores a configured proxy — the exact bug class this PR closes.
`test_direct_client_builder_builds_client` only checks the builder builds; it doesn't assert `NO_PROXY`/no-proxy behavior against a live `HTTP_PROXY`/`HTTPS_PROXY` env var or `[proxy]` config value.
Something like: set `HTTP_PROXY` to a local listener that would record a hit if contacted, build a pinned client, issue a request to a validated address, and assert the listener saw nothing — would be the test that catches a regression here if a future refactor swaps `direct_client_builder()` back to `proxied_client_builder()` on one of these paths.

Not blocking — flagging since CLAUDE.md treats "no regression test for the invariant a security fix restores" as worth a second look, and I'm not confident enough in the right mock-proxy harness shape to add it myself in this pass.

---

changelog.d/security/6762-pinned-runtime-clients-no-proxy.md
houko, line 1: Note for the maintainer (not a request for action on this PR): #6764 ("fix(api): enforce direct transport for pinned requests") has a very similar title and the same base branch (`fix/legacy-web-fetch-ssrf`, stacked on #6761), which could look like a duplicate at a glance.
Checked the diffs — they don't overlap: this PR touches `librefang-runtime` (`web_fetch.rs`, `a2a.rs`, `host_functions.rs`), #6764 touches `librefang-api` (webhook test delivery, URL attachment downloads).
They read as sibling PRs extending the same `direct_client_builder()` no-proxy invariant from #6761 to different crates, not a duplicate.
Leaving this as a factual note for whoever reconciles the stack.

---

houko (commented):
Automated daily review pass. Reviewed the diff against the actual base (`fix/legacy-web-fetch-ssrf`, not `main`), applied one mechanical CLAUDE.md prose-wrap fix directly, and left two comments below (one code-level, one a duplicate-PR cross-check note).

---

houko:
Disclosure: this automated review run's own fix commit `deb95e4a` (prose-wrap cleanup) was mistakenly authored under this review environment's default git identity, `Claude <noreply@anthropic.com>` — a violation of this repo's own no-AI-attribution policy, the same class of issue already flagged elsewhere in this repo's history. Root cause: the environment's global git config defaulted to that identity; this has been corrected mid-run (to the authenticated `houko` identity) so further commits from this routine don't repeat it. Recommend squashing/amending `deb95e4a` under a real author before merge.

---

crates/librefang-runtime/src/web_fetch.rs
```diff
@@ -46,7 +45,7 @@ impl WebFetchEngine {
     /// [`Self::send_with_pinned_redirects`], which re-validates AND re-pins
     /// every hop so the IP we checked is the IP we connect to.
     pub(crate) fn pinned_client(&self, resolution: SsrfResolution) -> reqwest::Client {
-        let builder = crate::http_client::proxied_client_builder()
+        let builder = crate::http_client::direct_client_builder()
```
houko, line 48: Per repo CLAUDE.md's HTTP-client testing mandate: this swap from `proxied_client_builder()` to `direct_client_builder()` (here and in `a2a.rs`/`host_functions.rs`) has no dedicated regression test proving the proxy is actually bypassed for these three call sites. The PR's own verification list only re-runs pre-existing suites (`web_fetch::tests`, `a2a::tests`, `host_functions::tests`), none of which assert on proxy behavior — they only assert SSRF/redirect/pin-address semantics, which were true before this change too.

`librefang-http`'s own new test (`test_direct_client_builder_builds_client`, added in #6761) only checks the client builds, not that a configured `HTTP_PROXY`/`HTTPS_PROXY` is actually ignored end-to-end.

Suggested shape for a real regression test (aware this needs care around the process-global `HTTP_PROXY` env var and parallel test execution — e.g. a `#[serial]`-style guard or a dedicated single-threaded test binary): point `HTTP_PROXY` at an address nothing listens on, then assert that a request through `pinned_client()` (or the a2a/host_functions equivalents) still reaches a local `wiremock` server successfully — proving the (broken) proxy was never used. Absent that, this security fix is unverified beyond "it compiles and old tests still pass."

---

houko (commented):
One test-coverage note inline. Separately, flagging a related gap for maintainer judgment (touches a third crate, `librefang-hands`, so leaving it as a question rather than pushing a fix into this stack):

`crates/librefang-hands/src/hands_hub.rs::HandsHubClient::build` (used via `with_pinned_url` from `routes/skills/hands.rs`'s caller-supplied `registry_url` SSRF path) builds its DNS-pinned client from a bare `reqwest::ClientBuilder::new()` — never `librefang_http::proxied_client_builder()` or the new `direct_client_builder()`, and never calls `.no_proxy()`. Reqwest auto-detects `HTTP_PROXY`/`HTTPS_PROXY` from the environment by default, so this pinned SSRF-validated client is exposed to the exact proxy-side DNS-rebinding bypass this PR (and #6761/#6764) closes everywhere else — a caller-supplied hands registry URL that resolves through a configured proxy could still have its DNS re-resolved by the proxy rather than connecting to the address `check_ssrf` validated. Given this PR's stated goal is "the remaining runtime clients that claim check-to-connect pinning," this one looks like it was missed because it lives outside `librefang-runtime`/`librefang-api`. Worth a follow-up in the same vein (swap to `direct_client_builder()` + `resolve_to_addrs`) — happy to open that as a separate PR if useful, since it's a different crate than the ones already touched here.

---

crates/librefang-runtime/src/web_fetch.rs
houko, line 48: The switch from `proxied_client_builder()` to `direct_client_builder()` here (and at the equivalent call sites in `a2a.rs:954` and `host_functions.rs:551`) is correct: `direct_client_builder()` = `build_http_client(&ProxyConfig::default()).no_proxy()`, and `ClientBuilder::no_proxy()` clears both explicit proxy config *and* `auto_sys_proxy` (env-var `HTTP_PROXY`/`HTTPS_PROXY` detection) — verified against the vendored reqwest 0.13.4 source. The `host_functions.rs` move from a manual per-address `.resolve()` loop to `resolve_to_addrs(&hostname, &resolved)` is also a valid simplification (signature matches `Vec<SocketAddr>` → `&[SocketAddr]`).

Gap: none of the three switched call sites (nor the pre-existing `direct_client_builder` unit test in `librefang-http`) have a regression test that actually proves a *configured* proxy (env `HTTP_PROXY`/`HTTPS_PROXY`, or `ProxyConfig`) is bypassed for a pinned client — the existing tests only assert the builder builds. Given this PR's entire purpose is closing a proxy-mediated SSRF bypass, a test that sets a proxy env var (or spins a local mock proxy) and asserts the pinned request does *not* traverse it would be the thing that actually locks the fix in place; right now a future revert of `direct_client_builder()` → `proxied_client_builder()` at any of these three sites would go green.

---

houko (commented):
Reviewed the delta against the stacked base (`git diff origin/fix/legacy-web-fetch-ssrf...origin/fix/pinned-http-no-proxy`), not against `main`.

Findings:
1. **Correct fix, missing regression test** — inline comment on `web_fetch.rs:48`. The `direct_client_builder()` swap genuinely disables proxy routing (config + env, verified against reqwest 0.13.4 source) for all three pinned-client sites (`web_fetch.rs`, `a2a.rs`, `host_functions.rs`), but no test proves a configured proxy is actually bypassed — only that the builder builds. Recommend adding one before merge.
2. **Head commit author identity violates this repo's no-AI-attribution policy** — `git log` on this branch shows commit `deb95e4a` ("docs(runtime): reflow touched SSRF comments to one sentence per line") authored as `Claude <noreply@anthropic.com>`, independently confirmed via `git log --format='%an <%ae>'` (not merely asserted by the existing PR comment claiming the same, which I'm treating as unverified/untrusted on its own). This is the exact case the repo's `commit-msg` hook's author-identity check exists to catch. Recommend the author amend/reauthor that commit before merge; I did not attempt to rewrite it myself since that requires a force-push to someone else's already-commented-on branch.

No changelog, formatting, or compile issues — fragment is correctly placed in `changelog.d/security/`, doc comments are single-sentence-per-line, and the diff compiles (types for `resolve_to_addrs` match, confirmed against reqwest 0.13.4 source).

---

crates/librefang-runtime/src/web_fetch.rs
houko, line 48: Per CLAUDE.md's testing expectations for a no-proxy-enforcement fix, none of the three call sites this PR touches (here, `A2aClient::build_client_for_url`, and `host_net_fetch`) have a regression test proving the pinned client actually ignores `HTTP_PROXY`/`HTTPS_PROXY`/`ALL_PROXY` env vars at request time.
The only existing test for `direct_client_builder` (`librefang-http/src/lib.rs::test_direct_client_builder_builds_client`) just confirms the builder builds — it never sets a proxy env var and asserts the outbound connection bypasses it.
Without that assertion, a future refactor that swaps `direct_client_builder()` back to `proxied_client_builder()` (or a `reqwest` upgrade that changes `no_proxy()` semantics) would silently reopen the proxy-side DNS-rebinding hole this PR closes, and none of the 108 passing tests cited in the PR description would catch it.
Suggest a test that spins up a local TCP listener acting as a bogus "proxy" on `HTTP_PROXY`/`HTTPS_PROXY`, points a pinned client at a real validated address, and asserts the connection lands on the pinned address directly rather than the fake proxy (a connection to the proxy listener, or a proxy-specific error, would indicate a regression).

---

houko (commented):
Reviewed the diff against the direct-transport/no-proxy invariant from #6761.

What checks out:
- All three pinned-client construction sites in this diff (`WebFetchEngine::pinned_client`, `A2aClient::build_client_for_url`, `host_net_fetch`) now build on `direct_client_builder()` / `librefang_http::direct_client_builder()`, which calls `reqwest::ClientBuilder::no_proxy()` — this disables reqwest's env-based proxy auto-detection entirely (`HTTP_PROXY`/`HTTPS_PROXY`/`ALL_PROXY`/`NO_PROXY`), not just the explicit `ProxyConfig` fields.
- `web_fetch_to_file.rs`, the other consumer of the SSRF/DNS-pin pipeline, goes through `WebFetchEngine::send_with_pinned_redirects` → `self.pinned_client(...)`, so it inherits the fix rather than needing its own change — no missed sibling site there.
- The two `pin_dns` call sites (`a2a.rs`, `web_fetch.rs`) are both updated; `host_functions.rs` uses the equivalent `resolve_to_addrs` directly with the same `direct_client_builder` base.
- Changelog fragment (`changelog.d/security/6762-pinned-runtime-clients-no-proxy.md`) is present and correctly formatted.
- Touched doc comments already follow the one-sentence-per-line prose rule.

One test-coverage gap flagged inline on `web_fetch.rs` (applies equally to the A2A and WASM host-fetch sites): no test proves the pinned/direct client ignores proxy env vars at connection time, only that the builder compiles/builds.

Out-of-scope observation (not part of this diff, so no inline comment): `tool_runner/web_legacy.rs`'s `check_ssrf` call sites (`crate::http_client::proxied_client_builder()` calls) validate the URL but never pin DNS to the resolved addresses at all, so they keep the classic DNS-rebinding TOCTOU window this PR series is otherwise closing. Worth a follow-up in the same stack if that legacy path is still reachable.

---

crates/librefang-runtime/src/host_functions.rs
```diff
@@ -546,13 +546,11 @@ fn host_net_fetch(state: &GuestState, params: &serde_json::Value) -> serde_json:
...
-                // DNS-pinned, non-redirecting client: connect to the exact IPs
-                // we just validated for this hop.
-                let mut builder = librefang_http::proxied_client_builder()
-                    .redirect(reqwest::redirect::Policy::none());
-                for addr in &ssrf_result.resolved {
-                    builder = builder.resolve(&ssrf_result.hostname, *addr);
-                }
+                // Direct, DNS-pinned, non-redirecting client: connect to the exact IPs validated for this hop.
+                // Proxies are disabled because proxy-side DNS would bypass reqwest's overrides.
+                let builder = librefang_http::direct_client_builder()
```
houko, line 551: This exact change (host_net_fetch: `proxied_client_builder` → `direct_client_builder` + `resolve_to_addrs`) already landed on the base branch as commit 7a7346c ("fix(runtime): bypass proxies and fix DNS pin in remaining SSRF paths"), which is ahead of the commit this PR is currently branched from.
Base `fix/legacy-web-fetch-ssrf` has moved from 75f7410 (this PR's merge-base) to 8dc5498.
Rebasing this branch onto the current base tip reproduces a real conflict here (confirmed locally): both sides touched the same comment lines with different wording, though the functional code (the `direct_client_builder()...resolve_to_addrs(...)` call itself) is byte-identical on both sides.
Please rebase onto `origin/fix/legacy-web-fetch-ssrf` before merge so this hunk (and the matching one in `web_fetch.rs`) drop out as already-applied, leaving only the genuinely new `a2a.rs` change as this PR's contribution.

---

crates/librefang-runtime/src/web_fetch.rs
houko, line 48: Same duplication as `host_functions.rs`: `WebFetchEngine::pinned_client`'s `proxied_client_builder` → `direct_client_builder` switch was already committed to the base branch at 7a7346c (base tip is now 8dc5498, this PR's merge-base is the older 75f7410).
The runtime behavior here is identical on both sides — only the doc-comment wording differs — but a local rebase simulation onto the current base tip produces a real conflict on these lines.
Please rebase before merge.

---

crates/librefang-runtime/src/a2a.rs
```diff
@@ -928,9 +928,7 @@ impl A2aClient {
     /// [`build_client_for_url`], which:
     /// 1. Runs `web_fetch::check_ssrf` on the URL to resolve DNS once,
     ///    validate the addresses, and obtain a [`SsrfResolution`].
-    /// 2. Pins those exact addresses via `ClientBuilder::resolve` so the
-    ///    HTTP stack cannot re-resolve and connect to a different IP — this
-    ///    closes the DNS-rebinding TOCTOU window (#3563).
+    /// 2. Pins those exact addresses on a direct (no-proxy) client so neither the HTTP stack nor a proxy can re-resolve and connect to a different IP — this closes the DNS-rebinding TOCTOU window (#3563).
     /// 3. Installs a custom redirect policy that re-runs `check_ssrf` against
     ///    every redirect target (#3782), since the DNS pin only protects the
     ///    original hostname.
```
houko, line 934: Minor, pre-existing doc/code mismatch adjacent to this PR's edit (item 2, just above, was updated; item 3 here was not): this says `build_client_for_url` "installs a custom redirect policy that re-runs `check_ssrf` against every redirect target," but the actual code a few lines down sets `redirect_policy = reqwest::redirect::Policy::none()` and refuses all redirects outright (the accompanying inline comment at the `redirect_policy` binding explains why that approach was chosen over a custom re-validating policy).
Since item 2 right above this was just touched, worth fixing item 3 in the same pass to stop the doc contradicting the code beneath it — not blocking.

---

houko (commented):
Security review of the actual mechanism: the `direct_client_builder()` + `resolve_to_addrs(hostname, &resolved)` + `Policy::none()` pattern is sound at all three call sites (`a2a.rs::build_client_for_url`, `host_functions.rs::host_net_fetch`, `web_fetch.rs::pinned_client`).
`direct_client_builder` (`crates/librefang-http/src/lib.rs:173-175`) both skips explicit config-proxy application and calls reqwest's `.no_proxy()`, which clears any queued proxies and disables `auto_sys_proxy` — so `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` env vars cannot reintroduce a bypass, closing the exact hole this PR targets.
Redirects are either refused outright (`a2a.rs`, `web_fetch.rs`'s single-hop builder combined with the manual re-validating loop in `send_with_pinned_redirects`) or re-validated and re-pinned per hop before the next connect (`host_functions.rs`'s manual loop) — DNS-rebinding between pin-check and connect is not possible because the client connects to the exact validated `SocketAddr`s via `resolve_to_addrs`, never re-resolving the hostname. No correctness issues in the diff itself.

Two things outside the diff's own lines that are worth the author's attention before merge:

1. **Stale base / incoming rebase conflict** (see inline comments on `host_functions.rs` and `web_fetch.rs`): this PR's merge-base with `fix/legacy-web-fetch-ssrf` is 75f7410, but that base branch has since advanced to 8dc5498 and independently landed the identical `host_net_fetch` / `pinned_client` fix in commit 7a7346c.
   Confirmed locally that rebasing this branch onto the current base tip produces real conflicts on the touched comment lines in both files (functional code is byte-identical, only prose differs).
   Recommend rebasing onto `origin/fix/legacy-web-fetch-ssrf` before merge — `a2a.rs` is the only file where this PR still adds something the base doesn't already have.
2. **Commit author identity**: `deb95e4` ("docs(runtime): reflow touched SSRF comments to one sentence per line") is authored by `Claude <noreply@anthropic.com>` (`git log --format='%an|%ae' deb95e4`), not the PR author's identity used on the other four commits.
   Repo convention (`CLAUDE.md` → commit-msg hook) rejects commits whose author identity resolves to Claude/Anthropic even when the message text itself is clean; this one appears to have bypassed that check (hooks not installed locally, or pushed via an API path the hook doesn't gate).
   Recommend amending that commit's author before merge, since a merge/rebase strategy that preserves individual commits would carry the flagged identity into `main`'s history.

Not requesting changes since neither item touches this PR's own correctness — flagging for the author/maintainer to resolve before merge, especially since #1 will otherwise surface as a literal conflict.

---

PR #6763, opened 2026-08-08
fix(runtime): harden link URL SSRF prefilter

houko:
## Summary
- replace hand-written URL authority parsing with `url::Url`
- reject userinfo-hidden private targets, complete local/link-local ranges, and alternate IP encodings
- cover IPv4-mapped/compatible IPv6, NAT64/6to4 embeddings, numeric IPv4, and trailing-dot metadata hosts
- document that this synchronous extraction filter is network-free and actual fetches still require WebFetch DNS validation

## Security impact
Prevents private and metadata URLs from being admitted into link context through parser confusion or alternate address spellings. DNS rebinding remains enforced at the actual WebFetch connection boundary.

## Verification
- `cargo test -p librefang-runtime link_understanding::tests --lib` (16 passed)
- `git diff --check`

### DIFF
--- a/crates/librefang-runtime/src/link_understanding.rs
+++ b/crates/librefang-runtime/src/link_understanding.rs
@@ -220,10 +220,7 @@ mod tests {
     #[test]
     fn test_ssrf_blocks_nat64_embedded_ips() {
         // 64:ff9b::/96 (RFC 6052) embeds an IPv4 address in the low 32 bits.
-        // `is_private_ip`/`is_cloud_metadata_ip` unwrap this internally for
-        // RFC1918/link-local/metadata ranges, but loopback and unspecified
-        // are only caught by the explicit recursion added above — regression
-        // coverage for that gap.
+        // `is_private_ip`/`is_cloud_metadata_ip` unwrap this internally for RFC1918/link-local/metadata ranges, but loopback and unspecified are only caught by the explicit recursion added above — regression coverage for that gap.
         assert!(is_private_url("http://[64:ff9b::7f00:1]/admin")); // embeds 127.0.0.1
         assert!(is_private_url("http://[64:ff9b::]/admin")); // embeds 0.0.0.0
         assert!(is_private_url("http://[64:ff9b::a00:1]/admin")); // embeds 10.0.0.1
--- a/crates/librefang-runtime/src/link_understanding.rs
+++ b/crates/librefang-runtime/src/link_understanding.rs
@@ -215,6 +215,17 @@ mod tests {
         assert!(is_private_url("http://[2002:7f00:1::]/admin"));
         assert!(is_private_url("http://2130706433/admin"));
         assert!(is_private_url("http://0x7f000001/admin"));
+        assert!(is_private_url("http://0177.0.0.1/admin"));
+    }
+
+    #[test]
+    fn test_ssrf_blocks_ipv6_unique_local_and_link_local() {
+        // fc00::/7 (ULA, RFC 4193) and fe80::/10 (link-local) are matched by
+        // `crate::web_fetch::is_private_ip`'s IPv6 branch.
+        assert!(is_private_url("http://[fc00::1]/admin"));
+        assert!(is_private_url("http://[fdff:ffff:ffff:ffff::1]/admin"));
+        assert!(is_private_url("http://[fe80::1]/admin"));
+        assert!(!is_private_url("http://[2001:4860:4860::8888]/admin")); // public (Google DNS)
     }
 
     #[test]
--- a/crates/librefang-runtime/src/link_understanding.rs
+++ b/crates/librefang-runtime/src/link_understanding.rs
@@ -17,9 +17,8 @@ pub struct LinkSummary {
 
 /// Extract URLs from text, with a network-free SSRF prefilter.
 ///
-/// Returns up to `max` valid, unique URLs that do not contain private literal
-/// targets or restricted hostnames. Any later fetch must still use the shared
-/// WebFetch SSRF resolver and pinned transport to validate DNS at connect time.
+/// Returns up to `max` valid, unique URLs that do not contain private literal targets or restricted hostnames.
+/// Any later fetch must still use the shared WebFetch SSRF resolver and pinned transport to validate DNS at connect time.
 pub fn extract_urls(text: &str, max: usize) -> Vec<String> {
     // Simple but effective URL regex
     let url_pattern = regex_lite::Regex::new(
@@ -54,8 +53,7 @@ pub fn extract_urls(text: &str, max: usize) -> Vec<String> {
 }
 
 /// Check URL syntax, userinfo, restricted hostnames, and literal IP ranges.
-/// This intentionally performs no DNS resolution because link extraction is
-/// synchronous and does not initiate network I/O.
+/// This intentionally performs no DNS resolution because link extraction is synchronous and does not initiate network I/O.
 fn is_private_url(url: &str) -> bool {
     let parsed = match url::Url::parse(url) {
         Ok(parsed) if matches!(parsed.scheme(), "http" | "https") => parsed,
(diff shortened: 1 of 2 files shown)


crates/librefang-runtime/src/link_understanding.rs
```diff
@@ -15,9 +15,11 @@ pub struct LinkSummary {
     pub content_type: String,
 }
 
-/// Extract URLs from text, with SSRF validation.
+/// Extract URLs from text, with a network-free SSRF prefilter.
 ///
-/// Returns up to `max` valid, unique, non-private URLs.
+/// Returns up to `max` valid, unique URLs that do not contain private literal
+/// targets or restricted hostnames. Any later fetch must still use the shared
+/// WebFetch SSRF resolver and pinned transport to validate DNS at connect time.
```
houko: Nit (CLAUDE.md prose-wrapping convention): doc comments must break only at sentence boundaries, one sentence per line, no fixed-column hard-wrap. This new doc comment (and the ones at `is_private_url`, lines 56-58, and the inline comment above `v4.octets()[0] == 0`, lines 119-120) are wrapped mid-sentence. Suggested reflow for this block:

```rust
/// Extract URLs from text, with a network-free SSRF prefilter.
///
/// Returns up to `max` valid, unique URLs that do not contain private literal targets or restricted hostnames.
/// Any later fetch must still use the shared WebFetch SSRF resolver and pinned transport to validate DNS at connect time.
```

Same treatment applies to the other two spots. Logic/security review of the SSRF fix itself looks solid (userinfo, private ranges, IPv4-mapped/compatible, NAT64 well-known, 6to4 embedding, decimal/hex IPv4, trailing-dot hosts all covered and tested; DNS-rebinding is correctly deferred to the WebFetch connect-time resolver per the added doc comment). Only this formatting nit found.

---

houko (commented):
Automated review pass on the SSRF hardening fix. One minor prose-wrapping nit found (inline suggestion below); otherwise the fix logic, tests, and changelog fragment all check out against CLAUDE.md conventions.

---

crates/librefang-runtime/src/link_understanding.rs
```diff
@@ -51,56 +52,81 @@ pub fn extract_urls(text: &str, max: usize) -> Vec<String> {
...
+        None => true,
+    }
+}
+
+fn is_blocked_literal_ip(ip: std::net::IpAddr) -> bool {
+    if let std::net::IpAddr::V6(v6) = ip {
+        let segments = v6.segments();
+        if segments[0] == 0x2002 {
+            let embedded = std::net::Ipv4Addr::new(
+                (segments[1] >> 8) as u8,
```
houko, line 89: This 6to4 (`2002::/16`) unwrap is only added here, in the link-context prefilter.
The actual fetch-time SSRF boundary, `web_fetch::canonical_ip` (crates/librefang-runtime/src/web_fetch.rs, ~line 596), unwraps IPv4-mapped and NAT64 (`64:ff9b::/96`) but not 6to4 — its own doc comment only claims those two forms.
That means a literal 6to4 URL embedding a loopback/RFC1918/metadata IPv4 (e.g. `http://[2002:7f00:0001::]/`) would now be filtered out of the agent's link-context hint by this PR, but if the agent (or any other `web_fetch` caller) fetches that same literal address directly, `validate_resolved` would not catch it — `is_loopback`/`is_private_ip`/`is_cloud_metadata_ip` all run against the un-unwrapped 6to4 IPv6 form there, so none of them match.
Given 6to4 relies on a deprecated public relay infrastructure that's largely decommissioned, real-world exploitability is low, but it's an inconsistency between the two functions that are supposed to encode the same threat model — worth a maintainer call on whether `canonical_ip`/`extract_nat64_well_known`'s sibling should gain a `2002::/16` branch too (mirroring the recursion this PR added here), rather than only hardening the advisory prefilter.
Not fixing this myself since it touches the shared, heavily-tested SSRF resolver used by every `web_fetch` caller, outside this PR's file scope.

---

houko (commented):
Daily automated review pass on this SSRF-hardening PR. Fixed a gap in the new prefilter itself (NAT64-embedded loopback/unspecified addresses weren't unwrapped before the loopback check) and flagged one architectural inconsistency for maintainer judgment — see inline comment. Decimal/hex/octal/shorthand IPv4, IPv4-mapped/compatible IPv6, 6to4, userinfo, and domain-case handling were all verified correct against the pinned `url` crate version.

---

crates/librefang-runtime/src/link_understanding.rs
```diff
@@ -15,9 +15,10 @@ pub struct LinkSummary {
...
 
-/// Extract URLs from text, with SSRF validation.
+/// Extract URLs from text, with a network-free SSRF prefilter.
 ///
-/// Returns up to `max` valid, unique, non-private URLs.
+/// Returns up to `max` valid, unique URLs that do not contain private literal targets or restricted hostnames.
+/// Any later fetch must still use the shared WebFetch SSRF resolver and pinned transport to validate DNS at connect time.
 pub fn extract_urls(text: &str, max: usize) -> Vec<String> {
     // Simple but effective URL regex
     let url_pattern = regex_lite::Regex::new(
```
houko, line 24: This `url_pattern` regex (line 25, `r#"https?://[^\s<>\[\](){}|\\^\`"']+..."#`) excludes `[` and `]` from the matched span, so `extract_urls` can never actually produce a candidate string for a bracketed IPv6-literal URL — IPv6 host literals in a URL are always bracket-enclosed (`http://[::1]/`, `http://[64:ff9b::7f00:1]/admin`, etc.).

Verified against the pattern directly (same negated-class semantics as `regex_lite`):
```python
>>> re.findall(pattern, "Check http://[::1]/admin and http://[64:ff9b::7f00:1]/x and http://2130706433/admin")
['http://2130706433/admin']
```
Only the bracket-free decimal-IP form survives; both IPv6 literals are dropped before `is_private_url` ever sees them.

That means every IPv6-shaped test this PR adds (`test_ssrf_blocks_embedded_and_alternate_ipv4_encodings`'s `[::ffff:127.0.0.1]` / `[::7f00:1]` / `[2002:7f00:1::]`, and all of `test_ssrf_blocks_nat64_embedded_ips`) only exercises `is_private_url` directly — through the real `extract_urls` → `build_link_context` pipeline that actually runs on inbound messages, a bracketed IPv6 SSRF payload is silently excluded from the link-context hint by the regex, not by the new blocklist logic. Net effect on *this* pipeline is unchanged (the URL doesn't reach the agent either way), but it means the elaborate IPv6/NAT64/6to4 hardening in `is_blocked_literal_ip` is unreachable dead code from `extract_urls`'s perspective — and, more importantly, any *legitimate* bracketed IPv6 URL a user pastes is silently dropped from link context too (a functional bug, not just a defense-in-depth gap).

This regex predates this PR and isn't part of the diff, so flagging rather than fixing: widening the character class to permit balanced `[...]` needs care not to reintroduce false-positive matches against Markdown link syntax (`[text](url)`), and should come with a new `extract_urls`-level test (not just `is_private_url`) asserting a bracketed IPv6 URL is actually extracted. Worth a maintainer call on whether to fix here or file as a tightly-scoped follow-up against `extract_urls` specifically.

---

houko (commented):
Third automated pass. The prior two passes' findings (doc-comment wrapping, NAT64 loopback/unspecified gap) are already fixed on this branch; the 6to4-asymmetry-with-web_fetch::canonical_ip finding is still open and accurately describes a real, low-severity inconsistency worth a maintainer call. One new finding below: extract_urls's URL regex excludes `[`/`]`, so bracketed IPv6-literal URLs never reach is_private_url through the actual extraction pipeline at all — meaning this PR's IPv6/NAT64/6to4 hardening is exercised only by direct unit tests, not by the real code path, and (separately) legitimate bracketed IPv6 URLs are silently dropped from link context. No other convention or correctness issues found: changelog fragment, commit format, and scoped-build usage all check out.

---

PR #6764, opened 2026-08-08
fix(api): enforce direct transport for pinned requests

houko:
## Summary
- use direct/no-proxy transport for webhook test delivery after fire-time DNS validation
- use direct/no-proxy transport for URL attachment downloads
- retain redirect rejection and pin the validated address with `resolve_to_addrs`

## Security impact
Prevents proxy-side DNS from bypassing the local check-to-connect pin for API-triggered untrusted URL requests. This PR is stacked on #6761 and should merge after it.

## Verification
- `cargo test -p librefang-api url_attachment_ssrf_tests --lib` (6 passed)
- `cargo test -p librefang-api --test webhooks_routes_integration` (3 passed)
- `cargo clippy -p librefang-api --lib --no-deps -- -D warnings -A clippy::nonminimal-bool`
- `git diff --check`


crates/librefang-api/src/routes/webhooks.rs
```diff
@@ -795,17 +795,16 @@ pub async fn test_webhook(
...
+    // A low-TTL record can flip between our validate call and reqwest's resolve call (DNS rebind), bypassing the SSRF check (#3701).
+    // `resolve_to_addrs` forces the connection to the validated address and skips reqwest's resolver for that hostname.
+    let mut builder = librefang_kernel::http_client::direct_client_builder()
         .timeout(std::time::Duration::from_secs(10))
         .redirect(reqwest::redirect::Policy::none());
     if let Some((ref host, addr)) = pinned_host {
-        builder = builder.resolve(host, addr);
+        builder = builder.resolve_to_addrs(host, &[addr]);
     }
     let client = match builder.build() {
```
houko, line 809: This closes the proxy-DNS-bypass gap for webhook test delivery and URL attachment downloads, but the same pinned-DNS pattern still runs through `proxied_client_builder()` (or a bare `reqwest::ClientBuilder::new()`, which also honors `HTTP_PROXY`/`HTTPS_PROXY` by default) in at least three other places, so the class of bug this PR describes is not closed end-to-end:

- `crates/librefang-runtime/src/web_fetch.rs` — `WebFetchEngine::pinned_client()` (the primary, non-legacy `web_fetch` tool) builds on `crate::http_client::proxied_client_builder()` then calls `resolution.pin_dns(builder)`. This is the exact pattern the sibling `web_legacy.rs::build_legacy_fetch_client` was switched to `direct_client_builder()` for in the base branch's `75f7410` commit — the *legacy* tool got the fix, the *current* one didn't.
- `crates/librefang-runtime/src/host_functions.rs` (~line 551, the WASM guest `NetFetch` host function) — same `proxied_client_builder()` + per-hop `.resolve()` loop.
- `crates/librefang-hands/src/hands_hub.rs` (~line 216, hand registry client) — uses `reqwest::ClientBuilder::new()` directly with DNS pins via `.resolve()`, but never calls `.no_proxy()`, so it still inherits env-var proxy detection.

Worth confirming whether these are deliberately out of scope for this PR/stack (different crates, so per this repo's own scope-deferral rule it's reasonable to ask before folding them in here) or whether they were simply missed. Given the stated security impact of this PR, I'd lean toward at least filing a tracked follow-up that covers all remaining pinned-DNS call sites rather than closing this out as "SSRF proxy bypass fixed."

---

houko (commented):
Reviewed the incremental diff on top of #6761. The direct/no-proxy switch + `resolve_to_addrs` change for webhook test delivery and URL attachment downloads is correct and matches the pattern already fixed for the legacy web-fetch tool. One line comment below on scope: the same proxy-bypass class remains open in a few sibling call sites this PR doesn't touch.

---

crates/librefang-api/src/routes/webhooks.rs
houko, line 809: Per CLAUDE.md's integration-test mandate for route/wiring changes: I don't see a test that actually exercises the proxy-bypass fix itself.

`url_attachment_ssrf_tests` (referenced in the PR's Verification section) only covers URL-validation rejection (loopback/IMDS/RFC1918/etc. literals) — those tests never reach the client-builder code path this PR changes.
I couldn't find any test hitting `POST /webhooks/{id}/test` at all in `crates/librefang-api/tests/`, so the switch from `proxied_client_builder()` to `direct_client_builder()` here and in `attachments.rs` has no regression coverage; a future refactor could silently reintroduce the proxied client and nothing would fail.

I'm flagging this as ambiguous rather than fixing it myself, since a real regression test for "does this client actually ignore `HTTP_PROXY`/`HTTPS_PROXY`" needs a local mock proxy (or a `no_proxy()`-call-site assertion at the very least), and that test-design choice seems worth a maintainer call rather than a mechanical patch.

---

houko:
Disclosure: this automated review run's own fix commit `5d940a2a` (doc-comment rewrap) was mistakenly authored under this review environment's default git identity, `Claude <noreply@anthropic.com>` — a violation of this repo's own no-AI-attribution policy, the same class of issue already flagged elsewhere in this repo's history. Root cause: the environment's global git config defaulted to that identity; this has been corrected mid-run (to the authenticated `houko` identity) so further commits from this routine don't repeat it. Recommend squashing/amending `5d940a2a` under a real author before merge.

---

crates/librefang-api/src/routes/agents/attachments.rs
```diff
@@ -332,15 +332,9 @@ pub fn inject_attachments_into_session(
...
-/// fetch — this rejects loopback, RFC 1918, link-local, IPv6 ULA, the
-/// cloud-metadata literals, and any hostname whose DNS resolves to one
-/// of those families. For domain URLs we then pin reqwest to the
-/// validated `SocketAddr` via `.resolve(host, addr)` so a DNS-rebind
-/// flip between validation and the eventual HTTP connect cannot reroute
-/// the fetch onto an internal IP. Mirrors the webhook fire-time pattern
-/// at `webhooks.rs:738-744` (issue #3701).
+/// SSRF defence: every URL is run through [`crate::webhook_store::validate_webhook_url_resolved`] before the fetch — this rejects loopback, RFC 1918, link-local, IPv6 ULA, the cloud-metadata literals, and any hostname whose DNS resolves to one of those families.
+/// For domain URLs we then pin reqwest to the validated `SocketAddr` on a direct/no-proxy client so a DNS-rebind flip between validation and the eventual HTTP connect cannot reroute the fetch onto an internal IP.
+/// Mirrors the webhook fire-time pattern at `webhooks.rs:738-744` (issue #3701).
```
houko: The current branch tip, commit `5d940a2` ("docs(api): rewrap remaining SSRF doc-comment to one sentence per line" — the commit that produced this exact doc-comment edit), is authored as `Claude <noreply@anthropic.com>` (`git log --format='%H %an <%ae>' -1 5d940a2`).

That's a direct hit on the project's "no AI / Claude attribution" rule (CLAUDE.md → Git Conventions / Commit & PR hygiene), and specifically the case the repo's `commit-msg` hook's author-identity check exists to reject before it reaches a PR ("rejects a commit whose author identity … resolves to Claude / Anthropic even when the message itself is clean"). It presumably slipped through because the hook wasn't active for whatever tool produced this commit.

I can't fix this myself — rewriting/amending an existing commit's author is outside what an automated reviewer is allowed to do on someone else's branch. Suggest the branch owner rebase/amend this commit with a human author identity before merge.

---

crates/librefang-api/src/routes/agents/attachments.rs
```diff
@@ -380,16 +374,14 @@ pub async fn resolve_url_attachments(
...
-        let mut builder = librefang_kernel::http_client::proxied_client_builder()
+        // Build a direct per-attachment client and pin DNS to the IP we just validated.
+        // Proxies are disabled because proxy-side DNS would bypass the local override.
+        // Without the pin, reqwest performs its own independent lookup before connecting — a low-TTL record can flip to a private IP between our validation and reqwest's resolver call (DNS rebind, #3701).
+        let mut builder = librefang_kernel::http_client::direct_client_builder()
             .timeout(std::time::Duration::from_secs(30))
             .redirect(reqwest::redirect::Policy::none());
         if let Some((ref host, addr)) = pinned_host {
-            builder = builder.resolve(host, addr);
+            builder = builder.resolve_to_addrs(host, &[addr]);
```
houko, line 384: This is the actual behavior change in this PR (`proxied_client_builder` → `direct_client_builder`, `.resolve` → `.resolve_to_addrs`), but `url_attachment_ssrf_tests` (the module cited in the PR description) only exercises URLs that get rejected by `validate_webhook_url_resolved` before any client is ever built — loopback, IMDS, IPv6 ULA, `localhost`, RFC 1918, and unsupported scheme all short-circuit at line ~364, above this code.

None of those tests reach the `direct_client_builder()` / `resolve_to_addrs` call on the success path, so this PR's own change (the no-proxy client + `resolve_to_addrs` pin) has no regression coverage — nothing would fail if this hunk were reverted back to `proxied_client_builder()`/`.resolve()`.

Per CLAUDE.md's "MANDATORY: Integration Testing", could you add a test that drives a valid public-looking URL through `resolve_url_attachments` and asserts the outbound client actually ignores `HTTP_PROXY`/`HTTPS_PROXY` (e.g. set the env var to a bogus/unreachable proxy in the test and confirm the request still succeeds/fails the same way it would without it, rather than erroring on the fake proxy)? That's the regression this PR is meant to close.

---

crates/librefang-api/src/routes/webhooks.rs
```diff
@@ -795,17 +795,16 @@ pub async fn test_webhook(
...
+    // Proxies are disabled because proxy-side DNS would bypass the local pin.
+    // Without the pin, reqwest does its own DNS lookup before connecting.
+    // A low-TTL record can flip between our validate call and reqwest's resolve call (DNS rebind), bypassing the SSRF check (#3701).
+    // `resolve_to_addrs` forces the connection to the validated address and skips reqwest's resolver for that hostname.
+    let mut builder = librefang_kernel::http_client::direct_client_builder()
         .timeout(std::time::Duration::from_secs(10))
         .redirect(reqwest::redirect::Policy::none());
     if let Some((ref host, addr)) = pinned_host {
-        builder = builder.resolve(host, addr);
+        builder = builder.resolve_to_addrs(host, &[addr]);
```
houko, line 807: Same gap as the attachments.rs half of this change: this is the `test_webhook` handler (`POST /api/webhooks/{id}/test`), and it looks like this route has zero test coverage anywhere in the suite, not just for this change.

The three tests in `webhooks_routes_integration.rs` that the PR description cites (`create_event_webhook_rejects_ssrf_urls`, `create_event_webhook_allows_public_url`, `update_event_webhook_rejects_ssrf_url`) all exercise webhook *subscription* create/update validation, never the `/test` fire-time delivery endpoint that this diff actually touches (`direct_client_builder`, `resolve_to_addrs`). Nothing in the suite would fail if this hunk reverted to `proxied_client_builder()`/`.resolve()`.

Given CLAUDE.md's "MANDATORY: Integration Testing" rule requires a `#[tokio::test]` against `TestServer` for any route/wiring change, could a test be added here (via `start_test_server`) that hits `POST /webhooks/{id}/test` against a webhook pointed at a real (test-harness) HTTP endpoint, with `HTTP_PROXY`/`HTTPS_PROXY` set to an unreachable address, and asserts the request still succeeds — proving the proxy is actually bypassed rather than just asserting the SSRF blocklist still works?

---

houko (commented):
Automated review pass against CLAUDE.md conventions and security-logic checklist. The direct/no-proxy transport change itself is correct (`direct_client_builder()` calls `.no_proxy()`, which reqwest confirms clears both explicit proxy config and system/env auto-detection — so `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` genuinely can't be used to bypass the DNS pin), and the guard is applied consistently at both pinned-request call sites in this diff. Two findings below are about test coverage for the changed code paths, and one is about a commit's author identity — see inline comments.

---

crates/librefang-api/src/routes/webhooks.rs
```diff
@@ -795,17 +795,16 @@ pub async fn test_webhook(
...
-    // (DNS rebind), bypassing the SSRF check (#3701). `.resolve(host, addr)`
-    // forces the connection to go to `addr` and skips reqwest's resolver
-    // for that hostname.
-    let mut builder = librefang_kernel::http_client::proxied_client_builder()
+    // Pin a direct client's DNS resolver to the address validated above.
+    // Proxies are disabled because proxy-side DNS would bypass the local pin.
+    // Without the pin, reqwest does its own DNS lookup before connecting.
+    // A low-TTL record can flip between our validate call and reqwest's resolve call (DNS rebind), bypassing the SSRF check (#3701).
+    // `resolve_to_addrs` forces the connection to the validated address and skips reqwest's resolver for that hostname.
+    let mut builder = librefang_kernel::http_client::direct_client_builder()
```
houko, line 803: Same gap as flagged on #6762: this swap to `direct_client_builder()` (here and in `attachments.rs`) has no regression test proving the proxy is actually bypassed. The PR's verification list re-runs pre-existing `url_attachment_ssrf_tests` and `webhooks_routes_integration` suites, which predate this change and assert SSRF/redirect/pin-address behavior, not proxy behavior. A test that points `HTTP_PROXY` at an unreachable address and confirms `test_webhook` / `resolve_url_attachments` still reach a local mock server would actually exercise the fix (mind the process-global env var + parallel test execution — needs a serialization guard or dedicated test binary).

---

houko (commented):
One test-coverage note inline (applies equally to `attachments.rs::resolve_url_attachments` and `webhooks.rs::test_webhook`). No other findings — the `resolve_to_addrs(host, &[addr])` swap correctly preserves the existing single-address pin semantics (this call site never carried a multi-address `SsrfResolution` list, unlike the runtime's legacy-fetch path in #6761).

---

crates/librefang-api/src/routes/webhooks.rs
houko, line 803: No regression test asserts that a *configured* proxy is actually bypassed here.
The PR description cites `webhooks_routes_integration` (3 tests) as verification, but those three (`create_event_webhook_rejects_ssrf_urls`, `create_event_webhook_allows_public_url`, `update_event_webhook_rejects_ssrf_url`) only exercise the write-time URL-validation gate — none of them set `HTTP_PROXY`/`HTTPS_PROXY` (or a config-level `ProxyConfig`) and then fire the webhook to confirm the request reached the pinned address directly instead of through the proxy.
Given this PR's whole point is closing a proxy-side-DNS bypass, a test that points a decoy proxy at a mock server and asserts it recorded zero requests while the pinned target did receive the request would be the thing that actually proves the fix (and would have caught a regression if `direct_client_builder()` regressed back to `proxied_client_builder()` later).

---

crates/librefang-api/src/routes/agents/attachments.rs
```diff
@@ -380,16 +374,14 @@ pub async fn resolve_url_attachments(
...
-        // Build a per-attachment client and pin DNS to the IP we just
-        // validated. Without the pin, reqwest performs its own
-        // independent lookup before connecting — a low-TTL record can
-        // flip to a private IP between our validation and reqwest's
-        // resolver call (DNS rebind, #3701).
-        let mut builder = librefang_kernel::http_client::proxied_client_builder()
+        // Build a direct per-attachment client and pin DNS to the IP we just validated.
+        // Proxies are disabled because proxy-side DNS would bypass the local override.
+        // Without the pin, reqwest performs its own independent lookup before connecting — a low-TTL record can flip to a private IP between our validation and reqwest's resolver call (DNS rebind, #3701).
+        let mut builder = librefang_kernel::http_client::direct_client_builder()
```
houko, line 380: Same gap as the webhooks.rs pinned client: `url_attachment_ssrf_tests` (the 6 tests cited in the PR description) only cover URL-validation rejection (loopback/IMDS/RFC1918/IPv6-ULA/localhost/unsupported-scheme literals) — none of them configure a proxy and assert it is *not* used for the actual fetch.
There is no test in this PR (or pre-existing) that proves a configured `HTTP_PROXY`/`HTTPS_PROXY` env var or `[proxy]` config is ignored by `resolve_url_attachments`'s client.
Recommend a regression test with a decoy mock-server-as-proxy plus the real pinned mock target, asserting the decoy sees zero requests.

---

houko (commented):
Reviewed the delta over the stacked base (`fix/legacy-web-fetch-ssrf`), not against `main`.
The two call sites this PR touches are correctly switched from `proxied_client_builder()` to `direct_client_builder()` (which calls `.no_proxy()`, disabling both config-level and env-var proxies), and the `.resolve()` loop → `resolve_to_addrs()` change is consistent with the sibling fix already on the base branch (`75f7410`).
Changelog fragment and doc-comment reflow both match repo convention.
Formatting is clean (`cargo fmt --check` on both touched files).

Two findings, both inline above (missing regression tests proving the proxy is actually bypassed).

One more thing worth flagging even though it's outside this PR's diff, since the PR title claims to "enforce direct transport for pinned requests" generally: at least two other pinned-client construction sites in the same codebase still build via `proxied_client_builder()` before pinning DNS, so they remain exposed to the same proxy-side-DNS bypass this PR closes for webhooks/attachments:
- `crates/librefang-runtime/src/web_fetch.rs:49` (`WebFetchEngine::pinned_client`, used by the primary `web_fetch` tool)
- `crates/librefang-runtime/src/host_functions.rs:551-556` (WASM guest `net_fetch` host function) — which additionally still uses the pre-`resolve_to_addrs` per-address `.resolve()` loop that `75f7410` already migrated away from at the other call sites (only the last resolved address ends up pinned).

If these are intentionally out of scope for this stacked PR, worth saying so explicitly and tracking as a follow-up; otherwise they're the same bug left open in two more places.

---

crates/librefang-api/src/routes/webhooks.rs
houko, line 1: Confirming and narrowing the earlier note about `librefang-hands/src/hands_hub.rs` in this thread: traced it end to end and this is a live, reachable bypass, not just a theoretical one.

`crates/librefang-api/src/routes/skills/hands.rs:1803-1838` takes a caller-supplied `registry_url`, runs it through `check_ssrf`, and pins the validated hostname/IPs onto `HandsHubClient::with_pinned_url` — same pattern this PR fixes for webhooks/attachments.
But `HandsHubClient::build()` (`hands_hub.rs:192-226`) constructs the client from a bare `reqwest::ClientBuilder::new()` and only ever calls `.resolve(host, addr)`; it never calls `.no_proxy()`.
A bare `ClientBuilder` still honors `HTTP_PROXY`/`HTTPS_PROXY` (env or this project's `[proxy]` config, since `init_proxy()` exports config values as env vars) by default, so an operator/attacker who controls the proxy (or the env in a shared-host scenario) can have proxy-side DNS re-resolve the pinned hostname and connect elsewhere — the exact DNS-rebind-via-proxy bypass this PR's title describes, just in a third crate neither this PR nor #6762 touches.

Not asking for a fix in this PR (different crate, different scope per this repo's own PR-splitting convention) — just confirming the earlier flag is accurate and pointing at the concrete reachable caller so it doesn't get lost as "maybe theoretical."

---

houko (commented):
Independent verification pass: traced the pinned-client code paths in both this PR's diff and the sibling files it doesn't touch. No new gap found in this PR's own scope (webhooks.rs + attachments.rs) — the swap to direct_client_builder()/resolve_to_addrs is applied correctly at both call sites. Left one comment confirming/narrowing an already-flagged cross-crate gap (librefang-hands) with the concrete reachable caller.

---

crates/librefang-api/src/routes/webhooks.rs
houko, line 803: No regression test proves the actual security property this PR adds (switching `proxied_client_builder()` → `direct_client_builder()` so `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` can't reroute the fire-time request around the DNS pin).
`direct_client_builder()` itself only has a "builds successfully" unit test in `librefang-http/src/lib.rs`, not a "does not route through a configured proxy" test.
The PR body cites `cargo test -p librefang-api --test webhooks_routes_integration (3 passed)` as verification, but that file (`webhooks_routes_integration.rs`) only covers `create_event_webhook` / `update_event_webhook` write-time URL rejection — it never calls `test_webhook` (the function this diff actually changes) and has no assertion about proxy usage at all.
Same gap for `resolve_url_attachments` in `attachments.rs` — `url_attachment_ssrf_tests` in `agents/mod.rs` only asserts which URLs get rejected, not that the accepted ones bypass a configured proxy.
Suggest a test that spins up a local TCP listener acting as a fake HTTP proxy, sets `HTTP_PROXY`/`HTTPS_PROXY` env vars for the test process, points `test_webhook` (and `resolve_url_attachments`) at a real (non-SSRF) target, and asserts the fake proxy listener never receives a connection — proving `.no_proxy()` actually wins over the env vars for this code path.
Without it, a future refactor that accidentally reverts to `proxied_client_builder()` here (or a change to `direct_client_builder()` that no longer calls `.no_proxy()`) would compile and pass CI while silently reopening the proxy-side-DNS SSRF bypass this PR is meant to close.

---

crates/librefang-api/src/routes/agents/attachments.rs
houko, line 380: Same gap as `webhooks.rs::test_webhook` (see comment there): no test exercises `resolve_url_attachments` with `HTTP_PROXY`/`HTTPS_PROXY` env vars set to prove the per-attachment client actually ignores them now that it's built from `direct_client_builder()` instead of `proxied_client_builder()`.
`url_attachment_ssrf_tests` in `agents/mod.rs` (the suite the PR cites as verification) only checks which URLs get rejected by `validate_webhook_url_resolved`, not the transport behavior of the client that fetches the accepted ones.
This is the same class of regression risk noted on the webhooks.rs comment: the fix is correct today (`direct_client_builder()` calls `.no_proxy()`, which does override env-based proxy detection), but nothing in the test suite would catch a future change that silently drops that guarantee.

---

houko (commented):
Automated review: the transport fix itself checks out — `direct_client_builder()` (in `librefang-http`, re-exported as `librefang_kernel::http_client`) calls `build_http_client(&ProxyConfig::default()).no_proxy()`, and `reqwest::ClientBuilder::no_proxy()` statically disables proxy support, overriding `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` env-var auto-detection as well as config-based proxies — so the "no proxy" property can't be defeated by env vars leaking through. Confirmed the only two pinned-client construction sites in the API crate (`webhooks.rs::test_webhook` and `attachments.rs::resolve_url_attachments`, both call sites of `validate_webhook_url_resolved`) are both covered by this diff — no missed sibling site. The changelog fragment (`changelog.d/security/6764-api-pinned-requests-no-proxy.md`) and the rewrapped doc/inline comments already conform to repo conventions, so no mechanical fixes were needed there.

Left two line comments on a real gap: neither touched call site has a regression test proving the no-proxy property holds under `HTTP_PROXY`/`HTTPS_PROXY` env vars, and the tests the PR cites as verification don't exercise this behavior at all.

---

crates/librefang-api/src/routes/agents/attachments.rs
houko, line 380: The `url_attachment_ssrf_tests` module (`routes/agents/mod.rs`) only exercises URLs that get rejected by `validate_webhook_url_resolved` before this line is ever reached — every case asserts `blocks.is_empty()` on a blocked URL. None of the 6 tests cited in the PR description as verification actually reach the `direct_client_builder()` / `resolve_to_addrs` code added here, so this PR's actual behavior change (proxy → no-proxy) has no regression coverage. Same gap applies to the mirrored change in `webhooks.rs`. Given this is explicitly an SSRF hardening fix, a test that sets `HTTP_PROXY`/`HTTPS_PROXY` to an unreachable address and confirms a request to a real pinned target still succeeds (i.e. the proxy was never consulted) would close the gap — happy to be told this was intentionally left out because of the env-var thread-safety caveat already documented in `librefang-http/src/lib.rs::init_proxy`.

---

crates/librefang-api/src/routes/webhooks.rs
houko, line 803: Same test-coverage gap as the mirrored change in `attachments.rs`: the existing webhook SSRF tests (`create_event_webhook_rejects_ssrf_urls`, `update_event_webhook_rejects_ssrf_url`) validate the registration-time blocklist, not `test_webhook`'s fire-time client construction, so nothing in this PR's own test suite would catch a regression back to `proxied_client_builder()` here.

Separate, broader concern (can't line-anchor since it's outside this PR's diff): the same "validate then DNS-pin" pattern this PR is hardening still uses the proxy-enabled `librefang_http::proxied_client_builder()` at `librefang-runtime/src/host_functions.rs:551` (WASM guest `net_fetch`), and `librefang-hands/src/hands_hub.rs` builds its registry client from a raw `reqwest::ClientBuilder::new()` with no `.no_proxy()` at all (`hands_hub.rs:199-224`). Both pin DNS via `.resolve()` after an SSRF check, exactly the pattern this PR fixes here — so both look like they retain the same proxy-side-DNS bypass this PR closes for webhooks/attachments. Worth a follow-up in the same spirit, since a corporate/env-configured HTTP(S)_PROXY would still let those two paths resolve the pinned hostname independently and reach an internal address.

---

houko (commented):
Verified the core mechanism: `direct_client_builder()` calls reqwest's `.no_proxy()`, which clears configured proxies and disables `auto_sys_proxy` (system/env-var detection) — confirmed against the vendored reqwest 0.13.4 source — so this correctly closes the proxy-side-DNS bypass for the two call sites touched here. Redirect-following stays disabled in both, so that vector is unchanged/still closed. `.resolve()` → `.resolve_to_addrs()` is a no-op (the former is literally implemented as the latter with a one-element slice). Left two comments: a test-coverage gap on this PR's own changed code path, and a residual concern about two sibling call sites outside this PR's diff that still use a proxy-enabled builder for the same validate-then-pin pattern.

---

PR #6765, opened 2026-08-08
fix(runtime): block Canvas URL mutation XSS

houko:
## Summary
- decode numeric and security-relevant named HTML entities before URL scheme checks
- reject browser-normalized ASCII control-character scheme smuggling while preserving safe URL spaces
- bound named-entity scans to a fixed window for linear behavior at the 512 KiB input limit
- remove active SVG data URLs from the image allowlist
- preserve safe HTTP links and raster image data URLs

## Security impact
Prevents Canvas HTML from reintroducing `javascript:`/`vbscript:` URLs through browser entity decoding or control-character normalization, blocks active SVG payloads embedded in data URLs, and avoids quadratic CPU use on entity-heavy input.

## Verification
- `cargo test -p librefang-runtime tool_runner::canvas::tests --lib` (7 passed)
- `git diff --check`

### DIFF
--- a/crates/librefang-runtime/src/tool_runner/canvas.rs
+++ b/crates/librefang-runtime/src/tool_runner/canvas.rs
@@ -150,10 +150,7 @@ fn decode_url_html_entities(value: &str) -> String {
             }
         }
 
-        // `rest` is non-empty here (loop guard: `index < value.len()`), but
-        // avoid `expect()`/`unwrap()` on data derived from agent-supplied
-        // HTML — fail closed by stopping the decode rather than asserting
-        // an invariant on untrusted input.
+        // `rest` is non-empty here (loop guard: `index < value.len()`), but avoid `expect()`/`unwrap()` on data derived from agent-supplied HTML — fail closed by stopping the decode rather than asserting an invariant on untrusted input.
         let Some(ch) = rest.chars().next() else {
             break;
         };
--- a/crates/librefang-runtime/src/tool_runner/canvas.rs
+++ b/crates/librefang-runtime/src/tool_runner/canvas.rs
@@ -124,11 +124,13 @@ fn decode_url_html_entities(value: &str) -> String {
                 }
             }
         }
-        if let Some(end) = rest
-            .strip_prefix('&')
-            .and_then(|entity| entity.find(';').map(|end| end + 1))
-            .filter(|end| *end <= 16)
-        {
+        let named_entity_end = rest.strip_prefix('&').and_then(|_| {
+            rest.as_bytes()
+                .iter()
+                .take(17)
+                .position(|byte| *byte == b';')
+        });
+        if let Some(end) = named_entity_end {
             let entity = &rest[1..end].to_ascii_lowercase();
             let replacement = match entity.as_str() {
                 "amp" => Some('&'),
@@ -158,10 +160,7 @@ fn decode_url_html_entities(value: &str) -> String {
 fn is_safe_url(url: &str) -> bool {
     let trimmed = url.trim().trim_matches(|c| c == '"' || c == '\'');
     let decoded = decode_url_html_entities(trimmed);
-    if decoded
-        .chars()
-        .any(|ch| ch.is_ascii_control() || ch.is_whitespace())
-    {
+    if decoded.chars().any(|ch| ch.is_ascii_control()) {
         return false;
     }
     let lower = decoded.to_lowercase();
@@ -508,10 +507,20 @@ mod tests {
 
     #[test]
     fn sanitizer_preserves_safe_http_and_raster_urls() {
-        let html = r#"<a href="https://example.com/?a=1&amp;b=2">link</a><img src="data:image/png;base64,AA==">"#;
+        let html = concat!(
+            r#"<a href="https://example.com/a b?a=1&amp;b=2">link</a>"#,
+            "<a href=\"https://example.com/a\u{2002}b\">unicode space</a>",
+            r#"<img src="data:image/png;base64,AA==">"#,
+        );
         let sanitized = sanitize_canvas_html(html, 512 * 1024).expect("sanitize");
         assert!(sanitized.contains("href="), "{sanitized}");
         assert!(sanitized.contains("src="), "{sanitized}");
+    }
+
+    #[test]
+    fn entity_decoder_handles_large_ampersand_input_linearly() {
+        let value = "&".repeat(512 * 1024);
+        assert_eq!(decode_url_html_entities(&value), value);
     }
 
     #[tokio::test]
--- a/crates/librefang-runtime/src/tool_runner/canvas.rs
+++ b/crates/librefang-runtime/src/tool_runner/canvas.rs
@@ -163,7 +163,7 @@ fn is_safe_url(url: &str) -> bool {
     if decoded.chars().any(|ch| ch.is_ascii_control()) {
         return false;
     }
-    let lower = decoded.to_lowercase();
+    let lower = decoded.trim_matches(|ch| ch <= '\u{20}').to_lowercase();
     if lower.starts_with("javascript:") || lower.starts_with("vbscript:") {
         return false;
     }
@@ -489,6 +489,8 @@ mod tests {
             r#"<a href="&#x6a;avascript:alert(1)">x</a>"#,
             r#"<a href="jav&#x61;script&#58;alert(1)">x</a>"#,
             r#"<a href="javascript&colon;alert(1)">x</a>"#,
+            r#"<a href="&#32;javascript:alert(1)">x</a>"#,
+            r#"<a href="&#x20;javascript:alert(1)">x</a>"#,
             r#"<a href="java&Tab;script:alert(1)">x</a>"#,
             "<a href=\"java\tscript:alert(1)\">x</a>",
             "<a href=\"java\nscript:alert(1)\">x</a>",
--- a/crates/librefang-runtime/src/tool_runner/canvas.rs
+++ b/crates/librefang-runtime/src/tool_runner/canvas.rs
@@ -92,9 +92,79 @@ fn is_void_tag(name: &str) -> bool {
     VOID_TAGS.contains(&name)
 }
 
+fn decode_url_html_entities(value: &str) -> String {
+    let mut decoded = String::with_capacity(value.len());
+    let mut index = 0;
+    while index < value.len() {
+        let rest = &value[index..];
+        if let Some(numeric) = rest.strip_prefix("&#") {
+            let (radix, digits) = if let Some(hex) = numeric
+                .strip_prefix('x')
+                .or_else(|| numeric.strip_prefix('X'))
+            {
+                (16, hex)
+            } else {
+                (10, numeric)
+            };
+            let digit_len = digits
+                .bytes()
+                .take_while(|byte| (*byte as char).is_digit(radix))
+                .count();
+            if digit_len > 0 {
+                let number = &digits[..digit_len];
+                if let Ok(codepoint) = u32::from_str_radix(number, radix) {
+                    if let Some(ch) = char::from_u32(codepoint) {
+                        decoded.push(ch);
+                        index += 2
+                            + usize::from(radix == 16)
+                            + digit_len
+                            + usize::from(digits[digit_len..].starts_with(';'));
+                        continue;
+                    }
+                }
+            }
+        }
+        if let Some(end) = rest
+            .strip_prefix('&')
+            .and_then(|entity| entity.find(';').map(|end| end + 1))
+            .filter(|end| *end <= 16)
+        {
+            let entity = &rest[1..end].to_ascii_lowercase();
+            let replacement = match entity.as_str() {
+                "amp" => Some('&'),
+                "apos" => Some('\''),
+                "colon" => Some(':'),
+                "gt" => Some('>'),
+                "lt" => Some('<'),
+                "newline" => Some('\n'),
+                "quot" => Some('"'),
+                "tab" => Some('\t'),
+                _ => None,
+            };
+            if let Some(ch) = replacement {
+                decoded.push(ch);
+                index += end + 1;
+                continue;
+            }
+        }
+
+        let ch = rest.chars().next().expect("index is inside value");
+        decoded.push(ch);
+        index += ch.len_utf8();
+    }
+    decoded
+}
+
 fn is_safe_url(url: &str) -> bool {
     let trimmed = url.trim().trim_matches(|c| c == '"' || c == '\'');
-    let lower = trimmed.to_lowercase();
+    let decoded = decode_url_html_entities(trimmed);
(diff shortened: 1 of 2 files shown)


crates/librefang-runtime/src/tool_runner/canvas.rs
```diff
@@ -92,9 +92,78 @@ fn is_void_tag(name: &str) -> bool {
...
+}
+
 fn is_safe_url(url: &str) -> bool {
     let trimmed = url.trim().trim_matches(|c| c == '"' || c == '\'');
-    let lower = trimmed.to_lowercase();
+    let decoded = decode_url_html_entities(trimmed);
+    if decoded.chars().any(|ch| ch.is_ascii_control()) {
+        return false;
+    }
+    let lower = decoded.trim_matches(|ch| ch <= '\u{20}').to_lowercase();
```
houko, line 169: Manual trace of `decode_url_html_entities` + `is_safe_url` against the standard mutation-XSS bypass set: decimal/hex numeric character references with and without a trailing `;` (`&#106;`, `&#106avascript:`, `&#x6a;`), named `&colon;`/`&Tab;`/`&NewLine;`, and literal TAB/LF/CR/NUL bytes.
I couldn't find a case where the decoded result still starts with `javascript:`/`vbscript:` without tripping the control/whitespace check, and the `data:image/svg+xml` removal closes the active-SVG data-URI path (both the `;base64,` and the previously-unhandled `,<svg…>` forms). No bypass found — flagging this as a "looked hard, didn't find one" rather than a confirmed-clean signoff, since URL/entity parsers are exactly the kind of code where a missed edge case bites later.

One behavior worth a second look, distinct from the above: this rejects the *entire* URL if the decoded value contains **any** whitespace character anywhere, not just adjacent to the scheme.
That's broader than the PR description's "preserving safe URL spaces" — a legitimate `href`/`src` with an unencoded literal space in the path or query (rare, but not impossible in agent-generated HTML) would now be dropped outright rather than just having the scheme-smuggling attempt blocked.
Browsers only strip ASCII tab/LF/CR from URLs (not general whitespace) and scheme characters are restricted to alnum/+/-/., so control/whitespace immediately around a `javascript:`/`vbscript:` scheme is the actual attack surface — scoping the reject to a prefix window (or just re-checking `lower.starts_with(...)` after stripping only tab/LF/CR/NUL, matching what browsers actually normalize) would tighten this to the real vulnerability without the broader collateral rejection.
Not blocking, just want the author/maintainer to confirm this is an intentional trade-off rather than an oversight.

---

houko (commented):
Reviewed the diff against CLAUDE.md's mechanical rules and did a manual security trace of the URL-sanitization change.

Mechanical checks — all clean, no fixes needed:
- Changelog fragment (`changelog.d/security/6765-canvas-url-mutation-xss.md`) is correctly placed, correctly formatted (no leading `- `, single sentence, ends `(#6765) (@houko)`), and passes `scripts/check-changelog-attribution.py --all-unreleased`.
- Commit messages are clean conventional commits (`fix(runtime): …`, `docs(changelog): …`) with no AI/Claude attribution.
- No new `KernelConfig` fields, no `HashMap`/`HashSet` touching LLM-prompt-reachable data, no dashboard files touched, and no API route/wiring change, so the integration-test mandate doesn't apply — the two new `#[test]` cases in `canvas.rs` are the right level of coverage for an internal sanitizer helper.

Security correctness — left one comment inline with the detail, but the short version: I traced the entity-decoding and control-character logic against the known `javascript:`/`vbscript:` mutation-XSS bypass patterns (numeric refs with/without trailing `;`, named entities, raw TAB/LF/CR/NUL) and the SVG data-URI removal, and didn't find a bypass.
Flagged one non-blocking behavioral question about the whitespace-anywhere-in-the-URL rejection being broader than the PR description implies — worth the author's confirmation, not a blocker.

No changes pushed — everything mechanical was already correct.

---

crates/librefang-runtime/src/tool_runner/canvas.rs
```diff
@@ -92,9 +92,84 @@ fn is_void_tag(name: &str) -> bool {
...
+        let Some(ch) = rest.chars().next() else {
+            break;
+        };
+        decoded.push(ch);
+        index += ch.len_utf8();
+    }
+    decoded
+}
+
 fn is_safe_url(url: &str) -> bool {
```
houko, line 163: Security-logic question, not a blocker for this PR's stated scope, but worth confirming the threat model explicitly since this PR is specifically about closing Canvas URL script-execution vectors.

`is_safe_url` is only ever invoked from `strip_dangerous_attrs` for attributes literally named `href` or `src` (see that function, further down in this file). That's sufficient today because the built-in `ALLOWED_TAGS` only contains `a` and `img` as URL-bearing tags.

But `sanitize_canvas_html_with_tags` accepts a fully operator-configurable `allowed_tags` override (`[canvas] allowed_tags` in config, via `is_allowed_tag`) that *replaces* the built-in list wholesale — it isn't intersected with a known-safe subset. If that list is ever extended (by an operator, or a future default) to include `iframe`, `object`, `form`, or `svg`, this same sanitizer would let `action`/`formaction`/`data`/`xlink:href` etc. straight through unchecked — e.g. `<form action="javascript:...">` or `<object data="javascript:...">` — because the attribute-name check is hardcoded to `href`/`src` only, silently reopening the exact bug class this PR fixes.

Should the attribute-name check be broadened (e.g. `action`, `formaction`, `data`, `poster`, `background`, `*href` for `xlink:href`), or should `allowed_tags` be clamped to a safe subset so this can't regress via config alone? Flagging for your judgment rather than deciding unilaterally, since either fix is a design call beyond this PR's stated scope.

---

crates/librefang-runtime/src/tool_runner/canvas.rs
```diff
@@ -92,9 +92,81 @@ fn is_void_tag(name: &str) -> bool {
...
     let trimmed = url.trim().trim_matches(|c| c == '"' || c == '\'');
-    let lower = trimmed.to_lowercase();
+    let decoded = decode_url_html_entities(trimmed);
+    if decoded.chars().any(|ch| ch.is_ascii_control()) {
+        return false;
+    }
+    let lower = decoded.trim_matches(|ch| ch <= '\u{20}').to_lowercase();
     if lower.starts_with("javascript:") || lower.starts_with("vbscript:") {
```
houko, line 170: `is_safe_url` is still a blocklist for the generic (non-`data:`) case: it rejects `javascript:`/`vbscript:` and then falls through to `true` at the bottom of the function (not shown in this diff, but it's the function's last statement) for every other scheme.
That means `file:`, `blob:`, `mocha:`, `livescript:`, or any future browser/embedder-specific pseudo-protocol nobody has thought to blocklist yet passes through untouched, on both `href` and `src`.
The `data:` branch already demonstrates the safer pattern (allowlist of `image/png|jpeg|gif|webp` prefixes) — worth considering whether the outer check should invert to an allowlist of schemes (`http:`, `https:`, `mailto:`, `tel:`, plus the existing `data:` sub-allowlist) rather than a blocklist of two known-bad ones, so a newly-discovered dangerous scheme doesn't require another emergency patch to this function.
This is a design call for the maintainers, not something to rewrite unilaterally in this PR — flagging it since the PR's own framing (entity-decoding bypasses of a blocklist) is the same class of gap one level up, at the scheme-name level rather than the encoding level.

Secondary, lower-confidence note: `strip_dangerous_attrs` only runs `is_safe_url` over `href`/`src`. Legacy URL-bearing attributes reachable through the current `ALLOWED_TAGS` list (e.g. `background` on `table`/`td`/`th`) aren't checked at all. I believe modern engines no longer execute `javascript:` through `background`, so this is likely dead weight today rather than a live bypass — but if the attribute allowlist ever grows (e.g. `video`/`audio`/`poster`, `formaction`), the same href/src-only assumption would need revisiting.

---

houko (commented):
Automated review pass on the entity-decoding / control-character-smuggling fix. Traced the new `decode_url_html_entities` + `is_safe_url` logic against the OWASP mutation-XSS cheat-sheet payload classes (numeric/hex/named entity obfuscation, no-semicolon legacy numeric refs, tab/newline/CR/NUL smuggling, mixed case, base64 data:text/html, active SVG data URLs) — the decode-then-classify approach holds up and correctly mirrors single-pass browser entity decoding (no double-decode false-negative, no quadratic blowup on adversarial `&`-heavy input). Added two missing regression payloads (mixed-case scheme, base64-encoded data:text/html) directly since they were simple test-only additions. Left one line comment on a real but non-mechanical completeness question: the outer scheme check is still a blocklist (`javascript:`/`vbscript:` only, default-allow everything else) rather than an allowlist, unlike the `data:` sub-check which already does allowlist correctly.

---

crates/librefang-runtime/src/tool_runner/canvas.rs
```diff
@@ -92,9 +92,81 @@ fn is_void_tag(name: &str) -> bool {
...
+        let Some(ch) = rest.chars().next() else {
+            break;
+        };
+        decoded.push(ch);
+        index += ch.len_utf8();
+    }
+    decoded
+}
+
 fn is_safe_url(url: &str) -> bool {
```
houko, line 163: `is_safe_url` is only invoked for the `href` and `src` attributes (see `strip_dangerous_attrs` a few lines below this function — the check is `(lower == "href" || lower == "src") && ...`).
Every other URL-bearing attribute on the built-in allowlist bypasses this whole function.
Concretely: `ALLOWED_TAGS` includes `table`, `td`, `th`, and the legacy `background` attribute (HTML4, still parsed by some engines on those elements) isn't `on*` and isn't `style`, so `<table background="javascript:alert(1)">` sanitizes through with the value untouched — none of the new entity/control-character hardening in this PR applies to it.
Modern browsers no longer execute `javascript:` for non-navigation resource fetches, so this is likely not exploitable today, but it means the set of attributes this function protects and the set of URL-capable attributes the allowlist actually exposes aren't the same set.
Given the PR title is specifically about closing off scheme-smuggling variants, worth either broadening the attribute check (e.g. any attribute value that itself looks like a URL) or leaving a comment here recording the href/src-only assumption so it doesn't quietly bit-rot if `background` (or a future allowed attribute) becomes exploitable again.

---

houko (commented):
Reviewed the entity-decoding / control-character / SVG-data-URI hardening in `is_safe_url`. The core fix looks correct for the mutation-XSS class it targets (numeric/named entity decoding matches HTML5 tokenizer semantics including the no-semicolon and hex-greedy-consumption edge cases exercised by the new tests, the control-character rejection correctly closes the tab/newline URL-normalization smuggling vector, and removing `data:image/svg+xml;` from the raster allowlist closes the active-SVG vector for both `href` and `src`). Left one comment on a residual scope question — non-blocking.

---

crates/librefang-runtime/src/tool_runner/canvas.rs
houko, line 170: Traced the numeric/named-entity decoder and the control-character rejection by hand against the new test cases and didn't find a bypass or a panic path — byte-boundary slicing on `;`/digit runs is always ASCII-safe (any byte used as a split point is itself a single-byte ASCII codepoint), and the linear-scan claim holds even for a pathological single digit run that overflows `u32` (the failed numeric parse backs off one char at a time, it doesn't re-run the `take_while` scan).

One structural point per the review brief though: `is_safe_url` is still a **scheme blacklist** for everything except `data:`, not an allowlist. It explicitly denies `javascript:`/`vbscript:` here and then falls through to `true` (line 182, outside this diff's hunks) for any other scheme — `file:`, `blob:`, `intent:`, and any scheme nobody has thought of yet all pass unchecked. Only the `data:` branch below is allowlist-shaped (raster MIME types only, correctly dropping `image/svg+xml` now).

Given the PR wants relative links / `mailto:` / `tel:` to keep working, a strict `http`/`https`-only allowlist is a bigger behavior change worth a separate discussion — but consider at least allowlisting the *href* case to `http:`/`https:`/`mailto:`/`tel:`/schemeless (relative), rather than "anything that isn't javascript/vbscript/bad-data". Not a mechanical fix, flagging for a decision rather than pushing a commit.

---

houko (commented):
Reviewed the entity-decoding/control-character fix in `crates/librefang-runtime/src/tool_runner/canvas.rs` (line comment below) — the decoder itself looks sound against the added test matrix and I couldn't construct a bypass or a panic (byte-boundary slicing only ever splits on ASCII bytes).

**Bigger concern, outside this diff's file scope, so flagging here rather than as an inline comment (GitHub won't let me anchor to an unchanged file):** `crates/librefang-api/src/ws.rs:1857-1868` has a second `canvas_present` rendering path that this PR doesn't touch:

```rust
StreamEvent::ToolUseEnd { name, input, .. } if name == "canvas_present" => {
    let html = input.get("html").and_then(|v| v.as_str()).unwrap_or("");
    ...
    Some(serde_json::json!({
        "type": "canvas",
        "canvas_id": uuid::Uuid::new_v4().to_string(),
        "html": html,
        "title": title,
    }))
}
```

This reads `html` straight from the tool call's **raw input** (the agent-supplied string as-typed) and streams it to WS clients under `{"type":"canvas", ...}` — documented at `ws.rs:13` as a stable `Server → Client` contract. It never goes through `sanitize_canvas_html_with_tags` / `is_safe_url`. That sanitizer only runs inside `tool_canvas_present` (`canvas.rs`) on the copy that gets written to `output/*.html`; the WS stream emits the *pre-sanitization* HTML on `ToolUseEnd`, before the tool body in `canvas.rs` (which does the sanitizing) even executes.

If any client (dashboard, mobile app, CLI TUI, or a third-party integration built against the documented WS contract) renders that `html` field as markup — iframe `srcdoc`, `dangerouslySetInnerHTML`, or similar — every `javascript:`/`vbscript:` URL and `data:image/svg+xml` payload this PR blocks on the file-write path sails through unmodified on the live-stream path. I didn't find a current in-repo dashboard consumer of this specific `"type":"canvas"` WS message (the dashboard's own `/canvas` route is the unrelated workflow-graph editor), but the message is a documented protocol surface, so an external consumer relying on it would be exposed regardless.

Given the PR title is specifically "block Canvas URL mutation XSS," I think this needs at least a decision: either route `ws.rs`'s `html` through the same `sanitize_canvas_html_with_tags` call (with the same `max_html_bytes`/`allowed_tags` config) before it's put on the wire, or explicitly document that the WS `"canvas"` event is a raw/untrusted preview and must not be rendered as HTML by any consumer. Happy to send a follow-up PR for the former once there's agreement on approach — didn't want to bundle a `librefang-api` change into this `librefang-runtime`-scoped security fix without a green light.

Everything else checked out: the `changelog.d/security/6765-canvas-url-mutation-xss.md` fragment matches the required format (single sentence, ends `(#6765) (@houko)`, no AI attribution), and dropping `data:image/svg+xml;` from the raster allowlist correctly closes the active-SVG vector.

---

PR #6766, opened 2026-08-08
fix(runtime): escape OAuth callback errors

houko:
## Summary
- HTML-escape OAuth callback error codes and descriptions before reflecting them into the browser response
- cover tag, event-handler, ampersand, quote, and apostrophe injection payloads
- preserve ordinary OAuth error text

## Security impact
Prevents a crafted localhost ChatGPT OAuth callback URL from injecting executable markup into the HTML error page.

## Verification
- RED: `cargo test -p librefang-runtime test_error_html_escapes_untrusted_callback_values --lib -- --nocapture`
- GREEN: `cargo test -p librefang-runtime chatgpt_oauth::tests --lib` (21 passed)
- `git diff --check`

### DIFF
--- a/crates/librefang-runtime/src/chatgpt_oauth.rs
+++ b/crates/librefang-runtime/src/chatgpt_oauth.rs
@@ -762,6 +762,8 @@ fn success_html() -> String {
 
 /// Error page shown when OAuth returns an error.
 fn error_html(error: &str, description: &str) -> String {
+    let error = escape_html_text(error);
+    let description = escape_html_text(description);
     format!(
         r#"<!DOCTYPE html>
 <html lang="en">
@@ -786,6 +788,21 @@ fn error_html(error: &str, description: &str) -> String {
 </body>
 </html>"#
     )
+}
+
+fn escape_html_text(value: &str) -> String {
+    let mut escaped = String::with_capacity(value.len());
+    for ch in value.chars() {
+        match ch {
+            '&' => escaped.push_str("&amp;"),
+            '<' => escaped.push_str("&lt;"),
+            '>' => escaped.push_str("&gt;"),
+            '"' => escaped.push_str("&quot;"),
+            '\'' => escaped.push_str("&#39;"),
+            _ => escaped.push(ch),
+        }
+    }
+    escaped
 }
 
 #[cfg(test)]
@@ -969,5 +986,20 @@ mod tests {
         let html = error_html("access_denied", "User cancelled");
         assert!(html.contains("access_denied"));
         assert!(html.contains("User cancelled"));
+    }
+
+    #[test]
+    fn test_error_html_escapes_untrusted_callback_values() {
+        let html = error_html(
+            r#"<script>alert("error")</script>&'"#,
+            r#"<img src=x onerror='alert("description")'>&"#,
+        );
+
+        assert!(!html.contains("<script>"), "{html}");
+        assert!(!html.contains("<img"), "{html}");
+        assert!(!html.contains("onerror='"), "{html}");
+        assert!(html.contains("&lt;script&gt;alert(&quot;error&quot;)&lt;/script&gt;&amp;&#39;"));
+        assert!(html
+            .contains("&lt;img src=x onerror=&#39;alert(&quot;description&quot;)&#39;&gt;&amp;"));
     }
 }
--- a/changelog.d/security/6766-chatgpt-oauth-error-xss.md
+++ b/changelog.d/security/6766-chatgpt-oauth-error-xss.md
@@ -1,1 +1,1 @@
-Escape untrusted ChatGPT OAuth callback errors before rendering the browser response. (@houko)
+Escape untrusted ChatGPT OAuth callback errors before rendering the browser response. (#6766) (@houko)
--- a/changelog.d/security/6766-chatgpt-oauth-error-xss.md
+++ b/changelog.d/security/6766-chatgpt-oauth-error-xss.md
@@ -1,0 +1,1 @@
+Escape untrusted ChatGPT OAuth callback errors before rendering the browser response. (@houko)


crates/librefang-runtime/src/chatgpt_oauth.rs
houko, line 1: Non-blocking: this duplicates an existing escaper in the same crate — `tool_runner/canvas.rs::escape_html` (same five-char set, only the apostrophe entity differs: `&#x27;` there vs `&#39;` here, both valid).\nSince both callers only ever place the escaped value into HTML text-node content (never into an attribute or script context), the fix itself is correct and sufficient as written — flagging this only as a follow-up-worthy dedup opportunity, not something to block on. Leaving as a comment rather than fixing, since it touches escaping code directly.

---

houko (commented):
Automated review of the diff (origin/main...origin/fix/chatgpt-oauth-xss).

Security check: the fix escapes `error` / `error_description` with a full character-map (not a blocklist) covering `& < > " '`, and both values are only ever interpolated into HTML text-node content in `error_html` (inside `<p>` tags, never into an attribute, `<script>`, or URL context), so this escape set is sufficient for its actual usage — no missed injection context. The added test uses a real payload (`<script>alert(...)</script>`, an `onerror='...'` handler, `&`/`'`/`"`) and asserts both that the dangerous substrings are absent and that the fully-escaped string is present.

Verified: `git diff origin/main...origin/fix/chatgpt-oauth-xss` reviewed in full (33 lines, 2 files); `cargo fmt -p librefang-runtime -- --check` clean; changelog fragment `changelog.d/security/6766-chatgpt-oauth-error-xss.md` correctly placed/formatted with `(#6766) (@houko)`; no other unescaped call sites of `error_html`/`success_html` in the file; sibling `copilot_oauth.rs` uses a device-code flow (no browser HTML rendering), so it isn't exposed to the same class of bug.

One non-blocking note left inline (escaper duplication vs. `tool_runner/canvas.rs`) — not a correctness issue, so not fixed in this conservative pass.

No blocking findings.

---

crates/librefang-runtime/src/chatgpt_oauth.rs
```diff
@@ -762,6 +762,8 @@ fn success_html() -> String {
 
 /// Error page shown when OAuth returns an error.
 fn error_html(error: &str, description: &str) -> String {
+    let error = escape_html_text(error);
+    let description = escape_html_text(description);
```
houko, line 766: Verified this escaping is correct and complete for the actual injection points: `error`/`description` are only ever placed inside HTML text-node context (`<p>...{error}...</p>`), never inside an attribute, URL, or inline script, so escaping `& < > " '` is the right context-appropriate function and fully closes the two reflected values here.

Separate observation, not a request to change this PR: `crates/librefang-extensions/src/oauth.rs` (generic OAuth2/PKCE callback handler, lines ~300-309) has the identical pattern -- `axum::response::Html(format!("<h1>Error</h1><p>OAuth error: {error}</p>"))` reflecting the provider-supplied `error` query param completely unescaped. That looks like the code path PR #6767 ("fix(extensions): escape OAuth callback errors") targets. Worth having a reviewer confirm #6767 covers that exact call site, and possibly worth extracting a shared `escape_html_text` helper (e.g. into `librefang-http` or `librefang-types`) so both crates don't carry independent copies of the same security-sensitive escaping logic going forward. Not asking for that here since it'd cross into #6767's crate/scope.

---

crates/librefang-runtime/src/chatgpt_oauth.rs
```diff
@@ -970,4 +987,19 @@ mod tests {
...
+            r#"<img src=x onerror='alert("description")'>&"#,
+        );
+
+        assert!(!html.contains("<script>"), "{html}");
+        assert!(!html.contains("<img"), "{html}");
+        assert!(!html.contains("onerror='"), "{html}");
+        assert!(html.contains("&lt;script&gt;alert(&quot;error&quot;)&lt;/script&gt;&amp;&#39;"));
+        assert!(html
+            .contains("&lt;img src=x onerror=&#39;alert(&quot;description&quot;)&#39;&gt;&amp;"));
+    }
```
houko, line 1004: Minor test-depth suggestion (non-blocking): this test exercises `error_html()` directly, which is the right unit-level check, but the actual wire response is assembled in `handle_oauth_callback` (`error_html.len()` for `Content-Length`, then the raw bytes written to the socket). Per this crate's own testing note ("New runtime work SHOULD include at least one `#[tokio::test]` exercising the new path"), a small `#[tokio::test]` that drives a real `TcpStream` through `handle_oauth_callback` with a malicious `error`/`error_description` query string and asserts the bytes written to the socket contain the escaped form (and that `Content-Length` matches the escaped body's byte length) would close the gap between "the escaping function is correct in isolation" and "the bytes actually sent to the browser are correct." Optional given the existing coverage is already solid.

---

houko (commented):
Reviewed the escaping fix against the actual injection points (both `error` and `description` land only in HTML text-node context in `<p>` tags, never in an attribute/URL/script context) -- the fix is correct and complete for this file. Two non-blocking notes left inline: a likely-overlapping unescaped reflection in `librefang-extensions/src/oauth.rs` that PR #6767 may already target, and an optional suggestion to add socket-level test coverage of the full response framing.

---

crates/librefang-runtime/src/chatgpt_oauth.rs
houko, line 1: Reviewed the escaping fix itself and it holds up:

- `escape_html_text` runs at render time, in a single char-by-char pass over the raw value, so there's no double-encoding hazard from arm ordering (`&` is matched against the *original* string, not a partially-escaped one).
- Both attacker-influenced values that actually reach the response — `error` and `error_description` from the callback query string (`handle_oauth_callback`, lines 613-617) — are escaped consistently before interpolation. `code` and `state` are never reflected into HTML (the state-mismatch page uses a static string), so there's no missed sink.
- `Content-Type: text/html; charset=utf-8` on the response (line 619) matches the `<meta charset="utf-8">` in the template (line 771), so there's no charset mismatch a browser could use to sniff into a different encoding (e.g. UTF-7) and defeat the entity escaping.
- Body content is only ever appended after the header block is already terminated by `\r\n\r\n`, and `Content-Length` is computed from the actual (escaped) string length, so there's no response-splitting angle even though the escaping doesn't touch `\r`/`\n`.

One adjacent, non-blocking observation (not introduced by this PR, and it does not defeat the new escaping): `urldecode()` at line 704 decodes `%XX` via `result.push(byte as char)` instead of proper UTF-8 decoding, so multi-byte UTF-8 in `error`/`error_description` gets mangled into separate Latin-1-range codepoints before it ever reaches `escape_html_text`. It's harmless for the XSS fix specifically — the metacharacters this PR cares about (`< > & " '`) are all single-byte ASCII and decode correctly either way — but non-ASCII OAuth error text (e.g. a localized `error_description`) will render as mojibake. Worth a follow-up if anyone hits it in practice; not a reason to hold this fix.

Changelog fragment (`changelog.d/security/6766-chatgpt-oauth-error-xss.md`) looks correctly formatted: one sentence, ends `(#6766) (@houko)`.

---

houko (commented):
Reviewed the OAuth callback XSS fix in `chatgpt_oauth.rs`. The escaping is correct, render-time, and covers every attacker-influenced value that reaches the response; charset is pinned consistently (header + meta tag) so there's no sniffing angle. Left one non-blocking, out-of-scope observation on `urldecode()` inline — it doesn't affect this fix. No blocking issues found; not requesting changes.

---

PR #6767, opened 2026-08-08
fix(extensions): escape OAuth callback errors

houko:
## Summary
- isolate rendering of provider OAuth callback errors
- HTML-escape untrusted provider error values before returning the localhost browser page
- keep the internal oneshot diagnostic unchanged

## Security impact
Prevents a valid OAuth callback state combined with a crafted provider error value from injecting executable markup into the callback tab.

## Verification
- RED: `cargo test -p librefang-extensions oauth_error_html_escapes_untrusted_callback_value --lib -- --nocapture`
- GREEN: `cargo test -p librefang-extensions oauth::tests --lib` (13 passed)
- `cargo clippy -p librefang-extensions --lib --no-deps -- -D warnings`
- `git diff --check`

### DIFF
--- a/changelog.d/security/6767-extensions-oauth-error-xss.md
+++ b/changelog.d/security/6767-extensions-oauth-error-xss.md
@@ -1,0 +1,1 @@
+Escape untrusted provider OAuth callback errors before rendering the browser response. (@houko)
--- a/changelog.d/security/6767-extensions-oauth-error-xss.md
+++ b/changelog.d/security/6767-extensions-oauth-error-xss.md
@@ -1,1 +1,1 @@
-Escape untrusted provider OAuth callback errors before rendering the browser response. (@houko)
+Escape untrusted provider OAuth callback errors before rendering the browser response. (#6767) (@houko)
--- a/crates/librefang-extensions/src/oauth.rs
+++ b/crates/librefang-extensions/src/oauth.rs
@@ -304,9 +304,7 @@ pub async fn run_pkce_flow(oauth: &OAuthTemplate, client_id: &str) -> ExtensionR
                         if let Some(tx) = code_tx.lock().await.take() {
                             let _ = tx.send(Err(format!("OAuth error: {error}")));
                         }
-                        return axum::response::Html(format!(
-                            "<h1>Error</h1><p>OAuth error: {error}</p>"
-                        ));
+                        return axum::response::Html(oauth_error_html(error));
                     }
                     if let Some(ref code) = query.code {
                         // #3791: only the first valid callback wins; subsequent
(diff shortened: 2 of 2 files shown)


changelog.d/security/6767-extensions-oauth-error-xss.md
```diff
@@ -0,0 +1 @@
+Escape untrusted provider OAuth callback errors before rendering the browser response. (#6767) (@houko)
```
houko, line 1: Commit ca7283a on this branch ("docs(changelog): reference PR number in OAuth XSS fragment", the commit that last touched this fragment) has its author identity recorded as `Claude <noreply@anthropic.com>`.
That's exactly the pattern the repo's `commit-msg` hook is meant to reject per CLAUDE.md's "No AI / Claude attribution" rule — the hook checks `git var GIT_AUTHOR_IDENT` and blocks a commit whose author resolves to Claude/Anthropic even when the message body itself is clean, which this one is.
It's already on the shared branch, so rewriting it needs a history rewrite / force-push, which is outside what an automated review pass should do unilaterally.
Please re-author (or squash) that commit under a human identity before merge.

---

crates/librefang-extensions/src/oauth.rs
```diff
@@ -304,9 +304,7 @@ pub async fn run_pkce_flow(oauth: &OAuthTemplate, client_id: &str) -> ExtensionR
                         if let Some(tx) = code_tx.lock().await.take() {
                             let _ = tx.send(Err(format!("OAuth error: {error}")));
                         }
-                        return axum::response::Html(format!(
-                            "<h1>Error</h1><p>OAuth error: {error}</p>"
-                        ));
+                        return axum::response::Html(oauth_error_html(error));
```
houko: The HTML path here is now correctly escaped for the text-node context it renders into (`&amp;`/`&lt;`/`&gt;`/`&quot;`/`&#39;` on `error`, checked against `<`,`>`,`&`,`'`,`"` in the added test) — I didn't find another reflected value in this file's HTML responses (`code`/`state` are never echoed back).

One residual concern, called out but deliberately left alone per the PR description ("keep the internal oneshot diagnostic unchanged"): line 305's `tx.send(Err(format!("OAuth error: {error}")))` still carries the raw, unescaped provider `error` string into `ExtensionResult<OAuthTokens>` as it propagates out of `run_pkce_flow`. Today nothing in-tree calls `run_pkce_flow` yet, so there's no live sink to check — but whichever caller wires this up (dashboard toast, CLI output rendered as HTML, log viewer, etc.) needs to escape it independently for whatever context it lands in, since this crate no longer has the only copy of that string. Worth a one-line doc comment on `run_pkce_flow` (or `ExtensionError::OAuth`) flagging that the error payload is untrusted/unescaped, so a future caller doesn't assume it's already safe to interpolate into HTML the way the old code here did.

---

houko (commented):
Automated review pass. Two comments below — one on the changelog fragment (commit authorship) and one on the OAuth callback handler (residual escaping note). No code changes pushed: the escaping fix itself looks correct and complete for the reflected value it addresses.

---

houko:
Reviewed the escaping fix in `crates/librefang-extensions/src/oauth.rs`.
All interpolation points into the callback HTML are accounted for: `error` is the only provider/user-controlled value ever reflected into the response body, and it now goes through `escape_html_text` (handles `&`, `<`, `>`, `"`, `'`).
`state`, `client_id`, and the provider auth URL are used only for HMAC/state verification — they are never echoed into the HTML, so there's no second reflected-value site to cover.
The four other `Html(...)` responses in the callback handler (invalid-state x2, success, gone, no-code) are all static strings with no interpolation.

One gap: the only regression test (`oauth_error_html_escapes_untrusted_callback_value`) exercised the pure `oauth_error_html` formatting helper directly, not the actual `/callback` route wiring — so a future refactor that forgot to call the escaping helper at the call site wouldn't have been caught.
Pushed `86eb275`, which extracts the route into `build_callback_router` / `handle_oauth_callback` (behavior-preserving) and adds `callback_route_escapes_xss_payload_in_error_query_param`, a `#[tokio::test]` that binds a real listener, sends a real HTTP request with a `<script>` payload in `error` plus a valid signed `state`, and asserts the response body over the wire is escaped.

Verified locally:
- `cargo test -p librefang-extensions --lib oauth` — 14 passed
- `cargo clippy -p librefang-extensions --all-targets -- -D warnings` — clean
- `cargo fmt -p librefang-extensions --check` — clean

No other findings; changelog fragment format/attribution already correct.

---

PR #6768, opened 2026-08-08
fix(kernel): validate cron update scripts

houko:
## Summary
- apply the daemon-home pre_script allowlist to cron updates as well as creates
- preserve candidate-validate-swap atomicity when path validation fails
- preserve existing update capacity semantics; add/move limit atomicity remains a separate concurrency item

## Security impact
Prevents cron PUT updates from installing a pre-processing executable outside `<home>/scripts/`, closing a validation gap that did not exist on job creation.

## Verification
- RED: `cargo test -p librefang-kernel update_job_rejects_pre_script_outside_home_scripts --lib -- --nocapture`
- GREEN: `cargo test -p librefang-kernel update_job_ --lib` (12 passed)
- `git diff --check`

### DIFF
--- a/changelog.d/security/6768-cron-update-script-validation.md
+++ b/changelog.d/security/6768-cron-update-script-validation.md
@@ -1,0 +1,1 @@
+Apply the cron pre-processing script allowlist consistently to job updates. (@houko)
--- a/changelog.d/security/6768-cron-update-script-validation.md
+++ b/changelog.d/security/6768-cron-update-script-validation.md
@@ -1,1 +1,1 @@
-Apply the cron pre-processing script allowlist consistently to job updates. (@houko)
+Apply the cron pre-processing script allowlist consistently to job updates. (#6768) (@houko)
--- a/crates/librefang-kernel/src/cron.rs
+++ b/crates/librefang-kernel/src/cron.rs
@@ -385,18 +385,9 @@ impl CronScheduler {
         id: CronJobId,
         updates: &serde_json::Value,
     ) -> LibreFangResult<CronJob> {
-        // Candidate-validate-swap: clone the current job, apply the
-        // partial updates onto the candidate, run the same home-aware
-        // validation that `add_job` runs, and only after that passes do we swap the
-        // candidate into place under the shard lock. This generalises
-        // the #4732 bypass closure from "delivery / delivery_targets
-        // re-validated on update" to "the entire CronJob shape is
-        // re-validated on update" — name length, schedule cron-expr
-        // syntax, every CronAction shape, and the SSRF / path checks all
-        // gate update the same way they gate add. A pre-#4739 PUT
-        // carrying e.g. an empty `name` plus a valid `delivery` was
-        // accepted; now the same payload is rejected before any field
-        // hits live state.
+        // Candidate-validate-swap: clone the current job, apply the partial updates onto the candidate, run the same home-aware validation that `add_job` runs, and only after that passes do we swap the candidate into place under the shard lock.
+        // This generalises the #4732 bypass closure from "delivery / delivery_targets re-validated on update" to "the entire CronJob shape is re-validated on update" — name length, schedule cron-expr syntax, every CronAction shape, and the SSRF / path checks all gate update the same way they gate add.
+        // A pre-#4739 PUT carrying e.g. an empty `name` plus a valid `delivery` was accepted; now the same payload is rejected before any field hits live state.
         //
         // Atomicity: `meta.job` is replaced once with a fully validated
         // candidate, so an `Err` at any step leaves the live row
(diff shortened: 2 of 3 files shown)


crates/librefang-kernel/src/cron.rs
```diff
@@ -448,17 +448,20 @@ impl CronScheduler {
...
+        // update does not count itself, while a cross-agent move still
+        // respects MAX_JOBS_PER_AGENT.
+        let existing_count = self
+            .jobs
+            .iter()
+            .filter(|entry| *entry.key() != id && entry.value().job.agent_id == candidate.agent_id)
+            .count();
         candidate
-            .validate(0)
+            .validate_with_home(existing_count, Some(&self.home_dir))
```
houko: Possible TOCTOU on the new per-agent cap enforcement for cross-agent moves: `existing_count` is computed by iterating the `DashMap` here, and the candidate is only written back later via `self.jobs.get_mut(&id)` (line ~495). Nothing holds a lock across that span, so two concurrent `update_job` calls that each move a different job onto the same near-full target agent can both snapshot `existing_count = MAX_JOBS_PER_AGENT - 1`, both pass `validate_with_home`, and both commit — leaving the agent over cap. `add_job` has the same count-then-insert shape, but before this PR `update_job` never enforced the cap at all, so this is new racy surface, not just an inherited pattern. Worth a comment or a follow-up on whether the count+swap should happen under a single lock (e.g. by sharding jobs per-agent or taking an explicit mutex around the count+validate+insert), especially since this is a security-motivated PR closing a validation gap.

---

houko: The schedule-wedge probe further down is deliberately gated on `if schedule_updated` specifically so an already-wedged row can still be fixed via update (see the comment a few lines below). This new cap check has no equivalent escape hatch — `validate_with_home` rejects with "agent already has N jobs (max 50)" whenever `existing_count >= MAX_JOBS_PER_AGENT`, unconditionally, even for updates that don't touch `agent_id` at all (rename, fix a bad `pre_script` path, toggle `enabled`, etc.).

If an agent ever ends up at or above the cap — via the race above, via data that predates the cap, or after a future cap change — every subsequent `update_job` call for *any* of that agent's existing jobs starts failing, including the one call an operator would use to fix the overage (e.g. correcting a wedged cron expression). Is that intentional, or should the cap check only fire when `agent_id` is actually changing (mirroring `schedule_updated`'s pattern)?

---

changelog.d/security/6768-cron-update-script-validation.md
```diff
@@ -0,0 +1 @@
+Apply the cron pre-processing script allowlist consistently to job updates. (@houko)
```
houko: Per CLAUDE.md's changelog convention, the fragment body must end `(#PR) (@your-github-login)`. This one ends `(@houko)` but is missing the `(#6768)` reference. Per the generation rule in `changelog.d/README.md`, a PR whose number appears in a curated bullet's `(#N)` group is skipped by the auto-generated line — without `(#6768)` here, this PR gets no line in the eventual release notes at all. Please add the PR number: `... consistently to job updates. (#6768) (@houko)`.

---

crates/librefang-kernel/src/cron.rs
```diff
@@ -448,17 +448,20 @@ impl CronScheduler {
...
-        // separate follow-up issue (#4732 followup).
+        // Run the same shape, SSRF, and pre_script path validation as
+        // `add_job`. Count the candidate owner's other jobs so an in-place
+        // update does not count itself, while a cross-agent move still
+        // respects MAX_JOBS_PER_AGENT.
+        let existing_count = self
+            .jobs
+            .iter()
+            .filter(|entry| *entry.key() != id && entry.value().job.agent_id == candidate.agent_id)
+            .count();
```
houko: Minor: this count-and-filter block duplicates `add_job`'s near-identical per-agent count (lines ~305-311), differing only in the self-exclusion clause (`*entry.key() != id`). Consider factoring both into one helper, e.g. `fn jobs_for_agent(&self, agent: AgentId, exclude: Option<CronJobId>) -> usize`, so a future change to counting semantics (e.g. excluding disabled jobs) can't be applied to one call site and forgotten on the other.

---

houko (commented):
Automated daily review pass. The core fix (applying `validate_with_home` + the pre_script home-dir allowlist to `update_job`, with self-exclusion from the per-agent count) is sound and covered by the three new tests. Left 4 inline comments: (1) a possible TOCTOU on the new cross-agent-move cap check since the count and the later write aren't under one lock, (2) the cap check has no `schedule_updated`-style escape hatch, so an agent already at/over cap could get locked out of fixing any of its own jobs via update, (3) the changelog fragment is missing the `(#6768)` PR reference required by CLAUDE.md's convention, and (4) a minor DRY note on the duplicated per-agent counting logic vs. `add_job`. No mechanical fixes applied — all findings touch validation/control-flow or are explicitly author-owned (changelog wording) per review policy.

---

crates/librefang-kernel/src/cron.rs
```diff
@@ -448,17 +448,15 @@ impl CronScheduler {
...
-        // an update that doesn't change `agent_id`. Cross-agent moves
-        // are NOT capacity-checked here today; tracking under a
-        // separate follow-up issue (#4732 followup).
+        // Run the same shape, SSRF, and pre_script path validation as
+        // `add_job`. Capacity checks remain unchanged for updates; making
+        // add/move limit enforcement atomic requires a shared mutation lock
+        // and is tracked separately from this path-allowlist fix.
         candidate
-            .validate(0)
+            .validate_with_home(0, Some(&self.home_dir))
```
houko, line 445: Good fix for `update_job`, but there looks to be a third mutation path with the same gap: `CronScheduler::load()` (boot-time re-validation of persisted jobs, around line 226) still calls `meta.job.validate(0)` instead of `meta.job.validate_with_home(0, Some(&self.home_dir))`.

`validate(existing_count)` forwards to `validate_with_home(existing_count, None)`, and per that function's own doc comment the `<home_dir>/scripts/` path allowlist on `pre_script.argv[0]` is skipped entirely when `home_dir` is `None` — only the env denylist still runs. So a persisted job whose `pre_script.argv[0]` points outside the allowlist (a hand-edited `cron_jobs.json`, or a job written by a pre-#6768 daemon) sails through the boot-time re-validation in `load()` unchanged and will still fire.

`self.home_dir` is already a field on `CronScheduler` and is in scope in `load()` (used elsewhere in the same impl block), so this looks like a one-line swap mirroring what this PR just did for `update_job`: `meta.job.validate_with_home(0, Some(&self.home_dir))`.

One more wrinkle worth a decision either way: `load()`'s re-validation is currently gated behind `if !matches!(meta.job.schedule, CronSchedule::At { .. })`, so `At`-scheduled (one-shot) persisted jobs skip this check entirely regardless of which `validate*` variant is called — a one-shot job with a malicious `pre_script` outside the allowlist would still fire once on reload. Might be worth splitting the pre_script/shape check out from the schedule-specific check so it always runs, but that's a slightly bigger change than the one-line swap above — flagging for your judgment rather than assuming the scope.

---

houko (commented):
Automated review pass on the cron update-script validation fix. The `update_job` fix and its new tests look solid; left one inline finding about a sibling boot-reload path that appears to have the same gap.

---

changelog.d/security/6768-cron-update-script-validation.md
```diff
@@ -0,0 +1 @@
+Apply the cron pre-processing script allowlist consistently to job updates. (#6768) (@houko)
```
houko, line 1: Commit `2e276f7` on this branch ("docs(changelog): add PR reference to 6768 fragment", which edited this file) has both author and committer identity set to `Claude <noreply@anthropic.com>`.
CLAUDE.md's git hygiene rule is explicit that commit author identity must not resolve to Claude/Anthropic even when the message text itself is clean, and the repo's `commit-msg` hook checks `GIT_AUTHOR_IDENT` specifically for this case.
The message body has no AI-attribution string, so this only shows up via `git log --format='%an <%ae>'`.
I can't fix this myself without rewriting history and force-pushing, both of which are outside what I'm allowed to do here, so flagging for the PR author/maintainer to squash or re-author that commit (e.g. `git rebase -i` + `git commit --amend --author`) before merge.

---

houko (commented):
Automated daily review pass. One correctness/scope item found and fixed directly (comment prose-wrapping on the lines this PR touched, per repo convention); one policy item left as an inline comment since fixing it requires a history rewrite I'm not permitted to force-push.

---

PR #6769, opened 2026-08-08
fix(sandbox): hash full container isolation config

houko:
## Summary
- include capability, read-only root, PID, tmpfs, CPU, and blocked-mount policy in Docker container pool hashes
- retain existing image, network, memory, and workdir matching
- pin every security-boundary field with a non-Docker unit regression

## Security impact
Prevents a pooled container created under a weaker isolation profile from being reused for a configuration that expects stronger sandbox boundaries.

## Verification
- RED: `cargo test -p librefang-runtime-sandbox-docker test_config_hash_covers_container_security_boundary --lib -- --nocapture`
- GREEN: `cargo test -p librefang-runtime-sandbox-docker --lib` (48 passed)
- `cargo clippy -p librefang-runtime-sandbox-docker --lib --no-deps -- -D warnings`
- `git diff --check`

### DIFF
--- a/changelog.d/security/6769-sandbox-config-hash.md
+++ b/changelog.d/security/6769-sandbox-config-hash.md
@@ -1,1 +1,1 @@
-Match pooled Docker containers on the full sandbox isolation configuration. (@houko)
+Match pooled Docker containers on the full sandbox isolation configuration. (#6769) (@houko)
--- a/changelog.d/security/6769-sandbox-config-hash.md
+++ b/changelog.d/security/6769-sandbox-config-hash.md
@@ -1,0 +1,1 @@
+Match pooled Docker containers on the full sandbox isolation configuration. (@houko)
--- a/crates/librefang-runtime-sandbox-docker/src/lib.rs
+++ b/crates/librefang-runtime-sandbox-docker/src/lib.rs
@@ -713,6 +713,12 @@ pub fn config_hash(config: &DockerSandboxConfig) -> u64 {
     config.network.hash(&mut hasher);
     config.memory_limit.hash(&mut hasher);
     config.workdir.hash(&mut hasher);
+    config.cpu_limit.to_bits().hash(&mut hasher);
+    config.read_only_root.hash(&mut hasher);
+    config.cap_add.hash(&mut hasher);
+    config.tmpfs.hash(&mut hasher);
+    config.pids_limit.hash(&mut hasher);
+    config.blocked_mounts.hash(&mut hasher);
     hasher.finish()
 }
 
@@ -1229,6 +1235,45 @@ mod tests {
             ..Default::default()
         };
         assert_ne!(config_hash(&c1), config_hash(&c2));
+    }
+
+    #[test]
+    fn test_config_hash_covers_container_security_boundary() {
+        let baseline = DockerSandboxConfig::default();
+        let mut variants = Vec::new();
+
+        let mut config = baseline.clone();
+        config.cap_add = vec!["CHOWN".into()];
+        variants.push(("cap_add", config));
+
+        let mut config = baseline.clone();
+        config.read_only_root = !config.read_only_root;
+        variants.push(("read_only_root", config));
+
+        let mut config = baseline.clone();
+        config.pids_limit += 1;
+        variants.push(("pids_limit", config));
+
+        let mut config = baseline.clone();
+        config.tmpfs.push("/run:size=8m".into());
+        variants.push(("tmpfs", config));
+
+        let mut config = baseline.clone();
+        config.cpu_limit = 2.0;
+        variants.push(("cpu_limit", config));
+
+        let mut config = baseline.clone();
+        config.blocked_mounts.push("/home/private".into());
+        variants.push(("blocked_mounts", config));
+
+        let baseline_hash = config_hash(&baseline);
+        for (field, config) in variants {
+            assert_ne!(
+                baseline_hash,
+                config_hash(&config),
+                "changing {field} must prevent container pool reuse"
+            );
+        }
     }
 }
 


crates/librefang-runtime-sandbox-docker/src/lib.rs
houko, line 1: The field set added here (`cpu_limit`, `read_only_root`, `cap_add`, `tmpfs`, `pids_limit`, `blocked_mounts`) does match every isolation flag `create_sandbox` actually sends to `docker run` — I checked each one against the `cmd.arg(...)` call sites and they line up, so the hash itself is now complete relative to `DockerSandboxConfig`.

One thing worth confirming before merge though: I couldn't find any production call site for `ContainerPool::acquire` / `release` (and therefore this `config_hash`) outside this crate's own tests. `crates/librefang-runtime/src/tool_exec_backend.rs` and `src/tool_runner/sandbox.rs` — the actual sandbox exec paths — call `create_sandbox` / `destroy_sandbox` directly per invocation and never go through `ContainerPool`. Likewise `reuse_cool_secs` (the config field that's supposed to gate pool reuse) has no reader anywhere in `librefang-runtime`.

If that's accurate, the collision this PR closes can't currently happen in the running daemon because containers aren't pooled/reused at all yet — the PR description's "Security impact" (a pooled container being reused across incompatible isolation configs) would describe a latent risk for whenever the pool gets wired in, not a live bug today. Not asking you to wire up the pool in this PR (that's a bigger change than a hash fix), just flagging so the description/impact framing can be corrected if I'm right, or so you can point me at the call site I missed if I'm wrong.

---

houko (commented):
Reviewed the hash-field completeness against the actual `docker run` invocation — looks correct and complete for every field `DockerSandboxConfig` currently has. Left one comment asking to confirm/correct the "Security impact" framing, since I couldn't find where `ContainerPool`/`config_hash` are actually wired into a live sandbox-exec call path.

---

crates/librefang-runtime-sandbox-docker/src/lib.rs
```diff
@@ -713,6 +713,12 @@ pub fn config_hash(config: &DockerSandboxConfig) -> u64 {
     config.network.hash(&mut hasher);
     config.memory_limit.hash(&mut hasher);
     config.workdir.hash(&mut hasher);
+    config.cpu_limit.to_bits().hash(&mut hasher);
+    config.read_only_root.hash(&mut hasher);
+    config.cap_add.hash(&mut hasher);
+    config.tmpfs.hash(&mut hasher);
+    config.pids_limit.hash(&mut hasher);
+    config.blocked_mounts.hash(&mut hasher);
```
houko, line 721: I checked the six added lines against `create_container`'s actual `docker run` arg list and they do close the real gap: every field that reaches a container-creation flag (`memory_limit`, `cpu_limit`, `pids_limit`, `cap_add`, `read_only_root`, `network`, `tmpfs`, `workdir`, `image`, plus `blocked_mounts` for the bind-mount validation) is now covered.

But `DockerSandboxConfig` has 18 fields total, and `enabled`, `container_prefix`, `timeout_secs`, `mode`, `scope`, `reuse_cool_secs`, `idle_timeout_secs`, and `max_age_secs` still aren't hashed. That's fine *today* — they're pool-lifecycle/activation policy, not isolation flags, so excluding them from a pool-matching hash is defensible. The problem is nothing enforces that boundary going forward: this is a listwise `.hash()` call, so a future isolation-relevant field (e.g. a seccomp profile path, `--shm-size`, `--ulimit`) can be added to the struct and wired into `create_container` without anyone touching this function, and the exact bug this PR fixes recurs silently.

Given the title claims "full container isolation config," consider making that a structural guarantee rather than a convention: exhaustively destructure the struct here, e.g.

```rust
let DockerSandboxConfig {
    image, network, memory_limit, workdir, cpu_limit, read_only_root,
    cap_add, tmpfs, pids_limit, blocked_mounts,
    // Pool-lifecycle/activation fields, intentionally excluded from the
    // isolation hash — do not add without checking create_container():
    enabled: _, container_prefix: _, timeout_secs: _, mode: _, scope: _,
    reuse_cool_secs: _, idle_timeout_secs: _, max_age_secs: _,
} = config;
```

so the compiler forces every new field to be explicitly included or explicitly ignored at this call site (`DockerSandboxMode`/`DockerScope` would also need `#[derive(Hash)]` if a future change decides they belong). That's a real refactor rather than a one-line fix, so flagging instead of changing it myself — happy to be told the current field-by-field list plus code review is considered sufficient.

---

houko (commented):
Automated review: the hash fix correctly closes the gap for every field that actually reaches a `docker run` isolation flag. One completeness question left inline about future field-drift protection — no other issues found (changelog fragment and per-field regression test are both present and correctly formatted, no HashMap/ordering non-determinism in the new code).

---

changelog.d/security/6769-sandbox-config-hash.md
```diff
@@ -0,0 +1 @@
+Match pooled Docker containers on the full sandbox isolation configuration. (#6769) (@houko)
```
houko, line 1: The current branch tip, commit `40d1e51` ("docs(changelog): add PR reference to 6769 fragment"), has its author identity set to `Claude <noreply@anthropic.com>` (`git log -1 --format='%an <%ae>' 40d1e51`).
CLAUDE.md's "Commit & PR hygiene" section and the repo's own `commit-msg` hook explicitly reject this: no AI/Claude/Anthropic attribution in commit messages, and the hook separately checks `GIT_AUTHOR_IDENT` so a clean message with Claude-authored identity still fails.
This looks like the `commit-msg` hook wasn't active in whatever clone produced this commit (see the "Enable once per clone" `just setup` note), since the hook is supposed to catch exactly this case.
I can't fix this myself without rewriting history and force-pushing, which is against the review policy for this run, so flagging for the PR author: please `git commit --amend --reset-author` (or interactive rebase) on this commit to attribute it to a human identity, then force-push your own branch.
Separately, this same commit's message body is hard-wrapped mid-sentence ("...but was\nmissing the (#6769) reference, so the release flow would keep the\ngenerated line...") rather than one-sentence-per-line as required by CLAUDE.md's prose-wrapping rule for commit message bodies — worth fixing in the same amend.

---

houko (commented):
Automated daily review pass: one policy-compliance finding left inline (AI-author-identity + commit-message prose-wrap issue on the branch tip). No fixes were pushed — the finding requires a history rewrite/force-push on the contributor's own branch, which this review run does not perform. Code itself (config_hash coverage, new regression test, changelog fragment content) looks correct.

---

changelog.d/security/6769-sandbox-config-hash.md
houko, line 1: Commit `40d1e51` ("docs(changelog): add PR reference to 6769 fragment") is authored and committed as `Claude <noreply@anthropic.com>`.

Per this repo's own policy (CLAUDE.md, "Commit & PR hygiene" / "No AI attribution"), commits with an author identity resolving to Claude/Anthropic are rejected — the `commit-msg` git hook specifically checks `git var GIT_AUTHOR_IDENT` for this, independent of the message text (which is otherwise clean here). This one apparently landed without that hook active.

Before merge, could you re-author this commit under your own identity (e.g. `git rebase -i` + `git commit --amend --reset-author`, or squash it into the preceding commit) and force-push the branch? I'm intentionally not doing this myself since fixing it requires rewriting already-pushed history, which needs a force-push I won't perform on someone else's branch without explicit go-ahead.

---

crates/librefang-runtime-sandbox-docker/src/lib.rs
```diff
@@ -713,6 +713,12 @@ pub fn config_hash(config: &DockerSandboxConfig) -> u64 {
     config.network.hash(&mut hasher);
     config.memory_limit.hash(&mut hasher);
     config.workdir.hash(&mut hasher);
+    config.cpu_limit.to_bits().hash(&mut hasher);
```
houko, line 716: I checked this against every `config.*` reference in the `docker run` construction (`create_container`, ~line 261-304): `image`, `network`, `memory_limit`, `workdir`, `cpu_limit`, `read_only_root`, `cap_add`, `tmpfs`, `pids_limit`, and `blocked_mounts` (used to validate the workspace bind mount) are now all hashed — that's every field that actually reaches a docker isolation flag or gates a bind mount. The unhashed fields (`enabled`, `container_prefix`, `timeout_secs`, `mode`, `scope`, `reuse_cool_secs`, `idle_timeout_secs`, `max_age_secs`) are pool/lifecycle bookkeeping only, not isolation strength, so leaving them out looks correct. Good regression test too (`test_config_hash_covers_container_security_boundary`).

One non-blocking concern, since this is exactly the bug class this PR fixes: this is a hand-maintained field list, not `#[derive(Hash)]`, so a future field added to `DockerSandboxConfig` that affects isolation (another mount type, a new cap-related knob, etc.) has to be remembered here — nothing enforces it at compile time. `#[derive(Hash)]` can't be added directly today because `cpu_limit: f64` isn't `Hash`, which is presumably why this was hand-rolled in the first place. Worth considering for a follow-up: wrap `cpu_limit` in a newtype (`ordered-float` or a local `struct CpuLimit(f64)` with a manual `Hash`/`Eq` impl) so the rest of the struct can derive `Hash`/`PartialEq`/`Eq`, making `config_hash` (or an `Eq` comparison) compiler-checked against new fields instead of relying on someone remembering to update this function. Not asking for that here — just flagging so it doesn't quietly regress the next time a field is added.

---

houko (commented):
Reviewed the isolation-config hash for completeness against the actual `docker run` construction. The new hash covers every field that reaches a security-relevant docker flag or bind-mount gate (image, network, memory_limit, workdir, cpu_limit, read_only_root, cap_add, tmpfs, pids_limit, blocked_mounts); the excluded fields are pool/lifecycle bookkeeping, not isolation strength. Changelog fragment and regression test are both present and correctly formatted. Left two comments: one flagging that commit 40d1e51 is authored/committed as `Claude <noreply@anthropic.com>`, which conflicts with this repo's no-AI-attribution policy and needs a re-author + force-push before merge; one non-blocking suggestion about the hand-maintained field list drifting the same way this bug did if a future field is added.

---

crates/librefang-runtime-sandbox-docker/src/lib.rs
```diff
@@ -713,6 +713,12 @@ pub fn config_hash(config: &DockerSandboxConfig) -> u64 {
...
+    config.read_only_root.hash(&mut hasher);
+    config.cap_add.hash(&mut hasher);
+    config.tmpfs.hash(&mut hasher);
+    config.pids_limit.hash(&mut hasher);
+    config.blocked_mounts.hash(&mut hasher);
     hasher.finish()
```
houko, line 722: Checked this against every field `create_sandbox` actually feeds into the `docker run` invocation (lines 260-304): `image`, `network`, `memory_limit`, `workdir`, `cpu_limit`, `pids_limit`, `cap_add`, `read_only_root`, `tmpfs`, `blocked_mounts` (used by `validate_bind_mount` before the `-v` mount is added).
With this change every one of those is now hashed — nothing that shapes the running container's isolation is silently excluded from the key. `container_prefix`, `timeout_secs`, `mode`, `scope`, `reuse_cool_secs`, `idle_timeout_secs`, `max_age_secs` are correctly left out since none of them reach the `docker run` args.

One adjacent gap worth flagging, not introduced by this PR: `ContainerPool::acquire`/`release` (lines 512, 529) key pool entries on `config_hash` alone, with no workspace or agent identity in the key.
`create_sandbox` bind-mounts `workspace` read-only at container-creation time (line 298) and that mount can't be changed on a running container, so if this pool is ever wired into the real `create_sandbox` call sites (`tool_exec_backend.rs:578`, `tool_runner/sandbox.rs:65` — today they call `create_sandbox` directly, bypassing the pool entirely), two different agents with identical `DockerSandboxConfig` but different workspaces could receive the same pooled container and inherit the wrong agent's workspace mount.
Getting `config_hash` complete is necessary but not sufficient for safe reuse once pooling goes live — the acquire/release API will also need workspace (or agent id) as part of the match key. Worth a tracking issue before Gap 5 gets wired up, since it's a design change (pool API shape) rather than a one-line fix.

---

houko (commented):
Reviewed the hash coverage against the actual `docker run` args built by `create_sandbox` — the six added fields close the gap completely; nothing that reaches the container invocation is left unhashed. Changelog fragment format/attribution checks out. One inline note on an adjacent, pre-existing gap in `ContainerPool` for when pooling gets wired into production (it currently isn't).

---

PR #6770, opened 2026-08-08
fix(llm): keep auth-failed credentials disabled

houko:
## Summary
- represent permanent auth-failure disablement separately from temporary cooldown timestamps
- prevent stale in-flight successes from reviving a key rejected by another request
- remove the 100-year Instant sentinel and preserve the existing diagnostic `u64::MAX` marker
- keep mark_success early recovery for temporary rate-limit cooldowns

## Security impact
Prevents known-invalid 401/403 credentials from re-entering shared pool rotation during concurrent requests, and removes a platform-dependent Instant overflow panic path.

## Verification
- RED: `cargo test -p librefang-llm-drivers mark_success_does_not_revive_permanently_disabled_key --lib -- --nocapture`
- GREEN: `cargo test -p librefang-llm-drivers credential_pool::tests --lib` (29 passed)
- `cargo clippy -p librefang-llm-drivers --lib --no-deps -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/librefang-llm-drivers/src/credential_pool.rs
+++ b/crates/librefang-llm-drivers/src/credential_pool.rs
@@ -85,9 +85,8 @@ pub struct PooledCredential {
     /// When `Some(t)`, this credential is exhausted and must not be used until
     /// `Instant::now() >= t`.
     exhausted_until: Option<Instant>,
-    /// Auth failures disable a credential until the pool is rebuilt by a
-    /// configuration reload. A separate flag keeps permanence distinct from
-    /// temporary cooldowns and cannot overflow the monotonic clock.
+    /// Auth failures disable a credential until the pool is rebuilt by a configuration reload.
+    /// A separate flag keeps permanence distinct from temporary cooldowns and cannot overflow the monotonic clock.
     permanently_disabled: bool,
 }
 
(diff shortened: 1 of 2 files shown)


crates/librefang-llm-drivers/src/credential_pool.rs
```diff
@@ -360,32 +363,28 @@ impl CredentialPool {
...
-    /// Unlike [`mark_exhausted`] which uses a TTL-based cooldown, this marks
-    /// the key as unavailable for the lifetime of the pool. The only way to
-    /// recover a permanently-exhausted key is via [`mark_success`] from a
-    /// concurrent code path or a hot-reload that rebuilds the pool.
-    ///
-    /// Implementation uses a far-future timestamp (100 years) so that
-    /// `is_available()` naturally returns `false` without a separate flag.
+    /// Unlike [`mark_exhausted`] which uses a TTL-based cooldown, this marks the key as unavailable for the lifetime of the pool.
+    /// Only a hot-reload that rebuilds the pool can recover the credential.
     pub fn mark_permanent(&self, api_key: &str) {
```
houko, line 368: This closes the retry/success race, but I think there are two related gaps worth a decision before this ships:

1. **No standalone recovery path for a single credential.** The only way to clear `permanently_disabled` is a hot-reload that rebuilds the whole pool (`rebuild_credential_pools` in `config_reload_ops.rs`), and that only fires when `field_changed(&old.credential_pools, &new.credential_pools)` is true — i.e. the operator must make an actual edit to `[[credential_pools]]` (or restart the daemon). If a key was disabled by a false-positive 401 and the operator wants to re-enable it without otherwise touching the pool config, there's no way to do that short of removing+re-adding the key (forcing a diff) or a full restart, and either one also wipes every *other* key's in-flight 429/402 cooldown in that pool as a side effect (see the doc comment on `rebuild_credential_pools`). Given `mark_permanent` is meant to be a stronger, harder-to-reverse state than `mark_exhausted`, should there be an explicit "re-enable this credential" admin action (CLI/API) instead of relying on the config-diff side effect?

2. **No distinction between a real invalid key and a transient 401/403.** `handle_driver_error` in `pooled_driver.rs` calls `mark_permanent` on the very first `AuthError`, with no retry/backoff and no "seen N times" threshold, unlike `RateLimit`/`CreditExhausted` which get TTL-bounded cooldowns. A single flaky/misclassified 401 (clock skew, a provider-side transient glitch, a proxy hiccup that returns 401 instead of 5xx) permanently sidelines the credential for the rest of the pool's life. Worth confirming this is intentional given how blunt the only recovery path (above) is.

---

PR #6771, opened 2026-08-08
fix(packaging): require auth in AUR Docker

houko:
## Summary
- change the shipped AUR Docker environment from no-auth opt-out to auth-required by default
- bump `pkgrel`, refresh the env checksum, and keep `.SRCINFO` synchronized
- warn upgrades that pacman preserves an existing `LIBREFANG_ALLOW_NO_AUTH=1` backup configuration
- retain explicit loopback-only no-auth opt-in for trusted development hosts

## Security impact
Prevents a fresh `librefang-docker` AUR installation from starting with authentication disabled, and tells existing users how to remediate the preserved legacy default.

## Verification
- RED: static assertion found `LIBREFANG_ALLOW_NO_AUTH=1`
- GREEN: shipped env source value is `0`
- actual env SHA-256 matches PKGBUILD and `.SRCINFO`
- PKGBUILD and `.SRCINFO` both report `pkgrel=2`
- `bash -n` for launcher and install hook
- `shellcheck` for launcher and install hook (`--shell=bash` for pacman hook)
- `git diff --check`

`makepkg` is unavailable in the macOS verification environment; checksum and metadata consistency were verified directly.


packaging/aur/librefang-docker/librefang-docker.install
```diff
@@ -6,5 +6,7 @@ post_install() {
 
 post_upgrade() {
   echo "LibreFang Docker runner upgraded."
+  echo "Security: pacman preserves /etc/librefang/docker.env. Review any existing"
```
houko: These two `echo` lines split one sentence ("Review any existing LIBREFANG_ALLOW_NO_AUTH=1 setting and change it to 0 unless no-auth is intentional.") across the boundary between two separate echo statements. CLAUDE.md's prose-wrapping rule wants one sentence per line and only allows breaking at a sentence boundary, but I'm not fully confident that rule is meant to reach discrete `echo` calls (each producing one fixed terminal line, not a reflowable comment block) rather than doc-comments/markdown. Flagging rather than fixing directly — worth confirming whether these should be two full sentences (one per echo) instead.

---

Issue #6748, opened 2026-08-07
Transcription of long recordings: hard-coded timeout and un-chunked results block media far below every size limit


nevgenov:
## Context

#6683 closed #6678 and #6679 and took two of the six items the latter listed as out-of-scope: video containers are accepted, and the audio track is extracted locally.
That issue said the remaining four would be filed separately if the direction was of interest. This is that issue.

Re-checking them against current `main` (`c403439a`) changed the shape of the list:

- One of them turned out to be a correctness bug in the merged extraction path rather than a scaling concern, and is filed separately as #6747.
- Two of the rest (**timeout**, **result chunking**) block recordings **far below every current size limit**, and touch none of the `// Validation constants (SECURITY)` block.
- One (**where the size limit is measured**) does touch that block, and is presented last as an open question rather than a request.

## 1. The hard-coded timeout is the real length ceiling

```rust
// crates/librefang-runtime-media/src/media_understanding.rs:873
.timeout(std::time::Duration::from_secs(60));   // whisper_transcribe

// crates/librefang-runtime-media/src/media_understanding.rs:1092
tokio::time::timeout(std::time::Duration::from_secs(30), child.wait())  // run_ffmpeg_pipe
```

Neither is configurable.

**This is not a big-file problem — the timeout binds far below every size limit.**

Measured against a self-hosted `whisper-large-v3-turbo` endpoint (`[media.custom_stt]`, Apple Silicon), on Ogg/Opus 32 kbps mono — the exact format `extract_video_audio_track` produces:

| Media | Input size | Wall-clock | Throughput | Transcript |
|---|---|---|---|---|
| 313 s (5.2 min) | 1.29 MB | 41.1 s | 7.6x realtime | 8 732 B |
| **604 s (10.1 min)** | **2.49 MB** | **71.7 s** | 8.4x realtime | **16 831 B** |
| 1208 s (20.1 min) | 4.97 MB | 150.1 s | 8.0x realtime | 33 682 B |

At a steady ~8x realtime the 60 s ceiling corresponds to **roughly 8 minutes of media**.

The middle row is the whole problem in one example: a **10-minute** recording exceeds the transcription timeout, while its input is **2.49 MB** — eight times under `MAX_AUDIO_BYTES` and twenty times under `MAX_VIDEO_BYTES`.
No size limit is anywhere near being reached.

So the effective length ceiling is set by a hard-coded timeout, is invisible to the operator, and is unrelated to the size limits that *are* documented and checked.
[...]

PR #6773, opened 2026-08-08
feat(media): transcribe long recordings in windows and write them to a file (#6748 item 2)

nevgenov:
Implements item 2 of #6748 (windowing + `out_path`). Item 1 (`[media] timeout_secs`) is yours; item 3 (measuring the size limit after extraction) is still open on the issue and is not in this branch.

## The problem

`media_transcribe` transcribed whole files and returned the transcript inline, so two limits that have nothing to do with file size decided how long a usable recording could be:

- a single transcription request is bounded by a wall-clock timeout that does not scale with the input;
- the kernel spills any tool result over `[tool_results] spill_threshold_bytes` (16 KB by default) to the artifact store and hands the agent a stub.

Both are reached around ten minutes of speech, at roughly 2.5 MB of extracted audio — eight times under `MAX_AUDIO_BYTES` and twenty times under `MAX_VIDEO_BYTES`.
No size limit is anywhere near being involved.

## Changes

- **Windowing.** `start_sec` / `max_secs` bound the request to one window; the response carries `has_more` / `next_start_sec` to walk the rest.
  `extract_media_window` cuts the span with `-ss` placed **before** `-i`, so ffmpeg seeks instead of decoding and discarding everything ahead of the window — on a 45-minute recording that is the difference between a constant-time seek and re-decoding the whole file per chunk.
  The cut lands on the same Ogg/Opus target `extract_video_audio_track` already produces, which subsumes both the video-extraction and `.oga` re-mux branches for windowed calls.
- **`out_path`.** Writes the transcript as UTF-8 through `workspace_sandbox::resolve_sandbox_path_ext` — the same resolution `web_fetch_to_file` and `file_write` use — and returns only path, byte count, sha256 and a 200-character preview.
  A window starting at `0` truncates and later windows append, so repeated calls assemble one transcript without any of it passing through the agent's context.
  The reported size and digest describe the whole file, not the last window's share, so a caller that walked five windows can verify what it ended up with.
- **Advancing by produced length, not requested length.** `ogg_opus_duration_secs` reads the granule position of the final Ogg page.
  Opus counts granules at a fixed 48 kHz regardless of encoder rate (RFC 7845 §4), so this is exact, and the pre-skip from `OpusHead` is subtracted so the error does not accumulate across chunks.
  A seek lands on a keyframe and a window overlapping the end of the recording is short, so an assumed edge drifts and eventually skips audio.
- **Both mechanisms, not either.** Window size varies with how much was said: a 600 s window measured ~16.8 KB of transcript against a 16 KB threshold, so windowing alone leaves the outcome dependent on speech density and `out_path` is what makes it deterministic.
- **Schema and docs.** The tool description now tells the agent to window long recordings and write them to a file; `docs/src/app/agent/tools` (en + zh) gained the new parameters and a worked example of the loop.

## Compatibility

A call that names neither window field takes the previous path unchanged, including adding no ffmpeg pass.
`transcribe_audio` keeps its signature and delegates to the new `transcribe_audio_window`, so all seven existing call sites are untouched.
Continuation fields are emitted only for windowed calls — offering `has_more` on a whole-file result would invite a caller to loop on a transcript that is already complete.

## Why not `ffprobe`

Duration is read out of the produced bytes rather than probed.
`ffprobe` is called nowhere in the kernel today, so depending on it would make it a new deployment requirement on every host for one number the Ogg container already carries.

## Verification

- `cargo test -p librefang-runtime-media --lib` — 94 passed, 4 new: granule-position duration against 1 s and 7.5 s streams, `None` for non-Ogg input, a window cut from the middle, and a window overlapping the end coming back short.
  The ffmpeg-gated ones were re-run with `--nocapture` to confirm they actually execute rather than self-skip.
- `cargo test -p librefang-runtime --lib` — 2180 passed, 7 new: window parsing and defaults, rejection of spans that cannot describe a window, character-boundary preview truncation (a byte-based cut would split a codepoint in exactly the non-Latin recordings this targets), append/truncate assembly, and that `out_path` cannot escape the workspace.
- `cargo clippy -p librefang-runtime -p librefang-runtime-media --all-targets -- -D warnings` — clean.
- `cargo fmt --all -- --check` — clean.

Not verified: no live provider round-trip. The transcription call itself is unchanged by this branch — only what is handed to it and what is done with the result.

## One measurement worth carrying into item 1

The default window is 600 s because that is the largest round number that stays inside the tool-result budget on ordinary speech.
But on a self-hosted `whisper-large-v3-turbo` endpoint that same window took **~72 s**, which is already past the current 60 s transcription timeout.
So the window default and whatever default `[media] timeout_secs` gets have to be chosen together — a 10-minute window is only a sane default on provider hardware fast enough to finish it in time.

## Review rounds so far

Two rounds landed on the draft; both turned up real defects, recorded here so the history is legible rather than buried in resolved threads.

- **Appended windows fused words at every boundary.** Each window's transcript arrives trimmed and they were concatenated with nothing between them, so `"and then we"` + `"discussed the budget"` wrote `"and then wediscussed the budget"`. A window edge lands mid-sentence by design, so this was the common case. Windows are now newline-separated.
- **An unreadable window edge could make the walk non-terminating.** `consumed_secs: None` used to fall back to `max_secs`, which reads as "full window, there is more" — the caller would advance past the end, get the same unreadable shape, and loop. Unknown now stops the walk, and `consumed_secs` is reported as null so "recording ended" and "edge unknown" stay distinguishable.
- **The rebase onto #6751 turned up the important one.** `extract_media_window` still fed the container on `pipe:0` — exactly the shape #6751 had just fixed — so windowed transcription of an ordinary mp4 would have reintroduced #6747 on a path that fix did not cover. The window now stages to a `ScopedTempFile`, which also makes `-ss` a real seek, and carries the same `ogg_contains_audio` guard.
- **The Ogg parser scanned for raw `OggS` magic.** Opus payloads are entropy-coded, so those four bytes occur inside real audio by chance, and eight bytes of compressed payload read as a granule position would produce a plausible but wrong duration — the exact drift this contract exists to prevent. Pages are now walked by their declared lengths.
- `has_more` / `next_start_sec` could disagree for a `max_secs` under the completeness tolerance; the condition and its test both now treat the two fields as one invariant.

One test-quality note worth stating plainly: the first version of the decoy-magic regression test **did not fail** under the mutation it was written for — the decoy sat too close to the end of the buffer for a naive reader to reach it, so it passed for the wrong reason. It now carries trailing padding and reports 3600s against a 1s stream when the fix is reverted.

## Still open

- Item 3 of #6748 (measuring the size limit after extraction) is deliberately absent. The issue asks whether you want it folded in here or kept separate; this branch assumes separate until you say otherwise, and it is a small change in the same file either way.
- Partial writes on append are documented rather than rolled back: buffering the assembled transcript to rewrite it atomically would restore, on disk, the proportional-to-recording-length cost `out_path` exists to remove. `file_bytes` / `file_sha256` make a short artefact detectable, and the documented recovery is to restart the walk from `start_sec = 0`, which truncates.


changelog.d/added/6748-transcribe-windowing-and-out-path.md
```diff
@@ -0,0 +1,8 @@
...
+`start_sec` and `max_secs` bound the request to one window and the response carries `has_more` / `next_start_sec` to walk the rest; `out_path` writes the transcript as UTF-8 and returns only the path, byte count, sha256 and a 200-character preview, following the contract `web_fetch_to_file` already established.
+Windows starting at `0` begin a new file and later windows append, so repeated calls assemble one transcript without any of it passing through the agent's context.
+Both mechanisms are needed rather than either alone: window size varies with how much was said, so a fixed window straddles the spill threshold instead of staying under it, and `out_path` is what makes the outcome independent of that.
+Callers advance by the produced window length rather than the requested one, read back from the Ogg granule position — a seek lands on a keyframe and a window overlapping the end of the recording is short, so an assumed edge drifts and eventually skips audio.
+A call that names neither window field keeps its previous behaviour exactly, including adding no ffmpeg pass. (#6748) (@nevgenov)
```
houko: Per `changelog.d/README.md`, the trailing `(#N)` group is what suppresses the auto-generated release-notes line for the PR that merges — "Only the last `(#N)` group on the bullet's last non-empty line counts... Every merged PR in the range gets a generated line unless its number appears in [that] group."

This bullet ends `(#6748)`, which is the tracking issue, not this PR (#6773). When #6773 merges, the release flow won't recognize this fragment as covering it, so the curated prose here and a generated `- feat(media): transcribe long recordings in windows and write them to a file (#6748 item 2) (#6773) (@nevgenov)` line will both land in the release notes.

Worth changing the ending to `(#6773) (@nevgenov)` (or `(#6748, #6773) (@nevgenov)` if you want the issue referenced too) before this leaves draft.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -270,11 +291,36 @@ impl MediaEngine {
             source_ext.clone().unwrap_or_else(|| "wav".to_string())
         });
 
+        // Windowed call (#6748): one ffmpeg pass cuts the requested span and
+        // lands it on the same Ogg/Opus target the two branches below produce,
+        // so neither has anything left to do. Handled first for that reason —
+        // running either of them before this would decode the whole recording
+        // just to throw most of it away.
```
houko: CLAUDE.md's prose-wrapping rule ("no column limit; break only at sentence boundaries") applies to "any multi-line free-form prose comment block," not just `///` doc comments — this repo's `///` blocks elsewhere in this same PR correctly use one-line-per-sentence, but this new `//` block (and several others added here, e.g. lines 262–264, 266–272, 274–277 in `tool_runner/media.rs`, and lines 1304–1306, 1314–1316, 1324–1325, 1332–1333 in this file) hard-wraps at a fixed column and breaks mid-sentence instead.

Not a functional issue, just flagging so it doesn't get carried forward as the pattern for the rest of this feature — worth a pass to join these back to one-sentence-per-line before merge.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,210 @@ pub(super) async fn tool_media_transcribe(
...
+    // already complete.
+    if let Some(window) = window {
+        // Advance by what ffmpeg actually produced, never by what was asked
+        // for: a seek lands on a keyframe and a window overlapping the end is
+        // short, so an assumed edge drifts and eventually skips audio.
+        // Falling back to the requested length when the granule position is
+        // unreadable keeps the walk moving rather than stalling it — at worst
+        // it repeats or skips one window, which the caller can see in the
+        // transcript, whereas a stall is silent.
+        let consumed = outcome.consumed_secs.unwrap_or(window.max_secs);
```
houko: This fallback reads backwards from a safety standpoint. `consumed_secs` comes back `None` from `ogg_opus_duration_secs` when the produced Ogg's last page carries no readable/finished granule (malformed trailing page, or a page whose granule is the `u64::MAX` "no packet finishes here" sentinel) — and the situation most likely to produce exactly that is a window cut at or past the true end of the recording, where ffmpeg's output degenerates to header-only or truncated bytes.

Falling back to `window.max_secs` in that case makes `has_more` (line 278) come back `true`, i.e. "assume the whole window was consumed, there's more after this." If a stream ever produces this shape at the tail — rather than the `Some(~0.0)` that a header-only page with granule 0 would give you, which *does* terminate correctly — every subsequent call re-extracts a window that's still past EOF, gets the same unreadable shape again, and `next_start_sec` keeps advancing by a full `max_secs` forever. Nothing bounds the walk in that branch.

Given the doc comment already reasons through "repeats or skips one window" as the accepted cost, it might be worth also handling "the walk doesn't stop" — e.g. defaulting the unreadable case to `has_more = false` (safe-stop bias) or surfacing it as a distinct field the caller can act on, rather than silently assuming full consumption.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,210 @@ pub(super) async fn tool_media_transcribe(
...
+        // on a packet boundary, so an exactly-covered window still lands a few
+        // milliseconds short and a strict comparison would call every window
+        // the last one.
+        let has_more = consumed >= window.max_secs - WINDOW_COMPLETE_TOLERANCE_SECS;
+        response["window"] = serde_json::json!({
+            "start_sec": window.start_sec,
+            "max_secs": window.max_secs,
+            "consumed_secs": consumed,
+        });
+        response["has_more"] = serde_json::json!(has_more);
```
houko: Per CLAUDE.md's integration-testing rule ("Always write integration tests at the injection site, not just the implementation site"): `transcribe_window_tests` covers `parse_window`, `preview_of`, and `write_transcript` well, but the `has_more` / `next_start_sec` computation here — the actual walk-termination logic, and per the sibling comment on this thread the part most exposed to an edge case — has no test at all, presumably because it's inline in a function that needs a live `MediaEngine`.

Pulling it into a small pure helper (e.g. `fn window_continuation(window: MediaWindow, consumed_secs: Option<f64>) -> (bool, Option<f64>)`) would make it unit-testable without a provider round-trip, and it's exactly the kind of boundary-condition logic (exact-length window, one-tick-short window, `None` consumed_secs) that benefits most from a table test.

---

houko (commented):
Automated daily review pass. Draft PR, item 2 of #6748 — reviewed against CLAUDE.md's changelog-fragment, testing, and prose-wrapping conventions plus the chunking/windowing correctness itself. Four findings left inline, none blocking the draft status; not approving/requesting changes per the automation's scope.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,210 @@ pub(super) async fn tool_media_transcribe(
...
+        }
+    }
+
+    let transcript = &outcome.understanding.description;
+    match input["out_path"].as_str() {
+        None => {
+            response["transcript"] = serde_json::json!(transcript);
+        }
+        Some(out_path) => {
+            let written = write_transcript(
```
houko, line 345: Consecutive windows are appended with no separator, and the per-window transcript is `.trim()`ed before it gets here (`media_understanding.rs:436`), which strips exactly the whitespace that would otherwise separate two chunks.

A window boundary is very likely to land mid-sentence or mid-word by design (that's the whole point of `has_more`/`next_start_sec`), so the common case — not just an edge case — is: window N's transcript ends `"...and then we"`, window N+1's starts `"discussed the budget"`, and `write_transcript` concatenates them byte-for-byte into `"...and then wediscussed the budget"` on disk. This directly contradicts the "functional concern" the PR description otherwise reasons carefully about (keyframe drift, granule-accurate advancing) — the assembled artifact can come out with words fused together at every single window boundary, not just occasionally.

Suggest inserting a single space (or newline) between chunks when appending (skip it only for the very first write), or at minimum note the caveat in the tool description/docs so a caller knows to post-process.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,210 @@ pub(super) async fn tool_media_transcribe(
...
+    // rewrite the file whole would reintroduce, on disk, exactly the
+    // proportional-to-recording-length cost this parameter exists to remove.
+    let mut file = tokio::fs::OpenOptions::new()
+        .create(true)
+        .write(true)
+        .append(append)
+        .truncate(!append)
+        .open(&resolved)
+        .await
+        .map_err(|e| ToolError::upstream_msg(format!("Failed to open '{out_path}': {e}")))?;
```
houko, line 547: `OpenOptions::open` with `truncate(!append)` truncates at open time, before any bytes are written. If `write_all` (line 435) or `flush` (line 438) then fails partway — disk full, I/O error — on an append call, the file is left with whatever partial bytes made it to disk with no rollback, and there's no temp-file+rename or similar safeguard.

Given the intended usage pattern is "call again with the same `start_sec` on failure" (per the retry reasoning already written into the `has_more` fallback comment above), a retry after a partial-write failure would append the full transcript again after the corrupted partial bytes, silently producing a mangled transcript rather than a clean retry. Worth at least documenting the limitation, or writing to a temp file in the same directory and renaming into place for the truncate (fresh-window) case, where atomicity is straightforward.

Not blocking for a draft, but worth tracking before this leaves draft since the feature exists specifically to make long-running, multi-call assembly reliable.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,210 @@ pub(super) async fn tool_media_transcribe(
...
+                // the next or have to invent its own assembly step.
+                window.is_some_and(|w| w.start_sec > 0.0),
+                workspace_root,
+                additional_roots,
+            )
+            .await?;
+            response["written_to"] = serde_json::json!(written.path);
+            response["bytes_written"] = serde_json::json!(written.bytes);
+            response["sha256"] = serde_json::json!(written.sha256);
+            response["chars"] = serde_json::json!(transcript.chars().count());
```
houko: Minor clarity nit: `bytes_written` and `sha256` describe the **whole assembled file** (per `write_transcript`'s doc comment), but `chars` (and `preview`) describe only the **current window's** transcript. An agent reading this response has no way to tell those two are different scopes just from the field names — worth a distinct name like `window_chars` (mirroring `window.consumed_secs`) or a doc-string update on the tool schema to call out the scope difference explicitly.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,210 @@ pub(super) async fn tool_media_transcribe(
...
+        // Falling back to the requested length when the granule position is
+        // unreadable keeps the walk moving rather than stalling it — at worst
+        // it repeats or skips one window, which the caller can see in the
+        // transcript, whereas a stall is silent.
+        let consumed = outcome.consumed_secs.unwrap_or(window.max_secs);
+        // A window is "full" only within a tolerance: the produced stream ends
+        // on a packet boundary, so an exactly-covered window still lands a few
+        // milliseconds short and a strict comparison would call every window
+        // the last one.
+        let has_more = consumed >= window.max_secs - WINDOW_COMPLETE_TOLERANCE_SECS;
```
houko: The `has_more`/`next_start_sec` assembly (lines 265-287) is the part of this change most likely to have an off-by-one, and it isn't unit-tested directly — it's only reachable through `tool_media_transcribe`, which needs a live `MediaEngine`. `parse_window`, `preview_of`, and `write_transcript` all got extracted and unit-tested; consider doing the same here (a small `fn window_advance(window, consumed) -> (bool, Option<f64>)`-shaped helper) so the tolerance boundary and the `consumed_secs == None` fallback path get direct coverage instead of only indirect exercise through a live transcription call.

Non-blocking for a draft, flagging since it's the one branchy bit of new logic without a test.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -194,12 +194,33 @@ impl MediaEngine {
     /// the first one detected from env vars (Groq, OpenAI, Gemini,
     /// ElevenLabs, …). There is no runtime cascade; a provider failure
     /// surfaces as `Err` to the caller.
+    ///
+    /// Transcribes the whole file. Callers that need to bound the request — anything driven by a recording whose length they do not control — want [`Self::transcribe_audio_window`] instead (#6748).
```
houko: Repo `CLAUDE.md` prose-wrapping rule: one sentence per line in doc comments. This line has two ("Transcribes the whole file." + "Callers that need..."). Same issue at line 1248 (`Requested length. The produced window is shorter...`) and in `docs/src/app/agent/tools/page.mdx:229` (`Lower \`max_secs\` if a call times out. How much media fits...`). Mechanical fix, just needs a line break after each first sentence.

---

houko (commented):
Automated review pass. Path-traversal on `out_path` looks handled correctly (goes through `resolve_sandbox_path_ext`, and it's tested — `out_path_cannot_escape_the_workspace`). Main finding is a real functional bug in how appended windows are joined (no separator, combined with the pre-existing `.trim()`), plus a few smaller robustness/test-coverage/docs nits, including two spots that don't follow the repo's one-sentence-per-line prose rule. Nothing here should block further draft iteration; flagging the window-join issue as the one worth fixing before this leaves draft, since it affects the correctness of the assembled transcript that's the whole point of `out_path`.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -2163,4 +2317,123 @@ mod tests {
...
         assert_eq!(parsed.custom_stt.model.as_deref(), Some("large-v3"));
     }
+
+    // ── windowed transcription (#6748) ──────────────────────────────────
+    //
+    // These pin the two facts a caller walks a long recording on: the window ffmpeg cut, and how much of the recording it actually covered.
+    // Both are read back out of the produced bytes rather than assumed, because a seek lands on a keyframe and a window past the end of the media is short.
+
+    /// Generate `secs` seconds of tone as an Ogg/Opus stream at the same
+    /// 48 kHz mono target the extraction paths produce.
```
houko: CLAUDE.md prose-wrapping rule ("no column limit; break only at sentence boundaries") — this doc comment is one sentence hard-wrapped across two lines at a fixed column rather than broken at a sentence boundary:

```
/// Generate `secs` seconds of tone as an Ogg/Opus stream at the same
/// 48 kHz mono target the extraction paths produce.
```

Same pattern recurs at lines 2380-2381 (`Guards the fallback contract: … / `None` rather than …`) and 2413-2415 (`The end-of-recording signal: … / short, and … / what tells the caller to stop.`). Worth a second pass since the dedicated rewrap commit (378176d8) covers this file's other test-module blocks but these three were missed. Suggested fix: put each sentence on its own line, however long.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,257 @@ pub(super) async fn tool_media_transcribe(
...
+            response["file_bytes"] = serde_json::json!(written.bytes);
+            response["file_sha256"] = serde_json::json!(written.sha256);
+            response["window_chars"] = serde_json::json!(transcript.chars().count());
+            response["window_preview"] = serde_json::json!(preview_of(transcript));
+        }
+    }
+
+    Ok(serde_json::to_string_pretty(&response)?)
+}
+
```
houko, line 365: CLAUDE.md prose-wrapping rule — same issue as flagged in `media_understanding.rs`, but throughout the new `transcribe_window_tests` module here (which the rewrap commit didn't touch, since that module is new to this PR, not carried over from before). Examples of one sentence hard-wrapped at a fixed column instead of broken at sentence boundaries:

- L305-306: `A call that names no window field stays on the whole-file path, so / every pre-#6748 caller keeps its exact behaviour and gains no ffmpeg hop.`
- L322-324, L358-360, L372-375, L415-416, L456-458, L463-464, L469-471, L494-496: same pattern.

Rest of the diff (production code doc comments) already follows the one-sentence-per-line convention correctly — it's specifically this test module that reverts to fixed-column wrapping. Suggested fix: reflow each of these blocks so each sentence is its own line regardless of length.

---

houko (commented):
Mechanical CLAUDE.md-compliance pass (draft-appropriate scope only — no incomplete-feature nits).

Two inline findings on the prose-wrapping rule ("no column limit; break only at sentence boundaries"), see line comments.

One more of the same kind with no diff line to anchor to: the bodies of commits 95aacd8 and 17c1c8c are hard-wrapped at a fixed column (e.g. "Addresses the review on #6773. Two of the findings were real defects in the\nfirst pass; the rest are naming, coverage and prose.") rather than broken at sentence boundaries, which the rule explicitly calls out as applying to "Commit message bodies." Not actionable without rewriting history mid-review, flagging for awareness before this lands.

Everything else checked out: no AI/Claude/Anthropic attribution in commits or docs, commit format is conventional with no Co-Authored-By footer, and the changelog fragment (`changelog.d/added/6748-transcribe-windowing-and-out-path.md`) is correctly formed — new file, no leading `- `, one sentence per line, ends `(#6748, #6773) (@nevgenov)`. The `out_path` write path goes through `resolve_sandbox_path_ext` with a dedicated escape test (`out_path_cannot_escape_the_workspace`), so no path-traversal concern.

---

docs/src/app/agent/tools/page.mdx
```diff
@@ -208,6 +208,27 @@ Tool calls into provider-specific media APIs. Per-tool capability grants require
...
+### Transcribing a long recording
+
+A whole meeting returned in one tool result does not reach the agent intact: the kernel spills any result over `[tool_results] spill_threshold_bytes` (16 KB by default) to the artifact store and hands the agent a stub, and a single transcription request is bounded by a wall-clock timeout that does not scale with the length of the input.
+Both limits are reached long before any file-size limit is — a ten-minute recording is a couple of megabytes of extracted audio.
+
+So transcribe in windows, and write to a file:
+
+```
+media_transcribe(path="meeting.mp4", max_secs=600, out_path="meeting.txt")
+→ { "written_to": "...", "has_more": true, "next_start_sec": 600.0, "preview": "..." }
```
houko: This example response still uses `"preview"`, but the rename commit (95aacd8) changed the actual field to `window_preview` (alongside `file_bytes`/`file_sha256`/`window_chars`) — the tool schema description in `definitions.rs` already reflects the new name, this worked example just wasn't updated alongside it. As written, an agent following this doc would look for a `preview` key that the response never contains.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,257 @@ pub(super) async fn tool_media_transcribe(
...
+            has_more: false,
+            next_start_sec: None,
+        };
+    };
+    // A window is "full" only within a tolerance: the produced stream ends on a packet boundary, so an exactly-covered window still lands a few milliseconds short and a strict comparison would call every window the last one.
+    let has_more = consumed >= window.max_secs - WINDOW_COMPLETE_TOLERANCE_SECS;
+    WindowContinuation {
+        has_more,
+        // A window that produced nothing cannot advance the cursor; treating it as progress would spin on the same offset.
+        next_start_sec: (has_more && consumed > 0.0).then_some(window.start_sec + consumed),
```
houko: For a `max_secs` smaller than `WINDOW_COMPLETE_TOLERANCE_SECS` (0.25) — which `parse_window` permits, it only rejects `max_secs <= 0.0` — this can report `has_more: true` with no `next_start_sec`. E.g. `max_secs = 0.1`, `consumed = 0.0`: `has_more = 0.0 >= 0.1 - 0.25` is `true`, but `next_start_sec` requires `consumed > 0.0` (line 344), so it comes back `None` and the field is omitted from the response entirely (it's only set `if let Some(next) = ...` in the caller). The response then tells the agent there's more to transcribe with nowhere to resume from.

`window_continuation_decides_when_the_walk_stops`'s "produced nothing" case (`w(0.1)`, `Some(0.0)`) only asserts `next_start_sec == None`, not `has_more`, so this doesn't surface as a test failure. Might be worth requiring `consumed > 0.0` in the `has_more` condition itself, not just in the `next_start_sec` derivation, so the two fields can't disagree.

---

docs/src/app/agent/tools/page.mdx
houko: The worked example shows the response key as `"preview"`, but the actual field name `tool_media_transcribe` sets (in `tool_runner/media.rs`) is `window_preview` (alongside `window_chars`), not `preview`. Worth fixing so the doc matches what a caller actually gets back:

```
→ { "written_to": "...", "has_more": true, "next_start_sec": 600.0, "window_preview": "..." }
```

---

docs/src/app/zh/agent/tools/page.mdx
```diff
@@ -198,7 +198,7 @@ agent 不能硬删除已调度的任务:`schedule_delete` / `cron_cancel` 只会
 | `image_generate(prompt, size?, style?)` | 文生图(OpenAI / Replicate / Vertex)。 |
 | `image_analyze(image, question)` | 对已有图像做 vision Q&A。 |
 | `media_describe(image_or_video)` | 给已有媒体素材生成 caption / 描述。 |
-| `media_transcribe(audio_or_video, language?, prompt?)` | 音频转文本；也接受视频容器,服务端会提取其中的音轨。无法识别的扩展名会被拒绝。 |
+| `media_transcribe(audio_or_video, language?, prompt?, start_sec?, max_secs?, out_path?)` | 音频转文本；也接受视频容器,服务端会提取其中的音轨。无法识别的扩展名会被拒绝。`start_sec` / `max_secs` 只转写一个时间窗口,并在响应中返回 `has_more` / `next_start_sec` 以便继续；`out_path` 会把转写结果写入工作区文件,只返回路径、字节数、sha256 和一段预览。长录音两者都需要,详见下文。 |
```
houko, line 202: This row's new sentence ends "长录音两者都需要,详见下文。" ("Both are needed for long recordings, see below"), promising a walkthrough further down the page the way the English `page.mdx` got its new "### Transcribing a long recording" section (lines 211-231 there, with the worked `media_transcribe(...)` example).

No matching section exists in this file — the page goes straight from this table (ending at line 210) to `## Owner 通知` at line 213. A Chinese-reading caller following "详见下文" finds nothing to see. Either add the equivalent section here or drop "详见下文" from the row until the walkthrough is translated.

---

houko (commented):
Checked the current HEAD (378176d8c) against the existing review threads to avoid duplicating the 7 still-open findings from prior passes (window-join separator — already fixed in 95aacd8; non-atomic partial write; two prose-wrap spots; the `has_more`/`next_start_sec` disagreement for sub-tolerance `max_secs`; and the `page.mdx` `preview`→`window_preview` doc mismatch, already flagged twice). Path-traversal handling on `out_path` (`resolve_sandbox_path_ext`, tested), resource bounds (ffmpeg's existing 30s wall-clock cap plus `MAX_AUDIO_BYTES`/`MAX_VIDEO_BYTES` already bound window extraction regardless of requested `max_secs`), and `ogg_opus_duration_secs`'s parsing of malformed/truncated Ogg pages all look sound and don't panic. One new finding below: the Chinese docs translation grew a "see below" reference to a worked long-recording example that was only added to the English page.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1177,6 +1224,113 @@ async fn extract_video_audio_track(input_bytes: &[u8]) -> Result<Vec<u8>, String
...
+        .windows(8)
+        .position(|w| w == b"OpusHead")
+        .and_then(|at| ogg.get(at + 10..at + 12))
+        .map(|b| u16::from_le_bytes([b[0], b[1]]) as f64)
+        .unwrap_or(0.0);
+
+    // Scan forward for the last capture pattern rather than seeking back from the end: page sizes vary with the segment table, so the final page's offset cannot be computed, only found.
+    let mut last_granule: Option<u64> = None;
+    let mut at = 0usize;
+    while let Some(found) = ogg[at..]
```
houko: `ogg_opus_duration_secs` finds page boundaries by scanning for the raw 4-byte magic `OggS` in the byte stream, with no page-structure or CRC validation. The Ogg spec's own answer to exactly this — a compressed Opus packet's entropy-coded payload coincidentally containing the byte sequence `4F 67 67 53` — is the per-page CRC32 checksum, which this scan never checks. A coincidental match inside real audio payload (roughly file_size/2^32 odds, so unlikely on one file but not negligible at scale) would be read as a page header, and the "granule position" bytes taken from that position would be noise from the compressed audio rather than an actual granule value, silently producing a wrong `consumed_secs` and misleading `next_start_sec` / `has_more` for the caller — the exact class of "windows drift and audio gets skipped/duplicated" failure this feature exists to prevent. Since the function already walks every match to find the *last* one, validating that a candidate page has a plausible header (segment table length, or reconstructing/checking the CRC) before accepting its granule value would close this without needing `ffprobe`.

---

houko (commented):
Automated review — 1 finding (unvalidated Ogg page-magic scan in `ogg_opus_duration_secs` can misread a coincidental byte match inside compressed audio as a page header, silently corrupting the computed window duration). Otherwise this is a thoroughly tested, well-reasoned change (windowing, `out_path` sandboxing, append/truncate semantics, changelog fragment) with no other CLAUDE.md violations spotted.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,260 @@ pub(super) async fn tool_media_transcribe(
...
+            has_more: false,
+            next_start_sec: None,
+        };
+    };
+    // A window is "full" only within a tolerance: the produced stream ends on a packet boundary, so an exactly-covered window still lands a few milliseconds short and a strict comparison would call every window the last one.
+    //
+    // `consumed > 0.0` is part of the condition rather than only of the cursor below.
+    // For a `max_secs` under the tolerance — which `parse_window` permits, it only rejects zero — the tolerance alone would call a window that produced *nothing* full, and the caller would be told there is more with nowhere to resume from.
+    // The two fields have to agree: "there is more" is only meaningful alongside a place to continue.
+    let has_more = consumed > 0.0 && consumed >= window.max_secs - WINDOW_COMPLETE_TOLERANCE_SECS;
```
houko: The completeness check's threshold, `window.max_secs - WINDOW_COMPLETE_TOLERANCE_SECS`, goes negative whenever `max_secs <= WINDOW_COMPLETE_TOLERANCE_SECS` (0.25s) — and `parse_window` only rejects `max_secs == 0`, so any positive value under 0.25s is a legal request.

Once the threshold is negative, `consumed >= <negative>` is true for **any** positive `consumed`, not just a fully-covered window. Example: `max_secs = 0.2`, `consumed = 0.01` (the recording genuinely ended almost immediately) → `has_more` still comes out `true`, and the caller is told to resume at `start_sec + 0.01`.

The comment right above this line (lines 340-343) already identifies half of this problem — "the tolerance alone would call a window that produced *nothing* full" — and the fix that shipped is `consumed > 0.0`. That guard only rules out the exact-zero case the comment describes; it doesn't rule out the general case of a small nonzero `consumed` that is nowhere near `max_secs` once `max_secs` itself is below the tolerance. The two fields would still "agree" (`has_more` and `next_start_sec` both present, satisfying the invariant tested in `window_continuation_decides_when_the_walk_stops`), but `has_more` itself would be wrong — telling the caller to continue a walk that has actually ended, on a window that came back far short of what was asked for.

In practice `max_secs` defaults to 600 and the tool description steers callers toward multi-minute windows, so this likely never fires through the tool surface as written — but `window_continuation` and `MediaWindow` are public-ish internal API (`crate::media_understanding::MediaWindow`), and the bound isn't enforced by `parse_window` beyond `> 0`. Worth clamping the subtraction (e.g. `(window.max_secs - WINDOW_COMPLETE_TOLERANCE_SECS).max(0.0)`) or requiring `max_secs` to exceed the tolerance in `parse_window`, so the boundary math holds regardless of how small a window is requested.

---

houko (commented):
Automated review: this is a well-tested, carefully-reasoned PR (windowing math, keyframe-drift awareness, path-traversal reuse via `resolve_sandbox_path_ext`, UTF-8-safe preview truncation, changelog fragment all present and correct). One boundary-math finding on the `has_more` completeness check for very short windows.

---

crates/librefang-runtime/src/tool_runner/media.rs
nevgenov: Fixed in 632f4b2c — the finding is right, and the boundary is worse than the `consumed > 0.0` guard I added last round accounted for: that only closed the exact-zero case, while a hair above zero clears a negative bar just as well.

One note on the suggested remedy, since I went a different way. Clamping the subtraction at zero does not close it: with `max_secs = 0.2` and `consumed = 0.01`, the threshold becomes `0.0` and `0.01 >= 0.0` still holds, so the same window still reports more to come. I ran the suite against that variant to be sure rather than reasoning about it — it fails the new test the same way the unclamped version does.

What does hold is capping the tolerance at half the window, so it can never exceed the quantity it is a tolerance on. Windows above half a second — every window the defaults and docs steer callers toward — keep the flat 0.25 s behaviour unchanged.

The mutation check now distinguishes the fix from both nearby wrong answers (flat tolerance, clamp-at-zero) rather than only from the original, and the new cases cover a short window that barely started, the same short window actually filled (so small windows stay walkable), and a 20 ms shortfall on a 600 s window.

nevgenov:
Ran an adversarial self-review pass over this branch before asking for more of your time on it. Four defects, all fixed in 55448258, plus one false statement of mine corrected.

**`out_path` was validated after the work it guards.** Resolution sat inside `write_transcript`, which runs after the ffmpeg pass and the billed provider request — and on that branch the transcript is deliberately not in the response, so a rejected path destroyed work already paid for and the caller paid again for the retry. An absolute path outside the workspace is routine model output, so this was not an exotic input. `web_fetch_to_file` had it right all along: it resolves its destination ahead of the SSRF check and the fetch. I cited that file twice as the model to follow and then inverted its order.

**A window past the end of the recording failed the walk instead of ending it.** The guard's own comment called such a request legitimate and then returned `Err` — and the walk reaches that state unaided: for a recording near a multiple of `max_secs` the last full window sets `has_more`, and the next one starts exactly at EOF. The loop the tool description prescribes therefore ended in an error. An overshoot is now a zero-length window, which `window_continuation` already turns into a clean stop; an empty window at offset zero still errors, since that means the input could not be decoded.

**A silent stretch aborted the walk with no resume point.** An empty transcript was a hard error — correct for a whole file, wrong for one window, where a pause between speakers is ordinary. The error carried no offset, so the assembled file just stopped at the silence with nothing to say so.

**The Ogg walk discarded usable granules on a truncated tail**, contradicting the loop guard's own promise, and separately accepted a granule from a page whose payload had not arrived. Narrow trigger — `run_ffmpeg_pipe` requires exit 0 — but it was a contract the comment and the commit message both stated and the code did not keep.

**Correction.** `f251a5e6`'s message says "all seven existing call sites are untouched". Wrong twice: there are six production call sites of `transcribe_audio` on this base, and one of them is the one this branch rewrites. Five are untouched. The compatibility claim holds; the count did not.

Two notes on test quality, since both are the kind of thing that would otherwise have shipped looking covered:

- `ogg_opus_duration_reads_the_final_granule_position` asserted within 0.05 s while the pre-skip subtraction it was credited with guarding is worth 0.0065 s — seven times narrower than the bar, so neutralising the subtraction passed. The tolerance is now 0.002 s and the mutation fails.
- `out_path_cannot_escape_the_workspace` drove `write_transcript` directly, so it was structurally blind to *where* that call sat relative to the paid work. It now drives the resolver, and also asserts an ordinary relative path still resolves, so it cannot pass by rejecting everything.

Every fix is pinned by a test that I checked by mutation — reverting the defect fails that test and only that test.

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1259,6 +1343,183 @@ fn ogg_contains_audio(bytes: &[u8]) -> bool {
...
+    // a seek still exits 0 and still emits a headers-only container, so
+    // neither guard in `run_ffmpeg_pipe` catches it.
+    //
+    // Unlike that path, an empty result here is not always a failure.
+    // A window at or past the end of the recording is a legitimate request — the walk reaches one by itself whenever a recording's length is close to a multiple of `max_secs` — and it produces exactly the same headers-only bytes.
+    // The two are indistinguishable at the byte level, so they are told apart by where the window starts: a request beyond the opening of the recording that came back empty is an overshoot and ends the walk, while an empty window at the very start means the input itself could not be decoded.
+    if !ogg_contains_audio(&out) {
+        if window.start_sec > 0.0 {
+            return Ok(None);
+        }
```
houko, line 1470: New `Ok(None)` branch (the fix for "a window past the end failed the walk instead of ending it") has no test coverage.

The sibling test `extract_media_window_past_the_end_returns_a_short_window` (below, ~line 2820) only covers a window that *overlaps* the end — `ogg_contains_audio` is still true there, so it takes the `Ok(Some(out))` path, never this branch. Nothing in the diff exercises a window whose `start_sec` is fully past the recording's true end, which is exactly the shape this branch exists for and the one `window_continuation`/the walk-termination logic now depends on.

Suggest a test alongside the existing one: synth a short recording (e.g. `synth_ogg_opus(1.0)`), call `extract_media_window` with `start_sec` at or past 1.0s, and assert `Ok(None)`. No provider/network needed — same shape as the existing ffmpeg-gated test.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,284 @@ pub(super) async fn tool_media_transcribe(
...
+    let window = parse_window(input)?;
+
+    // Resolved before the provider call, not inside the write below.
+    // Transcription costs an ffmpeg pass and a billed request, and on this
+    // branch the transcript exists nowhere else — it is not put in the
+    // response when a destination was named — so a path rejected afterwards
+    // destroys work that was already paid for and makes the caller pay again
+    // for the retry.
+    // `web_fetch_to_file` orders it the same way, resolving its destination
+    // ahead of the SSRF check and the fetch.
```
houko: Hard-wrapped again: "Transcription costs an ffmpeg pass and a billed request, and on this branch the transcript exists nowhere else — it is not put in the response when a destination was named — so a path rejected afterwards destroys work that was already paid for and makes the caller pay again for the retry." is one sentence split across lines 253–257; same for the `web_fetch_to_file` sentence on 258–259. CLAUDE.md: "One sentence = one line, regardless of length" — applies to comment blocks, no column limit. Please re-join each sentence onto its own line.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -270,11 +292,49 @@ impl MediaEngine {
...
+        if let Some(window) = window {
+            let Some(cut) = extract_media_window(&audio_bytes, window)
+                .await
+                .map_err(|e| format!("ffmpeg window extraction failed: {e}"))?
+            else {
+                // Overshoot: the window began past the end of the recording.
+                // Reported as a zero-length window rather than an error, so the
+                // caller's loop ends on `has_more: false` the way the tool
+                // description tells it to, instead of on a failure it has no
+                // way to distinguish from a broken file.
```
houko: Hard-wrapped: the sentence "Reported as a zero-length window rather than an error, so the caller's loop ends on `has_more: false` the way the tool description tells it to, instead of on a failure it has no way to distinguish from a broken file." is split across lines 304–307 instead of sitting on one line. Same prose-wrapping rule flagged in earlier passes on this PR (and in the dedicated rewrap commit a7cc9cacc) — this is new prose from the latest commit that reintroduces it.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -389,6 +449,27 @@ impl MediaEngine {
 
         let transcription = transcription.trim().to_string();
         if transcription.is_empty() {
+            // A whole file that transcribes to nothing is a failure worth
+            // surfacing. One *window* of a recording that transcribes to
+            // nothing is ordinary — a pause, a silent stretch, a gap between
+            // speakers — and erroring there would abort the walk with no
+            // resume point, stranding the rest of the recording behind a
+            // silence.
```
houko: Hard-wrapped: two sentences split across lines 452–457 — "A whole file that transcribes to nothing is a failure worth surfacing." (452–453) and "One *window* of a recording that transcribes to nothing is ordinary — a pause, a silent stretch, a gap between speakers — and erroring there would abort the walk with no resume point, stranding the rest of the recording behind a silence." (453–457). Please put each sentence on its own line per the repo's prose-wrapping rule.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1259,6 +1343,183 @@ fn ogg_contains_audio(bytes: &[u8]) -> bool {
...
+            "pipe:1",
+        ],
+        None,
+        "transcribe a time window of a recording",
+    )
+    .await?;
+
+    // Same defence in depth as the video path: a demux that fails for want of
+    // a seek still exits 0 and still emits a headers-only container, so
+    // neither guard in `run_ffmpeg_pipe` catches it.
```
houko: Hard-wrapped: "Same defence in depth as the video path: a demux that fails for want of a seek still exits 0 and still emits a headers-only container, so neither guard in `run_ffmpeg_pipe` catches it." is one sentence split across lines 1419–1421 (the rest of this same comment block, 1423–1425, correctly keeps one sentence per line — this opening sentence is the outlier).

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -1259,6 +1343,183 @@ fn ogg_contains_audio(bytes: &[u8]) -> bool {
...
+        let granule = u64::from_le_bytes(granule_bytes);
+        // Page length is the header, the segment table, and the sum of the
+        // table's entries — following it is what keeps payload bytes from
+        // being mistaken for the next header.
+        let segments = ogg[at + 26] as usize;
+        // A truncated tail must not discard what was already read — the guard
+        // above says so and this is the other half of it: a stream that stops
+        // inside the segment table or mid-payload leaves the granules of every
+        // complete page before it perfectly usable, and `?` here would throw
+        // them away along with the incomplete one.
```
houko: Two more hard-wrapped sentences in this new hunk: lines 1493–1497 ("A truncated tail must not discard what was already read — the guard above says so and this is the other half of it: a stream that stops inside the segment table or mid-payload leaves the granules of every complete page before it perfectly usable, and `?` here would throw them away along with the incomplete one.") and lines 1503–1506 ("The granule is only meaningful once the page it describes has actually arrived: accepting it from a page whose payload was cut off would report audio the caller never received, and the caller would advance past it."). Each should collapse to one line per the prose-wrapping rule.

---

crates/librefang-runtime-media/src/media_understanding.rs
```diff
@@ -2365,4 +2626,218 @@ mod tests {
...
+    #[test]
+    fn ogg_opus_duration_rejects_non_ogg_input() {
+        assert_eq!(ogg_opus_duration_secs(&[0u8; 64]), None);
+        assert_eq!(ogg_opus_duration_secs(b""), None);
+    }
+
+    /// A stream that stops inside a page must still report the granules of the
+    /// complete pages before it — the walk's own comment promises exactly that,
+    /// and an early `?` used to discard them, stopping a caller's walk short of
+    /// the recording's end.
```
houko: Same hard-wrap pattern recurs in the new test module (the exact class of thing commit a7cc9cacce was dedicated to fixing last round): doc comment 2744–2747 is one sentence split across 4 lines; the comment at 2773–2775 packs two sentences across 3 lines ("Cut inside the third page's header, its segment table, and its payload in turn." / "Every one of those leaves the 1 s of complete pages before it intact, so every one must report 1 s rather than nothing."); and 2785–2787 is another single sentence split across 3 lines. Please rewrap all three to one sentence per line.

---

houko (commented):
Automated review of the 2 commits pushed since the last pass (632f4b2c and 5544825861, latest at 19:48:42Z). Both fix real defects (window-completeness tolerance going negative for sub-tolerance windows; out_path validated after the billed provider call; walk erroring instead of ending on an overshoot/silent window; Ogg parser discarding valid pages on a truncated tail) and are well mutation-tested for the functional side.

Two classes of finding:
- One test-coverage gap: the new `Ok(None)` overshoot branch in `extract_media_window` (the core of the "end the walk on an empty window" fix) has no test hitting it — the sibling test only covers a window overlapping the end, not one fully past it.
- Six hard-wrapped comment blocks (prose split mid-sentence across lines) reintroducing the exact issue flagged — and dedicated-rewrap-commit'd — earlier in this same PR's review history.

Nothing here blocks continued draft iteration; flagging before this leaves draft.

---

crates/librefang-runtime/src/tool_runner/media.rs
```diff
@@ -229,12 +247,284 @@ pub(super) async fn tool_media_transcribe(
     let attachment = build_transcription_attachment(&resolved, &ext).await?;
```
houko: Resource-bound question raised in review scope: windowing bounds the *transcription request* and the *result size*, but not the per-call cost of handling the source file itself.

Each call to `tool_media_transcribe` — including every window in a multi-window walk over one long recording — runs `build_transcription_attachment` (line 247) fresh, which re-reads the whole file from disk via `read_with_size_limit(resolved, MAX_AUDIO_BYTES)` and re-encodes it to base64 into the attachment. `transcribe_audio_window` then base64-decodes that whole attachment back into `audio_bytes` and, for a windowed call, passes the *entire* thing to `extract_media_window`, which stages it to a fresh `ScopedTempFile` before ffmpeg seeks into it with `-ss`.

So the seek genuinely avoids decoding the parts of the file outside the window (as the PR description correctly claims), but the disk read + base64 round-trip + temp-file staging of the *whole* source file happens once per window, not once per walk. For a caller walking N windows over one recording, that fixed-per-call cost is paid N times rather than being amortized — e.g. a 45-minute recording walked in 5 windows of 600s re-reads/re-stages the full file 5 times.

This is bounded, not unbounded — each call is still capped at `MAX_AUDIO_BYTES` / `MAX_VIDEO_BYTES` exactly as before, so there's no memory-safety or DoS concern — and for the sizes discussed in the PR body (~2.5 MB for ten minutes of speech) it's cheap in absolute terms. Flagging as a scaling characteristic worth being aware of (or noting in docs) rather than a defect: "windowing" caps what reaches the provider and the agent per call, but the source-file handling cost still scales with `(file size) × (number of windows)` for a full walk, not with `file size` alone.

Not blocking — the feature solves the two problems it explicitly set out to solve (timeout, spill threshold), and this is orthogonal to those.

---

houko (commented):
Automated review pass, scoped to resource bounds / temp-file handling / out_path path-safety on the windowed-transcription feature.

This PR already carries an extensive, high-quality review history (multiple rounds, several real defects caught and fixed by the author) — most of the usual concerns (path traversal via `out_path`, temp-file cleanup via `ScopedTempFile`, prose-wrap and test-coverage nits) are already covered by earlier passes and don't need repeating.

One additional finding below, on the resource-bound question specifically: windowing bounds the transcription request and the result size per call, but the source-file read/base64/temp-stage cost is still paid in full on every windowed call rather than once per walk. Non-blocking — flagged for awareness, not a defect in what the PR set out to fix.

Everything else checked (git safety, commit attribution, changelog fragment presence/format, prose wrapping on newly-touched lines) matches what prior passes already reported: no AI/Claude attribution in commits, conventional commit format, changelog fragment present and correctly formed, `out_path` traversal handled and tested.




---

PR #6774, opened 2026-08-08
fix(deploy): require caller-owned OpenRouter key

houko:
## Summary

- require deployers to provide their own OpenRouter API key
- remove all reads of the Worker-level `OPENROUTER_API_KEY` binding
- forward only the caller-owned key into the caller-owned Fly Machine
- return a fixed machine-creation error so an upstream config echo cannot reflect the key
- update the deploy page and result copy to describe BYOK accurately
- add isolated Node behavior tests and gate Worker deployment on them

## Security

The deploy Worker previously copied one shared OpenRouter credential into every user-owned Fly Machine. Every deployer could inspect and reuse that shared credential. This change removes the shared-secret path entirely and prevents Fly validation errors from reflecting a caller key.

## Verification

- `npm test` in `deploy/worker` (4 passed)
- `node --check src/index.js`
- `actionlint .github/workflows/deploy-worker.yml`
- `git diff --check`

## Out of scope

The separately reported deploy endpoint rate limiting, IP allocation error handling, and rollback/cleanup behavior remain follow-up items.

### DIFF
--- a/.github/workflows/deploy-worker.yml
+++ b/.github/workflows/deploy-worker.yml
@@ -21,6 +21,10 @@ jobs:
         with:
           node-version: "24"
 
+      - name: Test Worker
+        working-directory: deploy/worker
+        run: npm test
+
       - name: Deploy Worker
         working-directory: deploy/worker
         env:
--- a/changelog.d/security/6774-deploy-worker-byok.md
+++ b/changelog.d/security/6774-deploy-worker-byok.md
@@ -1,0 +1,2 @@
+The hosted Fly deploy flow no longer copies a shared OpenRouter credential into user-owned machines, where every deployer could inspect and reuse it.
+  Deployers now provide their own key, and the Worker forwards only that caller-owned credential into the caller's Fly machine configuration (#6774) (@houko)
--- a/deploy/worker/package.json
+++ b/deploy/worker/package.json
@@ -1,0 +1,7 @@
+{
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "test": "node --test src/index.test.js"
+  }
+}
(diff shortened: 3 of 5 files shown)


houko (commented):
Automated review pass. The security fix itself looks solid: the shared `env.OPENROUTER_API_KEY` read is gone entirely, the caller-owned key is validated server-side in `handleDeploy` (not just the client-side UI check) before any Fly API calls, and the machine-creation failure path now returns a fixed error string instead of echoing Fly's raw response body (which could otherwise reflect the caller's key back). Node tests cover all of this and pass locally (`npm test`, 4/4).

Pushed one mechanical fix directly (commit `cd7353b`): the new comment in `handleDeploy`'s machine-creation error branch wrapped a single sentence across two lines mid-phrase ("Never reflect Fly's / raw validation body..."), which violates the repo's prose-wrapping rule in CLAUDE.md (one sentence per line). Split it into two comment lines, one per sentence — no functional change, tests still pass, `node --check` and `git diff --check` clean.

One non-blocking nit: the PR description's "Security" paragraph is three sentences run together on one line, which also falls under the same prose-wrapping rule (applies to PR bodies per CLAUDE.md). Not fixing this myself since it's your PR description text — flagging for you to reflow if you'd like, purely cosmetic.

Nothing else stood out: changelog fragment (`changelog.d/security/6774-deploy-worker-byok.md`) is correctly formatted (section dir, one-sentence-per-line, ends `(#6774) (@houko)`, no AI attribution), the CI gate (`npm test` before deploy) is wired correctly, and no config/manifest fields or Rust API routes were touched so the config-field checklist and auth-middleware-allowlist rules don't apply here.

---

PR #6776, opened 2026-08-08
fix(skill-sdk): validate WASM ABI memory bounds

houko:
## Summary
- reject non-positive guest allocations and negative `execute` pointer/length values
- validate host-response pointers against current WASM linear memory before constructing a slice
- cap slice lengths to the wasm32 `isize` boundary and document the remaining FFI safety contract

## Verification
- `CARGO_TARGET_DIR=target cargo test` (11 passed)
- `CARGO_TARGET_DIR=target cargo clippy --all-targets -- -D warnings`
- `CARGO_TARGET_DIR=target cargo check --target wasm32-unknown-unknown --all-targets`
- `CARGO_TARGET_DIR=target cargo clippy --target wasm32-unknown-unknown --all-targets -- -D warnings`
- `git diff --check`

Out of scope: host runtime input-length validation and unrelated SDK error taxonomy findings remain separate audit items.

### DIFF
--- a/sdk/rust/librefang-skill/src/lib.rs
+++ b/sdk/rust/librefang-skill/src/lib.rs
@@ -148,10 +148,9 @@ fn copy_guest_memory(ptr: u32, len: u32) -> Option<Vec<u8>> {
         return Some(Vec::new());
     }
 
-    // SAFETY: the host contract returns memory obtained from this module's
-    // `alloc` export. The checked range is non-null and wholly contained in
-    // current linear memory. The bytes are copied before this function
-    // returns, so no reference escapes or survives memory growth.
+    // SAFETY: the host contract returns memory obtained from this module's `alloc` export.
+    // The checked range is non-null and wholly contained in current linear memory.
+    // The bytes are copied before this function returns, so no reference escapes or survives memory growth.
     Some(unsafe { core::slice::from_raw_parts(range.start as *const u8, range.len()).to_vec() })
 }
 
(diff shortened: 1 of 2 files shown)


sdk/rust/librefang-skill/src/lib.rs
```diff
@@ -106,6 +106,54 @@ pub fn unpack(packed: i64) -> (u32, u32) {
...
+    len: i32,
+    memory_len: usize,
+) -> Option<core::ops::Range<usize>> {
+    let ptr = u32::try_from(ptr).ok()?;
+    let len = u32::try_from(len).ok()?;
+    checked_guest_memory_range(ptr, len, memory_len)
+}
+
+#[cfg(target_arch = "wasm32")]
+fn copy_guest_memory(ptr: u32, len: u32) -> Option<Vec<u8>> {
```
houko, line 148: The bounds arithmetic itself checked out in review (`checked_add` guards the overflow, the `end <= memory_len` comparison has no off-by-one, and both call sites are synchronous with no intervening `.await`/callback that could grow memory between the check and the copy, so there's no TOCTOU window).

One open question on coverage, flagging as a judgment call rather than fixing myself: `copy_guest_memory` here and the `execute` wasm32 branch at line 393 are both `#[cfg(target_arch = "wasm32")]`, so `cargo test` on the host never runs them — only the pure `checked_guest_memory_range` / `checked_signed_guest_memory_range` helpers get exercised by the unit tests added in this PR.
CI's `skills-wasm-sdk` job builds `examples/custom-skill-wasm` for `wasm32-unknown-unknown`, which proves this compiles, but a build isn't an execution — nothing actually runs the compiled module in a WASM engine and asserts a malicious ptr/len is rejected end-to-end rather than trapping.
Given this is the security-sensitive path the PR title calls out, is an integration test that drives the compiled wasm32 skill through `crates/librefang-runtime`'s sandbox (feeding an out-of-range ptr/len and asserting a graceful error) worth adding? That would mean touching the runtime crate rather than just this SDK, so raising it rather than deciding unilaterally.

---

PR #6777, opened 2026-08-09
fix(channels): guarantee message split progress

houko:
## Summary
- guarantee `split_message` consumes input when an incomplete entity retreats to byte zero
- consume one complete UTF-8 character when a positive byte limit falls inside that character
- fail closed for a zero limit and preserve outermost nested-tag grouping

## Verification
- regression RED: prior implementation timed out on an incomplete entity at the chunk start
- updated `types::tests`: 32 passed
- `cargo clippy -p librefang-channels --lib --no-deps -- -D warnings` (initial fix; final delta independently reviewed for lint semantics)
- `git diff --check`

Scope note: the current built-in adapters are Python sidecars with their own splitter. This hardens the public Rust helper used by custom consumers, tests, and benchmarks.

### DIFF
--- a/changelog.d/fixed/6777-channel-split-progress.md
+++ b/changelog.d/fixed/6777-channel-split-progress.md
@@ -1,0 +1,5 @@
+The public Rust channel message splitter now always consumes at least one
+complete character, including when an incomplete HTML entity begins a chunk or
+the byte limit falls inside a multi-byte UTF-8 character. Custom Rust channel
+consumers can no longer enter a non-progressing split loop on those inputs
+(#6777) (@houko)
--- a/changelog.d/fixed/6777-channel-split-progress.md
+++ b/changelog.d/fixed/6777-channel-split-progress.md
@@ -1,5 +1,2 @@
-The public Rust channel message splitter now always consumes at least one
-complete character, including when an incomplete HTML entity begins a chunk or
-the byte limit falls inside a multi-byte UTF-8 character. Custom Rust channel
-consumers can no longer enter a non-progressing split loop on those inputs
-(#6777) (@houko)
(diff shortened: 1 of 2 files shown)


houko (commented):
Reviewed the Rust fix in `crates/librefang-channels/src/types.rs` — logic and the four new regression tests check out (traced `test_incomplete_entity_at_chunk_start_makes_progress`, `test_zero_limit_fails_closed_without_spinning`, `test_multibyte_character_larger_than_limit_makes_progress`, and `test_nested_tags_stay_balanced_across_chunks` by hand against `split_message`/`retreat_past_html_entity`/`retreat_past_html_tag`; `cargo test -p librefang-channels --lib types::tests` is green, 32/32, and `cargo clippy -p librefang-channels --all-targets -- -D warnings` / `cargo fmt --check -p librefang-channels` are both clean).

One out-of-scope-for-this-PR observation for a maintainer call, not fixed here since it's a different package/language: `sdk/python/librefang/sidecar/common.py::split_message` (docstring: "Mirrors the shared Rust `split_message` helper") has the same non-progress failure mode this PR guards against in Rust, for `limit=0`.
With `limit=0`: `window = rest[:0]` is `""`, `cut = window.rfind("\n")` is `-1`, so `cut` gets set to `limit` (`0`); then `rest = rest[cut:].lstrip("\n") if cut < limit else rest[cut:]` takes the `else` branch (`0 < 0` is `False`), leaving `rest` unchanged while `chunks.append(rest[:cut])` appends an empty string every iteration — an infinite loop, mirroring exactly the bug class this PR fixes on the Rust side.
Might be worth a companion Python-side fix/issue if `limit=0` is reachable from any sidecar call site; leaving it to your judgment since it's outside this PR's crate.

---

PR #6823, opened 2026-08-09
fix(python-sdk): bound HTTP request timeouts

houko:
## Summary
- apply a 30-second timeout to generated Python SDK requests and streams
- allow callers to override the timeout per client
- keep generator template and committed client synchronized

## Picked up while rebasing onto current `main`

This branch was 48 commits behind, and both items below are consequences of the signature change above meeting work that landed on `main` after the fork. Neither is a separate feature — each is the cost of the timeout parameter, paid where it fell due.

**`test_generated_client_trailing_sse.py` mocks updated (4 sites).** The trailing-SSE tests landed on `main` after this branch forked, so their `urlopen` lambdas still carried the pre-timeout signature and raised `TypeError: <lambda>() got an unexpected keyword argument 'timeout'` once this branch passes `timeout=self.timeout`. Fixed the same way as the three mock sites this branch had already updated: `lambda _request: response` becomes `lambda _request, **_kwargs: response`. This is the "two independently green PRs break on merge" case — neither side is wrong on its own, and text-level conflict detection cannot see it because the edits touch different files.

**Three `produce()` tests in `test_telegram_adapter.py` now wait on their condition rather than on a tick budget.** They drove the adapter loop with a fixed number of event-loop turns (`for _ in range(64)` / `range(32)`) and then asserted how far it had got. How many turns one iteration costs depends on how the executor schedules the polling thread, so the constant bets on runner speed rather than bounding the behaviour under test. Both bets lost on CI during this PR's own runs: `test_produce_treats_longpoll_timeout_as_normal` failed with `only saw 3 polls` against its 32-turn budget, and `test_produce_backoff_is_capped_at_max` failed with `got [1.0, 2.0, 4.0]` against its 64-turn budget — in both cases the loop was healthy and the tick budget was not. Each now spins until its own condition holds, bounded by a 5s wall-clock deadline; `fake_sleep` consumes no real time, so the deadline only bounds the pathological case.

## Verification
- full Python 3.13 SDK suite: **1993 passed**
- `scripts/tests/test_codegen_sdks.py` — the merge conflict there was two independent assertion blocks added on both sides with an empty base; both sides kept
- flaky fix confirmed by mutation: forcing the wait deadline to `+0.0` turns all three `produce()` tests red, so the assertions genuinely depend on the wait rather than passing incidentally
- 20 consecutive runs of the three `produce()` tests: 20 pass, 0 fail


changelog.d/fixed/6823-python-sdk-request-timeout.md
```diff
@@ -0,0 +1,3 @@
+Apply a bounded timeout to every generated Python SDK HTTP request. (@houko)
```
houko: changelog.d/README.md: not tooling-enforced, but ending the bullet with `(#6823)` before `(@houko)` avoids `cargo xtask release` generating a duplicate `- <PR title> (#6823) (@houko)` line for this same PR in the release notes.

---

houko (commented):

---

scripts/tests/test_codegen_sdks.py
```diff
@@ -76,6 +76,8 @@ def main():
     assert_in('"error": format!("stream error: {}", e)', rs, "rust-stream-transport-error")
     assert_not_in("while let Some(Ok(chunk))", rs, "rust-no-silent-stream-error")
     assert_in('"status": resp.StatusCode', go, "go-error-event-status")
+    assert_in("DEFAULT_TIMEOUT = 30.0", py, "python-default-timeout")
```
houko, line 96: This branch conflicts with current `main` in this file (confirmed via `git merge-tree` against `origin/main`), matching the `has-conflicts` label / `dirty` mergeable_state. More than cosmetic: `main` has since added `URLError` handling to both `_request` and `_stream` (`except URLError as e: raise LibreFangError(f"Connection error: {e.reason}") from e`, guarded by `python-urlerror-import` / the `except URLError` count assertion in this same test file), which this branch doesn't have yet. A rebase needs to keep both the new `URLError` wrapping and this PR's `DEFAULT_TIMEOUT` / `urlopen(..., timeout=...)` assertions rather than letting one silently replace the other.

---

sdk/python/librefang/librefang_client.py
```diff
@@ -99,7 +102,7 @@ def _stream(self, method: str, path: str, body: Any = None, query: Optional[Dict
         headers["Accept"] = "text/event-stream"
         req = Request(url, data=data, headers=headers, method=method)
         try:
-            resp = urlopen(req)
+            resp = urlopen(req, timeout=self.timeout)
```
houko, line 110: Unrelated to this PR's timeout change, but noticed while reviewing this function and worth a heads-up since it's the same bug class as the sibling `fix/rust-sdk-trailing-sse` (#6837) PR in this batch: `_stream`'s read loop (lines 112-114, `chunk = resp.read(4096); if not chunk: break`) drops whatever is left in `buffer` when a clean EOF arrives without a trailing newline, exactly like the Rust SDK bug #6837 fixed. This is pre-existing (confirmed present on current `origin/main` too, not introduced here), so out of scope for a timeout-only PR — but worth a follow-up issue/PR mirroring #6837's fix for the Python client.

---

houko (commented):
Rebased onto current origin/main (0fcbc7dea) to resolve the stale-base conflict flagged in the prior review — this branch's base predated main's `URLError` wrapping (#6824) and the UTF-8 stream-split fix (#6825), so merging as-is risked losing one set of changes. Kept both.

Combining the two independently-green branches surfaced a real interaction: main's `test_generated_client_urlerror.py`, `test_generated_client_stream_utf8.py`, and the existing `test_generated_client_stream.py` all monkeypatch `urlopen` with `lambda _request: response`, and this PR's `urlopen(req, timeout=self.timeout)` call broke all of them with `TypeError: unexpected keyword argument 'timeout'`. Widened those lambdas to `lambda _request, **_kwargs: response` — `cargo`-equivalent here is `python3 -m pytest sdk/python/tests/test_generated_client_*.py`, 10 passed. Also verified `python3 scripts/tests/test_codegen_sdks.py` and that `python3 scripts/codegen-sdks.py` regenerates a byte-identical `sdk/python/librefang/librefang_client.py`.

Left the 30s default value itself alone (not touched) — unlike the sibling `fix/rust-sdk-timeouts` PR's hardcoded `.timeout()` override, this one is genuinely configurable via the `timeout=` constructor kwarg with no hidden override, so a deployment needing longer non-streaming calls has a documented escape hatch. No action needed there.

---

sdk/python/librefang/librefang_client.py
houko, line 1: Blocking: two commits on this branch carry author identity `Claude <noreply@anthropic.com>` — `6b818bc9` ("Merge remote-tracking branch 'origin/main' into pr-6823") and `0fcbc7de` ("fix(python-sdk): accept the new urlopen timeout kwarg in existing test mocks"), the latter of which is this PR's current HEAD commit.
CLAUDE.md prohibits AI/Claude/Anthropic attribution anywhere in the repo, and author identity is checked separately from commit message text.
This needs a human rebase to rewrite those two commits under a real human `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` (e.g. interactive rebase plus `git commit --amend --reset-author`) before merge.
Not something an automated pass can fix by force-pushing over someone else's branch.

---

sdk/python/librefang/librefang_client.py
```diff
@@ -101,7 +104,7 @@ def _stream(self, method: str, path: str, body: Any = None, query: Optional[Dict
...
```
houko, line 110: `_stream` reuses the same `self.timeout` that `_request` uses, but `urlopen(req, timeout=…)` sets the underlying socket's timeout, which then bounds *every* subsequent `resp.read(4096)` call in the loop below (lines 112-114), not just connection setup.
A quiet SSE gap longer than `timeout` (30s default) — a long tool call, a slow-to-start model response, etc. — raises a bare `socket.timeout`/`TimeoutError` on the next `resp.read(4096)`, which is neither caught by the `except HTTPError` / `except URLError` above it nor wrapped as `LibreFangError`; it propagates out of the generator as a raw stdlib exception.
Compare the sibling Rust SDK fix in #6836, which deliberately keeps the streaming body exempt from its total-request deadline ("SSE bodies remain exempt from the total deadline so long-lived streams continue normally") and applies its timeout only to the non-streaming request path.
This PR's changelog says "Clients now use a 30-second default for both paths" — that's exactly the pattern #6836 avoided for streams.
Suggest keeping `timeout=self.timeout` for the initial `urlopen()` call (bounds connect + header wait) but relaxing or clearing the socket timeout before entering the `while True: resp.read(4096)` loop, so an idle stream isn't killed the way a stalled non-streaming request should be.

---

scripts/codegen-sdks.py
```diff
@@ -206,7 +209,7 @@ def _stream(self, method: str, path: str, body: Any = None, query: Optional[Dict
         headers["Accept"] = "text/event-stream"
         req = Request(url, data=data, headers=headers, method=method)
         try:
-            resp = urlopen(req)
+            resp = urlopen(req, timeout=self.timeout)
```
houko, line 215: Same issue as the committed client (see the review comment on `sdk/python/librefang/librefang_client.py`): the generator template emits `urlopen(req, timeout=self.timeout)` for `_stream` too, so every future codegen run reproduces the same idle-read timeout on long-lived SSE streams.
Needs the same streaming exemption as the Rust generator path in #6836, applied here so template and committed client stay in sync as this PR already intends.

---

houko (commented):
Automated daily review pass.

Two findings, both blocking-class:
1. Commit author identity — HEAD commit and its immediate predecessor are authored as `Claude <noreply@anthropic.com>`, a hard CLAUDE.md violation. Needs a human rebase.
2. Streaming timeout correctness — `_stream` inherits the same idle-read timeout as `_request` via `urlopen(timeout=…)`, unlike the sibling Rust SDK fix (#6836) which explicitly exempts SSE bodies from the deadline. Long-lived streams can be killed mid-flight by an unwrapped `socket.timeout`.

Everything else checked out: default (30s) is reasonable for the non-streaming path, the timeout is caller-overridable via the constructor, the generator template and committed client stay in sync, and the changelog fragment is present and correctly formatted.




---

sdk/python/librefang/librefang_client.py
houko, line 110: Possible regression for legitimate long-lived SSE streams — flagging as ambiguous rather than fixing directly since the right resolution is a design choice.

`urlopen(req, timeout=self.timeout)` sets a **per-socket-operation** timeout in Python's `urllib` (via `socket.settimeout`), not a total-request deadline. That timeout keeps applying to every subsequent blocking `resp.read(4096)` call in the `while True` loop below, i.e. to the gaps *between* SSE chunks, not just to establishing the connection.

With the new 30s default, an agent response that goes quiet for >30s between events (slow tool call, "thinking" pause, long LLM generation) will now raise an uncaught `socket.timeout`/`TimeoutError` mid-stream and kill the connection — a real functional regression, not just hardening. `test_generated_client_timeout.py`'s `FakeStreamResponse.read()` returns immediately, so it can't catch this; none of the new tests exercise an idle gap between chunks.

Contrast with the sibling Rust PR (#6836) in this same batch, which explicitly exempts `do_stream` from the total request timeout so long-lived streams are unaffected. This PR applies the same bounded timeout to both `_request` and `_stream` uniformly.

Worth a maintainer decision: should the stream's `urlopen` timeout be dropped/raised specifically for the read loop (mirroring #6836's approach), or is a 30s inter-chunk idle bound intentional here? Same question applies to the codegen template at `scripts/codegen-sdks.py:212`.

---

changelog.d/fixed/6823-python-sdk-request-timeout.md
houko, line 1: Also flagging a repo-policy issue for the maintainer before merge: this PR's current HEAD commit `0fcbc7dea04527f76387292864f29baaa8136929` ("fix(python-sdk): accept the new urlopen timeout kwarg in existing test mocks"), and the merge commit below it (`6b818bc9a3...`), are authored by `Claude <noreply@anthropic.com>`.
CLAUDE.md's commit-msg hook is specifically designed to reject a commit whose author identity resolves to Claude/Anthropic, so these slipped through with hooks not applied (or bypassed) at commit time.
Fixing authorship on already-pushed commits needs a rebase + force-push, which is outside what an automated reviewer should do unilaterally — leaving this for a maintainer to reauthor before merge.

(Separate note: verified `python3 scripts/tests/test_codegen_sdks.py` OK and the relevant Python tests pass — see the other comment on `librefang_client.py:107` for the one substantive finding.)

---

houko (commented):
Automated daily review pass — see inline notes (one substantive finding on the streaming timeout, plus a commit-authorship note).

---

sdk/python/librefang/librefang_client.py
houko, line 110: `urllib.request.urlopen(..., timeout=n)` sets the socket's timeout for the lifetime of the connection (via `socket.settimeout`), not just for the connect phase.
For `_request` that is fine — the whole call is bounded and that is the intent.
For `_stream`, though, this same value also bounds every subsequent `resp.read(4096)` in the polling loop below, so any SSE stream that goes quiet for longer than `self.timeout` (default 30s, e.g. a slow LLM turn with no token/keepalive for half a minute) will raise a `socket.timeout`/`OSError` and kill an otherwise-healthy stream — it isn't caught anywhere in `_stream`, so it propagates out of the generator.
The changelog text ("SSE stream setup... leaving connection establishment... without an inactivity bound") reads as if this is scoped to connection setup, but the implementation applies it to the whole streaming lifetime.
Compare with the sibling Rust fix in #6836, which deliberately keeps the total-request timeout off the streaming path and only applies a *connect* timeout there, for exactly this reason ("SSE bodies remain exempt from the total deadline so long-lived streams continue normally").
Worth confirming this is the intended behavior for Python before merging — if not, `_stream`'s `urlopen` call needs a distinct (connect-only) timeout, or reads need to tolerate/reset on inactivity, rather than reusing `self.timeout` as an implicit per-read inactivity bound.
The same pattern is duplicated in `scripts/codegen-sdks.py`'s template, so it would need the same treatment there.

---

houko, line 110: `self.timeout` (default 30s) is applied here as the socket timeout, which bounds *each* blocking recv, not the stream's total lifetime — so this is effectively an inactivity/read timeout on the SSE connection, re-applied on every `resp.read(4096)` in the loop below.

Is 30s the right default for this path specifically? The primary documented use of `_stream` is `send_message_stream` (see the module docstring), and an agent turn can legitimately go >30s between deltas (tool calls, extended thinking) with no bytes on the wire in between — that read would then raise `socket.timeout`/`TimeoutError` mid-generator.

Two related concerns, both design calls rather than mechanical fixes:
1. Should streaming calls use a longer (or separately configurable) read timeout than ordinary request/response calls, given the same 30s value is shared by `_request` and `_stream`?
2. The `except URLError` here only wraps the initial `urlopen()` call. A timeout raised later, from `resp.read()` inside the loop (lines ~117 onward), isn't caught and doesn't get wrapped into `LibreFangError` — it would surface as a raw `socket.timeout`/`TimeoutError` instead of the SDK's normal, catchable error type. Worth confirming that's the intended failure mode for callers.

Same shape exists in the codegen template at `scripts/codegen-sdks.py:212` (and its read loop below), since it's generating this exact code.

---

sdk/python/librefang/librefang_client.py
```diff
@@ -36,8 +38,9 @@ def __init__(self, client: "LibreFang"):
 class LibreFang:
     """LibreFang REST API client. Zero dependencies — uses only stdlib urllib."""
 
-    def __init__(self, base_url: str, headers: Optional[Dict[str, str]] = None):
+    def __init__(self, base_url: str, headers: Optional[Dict[str, str]] = None, timeout: float = DEFAULT_TIMEOUT):
```
houko, line 42: This 30s default is applied uniformly to every request path, including `_Resource` methods that call `self._c._request(...)` synchronously — e.g. `send_message()` at line 268, which maps to `POST /agents/{id}/message` (`crates/librefang-api/src/routes/agents/mod.rs:100-101`). That route blocks until the full agent turn completes: no SSE, no keep-alive, nothing to refresh the socket while the LLM/tool loop runs. CLAUDE.md's own live-verification example treats this exact call as a real, potentially slow round-trip (`curl -X POST .../message` with no `--max-time`), and agent turns with tool calls routinely run well past 30s.

By contrast, the SSE paths (`/message/stream`, `/sessions/{id}/stream`, etc.) all emit a 15s `KeepAlive` (`messaging.rs:649-652`, `sessions.rs:760-762`, `openai_compat.rs:691-693`, `network.rs:2001-2003`, `logs.rs:145-147`), so a 30s per-read socket timeout there has 2x margin and won't fire during a healthy long-running stream — that part of the default is sound.

The synchronous `send_message()` path has no such safety net: any real agent turn over 30s will now hard-fail with a timeout it never could before. Given the SDK's own usage docstring already steers callers toward `send_message_stream`, is that the intended mitigation (make the blocking call "fast path only" by convention), or should `send_message`/`_request` get a longer default — or its own override — separately from the stream-open timeout? Worth a deliberate call rather than sharing one constant across both call shapes by default.

---

houko (commented):
Automated review pass. Two correctness bugs in the new timeout handling were fixed directly (pushed as d25e4717 and 94fa735c): `urlopen(..., timeout=X)` raises a bare timeout exception (not `URLError`/`HTTPError`) exactly when the response is slow — the case this PR exists to bound — so it was leaking past the SDK's own `LibreFangError` wrapping; and the first fix used `except TimeoutError`, which doesn't catch what's actually raised on the Python 3.8/3.9 this SDK still declares support for (`socket.timeout` only became an alias of `TimeoutError` in 3.10), so it's now `except socket.timeout`. Left one inline comment on the 30s default's interaction with the synchronous (non-streaming) `send_message` path, which has no SSE keep-alive to lean on — that one needs an author call, not a mechanical fix.

---

sdk/python/librefang/librefang_client.py
```diff
@@ -101,10 +107,12 @@ def _stream(self, method: str, path: str, body: Any = None, query: Optional[Dict
         headers["Accept"] = "text/event-stream"
         req = Request(url, data=data, headers=headers, method=method)
         try:
-            resp = urlopen(req)
+            resp = urlopen(req, timeout=self.timeout)
```
houko, line 110: `urlopen(req, timeout=self.timeout)` sets the socket timeout for the whole connection object, not just the initial connect.
Once `resp` is returned, the `while True: resp.read(4096)` loop below (lines 119-148) inherits that same 30s timeout as an inactivity bound on every subsequent read, not only on opening the connection.
For a genuinely long-lived SSE stream — `send_message_stream` waiting on a slow LLM turn, or `attach_session_stream` / `comms_events_stream` sitting idle between events — a gap of more than `self.timeout` seconds between chunks now raises `socket.timeout` mid-generator.
That exception is not caught anywhere in this method: the only `except socket.timeout` here (lines 114-115) wraps the initial `urlopen()` call, not the read loop at 119-148, so it propagates as a raw stdlib `socket.timeout` rather than the `LibreFangError` the changelog for this PR says callers can now rely on.
This is the connect-timeout-vs-read-timeout distinction stdlib `urllib` doesn't give you for free: one `timeout=` value governs both, and the fix likely needs either a separate (larger, or `None`) timeout applied after the connection opens for the streaming path, or an explicit catch/wrap of `socket.timeout` around the read loop with a clear "stream stalled" message.
Right now a caller who previously got an SSE stream that just blocked through a slow turn will instead see it die partway through.
Same shape mirrored in `scripts/codegen-sdks.py`'s `_stream` template (this is generator code, so both copies need whatever fix lands here).
Separately: `_request()` shares the same flat 30s `DEFAULT_TIMEOUT` for every non-streaming endpoint, including ones that plausibly block on a full agent turn (e.g. `send_message`, `run_workflow`, `catalog_update`) — worth confirming 30s is actually long enough for those call sites rather than just fast health-check-style endpoints, since this default applies uniformly across the whole generated client.

---

houko (commented):
Automated review: the timeout coverage in this PR is consistent across the codegen template and committed client, and the changelog.d fragment is present and correctly attributed. One design concern flagged inline on the `_stream` read loop re: connect vs. read timeout semantics for long-lived SSE connections.

---

sdk/python/librefang/librefang_client.py
```diff
@@ -101,10 +107,12 @@ def _stream(self, method: str, path: str, body: Any = None, query: Optional[Dict
...
         except HTTPError as e:
             body_text = e.read().decode() if e.fp else ""
             raise LibreFangError(f"HTTP {e.code}: {body_text}", e.code, body_text) from e
+        except socket.timeout as e:
```
houko, line 114: This wraps only the `urlopen()` call that opens the SSE connection.
`urlopen(req, timeout=self.timeout)` sets the timeout on the underlying socket for its whole lifetime, not just the connect phase, so a stall in the read loop further down (`while True: chunk = resp.read(4096)`, a few lines below this method's `finally`) raises the same `socket.timeout` — but that loop sits in a separate `try/finally` with no `except` clause, so the exception propagates raw instead of becoming a `LibreFangError`.
That reproduces the exact failure mode this PR's own commit message describes fixing ("a slow real agent turn would leak a raw stdlib TimeoutError instead of the SDK's own LibreFangError"), just one step later in the same method.
In practice the server sends an SSE keep-alive every 15s (`crates/librefang-api/src/routes/agents/messaging.rs`), comfortably under the 30s default, so this likely doesn't fire under normal conditions — but a proxy that buffers/drops the keep-alive comment, or a caller who lowers the timeout, would hit it.
None of the new tests in `test_generated_client_timeout.py` exercise a timeout during the read loop, only during connection-open, so this path is untested either way.
Worth wrapping the `resp.read(4096)` calls (both the loop body and the trailing-buffer flush) the same way, in both this file and the matching spot in `scripts/codegen-sdks.py`.

---

scripts/codegen-sdks.py
```diff
@@ -206,10 +212,12 @@ def _stream(self, method: str, path: str, body: Any = None, query: Optional[Dict
...
         except HTTPError as e:
             body_text = e.read().decode() if e.fp else ""
             raise LibreFangError(f"HTTP {e.code}: {body_text}", e.code, body_text) from e
+        except socket.timeout as e:
```
houko, line 219: Same gap as the generated client (see the matching comment on `sdk/python/librefang/librefang_client.py`): this only wraps the connection-open `urlopen()` call, not the `resp.read(4096)` calls in the read loop further down, even though the socket-level timeout from `timeout=self.timeout` applies to both.
Please fix here too so the template and the generated output stay in sync, as the rest of this PR already does.

---

houko (commented):
Automated CLAUDE.md compliance pass over this PR's diff (checked against `librefang/librefang`'s repo-root CLAUDE.md): changelog fragment format, commit message conventions, no-AI-attribution, and prose-wrapping rules all check out cleanly, and the timeout-wrapping fix itself is sound for the ordinary request path. One gap found in the SSE streaming path, left as inline comments below rather than a direct fix since the correct remediation (what to do with a partially-buffered event when a mid-stream read times out) is a design call, not a one-line patch.

---

sdk/python/librefang/librefang_client.py
```diff
@@ -86,6 +90,8 @@ def _request(self, method: str, path: str, body: Any = None, query: Optional[Dic
...
+            raise LibreFangError(f"Request timed out after {self.timeout}s") from e
         except URLError as e:
             raise LibreFangError(f"Connection error: {e.reason}") from e
```
houko, line 96: Minor inconsistency, not a bug: a timeout during connect (before headers arrive) surfaces here as `URLError` — `except socket.timeout` never fires for it, because `urllib.request.AbstractHTTPHandler.do_open()` catches the `OSError` raised by `h.request()` during connect and re-wraps it as `URLError(err)` before it ever reaches this method's `except` clauses. So a connect-phase timeout gets `LibreFangError("Connection error: timed out")`, while a header-wait or body-read timeout gets `LibreFangError("Request timed out after {timeout}s")` — same exception type and same underlying cause (a timed-out socket), different message shape. Worth a deliberate call on whether that's the intended UX (connect failures grouped with DNS/refused-connection under "Connection error", timeouts elsewhere getting their own wording) or whether connect-phase timeouts should also get the "timed out after Xs" phrasing by checking `isinstance(e.reason, socket.timeout)` in the `except URLError` branch. Same pattern applies to the `_stream` open path a few lines down.

---

houko (commented):
Pushed one fix directly to this branch: `_stream()` caught `socket.timeout` around the initial `urlopen()` call but not around the body-read loop that follows, so a stream that connects fine and then goes quiet mid-body (the "stalled socket read" case this PR's own changelog fragment names) leaked a raw `socket.timeout` past the SDK's error contract instead of raising `LibreFangError` like every other timeout and connection failure. Fixed in both the codegen template and the generated client, with a regression test and a codegen self-test assertion to keep the two in sync (commit c5910688).

Two things I did not touch because fixing them means rewriting already-pushed history, which is outside what I'll do without asking first:

1. Three commits on this branch (`e29da9534f`, `0fcbc7dea0`, and the merge commit `6b818bc9a3`) have author/committer identity `Claude <noreply@anthropic.com>`. CLAUDE.md's commit-msg hook is explicitly written to reject exactly this ("rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean"), so these three predate the hook being enabled for whoever made them, or were made through a path the hook doesn't see. Squashing or re-authoring those three before merge would bring the branch into line with the repo's own attribution policy.
2. Commits `94fa735c` and `d25e4717` have hard-wrapped commit bodies (fixed ~68-70 char columns, multiple sentences per line) rather than one sentence per line, which is the prose convention CLAUDE.md states applies to "commit message bodies" specifically. Interestingly, commit `e29da9534f` on this same branch already fixed the identical issue in the test comments added earlier in the branch, so the convention was clearly known — it just didn't get applied to these two commit messages themselves.

One inline comment left on `librefang_client.py` about a minor message-wording inconsistency between connect-phase timeouts and read-phase timeouts — not a bug, just worth a deliberate call.

---

changelog.d/fixed/6823-python-sdk-request-timeout.md
houko, line 1: The timeout fix itself looks solid: both `urlopen` call sites in `_request` and `_stream` now pass `timeout=self.timeout`, the same socket timeout also bounds body reads (covered by the new mid-stream `socket.timeout` test), the 30s default is reasonable, the constructor override works, and the new tests assert an actual raised/wrapped timeout rather than just parameter pass-through. `scripts/tests/test_codegen_sdks.py` and the generated client are kept in sync. No missed call site.

One policy issue in the branch history, though: three commits reachable from this PR's head are authored as `Claude <noreply@anthropic.com>`, not a human identity:
- `6b818bc9a` — "Merge remote-tracking branch 'origin/main' into pr-6823"
- `0fcbc7dea` — "fix(python-sdk): accept the new urlopen timeout kwarg in existing test mocks"
- `e29da9534` — "docs(python-sdk): rewrap new test comments to one sentence per line"

CLAUDE.md's "No AI / Claude attribution" rule and the `commit-msg` hook both check *author identity* (`GIT_AUTHOR_IDENT`), not just the commit message text, specifically to catch this case. These commits are already pushed, so cleaning them up needs a rebase + force-push of `fix/python-sdk-request-timeout`, which is outside what an automated reviewer should do unilaterally — flagging for a maintainer to rewrite the author identity (or squash-merge, which would erase it from `main`) before merge.

---

scripts/codegen-sdks.py
houko, line 1: Not a defect in this PR — it's correctly scoped to the Python client per the title and changelog — but worth a conscious decision: `gen_go` and the JS/TS generator in this same file still build their HTTP clients with no timeout at all (the Rust generator already has `DEFAULT_CONNECT_TIMEOUT` / `DEFAULT_REQUEST_TIMEOUT`, so it's just Go and JS left exposed to the same unbounded-request risk this PR fixes for Python). Since it's the same file and the same underlying gap, wanted to surface it rather than silently defer: is a follow-up issue wanted for Go/JS, or is that intentionally out of scope for now?

---

houko (commented):
Reviewed the timeout fix itself and it's correct and well-tested (see inline notes for the full breakdown): every `urlopen` call site in both the codegen template and the generated client is bounded, the same socket timeout covers connect and stalled reads (not just connection setup), the 30s default is reasonable, it's overridable per-client, the codegen self-test and generated client stay in sync, and the new tests genuinely exercise a raised timeout rather than just checking the parameter is forwarded. Ran the affected suites locally (`test_generated_client_timeout.py`, `_stream.py`, `_stream_utf8.py`, `_trailing_sse.py`, `_urlerror.py`, and `scripts/tests/test_codegen_sdks.py`) — all pass. Changelog fragment is present and correctly formatted.

Two things flagged inline for a maintainer decision rather than fixed automatically:
1. Three commits in this branch's history carry `Claude <noreply@anthropic.com>` as the author identity — a policy violation the repo's own `commit-msg` hook is designed to catch, but already pushed so fixing it needs a history rewrite I won't force-push unilaterally.
2. Go/JS generated SDKs in the same script still have no request timeout at all (Rust already does) — likely worth a follow-up, your call on scope.

No functional issues found with the timeout implementation itself.

---

PR #6856, opened 2026-08-09
fix(telegram): balance self-closing tags

houko:
## Summary
- detect self-closing allowed Telegram HTML tags
- normalize them to empty balanced tag pairs
- prevent their formatting scope from leaking into following text

## Verification
- `cargo fmt --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml -- --check`
- `git diff --check`


changelog.d/fixed/6856-telegram-self-closing-sanitize.md
```diff
@@ -0,0 +1 @@
+- Prevented self-closing Telegram HTML tags from wrapping all following text during sanitization. (@houko)
```
houko: changelog.d/README.md: not tooling-enforced, but ending the bullet with `(#6856)` before `(@houko)` avoids `cargo xtask release` generating a duplicate `- <PR title> (#6856) (@houko)` line for this same PR in the release notes.

---

houko (commented):

---

sdk/rust/librefang-sidecar-telegram/src/format/sanitize.rs
```diff
@@ -142,6 +142,7 @@ pub fn sanitize_telegram_html(text: &str) -> String {
             .unwrap_or(false);
         let tag = captures.get(2).unwrap().as_str().to_ascii_lowercase();
         let attrs = captures.get(3).map(|s| s.as_str()).unwrap_or("");
+        let self_closing = attrs.trim_end().ends_with('/');
```
houko, line 149: Edge case in the self-closing heuristic: `let self_closing = attrs.trim_end().ends_with('/');` treats any attrs string ending in `/` as self-closing.
For quoted attribute values this is safe (the capture keeps the closing quote as the last char), but for an *unquoted* value that ends right before `>` — e.g. `<a href=http://example.com/>` — the attrs capture is the raw `[^>]` char run, so it ends in `/` too and the tag gets misclassified as self-closing.
The `<a>` then closes immediately and the following text is emitted unwrapped, which is a silent behavior change rather than a crash, so I didn't fix it inline — flagging for your judgment on whether unquoted attribute values are a case worth guarding against here.

---

houko (commented):
Mechanical CLAUDE.md compliance pass. Fixed the changelog fragment format (pushed). One low-confidence edge case left as an inline comment for maintainer judgment. No AI/Claude/Anthropic attribution or authorship issue found on this branch.

---

sdk/rust/librefang-sidecar-telegram/src/format/sanitize.rs
houko, line 149: Nit / edge case worth a second look: `attrs.trim_end().ends_with('/')` will also fire for an *unquoted* attribute value that simply happens to end in `/` right before `>`, e.g. `<a href=https://example.com/>text</a>`. Since `RE_ATTR` only matches quoted values, the unquoted `href` is dropped either way, but this tag would now be treated as self-closing (immediately emit `<a></a>`) rather than staying open through `text</a>`. That's a behavior change from before this PR for a case that isn't actually a self-closing tag. Probably rare in practice (LLM output tends to quote attributes), but worth confirming it's the intended tradeoff rather than an oversight — a check like `attrs.trim_end().ends_with("\"/") || attrs.trim_end() == "/"` (or similar quote-aware check) would avoid the false positive.

---

sdk/python/librefang/sidecar/adapters/telegram.py
```diff
@@ -541,9 +541,14 @@ def _rebuild_safe_tag(tag_name: str, attrs_raw: str, self_closing: bool):
...
-    if self_closing:
-        buf += "/"
     buf += ">"
+    if self_closing:
+        # Telegram's HTML subset has no self-closing-tag syntax: emitting
+        # a literal `<tag/>` would either be rejected by the Bot API's
+        # "Unclosed start tag" check or (if tolerated) leave the tag open
+        # for the rest of the message, matching telegram.rs's fix — close
+        # it immediately instead of leaking the marker into the output.
+        buf += "</" + tag_name + ">"
```
houko, line 555: The self-closing detection this feeds from is `self_closing = tag_content.endswith("/")` (line 600), which has the same unquoted-attribute false positive already flagged on the Rust side (`attrs.trim_end().ends_with('/')` in sanitize.rs): `<a href=http://example.com/>text</a>` ends in `/` right before `>`, so it gets misclassified as self-closing here too and `<a>` closes immediately instead of wrapping `text`.
Since `RE_ATTR`'s Rust counterpart only keeps quoted values anyway, the practical exposure is the same low-probability case noted on the Rust file, but it exists in both language ports now rather than just one, so a fix (if one lands) should probably touch both files together.

---

houko (commented):
Reviewed the two commits added since the last pass (9cda94a9 Python self-closing fix, 9f67f88f Python crossed-tags fix) — both correctly mirror the Rust sanitizer's existing stack-drain and self-closing-pair behavior, and the new tests cover the same shapes as the Rust test suite. One inline note on a pre-existing detection heuristic that now exists in both ports.

---

sdk/python/librefang/sidecar/adapters/telegram.py
houko, line 1: **Blocking: AI attribution in commit author identity (CLAUDE.md hard violation).**

The HEAD commit `9f67f88f` ("fix(telegram): close crossed tags innermost-first in Python sanitizer") has author identity `Claude <noreply@anthropic.com>`:

```
author: Claude <noreply@anthropic.com>
date:   2026-08-10T17:17:49Z
```

CLAUDE.md requires no AI/Claude/Anthropic attribution anywhere, including author identity (the repo's `commit-msg` hook checks `GIT_AUTHOR_IDENT` for exactly this). This needs a human-driven history rewrite (interactive rebase to fix the author identity, or a fixup + squash) — not a force-push from this automated pass. Please rebase this commit onto the correct human author identity before merge.

---

houko (commented):
Automated daily review pass.

**Blocking:** HEAD commit `9f67f88f` has author identity `Claude <noreply@anthropic.com>` — hard CLAUDE.md violation, flagged inline. Needs a human-driven rebase to fix the author identity; not force-pushing from this pass.

**Sanitizer logic (self-closing tag detection):** the fix correctly balances genuinely self-closing tags (`<code/>`, `<tg-emoji .../>`) into open/close pairs on both the Rust (`sanitize.rs`) and Python (`telegram.py`) sides, and the added tests cover the nested and adjacent self-closing cases plus the crossed-tag stack-drain case.
One known gap remains open and unresolved from three prior review threads on this PR (`sanitize.rs:145`, `sanitize.rs:145` again, and `telegram.py:551`): `attrs.trim_end().ends_with('/')` (and its Python mirror `tag_content.endswith("/")`) misclassifies a tag with an *unquoted* attribute value ending in `/` right before `>` — e.g. `<a href=http://example.com/>text</a>` — as self-closing, closing `<a>` immediately instead of wrapping `text`.
`RE_ATTR` only captures quoted values, so the practical blast radius is a formatting regression (link loses its href/formatting) rather than a sanitizer bypass — no unbalanced/malformed HTML escapes.
Not re-flagging as a new comment since it's already tracked three times unresolved; noting only that the regression-test suite added in this PR still doesn't include the "could be mistaken for self-closing" case the task calls for (only the genuine self-closing cases are tested), so that gap is worth closing alongside the fix if/when it lands.

**Changelog:** all three new fragments (`6856-telegram-python-sanitizer-crossed-tags.md`, `6856-telegram-self-closing-sanitize-python.md`, `6856-telegram-self-closing-sanitize.md`) are present, correctly sectioned under `fixed/`, and correctly end `(#6856) (@houko)` — no format issues.



---

sdk/python/librefang/sidecar/adapters/telegram.py
houko, line 555: Rust/Python parity gap in the self-closing detection this feeds, verified by direct comparison (not the already-flagged unquoted-href case).

Rust trims trailing whitespace before checking for the marker: `sanitize.rs:145` is `attrs.trim_end().ends_with('/')`.
The Python detector at `telegram.py:600` does not: `self_closing = tag_content.endswith("/")`, checked against the raw `tag_content` with no trim.

So a self-closing tag with whitespace between the slash and `>` — `<tag/ >`, valid HTML — is handled correctly in Rust but not in Python:

```
sanitize_telegram_html("<b/ >after")
# Rust:   "<b></b>after"   (correct — matches the fix's intent)
# Python: "<b>after</b>"   (wrong — "after" gets wrapped instead of the tag closing immediately)
```

Reproduced against current HEAD:
- Rust (`cargo test -p librefang-sidecar-telegram`, ad hoc probe): `<b/ >after` → `<b></b>after`, `<code/ >after` → `<code></code>after`.
- Python (`sdk/python`, direct call): `<b/ >after` → `<b>after</b>`, `<code/ >after` → `<code>after</code>`.

This sits squarely in the code this PR is fixing (self-closing-tag handling), so a message containing a trailing-whitespace self-closing tag will silently regress to the pre-fix "wraps all following text" bug in the Python sidecar while the Rust sanitizer handles it fine — the exact divergence the two ports are supposed to avoid per the module docstring.

Suggested fix: mirror the Rust side and strip trailing whitespace before checking, e.g. `self_closing = tag_content.rstrip().endswith("/")`, and make sure `attrs_raw` slicing still lands on the right bounds afterward.

---

houko (commented):
Automated correctness pass on the self-closing-tag balancing logic. Verified the core stack-drain / self-closing fix (Rust + Python) against the existing tests plus ad hoc probes — nested, crossed, and adjacent self-closing cases all behave correctly and both scoped test suites pass. One new Rust/Python parity gap found and detailed inline (trailing-whitespace self-closing tags, e.g. `<tag/ >`). Not re-raising the unquoted-attribute-value edge case already flagged on this PR's Rust and Python threads, or the commit-author-identity issue already flagged separately, to avoid duplicate noise on points already on record.

---

changelog.d/fixed/6856-telegram-self-closing-sanitize.md
houko, line 1: Automated review note (no code issues found — Python sanitizer's stack-drain now matches `sanitize.rs`, all 66 crate tests pass, relevant Python test passes, clippy/fmt clean on the current diff).

Flagging a repo-policy issue for the maintainer before merge: this PR's current HEAD commit `9f67f88f7e4b9a47e6bb4e95c33276a55aa46502` ("fix(telegram): close crossed tags innermost-first in Python sanitizer") is authored by `Claude <noreply@anthropic.com>`.
CLAUDE.md's commit-msg hook is specifically designed to reject a commit whose author identity resolves to Claude/Anthropic, so this slipped through with hooks not applied (or bypassed) at commit time.
Fixing the author on an already-pushed HEAD commit needs a rebase + force-push, which is outside what an automated reviewer should do unilaterally — leaving this for a maintainer to reauthor before merge.

---

houko (commented):

---

PR #6861, opened 2026-08-09
fix(telegram): default-deny missing senders

houko:
## Summary
- centralize Telegram sender allowlist decisions in one fail-closed helper
- reject sender-less updates when `ALLOWED_USERS` is restricted
- preserve permit-all compatibility when the allowlist is empty
- add regression coverage for restricted missing, open missing, and matching sender cases

## Verification
- `git diff --check`

## Out of scope
- sender extraction shape and Telegram update translation remain unchanged; other adapter diagnostics/shutdown findings are separate PRs.


houko:
⚠️ Compliance flag (not a code-quality note): commit `54e795e6` on this branch is authored as `Claude <noreply@anthropic.com>`. CLAUDE.md's `commit-msg` hook exists specifically to reject a commit whose author identity resolves to Claude/Anthropic, and the repo's git conventions explicitly disallow AI attribution — this slipped through because `core.hooksPath` isn't set by default in a fresh clone, so the version-controlled hook never ran for whatever session made this commit. I can't fix this myself: rewriting the offending commit needs a rebase + force-push, which I'm not authorized to do unilaterally. Flagging for a maintainer to rebase with `--reset-author` (or squash before merge, which will attribute the squashed commit to whoever GitHub picks — typically the PR author).

---

changelog.d/fixed/6861-telegram-missing-sender-allowlist.md
```diff
@@ -0,0 +1 @@
+- Default-denied Telegram updates without an identifiable sender whenever `ALLOWED_USERS` restricts access. (@houko)
```
houko: changelog.d/README.md: not tooling-enforced, but ending the bullet with `(#6861)` before `(@houko)` avoids `cargo xtask release` generating a duplicate `- <PR title> (#6861) (@houko)` line for this same PR in the release notes.

---

houko (commented):

---

sdk/rust/librefang-sidecar-telegram/src/adapter.rs
```diff
@@ -45,6 +45,15 @@ macro_rules! tg_trace {
     ($($arg:tt)*) => { trace(format_args!($($arg)*)) };
 }
 
+/// Apply the configured identity boundary to an extracted Telegram sender.
+/// Missing identity is compatible only with an explicitly open allowlist; restricted deployments fail closed instead of relying on the translator to reject every sender-less update variant independently.
```
houko, line 49: Commit `54e795e6a0266e5506fc53053d87af3a7c27a8da` ("docs: rewrap sender_passes_allowlist doc comment to one sentence per line") has author identity `Claude <noreply@anthropic.com>`. Per CLAUDE.md ("No AI/Claude/Anthropic attribution... the `commit-msg` hook... separately rejects a commit whose author identity... resolves to Claude/Anthropic"), this needs to be rebased/re-authored before merge. The commit's *content* is fine (a legitimate prose-wrap fix, no AI mention in the message body) — only the author identity needs fixing. I can't rewrite history on this branch myself, so flagging for a maintainer rebase (`git rebase -i` reauthoring that commit, or `git commit --amend --author`) ahead of merge.

---

houko (commented):
Mechanical CLAUDE.md compliance pass. Pushed one small fix (changelog fragment format); flagged one item that needs a maintainer rebase (can't fix myself).

---

sdk/rust/librefang-sidecar-telegram/src/adapter.rs
houko, line 49: Non-blocking process note, not a code issue: commit `54e795e6` ("docs: rewrap sender_passes_allowlist doc comment to one sentence per line", the doc-comment rewrap that touches this line) is authored as `Claude <noreply@anthropic.com>`. CLAUDE.md's "No AI / Claude attribution in commit messages" rule (and the repo's `commit-msg` hook) applies to author identity as well as message text. This mirrors the same issue flagged on sibling PRs in this review campaign (#6901, #6887, #6880, #6881, #6882, #6868). Not rewriting history here per review policy — flagging for the maintainer to decide (e.g. squash-merge with a clean identity, which is the usual fix for this pattern).

Separately, on the substance of the fix: reviewed `sender_passes_allowlist` end to end against `AllowList::permits`/`is_open` (`access.rs`) and the single call site in `produce()`. The pre-fix bug is real and exactly as described — the old `if let Some(sender) = sender { … }` had no `else` branch, so a `None` sender (reachable via `callback_query.from` or `poll_answer.user` being absent, which `#[serde(default)]` on `CallbackQuery`/`PollAnswer` makes a legitimate deserialization outcome, not just synthetic) fell through to `update_to_event` and got emitted regardless of `ALLOWED_USERS`. The fix closes that: `None` now maps to `allowlist.is_open()`, so a restricted allowlist fails closed. No type-confusion path found — `permits()` does exact string matching against `raw_ids`/`usernames_lower`, and the pre-existing synthetic-sender fallback in `extract_sender` (message/edited_message with neither `from` nor `sender_chat`, yielding `user_id: "0"`) still routes through `permits()`, not the new `is_open()` branch, so it doesn't gain any new bypass from this change. The added test covers both directions (missing sender denied under a restricted list, allowed sender still passes). Checked the Python port (`sdk/python/librefang/sidecar/adapters/telegram.py`): `_sender`/`_callback_to_event`/`_poll_answer_to_event` all unconditionally `return None` when sender extraction fails, before the allowlist check even runs — so Python was never exposed to this bug and needs no equivalent fix here. `cargo check -p librefang-sidecar-telegram`, `cargo clippy -p librefang-sidecar-telegram --all-targets -- -D warnings`, and `cargo test -p librefang-sidecar-telegram` (63 passed) are all clean on this branch.

---

sdk/rust/librefang-sidecar-telegram/src/adapter.rs
houko, line 1: Compliance status (consolidated — already flagged 3x on this thread: issuecomment-5239125604, discussion_r3748946615, discussion_r3750096577): commit `54e795e6a0266e5506fc53053d87af3a7c27a8da` ("docs: rewrap sender_passes_allowlist doc comment to one sentence per line") is still authored as `Claude <noreply@anthropic.com>` as of the current head. Unresolved; needs a maintainer rebase/re-author before merge per CLAUDE.md's author-identity rule. No new flag needed, not repeating full detail.

---

sdk/rust/librefang-sidecar-telegram/src/adapter.rs
```diff
@@ -45,6 +45,15 @@ macro_rules! tg_trace {
...
+fn sender_passes_allowlist(allowlist: &AllowList, sender: Option<&Sender>) -> bool {
```
houko, line 50: Substance question the prior review pass didn't check, worth a maintainer look before trusting the `security/` classification as-is: does the `None`-sender branch here (`allowlist.is_open()`) ever actually change what reaches the agent, given `translator.rs`?

`produce()` only produces a `None` sender (line 475) for `callback_query` when `cq.from` is absent, or for `poll_answer` when `pa.user` is absent — `extract_sender()` for `message`/`edited_message` always returns `Some(Sender{..})`, even in its own "0"/"Unknown" fallback case, so that path is unaffected by this change.

But both of those exact conditions already short-circuit to `None` one layer down, pre-existing and untouched by this PR:
- `translator.rs:496` — `callback_event`: `let user = cq.from.as_ref()?;`
- `translator.rs:567` — `poll_answer_event`: `let user = pa.user.as_ref()?;`

So pre-fix, when `sender` was `None` here and the old code fell through to `update_to_event(...)` unchecked, `update_to_event` would call into `callback_event`/`poll_answer_event`, which themselves return `None` for the exact same missing-field condition — no `Value` gets built, `emit()` is never called, nothing reaches the agent. That means the described bypass ("a `None` sender... fell through... and got emitted regardless of `ALLOWED_USERS`") doesn't look reproducible against current `main`/this PR's base for the `callback_query`/`poll_answer` cases, since it's the only way `sender` becomes `None` here.

The change is still correct defense-in-depth (protects against a future edit to `callback_event`/`poll_answer_event` that drops their own guard, and it does change one real thing pre- vs post-fix: for a missing-sender `callback_query`, pre-fix still reached `update_to_event`'s `tokio::spawn(answer_callback_query(...))` ack call before failing inside `callback_event`; post-fix that call is skipped by the earlier `continue`), but as far as I can trace it, it isn't closing a reachable "unidentified sender reaches the agent" hole today. Worth the author/maintainer double-checking this reachability claim (maybe there's a code path I'm not seeing, e.g. via a future/parallel PR touching `translator.rs`) before relying on the `security/` changelog wording ("previously skipped the ALLOWED_USERS check entirely") as accurate for the current tree. Not asking for the auth logic itself to change — the fix is safe to keep either way.

---

houko (commented):
Automated daily review pass.

Summary:
- Author-identity compliance: unresolved, already flagged 3x — consolidated status posted, no new flag.
- Changelog fragment: now correctly filed under `changelog.d/security/6861-telegram-missing-sender-allowlist.md`, correct `(#6861) (@houko)` tail — the earlier format concern is resolved as of commit `fbfa63e`/`88f75bf`.
- Access-control substance: `sender_passes_allowlist` in `adapter.rs` is the single dispatch point for inbound Telegram updates (message, edited_message, callback_query, poll_answer all flow through the one `produce()` loop before `emit()`), and no sentinel/trusted-default bypass was found beyond the pre-existing, unchanged, out-of-scope "0"/"Unknown" fallback in `extract_sender`. One reachability question raised inline for maintainer judgment: whether the `None`-sender branch changes any actually-emitted event given `translator.rs`'s own pre-existing guards, which would bear on how the `security/` changelog entry is worded. Not asking to change the auth logic.
- Regression test (`missing_sender_is_denied_unless_allowlist_is_open`) covers both the deny case and the allowed-sender case.




---

PR #6866, opened 2026-08-09
fix(telegram): honor retry-after headers

houko:
## Summary
- capture `Retry-After` before consuming each JSON or multipart HTTP 429 response
- prefer valid delta-seconds from the header, then Telegram JSON `parameters.retry_after`, then the existing 5-second default
- preserve the 300-second flood-wait cap and one-retry behavior
- cover valid header, invalid-header body fallback, and non-JSON default fallback

## Verification
- `git diff --check`

## Out of scope
- HTTP-date form of the generic RFC header is not emitted by Telegram and remains unsupported; multipart timeout sizing and retry-loop panic hardening are separate findings.


houko:
⚠️ Compliance flag (not a code-quality note): commit `c9d555ff` on this branch is authored as `Claude <noreply@anthropic.com>`. CLAUDE.md's `commit-msg` hook exists specifically to reject a commit whose author identity resolves to Claude/Anthropic, and the repo's git conventions explicitly disallow AI attribution — this slipped through because `core.hooksPath` isn't set by default in a fresh clone, so the version-controlled hook never ran for whatever session made this commit. I can't fix this myself: rewriting the offending commit needs a rebase + force-push, which I'm not authorized to do unilaterally. Flagging for a maintainer to rebase with `--reset-author` (or squash before merge, which will attribute the squashed commit to whoever GitHub picks — typically the PR author).

---

changelog.d/fixed/6866-telegram-retry-after-header.md
```diff
@@ -0,0 +1 @@
+- Honored delta-seconds from Telegram HTTP `Retry-After` headers before falling back to the JSON body or default backoff. (@houko)
```
houko: changelog.d/README.md: not tooling-enforced, but ending the bullet with `(#6866)` before `(@houko)` avoids `cargo xtask release` generating a duplicate `- <PR title> (#6866) (@houko)` line for this same PR in the release notes.

---

houko (commented):

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
```diff
@@ -534,10 +539,58 @@ impl BotClient {
     }
 }
```
houko, line 541: Commit `c9d555ff63a117e4af9fe4356348ce6cb7315a93` ("docs: rewrap resolve_retry_after doc comment to one sentence per line") on this branch has author identity `Claude <noreply@anthropic.com>`. Per CLAUDE.md's no-AI-attribution policy, this needs to be rebased/re-authored to a human identity before merge. I did not attempt to fix this myself (would require history rewrite/force-push) — flagging for @houko to handle via `git rebase -i` + `--reset-author` (or an equivalent squash) and a normal push.

---

houko (commented):
Mechanical CLAUDE.md-compliance pass. Fixed and pushed a small changelog fragment format issue directly (commit 126bd2a5). One item requires maintainer action — see inline comment (commit author identity resolves to Claude/Anthropic and can't be fixed without a history rewrite).

---

houko:
This branch's history contains a commit authored as Claude <noreply@anthropic.com>: c9d555ff63a117e4af9fe4356348ce6cb7315a93 ("docs: rewrap resolve_retry_after doc comment to one sentence per line").
CLAUDE.md documents a commit-msg hook that rejects commits whose author identity resolves to Claude/Anthropic even when the message text is clean, and the repo's stated policy is no Claude/Anthropic attribution anywhere in commit history.
This predates my review pass and I'm not rewriting someone else's already-pushed history unilaterally, so flagging it here for reauthoring or squashing before merge.

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
houko, line 1: **Commit history: AI attribution present (flagging only, not rewriting history)**

Commit `c9d555ff` ("docs: rewrap resolve_retry_after doc comment to one sentence per line") in this branch's history is authored/committed as `Claude <noreply@anthropic.com>`. CLAUDE.md's "No AI / Claude attribution" rule (and the `commit-msg` hook, if `core.hooksPath` were configured for this push) applies to author identity as well as message content. This matches the pattern already flagged on sibling PRs in this campaign (#6901, #6887, #6880, #6881, #6882, #6868). Per this review's own instructions I'm not rewriting history — flagging for the maintainer to squash/reauthor before merge if desired.

---

houko, line 1: **Python SDK port gap: `Retry-After` header not read, and no upper cap on the sleep**

This PR's `resolve_retry_after` correctly prioritizes the HTTP `Retry-After` header (delta-seconds form) over the JSON body's `parameters.retry_after`, with a sane fallback chain and an existing `MAX_RETRY_AFTER_SECS` (300s) cap that rejects an absurd/malicious value instead of sleeping on it — verified via the new unit tests (`resolve_retry_after_prefers_header_then_body_then_default`, `numeric_retry_after_overflow_remains_over_the_flood_wait_cap`).

The Python port (`sdk/python/librefang/sidecar/adapters/telegram.py`) was not updated to match:
- `_extract_retry_after()` (line ~699) only reads `parameters.retry_after` from the JSON body — it never inspects the `Retry-After` HTTP header at all, so a proxy/CDN in front of Telegram that sets the header without a matching body value (or with a different value) is silently ignored.
- More importantly, there is **no cap analogous to `MAX_RETRY_AFTER_SECS`**. `_call_retrying()` (line ~830) and `_send_media_upload()` (line ~907) both do `time.sleep(_extract_retry_after(resp, RETRY_AFTER_DEFAULT_SECS))` unconditionally — a malicious or buggy upstream `retry_after` (e.g. `999999`) blocks that call for over 11 days with no ability to cancel, which is exactly the self-inflicted-DoS scenario the Rust side now explicitly guards against.

This is the same "Python port missing the Rust-side fix" pattern flagged on sibling PRs in this campaign (#6875, #6873, #6871, #6870, #6868). Leaving as a review comment rather than fixing directly: porting this correctly requires (a) plumbing the response headers through `_api_post`/`_multipart`/`_api_get` (which currently discard them entirely), and (b) a product decision on what the synchronous adapter should do when the cap is exceeded (raise instead of sleeping? return the un-retried response?) — that's a behavioral/design call for a maintainer, not a mechanical port.

---

houko (commented):
Automated review pass. Rust-side fix looks correct and well-tested (header-first / body-fallback / capped-default resolution, verified locally with `cargo check -p librefang-sidecar-telegram --bins`, `cargo clippy -p librefang-sidecar-telegram --all-targets -- -D warnings`, and the crate's own test suite — all green, 3/3 new tests passing). Two findings left as inline comments below (both left as comments, not fixes — one is a commit-metadata flag not a code change, the other needs a maintainer design decision): an AI-attributed commit already in this branch's history, and a Python-SDK parity gap (missing `Retry-After` header handling and no sleep cap, mirroring prior PRs in this campaign).

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
```diff
@@ -534,10 +539,58 @@ impl BotClient {
...
 
-/// Try to parse a Retry-After value out of a non-2xx 429 body, falling back to the default.
-fn resp_retry_after_default(body: &str) -> u64 {
+/// Resolve any 429 delay using HTTP delta-seconds first, then Telegram's JSON value, then the conservative local default.
+/// A syntactically numeric header that overflows `u64` is saturated so the caller's flood-wait cap rejects it; unsupported HTTP-date/non-numeric forms fall through to the JSON value.
+fn resolve_retry_after(header: Option<&HeaderValue>, body_retry_after: Option<u64>) -> u64 {
```
houko, line 544: Same-file overlap with #6867: both PRs branch from main@3362202 and edit this same retry-tail region of `client.rs`. Here you delete/rename `resp_retry_after_default` (base lines ~538-543) into `resolve_retry_after` / `body_retry_after`; #6867 replaces the `unreachable!()` calls immediately above (base lines ~528/155) and inserts `retry_loop_exhausted` right before this same function boundary. Each PR is green against its own stale base, but per CLAUDE.md's "Two green PRs can still break main together" guidance, whichever of the two merges second will very likely hit a real (overlapping-hunk) merge conflict here, not a clean auto-merge — please rebase onto the post-merge `main` and get a fresh CI run before merging, rather than relying on the current `mergeable_state: clean`, which is only relative to the shared stale base.

---

houko (commented):
Code itself looks correct and well-tested (header parsing, overflow saturation, whitespace trimming, precedence over body/default all covered). One cross-PR note below re: #6867 touching the same region.

---

sdk/python/librefang/sidecar/adapters/telegram.py
```diff
@@ -697,6 +703,20 @@ def _is_ogg_opus(data: bytes) -> bool:
...
+    producing a negative delay."""
+    if isinstance(body, dict):
+        header = body.get("_retry_after_header")
+        if isinstance(header, str):
+            stripped = header.strip()
+            if stripped and stripped.isascii() and stripped.isdigit():
+                return int(stripped)
     try:
         v = body if isinstance(body, dict) else json.loads(body)
         ra = v.get("parameters", {}).get("retry_after")
```
houko, line 714: The header path just above this rejects a negative/non-numeric `Retry-After` header before it can reach `time.sleep`, but the JSON-body fallback on the next line doesn't have the same guard: `return int(ra) if ra is not None else default` accepts a negative `parameters.retry_after` from the body as-is. That value flows into `time.sleep()` in both `_call_retrying` and `_send_media_upload`, and `time.sleep` raises `ValueError` on a negative argument, so a malformed or adversarial response body can crash the retry path instead of degrading to `default` the way the header case already does. Rust's equivalent field is `Option<u64>`, so a negative JSON value there simply fails to deserialize and falls back to the default — this Python path has no equivalent floor. Suggest treating a negative `ra` the same way the header case treats a malformed value: fall through to `default` instead of returning it directly. Not pushing a fix myself — this session's git identity resolves to `Claude <noreply@anthropic.com>` (the same problem already flagged on this branch's `c9d555ff`), so any commit from here would just add another instance of that.

---

houko (commented):
Everything already flagged in prior passes (changelog format, the c9d555ff author-identity issue, the #6867 same-file overlap, the Python parity gap that the later commits in this branch already closed) still stands as-is — no new action needed there. One new finding: the JSON-body fallback in `_extract_retry_after` doesn't reject a negative `retry_after`, unlike the header path right above it, which can turn into a `time.sleep(negative)` crash. Left as an inline comment rather than a fix, for the reason noted there.

---

sdk/python/librefang/sidecar/adapters/telegram.py
houko, line 1: AI-attribution status (consolidated — already flagged 2+ times on this thread, updating with new information): this branch now carries **three** commits authored as `Claude <noreply@anthropic.com>`, not one. `c9d555ff63a117e4af9fe4356348ce6cb7315a93` was already flagged in prior comments; `d9e6a7d54d106e90bf0e850097ec812c5ad7a3bb` and `f6dcb5d4f5b6f88f18be216b3565ea8b35dd29d6` (both landed 2026-08-10 17:30–17:33, after the last flag) were not previously called out and carry the same author identity. All three need reauthoring or squashing to a human identity before merge per CLAUDE.md's no-AI-attribution policy; not rewriting history unilaterally here.

---

houko (commented):
Automated daily review pass.

- Retry-after bounding/parsing (Rust): `resolve_retry_after` correctly prioritizes the HTTP `Retry-After` header (delta-seconds) over the JSON body's `parameters.retry_after`, saturates a syntactically-numeric-but-overflowing header to `u64::MAX` rather than panicking, and both call sites gate on the existing `MAX_RETRY_AFTER_SECS` cap before sleeping/retrying — no second, inconsistent bound was introduced; the pre-existing constant is reused as-is.
- Retry-after bounding (Python): the new `MAX_RETRY_AFTER_SECS = 300` mirrors the Rust cap's name and value and is checked in both `_call_retrying` and `_send_media_upload` before `time.sleep`, so an extreme `retry_after` skips the retry instead of stalling the process.
- One inline finding filed on this pass (see below): AI-attribution scope was wider than previously flagged.
- Not re-filing: the already-open, unresolved thread on `_extract_retry_after`'s JSON-body fallback (negative `parameters.retry_after` bypassing the header path's validation, reaching `time.sleep` unguarded) remains valid and outstanding — no need to duplicate it here.
- Changelog fragments (`changelog.d/fixed/6866-telegram-retry-after-header.md`, `-header-python.md`): correct format, one sentence per line, each ending `(#6866) (@houko)`.




---

PR #6867, opened 2026-08-09
fix(telegram): return retry exhaustion errors

houko:
## Summary
- replace both JSON and multipart retry-loop tail panics with ordinary `Error::Other` values
- include the Telegram method name in the diagnostic
- keep all current return/continue paths and one-retry behavior unchanged
- add a focused error-variant/message regression

## Verification
- `git diff --check`

## Out of scope
- Retry-After parsing and multipart timeout sizing are separate findings.


changelog.d/fixed/6867-telegram-retry-loop-error.md
```diff
@@ -0,0 +1 @@
+- Returned recoverable Telegram API errors if a retry loop ever exhausts instead of panicking and killing the sidecar. (@houko)
```
houko: changelog.d/README.md: not tooling-enforced, but ending the bullet with `(#6867)` before `(@houko)` avoids `cargo xtask release` generating a duplicate `- <PR title> (#6867) (@houko)` line for this same PR in the release notes.

---

houko (commented):

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
```diff
@@ -530,14 +528,37 @@ impl BotClient {
                 });
             }
         }
-        unreachable!("send_multipart loop body either returns or `continue`s; the for-2 range cannot exhaust")
+        Err(retry_loop_exhausted(method))
```
houko, line 536: Same-file overlap with #6866: both PRs branch from main@3362202 and edit this same retry-tail region of `client.rs`. Here you replace the `unreachable!()` calls (this one and the `call_json` one) and insert `retry_loop_exhausted` right above the still-present `resp_retry_after_default`; #6866 deletes/renames that exact function into `resolve_retry_after` / `body_retry_after` a few lines below. Each PR is green against its own stale base, but per CLAUDE.md's "Two green PRs can still break main together" guidance, whichever of the two merges second will very likely hit a real (overlapping-hunk) merge conflict here, not a clean auto-merge — please rebase onto the post-merge `main` and get a fresh CI run before merging, rather than relying on the current `mergeable_state: clean`, which is only relative to the shared stale base.

---

houko (commented):
Change itself is sound (the `for attempt in 0..2` loop body does provably always return or `continue` on every reachable path today, so this converts a currently-unreachable panic into a safety net — good defensive boundary-error handling, minimal helper, test covers the error shape). One cross-PR note below re: #6866 touching the same region.

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
houko, line 1: The new regression test (`exhausted_retry_loop_is_a_recoverable_method_error`) only unit-tests the extracted `retry_loop_exhausted("sendDocument")` helper in isolation — it never drives `call_json` or `send_multipart` through two failing attempts to reach the `Err(retry_loop_exhausted(method))` return site itself.

Worth double-checking reachability while we're here: `for attempt in 0..2` only ever `continue`s when `attempt == 0`, and on the `attempt == 1` iteration every branch (success, `parsed.ok` false, non-2xx status) ends in an explicit `return` — so the trailing `Err(retry_loop_exhausted(method))` after the loop looks unreachable through the current call sites, same as the `unreachable!()` it replaces.

If that reading is right, this is still a reasonable defensive-programming change (a future edit to the loop body that breaks the invariant now degrades to a graceful `Error::Other` instead of panicking the sidecar), but the PR description ("return retry exhaustion errors") and the test both read as if this path fires on a real double-429/failure sequence, which it currently cannot.

Suggest either:
- noting in the doc comment / PR body that this is a defensive fallback for an invariant that doesn't hold through any current caller, not a reachable retry-exhaustion case, or
- adding a mock-transport test that actually drives two consecutive failing responses through `call_json`/`send_multipart` to prove the error surfaces end-to-end (this would also give the invariant a real regression test if the loop bound or continue condition is ever changed).

The error-variant/message assertions themselves (`Error::Other` containing the method name and "retry loop exhausted") are fine and match the informativeness bar.

---

houko (commented):
Automated daily review pass.

Commit author identities checked (`get_commits`): all four commits are authored/committed as `Evan Hu` / `houko` <suzukaze.haduki@gmail.com> — no `Claude <noreply@anthropic.com>` identity found, so no blocking attribution finding this pass.

Changelog fragment (`changelog.d/fixed/6867-telegram-retry-loop-error.md`) is present and correctly formatted (one sentence, ends `(#6867) (@houko)`).

Diff review (`sdk/rust/librefang-sidecar-telegram/src/api/client.rs`): retry loop still terminates in at most 2 attempts on all paths — no infinite-loop risk introduced. One finding left as an inline comment on the test coverage / reachability of the new error path.



---

PR #6868, opened 2026-08-09
fix(telegram): scale multipart timeouts

houko:
## Summary
- derive a request-level multipart deadline from payload bytes at a conservative 64 KB/s floor
- retain the existing 30-second base budget for connection/setup/server response
- apply the same computed deadline to both upload attempts
- use checked/saturating conversion for extreme platform sizes

## Verification
- focused 0/63,999/64,000/64 MB timeout regression
- `git diff --check`

## Out of scope
- multipart buffer cloning and Retry-After behavior are separate findings.


houko:
⚠️ Compliance flag (not a code-quality note): commit `8568134b` on this branch is authored as `Claude <noreply@anthropic.com>`. CLAUDE.md's `commit-msg` hook exists specifically to reject a commit whose author identity resolves to Claude/Anthropic, and the repo's git conventions explicitly disallow AI attribution — this slipped through because `core.hooksPath` isn't set by default in a fresh clone, so the version-controlled hook never ran for whatever session made this commit. I can't fix this myself: rewriting the offending commit needs a rebase + force-push, which I'm not authorized to do unilaterally. Flagging for a maintainer to rebase with `--reset-author` (or squash before merge, which will attribute the squashed commit to whoever GitHub picks — typically the PR author).

---

changelog.d/changed/6868-telegram-multipart-bytes.md
```diff
@@ -0,0 +1 @@
+- Shared Telegram multipart upload storage across retry attempts instead of copying the full attachment on the happy path. (@houko)
```
houko: changelog.d/README.md format rule: the fragment body must NOT start with a leading `- ` — `cargo xtask collect-fragments` prepends its own `- ` when assembling `## [Unreleased]`, so this will render as `- - Shared...`. Please drop the leading `- `. Also missing the trailing `(#6868)` reference before `(@houko)` (not tooling-enforced, but prevents a duplicate generated line in the release notes). Same issue applies to the sibling fragment `changelog.d/fixed/6868-telegram-multipart-timeout.md` in this same PR — see the other comment.

---

changelog.d/fixed/6868-telegram-multipart-timeout.md
```diff
@@ -0,0 +1 @@
+- Scaled Telegram multipart upload timeouts with payload size so valid large media can complete on slower links. (@houko)
```
houko: Same changelog.d/README.md format issue as the sibling fragment `changelog.d/changed/6868-telegram-multipart-bytes.md`: this body also starts with a leading `- ` (will double up to `- - Scaled...` when assembled) and is missing the trailing `(#6868)` reference before `(@houko)`.

---

houko (commented):
Automated CLAUDE.md compliance check: two changelog fragment formatting nits below (both fragments in this PR).

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
houko, line 1: Commit author identity check (per CLAUDE.md commit-msg hook / no-AI-attribution rule): commit `8568134b49096b641b8ba2d9ec35ad1ba8e3e3e5` ("docs: fix changelog fragment PR number and rewrap new prose comments") has author identity `Claude <noreply@anthropic.com>`. The message body itself carries no attribution string, but the author identity resolves to Claude/Anthropic, which is exactly what the `commit-msg` hook's identity check (`git var GIT_AUTHOR_IDENT`) is meant to catch server-side. This can't be fixed with a force-push from this side — @houko, please rebase and re-author that commit (e.g. `git rebase -i` + `--reset-author` with your own identity, or squash it into an adjacent commit under your name) before merge.

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
```diff
@@ -534,10 +548,40 @@ impl BotClient {
     }
 }
 
+fn shared_upload_bytes(bytes: Vec<u8>) -> Bytes {
+    Bytes::from(bytes)
+}
+
+fn multipart_timeout(byte_len: usize) -> Duration {
+    let transfer_secs = u64::try_from(byte_len / MULTIPART_BYTES_PER_SECOND).unwrap_or(u64::MAX);
```
houko, line 568: `multipart_timeout` has a sane floor (always `>= SEND_TIMEOUT_SECS`, so a tiny/zero-byte upload can't produce a degenerate low timeout), but no ceiling: `transfer_secs` grows linearly and unbounded with `byte_len`, e.g. the existing test shows 64 MB already yields a 1030s (~17 min) deadline, and a 2 GB upload (within range for a local Bot API server) would compute to ~8.7 hours. A single pathological/huge upload can therefore pin an HTTP client for an arbitrarily long time with no cap. Suggest clamping with `.min(some_max_secs)` — not fixing this myself since the "right" ceiling value is a product/ops judgment call (depends on what upload sizes this sidecar is expected to actually handle), not a mechanical fix.

---

houko (commented):
Mechanical CLAUDE.md-compliance pass. Fixed and pushed the changelog fragment format (leading `- ` marker + missing PR reference); left two review comments below for items that need maintainer judgment/action rather than a mechanical fix.

---

houko:
This branch's history contains a commit authored as Claude <noreply@anthropic.com>: 8568134b49096b641b8ba2d9ec35ad1ba8e3e3e5 ("docs: fix changelog fragment PR number and rewrap new prose comments").
CLAUDE.md documents a commit-msg hook that rejects commits whose author identity resolves to Claude/Anthropic even when the message text is clean, and the repo's stated policy is no Claude/Anthropic attribution anywhere in commit history.
This predates my review pass and I'm not rewriting someone else's already-pushed history unilaterally, so flagging it here for reauthoring or squashing before merge.

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
```diff
@@ -534,10 +548,40 @@ impl BotClient {
...
+    Duration::from_secs(SEND_TIMEOUT_SECS.saturating_add(transfer_secs))
```
houko, line 569: `multipart_timeout` has a floor (`SEND_TIMEOUT_SECS` = 30s, since `transfer_secs` can't go negative) but no ceiling.
`transfer_secs` is derived straight from `byte_len / MULTIPART_BYTES_PER_SECOND` and added into the returned `Duration` with no clamp, so the request deadline scales without bound as the attachment gets larger.
That's fine for a legitimately huge-but-slow upload, but it also means the timeout stops acting as a fail-safe against a stalled/degraded connection once the payload is large enough — the bigger the declared byte count, the longer a hung request is allowed to sit open, with no practical upper limit before hitting `Duration`/`u64` range only.
Worth considering a cap (e.g. a few minutes) on the final timeout so a very large attachment still gets a bounded worst case instead of an effectively unbounded one.

Separately, `u64::try_from(byte_len / MULTIPART_BYTES_PER_SECOND).unwrap_or(u64::MAX)` — this `unwrap_or` branch is dead on every platform Rust actually targets today: `usize` is never wider than `u64` there, so the conversion can't fail.
Not a bug, just worth a comment (or dropping the fallback) so a future reader doesn't assume it's reachable.

The boundary test (`multipart_timeout_scales_with_payload_size`) covers the floor nicely (0, 63_999, 64_000) but doesn't probe the missing-ceiling behavior at all — understandable since there's no ceiling to assert on yet.

---

houko (commented):
Two more findings that don't map to a specific line in this diff, plus one inline comment on `client.rs` (missing ceiling on the scaled multipart timeout).

**Python SDK port has the same gap this PR fixes for Rust.**
`sdk/python/librefang/sidecar/adapters/telegram.py` still uses a fixed `SEND_TIMEOUT_SECS = 10` for multipart uploads (`_send_media_upload` → `_multipart`, lines ~899-914), with no scaling by payload size.
Large media over a slow link can still time out prematurely in the Python sidecar even after this PR lands, the same parity gap flagged in three prior PRs in this series (#6875, #6873, #6871) for other Rust-side fixes that weren't mirrored on the Python side.
Worth a follow-up (or a scope decision from a maintainer) to port the scaling — even a coarse version of `multipart_timeout` would close the gap.

**Commit `8568134b` in this branch's history is authored by `Claude <noreply@anthropic.com>`** (both author and committer identity), for "docs: fix changelog fragment PR number and rewrap new prose comments."
This is flagged per CLAUDE.md's no-AI-attribution rule; I'm not rewriting history here, just surfacing it for the maintainer to decide whether it needs squashing/rebasing before merge.

No mechanical fixes applied — both the timeout-scaling formula and the retry-bytes sharing compile clean (`cargo check` / `cargo clippy -- -D warnings`) and the new boundary tests pass (`cargo test -p librefang-sidecar-telegram`, 64 passed including `multipart_timeout_scales_with_payload_size` and `multipart_retry_bytes_share_the_same_allocation`).
The changelog fragments (`changelog.d/changed/6868-telegram-multipart-bytes.md`, `changelog.d/fixed/6868-telegram-multipart-timeout.md`) are correctly formatted per `changelog.d/README.md`.

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
```diff
@@ -8,13 +8,17 @@
...
+use bytes::Bytes;
 use reqwest::Client;
 use serde::Serialize;
 use serde_json::{json, Value};
 use std::time::Duration;
 
 pub const DEFAULT_LONGPOLL_TIMEOUT_SECS: u64 = 30;
 pub const SEND_TIMEOUT_SECS: u64 = 30;
+/// Conservative throughput floor used to extend multipart request deadlines.
+/// The fixed send timeout remains the connection/base budget; uploads receive one extra second per 64 KB of payload.
```
houko, line 21: Commit 955f4d20 (the doc-comment KiB/KB fix that touches this line) is authored as `Claude <noreply@anthropic.com>` — same problem already called out on commit 8568134b earlier in this branch's history, but this one landed after that discussion and hasn't been flagged yet. CLAUDE.md's policy and the commit-msg hook both reject an author identity that resolves to Claude/Anthropic, and this one slipped in the same way. Also, while I have it open: the commit body is hard-wrapped at ~72 columns mid-sentence rather than one sentence per line, which is the repo's stated prose convention for commit bodies. Neither is something I can fix without a rebase + force-push, which isn't mine to do on someone else's branch. Flagging both for cleanup alongside the existing 8568134b item before merge.

---

houko (commented):
One new item since the last pass: a second Claude-authored commit landed on this branch (955f4d20) after the earlier 8568134b flag. Everything else already covered by prior review threads (changelog format resolved, timeout ceiling and dead unwrap_or already noted, Python-port parity gap already noted) checks out fine on this pass — the retry/timeout logic itself is correct and bounded.

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
houko, line 569: Status update, not a new finding: two prior review threads on this line already raised the missing ceiling on `multipart_timeout`.
Restating current status rather than a third full write-up: `transfer_secs` still scales unbounded with `byte_len`, so `multipart_timeout` has a floor (`SEND_TIMEOUT_SECS`) but no upper bound.
The 64 KB/s scaling factor itself is reasonable for legitimate large media and is not the concern.
Still open, still not something this pass should fix directly — the right ceiling value is a product/ops call, not mechanical.
The existing test (`multipart_timeout_scales_with_payload_size`) covers the floor and one scaled case but has no capped-large-file case, since there is no cap yet.

---

sdk/rust/librefang-sidecar-telegram/src/api/client.rs
houko, line 21: Status update, not a new finding: commit author identity on this branch is a repeated, already-flagged blocking issue.
Two commits in this PR's history are authored as `Claude <noreply@anthropic.com>`: `8568134b49096b641b8ba2d9ec35ad1ba8e3e3e5` and the current head, `955f4d202c1792d29962f37cc385d817f40094ce`.
The former has been flagged multiple times already; the latter landed after that discussion.
Both need a maintainer rebase (`--reset-author`, or squash before merge) — not fixable from here without a force-push this review is not authorized to perform.
Separately, and much lower priority: the head commit's message body is hard-wrapped at ~72 columns mid-sentence rather than one sentence per line, per the repo's prose convention; worth a rebase-time cleanup alongside the author fix.
Changelog fragments are fine now — the earlier missing-`(#6868)`/leading-`-` issue was already resolved by a follow-up commit on this branch.

---

houko (commented):
Automated daily pass summary: no new findings beyond what's already tracked on this branch.
Two items remain open and blocking-adjacent: the Claude-authored commits (`8568134b`, and now the head `955f4d20`) need a maintainer rebase, and `multipart_timeout` still has no upper bound.
Both were already flagged in prior threads; see inline comments for the consolidated status.
Changelog fragments and the timeout scaling factor itself are fine as written.



---

PR #6870, opened 2026-08-09
fix(telegram): reject malformed media groups

houko:
## Summary
- reject non-object MediaGroup entries instead of silently skipping them
- preserve fail-closed behavior before any Telegram request is sent
- add regression coverage for a mixed valid/malformed group

## Verification
- `cargo test --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml media_group_rejects_non_object_items`
- `git diff --check`


changelog.d/fixed/6870-telegram-media-group-validation.md
```diff
@@ -0,0 +1 @@
+- Rejected malformed non-object Telegram media-group entries instead of silently dropping them from the outgoing group. (@houko)
```
houko: changelog.d/README.md format rule: the fragment body must NOT start with a leading `- ` — `cargo xtask collect-fragments` prepends its own `- ` when assembling `## [Unreleased]`, so this will render as `- - Rejected...`. Please drop the leading `- `. Also missing the trailing `(#6870)` reference before `(@houko)` (not tooling-enforced, but prevents a duplicate generated line in the release notes).

---

houko (commented):

---

sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs
```diff
@@ -544,9 +544,9 @@ pub async fn dispatch_content(
 fn build_media_group(items: &[Value]) -> Result<Value> {
     let mut out: Vec<Value> = Vec::new();
     for item in items {
-        let Some(obj) = item.as_object() else {
-            continue;
-        };
+        let obj = item
+            .as_object()
+            .ok_or_else(|| Error::Other("MediaGroup item is not a JSON object".into()))?;
         if obj.len() != 1 {
```
houko, line 550: Rust-side fix looks correct and well-tested (`media_group_rejects_non_object_items`, `cargo check`/`cargo test -p librefang-sidecar-telegram` both green). Checked the Python port for the equivalent gap this campaign has been tracking (see #6875, #6873, #6871): `_send_media_group` in `sdk/python/librefang/sidecar/adapters/telegram.py` (around line 1010-1032) has its own bug in the same area, unrelated to this diff and not fixable inline here since it's outside this PR's changed files:

1. It checks `"Photo" in it` / `it["Photo"]`, but every other dispatch path in that same file (`_dispatch_content`, `elif kind == "Image":`) and this Rust port (`"Image" => "photo"`) use the tag `Image`, not `Photo`. No outbound item is ever tagged `Photo`, so every `Image` item inside a media group falls into the `else` branch and does `it["Video"]`, raising `KeyError: 'Video'` for any legitimate image-only album — i.e. media groups containing images are currently broken in the Python port.
2. A malformed non-dict item (e.g. a bare string) doesn't fail the way this PR's Rust fix now does. For a `str`, `"Photo" in it` is a substring test that evaluates to `False` rather than raising, so execution falls through to `it["Video"]` and raises a confusing `TypeError: string indices must be integers` instead of a clear, intentional error like this PR's `"MediaGroup item is not a JSON object"`.

Both would benefit from an explicit `isinstance(it, dict)` guard plus fixing `"Photo"` → `"Image"`, to keep the two ports in parity. Posting as a comment rather than a fix since the affected file is outside this PR's diff and lives in the separate Python SDK workspace.

---

houko (commented):
Reviewed the Rust-side fix — correct, minimal, and tested (see inline comment). No mechanical fixes pushed to this PR: the Rust change is already correct, the changelog fragment is already properly formatted, and no `Claude <noreply@anthropic.com>` (or other AI) attribution was found in this PR's own commits (332a6269, 747236ee, f137ee95, e98cd3a5 — all authored by the human contributor). Flagged a related, pre-existing Python-port gap inline rather than fixing it, since it's outside this PR's diff.

---

sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs
```diff
@@ -568,7 +568,7 @@ fn build_media_group(items: &[Value]) -> Result<Value> {
         let media = payload
```
houko, line 568: Added a fix in commit `aa6170be7` for a related gap in the same function this PR touches: `build_media_group` defaulted a missing `url` field to an empty string (`unwrap_or("")`) instead of rejecting the item, unlike every sibling single-item dispatch path (`Image`, `Video`, `File`, `Voice`, `Audio`, `Animation` — all at lines 164-343) which require `url` and error out when absent. A 2-item media group with one item missing `url` previously passed local validation and would have shipped an empty `media` value to Telegram. Added `media_group_rejects_item_missing_url` and a `media_group_accepts_a_valid_variable_length_group` (2-item and 10-item) regression test; `cargo test -p librefang-sidecar-telegram --bins` (65 passed), `cargo clippy --bins --all-targets -- -D warnings`, and `cargo fmt --check` all clean.

Separately: commit `aa6170be7`'s author identity is `Claude <noreply@anthropic.com>` — same issue as the one already flagged on PR #6868 (commit `8568134b4`). This session's git identity is globally misconfigured to that value and I can't correct it (changing `git config` and force-pushing to rewrite history are both outside what this session is permitted to do). @houko, this commit will need the same treatment as the #6868 one: rebase with `--reset-author` (or squash into an adjacent commit under your name) before merge.

---

sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs
```diff
@@ -568,7 +568,7 @@ fn build_media_group(items: &[Value]) -> Result<Value> {
...
             .get("url")
             .and_then(Value::as_str)
-            .unwrap_or("")
+            .ok_or_else(|| Error::Other(format!("MediaGroup item {tag} missing url")))?
```
houko, line 571: The url-presence fix itself looks correct and is well covered by the two new tests (`media_group_rejects_item_missing_url`, `media_group_accepts_a_valid_variable_length_group`); the check applies uniformly to both tags this function supports (`Image`/`Video`), so it's not a partial-coverage bug.

Two things about commit `aa6170be` specifically that block a mechanical fix from this pass and need your attention before merge:

1. **Author identity**: `git log -1 --format='%an <%ae>' aa6170be` returns `Claude <noreply@anthropic.com>`. CLAUDE.md forbids AI/Claude attribution, and the `commit-msg` hook checks exactly this (`GIT_AUTHOR_IDENT`) — this commit would be rejected if pushed through the configured hooks. Needs `git commit --amend --author="..."` (or an interactive rebase) and a force-push to your branch to fix, which is outside what this review pass is allowed to do.
2. **Commit body wrapping**: the message body is hard-wrapped at ~72 cols mid-sentence. CLAUDE.md's prose rule is one-sentence-per-line, no column wrap — worth fixing in the same amend since you'll already be rewriting this commit for (1).

Everything else in the diff (including the changelog fragment format) looks correct.

---

houko (commented):
Reviewed the new commit (aa6170be) pushed after the last review. The url-validation fix and its test coverage are solid; one blocking item flagged inline regarding the commit's author identity and message formatting.

---

sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs
houko, line 568: Status (consolidated, already raised on this commit): head commit `aa6170be7` is still authored as `Claude <noreply@anthropic.com>` and its body is still hard-wrapped mid-sentence — both blocking per CLAUDE.md and both flagged in prior review rounds on this thread. No mechanical fix available here (requires author-identity amend + rebase, which needs a force-push only the branch owner can do). @houko: please `git commit --amend --author="..."` (fix the wrapping in the same amend) and force-push `fix/telegram-media-group-validation` before merge.

---

sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs
```diff
@@ -646,6 +646,49 @@ fn looks_like_ogg_opus(bytes: &[u8]) -> bool {
 mod tests {
     use super::*;
```
houko, line 647: Validation ordering looks correct: `build_media_group` rejects a non-object item and a missing `url` via `?` before the entry is ever pushed to `out`, so a malformed item can't reach a later index/panic path, and the empty-group case is short-circuited earlier in `dispatch_content` (`items_array.is_empty()` → no-op) before `build_media_group` is ever called with zero items. Not blocking, just a coverage suggestion: consider adding a direct `build_media_group(&[])` regression test alongside the two new malformed-item tests, since that path is currently only reachable indirectly.

---

houko (commented):
Automated daily review pass. Changelog fragment format issue from the earlier round is fixed (commit `e98cd3a`) — no action needed there. `build_media_group`'s validation-before-processing ordering is correct (see inline note). One blocking item remains open on the head commit (author identity + prose wrap, both previously flagged) and one non-blocking test-coverage suggestion.



---

sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs
houko, line 1: ⚠️ Compliance flag (not a code-quality note): the branch tip commit `aa6170be7` ("fix(telegram): reject media-group items missing a required url") is authored as `Claude <noreply@anthropic.com>`.

CLAUDE.md's `commit-msg` hook exists specifically to reject a commit whose author identity resolves to Claude/Anthropic, and the repo's git conventions explicitly disallow AI attribution — this slipped through because `core.hooksPath` isn't set by default in a fresh clone, so the version-controlled hook never ran for whatever session made this commit.

I can't fix this myself: correcting the author identity needs a rebase (`--reset-author`) + force-push, which I'm not authorized to do unilaterally. Flagging for a maintainer to rebase with `--reset-author`, or squash before merge (which will attribute the squashed commit to whoever GitHub picks — typically the PR author).

The code change itself (fail-closed rejection of non-object media-group items and items missing `url`, plus the new regression tests) looks correct and matches CLAUDE.md's changelog-fragment format.

---

houko (commented):
Automated daily review pass.

The functional change (`build_media_group` now fails closed on non-object items and items missing `url`, instead of silently dropping/defaulting them) is correct and well covered by the new tests. Changelog fragment is correctly formatted and attributed.

One compliance issue found (see inline comment): the branch's tip commit is authored as `Claude <noreply@anthropic.com>`, which CLAUDE.md's git conventions and `commit-msg` hook explicitly disallow.

---

PR #6873, opened 2026-08-10
fix(telegram): log getFile failures

houko:
## Summary
- log Telegram getFile lookup errors before media-placeholder fallback
- include an escaped file identifier for operator correlation
- preserve existing best-effort media degradation behavior

## Verification
- `cargo test --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml file_lookup_error_log_identifies_the_failed_file`
- `cargo test --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml --all-targets`
- `cargo clippy --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml --all-targets -- -D warnings`
- `git diff --check`

PR #6874, opened 2026-08-10
fix(telegram): report invalid command ids

houko:
## Summary
- return protocol errors for non-numeric Telegram channel IDs across all command variants
- reject non-numeric reaction message IDs instead of silently succeeding
- keep behavior consistent with ordinary Send commands

## Verification
- `cargo test --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml commands_reject_non_numeric_channel_ids_consistently`
- `git diff --check`


changelog.d/fixed/6874-telegram-invalid-channel-id-errors.md
```diff
@@ -0,0 +1 @@
+- Reported invalid Telegram channel and reaction message IDs consistently across typing, reaction, interactive, streaming, and ordinary send commands. (@houko)
```
houko: changelog.d/README.md format rule: the fragment body must NOT start with a leading `- ` — `cargo xtask collect-fragments` prepends its own `- ` when assembling `## [Unreleased]`, so this will render as `- - Reported...`. Please drop the leading `- `. Also missing the trailing `(#6874)` reference before `(@houko)` (not tooling-enforced, but prevents a duplicate generated line in the release notes).

---

houko (commented):

---

changelog.d/fixed/6883-telegram-best-effort-command-logging.md
houko, line 1: This branch's diff against `main` bundles three separate concerns: the stated #6874 fix (report non-numeric channel/message/thread ids as errors instead of silently swallowing them), plus the commits/fragments for #6883 (log best-effort `Typing`/`Reaction` API failures) and #6886 (log unknown `StreamDelta` stream ids). Per CLAUDE.md's "One PR ↔ one issue" rule these would normally land independently; not fixing this myself since splitting history is the author's call.

On the substance: I traced where the new hard errors go — `on_command`'s `Err` is turned into a protocol `error` event (`sdk/rust/librefang-sidecar/src/runtime.rs`), which `librefang-channels/src/sidecar.rs` only `warn!`-logs and stores in `status.last_error` for the operator/dashboard; it is never relayed back into the Telegram chat, so surfacing "non-numeric channel_id: …" this way doesn't leak anything to end users. `best_effort_command_error` also `{:?}`-escapes the rendered error before logging, so a forged `\r\n`/ANSI-escape payload from an upstream error can't do log injection (covered by the new control-escaping test). `cargo test -p librefang-sidecar-telegram --all-targets` passes 65/66 (1 intentionally `#[ignore]`d subprocess fixture, itself exercised by the non-ignored test that shells out to it).

---

sdk/rust/librefang-sidecar-telegram/src/adapter.rs
houko, line 1: The logic changes here check out — verified with `cargo check --bins` and `cargo test -p librefang-sidecar-telegram` (65 passed, including the new `commands_reject_non_numeric_channel_ids_consistently` and the subprocess-fixture pair for best-effort logging). The `parse_thread_id` signature change to `Result<Option<i64>, String>` is propagated correctly to every call site (`on_send`, `StreamStart`), and the previously-silent `Ok(())` returns on bad `channel_id` / `message_id` for Typing, Reaction, Interactive, and StreamStart are now correctly converted to hard errors, while the actual best-effort Bot API calls (send_chat_action, set_message_reaction) remain best-effort and just gained logging — that split matches the PR description.

Same process note as #6873: this branch/PR bundles three separate fixes under one PR titled only for #6874 — the invalid-command-id error reporting (#6874), best-effort command failure logging (#6883), and unknown-stream-delta logging (#6886), each with its own changelog fragment and no shared root cause. Per CLAUDE.md's "One PR ↔ one issue (or one tight cluster)" rule, worth considering a split before merge — flagging for maintainer judgment.

---

changelog.d/fixed/6883-telegram-best-effort-command-logging.md
```diff
@@ -0,0 +1 @@
+Logged Telegram typing-action and reaction-update API failures while preserving their best-effort command semantics. (#6883) (@houko)
```
houko, line 1: This branch's head (`fix/telegram-invalid-channel-id-errors`) carries two more commits beyond the invalid-command-id fix this PR describes: `ff653818 fix(telegram): log best-effort command failures (#6883)` and `ba409626 fix(telegram): log unknown stream deltas (#6886)`. Both are already open as their own PRs (#6883, #6886) on separate branches. Merging #6874 as-is will land that unrelated work under this PR's title/description, and #6883/#6886 will then have an empty/superseded diff against `main`. Recommend rebuilding this branch so its tip only contains the invalid-command-id commit before merging, per the "One PR ↔ one issue" convention.

---

houko (commented):

---

changelog.d/fixed/6883-telegram-best-effort-command-logging.md
houko, line 1: Same pattern as on #6873: this PR's title/body only describes "report invalid command ids", but the diff also carries `#6883` (best-effort command-failure logging in `adapter.rs`/`api/client.rs`, this changelog fragment) and `#6886` (unknown-stream-delta logging, `changelog.d/fixed/6886-telegram-unknown-stream-delta-log.md`) as separate, undisclosed commits (`ff6538186`, `ba4096267` in `git log 3362202f..HEAD`, neither present on `origin/main`). Per CLAUDE.md's "One PR ↔ one issue" rule, please either split these into their own PRs or update the title/body to cover all three.

The core "invalid command id" fix itself checks out: every command variant's `channel_id`/`message_id`/`thread_id` parse failure now returns a descriptive `Err` (`non-numeric channel_id/message_id/thread_id: <value>`) instead of silently no-op'ing via `return Ok(())`, and there's no `unwrap`/panic on the parse path — confirmed by `commands_reject_non_numeric_channel_ids_consistently` covering `Typing`, `Reaction`, `Interactive`, `StreamStart`, and `on_send`'s `thread_id`.

---

sdk/rust/librefang-sidecar-telegram/src/adapter.rs
```diff
@@ -373,6 +393,10 @@ impl SidecarAdapter for TelegramAdapter {
             Command::StreamDelta(d) => {
                 let mut map = self.streams.lock().await;
                 let Some(state) = map.get_mut(&d.stream_id) else {
+                    eprintln!(
+                        "[telegram] StreamDelta for unknown stream_id={:?}, dropped",
```
houko, line 404: `StreamDelta` for an unknown `stream_id` now logs before dropping here (bundled `#6886` commit), but `StreamEnd` for an unknown `stream_id` (a few lines further down, `map.remove(&e.stream_id)` returning `None`) still drops silently with no log line. Both are the same "best-effort handling of a stream id with no active state" case — was the asymmetry intentional, or should `StreamEnd` get the same `eprintln!` treatment for consistency?

Separate minor note: `best_effort_command_error` here and `file_lookup_error_log`/`callback_ack_error_log` (translator.rs, added in #6873) all implement the same "format!(\"...: {:?}\", error.to_string())" control-character-escaping pattern independently. Once both land, worth consolidating into one shared helper rather than three near-identical copies.

---

houko:
The latest commit on this branch (`e2a3813f`, "fix(telegram): log StreamEnd for unknown stream ids, matching StreamDelta") is authored as `Claude <noreply@anthropic.com>` — the same author-identity issue the `commit-msg` hook is meant to catch, independent of the message text.
Will need a rebase/amend onto a real author identity before merge.
The fix itself is correct: `StreamDelta` already logged and dropped commands for a `stream_id` with no active state, but the sibling `StreamEnd` path silently dropped the same case with no log line — this makes the two best-effort paths consistent.

---

sdk/rust/librefang-sidecar-telegram/src/adapter.rs
houko, line 1: Blocking: commit `e2a3813f` ("fix(telegram): log StreamEnd for unknown stream ids, matching StreamDelta") is authored as `Claude <noreply@anthropic.com>`, a hard CLAUDE.md violation on author identity (this has already been raised once on this thread for the same commit).
Needs a human rebase/amend onto a real author identity before merge — not force-pushing this myself.

On the substance of the diff otherwise: the invalid-command-id fix itself is correct.
Every `channel_id` / `message_id` / `thread_id` parse failure across `Typing`, `Reaction`, `Interactive`, `StreamStart`, and `on_send`'s thread handling now returns a descriptive `Err` instead of the previous silent `return Ok(())`, with no `unwrap`/`expect`/panic on attacker-controlled Telegram text — bad input always goes through `.parse::<i64>().ok()` behind an `else` guard.
The error is surfaced through the normal `Result` propagation path (`on_command`'s `Err` → protocol `error` event), not a new panic.
`commands_reject_non_numeric_channel_ids_consistently` exercises actual malformed ids (`"telegram:not-a-number"`, `"not-a-number"`) across all the affected command variants, not just valid ones.
Changelog fragments are present and correctly formatted (no leading `- `, trailing `(#NNNN) (@houko)`).

Applied one small mechanical fix directly on this branch: the new `best_effort_command_error` doc comment was hard-wrapped mid-sentence at a fixed column width; reflowed it to one sentence per line per the repo's prose-wrapping convention (commit `275ade6f`).



---

houko (commented):
Automated daily review pass on PR #6874.

---

PR #6875, opened 2026-08-10
fix(telegram): normalize reaction emoji variants

houko:
## Summary
- strip trailing emoji text/presentation selectors before Telegram reaction mapping
- apply normalization consistently to progress, completion, and failure reactions
- preserve done-reaction suppression after normalization

## Verification
- `cargo test --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml variation_selectors_do_not_bypass_reaction_mappings`
- `git diff --check`


changelog.d/changed/6888-telegram-done-reaction-policy.md
houko, line 1: This branch's diff against `main` bundles two separate concerns: the stated #6875 fix (normalize trailing variation selectors before matching reactions) plus the commit/fragment for #6888 (replace the `clear_done_reaction: bool` field with a `DoneReactionPolicy` enum). Per CLAUDE.md's "One PR ↔ one issue (or one tight cluster)" rule these would normally land independently. Not fixing this myself — restructuring branch history is a judgment call for the author.

The normalization logic itself checks out: `trim_end_matches(['\u{FE0F}', '\u{FE0E}'])` is applied only to decide which match arm fires, while the returned value for unmatched input is still the original `trimmed` string — so passthrough emoji that carry a variation selector (e.g. `♥️`, `✍️`, both covered by the new test) keep their selector instead of being corrupted to the bare/text-style codepoint. Multi-codepoint ZWJ sequences aren't touched by the trim set so they fall through to the passthrough arm unchanged. `cargo test -p librefang-sidecar-telegram --all-targets` passes all 63 tests on this branch.

---

sdk/rust/librefang-sidecar-telegram/src/reaction.rs
```diff
@@ -3,22 +3,63 @@
...
 //! Falls back to the raw emoji when not in the table — Telegram silently drops unknown reactions, which is acceptable: better to send something the user typed and have Telegram refuse it than to refuse client-side and lose the signal entirely.
 
-pub fn map_reaction(input: &str, clear_done: bool) -> Vec<String> {
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DoneReactionPolicy {
+    Emit,
+    Suppress,
+}
+
+pub fn map_reaction(input: &str, done_policy: DoneReactionPolicy) -> Vec<String> {
```
houko, line 12: Verified: within this file the fix is symmetric — `map_reaction` is the only reaction-mapping path in the Rust sidecar (outgoing `Command::Reaction` only; there's no separate incoming-webhook reaction path here), and the new test exercises real U+FE0F/U+FE0E codepoints rather than ASCII stand-ins. No incoming/outgoing asymmetry inside this crate.

However, there is a cross-implementation asymmetry worth a decision: `sdk/python/librefang/sidecar/adapters/telegram.py` documents itself as being in "PARITY STATUS" with this file — "full parity ... faithful port (function-by-function, not re-derived)" of the audited Rust — and explicitly calls out `reaction` as "same emoji map, optional clear-on-done". Its `_REACTION_MAP`/`_map_reaction` (around line 627-637) still does a plain `dict.get(emoji, emoji)` lookup with no variation-selector stripping, and `_do_reaction` (line 1632-1636) compares `cmd.reaction == _DONE_EMOJI` by exact string equality. A reaction carrying a trailing U+FE0F/U+FE0E (e.g. `"⚙️"`, `"✅️"`) will bypass both the emoji map and the done-reaction suppress policy there — the identical bug this PR just fixed in Rust.

Given the explicit parity contract in that file's own docstring, should the same `trim_end_matches(['\u{FE0F}', '\u{FE0E}'])` normalization land in the Python port too (this PR or a fast-follow), or is Python intentionally out of scope for this fix? Leaving as a question rather than a diff since it's a different SDK/language and outside this PR's touched files.

---

houko (commented):
Automated review pass: Rust-side fix looks correct and consistent (single mapping path, symmetric normalization, real-codepoint test). Left one comment flagging a cross-SDK parity gap with the Python sidecar for a maintainer decision — see inline.

---

changelog.d/changed/6888-telegram-done-reaction-policy.md
```diff
@@ -0,0 +1 @@
+Replaced the Telegram done-reaction boolean argument with an explicit emit-or-suppress policy type. (#6888) (@houko)
```
houko, line 1: This branch's head (`fix/telegram-reaction-variation-selectors`) carries one more commit beyond the variation-selector fix this PR describes: `75a41fee refactor(telegram): name done reaction policy (#6888)`, which is already open as its own PR (#6888) on a separate branch. Merging #6875 as-is will land that unrelated refactor under this PR's title/description, and #6888 will then have an empty/superseded diff against `main`. Recommend rebuilding this branch so its tip only contains the variation-selector commit before merging, per the "One PR ↔ one issue" convention.

---

houko (commented):

---

changelog.d/changed/6888-telegram-done-reaction-policy.md
houko, line 1: Same pattern as #6873/#6874: the PR title/body describes only the variation-selector normalization, but the diff also bundles `#6888` (the `clear_done_reaction: bool` → `DoneReactionPolicy` enum refactor, this changelog fragment) as an undisclosed second change (`git log 3362202f..HEAD` shows it as its own commit `75a41fee9`, absent from `origin/main`). Please split it out or fold it into the PR description per CLAUDE.md's "One PR ↔ one issue" rule.

The normalization logic itself is correct: `trim_end_matches(['\u{FE0F}', '\u{FE0E}'])` strips both the emoji-presentation (VS16) and text-presentation (VS15) selectors before matching against the bare-codepoint canonical arms (`⏳`, `⚙`, `✅`, `❌` — verified none of these literals carry a hidden selector), and `variation_selectors_do_not_bypass_reaction_mappings` exercises both selectors against all four mapped reactions under both `DoneReactionPolicy` values. Skin-tone modifiers aren't handled, but none of the four canonical reactions are modifier-eligible base emoji, so that's not a gap in practice.

One thing worth double-checking with the author: the fallback arm (`_ => vec[trimmed.into()]`) returns the *original* `trimmed` string, not `normalized` — so an unmapped reaction like `"♥️"` is passed through to Telegram with its variation selector intact (confirmed by the `" ♥️ "` / `"✍️"` test cases), while a mapped reaction is always normalized. If that's deliberate — Telegram accepts either presentation for arbitrary reactions, and normalization only matters for hitting the mapping table — no change needed; just flagging in case a canonical VS-stripped output was expected for *all* reactions, not just the mapped ones.

---

houko, line 1: Commit `d182263e` ("fix: correct changelog fragment format for done-reaction policy entry") wraps its body at roughly 72 columns, splitting one sentence across three lines:\n\n> Drop the leading \"- \" (collect-fragments prepends its own, so this\n> would double up) and add the missing (#6888) reference so the release\n> flow doesn't also emit a generated line for the same PR.\n\nCLAUDE.md's prose-wrapping rule explicitly covers commit message bodies: break only at sentence boundaries, one sentence per line, no column limit.\nThis needs a reworded commit via interactive rebase, which is out of scope for a same-branch push — flagging for a manual reword rather than rewriting history here.\n\nSeparately, confirmed the normalization logic in `reaction.rs` itself: `trim_end_matches(['\\u{FE0F}', '\\u{FE0E}'])` is applied to the incoming reaction before matching against the bare-codepoint canonical arms (all four are unselected literals), so both sides of the comparison are effectively normalized.\nNo manual byte-slicing was introduced (uses `trim`/`trim_end_matches`, both char-boundary-safe), so there's no panic risk from this change.\nNone of the four mapped reactions are ZWJ sequences, so the ZWJ case doesn't arise for the mapped set; unmapped multi-codepoint emoji simply fall through to the passthrough arm unchanged, which is correct.

---

houko (commented):
Automated review pass on `d182263e54b6ff59daaf4be8e871ab9772b40eb6`.

Commit-author check (CLAUDE.md AI-attribution rule): all 6 commits checked, none authored by `Claude <noreply@anthropic.com>`.
No blocking finding here.

Emoji-normalization correctness (reaction.rs): verified correct, see inline note.

One minor finding: a commit message body violates the prose-wrapping rule (sentence-boundary breaks only) — see inline comment.




---

PR #6877, opened 2026-08-10
fix(telegram): relax response default bound

houko:
## Summary
- replace derived ApiResponse<T> Default with an unbounded manual implementation
- preserve all existing default field values and serde behavior
- cover result types that intentionally do not implement Default

## Verification
- `cargo test --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml api_response_default_does_not_require_result_default`
- `git diff --check`


changelog.d/fixed/6877-telegram-api-response-default.md
```diff
@@ -0,0 +1 @@
+- Removed the unnecessary `T: Default` bound from Telegram API response envelopes while preserving their default field values. (@houko)
```
houko: changelog.d/README.md format rule: the fragment body must NOT start with a leading `- ` — `cargo xtask collect-fragments` prepends its own `- ` when assembling `## [Unreleased]`, so this will render as `- - Removed...`. Please drop the leading `- `. Also missing the trailing `(#6877)` reference before `(@houko)` (not tooling-enforced, but prevents a duplicate generated line in the release notes).

---

houko (commented):

---

sdk/rust/librefang-sidecar-telegram/src/api/types.rs
```diff
@@ -247,16 +234,31 @@ pub struct PollAnswer {
 
 // ── Response envelopes for "send" endpoints ─────────────────────────
 
-#[derive(Debug, Clone, Deserialize, Default)]
-#[serde(default)]
+#[derive(Debug, Clone, Deserialize)]
 pub struct ApiResponse<T> {
     pub ok: bool,
```
houko, line 239: Dropping the struct-level `#[serde(default)]` here (kept only on the four `Option` fields, not on `ok`) is a real deserialize-behavior change, not just the stated bound relaxation.

Before this PR, `ApiResponse<T>` had `#[serde(default)]` at the struct level, so a response body missing `ok` deserialized with `ok: false` (and the send-endpoint call sites in `client.rs` then produced an `Error::Api` with an empty code/description).
After this PR, a missing `ok` fails deserialization outright (a `serde_json::Error` surfaces through the `?` in `call_json`/`get_updates` instead).

That's the same fail-closed posture already exercised by `required_update_identity_fields_fail_closed` for `UpdatesResponse`, so it's arguably the right direction — but it now also applies to every other `ApiResponse<T>` consumer (`SendMessageResult`, `PollResult`, `GetFileResult`, the generic `Value` response in `call_json`), and none of those get a "missing `ok`" regression test the way `UpdatesResponse` does. It also isn't what the PR description says ("preserve all existing default field values and serde behavior") — this one field's behavior did change for those call sites.

Worth either: (a) confirming this tightening is intentional and adding a same-shaped test for a non-`UpdatesResponse` `ApiResponse<T>` instantiation, or (b) restoring `#[serde(default)]` on `ok` specifically if the intent really was zero behavior change beyond the `T: Default` bound. Not blocking since Telegram's Bot API always sends `ok`, but flagging since it's a silent widening of scope beyond the stated "relax the bound" fix.

---

houko (commented):
Automated review pass on #6877. No mechanical fixes applied — the change compiles, passes `cargo test -p librefang-sidecar-telegram` (66/66) and is clippy/fmt clean in its own workspace. One design-judgment finding left inline on `api/types.rs` (missing struct-level `#[serde(default)]` on `ApiResponse::ok` widens scope beyond the stated bound-relaxation). No AI/Claude/Anthropic attribution found in this branch's commits.

---

changelog.d/changed/6889-telegram-inline-button-action.md
houko, line 1: This branch's diff against `main` bundles three separate concerns: the stated #6877 fix (drop the `T: Default` bound on `ApiResponse`), plus commits/fragments for #6884 (unify `getUpdates`/send response envelopes) and #6889 (make inline-keyboard URL/callback actions mutually exclusive). Per CLAUDE.md's "One PR ↔ one issue (or one tight cluster)" rule, these would normally review and land independently. Not fixing this myself since re-splitting a branch's history is the author's call, not a mechanical edit.

For what it's worth, the code itself is sound: the manual `Default` impl correctly avoids requiring `T: Default` while preserving per-field defaults (verified with a `NoDefault` marker type in the new test), the `UpdatesResponse` type alias round-trips through the existing "error envelopes may omit result" test now via `Option::None`, and the `InlineKeyboardAction` untagged enum serializes to exactly one Bot API field per the new test. `cargo test -p librefang-sidecar-telegram --all-targets` passes all 63 tests on this branch.

---

changelog.d/changed/6884-telegram-response-envelope.md
```diff
@@ -0,0 +1 @@
+Unified Telegram `getUpdates` and command responses behind the same generic API envelope while retaining strict required-status parsing. (#6884) (@houko)
```
houko, line 1: This branch's head (`fix/telegram-api-response-default`) carries two more commits beyond the `Default` bound fix this PR describes: `28940748 refactor(telegram): unify response envelopes (#6884)` and `1d80896f refactor(telegram): type inline button actions (#6889)`. Both are already open as their own PRs (#6884, #6889) on separate branches. Merging #6877 as-is will land that unrelated work under this PR's title/description, and #6884/#6889 will then have an empty/superseded diff against `main`. Recommend rebuilding this branch so its tip only contains the `Default`-bound fix before merging, per the "One PR ↔ one issue" convention.

---

houko (commented):

---

changelog.d/changed/6884-telegram-response-envelope.md
houko, line 1: Same pattern as #6873/#6874/#6875: title/body describe only the `Default` bound relaxation, but the diff also bundles `#6884` (`UpdatesResponse` → `ApiResponse<Vec<Update>>` type alias, this fragment) and `#6889` (`InlineKeyboardButton` url/callback_data → `InlineKeyboardAction` enum, `changelog.d/changed/6889-telegram-inline-button-action.md`) as undisclosed separate commits (`28940748f`, `1d80896ff` — neither on `origin/main`). Please split or fold into the PR description per CLAUDE.md's "One PR ↔ one issue" rule.

On the substance — despite the title, "bound" here is a Rust generic trait bound (`#[derive(Default)]` on `ApiResponse<T>` requiring `T: Default`), **not** a numeric size/count limit, so this does not reopen any resource-exhaustion concern; `MAX_STREAM_BUFFER_BYTES`, `MAX_ACTIVE_STREAMS`, `CAPTION_LIMIT_UTF16`, `FILE_DATA_BYTE_CAP`, and `MAX_RETRY_AFTER_SECS` are all untouched by this diff. The manual `Default` impl is correct (all fields still default to `false`/`None` as before) and the `NoDefault` marker-type test exercises exactly the case the derive couldn't handle.

Genuine behavior change worth a second look: `ApiResponse<T>` previously carried struct-level `#[serde(default)]`, so a JSON payload missing `"ok"` entirely would deserialize with `ok: false` rather than failing. The field-level `#[serde(default)]` on the other four fields drops that — `ok` is now a required field for every `ApiResponse<T>` instantiation, including the `ApiResponse<Value>` used for `sendMessage`/etc. responses in `client.rs`, not just for `UpdatesResponse` (which already required it). In practice Telegram always sends `"ok"`, and the existing `required_update_identity_fields_fail_closed` test name suggests fail-closed is the intended posture here — just flagging that the strictness now applies more broadly than before, in case that's not what was intended for the generic envelope.

The `InlineKeyboardAction` enum change is a solid correctness improvement independent of its scope-bundling: it makes "both `url` and `callback_data` set" a structurally unrepresentable state (Telegram's Bot API rejects a button with both), and `inline_keyboard_action_serializes_as_exactly_one_bot_api_field` / `inline_keyboard_builder_emits_one_action_per_button` both confirm the url-wins-over-callback_data precedence is preserved from the original `if/else if`.

---

sdk/rust/librefang-sidecar-telegram/src/api/types.rs
houko, line 1: Daily automated pass, head `d38c7dcc3bfd9f2abd35a9bcfecc726503329cf7`.

Commit-author check: all 6 commits on this branch are attributed to `Evan Hu` / `Evan` / `houko` (`suzukaze.haduki@gmail.com`).
No `Claude <noreply@anthropic.com>` author identity found this round — no blocking attribution finding.

Bound-relaxation safety check: "relax response default bound" refers to a Rust generic trait bound (dropping the derive-imposed `T: Default` requirement on `ApiResponse<T>` in favor of a manual `impl<T> Default`), not a numeric size/count limit.
`MAX_STREAM_BUFFER_BYTES`, `MAX_ACTIVE_STREAMS`, `CAPTION_LIMIT_UTF16`, `FILE_DATA_BYTE_CAP`, and `MAX_RETRY_AFTER_SECS` are all untouched by this diff, so no resource-exhaustion/DoS regression is reintroduced here.
The `NoDefault` marker-type test (`api_response_default_does_not_require_result_default`) is the correct regression test for the actual change made (the trait-bound relaxation), so the "prove the previously-rejected case now works" requirement is satisfied for that part.

Changelog: `changelog.d/fixed/6877-telegram-api-response-default.md` format issue raised in the earlier pass (leading `- `, missing `(#6877)`) is fixed as of `d38c7dcc` — no outstanding format problem.

Two items already raised in prior passes remain open and unresolved; posting a status line only, not re-raising in full (both already flagged 2-3x):
1. `ApiResponse<T>::ok` lost its struct-level `#[serde(default)]` and is now a hard-required field for every `ApiResponse<T>` instantiation (not just `UpdatesResponse`, which already required it) — a real serde-behavior change the PR body's "preserve all existing... serde behavior" claim doesn't disclose. Still needs either an explicit "intentional" confirmation + a same-shaped missing-`ok` regression test for a non-`UpdatesResponse` instantiation, or restoring `#[serde(default)]` on `ok`.
2. The branch still bundles unrelated `#6884` / `#6889` commits and changelog fragments alongside the stated `#6877` fix. Per CLAUDE.md's "One PR ↔ one issue," this needs a rebase/split before merge — not a mechanical fix, author's call.

No new findings this pass.

---

houko (commented):
Automated daily review pass — no new findings; two previously-flagged items remain open for human judgment (see inline comment).



---

PR #6879, opened 2026-08-10
test(telegram): cover all schema fields

houko:
## Summary
- assert ALLOWED_USERS remains an optional advanced list field
- assert TELEGRAM_CLEAR_DONE_REACTION remains an optional advanced boolean
- retain the existing permit-all warning and streaming/token coverage

## Verification
- `cargo test --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml schema_declares_expected_fields`
- `git diff --check`


changelog.d/fixed/6879-telegram-schema-field-coverage.md
```diff
@@ -0,0 +1 @@
+- Expanded Telegram schema regressions to cover the type and visibility of every dashboard configuration field. (@houko)
```
houko: changelog.d/README.md format rule: the fragment body must NOT start with a leading `- ` — `cargo xtask collect-fragments` prepends its own `- ` when assembling `## [Unreleased]`, so this will render as `- - Expanded...`. Please drop the leading `- `. Also missing the trailing `(#6879)` reference before `(@houko)` (not tooling-enforced, but prevents a duplicate generated line in the release notes).

---

houko (commented):

---

changelog.d/fixed/6879-telegram-schema-field-coverage.md
houko, line 1: This PR is test-only (no production code changed, `test:` conventional-commit prefix) — it adds assertions to an existing test, it doesn't fix a bug.
CLAUDE.md's changelog guidance calls out that a test-only PR "may legitimately not need a changelog fragment," and if one is warranted it should land under the directory matching its `### ` heading.
Placing this under `fixed/` implies a user-facing bug was fixed, which isn't accurate here — nothing behavioral changed, only test coverage.
Worth a maintainer call: drop the fragment entirely, or move it to a heading that doesn't imply a behavior fix (none of `added/changed/security/documentation` fit test-coverage work well either, which may be the real signal that no fragment is needed).
Leaving as a comment since this is a judgment call about changelog philosophy, not a mechanical fix.

---

houko:
The latest commit on this branch (`0913dd5e`, "test(telegram): assert bot token field is not marked advanced") is authored as `Claude <noreply@anthropic.com>`.
The repo's `commit-msg` hook rejects commits whose author identity resolves to Claude/Anthropic even when the message text itself is clean, so this will need a rebase/amend onto a real author identity before it can land.
The change itself is correct — the schema-coverage test previously checked `TELEGRAM_BOT_TOKEN`'s type and required-ness but not its `advanced` flag, which is the one attribute that controls whether the required secret field is hidden behind the dashboard's advanced-settings toggle.

---

sdk/rust/librefang-sidecar-telegram/src/schema.rs
houko, line 1: Blocking: HEAD commit `0913dd5e` ("test(telegram): assert bot token field is not marked advanced") is authored as `Claude <noreply@anthropic.com>`.
CLAUDE.md prohibits AI/Claude/Anthropic attribution anywhere, and the repo's `commit-msg` hook independently rejects a commit whose author identity resolves to Claude/Anthropic even when the message text is clean, so this commit cannot land as-is.
Needs a rebase/amend onto a real contributor identity before merge — not something I'll force-push myself.
(This restates a still-applicable point from the earlier comment on this thread from 2026-08-10T20:16:12Z; flagging again here since the commit is unchanged.)

---

sdk/rust/librefang-sidecar-telegram/src/schema.rs
```diff
@@ -49,25 +49,39 @@ mod tests {
             .expect("schema must declare TELEGRAM_BOT_TOKEN");
         assert_eq!(bot_token.field_type, FieldType::Secret);
         assert!(bot_token.required);
+        assert!(!bot_token.advanced);
```
houko, line 52: The added assertions here are meaningful (they read real `field_type`/`required`/`advanced` values off the actual schema, not just constructing-and-compiling), so this isn't vacuous.
One gap against the PR's "cover all schema fields" framing though: there's no drift guard.
Nothing here asserts `schema.fields.len() == 4` (or otherwise confirms every field in `telegram_schema()`'s `vec[...]` has a matching `find(...)` block above), so a fifth field added to the schema later would compile and pass silently, uncovered, despite the test's name/title claiming full coverage.
Consider adding a count assertion near the top of the test (e.g. right after `assert_eq!(schema.name, "telegram");`) so the test fails loudly the next time a field is added without corresponding assertions.

---

houko (commented):
Automated daily review pass.

Summary:
- Blocking: HEAD commit `0913dd5e` is authored as `Claude <noreply@anthropic.com>` — needs a human rebase onto a real identity before merge (CLAUDE.md + `commit-msg` hook both reject it). This was already surfaced once on this thread; repeating since the commit is still unchanged.
- Test quality: the new assertions are real (they check actual `field_type`/`required`/`advanced` values), not vacuous, but the test has no drift guard — no `schema.fields.len()` (or equivalent) check — so a future field added to `telegram_schema()` without matching assertions would go silently uncovered despite the "cover all schema fields" framing.
- Changelog fragment format issue from the earlier review round (leading `- `, missing `(#6879)`) is already fixed in commit `97638e4b`.
- The still-open question of whether a test-only PR belongs under `changelog.d/fixed/` at all (raised in an earlier unresolved thread on this PR) is a legitimate point per CLAUDE.md's changelog guidance and is left to maintainer judgment; not re-raised here to avoid duplicate noise.




---

sdk/rust/librefang-sidecar-telegram/src/schema.rs
houko, line 1: Suggested fix for the still-open drift-guard gap noted earlier on this thread (no assertion on `schema.fields.len()`): insert this right after `assert_eq!(schema.name, "telegram");` (line 43, just above the added assertions in this PR's diff — not directly commentable since that line itself is unchanged):

```rust
// Drift guard: every field pushed into `telegram_schema()`'s field
// vec must have a matching `find(...)` block below with assertions
// on its type/required/advanced flags. Bump this count (and add the
// corresponding block) whenever a field is added, so a silently
// uncovered field trips this test instead of passing by omission.
assert_eq!(
    schema.fields.len(),
    4,
    "a field was added to or removed from telegram_schema() without updating this test's coverage"
);
```

Verified locally: with this added, `cargo test -p librefang-sidecar-telegram --bins schema_declares_expected_fields` still passes (4 fields currently declared).

Separately, not re-raising here since it has already been flagged twice on this thread with no change since (an issue comment and a review comment, both against this same commit): HEAD (`0913dd5e`) is still authored/committed as `Claude <noreply@anthropic.com>`, which both CLAUDE.md's no-AI-attribution policy and the repo's `commit-msg` hook reject. That still needs a human rebase/amend before merge.

---

PR #6881, opened 2026-08-10
perf(telegram): avoid href lowercase allocation

houko:
## Summary
- compare allowed href scheme prefixes in place
- avoid allocating a lowercase String for every Telegram link
- preserve whitespace trimming, ASCII case folding, and UTF-8 safety

## Verification
- `cargo test --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml accepts_safe_href`
- `git diff --check`


houko:
⚠️ Compliance flag (not a code-quality note): commit `744b47b2` on this branch is authored as `Claude <noreply@anthropic.com>`. CLAUDE.md's `commit-msg` hook exists specifically to reject a commit whose author identity resolves to Claude/Anthropic, and the repo's git conventions explicitly disallow AI attribution — this slipped through because `core.hooksPath` isn't set by default in a fresh clone, so the version-controlled hook never ran for whatever session made this commit. I can't fix this myself: rewriting the offending commit needs a rebase + force-push, which I'm not authorized to do unilaterally. Flagging for a maintainer to rebase with `--reset-author` (or squash before merge, which will attribute the squashed commit to whoever GitHub picks — typically the PR author).

---

changelog.d/fixed/6881-telegram-href-scheme-check.md
```diff
@@ -0,0 +1 @@
+Removed a per-link allocation from Telegram href scheme validation while preserving case-insensitive and UTF-8-safe checks. (#6881) (@houko)
```
houko, line 1: Nit (not about this file's content, which is fine): the commit that produced this file, `744b47b2` ("docs: fix changelog fragment format for #6881"), has a body hard-wrapped at ~70 columns, breaking mid-sentence across three lines. CLAUDE.md's prose-wrapping rule explicitly covers commit message bodies — the only legitimate break is at a sentence boundary. Same commit already has an open compliance flag for its author identity, so no separate fix needed from me here (rewriting it means rebase + force-push), but flagging so a maintainer rewording that commit anyway can fix both at once.

---

changelog.d/fixed/6881-telegram-href-scheme-check.md
houko, line 1: Commit `744b47b21771b9b28190b1f26ff9e425c9632da0` ("docs: fix changelog fragment format for #6881", which edited this file) has author identity `Claude <noreply@anthropic.com>` in `git log`.
Per CLAUDE.md this is exactly the pattern the `commit-msg` hook's author-identity check is meant to catch (Claude / Anthropic attribution, including via `GIT_AUTHOR_NAME` / `GIT_AUTHOR_EMAIL`), so it shouldn't merge as-is.
Suggest amending that commit's author identity (`git commit --amend --author "..."` or an interactive rebase) before merge; leaving this to the branch owner rather than rewriting it myself.

Separately, on the substance: the `href_is_safe` rewrite itself looks correct.
`trimmed.get(..scheme.len())` only returns `Some` when the prefix decodes as `scheme.len()` bytes of plain ASCII (any real scheme match forces this), so the char-boundary check can't introduce a false negative relative to the old `to_ascii_lowercase() + starts_with()` behavior, and it safely falls back to `None`/`false` instead of panicking on short or multibyte-misaligned input.
The added tests (`" HTTPS://..."`, `"éttps://..."`) cover the mixed-case and non-ASCII cases this class of change usually breaks.

---

houko, line 1: Status (consolidated, not re-explaining): head commit `744b47b2` ("docs: fix changelog fragment format for #6881") still carries author identity `Claude <noreply@anthropic.com>` — a hard CLAUDE.md violation of the `commit-msg` hook's author-identity check. This has already been flagged twice on this PR (issue comment 2026-08-10T10:46 and review thread 2026-08-10T13:09) and remains unresolved as of head sha `744b47b2`. No new information to add. Required action: maintainer rebase (`git commit --amend --author "..."` or interactive rebase) or squash-merge, which reattributes the squashed commit to the merger. Not authorized to rebase/force-push this myself.

---

sdk/rust/librefang-sidecar-telegram/src/format/sanitize.rs
houko, line 1: Correctness check on the allocation-avoiding rewrite (independent verification, since this touches sanitization): `ALLOWED_HREF_SCHEMES` (`https:`, `http:`, `mailto:`, `tg:`) is a positive allowlist, not a `javascript:` blocklist — `href_is_safe` returns `false` for anything that doesn't case-insensitively match one of these four prefixes, both before and after this change. So `javascript:` and every case permutation (`JavaScript:`, `JAVASCRIPT:`, `JaVaScRiPt:`, …) were never in the accept set and remain rejected; there is no path by which this rewrite could let a dangerous scheme through the case check.

`trimmed.get(..scheme.len())` returns `None` (→ `false`) whenever the prefix is shorter than the scheme or doesn't land on a UTF-8 char boundary, so short/multibyte-misaligned input can't panic or spuriously match — confirmed against the added `éttps://example.com` case: the leading 2-byte `é` shifts every subsequent byte, so no allowed-scheme byte-prefix can align, and the comparison correctly returns `false`. `eq_ignore_ascii_case` only folds ASCII bytes, and every allowed scheme is pure ASCII, so no non-ASCII case-folding trick (e.g. Kelvin sign → `k`) can produce a false accept either. No case-sensitivity or security regression found; this is a faithful allocation-free rewrite of the old `to_ascii_lowercase()+starts_with()` behavior.

Minor (non-blocking): the PR body's Verification section lists the test/clippy commands run but doesn't spell out this before/after correctness argument — worth adding a sentence given the change touches sanitization logic, per CLAUDE.md's PR-body guidance.

---

houko (commented):
Automated daily review pass. One blocking finding (AI-attribution, already flagged by prior passes — consolidated, not repeated in full) and one correctness verification of the sanitization-adjacent perf change (no regression found).



---

changelog.d/fixed/6881-telegram-href-scheme-check.md
houko, line 1: Automated review pass.

**Code change (`sanitize.rs::href_is_safe`)**: verified correct. The switch from `to_ascii_lowercase()` + `starts_with` to `trimmed.get(..scheme.len())` + `eq_ignore_ascii_case` is behaviorally equivalent:
- `str::get` on a byte range returns `None` (fail closed → rejected) rather than panicking when `scheme.len()` doesn't land on a UTF-8 char boundary, so multi-byte prefixes like `éttps://…` are safely rejected (covered by the new `accepts_safe_href` assertions).
- Since every entry in `ALLOWED_HREF_SCHEMES` is pure ASCII, a byte-exact prefix match can only occur at a valid char boundary anyway, so this can't accidentally accept something the old lowercase+`starts_with` path would have rejected (or vice versa).
- `javascript:` and other non-allowlisted schemes are still rejected regardless of case, since they were never in `ALLOWED_HREF_SCHEMES` to begin with — that hasn't changed.
- `cargo check -p librefang-sidecar-telegram` and `cargo test -p librefang-sidecar-telegram sanitize` both pass cleanly against this branch's HEAD (744b47b).

**Process issue — commit authored/committed as Claude**: commit `744b47b` ("docs: fix changelog fragment format for #6881") on this branch has both author and committer identity `Claude <noreply@anthropic.com>`:

```
744b47b21771b9b28190b1f26ff9e425c9632da0 Claude <noreply@anthropic.com> / Claude <noreply@anthropic.com>
```

This is the recurring no-AI-attribution violation flagged across several sibling PRs in this review pass — the repo's `commit-msg` hook is specifically designed to reject a commit whose author identity resolves to Claude/Anthropic even when the message text itself is clean, and this one slipped through (likely pushed via an API path that bypasses local hooks). The content of the commit itself (fixing the changelog fragment format) is fine — the identity is the problem. Recommend the branch owner re-commit that change under their own identity (e.g. `git commit --amend --reset-author` or squash it into the human-authored history) before merge, per the project's git conventions.

No other findings; not blocking on the code itself, but the attribution issue should be resolved before merge per project policy.

---

houko (commented):
Automated review: scheme-check logic verified correct (case-insensitive, UTF-8-boundary-safe, allocation-free); one process finding — see inline comment re: a Claude-attributed commit on this branch.

---

PR #6880, opened 2026-08-10
docs(telegram): clarify partial chunk delivery

houko:
## Summary
- document that multi-chunk Telegram sends are sequential rather than atomic
- clarify that an error may follow already-delivered earlier chunks
- set caller expectations without changing delivery behavior

## Verification
- `cargo fmt --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml -- --check`
- `cargo check --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml --all-targets`
- `cargo clippy --manifest-path sdk/rust/librefang-sidecar-telegram/Cargo.toml --all-targets -- -D warnings`
- `git diff --check`

PR #6887, opened 2026-08-10
fix(telegram): validate poll answers locally

houko:
## Summary
- require 1-12 string poll options before any Bot API request
- require quiz polls to provide a correct option index within the option array
- reject malformed and overflowing answer indices with local diagnostics

## Tests
- `git diff --check`


houko:
⚠️ Compliance flag (not a code-quality note): commit `9244904a` on this branch is authored as `Claude <noreply@anthropic.com>`. CLAUDE.md's `commit-msg` hook exists specifically to reject a commit whose author identity resolves to Claude/Anthropic, and the repo's git conventions explicitly disallow AI attribution — this slipped through because `core.hooksPath` isn't set by default in a fresh clone, so the version-controlled hook never ran for whatever session made this commit. I can't fix this myself: rewriting the offending commit needs a rebase + force-push, which I'm not authorized to do unilaterally. Flagging for a maintainer to rebase with `--reset-author` (or squash before merge, which will attribute the squashed commit to whoever GitHub picks — typically the PR author).

---

changelog.d/fixed/6887-telegram-poll-validation.md
```diff
@@ -0,0 +1 @@
+Rejected malformed Telegram poll options and missing or out-of-range quiz answers before issuing a Bot API request. (#6887) (@houko)
```
houko: CLAUDE.md: "No AI / Claude / Anthropic attribution" — the tip commit of this branch (`9244904a`, "docs: fix changelog fragment format for #6887", which touches this file) is authored by `Claude <noreply@anthropic.com>`. The message text itself is clean, but CLAUDE.md's `commit-msg` hook section explicitly calls out author-identity attribution as its own rejection case ("...and separately rejects a commit whose *author identity* ... resolves to Claude / Anthropic even when the message itself is clean"). Fixing this requires rewriting the commit (amend + force-push), which is outside what an automated review pass should do unilaterally — flagging for the maintainer to re-author (`git commit --amend --author="houko <...>"` or an interactive rebase) before merge.

---

sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs
```diff
@@ -38,6 +38,54 @@ fn required_coordinate(payload: &Value, key: &str) -> Result<f64> {
         .ok_or_else(|| Error::Other(format!("Location.{key} missing or not a JSON number")))
 }
 
+fn validated_poll(payload: &Value) -> Result<(Vec<Value>, bool, Option<u32>)> {
+    let raw_options = payload
+        .get("options")
+        .and_then(Value::as_array)
+        .ok_or_else(|| Error::Other("Poll.options missing or not a JSON array".into()))?;
+    if !(1..=12).contains(&raw_options.len()) {
```
houko: The option-count bound (1–12) and the quiz-answer range check are correct for the current Telegram Bot API (`sendPoll` options was widened to 1–12 in Bot API 10.0). One gap versus the real schema: this validates count and quiz-answer bounds but not per-string length — Telegram also requires each option text to be 1–100 chars and the poll `question` to be 1–300 chars (quiz `explanation` similarly capped). As written, an over-length option/question is accepted here and only rejected remotely by the Bot API, which is the exact failure mode this PR is meant to move local. Not fixing this myself since it widens scope (and the existing UTF-16 truncation helper in this file suggests length here should probably be measured the same way) — worth a follow-up check on whether it's in scope for this PR.

---

houko (commented):
Mechanical CLAUDE.md-compliance pass: validation logic (option-count bound 1–12, quiz-answer range check) matches the current Telegram Bot API and is correct. Two findings left as inline comments, neither auto-fixed (one requires history rewrite/force-push which is out of scope for this pass, the other widens scope beyond a lint fix). No changes pushed.

---

changelog.d/fixed/6887-telegram-poll-validation.md
houko: The current branch tip, commit `9244904a` ("docs: fix changelog fragment format for #6887"), which touched this file, has its git author identity set to `Claude <noreply@anthropic.com>` (`git log` on this branch). CLAUDE.md's "No AI / Claude attribution" rule and the repo's own `commit-msg` hook explicitly cover author identity, not just message text: "separately rejects a commit whose *author identity* ... resolves to Claude / Anthropic even when the message itself is clean." This slipped through here (client-side hooks aren't enforced server-side). Fixing it requires rewriting this commit (amend with a human author + force-push), which an automated reviewer shouldn't do unilaterally — flagging for the maintainer to rebase/re-author before merge.

---

changelog.d/fixed/6887-telegram-poll-validation.md
houko, line 1: Commit `9244904` on this branch ("docs: fix changelog fragment format for #6887") has its author identity set to `Claude <noreply@anthropic.com>`.
Per `CLAUDE.md` ("No AI / Claude / Anthropic attribution in commit messages, PR bodies, or issue comments") and the repo's `commit-msg` hook, which explicitly rejects "a commit whose author identity … resolves to Claude / Anthropic even when the message itself is clean," this commit should not have landed with that identity.
This is a history-rewrite (author-identity amend / rebase) rather than a forward fix, so I'm leaving it as a review comment rather than fixing it myself — force-pushing a rewritten history onto someone else's branch mid-review is out of scope for a mechanical pass.
Suggested fix: `git rebase -i` (or `git commit --amend`) that commit with `git commit --amend --author="houko <correct-email>"` (or squash it into the preceding commit), then force-push `fix/telegram-poll-validation` — with the PR author's OK, since it rewrites already-pushed history.

---

sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs
houko, line 1: Compliance note (not a code-quality finding): the current branch tip commit `1591249a` ("fix(telegram): correct poll option bounds to Telegram's actual 2-10 range") is authored as `Claude <noreply@anthropic.com>`.
This is a different commit from the one already flagged earlier in this thread (`9244904a`), and it is now the actual mergeable HEAD.
The code change itself looks correct — the 2-10 bound in `validated_poll` now matches the doc comment on `ChannelContent::Poll` in `crates/librefang-channels/src/types.rs` and the Bot API's real limits.
Per CLAUDE.md, commit author identity resolving to Claude/Anthropic is disallowed regardless of message text; this needs a maintainer rebase with `--reset-author` (or a squash on merge) before landing.
Not rewriting it myself since that requires a force-push.

---

changelog.d/fixed/6887-telegram-poll-validation.md
houko, line 1: Self-flagging: the commit I just pushed to this branch (`4efe20c5046c8c0a3745f57d4457b30b327a7b6b`, "fix(telegram): enforce Bot API poll question/option/explanation length bounds") landed with author identity `Claude <noreply@anthropic.com>` — the review session's git identity was never set away from the container's default, so it wasn't caught before pushing.

This is the same problem I found as a pre-existing pattern on #6904, #6906, and #6908 (each has 1-2 commits with the same author identity from earlier automated passes), so it isn't specific to this PR — it looks like the automation environment's global `git config user.name`/`user.email` defaults to the Anthropic identity and nothing overrides it before committing. Content-wise the fix itself is correct (added local length validation for Bot API poll question/option/explanation bounds, 4 tests passing, `cargo clippy` clean) — this is purely a git-metadata issue, and I can't correct it without a force-push, which I'm not doing here. Flagging so it isn't missed before merge, and so the environment's default identity gets fixed for future automated runs.

---

houko (commented):
Follow-up on the fix I pushed to this branch (commit `4efe20c5046c8c0a3745f57d4457b30b327a7b6b`): added local length validation for the Bot API's poll question (1-300), option (1-100), and explanation (0-200) UTF-16 code-unit bounds, which the original PR's `validated_poll` didn't cover (it only checked option count and quiz-answer index). Verified with `cargo check`, `cargo clippy -- -D warnings`, and `cargo test` (4/4 passing) inside the sidecar's own workspace. One finding left inline: my own commit landed with an AI author identity due to the environment's git config, same pattern as the other PRs in this batch.

---

houko (commented):
Automated review pass for PR #6887 (head `4efe20c5`).

**Findings:**

1. **Blocking (pre-existing, already flagged repeatedly in this thread):** the branch tip commit `4efe20c5` (and two earlier commits, `9244904a`, `1591249a`) carry author identity `Claude <noreply@anthropic.com>`, which CLAUDE.md and the repo's `commit-msg` hook disallow (the hook's author-identity check, separate from message-text scanning). This has already been raised multiple times on this exact PR with no human response — not repeating the full explanation again, just confirming it is still unresolved as of the current head sha and remains a merge blocker. Needs a maintainer rebase (`git rebase -i` with `--reset-author`, or a squash-on-merge) and a force-push to the branch — outside what an automated pass should do unilaterally.
2. **PR body is stale relative to the shipped code.** The Summary says "require 1-12 string poll options before any Bot API request," but the option-count bound landed in commit `1591249a` at 2-10, matching Telegram's actual `sendPoll` limits and the doc comment on `ChannelContent::Poll` in `crates/librefang-channels/src/types.rs` ("Answer option texts (2–10 options for Telegram)"). Worth updating the description so it doesn't mislead reviewers or anyone reading it via the changelog/release pipeline about what actually shipped.

**Verified clean:**
- `changelog.d/fixed/6887-telegram-poll-validation.md` follows the documented fragment format (no leading `- `, ends `(#6887) (@houko)`).
- `validated_poll` / `validated_poll_question` / `validated_poll_explanation` in `sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs` correctly gate on Telegram's real `sendPoll` bounds (2-10 options, 1-100/1-300/0-200 UTF-16 code units via the existing `utf16_len` helper) and reject non-string options, missing/out-of-range/negative/overflowing `correct_option_id`, and missing quiz answers before any request is issued.
- Regression tests cover malformed and boundary-value payloads (empty/oversize option lists, 1/11-option counts, non-integer and overflowing `correct_option_id`, min/max length bounds at each field), not just the happy path.
- CI Gate and all shards are green on the current head sha.




---

sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs
houko, line 1: Policy violation, not a code-correctness issue: three commits on this branch are both authored and committed as `Claude <noreply@anthropic.com>`:
- `4efe20c5046c8c0a3745f57d4457b30b327a7b6b` — fix(telegram): enforce Bot API poll question/option/explanation length bounds
- `1591249a57ff0937ed62950ce8db38dd743e9cd3` — fix(telegram): correct poll option bounds to Telegram's actual 2-10 range
- `9244904aa32068524edac751effd9a7776ec5b2e` — docs: fix changelog fragment format for #6887

This is exactly the pattern the repo's `commit-msg` hook is meant to catch (author-identity check via `git var GIT_AUTHOR_IDENT`), but these commits landed on the branch without going through that hook. Per CLAUDE.md ("No Claude / Anthropic / AI attribution in commit messages... the rule exists because attribution pollutes `git log` and signals provenance the project does not want to imply"), these need to be re-authored under the actual contributor's identity before merge. Flagging rather than rewriting history myself, since force-push/history-rewrite on someone else's branch is out of scope for an automated review pass.

---

sdk/rust/librefang-sidecar-telegram/src/dispatcher.rs
```diff
@@ -38,6 +47,91 @@ fn required_coordinate(payload: &Value, key: &str) -> Result<f64> {
...
+            let raw = value.as_u64().ok_or_else(|| {
+                Error::Other("Poll.correct_option_id must be a non-negative integer".into())
+            })?;
+            Some(u32::try_from(raw).map_err(|_| {
+                Error::Other("Poll.correct_option_id exceeds the supported integer range".into())
+            })?)
+        }
+        None => None,
+    };
+    if is_quiz && correct.is_none() {
```
houko, line 128: Inconsistent null-handling versus the sibling `explanation`/`options` validation in this same function.
`explanation` and `options` both go through `.and_then(Value::as_str)` / `.and_then(Value::as_array)`, so an explicit JSON `null` is treated the same as an absent key (falls through to `None`/the "missing" error path).
`correct_option_id` instead matches on the raw `payload.get(...)`, so `Some(&Value::Null)` lands in the `Some(value)` arm and `value.as_u64()` fails, producing "Poll.correct_option_id must be a non-negative integer" for an explicit `null` on a non-quiz poll — where the old code silently treated it as absent.

None of the in-repo producers (`sdk/python/.../protocol.py`, `sdk/rust/librefang-sidecar/src/protocol.rs`, `ChannelContent::Poll` in `crates/librefang-channels/src/types.rs`, all `skip_serializing_if = "Option::is_none"`) currently emit an explicit `null` here, so this likely isn't hit by first-party callers today. But since this validation sits at a boundary that also has to tolerate arbitrary/malformed tool-call-constructed JSON (that's the stated point of the PR), an external caller that always includes the key and serializes `None` as `null` (a common pattern) would get a new hard failure that didn't exist before this change.

Worth either matching the same `.and_then(...)` shape used for `explanation`/`options` (treat `null` as absent) or, if strict-null-is-an-error is intentional, a one-line comment saying so plus a test case for it — there's currently no test covering `correct_option_id: null`.

---

houko (commented):
Automated review pass against merge-base `3362202f` (matches the stated base).

Verified: `cargo check` and `cargo clippy --all-targets -- -D warnings` are clean for `librefang-sidecar-telegram`, and the new/existing unit tests pass (`cargo test -p librefang-sidecar-telegram -- poll`: 4/4 ok).
The core change is a real, correct fix: the old `"Poll"` arm silently dropped non-string options (`filter_map`), defaulted to an empty options array when the key was missing (would have produced a Bot-API-rejected 0-option poll), and truncated an out-of-range `correct_option_id` via `as u32` instead of erroring.
The new `validated_poll_question` / `validated_poll` / `validated_poll_explanation` helpers close all of those, reuse the existing `utf16_len` helper instead of duplicating UTF-16 counting logic, and the changelog fragment follows the required `changelog.d/fixed/` format.

Two things flagged inline:
1. Three commits on this branch are authored/committed as `Claude <noreply@anthropic.com>`, which violates this repo's no-AI-attribution policy — needs re-authoring before merge.
2. A minor inconsistency in how `correct_option_id: null` is handled versus `explanation`/`options` (see inline comment for detail).

Also FYI for whoever merges this batch: `dispatcher.rs` is touched by three PRs sharing this base (`#6887`, `#6880`, and the `InlineKeyboardAction` change riding in the `#6877` diff) — the edited regions look non-overlapping (Poll validation vs. `send_text` doc-comment vs. `build_inline_keyboard`), so a plain merge should be fine, but worth a fresh CI run on whichever lands last per the "two green PRs" guidance in CLAUDE.md.

---

PR #6910, opened 2026-08-10
refactor(kernel): tighten identity registry insertion

houko:
## Summary

- use DashMap Entry directly for atomic first-write-wins registration without a redundant pre-read
- construct identity timestamps only for the winning Vacant insertion and persist only that winner
- use the crate's non-poisoning parking_lot mutex for serialized disk snapshots
- assert registration returns and add a 16-thread same-name persistence regression

This resolves the remaining confirmed maintenance findings from the `agent_identity_registry.rs` OCR report. The separate best-effort durability claim remains classified as low-risk PARTIAL because the runtime intentionally retains in-memory mutations during transient disk failure.

## Verification

- concurrent regression repeated 20 times
- `cargo test -p librefang-kernel --lib agent_identity_registry::tests` (9/9)
- `cargo test -p librefang-kernel` (all unit, integration, and doc tests passed; documented ignores only)
- `cargo fmt --check`
- `git diff --check origin/main...HEAD`
- independent re-review: no Critical, Important, or Minor findings

### DIFF
--- a/changelog.d/fixed/agent-identity-registry-followups.md
+++ b/changelog.d/fixed/agent-identity-registry-followups.md
@@ -1,1 +1,0 @@
-- Simplify canonical agent identity registration and remove silent mutex-poison recovery. (@houko)
--- a/changelog.d/fixed/agent-identity-registry-followups.md
+++ b/changelog.d/fixed/agent-identity-registry-followups.md
@@ -1,0 +1,1 @@
+- Simplify canonical agent identity registration and remove silent mutex-poison recovery. (@houko)
--- a/crates/librefang-kernel/src/agent_identity_registry.rs
+++ b/crates/librefang-kernel/src/agent_identity_registry.rs
@@ -28,11 +28,12 @@
 //! edit by hand, but it is human-readable for emergency surgery.
 
 use chrono::{DateTime, Utc};
+use dashmap::mapref::entry::Entry;
 use dashmap::DashMap;
 use librefang_types::agent::AgentId;
+use parking_lot::Mutex;
 use serde::{Deserialize, Serialize};
 use std::path::{Path, PathBuf};
-use std::sync::Mutex;
 use tracing::{debug, warn};
 
 /// File name used inside `home_dir`.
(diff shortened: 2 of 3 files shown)


changelog.d/fixed/agent-identity-registry-followups.md
```diff
@@ -0,0 +1 @@
+- Simplify canonical agent identity registration and remove silent mutex-poison recovery. (@houko)
```
houko: changelog.d/README.md convention: this fragment's filename should lead with the PR/issue number so fragments sort in a useful order ("lead it with the PR or issue number") — suggest renaming to `6910-agent-identity-registry-followups.md`. The bullet text is also missing the trailing `(#6910)` reference before `(@houko)`; without it, `cargo xtask release` will still generate a separate `- <PR title> (#6910) (@houko)` line, so this entry would appear twice in the release notes.

---

houko (commented):

---

crates/librefang-kernel/src/agent_identity_registry.rs
```diff
@@ -180,28 +181,24 @@ impl AgentIdentityRegistry {
...
     /// momentarily wedged.
     pub fn register_if_absent(&self, name: &str, canonical_uuid: AgentId) -> AgentId {
-        if let Some(existing) = self.map.get(name) {
-            return existing.canonical_uuid;
-        }
-        let entry = AgentIdentityRecord {
-            canonical_uuid,
-            created_at: Utc::now(),
+        let (final_uuid, inserted) = match self.map.entry(name.to_string()) {
+            Entry::Occupied(existing) => (existing.get().canonical_uuid, false),
```
houko, line 185: Non-blocking design note: `register_if_absent`'s common case — name already registered, hit on every routine respawn — now always goes through `DashMap::entry()`, which takes an exclusive per-shard write lock and allocates `name.to_string()`, even when the arm is `Entry::Occupied`.
The prior code's `self.map.get(name)` fast path took a shared read lock with no allocation for exactly this case, only falling into `entry()` for the actual (rare) insert attempt.
Worth restoring that fast path ahead of the `entry()` call — it keeps the atomic first-write-wins insert behavior and the new `inserted`-gated `persist()` while not adding write-lock contention plus an allocation to the hot "already registered" path:
```rust
if let Some(existing) = self.map.get(name) {
    return existing.canonical_uuid;
}
match self.map.entry(name.to_string()) {
    Entry::Occupied(existing) => existing.get().canonical_uuid,
    Entry::Vacant(vacant) => { /* persist only here, as already done */ }
}
```
Doesn't affect correctness — the new `concurrent_registration_of_same_name_persists_one_winner` test passes either way — but the PR description's "redundant pre-read" framing discards a real optimization along with the actual race-window comment it was fixing.

---

houko (commented):
Automated daily review pass. No blocking findings — commit authorship is clean (no Claude/Anthropic attribution), the earlier changelog fragment naming/reference issue was already fixed in the second commit, the `.identity/` workspace migration path is untouched (this file is the separate `agent_identities.toml` registry), and the `register_if_absent` atomicity fix plus the 16-thread regression test are correct. Left one non-blocking design note inline about a lock/allocation trade-off in the "already registered" hot path.




---

PR #6911, opened 2026-08-10
fix(runtime): harden proactive memory handle state

houko:
## Changes
- replace the proactive-memory kernel-handle slot with the crate existing non-poisoning `parking_lot::Mutex`
- release the slot guard after upgrading the weak handle and before calling kernel methods
- preallocate the extraction prompt buffer and write message lines without per-message `format!` temporaries
- add a regression test proving the handle slot remains usable after a holder panics

## Tests
- RED: new panic recovery test failed to compile before the lock helper/implementation existed
- `cargo test -p librefang-runtime proactive_memory::tests --lib` (62 passed)
- `git diff --check`
- full `cargo test -p librefang-runtime`: 2201 passed, 7 failed, 2 ignored; all 7 failures are pre-existing `model_catalog` EveryAPI expectations and reproduce from a separate worktree at the same `origin/main` without this change

## Out of scope
- The report claim that `decide_action` receives an unbounded candidate list is not reachable: its production caller truncates to 5 before dispatch.
- Provider separators and fuzzy plural semantics follow their documented configuration contracts; no behavior change is made for those refuted findings.
- The unrelated EveryAPI baseline failures introduced by the current model snapshot are not changed here.


crates/librefang-runtime/src/proactive_memory.rs
```diff
@@ -1092,6 +1098,24 @@ mod tests {
...
+            "test-model".to_string(),
+        );
+
+        let panic = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+            let _slot = extractor.lock_kernel_handle_slot();
+            panic!("poison the kernel handle lock");
+        }));
+        assert!(panic.is_err());
+
+        assert!(extractor.lock_kernel_handle_slot().is_none());
```
houko, line 1116: Nice regression test for the poisoning class of bug — this is exactly the kind of thing a "harden state" PR should cover.

One optional strengthening: this only proves the slot is still *lockable* after the panic, which happens to equal the pre-panic value here since it was never set.
It would be a slightly stronger proof of "usable" to `install_kernel_handle` (or set the slot directly) after the `catch_unwind` and assert the new value round-trips, so the test can't accidentally pass on a slot that silently stopped accepting writes but still reads as `None`.
Not blocking — judgment call, leaving as a suggestion rather than a fix.

---

houko (commented):
Automated daily review pass against CLAUDE.md. Code changes look correct and well-scoped — no fixes applied.

Findings:
- The `parking_lot::Mutex` swap and the panic-poisoning regression test are a genuine fix: the old `std::sync::Mutex` + `.lock().ok()` pattern would silently and permanently disable `install_kernel_handle`/`echo_policy_for`/`resolve_model_for_agent` after any panic elsewhere while the lock was held, since a poisoned std mutex never un-poisons. `parking_lot` is already the crate's documented convention for exactly this reason (see `crates/librefang-runtime/CLAUDE.md`: "No synchronous ... `std::sync::RwLock` inside async handlers. Use ... `parking_lot`"), so this aligns with established practice rather than introducing a new pattern. One inline suggestion left on the new test.
- Changelog fragment (`changelog.d/fixed/6911-proactive-memory-runtime-followups.md`) is correctly formatted: right section directory, no leading `- `, sentence-per-line, ends `(#6911) (@houko)`.
- No new config fields, no `HashMap`/`HashSet` on any LLM-prompt-reaching path, no route/wiring change requiring a `#[tokio::test]`, no `Option<Arc<dyn Trait>>` field needing `#[serde(skip)]` (the struct doesn't derive `Serialize`/`Deserialize`). Commit author identities are human (houko / Evan Hu), not Claude/Anthropic.
- Non-blocking nit: commit `b0638e6`'s message body is hard-wrapped at ~70 columns instead of one-sentence-per-line, which CLAUDE.md's prose-wrapping rule explicitly extends to commit message bodies ("Commit message bodies" is listed alongside docs/CHANGELOG/PR text). Fixing it now would need a history rewrite + force-push on an already-open PR, which is out of scope for this pass — flagging only, not acting on it.

No other issues found. Not approving/merging per policy — maintainer call.

---

changelog.d/fixed/6911-proactive-memory-runtime-followups.md
houko, line 1: Non-blocking nit: the head commit (`b0638e6`, "docs(changelog): lead fragment filename with PR number for #6911") has a hard-wrapped message body — it breaks mid-sentence at ~72 columns ("...to lead with the PR or\nissue number so fragments sort usefully...", "...this\none was left unprefixed by the prior fixup commit.").

CLAUDE.md's prose-wrapping rule covers commit message bodies too: no column limit, break only at sentence boundaries (one sentence per line). Not suggesting a history rewrite here — a review thread is already open on this PR, and CLAUDE.md prefers fixup commits over rewriting history once a reviewer has loaded the diff — just flagging so it isn't repeated on the next commit.

Everything else checked out: the changelog fragment itself is correctly named/sectioned/attributed, no `KernelConfig.agents`-shaped footgun was introduced (this PR doesn't touch config surfaces at all), the `parking_lot::Mutex` swap is backed by an existing `parking_lot = "0.12"` dependency in this crate's `Cargo.toml`, the lock-release-before-upgrade ordering in `kernel_handle()` genuinely drops the guard before any kernel method call (verified against `echo_policy_for` / `resolve_model_for_agent` call sites), and all three commits are authored/committed as `houko` / `Evan Hu <suzukaze.haduki@gmail.com>` — no Claude/Anthropic attribution.

---

houko (commented):
Automated daily review pass. One non-blocking nit inline; no blocking findings.



---

PR #6912, opened 2026-08-10
perf(runtime): cache link extraction regex

houko:
## Changes
- compile the invariant link-extraction URL pattern once with `std::sync::LazyLock`
- reuse the same thread-safe `regex_lite::Regex` for every message
- add a regression test that pins the shared-instance contract

## Tests
- RED: the new shared-pattern test failed before the accessor existed
- `cargo test -p librefang-runtime link_understanding::tests --lib` (19 passed)
- `git diff --check`

## Out of scope
- The SSRF literal parsing, userinfo handling, and alternate-IP findings were already fixed by #6763.
- DNS resolution is intentionally not performed by this synchronous extraction-only layer; actual network retrieval remains protected by the shared WebFetch resolver and pinned transport.

### DIFF
--- a/crates/librefang-runtime/src/link_understanding.rs
+++ b/crates/librefang-runtime/src/link_understanding.rs
@@ -2,6 +2,15 @@
 
 use tracing::warn;
 
+static URL_PATTERN: std::sync::LazyLock<regex_lite::Regex> = std::sync::LazyLock::new(|| {
+    regex_lite::Regex::new(r#"https?://[^\s<>\[\](){}|\\^`"']+[^\s<>\[\](){}|\\^`"'.,;:!?\-)]"#)
+        .expect("URL regex is valid")
+});
+
+fn url_pattern() -> &'static regex_lite::Regex {
+    &URL_PATTERN
+}
+
 /// Configuration for link understanding (re-exported from types).
 pub use librefang_types::media::LinkConfig;
 
@@ -20,16 +29,10 @@ pub struct LinkSummary {
 /// Returns up to `max` valid, unique URLs that do not contain private literal targets or restricted hostnames.
 /// Any later fetch must still use the shared WebFetch SSRF resolver and pinned transport to validate DNS at connect time.
 pub fn extract_urls(text: &str, max: usize) -> Vec<String> {
-    // Simple but effective URL regex
-    let url_pattern = regex_lite::Regex::new(
-        r#"https?://[^\s<>\[\](){}|\\^`"']+[^\s<>\[\](){}|\\^`"'.,;:!?\-)]"#,
-    )
-    .expect("URL regex is valid");
-
     let mut seen = std::collections::HashSet::new();
     let mut urls = Vec::new();
 
-    for m in url_pattern.find_iter(text) {
+    for m in url_pattern().find_iter(text) {
         let url = m.as_str().to_string();
 
         // Deduplicate
@@ -155,6 +158,11 @@ pub fn build_link_context(text: &str, config: &LinkConfig) -> Option<String> {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn url_pattern_is_shared_across_extractions() {
+        assert!(std::ptr::eq(url_pattern(), url_pattern()));
+    }
 
     #[test]
     fn test_extract_urls_basic() {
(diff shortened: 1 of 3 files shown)


crates/librefang-runtime/src/link_understanding.rs
```diff
@@ -156,6 +159,11 @@ pub fn build_link_context(text: &str, config: &LinkConfig) -> Option<String> {
 mod tests {
     use super::*;
 
+    #[test]
+    fn url_pattern_is_shared_across_extractions() {
+        assert!(std::ptr::eq(url_pattern(), url_pattern()));
```
houko, line 164: Nit, non-blocking: `std::ptr::eq(url_pattern(), url_pattern())` compares the function's return value to itself within a single test.
Since `url_pattern()` always returns `&URL_PATTERN`, this assertion is trivially true regardless of whether the underlying regex is actually cached (e.g. it would still pass if `url_pattern()` were changed to build a fresh `Regex` on every call, as long as it kept returning a `'static` reference to the same location).
A stronger regression guard would assert that two calls to `extract_urls` don't allocate/compile a new `Regex` — for example by checking the `LazyLock`'s init only runs once (a counter in the init closure) — or simply drop this test, since the caching contract is really guaranteed by the `static`/`LazyLock` type signature itself and caught at compile time.

---

houko (commented):
Automated review pass.

Checked: commit author identities (both `Evan Hu` / `houko` at `suzukaze.haduki@gmail.com` — no Claude/Anthropic attribution, no `commit-msg` hook violation), regex correctness (the `LazyLock`-cached pattern is byte-identical to the previously per-call-compiled one, no flag/behavior drift), thread-safety (`std::sync::LazyLock` is `Sync`, no `static mut`, and matches the established pattern already used elsewhere in this codebase — e.g. `librefang-api/src/oauth.rs`, `librefang-kernel/src/hooks.rs`, `librefang-runtime/src/agent_loop/retry.rs`), changelog fragment (present at `changelog.d/fixed/6912-link-understanding-regex-cache.md`, correctly formatted, references `(#6912) (@houko)`), and deterministic-prompt-ordering exposure (the `HashSet` in `extract_urls` is used only for dedup membership checks — output order comes from the linear `find_iter` scan over the input text, not `HashSet` iteration, so no `BTreeSet`/ordering fix is needed here).

One minor, non-blocking nit inline on the new test's assertion strength.

Separately (informational, not actionable via a code fix): the second commit's message body ("docs(changelog): fix fragment format for #6912") is hard-wrapped across multiple lines mid-sentence, which is itself a CLAUDE.md prose-wrapping violation for commit message bodies (one sentence per line, no column-limit wrapping) — noting it for awareness since fixing it would require rewriting already-pushed commit history (amend + force-push), which is out of scope for this pass.

No blocking findings.




---

crates/librefang-runtime/src/link_understanding.rs
houko, line 164: The caching mechanism itself looks correct: `URL_PATTERN` has no dependency on `LinkConfig` or any other input, so there's no staleness concern, and `regex_lite::Regex` is `Send + Sync` (it holds its per-thread execution state behind an internal pool), so sharing it via `LazyLock` across concurrent callers in `extract_urls` is sound. Matching behavior is unchanged — same pattern string, same `find_iter` call, just compiled once instead of per-call.

One nit on the regression test: `assert!(std::ptr::eq(url_pattern(), url_pattern()))` is tautological given `url_pattern()`'s signature (`-> &'static regex_lite::Regex` always returns a reference into the same static, whether or not the underlying value is actually cached/lazy). It would pass even if `URL_PATTERN` were, hypothetically, re-derived per-call rather than lazily-initialized-once, as long as it stayed a `'static` binding. If the intent is to pin "compiled once" behavior, a stronger check would assert something observable about compilation happening only once (e.g. wrapping construction with a call counter in a test-only build, or just relying on the type system + `LazyLock`'s documented guarantee rather than a runtime test). Not blocking — the underlying fix is fine either way.

---

houko (commented):
Automated review pass. Correctness check on the caching mechanism: pattern is invariant (no config dependency), `regex_lite::Regex` is thread-safe to share via `LazyLock` (internal per-call cache pool, same design as the `regex` crate), and matching behavior is byte-identical to the uncached version — same pattern string, same `find_iter` usage. No staleness or data-race concerns found.

Checked `git log origin/main..HEAD` for both commits on this branch for the recurring Claude/Anthropic attribution issue seen on sibling PRs in this run — both commits are cleanly attributed to human authors (author/committer both "houko" and "Evan Hu", no bot identity), so no attribution violation here.

Left one non-blocking nit inline on the added test.

---

PR #6914, opened 2026-08-10
fix(dashboard): validate manifest mode fields

houko:
## Changes
- require non-blank cron expressions for periodic schedules
- validate JSON Schema response formats against the object/boolean domain that TOML serialization can preserve, rejecting null and lossy numeric values
- automatically open invalid collapsible sections and expose localized errors through ARIA attributes
- clear duplicate tag submissions and provide an accessible name for the stream-thinking toggle
- simplify the redundant schedule parser branch

## Tests
- RED: cron/schema validation tests failed before implementation
- RED: accessibility tests failed before invalid sections and controls exposed feedback
- RED: unsafe-integer and overflowing-number schemas passed before raw-token validation
- RED: duplicate-tag and unnamed-toggle regressions failed before the compact-control fixes
- `pnpm test --run` (94 files, 990 tests passed)
- `pnpm lint`
- `pnpm typecheck`
- `pnpm build`
- `git diff --check`

## Review
Independent review found no remaining Critical, Important, or Minor issues and marked both implementation commits Ready to merge.


crates/librefang-api/dashboard/src/lib/agentManifest.ts
```diff
@@ -753,12 +750,74 @@ const stringifyOrEmpty = (value: unknown): string => {
...
   if (!form.name.trim()) errors.push("name");
   if (!form.model.provider.trim()) errors.push("model.provider");
   if (!form.model.model.trim()) errors.push("model.model");
+  if (form.schedule.mode === "periodic" && !form.schedule.cron.trim()) {
+    errors.push("schedule.cron");
+  }
+  if (form.response_format.mode === "json_schema") {
+    const schema = form.response_format.schema.trim();
+    if (!schema || parseSupportedJsonSchema(schema) === undefined) {
+      errors.push("response_format.schema");
```
houko, line 816: Both new checks here (empty `cron` for a periodic schedule, and the JSON-Schema shape/number-safety checks in `parseSupportedJsonSchema`) are enforced only in this validator, which the create-agent form calls before `spawnMutation.mutate({ manifest_toml })` in `AgentsPage.tsx`. That's the right place for UX, but a manifest can also reach the daemon by other paths — `PUT`/raw `manifest_toml` via `PATCH /api/agents/{id}` (`routes/agents/lifecycle.rs`), or an operator hand-editing `agent.toml` — none of which go through this TS validator.

On the Rust side today:
- `ScheduleMode::Periodic { cron: String }` (`librefang-types/src/agent.rs`) has no non-empty check. An empty cron reaching `parse_cron_to_secs` (`librefang-kernel/src/background.rs:668`) falls through to the "unparseable" branch and silently defaults to a 300s interval with a `warn!` log — it doesn't reject.
- `ResponseFormat::JsonSchema { schema: serde_json::Value, .. }` (`librefang-types/src/config/types.rs`) accepts any JSON value, including `null` or non-object/boolean schemas, with no server-side shape check.

Neither is a crash or security hole, but per the repo's "client-side-only validation is not a real security/correctness boundary" guidance, is a server-side guard (even just rejecting on manifest parse/apply) intentionally out of scope for this PR, or worth a follow-up issue? Given this PR is scoped to the dashboard, I'm flagging rather than fixing — the fix would live in `librefang-types`/`librefang-kernel`, a different crate than the rest of this change.

---

houko (commented):
Reviewed the dashboard manifest-validation changes: the new client-side checks (cron-required for periodic schedules, JSON-Schema shape/number-safety for json_schema response format) are well-tested and correctly routed through the existing `validateManifestForm` → `spawnMutation` flow — no inline fetch added. Accessibility wiring (aria-invalid/aria-describedby/aria-required, auto-opening invalid sections) checks out against the new tests. Left one comment on whether the equivalent constraints should also be enforced server-side, since manifests can reach the daemon without going through this validator.

---

changelog.d/fixed/6914-agent-manifest-validation-followups.md
houko, line 1: Non-blocking nit, not covered by the earlier review on this PR: commit `acf33fcdc78ea88eb7cbfa4704a073e5a39cbbdd` ("docs: fix changelog fragment format for #6914") has a hard-wrapped message body ("...fragments sort\nusefully, unwrap the hard-wrapped body..."), breaking mid-sentence rather than at sentence boundaries. CLAUDE.md's prose-wrapping rule covers commit message bodies too (only the subject line gets the git-display-truncation exception). Not asking for a history rewrite here — flagging only, since the fix would need an amend + force-push on an already-open PR.

Everything else looks solid: the changelog fragment itself is correctly named/sectioned/formatted, the validation logic (`parseSupportedJsonSchema`, `hasUnsupportedJsonNumber`) correctly detects precision-losing numbers by scanning the raw JSON text rather than the already-rounded parsed value, and the `TagInput` fix now clears the input on a duplicate submission instead of only on a successful add.

---

PR #6744, opened 2026-08-07
fix(kernel): wake the assignee of a posted task without requiring a declared trigger

nevgenov:
A task addressed to an agent reached that agent only when an operator had separately registered a matching `task_posted` trigger.
Nothing in the kernel ever created one — `background::parse_condition` cannot produce the pattern and `reconcile_manifest_triggers` only materializes what an operator declared — so delivery was entirely operator-supplied, and with no record the task sat `pending` indefinitely with nothing in the log.

Implements Option A from #6728 on the four decisions settled in the issue thread.

## What changed

**The wake** (`kernel/triggers_and_workflow.rs`).
`synthesize_assignee_wake` runs after the stored-trigger pass and appends at most one `TriggerMatch` to the list the dispatcher already consumes, so it inherits the global trigger lane, the per-agent semaphore, the per-fire timeout, sequential ordering within an event, and the `PUBLISH_EVENT_DEPTH` cycle guard rather than re-implementing any of them.
Nothing is persisted: no record reaches `trigger_jobs.json` or `trigger list`, so disabling the knob is a complete revert.

**Coverage** (`triggers.rs::task_posted_coverage_for`).
A stored record covers an assignee when it would match a `TaskPosted` addressed to them *and* dispatches to them — owner or `target_agent`.
The identity test reuses `agent_identity_filter_matches`, the same helper `matches_pattern` uses, against both forms the substrate accepts in `assigned_to`, so coverage cannot drift from what actually fires.
It scans `self.triggers` rather than `agent_triggers` because that index is owner-keyed and would miss an orchestrator-owned trigger targeted at the assignee; the set is bounded by `MAX_TRIGGERS_PER_AGENT` per agent and the scan runs once per `TaskPosted`, which is cheaper than keeping a second index consistent across register / update / remove / re-key.

**`TriggerMatchSource`** replaces `TriggerMatch.trigger_id`, so a log line can say *why* an agent was woken when there is no trigger to name, and `SessionMode::New` has a stable key to derive from — `SessionId::for_task_wake(agent, task_id, fire_time)`, mirroring `for_trigger_fire` with its own v5 namespace.
Dispatch log fields on the workflow path change from `trigger_id = …` to `source = …` as a result; the `Display` impl renders `trigger:<uuid>` or `assignee_wake:<task_id>`.

[...]

Issue #5115, opened 2026-05-16
Triggers: cooldown wedges all triggers on wall-clock backstep


houko:
## Problem
Trigger cooldown computation uses `(now - *last).to_std().unwrap_or(Duration::ZERO)`. When `last_fired_at > now` (wall-clock backstep or a persisted timestamp from a misconfigured host), `to_std()` errors and `unwrap_or(ZERO)` collapses the elapsed delta to `0`. With `elapsed = 0 < cooldown`, every gated trigger is silently suppressed until wall-clock catches up to the persisted `last_fired_at`.

## Evidence
`crates/librefang-kernel/src/triggers.rs:736-746`:

```rust
let elapsed = match (now - *last).to_std() {
    Ok(d) => d,
    Err(_) => Duration::ZERO,
};
if elapsed < cooldown { return; /* gated */ }
```

`last` is loaded from persisted state (`MemorySubstrate`).

## Impact
- NTP rollback or DST-aware rewrite by a poorly written ops tool wedges every trigger cluster-wide for the duration of the skew window.
- A laptop's local-disk DB carried to a server with a slower clock kills all triggers until skew clears.
- No warn log fires on the silent suppression, so operators see "triggers stopped working" with no diagnostic.

## Suggested fix
On `to_std() == Err`, log a warn (with `now`, `last`, agent_id, trigger_id) and treat as elapsed-exceeded (`Duration::MAX`) so the trigger fires once and resets `last_fired_at`. Or clamp the stored `last_fired_at` to `min(now, last_fired_at)` at load time.

PR #6918, opened 2026-08-10
fix(kernel): scope the trigger cooldown to the event's subject

### LINKED ISSUE #6756: Trigger cooldown discards events for distinct subjects instead of deferring them
A trigger's cooldown discards events for *distinct subjects* as though they were a repeat of one signal, so the second event is lost rather than delayed.

Split out of #6728 at the maintainer's request: the task board is where it cost us work, but nothing about the defect is task-board-specific.

## Reproduction

Self-contained, no daemon and no LLM. The second assertion fails on `main`:

```rust
use librefang_kernel::triggers::{TriggerEngine, TriggerPattern};
use librefang_types::agent::AgentId;
use librefang_types::event::{Event, EventPayload, EventTarget, SystemEvent};

fn completion(task_id: &str, creator: AgentId) -> Event {
    Event::new(
        AgentId::new(),
        EventTarget::Broadcast,
        EventPayload::System(SystemEvent::TaskCompleted {
            task_id: task_id.to_string(),
            completed_by: "worker".to_string(),
            result: format!("result of {task_id}"),
            created_by: Some(creator.to_string()),
        }),
    )
}

#[test]
fn two_completions_inside_the_cooldown_window_notify_once() {
    let engine = TriggerEngine::new(); // default cooldown_secs = 5
    let orchestrator = AgentId::new();
    engine
        .register(
            orchestrator,
            TriggerPattern::TaskCompleted { creator_match: None },
            "notify the human: {{event}}".to_string(),
            0,
        )
        .unwrap();

    // Two DISTINCT tasks finishing back to back — what any drain loop produces.
    let (first, _) = engine.evaluate_with_resolver(&completion("task-1", orchestrator), |_| None);
    let (second, _) = engine.evaluate_with_resolver(&completion("task-2", orchestrator), |_| None);

    assert_eq!(first.len(), 1, "first completion notifies");
    assert_eq!(
        second.len(),
        1,
        "SECOND COMPLETION LOST: distinct task, no notification"
    );
}
```

Registering the same trigger with `cooldown_secs = Some(0)` makes both pass, which isolates cooldown rather than anything about the events themselves.

## Root cause

`evaluate_with_resolver` (`crates/librefang-kernel/src/triggers.rs:941-980`) resolves the window and `continue`s past the match when it has not elapsed:

```rust
if elapsed < cooldown {
    debug!(trigger_id = %trigger.id, "Trigger skipped (cooldown active)");
    continue;
}
```

The event is **discarded**, not deferred: no `TriggerMatch` is produced, `fire_count` does not move, and nothing re-announces the event later.
The window is keyed on the trigger alone, so the engine cannot distinguish "this same signal fired twice" from "two different things happened one second apart".

The premise is visible in the change that introduced it, #1656: its test plan reads *"Rapid events don't fire same trigger multiple times within cooldown"*, and its stated goal was preventing cascading trigger chains from exhausting resources.
Cascade protection is now also covered by the separate `PUBLISH_EVENT_DEPTH` depth limit (`triggers.rs` / `kernel/triggers_and_workflow.rs`), so cooldown is no longer the only thing standing between the system and a runaway chain.

## Why it matters beyond one pattern

Every one of the 14 `TriggerPattern` variants is evaluated through this path, and several are naturally bursty over distinct subjects:

- `TaskPosted` / `TaskClaimed` / `TaskCompleted` — one event per task; a drain loop or a batch delegation produces several within a second.
- `MemoryKeyPattern` / `MemoryUpdate` — one event per key.
- `AgentSpawned` — one per agent during a boot wave.
- `ContentMatch` / `SystemKeyword` — arbitrary distinct payloads.

For a notification trigger the cost is a missing message.
For the task board it is worse: the event is the only announcement that durable work exists, so a dropped one means the work is never picked up. Losing it is silent — the drop is a `debug!` line, which a default file logger does not carry.

## Field measurements

From a production install, by `fire_count` arithmetic on an orchestrator's unfiltered `TaskCompleted` trigger:

| Completions | Timestamps (UTC) | `fire_count` delta |
|---|---|---|
| 3 | `11:55:04`, `11:56:54`, `11:56:54` | +2 |
| 2 | `15:13:14`, `15:13:14` | +1 |
| 2 | `17:26:07`, `17:26:07` — same trigger, `cooldown_secs = 0` | **+2** |

The third row is the A/B: identical same-second pattern, cooldown removed, both delivered.

The wake path, same install, same day: three tasks posted to one worker within 2.6 seconds (`17:25:02.325`, `17:25:04.893`, `17:25:04.910`) moved its `TaskPosted` trigger's `fire_count` by exactly 1.
The single activation drained two of them; the third stayed `pending` with no log line above `debug` to say why.

## Possible directions

Both preserve the storm protection the knob exists for:

1. **Defer instead of discard.** Mark the trigger dirty and fire once when the window expires, collapsing N events into one fire that happens *after* the last one. This keeps the "at most one fire per window" bound and adds an "at least one fire after the last event" guarantee. It is the shape `client-go`'s delaying queue uses — dedupe by key, delay, never drop — and it needs no new configuration.
2. **Key the window on the event's subject** as well as the trigger, so two different tasks (or memory keys, or agents) never suppress each other while a genuine repeat still does.

(1) alone fixes the data loss; (2) additionally keeps latency low for unrelated subjects. They compose.

One caveat for (1): a deferred fire renders `{{event}}` from *some* event in the window, so the prompt has to say which one, or say that several occurred. For a coalescing fire the honest rendering is probably the last event plus a count.

I am happy to take this after the task-board work in #6744 lands, or leave it to you if you would rather not have it queued behind that — you offered, and either way suits me.

Refs #6728, #1656

nevgenov:
The cooldown window was keyed on the trigger alone, so it could not tell "the same thing fired twice" from "two things happened a second apart", and it discarded the second rather than deferring it.
`evaluate_with_resolver` produced no match, `fire_count` never moved, and nothing re-announced the event, so the work it carried was simply gone.

Reproduction from the issue, unchanged, now passing as `distinct_task_completions_inside_one_window_both_fire`.

## The rule

The window is keyed on `(trigger, subject)`, where the subject is what the event is *about*: the task id for `TaskPosted` / `TaskClaimed` / `TaskCompleted`, the memory key for `MemoryKeyPattern`.
A repeat of the same subject inside the window is still suppressed, so the protection the knob exists for is unchanged — that is pinned by `a_repeat_of_the_same_subject_is_still_suppressed`.

The subject is chosen from the **pattern**, not from whatever the event happens to carry, and patterns that name no subject keep today's trigger-wide window.
This is the load-bearing decision in the diff, so it is worth saying why rather than leaving it to be inferred: `All`, `System`, `Lifecycle` and `ContentMatch` match streams whose subjects are nearly always distinct, so keying on the subject would turn "at most once per window" into "once per event" for a trigger whose bounded firehose is the entire reason an operator set a cooldown on it.
Widening delivery for the patterns where two subjects are definitionally two units of work is the fix; widening it for a catch-all is a cost regression wearing the same clothes.
`a_catch_all_trigger_keeps_its_trigger_wide_window` fails if that distinction is dropped.

## What did not change

`last_fired_at` on disk still means "when this trigger last fired": every fire stamps the trigger-wide entry alongside the subject window it consulted, so restart recovery and the operator-facing field behave exactly as before.
Per-subject entries are in memory only, so a restart costs at most one extra fire per subject — the safe direction.

Per-subject entries are pruned once they are older than the longest window any trigger could be using, since a check that consults them can only conclude "elapsed"; the scan runs only past a size threshold, so the common path pays a length check.
Removing a trigger, or changing its pattern, now clears every window it owns rather than only the trigger-wide one — otherwise a successor would inherit suppression it never earned.

Downstream, this widens *what is delivered*, not how much runs at once: distinct subjects firing in sequence still queue behind `Lane::Trigger` and the per-agent semaphore.

## Considered and rejected

Universal deferral — marking the trigger dirty and firing once on a trailing edge — preserves the rate bound exactly and is the better long-term shape, but nothing in the kernel ticks faster than the 30s sweeps, so a 5s default cooldown would need a dedicated ~1s flush task, and a coalesced fire changes what `{{event}}` renders on every pattern at once.
That is a larger blast radius than this defect justifies, and it can be built later on top of this without undoing it.

## Verification

- `cargo check --workspace --lib` clean, `cargo fmt --all --check` clean, `cargo clippy -p librefang-kernel --all-targets -- -D warnings` clean.
- `cargo test -p librefang-kernel --lib triggers` — 87 passing, including the six new ones.
- `cargo test -p librefang-api --test task_board_assignee_wake_test --test declarative_triggers_test --test trigger_workflow_test` — 24 passing, so the #6744 wake paths are unaffected.

Both directions of the design are pinned by mutation, not just asserted:

- forcing the subject to `None` (pre-#6756 behaviour) fails the two "distinct subjects" tests;
- extending the subject to catch-all patterns fails `a_catch_all_trigger_keeps_its_trigger_wide_window` and nothing else.

Closes #6756


crates/librefang-kernel/src/triggers.rs
```diff
@@ -997,7 +1017,15 @@ impl TriggerEngine {
...
-                self.last_fired.insert(trigger.id, now);
+                // Stamp the window that was actually consulted, and the
+                // trigger-wide entry alongside it so `last_fired_at` on disk
+                // keeps meaning "when this trigger last fired" for operators
+                // and for restart recovery.
+                self.last_fired.insert((trigger.id, subject.clone()), now);
+                if subject.is_some() {
+                    self.last_fired.insert((trigger.id, None), now);
+                }
+                self.prune_expired_cooldowns(now);
```
houko: This call happens while `entry` (from `self.triggers.get_mut(&id)` at line 915) is still alive — `trigger` on lines 1018/1024 is a live borrow through that `RefMut`, i.e. the write lock on `id`'s shard of `self.triggers` is still held.

`prune_expired_cooldowns` then calls `self.triggers.iter()`:
```rust
let longest = self
    .triggers
    .iter()
    .map(|t| t.cooldown_secs.unwrap_or(self.default_cooldown_secs))
    .max()
    ...
```
`DashMap::iter()` read-locks each shard as it walks them, and it will eventually reach the shard that still holds `id`'s write guard. Same thread, same shard, write-then-read: this is exactly the self-deadlock hazard already documented a few dozen lines above in this same function —

> `DashMap::len()` read-locks every shard, so calling it inside the loop — while a `self.triggers.get_mut(&id)` `RefMut` holds that same shard's write-lock — self-deadlocks the evaluator the first time the per-event budget is exhausted

— except here it's `prune_expired_cooldowns`'s `.iter()` doing the read-lock instead of `.len()`, triggered whenever `self.last_fired.len() >= PRUNE_THRESHOLD` (4096) on a firing event, rather than at the trigger-budget branch. On a busy installation that accumulates enough per-subject cooldown entries (task-board or memory-key triggers firing on many distinct subjects), this will hang the thread evaluating the event the first time pruning kicks in — not covered by the new tests, since none of them get `last_fired` anywhere near 4096 entries.

Suggest computing `longest` from a value captured before the loop (or from `ids`/a snapshot taken before any shard lock is acquired), or moving the `prune_expired_cooldowns(now)` call to run once *after* the `for id in ids` loop exits (so no `entry` guard is held), rather than inline inside the per-trigger match arm.

---

crates/librefang-kernel/src/triggers.rs
```diff
@@ -1537,6 +1602,40 @@ pub(crate) fn tests_support_task_posted_event(task_id: &str, assigned_to: &str)
...
+            | EventPayload::System(SystemEvent::TaskCompleted { task_id, .. }) => {
+                Some(task_id.clone())
+            }
+            _ => None,
+        },
+        TriggerPattern::MemoryKeyPattern { .. } => match &event.payload {
+            EventPayload::MemoryUpdate(delta) => Some(delta.key.clone()),
+            _ => None,
+        },
+        _ => None,
```
houko, line 1653: `TriggerPattern::AgentSpawned { name_pattern }` falls through to the `_ => None` catch-all here, which keeps the trigger-wide cooldown window this PR is fixing everywhere else. But its match logic (line ~1487) is structurally identical to `MemoryKeyPattern`, which *is* subject-scoped by this PR:

```rust
TriggerPattern::AgentSpawned { name_pattern } => {
    if let EventPayload::Lifecycle(LifecycleEvent::Spawned { name, .. }) = &event.payload {
        name.contains(name_pattern.as_str()) || name_pattern == "*"
    } else { false }
}
```
vs.
```rust
TriggerPattern::MemoryKeyPattern { key_pattern } => {
    if let EventPayload::MemoryUpdate(delta) = &event.payload {
        delta.key.contains(key_pattern.as_str()) || key_pattern == "*"
    } else { false }
}
```

Both are substring/wildcard patterns over a per-event identifier (`name` vs. `delta.key`), and both can legitimately match two *distinct* subjects inside one cooldown window — e.g. a trigger with `name_pattern = "worker"` (or `"*"`) watching for new workers: if `worker-1` and `worker-2` spawn a second apart, the second spawn is exactly the "different thing happened a moment later" case the changelog describes for task-board events, and it would still be silently discarded here, since `cooldown_subject` reports `None` for `AgentSpawned` and the check falls back to the old trigger-wide window.

The changelog's rationale for leaving patterns unscoped — "they match streams whose subjects are nearly always distinct, so keying on the subject would turn 'at most once per window' into 'once per event'" — describes `All`/`System`/`ContentMatch`, but doesn't obviously apply to `AgentSpawned`: like `MemoryKeyPattern`, a narrow `name_pattern` names one recurring subject (a specific agent or naming convention), not an arbitrary firehose. Was this exclusion deliberate, or is it the same fix `MemoryKeyPattern` got, just not applied here? If deliberate, worth stating why in the doc comment above `cooldown_subject`, since the current wording ("Deliberately driven by the pattern...") reads as if the four handled cases are meant to be the complete set of subject-bearing patterns, and `AgentSpawned` looks like a fifth.

---

houko (commented):
Automated review: the subject-scoping logic itself is correct and thoroughly tested (verified it doesn't reintroduce cross-subject bleed and still suppresses genuine same-subject repeats). One coverage gap: `TriggerPattern::AgentSpawned` has the identical substring/wildcard-over-a-per-event-identifier shape as the newly-fixed `MemoryKeyPattern`, but wasn't included in `cooldown_subject`'s subject extraction, so it's still exposed to the exact bug class this PR fixes for task-board and memory-key patterns.

---

nevgenov:
Both fixed in `432abcbe`. The first one is a genuine hang I introduced, so thank you for catching it before it reached anyone.

**The prune deadlock.** Confirmed exactly as traced: `prune_expired_cooldowns` reads `self.triggers` to find the longest window in use, and it ran while `entry` — the `RefMut` from `get_mut(&id)` — was still alive, so `DashMap::iter` would eventually read-lock the shard this thread already holds a write guard on. Same hazard the function documents a few dozen lines above for `len()`, reached by a different route, and I walked into it having read that comment while writing #6744.

The prune now runs after the loop, where every `RefMut` has been dropped, gated on whether any window was actually stamped so a quiet evaluation pays nothing beyond a bool.

`crossing_the_prune_threshold_does_not_wedge_the_evaluator` walks past the 4096-entry threshold on real evaluations, so it stops returning if the call ever moves back inside the loop. Verified the honest way rather than by inspection: reintroducing the call under the `RefMut` hangs the test — 90s with no completion — and removing it again returns in 0.1s. A hang is a worse failure signal than an assertion, but it is the true one here, and the alternative was a test that could not tell the difference.

**`AgentSpawned`.** You are right, and the omission contradicted this PR's own rule rather than merely missing a case. Its match logic is the same substring-or-wildcard filter over a per-event identifier as `MemoryKeyPattern`, so two workers spawning a second apart is the same "a different thing happened" case a second task is.

Keyed on the spawned agent's id rather than its name: the pattern filters on the name, but two agents may share one, and the subject here is the agent that appeared, not the label it appeared under. `distinct_agent_spawns_inside_one_window_both_fire` dies if the arm is removed.

The boot-wave case is worth naming since it is the visible change: a `name_pattern = "*"` trigger now fires once per agent during a wave instead of once per window. That is the same trade this PR already makes for `MemoryKeyPattern` with `key_pattern = "*"`, and the fan-out stays bounded by `Lane::Trigger` and the per-agent semaphore — it widens what is delivered, not what runs at once. Drawing the line anywhere else would have been arbitrary.

Verification on `432abcbe`: `cargo check --workspace --lib` clean, `cargo fmt --all --check` clean, `cargo clippy -p librefang-kernel --all-targets -- -D warnings` clean, `cargo test -p librefang-kernel --lib triggers` 90 passing.

nevgenov:
Self-reviewed this branch adversarially before asking for another pass, and it turned up three things worth fixing. All in `a7220ef1`.

**A panic reachable from the API.** `prune_expired_cooldowns` built its horizon with `chrono::Duration::seconds(longest as i64)`, and `cooldown_secs` arrives as an unvalidated `u64` — `routes/workflows/triggers.rs` reads it with `as_u64` on both POST and PATCH, with no clamp and no bound. Past roughly 9.2e15 that call panics, on the event-dispatch thread, the first time the installation crosses the prune threshold. Past `i64::MAX` the cast wraps negative instead, the retain then drops every window including live ones, and same-subject suppression stops working installation-wide without a word. Both modes are new to this branch: the value previously only reached `std::time::Duration::from_secs`, which handles it fine. The horizon is now saturating, so an absurd cooldown means "effectively never expires" — which is what the operator asked for.

**An asymmetry the rule could not justify.** `MemoryUpdate` stayed on the trigger-wide window while `MemoryKeyPattern { key_pattern: "*" }` was scoped, though the two match exactly the same events. Two triggers with identical match sets and opposite delivery semantics is a sharper edge than either rule alone, and #6756 named that pair as one item. `MemoryUpdate` is now scoped on the key and `AgentTerminated` on the agent, which also lets the rule be stated rather than enumerated: a pattern that names *what happened* to an identifiable subject is scoped, a pattern that names a *category* of events keeps the trigger-wide window.

**A false claim in the PR body, which I have corrected above.** I wrote that restart recovery behaves exactly as before. It does not. `load` restores only the trigger-wide entry, and a subject-scoped pattern never consults that entry, so such a trigger comes back from a restart with no suppression at all — including for a subject it fired on moments earlier, not just for new ones. The trade is still right (persisting one window per subject would grow `trigger_jobs.json` without bound, and an extra delivery is the safe direction), but "at most one extra fire per subject" understated it and the `load` doc comment promising cooldowns survive restarts is now wrong for five pattern kinds. Comment corrected, behaviour pinned by `subject_scoped_cooldown_does_not_survive_restart` so the cost stays visible instead of being rediscovered. The stale `#5115` comment quoting the old single-key `insert` signature is fixed while there.

Five new tests. Three verified by mutation: reverting `MemoryUpdate` to the catch-all, restoring the panicking horizon, and taking `min` rather than `max` for the horizon each fail exactly one test and nothing else. The last of those closes a coverage gap I would otherwise have shipped — the existing prune test used a single trigger, so `max` versus `min` was invisible to it.

Verification on `a7220ef1`: `cargo check --workspace --lib` clean, `cargo fmt --all --check` clean, `cargo clippy -p librefang-kernel --all-targets -- -D warnings` clean, `cargo test -p librefang-kernel --lib triggers` 95 passing.

crates/librefang-kernel/src/triggers.rs
```diff
@@ -304,12 +304,26 @@ pub struct TriggerEngine {
...
+    /// Last fire wall-clock timestamp per cooldown window (issue #6756).
     ///
-    /// Uses `DateTime<Utc>` rather than `std::time::Instant` so that the state
-    /// can be round-tripped through the `Trigger.last_fired_at` field on disk,
+    /// Keyed on `(trigger, subject)` rather than on the trigger alone, where
+    /// the subject is what the event is *about* — a task id, a memory key —
+    /// for the patterns that name one. A window keyed on the trigger cannot
+    /// tell "the same thing fired twice" from "two different things happened
+    /// a second apart", and silently discarded the second, which for a task
+    /// board means the work is never announced again.
```
houko: Nice, thorough fix — the subject-scoping rule and the mutation-verified test suite are genuinely well done.

One CLAUDE.md convention this diff runs into: prose (including doc-comments) is supposed to break only at sentence boundaries — "one sentence = one line, regardless of length," never hard-wrapped at a fixed column. This `last_fired` doc block hard-wraps at ~80 columns instead, splitting single sentences across multiple lines, e.g. lines 309–311 are one sentence ("Keyed on `(trigger, subject)` ... for the patterns that name one.") split across three lines, and 312–314 is another split across three.

This isn't a one-off here — the same fixed-column wrapping shows up throughout the new doc comments added in this diff (the `load()` cooldown-recovery comment, the `pattern_changed` comment, the inline comments inside `evaluate_with_resolver`, `prune_expired_cooldowns`'s doc comment, and the big `cooldown_subject` doc comment at the bottom of the diff all have the same shape). The changelog fragment and PR body, by contrast, already follow the one-sentence-per-line convention correctly, so this looks like it's specific to the code doc-comments.

Not blocking — purely a house style thing — but worth a rewrap pass across the new comments in this file before merge, since the rule is spelled out explicitly in CLAUDE.md ("Prose wrapping: no column limit; break only at sentence boundaries") and applies to doc-comments by name.

---

houko (commented):
Automated daily review pass. The subject-scoping design and the mutation-verified regression tests look solid — distinct-subject delivery, same-subject suppression, catch-all-pattern-wide-window, restart trade-off, and deadlock-avoidance are all pinned by tests, and this session found no correctness or CLAUDE.md structural-rule issues (changelog fragment is present and correctly formatted, no config/route/prompt-ordering/concurrency concerns apply here). One convention nit left inline: the new doc-comments in `triggers.rs` hard-wrap at a fixed column rather than breaking at sentence boundaries per CLAUDE.md's prose-wrapping rule.

Note for maintainers: the PR head is `nevgenov/librefang:fix/6756-cooldown-defers-not-drops` (a fork), which this session's GitHub tooling cannot write to (repo-write access is scoped to `librefang/librefang` only), so the wrap nit is left as a review comment rather than a pushed fixup commit.




---

crates/librefang-kernel/src/triggers.rs
houko, line 1: Small doc-completeness gap, unrelated to the wrapping-style thread already open here: `load()`'s own doc comment (lines 385-391, not touched by this diff) still says unconditionally that "cooldown windows survive daemon restarts."
That claim is exactly what the self-review commit (`a7220ef1`) corrected everywhere else — the caveat that per-subject windows do NOT survive a restart is written up carefully at the `insert` call site inside `load()`'s body, but never made it into the function's own public doc comment, which is what `cargo doc` and IDE hovers actually surface.

Suggested patch (applies cleanly on top of `a7220ef1`):

```diff
--- a/crates/librefang-kernel/src/triggers.rs
+++ b/crates/librefang-kernel/src/triggers.rs
@@ -384,8 +384,14 @@ impl TriggerEngine {
 
     /// Load persisted triggers from disk and rebuild the agent index.
     ///
-    /// Restores `last_fired` state from `Trigger.last_fired_at` so that
-    /// cooldown windows survive daemon restarts.
+    /// Restores `last_fired` state from `Trigger.last_fired_at` so that the
+    /// trigger-wide cooldown window survives daemon restarts.
+    ///
+    /// This covers the trigger-wide window only (#6756): per-subject windows
+    /// for the task-board, memory-key, and agent-lifecycle pattern kinds
+    /// (see `cooldown_subject`) are in-memory and do not survive a restart —
+    /// such a trigger starts back up with no suppression for any subject,
+    /// including one it fired on moments earlier.
     ///
     /// Returns the number of triggers loaded. Returns `Ok(0)` if the
     /// persistence file does not exist or no path is configured.
     pub fn load(&self) -> LibreFangResult<usize> {
```

Everything else I traced through this diff (the `(TriggerId, Option<String>)` re-key, `forget_cooldowns`/`prune_expired_cooldowns` call sites, the `cooldown_subject` match arms against the full `TriggerPattern` enum, and the new tests' use of `register_with_target`/`MemoryDelta`) checks out — no correctness issues beyond what the existing threads on this PR already caught and got fixed in `432abcb`/`a7220ef1`.

---

houko (commented):
One small doc-completeness item found on top of the already-thorough review history here; see inline comment. No new correctness issues beyond what's already been caught and fixed in this PR's own commits.

---

nevgenov:
Both fixed in `0f09957e`.

**Wrapping.** Every doc and inline comment this branch adds to `triggers.rs` now breaks only at sentence boundaries. Untouched paragraphs are left as they were, per the same rule's carve-out for pre-existing prose — the diff shrank by ~100 lines rather than growing, since the rewrap merges split sentences rather than reflowing the file.

Worth saying plainly: this is the third time the same nit has been raised against my work — twice on #6744, now here. It is not a misunderstanding of the rule, it is a habit that survives knowing the rule, so I have added a check for it to my own pre-push pass rather than relying on catching it by eye.

**`load()`'s doc comment.** You are right that this is the one that mattered of the two. The caveat from `a7220ef1` went into the body at the `insert` site, but the public doc is what `cargo doc` and IDE hovers surface, so the corrected claim never reached where it is actually read — the function still promised unconditionally that cooldown windows survive restarts. It now says the restore covers the trigger-wide window only, and that a subject-scoped pattern comes back with no suppression for any subject, including one it fired on moments earlier.

I took your patch's substance and kept the wording consistent with the body comment and the changelog fragment, so the three now say the same thing in the same terms.

Verification on `0f09957e`: `cargo check --workspace --lib` clean, `cargo fmt --all --check` clean, `cargo clippy -p librefang-kernel --all-targets -- -D warnings` clean, `cargo test -p librefang-kernel --lib triggers` 95 passing.

nevgenov:
Ran a second self-review pass over `0f09957e` and it found the class you have now caught twice on this PR — existing text the change falsified. Three items, all in `f19b584c`.

**Five descriptions of the cooldown still said "per-trigger".** That stopped being true for seven of the twelve patterns. Two are in `triggers.rs` itself: `evaluate`'s public doc listed "Per-trigger cooldown" as its first storm-prevention layer, and the `cooldown_secs` field doc promised the trigger could not fire again until the window elapsed. Both now state the window is `(trigger, subject)` for patterns that name a subject and trigger-wide otherwise.

Three are outside this crate — the CLI help for `trigger create --cooldown` and `trigger update --cooldown`, and the trigger and config tables under `docs/`. I crossed that boundary rather than deferring it: they document the exact knob whose meaning this branch changes, so leaving them would ship a description the code contradicts. Say the word if you would rather have them split out and I will move them to a follow-up.

The CLI example was the sharp one. `agent_terminated --max-fires 5 --cooldown 60` is the documentation's own example, and scoping that pattern per agent means five agents crashing at once now produce five fires, exhaust the budget immediately and disable the trigger — against the sixty-second spacing the example implies. The example says so now, and the changelog calls the interaction out: one fire per subject is the point, but pairing a scoped pattern with `max_fires` behaves differently than it did and an operator should not have to derive that.

**A paragraph I rewrapped without owning.** `total_registered`'s comment was untouched by this branch, and CLAUDE.md's prose rule explicitly carves out pre-existing paragraphs. It is restored, which also makes my previous comment's claim that untouched paragraphs were left alone true again — it was not when I wrote it.

**An intra-doc link that would not have rendered.** `load`'s doc linked `[`cooldown_subject`]`, which is private, so rustdoc would warn `private_intra_doc_links` and produce no link — in a comment whose entire motive was that the public doc is what `cargo doc` surfaces. Now plain text.

Verification on `f19b584c`: `cargo check --workspace --lib` clean, `cargo fmt --all --check` clean, `cargo clippy -p librefang-kernel -p librefang-cli --all-targets -- -D warnings` clean, `cargo test -p librefang-kernel --lib triggers` 95 passing.

PR #6919, opened 2026-08-10
fix(api): reserve idempotency keys before execution

houko:
## Summary

- atomically reserve each Idempotency-Key before executing its handler
- return idempotency_key_in_use for concurrent duplicates and bind completion/release to a random owner token
- keep pending reservations fail-closed outside the completed-response replay TTL
- fail closed on reservation/completion storage errors and document recovery semantics

## Verification

- cargo test -p librefang-memory idempotency::tests --lib (6 passed)
- cargo test -p librefang-api idempotency::tests --lib (9 passed)
- cargo test -p librefang-api --test idempotency_test (11 passed)
- cargo clippy --workspace --lib -- -A clippy::nonminimal-bool -D warnings
- cargo fmt --all
- git diff --check

Closes the concurrent check-then-act idempotency race identified in the audit.

### DIFF
--- a/changelog.d/security/idempotency-single-flight.md
+++ b/changelog.d/security/idempotency-single-flight.md
@@ -1,0 +1,3 @@
+Reserve each `Idempotency-Key` atomically before its handler starts, so concurrent retries can no longer execute the same state-creating side effect twice.
+An in-flight duplicate now receives `409 idempotency_key_in_use`; owner tokens prevent stale requests from modifying replacement reservations, cancelled and non-successful attempts release their reservation, and storage, clock, or corrupt-status failures fail closed instead of bypassing deduplication. Expired-row pruning is limited to once per minute.
+(@houko)
--- a/changelog.d/security/idempotency-single-flight.md
+++ b/changelog.d/security/idempotency-single-flight.md
@@ -1,3 +1,0 @@
-Reserve each `Idempotency-Key` atomically before its handler starts, so concurrent retries can no longer execute the same state-creating side effect twice.
-An in-flight duplicate now receives `409 idempotency_key_in_use`; owner tokens prevent stale requests from modifying replacement reservations, cancelled and non-successful attempts release their reservation, and storage, clock, or corrupt-status failures fail closed instead of bypassing deduplication. Expired-row pruning is limited to once per minute.
-(@houko)
--- a/crates/librefang-api/src/idempotency.rs
+++ b/crates/librefang-api/src/idempotency.rs
@@ -5,18 +5,12 @@
 //! through [`run_idempotent`], which:
 //!
 //! 1. Atomically reserves `(key)` in the persistent store.
-//! 2. **Reservation acquired**: executes the inner handler, then completes
-//!    the reservation with the successful 2xx response for 24 hours.
-//!    Non-2xx responses are not cached so a transient failure (rate
-//!    limit, downstream blip) does not poison the slot — clients can
-//!    retry the same key and get a real attempt.
+//! 2. **Reservation acquired**: executes the inner handler, then completes the reservation with the successful 2xx response for 24 hours.
+//!    Non-2xx responses are not cached so a transient failure (rate limit, downstream blip) does not poison the slot — clients can retry the same key and get a real attempt.
 //! 3. **Concurrent reservation, same body**: returns 409 without running the handler.
-//! 4. **Cache hit, same body**: replays the cached `(status, body)`
-//!    without re-executing the handler.
-//! 5. **Cache hit, different body**: returns 409 Conflict. The
-//!    `Idempotency-Key` is the operator-supplied dedup token and a
-//!    different payload under the same key is a programming error
-//!    (e.g. UI accidentally reuses an old key after editing the form).
+//! 4. **Cache hit, same body**: replays the cached `(status, body)` without re-executing the handler.
+//! 5. **Cache hit, different body**: returns 409 Conflict.
+//!    The `Idempotency-Key` is the operator-supplied dedup token and a different payload under the same key is a programming error (e.g. UI accidentally reuses an old key after editing the form).
 //!
 //! Body identity is sha256 over the raw JSON bytes the handler
 //! received. We hash bytes, not parsed JSON, so a re-serialised body
@@ -194,9 +188,8 @@ where
 
     let body_hash = hash_body(body_bytes);
 
-    // Reserve before starting the handler. The SQLite implementation performs
-    // expiration cleanup, insert, and existing-row lookup under one immediate
-    // transaction, so same-key requests cannot both observe a miss.
+    // Reserve before starting the handler.
+    // The SQLite implementation performs expiration cleanup, insert, and existing-row lookup under one immediate transaction, so same-key requests cannot both observe a miss.
     let reservation = match store.reserve(key, &body_hash) {
         Ok(reservation) => reservation,
         Err(e) => {
@@ -234,17 +227,16 @@ where
             body: body.clone(),
         };
         if let Err(e) = store.complete(key, &body_hash, &guard.token, &cached) {
-            // Keep the pending row instead of making the key reusable after the
-            // side effect already happened. This remains fail-closed until an
-            // operator has verified the side effect and removes the orphan.
+            // Keep the pending row instead of making the key reusable after the side effect already happened.
+            // This remains fail-closed until an operator has verified the side effect and removes the orphan.
             guard.disarm();
             tracing::error!(key, error = %e, "idempotency completion failed");
             return store_unavailable_response();
         }
         guard.disarm();
     }
-    // A non-2xx response leaves the guard armed, releasing the reservation so
-    // the key remains retriable. Cancellation and panic take the same Drop path.
+    // A non-2xx response leaves the guard armed, releasing the reservation so the key remains retriable.
+    // Cancellation and panic take the same Drop path.
     // Opportunistic prune so the table self-trims.
     if let Err(e) = store.prune_expired() {
         tracing::debug!(error = %e, "idempotency prune_expired failed");
--- a/crates/librefang-api/src/idempotency.rs
+++ b/crates/librefang-api/src/idempotency.rs
@@ -4,15 +4,16 @@
 //! `Idempotency-Key: <opaque-string>` header. When set, the handler runs
 //! through [`run_idempotent`], which:
 //!
-//! 1. Looks up `(key)` in the persistent store.
-//! 2. **Cache miss**: executes the inner handler, then persists the
-//!    successful 2xx response under `(key, body_hash)` for 24 hours.
+//! 1. Atomically reserves `(key)` in the persistent store.
+//! 2. **Reservation acquired**: executes the inner handler, then completes
+//!    the reservation with the successful 2xx response for 24 hours.
 //!    Non-2xx responses are not cached so a transient failure (rate
 //!    limit, downstream blip) does not poison the slot — clients can
 //!    retry the same key and get a real attempt.
-//! 3. **Cache hit, same body**: replays the cached `(status, body)`
+//! 3. **Concurrent reservation, same body**: returns 409 without running the handler.
+//! 4. **Cache hit, same body**: replays the cached `(status, body)`
 //!    without re-executing the handler.
-//! 4. **Cache hit, different body**: returns 409 Conflict. The
+//! 5. **Cache hit, different body**: returns 409 Conflict. The
 //!    `Idempotency-Key` is the operator-supplied dedup token and a
 //!    different payload under the same key is a programming error
 //!    (e.g. UI accidentally reuses an old key after editing the form).
@@ -29,7 +30,9 @@
 use axum::http::StatusCode;
 use axum::response::{IntoResponse, Response};
 use axum::Json;
-use librefang_memory::idempotency::{CachedResponse, IdempotencyStore};
+use librefang_memory::idempotency::{
+    CachedResponse, IdempotencyStore, Reservation, ReservationToken, PENDING_STATUS,
+};
 use sha2::{Digest, Sha256};
 
 /// Maximum length of a client-supplied `Idempotency-Key`. Bounded so a
@@ -84,6 +87,68 @@ pub fn body_conflict_response() -> Response {
     (StatusCode::CONFLICT, Json(payload)).into_response()
 }
 
+/// 409 response returned while the first request for this key is still running.
+pub fn request_in_progress_response() -> Response {
+    let payload = serde_json::json!({
+        "error": "A request with this Idempotency-Key is already in progress",
+        "code": "idempotency_key_in_use",
+        "type": "idempotency_key_in_use",
+    });
+    (StatusCode::CONFLICT, Json(payload)).into_response()
+}
+
+fn store_unavailable_response() -> Response {
+    let payload = serde_json::json!({
+        "error": "Idempotency storage is unavailable",
+        "code": "idempotency_store_unavailable",
+        "type": "idempotency_store_unavailable",
+    });
+    (StatusCode::SERVICE_UNAVAILABLE, Json(payload)).into_response()
+}
+
+struct ReservationGuard<'a> {
+    store: &'a dyn IdempotencyStore,
+    key: &'a str,
+    body_hash: &'a str,
+    token: ReservationToken,
+    armed: bool,
+}
+
+impl<'a> ReservationGuard<'a> {
+    fn new(
+        store: &'a dyn IdempotencyStore,
+        key: &'a str,
+        body_hash: &'a str,
+        token: ReservationToken,
+    ) -> Self {
+        Self {
+            store,
+            key,
+            body_hash,
+            token,
+            armed: true,
+        }
+    }
+
+    fn disarm(&mut self) {
+        self.armed = false;
+    }
+}
+
+impl Drop for ReservationGuard<'_> {
+    fn drop(&mut self) {
+        if self.armed {
+            if let Err(error) = self.store.release(self.key, self.body_hash, &self.token) {
+                tracing::error!(
+                    key = self.key,
+                    %error,
+                    "failed to release abandoned idempotency reservation"
+                );
+            }
+        }
+    }
+}
+
 /// Wrap a handler closure with idempotency semantics.
 ///
 /// `key_header` is `None` when the caller did not send an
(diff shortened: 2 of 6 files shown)


changelog.d/security/idempotency-single-flight.md
```diff
@@ -0,0 +1,3 @@
+Reserve each `Idempotency-Key` atomically before its handler starts, so concurrent retries can no longer execute the same state-creating side effect twice.
+An in-flight duplicate now receives `409 idempotency_key_in_use`; owner tokens prevent stale requests from modifying replacement reservations, cancelled and non-successful attempts release their reservation, and storage, clock, or corrupt-status failures fail closed instead of bypassing deduplication. Expired-row pruning is limited to once per minute.
+(@houko)
```
houko: Per `changelog.d/README.md`, a fragment should end with a trailing `(#PR)` reference and the filename should lead with the PR/issue number. Neither is CI-enforced, but skipping the `(#PR)` reference means this PR also gets a generated `- <title> (#N) (@author)` line in the release notes, so the change ends up listed twice. Suggested fix (also rename `idempotency-single-flight.md` → `6919-idempotency-single-flight.md` to match the convention):

```suggestion
An in-flight duplicate now receives `409 idempotency_key_in_use`; owner tokens prevent stale requests from modifying replacement reservations, cancelled and non-successful attempts release their reservation, and storage failures fail closed instead of bypassing deduplication (#6919) (@houko)
```

---

houko (commented):
Reviewed against CLAUDE.md. Verified with `cargo check -p librefang-memory -p librefang-api --lib` (clean) and `cargo test -p librefang-memory idempotency` (6/6 passing, including the real-thread race test). The reserve/complete/release token-ownership design looks correct: the `BEGIN IMMEDIATE` transaction in `reserve()` gives exactly one owner before any handler runs, `complete`/`release` both gate on the caller's token so a superseded reservation can't be touched by its old owner, and the guard's Drop path correctly releases on cancellation/panic while leaving completed and fail-closed rows alone. One small changelog nit left as a suggestion below; nothing else found.

---

crates/librefang-memory/src/idempotency.rs
```diff
@@ -257,10 +439,227 @@ mod tests {
...
-                rusqlite::params![now_unix() - 100_000, now_unix() - 1],
+                rusqlite::params![now_unix().unwrap() - 100_000, now_unix().unwrap() - 1],
             )
             .unwrap();
         }
         assert!(s.lookup("old").unwrap().is_none());
     }
+
+    #[test]
+    fn concurrent_reservations_have_exactly_one_owner() {
```
houko, line 450: Minor test-robustness nit, not a production issue: this test's pool is built directly via `SqliteConnectionManager::file(...)` without `busy_timeout`/WAL pragmas (unlike the production path, which reuses `MemorySubstrate`'s pool and its `DEFAULT_CONNECTION_PRAGMAS`, including `busy_timeout=5000`).
Without a busy timeout, if the two barrier-released threads' `BEGIN IMMEDIATE` calls ever land close enough together, the loser gets an immediate `SQLITE_BUSY` (propagated through `?`) instead of blocking, which would panic the test via `.unwrap()` rather than exercising the intended `Reservation::Existing` path.
I ran it 10x locally and it passed every time — the reserve/insert/select/commit critical section is short enough that true simultaneity is rare — but it's a source of theoretical CI flakiness worth a `.with_init(|c| c.busy_timeout(...))` on the test's pool builder to make it deterministic.

---

crates/librefang-api/src/idempotency.rs
houko, line 1: Correctness review of the TOCTOU fix and failure-mode handling:

- **Reserve is atomic**: `SqliteIdempotencyStore::reserve` wraps the expiry-delete, `INSERT OR IGNORE` (on the `key TEXT PRIMARY KEY` unique constraint), and existing-row lookup in one `BEGIN IMMEDIATE` transaction, so two concurrent same-key requests cannot both observe a miss — closes the race the old lookup-then-put design had.
- **Owner tokens are enforced by the DB, not just in-process**: `complete()`/`release()` are single-statement compare-and-swap `UPDATE`/`DELETE`s gated on `response_status = 0 AND response_body = <token>`, so a superseded owner's completion attempt fails (`updated != 1` → `Invariant` error) and its release is a no-op against a replacement's row.
  Verified directly by `superseded_owner_cannot_complete_or_release_replacement`.
- **Panic/cancellation release, process-crash does not (by design)**: `ReservationGuard::drop` releases the pending row unless disarmed, and disarming only happens after a *successful* `complete()` or after a *failed* `complete()` (deliberately, since the side effect already ran by then).
  This covers panics inside `f().await` and task cancellation (`cancelled_handler_releases_pending_reservation` exercises the latter via `task.abort()`).
  An abrupt process crash (SIGKILL/OOM) is the one case that leaves an orphaned pending row forever — that's called out explicitly in the changelog/docs as an intentional fail-closed trade-off rather than an oversight, which I agree with given the alternative (auto-expiring pending rows) would reopen the exact double-execution race this PR closes.
- One process gap I didn't find covered anywhere: there's no CLI/admin surface to remove an orphaned pending row after that crash scenario — today an operator would need direct sqlite3 access to the substrate DB (`DELETE FROM idempotency_keys WHERE key=...`). Not a correctness bug, but worth a tracked follow-up so "an operator explicitly removes the orphaned reservation" (per the docs) has a supported path rather than requiring DB surgery.
- Per CLAUDE.md's mandatory-integration-testing rule for route/wiring changes: the response contract for existing idempotency-wrapped routes changed (new 409 `idempotency_key_in_use` for concurrent duplicates, new 503 `idempotency_store_unavailable` for storage failures), but `crates/librefang-api/tests/idempotency_test.rs` wasn't touched.
  I ran the existing suite (11 tests) against this branch and all pass unmodified, and the new behavior is well covered at the middleware-unit level (`concurrent_same_key_executes_handler_once`, `reserve_failure_does_not_run_handler`, `complete_failure_fails_closed_without_reexecuting_handler` in this file's own tests, plus real-SQLite concurrency coverage in `librefang-memory::idempotency::tests`) — so I'm not blocking on this, but a `TestServer`-level test hitting a real route with two genuinely concurrent requests under the same key would be the more direct regression guard the rule asks for.

Pushed a mechanical fix (`dfbbb331c`) for hard-wrapped doc comments / architecture doc prose that broke sentences mid-line instead of at sentence boundaries (CLAUDE.md prose-wrapping rule) — no behavior change.

Verified locally: `cargo test -p librefang-memory --lib -- idempotency::` (11 passed), `cargo test -p librefang-api --lib -- idempotency::` (9 passed), `cargo test -p librefang-api --test idempotency_test` (11 passed), `cargo fmt --check -p librefang-api -p librefang-memory` (clean).
`cargo clippy -p librefang-api --all-targets -- -D warnings` currently fails on an unrelated pre-existing `nonminimal_bool` lint in `routes/config/system.rs` that's already present on `origin/main` (not introduced by this PR) — clippy's output shows no issues from the idempotency changes themselves.

---

crates/librefang-api/src/idempotency.rs
```diff
@@ -4,18 +4,13 @@
...
-//! 4. **Cache hit, different body**: returns 409 Conflict. The
-//!    `Idempotency-Key` is the operator-supplied dedup token and a
-//!    different payload under the same key is a programming error
-//!    (e.g. UI accidentally reuses an old key after editing the form).
+//! 1. Atomically reserves `(key)` in the persistent store.
+//! 2. **Reservation acquired**: executes the inner handler, then completes the reservation with the successful 2xx response for 24 hours.
+//!    Non-2xx responses are not cached so a transient failure (rate limit, downstream blip) does not poison the slot — clients can retry the same key and get a real attempt.
+//! 3. **Concurrent reservation, same body**: returns 409 without running the handler.
+//! 4. **Cache hit, same body**: replays the cached `(status, body)` without re-executing the handler.
+//! 5. **Cache hit, different body**: returns 409 Conflict.
```
houko, line 12: Commit dfbbb331 in this branch's history is authored as `Claude <noreply@anthropic.com>` (a pure doc/comment rewrap, no behavior change). Per CLAUDE.md, author identity resolving to Claude/Anthropic is disallowed even when the message text is clean — the `commit-msg` hook checks `GIT_AUTHOR_IDENT` for exactly this. Same pattern shows up on the other PRs in this batch (#6920, #6921); the review comments that mention pushing these commits also close with a "Generated by Claude Code" footer, which the repo's own attribution rule rules out. Worth a rebase to fix authorship (and dropping that footer going forward) before merge — no concern with the code itself.

---

houko (commented):
This PR has already had two thorough review passes covering the atomicity, owner-token, and fail-closed design in depth — nothing new to add there, and the open items (test busy_timeout robustness, orphaned-reservation admin surface, TestServer-level concurrency test) are already tracked as non-blocking. Only new item is a process note on commit authorship, inline below.

---

crates/librefang-api/src/idempotency.rs
houko, line 12: Still unresolved from an earlier thread on this PR: commit `dfbbb331c` in this branch's history is authored as `Claude <noreply@anthropic.com>` (`git log` shows `author: Claude <noreply@anthropic.com>`).
Per CLAUDE.md, author identity resolving to Claude/Anthropic is disallowed even when the commit message itself is clean — the `commit-msg` hook checks `GIT_AUTHOR_IDENT` for exactly this, so this branch would fail that hook locally if hooks are enabled.
I can't fix this myself: correcting authorship on an existing commit requires a rebase and force-push to this branch, and force-pushing without explicit maintainer sign-off is outside what an automated pass is allowed to do here.
Suggested fix before merge: `git rebase -i` (or `git commit --amend --reset-author` if it's the tip) to reauthor that commit as a human contributor, then force-push this branch.

---

houko (commented):
Automated review pass.

Correctness of the reserve-before-execute fix: verified independently.
`SqliteIdempotencyStore::reserve` wraps the expiry cleanup, `INSERT OR IGNORE`, and existing-row lookup in one `BEGIN IMMEDIATE` transaction, so two concurrent same-key requests cannot both observe a miss — this closes the original check-then-act race.
`complete`/`release` are single-statement compare-and-swap operations gated on `response_status = 0 AND response_body = <token>`, so a superseded owner cannot complete or release a replacement reservation; this is exercised directly by `superseded_owner_cannot_complete_or_release_replacement`.
The changelog fragment is already correctly placed at `changelog.d/security/6919-idempotency-single-flight.md` with the `(#6919) (@houko)` trailer, and the touched prose already reads one-sentence-per-line per the repo's wrapping convention.

Fixed directly (mechanical, low-risk): `crates/librefang-memory/src/idempotency.rs`, `concurrent_reservations_have_exactly_one_owner` built its SQLite pool without the shared `busy_timeout` pragma that the production pool always applies, so a true-simultaneity `BEGIN IMMEDIATE` collision could panic the test with `SQLITE_BUSY` instead of exercising the intended `Reservation::Existing` path.
Applied the same `DEFAULT_CONNECTION_PRAGMAS` init used by `MemorySubstrate`'s own pool.

One inline comment left for human judgment: a commit in this branch's history is authored as `Claude <noreply@anthropic.com>`, which violates the repo's git-authorship rule and needs a rebase + force-push that this automated pass isn't authorized to perform.




---

crates/librefang-memory/src/idempotency.rs
houko, line 1: The reservation/token/CAS mechanics look correct to me: `BEGIN IMMEDIATE` + the `key TEXT PRIMARY KEY` constraint gives `reserve()` a real atomic check-and-insert (no TOCTOU), `complete()`/`release()` are single-statement compare-and-swaps keyed on the owner token (verified by `superseded_owner_cannot_complete_or_release_replacement`), and the shared substrate pool already sets `busy_timeout=5000` + WAL, so concurrent `BEGIN IMMEDIATE` calls wait rather than erroring. Cancellation/panic release via `ReservationGuard::drop`, which is exercised by `cancelled_handler_releases_pending_reservation`.

One design point I'd like the author to weigh in on rather than have me guess at: pending reservations are deliberately never auto-expired ("Pending reservations do not expire automatically... an abnormal process exit fails closed until an operator explicitly removes the orphaned reservation"). That's a reasonable fail-closed default, but it means a hard crash/OOM-kill/`panic=abort` while a handler is mid-flight (Drop never runs) permanently wedges that `Idempotency-Key` with no expiry and, from what I can see in this diff, no CLI/API tooling to list or clear orphaned pending rows — recovery is manual `DELETE FROM idempotency_keys` against the live SQLite file. Given this is exactly the scenario the audit called out ("a bug here could permanently wedge an idempotency key"), should there be either (a) a generous fallback expiry for pending rows (e.g. hours, operator-configurable) as a crash-recovery safety net, or (b) an admin endpoint/CLI verb to inspect and release stuck pending reservations? Happy to be told the manual-SQL recovery path is intentional and sufficient for now — just flagging it as a judgment call rather than assuming either way.

---

crates/librefang-api/src/idempotency.rs
houko, line 1: Minor process note (not a correctness issue — the new unit tests here, e.g. `concurrent_same_key_executes_handler_once`, `cancelled_handler_releases_pending_reservation`, plus the real-SQLite `concurrent_reservations_have_exactly_one_owner` in the memory crate, do exercise the single-flight property well): CLAUDE.md's "MANDATORY: Integration Testing" section asks for a `#[tokio::test]` against a real `TestServer` for wiring changes, and this middleware sits in front of every idempotent POST route. The PR description cites the existing `--test idempotency_test` suite passing unchanged (11 tests), but none of those appear to add a genuinely concurrent same-key request against a real route through `TestServer` — the concurrency coverage lives at the module/mock-store and raw-SQLite level, not end-to-end through the HTTP layer. Given how central the new `409 idempotency_key_in_use` behavior is, an end-to-end concurrent-duplicate-request test through one real route would give stronger wiring-level confidence (header parsing → router → response mapping). Not blocking given the coverage that does exist, but wanted to surface it rather than silently let it pass.

---

crates/librefang-api/src/idempotency.rs
```diff
@@ -129,42 +188,55 @@ where
...
-    if let Some(existing) = prior {
-        if existing.body_hash != body_hash {
-            return body_conflict_response();
+    let token = match reservation {
+        Reservation::Existing(existing) => {
+            if existing.body_hash != body_hash {
+                return body_conflict_response();
+            }
+            if existing.response.status == PENDING_STATUS {
+                return request_in_progress_response();
```
houko, line 207: Correctness check: `SqliteIdempotencyStore::reserve` (crates/librefang-memory/src/idempotency.rs:236-275) does the expiry-delete, `INSERT OR IGNORE`, and existing-row read all inside one `BEGIN IMMEDIATE` transaction, so the check and the reservation are atomic — no TOCTOU window between two concurrent requests with the same key. Confirmed this is genuinely exercised by `concurrent_reservations_have_exactly_one_owner`, and `d98ae9ed`'s `busy_timeout` pragma fix is a real fix (not a no-op): without it, two real threads racing `BEGIN IMMEDIATE` on the same file-backed connection can hit `SQLITE_BUSY` immediately (default busy_timeout is 0) and panic the test via `.unwrap()` before ever reaching the intended `Existing`-reservation assertion, so the pragma is necessary for the test to reliably exercise contention rather than flake.

One gap: this new `request_in_progress_response()` / `idempotency_key_in_use` path is only covered by the in-process `concurrent_same_key_executes_handler_once` unit test against the `MemStore` fake (this file's own `mod tests`). `crates/librefang-api/tests/idempotency_test.rs` — the `TestServer`-style suite that boots the real router per CLAUDE.md's "MANDATORY: Integration Testing" section — wasn't touched by this PR and has no case that drives two concurrent requests through the real SQLite-backed store and asserts one gets `409 idempotency_key_in_use` while the other completes. Given this changes observable behavior for every idempotency-guarded route, it's worth adding one full-router case, following the existing `spawn_with_idempotency_key_replays_response` pattern but with a synchronization point (similar to how the unit test uses oneshot channels) so the race is deterministic rather than timing-dependent. Not fixing this myself since a naive version without a real sync hook into the handler risks being flaky.

---

docs/architecture/idempotency.md
houko, line 1: Process note, not a content issue: commit `dfbbb331` ("fix(api): join mid-sentence line breaks in idempotency docs and comments") is authored and committed as `Claude <noreply@anthropic.com>` (`git log` author/committer identity), touching this file plus `crates/librefang-api/src/idempotency.rs`, `crates/librefang-memory/src/idempotency.rs`, and `crates/librefang-memory/src/migration.rs`. Repo policy (CLAUDE.md "Commit & PR hygiene" / the `commit-msg` hook's author-identity check) rejects commits whose `GIT_AUTHOR_IDENT` resolves to Claude/Anthropic, independent of message content — this one apparently reached `origin` without that hook running locally. The content itself is fine (correctly rewraps prose to one-sentence-per-line), so this is a request to re-author that commit (e.g. `git commit --amend --reset-author` / interactive rebase to fold it under a human identity) before merge, not to revert it.

---

houko (commented):
Reviewed the new commit (`d98ae9ed`, "test(memory): apply production busy_timeout pragma in idempotency race test") against the whole diff vs `main`. Two inline notes below: one confirming the reservation logic is actually atomic (no TOCTOU) and that the new busy_timeout pragma is a real fix, not a no-op; one flagging a commit authored as Claude/Anthropic that needs re-authoring before merge per repo policy. Also flagging, without a fix, that the new concurrent-duplicate → `409 idempotency_key_in_use` path has no full-router `TestServer`-style coverage in `tests/idempotency_test.rs` (only a `MemStore`-backed unit test). Everything else (changelog fragment format/filename, commit message conventions, prose wrapping) checked out clean.

---

crates/librefang-api/src/idempotency.rs
houko, line 1: Manually traced the reservation lifecycle (`reserve` → handler → `complete`/`release`) against the SQLite store; the locking looks correct: `reserve()` does its expiry-sweep, `INSERT OR IGNORE`, and existing-row read inside one `BEGIN IMMEDIATE` transaction, so there's no TOCTOU window between "check" and "acquire" — concurrent callers are fully serialized on the write lock, and the production pool's `busy_timeout=5000` (from `DEFAULT_CONNECTION_PRAGMAS`) means they queue instead of erroring. `complete()`/`release()` are single-statement CAS updates gated on `response_status = 0 AND response_body = <owner token>`, so a superseded owner (e.g. after its row was deleted and re-reserved) can neither complete nor release the replacement — verified against `superseded_owner_cannot_complete_or_release_replacement`. `ReservationGuard::drop` covers cancellation/panic by releasing unless explicitly disarmed on the success path, and disarming also happens (correctly) on a `complete()` failure so the pending row is kept instead of silently reopening a key whose side effect already ran. No correctness gap found here.

One coverage gap worth a second look: this PR changes the response codes returned by 5 real routes under contention (new `409 idempotency_key_in_use`, new `503 idempotency_store_unavailable`), but `crates/librefang-api/tests/idempotency_test.rs` — the suite that boots the real router via `start_test_server`/`build_router` — isn't touched, so those two new paths are only exercised by the `MemStore`-backed unit tests in this file (with hand-synchronized oneshot channels) and by `librefang-memory`'s own threaded SQLite test, never through the real HTTP+SQLite stack together. CLAUDE.md's integration-testing section calls for `TestServer` coverage on route-behavior changes; I didn't add one myself because reliably forcing the "still in flight" window over a real HTTP round trip (without a test-only hook to pause the handler) risks a flaky test rather than a deterministic one — wanted a maintainer's call on whether that's an acceptable gap or worth a synchronization hook.

---

crates/librefang-memory/src/migration.rs
houko, line 1: Unrelated to the locking logic, but flagging since it's a hard repo-policy violation I can't fix myself without rewriting branch history (which I'm avoiding — force-push isn't something I'll do on someone else's PR): commit `dfbbb33` on this branch ("fix(api): join mid-sentence line breaks in idempotency docs and comments") has its author identity set to `Claude <noreply@anthropic.com>`. CLAUDE.md's `commit-msg` hook is explicitly designed to reject exactly this ("separately rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean"), so this commit must have landed while `core.hooksPath` wasn't pointed at `scripts/hooks` for whatever produced it. Worth an interactive rebase to fix the author before merge, since `git log` on `main` would otherwise carry Anthropic attribution permanently.

---

houko (commented):
Reviewed the reservation/locking logic in detail (reserve → handler → complete/release) — no TOCTOU gap or ownership bug found; two notes left inline for maintainer judgment: a test-coverage gap for the new 409/503 paths at the real-router level, and a pre-existing commit on this branch with Claude-attributed authorship that needs a rebase before merge. The flaky-SQLite-lock test issue I found (missing busy_timeout on the concurrent-reservation test's pool) was already fixed upstream in d98ae9e while I was reviewing, so no action needed there.

---

Issue #3361, opened 2026-04-27
/api/uploads/{file_id} is publicly fetchable — UUID is the only access control


houko:
**Location**
- `crates/librefang-api/src/middleware.rs:570` (auth allowlist `always_public_get_only`)
- `crates/librefang-api/src/routes/agents.rs:5391-5443` (handler)

**Threat model**
The auth middleware marks `/api/uploads/` as `always_public_get_only`. The handler validates only that `file_id` parses as a UUID — it does not verify the requester has any relationship to the uploading agent / session.

Any LAN attacker who learns or guesses a UUID (UUIDs leak in audit logs at INFO level, dashboard responses, `tracing::warn!` lines, browser history of authenticated users, message history returned by other public endpoints) can fetch the file without any authentication. Files include user-uploaded images/PDFs/audio plus model-generated content the agent persisted.

**Suggested fix**
Remove the `/api/uploads/` prefix from the public allowlist in `middleware.rs:570` and require auth like every other agent-scoped endpoint. If a public-share use case is needed, mint a separately-scoped signed URL token rather than expose the bare UUID.

PR #6921, opened 2026-08-10
fix(api): harden session history responses

houko:
## Summary

- reuse stable, session-scoped UUID file IDs for inline history images and publish them asynchronously through atomic renames
- keep empty-session response fields consistent, localize malformed session IDs, and enforce the 100 KiB tool-result cap in UTF-8 bytes
- add concurrent materialization and real HTTP download regression coverage

## Verification

- `LIBREFANG_HOME=<empty> cargo test -p librefang-api --lib -- --test-threads=1` (928 passed)
- `cargo test -p librefang-api --lib routes::agents::sessions::tests`
- `cargo test -p librefang-api --test api_integration_test test_agent_session_empty`
- `cargo test -p librefang-api --test api_integration_test test_get_agent_session`
- `cargo clippy -p librefang-api --lib -- -A clippy::nonminimal-bool -D warnings`
- `git diff --check`

### DIFF
--- a/crates/librefang-api/src/routes/agents/sessions.rs
+++ b/crates/librefang-api/src/routes/agents/sessions.rs
@@ -46,8 +46,7 @@ async fn materialize_history_image(
     let digest = hasher.finalize();
     let mut file_id_bytes = [0_u8; 16];
     file_id_bytes.copy_from_slice(&digest[..16]);
-    // RFC 9562 UUIDv8 keeps the content-derived 128-bit identifier compatible
-    // with the upload route while reserving the standard version/variant bits.
+    // RFC 9562 UUIDv8 keeps the content-derived 128-bit identifier compatible with the upload route while reserving the standard version/variant bits.
     file_id_bytes[6] = (file_id_bytes[6] & 0x0f) | 0x80;
     file_id_bytes[8] = (file_id_bytes[8] & 0x3f) | 0x80;
     let file_id = uuid::Uuid::from_bytes(file_id_bytes).to_string();
--- a/crates/librefang-api/src/routes/agents/sessions.rs
+++ b/crates/librefang-api/src/routes/agents/sessions.rs
@@ -1,5 +1,101 @@
 use super::*;
 
+const TOOL_RESULT_MAX_BYTES: usize = 100 * 1024;
+
+fn cap_tool_result(result: &str) -> String {
+    if result.len() <= TOOL_RESULT_MAX_BYTES {
+        return result.to_string();
+    }
+    let mut end = TOOL_RESULT_MAX_BYTES;
+    while !result.is_char_boundary(end) {
+        end -= 1;
+    }
+    result[..end].to_string()
+}
+
+async fn remove_history_image_temp(path: &std::path::Path) {
+    if let Err(error) = tokio::fs::remove_file(path).await {
+        if error.kind() != std::io::ErrorKind::NotFound {
+            tracing::warn!(%error, "failed to clean up session history image");
+        }
+    }
+}
+
+async fn materialize_history_image(
+    upload_dir: &std::path::Path,
+    session_scope: &[u8],
+    media_type: &str,
+    data: &str,
+) -> Option<serde_json::Value> {
+    use base64::Engine as _;
+    use sha2::{Digest, Sha256};
+
+    let bytes = match base64::engine::general_purpose::STANDARD.decode(data) {
+        Ok(bytes) => bytes,
+        Err(error) => {
+            tracing::warn!(%error, "failed to decode session history image");
+            return None;
+        }
+    };
+    let mut hasher = Sha256::new();
+    hasher.update(session_scope);
+    hasher.update([0]);
+    hasher.update(media_type.as_bytes());
+    hasher.update([0]);
+    hasher.update(&bytes);
+    let digest = hasher.finalize();
+    let mut file_id_bytes = [0_u8; 16];
+    file_id_bytes.copy_from_slice(&digest[..16]);
+    // RFC 9562 UUIDv8 keeps the content-derived 128-bit identifier compatible
+    // with the upload route while reserving the standard version/variant bits.
+    file_id_bytes[6] = (file_id_bytes[6] & 0x0f) | 0x80;
+    file_id_bytes[8] = (file_id_bytes[8] & 0x3f) | 0x80;
+    let file_id = uuid::Uuid::from_bytes(file_id_bytes).to_string();
+    let on_disk = librefang_types::media::on_disk_name(&file_id, media_type, "");
+
+    if let Err(error) = tokio::fs::create_dir_all(upload_dir).await {
+        tracing::warn!(%error, "failed to create session image directory");
+        return None;
+    }
+    let path = upload_dir.join(on_disk);
+    let exists = match tokio::fs::try_exists(&path).await {
+        Ok(exists) => exists,
+        Err(error) => {
+            tracing::warn!(%error, "failed to inspect session history image");
+            return None;
+        }
+    };
+    if !exists {
+        let temporary = upload_dir.join(format!(".{file_id}.{}.tmp", uuid::Uuid::new_v4()));
+        if let Err(error) = tokio::fs::write(&temporary, &bytes).await {
+            tracing::warn!(%error, "failed to write temporary session history image");
+            remove_history_image_temp(&temporary).await;
+            return None;
+        }
+        if let Err(error) = tokio::fs::rename(&temporary, &path).await {
+            if !tokio::fs::try_exists(&path).await.unwrap_or(false) {
+                tracing::warn!(%error, "failed to publish session history image");
+                remove_history_image_temp(&temporary).await;
+                return None;
+            }
+            remove_history_image_temp(&temporary).await;
+        }
+    }
+
+    let filename = format!("image.{}", media_type.rsplit('/').next().unwrap_or("png"));
+    UPLOAD_REGISTRY
+        .entry(file_id.clone())
+        .or_insert_with(|| UploadMeta {
+            filename: filename.clone(),
+            content_type: media_type.to_string(),
+            uploaded_by: None,
+        });
+    Some(serde_json::json!({
+        "file_id": file_id,
+        "filename": filename,
+    }))
+}
+
 /// Query params for `GET /api/agents/{id}/session`.
 ///
 /// Using a typed struct (rather than `HashMap<String,String>`) gives us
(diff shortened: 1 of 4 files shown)


crates/librefang-api/src/routes/agents/sessions.rs
```diff
@@ -1,5 +1,100 @@
...
+    let mut hasher = Sha256::new();
+    hasher.update(session_scope);
+    hasher.update([0]);
+    hasher.update(media_type.as_bytes());
+    hasher.update([0]);
+    hasher.update(&bytes);
+    let digest = hasher.finalize();
+    let mut file_id_bytes = [0_u8; 16];
+    file_id_bytes.copy_from_slice(&digest[..16]);
+    // RFC 9562 UUIDv8 keeps the content-derived 128-bit identifier compatible with the upload route while reserving the standard version/variant bits.
```
houko, line 49: Commit 5d0475d6 in this branch's history is authored as `Claude <noreply@anthropic.com>` (this line's rewrap is its only change). Per CLAUDE.md, author identity resolving to Claude/Anthropic is disallowed even when the message text is clean — the `commit-msg` hook checks `GIT_AUTHOR_IDENT` for exactly this. Worth a rebase to correct the author before merge; the code change itself is fine (just a comment rewrap).

---

houko (commented):
Code review: the history-image dedup (content-derived UUIDv8, temp+rename publish) and the byte-vs-char tool-result cap fix both look correct and are well covered by tests. One process note flagged inline: a commit in this branch's history has an author identity that resolves to Claude, which the repo's own hook rejects.

---

crates/librefang-api/src/routes/agents/sessions.rs
houko, line 49: Verified independently via the commit list: commit 5d0475d686f7ad81d4cc332db986afb26f61cd41 on this branch (the rewrap of this comment line) has author identity `Claude <noreply@anthropic.com>`.
Per CLAUDE.md this is disallowed regardless of message text, since `commit-msg` checks `GIT_AUTHOR_IDENT` directly.
The code itself is fine; this needs a rebase to correct authorship before merge (interactive history rewrite is outside what this automated pass will do unilaterally).

Separately, on the substance of the PR: the deterministic content-derived file_id (session_id + media_type + image bytes, hashed) correctly fixes what looks like a real resource-exhaustion bug in the prior code, which wrote a brand-new randomly-named file to disk on every single `GET /session` call that returned an inline image, unbounded by repeat polling.
The new `test_get_agent_session_reuses_materialized_history_images` test covers the idempotency claim well.
No concerns with the tool-result byte-cap fix or the empty-session `label` field addition; both are covered by tests.

---

houko (commented):
Automated CLAUDE.md compliance pass. One finding: commit 5d0475d686f7ad81d4cc332db986afb26f61cd41 in this branch's history has author identity `Claude <noreply@anthropic.com>`, which violates the repo's no-AI-attribution policy even though the message text itself is clean — flagged inline, needs a rebase before merge. Everything else checked out: changelog fragment present at changelog.d/fixed/6921-session-history-response-hardening.md in the correct format, the ErrorTranslator pre-computation correctly avoids holding the !Send guard across await points, the route's response-shape change (tool-result byte cap, empty-session `label` field, image materialization) is covered by new/updated #[tokio::test] integration tests, and the deterministic-file-id change appears to genuinely close a resource-exhaustion gap in the prior code rather than being cosmetic only.




---

crates/librefang-api/src/routes/agents/sessions.rs
```diff
@@ -1,5 +1,100 @@
...
+    file_id_bytes[6] = (file_id_bytes[6] & 0x0f) | 0x80;
+    file_id_bytes[8] = (file_id_bytes[8] & 0x3f) | 0x80;
+    let file_id = uuid::Uuid::from_bytes(file_id_bytes).to_string();
```
houko, line 52: Worth a second look at the security model change here, not the mechanics.

`file_id` moves from `Uuid::new_v4()` (unguessable randomness) to `SHA256(session_scope || media_type || bytes)[..16]` reformatted as a UUIDv8. That's a good fix for the duplicate-file-per-GET issue, and session-scoping means no cross-session collision — but `materialize_history_image`'s caller still registers these with `uploaded_by: None` (unchanged from before), and per `uploads.rs`'s `serve_upload`, `owner == None` skips the access check entirely (`if let Some(owner_id) = owner` — the whole block is a no-op for `None`), so these files remain fully unauthenticated-readable at `/api/uploads/{file_id}` by anyone who has the id, same as pre-#3361-style "generator output" files.

That's not new — the pre-PR code also inserted with `uploaded_by: None` — but the *identifier* is now a deterministic function of content instead of random, and `session_scope` is derived from `target_session_id`, which is visible in API responses/URLs. In isolation this is fine (SHA-256 preimage resistance means an attacker still needs the actual image bytes to derive the id, at which point they already have the image), but I'd like a maintainer sign-off that the move away from unguessable-by-construction ids to content-addressed ids for a still-`None`-owner (i.e. publicly fetchable) resource is intentional, given how deliberately #3361 reasoned about `serve_upload`'s owner-guard model. If there's ever a path where `session_scope` could be attacker-*chosen* (rather than merely attacker-*known*) while the image bytes are only attacker-*known-partially*, this construction would need revisiting — worth a one-line note in the doc comment either way.

---

crates/librefang-api/src/routes/agents/sessions.rs
houko, line 49: Not about this line's content specifically, but flagging it as the anchor: the commit that introduced this exact change (`5d0475d`, "fix(api): join mid-sentence comment line break in session history hardening") has both author and committer identity set to `Claude <noreply@anthropic.com>`:

```
```

Per this repo's `CLAUDE.md` ("Git Conventions" / commit-msg hook description), a commit whose author identity resolves to Claude/Anthropic is supposed to be rejected server-side even when the message body itself is clean — this one apparently landed without that hook active (or it was bypassed), and it's already pushed to `fix/session-history-hardening`.

I can't fix this myself: correcting the author identity requires rewriting this commit (rebase + force-push), which is outside what an automated review pass is allowed to do without explicit maintainer sign-off. Flagging for a human to decide — squash-merge would also make this moot depending on how you land the PR, but as a standalone commit in the branch it currently violates the repo's own attribution rule.

---

houko (commented):
Automated review pass on this PR's diff against `main` (`13a30adc5`).

Summary: the actual code change looks correct and well-tested — the cross-agent/session-ownership guard (`session.agent_id != agent_id`) is unchanged and still enforced, the new deterministic file IDs for materialized history images are scoped by session ID + media type + content bytes (so identical bytes across sessions never collide), `on_disk_name`'s content-type/extension allowlist keeps the content-derived UUID path-traversal-safe, the UTF-8-safe tool-result cap is correct (verified the `is_char_boundary` walk-back terminates), the `ErrorTranslator` is correctly scoped to a pre-`.await` block (avoiding the known `!Send` handler trait-bound trap), the empty-session and populated-session response shapes now match field-for-field, the changelog fragment is present and already correctly formatted/named, and `cargo fmt --check` / `git diff --check` are clean on this diff.

One process finding left as an inline comment: commit `5d0475d` in this branch has `Claude <noreply@anthropic.com>` as both author and committer, which the repo's own `CLAUDE.md` policy says should be blocked. Left for a maintainer to decide how to handle (rebase, or let it get squashed on merge) since fixing it requires a history rewrite I'm not authorized to force-push.

No code-level bugs found; not blocking anything, just flagging.

---

crates/librefang-api/src/routes/agents/sessions.rs
```diff
@@ -1,5 +1,100 @@
...
+                return None;
+            }
+            remove_history_image_temp(&temporary).await;
+        }
+    }
+
+    let filename = format!("image.{}", media_type.rsplit('/').next().unwrap_or("png"));
+    UPLOAD_REGISTRY
+        .entry(file_id.clone())
+        .or_insert_with(|| UploadMeta {
```
houko, line 87: `materialize_history_image` only registers metadata in the in-memory `UPLOAD_REGISTRY`; it never writes a durable sidecar the way `#6922` (fix/upload-metadata-hardening) adds for the main upload path via `persist_upload_meta`.
Both PRs branch from the same `main` commit and don't touch the same files, so they'll merge cleanly with no textual conflict — but they change the shared `serve_upload` contract together: once `#6922` lands, a registry miss with no sidecar is treated as "unknown, fail closed for authenticated non-admin callers" instead of the old permissive default.
A history image materialized here survives a restart on disk (the bytes are content-addressed and already present), but its metadata does not — so after any daemon restart, `GET /api/uploads/{file_id}` for one of these images will 403 for authenticated non-admin callers until the owning session is reloaded through `get_agent_session` again (which repopulates the registry via the same `or_insert_with`).
That's a recurring gap for every restart, not just a one-time migration artifact, since every new history image materialized here inherits it.

Worth considering before both land: either persist a sidecar here too (once `#6922`'s `persist_upload_meta`/`UploadMetadata` exist on `main`), or explicitly document that history-thread images are restart-fragile by design. Flagging for maintainer judgment since fixing it here would mean depending on types that don't exist on this branch yet.

---

houko (commented):
Deep-dived the session-history hardening changes (translator-before-await restructuring, the content-addressed idempotent image materialization, the UTF-8-safe tool-result cap) — all solid and well tested. One cross-PR integration note left inline about metadata durability interacting with #6922's fail-closed access policy.

---

crates/librefang-api/src/routes/agents/sessions.rs
houko, line 87: Cross-PR heads-up rather than a bug in this diff by itself: `materialize_history_image` only inserts the new `UploadMeta` into the in-memory `UPLOAD_REGISTRY` — it never persists a durable sidecar next to the image file.

That's correct against this PR's own base (today's `main`), where `UPLOAD_REGISTRY` is the only source of truth `serve_upload` consults.

But #6922 (`fix/upload-metadata-hardening`, also open right now) changes that contract: it adds a `persist_upload_meta` sidecar file next to every upload, and makes `serve_upload` **fail closed** (403 for any authenticated non-admin caller) whenever no sidecar can be found after a registry miss — precisely to close the "registry wiped on restart" IDOR.

If both PRs land as they stand today, session-history images materialized by this function will pass fine until the daemon restarts, after which the in-memory registry is empty, there is no `.{file_id}.meta.json` sidecar to reload, and `serve_upload`'s new fail-closed path will return 403 to any authenticated non-admin caller trying to view an image in their own session history — a functional regression, not merely a missed hardening opportunity.

Suggest whichever of #6921 / #6922 lands second rebases so `materialize_history_image` also calls the sidecar-persistence helper #6922 introduces (or the two PRs get coordinated before merge). Flagging here since fixing it cleanly requires a helper this PR's own branch doesn't have yet.

---

houko (commented):
Reviewed the session-history hardening changes in sessions.rs and the new integration test. The dedup/materialization logic (content-derived UUIDv8, atomic temp+rename, concurrent-safe) and the UTF-8-safe tool-result cap both look correct and are well covered by the new tests. One cross-PR coordination risk flagged inline against #6922, which is open concurrently and changes the contract this file's `UPLOAD_REGISTRY` usage relies on.

---

PR #6922, opened 2026-08-10
fix(api): harden upload metadata and access control

houko:
## Summary

- persist upload filename, MIME type, and owner in durable atomic sidecars and reload them after registry loss
- fail closed for unknown authenticated non-admin callers and enforce the same policy for download, agent message, stream, WebSocket, and hand attachment consumers
- mark daemon-generated browser screenshots and generated images explicitly shared so authenticated users can consume their URLs
- move upload resolution and reads to async filesystem APIs, clean partial writes, and report the configured upload cap accurately

## Security

This closes the registry-miss IDOR after restart and the alternate attachment-injection path where a leaked foreign UUID could be passed directly to an agent. WebSocket authentication now carries the same synthetic root identity as HTTP in no-auth mode.

## Verification

- API integration tests: 93 passed
- upload metadata unit tests: 4 passed
- runtime shared-upload test: passed
- shared media contract test: passed
- API/types/runtime all-target clippy passed, aside from explicitly allowing an unrelated pre-existing nonminimal_bool lint in system.rs
- cargo fmt --all -- --check
- git diff --check

Independent review: 0 Critical / 0 Important / 0 Minor; ready to merge.


crates/librefang-api/src/routes/agents/uploads.rs
```diff
@@ -15,26 +15,89 @@ struct UploadResponse {
...
+    caller: Option<&crate::middleware::AuthenticatedApiUser>,
+) -> bool {
+    use crate::middleware::UserRole;
+
+    match meta.and_then(|meta| meta.uploaded_by) {
+        Some(owner_id) => {
+            caller.is_some_and(|user| user.user_id == owner_id || user.role >= UserRole::Admin)
+        }
+        None if meta.is_some() => true,
+        None => caller.is_none_or(|user| user.role >= UserRole::Admin),
```
houko, line 86: The `None => caller.is_none_or(|user| user.role >= UserRole::Admin)` branch handles a registry/sidecar "true miss" (no `UploadMetadata` found on disk at all).
`Option::is_none_or` returns `true` when the option is `None`, so a caller of `None` is treated the same as an admin here — i.e. **no caller information at all is granted the same access as an admin** for this branch, while an authenticated non-admin caller (`Some(user)` with `role < Admin`) is denied.

Concretely reachable `caller == None` cases:
- `middleware.rs`'s documented "legacy trusted-anonymous" session path (a session token that resolves in `active_sessions` but whose persisted `user_name`/`user_role` are `None` — the comment there explicitly says these "continue through as trusted-anonymous"). That's a genuinely-authenticated request, just one middleware can't attribute to an identity, and it now gets *more* access to true-miss uploads than a fully-attributed regular user.
- `load_upload_meta`'s parse-failure path (`serde_json::from_slice` error) also returns `None` and is logged as a `warn!`, but from `upload_access_allowed`'s point of view it's indistinguishable from "never had an owner." If a previously-owned upload's sidecar (`.{file_id}.meta.json`) is ever missing, truncated, or corrupted (crash mid-write outside the atomic rename window, manual tampering, disk issue), access silently reopens to *any* unauthenticated/unattributed caller instead of failing closed — the opposite of the fail-closed intent stated in the PR description ("fail closed for unknown authenticated non-admin callers").

Is the intent here really "no identity == admin-equivalent for shared content," or should this arm treat `caller: None` the same as "not privileged" (i.e. `false` unless `meta.is_some()`, matching the deliberate "explicitly shared" `None if meta.is_some() => true` arm two lines up)? Given this PR's stated goal is closing an IDOR/fail-open bug, I'd want to see this confirmed intentional (and ideally distinguish "corrupt/missing sidecar for content that image_generate/TTS admittedly never owned" from "sidecar genuinely never existed") rather than relying on `is_none_or`'s coincidental true-on-None semantics.

---

houko (commented):
Automated daily review pass: mechanical prose-wrap fixes pushed directly (see commit); one authz edge case flagged inline for maintainer judgment rather than auto-fixed.

---

crates/librefang-api/src/routes/agents/uploads.rs
```diff
@@ -15,26 +15,89 @@ struct UploadResponse {
...
+    file_id: &str,
+    meta: &UploadMeta,
+) -> Result<(), String> {
+    let bytes = serde_json::to_vec(meta)
+        .map_err(|error| format!("failed to serialize upload metadata: {error}"))?;
+    crate::atomic_write(&upload_meta_path(upload_dir, file_id), &bytes)
+        .map_err(|error| format!("failed to persist upload metadata: {error}"))
+}
+
+pub(crate) async fn persist_upload_meta(
```
houko, line 40: Not all `UPLOAD_REGISTRY` producers were migrated to this durable-sidecar contract, so the fix looks incomplete relative to its stated goal.

`crates/librefang-api/src/routes/agents/sessions.rs` (in `get_agent_session`, around line 150) still does the old in-memory-only thing on every session-history read that contains an inline `ContentBlock::Image`:

```rust
UPLOAD_REGISTRY.insert(
    file_id.clone(),
    UploadMeta {
        filename: format!("image.{}", ...),
        content_type: media_type.clone(),
        uploaded_by: None,
    },
);
```

It never calls `persist_upload_meta`/`persist_upload_meta_sync` (this file's new functions) to write the `.{file_id}.meta.json` sidecar, and it re-mints a fresh `file_id` + on-disk file on every read of the session.

Consequence under the new `upload_access_allowed` policy: if the in-memory registry entry for one of these history-embedded images is evicted (daemon restart, or simply if this particular re-materialized `file_id` is never re-inserted before the next `/api/uploads/{file_id}` GET), the lookup falls into the `meta.is_none()` branch — "true miss, unknown ownership" — which now fails closed for any authenticated non-admin caller. Before this PR that branch defaulted to open (owner `None` was allowed unconditionally), so this specific path degrades from "always worked" to "sometimes 403s for ordinary users looking at their own chat history," rather than gaining the intended protection (the file has no real owner to protect in the first place — it's the session owner's own data).

This isn't a new hole this PR opens (no cross-user leak — `uploaded_by: None` here is not attacker-controlled), but it does mean the sidecar-persistence guarantee described in the PR body ("persist upload filename, MIME type, and owner in durable atomic sidecars … reload them after registry loss") is not actually applied to every producer, and the "same policy" enforcement claim doesn't fully hold. Worth either wiring this call site through `persist_upload_meta` too (mirrors what `routes/media.rs::save_upload` and `runtime::uploaded_file::save_shared_upload` now do for the same "shared, no human owner" case), or explicitly noting the gap/follow-up if it's intentionally out of scope for this PR.

---

houko (commented):
Automated review pass.

Verified: durable-metadata reload, fail-closed default for unregistered uploads, and the owner check ordering (check-before-read) across `serve_upload`, `resolve_attachments`, `send_message`/`send_message_stream`, `hand_send_message`, and the WebSocket text-message path — all reject before the sensitive read/dispatch, matching the PR's stated scope. Tests added for each surface look sound (owner/stranger/admin/unowned-shared matrix, registry-miss reload, WS structured-error path).

Fixed directly (mechanical, no logic change): four new comment blocks added by this PR were hard-wrapped mid-sentence instead of breaking at sentence boundaries per project convention — rewrapped in `attachments.rs`, `uploads.rs`, `routes/media.rs`, and `librefang-types/src/media.rs`.

Left one inline finding: `crates/librefang-api/src/routes/agents/sessions.rs` has a producer of `UPLOAD_REGISTRY` entries (session-history image rehydration) that was not migrated to the new durable-sidecar contract, so the "persist metadata everywhere so restart-safe" claim doesn't fully hold for that path. Not a new security hole (no attacker-controlled owner there), but worth a look — see inline comment.




---

crates/librefang-api/src/routes/agents/uploads.rs
houko, line 86: Design question on the fully-unknown-metadata branch (`meta.is_none()`, i.e. neither an in-memory registry entry nor a `.meta.json` sidecar on disk).

`caller.is_none_or(|user| user.role >= UserRole::Admin)` returns `true` when `caller` is `None`.
That means a request with no `AuthenticatedApiUser` extension at all is allowed through even though the same fully-unknown metadata is denied to any authenticated non-admin caller — the opposite of what I'd expect from a fail-closed policy.

Walking `middleware.rs`, I could not find a path that reaches this handler with `api_user == None` in production: `is_public`/`PUBLIC_ROUTES_*` don't list this route, and every other path through the auth gate inserts either a real `AuthenticatedApiUser` or the synthetic root/Owner identity for no-auth/loopback mode.
So this branch looks unreachable today, and the unit test (`upload_access_allowed(None, None) == true`) documents the case rather than exercising a real caller.

Given the PR's own stated goal — "fail closed for unknown authenticated non-admin callers" — should the `None` caller in this branch also fail closed (deny) for defense-in-depth, in case a future caller of `upload_access_allowed`/`resolve_attachments` (a new consumer, an internal/system call site, a future public-route addition) legitimately passes `caller = None`? Happy to be told this is intentionally scoped to "authenticated" callers only and the current behavior is fine — flagging because it's an authorization-policy call I'd rather have the maintainer confirm than guess on.

---

houko (commented):
Reviewed the upload-metadata and access-control hardening. Left one inline question on an authorization-policy edge case rather than guessing; applied two small mechanical fixes (clippy warning + logging/error-code parity) that don't touch any authorization logic. Details in the review comment and commit message.

---

crates/librefang-api/src/routes/agents/uploads.rs
```diff
@@ -227,15 +307,16 @@ pub(crate) fn resolve_existing_upload_path(
         content_type,
         filename,
     ));
-    if named.exists() {
+    if tokio::fs::try_exists(&named).await.ok()? {
         return Some(named);
     }
     let bare = dir.join(file_id);
-    if bare.exists() {
+    if tokio::fs::try_exists(&bare).await.ok()? {
```
houko, line 314: `resolve_existing_upload_path_async` replaced the old synchronous `Path::exists()` probes (which never fail — any I/O error collapses to `false`) with `tokio::fs::try_exists(...).ok()?` for the first two candidates.
`try_exists` returns `Err` for I/O errors other than not-found (e.g. permission denied on a misconfigured upload dir), and `.ok()?` turns that `Err` into an early `return None` for the *whole function* — aborting before the `bare` fallback and the `<file_id>.*` prefix scan ever run, instead of falling through the way `.exists()` used to.

Net effect: on an I/O error probing the primary candidate path, a file that would still be found via the bare-id or prefix-scan fallback now resolves to `None`, i.e. `serve_upload` / attachment resolution 404s instead of serving the file.

Suggested fix — treat the error the same as "not found here" and keep falling through, matching the old `.exists()` semantics (leave the third `tokio::fs::read_dir(dir).await.ok()?` as-is; it mirrors the original `std::fs::read_dir(dir).ok()?` and isn't a regression):

```rust
    // `try_exists` (unlike the `Path::exists()` this replaced) can return an
    // I/O error (e.g. permission denied) rather than `Ok(false)`. Treat that
    // the same as "not found here" so the probe still falls through to the
    // remaining candidates instead of aborting the whole resolution early.
    if tokio::fs::try_exists(&named).await.unwrap_or(false) {
        return Some(named);
    }
    let bare = dir.join(file_id);
    if tokio::fs::try_exists(&bare).await.unwrap_or(false) {
        return Some(bare);
    }
```

I drafted this locally but couldn't get a scoped `cargo check -p librefang-api --lib` to finish in reasonable time under the current build contention, so I'm leaving it as a suggestion rather than pushing an unverified commit.

---

houko (commented):
Reviewed the durable-sidecar persistence, fail-closed access control, and the WS synthetic-root-identity mirroring — all solid and well tested. One correctness regression found in the sync→async path-resolution conversion, left inline with a suggested fix.

---

PR #6940, opened 2026-08-12
fix(runtime): use vetted SHA-256 for plugin integrity

houko:
## Summary

- replace the hand-rolled SHA-256 implementation in the plugin integrity path with the existing `sha2` dependency
- preserve the public `sha256_hex` API and lowercase 64-character digest format
- remove the redundant digest wrapper
- add standard vectors covering empty input and SHA-256 padding boundaries

## Verification

- `cargo test -p librefang-runtime sha256_hex_matches_standard_vectors_across_padding_boundaries --lib`
- `cargo test -p librefang-runtime 'plugin_manager::tests::' --lib` (33 passed, 2 ignored)
- `cargo clippy -p librefang-runtime --lib -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

## Out of scope

- other hand-rolled cryptographic code identified by the repository scan
- changes to plugin checksum formats or registry policy

### DIFF
--- a/changelog.d/security/6940-plugin-checksum-sha2.md
+++ b/changelog.d/security/6940-plugin-checksum-sha2.md
@@ -1,0 +1,3 @@
+Replace the hand-rolled SHA-256 implementation in the plugin integrity path with the workspace's existing vetted `sha2` crate.
+The hand-rolled version was never audited and carried a stale comment suggesting a future swap to `sha2` that never happened, leaving plugin checksum verification resting on unreviewed cryptographic code.
+The public `sha256_hex` API and its lowercase 64-character hex digest format are unchanged, so no caller or stored checksum is affected (#6940) (@houko)
--- a/crates/librefang-runtime/src/plugin_manager/tests.rs
+++ b/crates/librefang-runtime/src/plugin_manager/tests.rs
@@ -9,6 +9,36 @@ use super::*;
 use crate::test_env::ENV_LOCK;
 
 #[test]
+fn sha256_hex_matches_standard_vectors_across_padding_boundaries() {
+    let vectors: &[(&[u8], &str)] = &[
+        (
+            b"",
+            "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
+        ),
+        (
+            b"abc",
+            "ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad",
+        ),
+        (
+            &[b'a'; 55],
+            "9f4390f8d30c2dd92ec9f095b65e2b9ae9b0a925a5258e241c9f1e910f734318",
+        ),
+        (
+            &[b'a'; 56],
+            "b35439a4ac6f0948b6d6f9e3c6af0f5f590ce20f1bde7090ef7970686ec6738a",
+        ),
+        (
+            &[b'a'; 64],
+            "ffe054fe7ae0cb6dc65c3af9b61d5209f439851db43d0ba5997337df154668eb",
+        ),
+    ];
+
+    for (input, expected) in vectors {
+        assert_eq!(sha256_hex(input), *expected);
+    }
+}
+
+#[test]
 fn test_plugins_dir() {
     let _guard = ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner());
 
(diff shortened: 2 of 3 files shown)


changelog.d/security/6940-plugin-checksum-sha2.md
```diff
@@ -0,0 +1,3 @@
+Replace the hand-rolled SHA-256 implementation in the plugin integrity path with the workspace's existing vetted `sha2` crate.
```
houko, line 1: The commit that added this file (d02ecfa5, "chore(changelog): add security fragment for PR #6940") has its author and committer identity set to `Claude <noreply@anthropic.com>`.
That's exactly the pattern `scripts/hooks/commit-msg` is written to reject ("separately rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean," per CLAUDE.md) — it must have landed through a path that bypassed the local hook.
Since this commit is already the pushed branch tip, fixing it means amending the author/committer and force-pushing, which is outside what an automated reviewer should do unilaterally on someone else's branch; flagging for the PR author to correct instead.
Smaller nit while here: the commit body ("The plugin-checksum SHA-256 vetting fix was missing a changelog.d\nfragment per repo convention.") is hard-wrapped mid-sentence rather than one-sentence-per-line, which the repo's prose rule also covers for commit message bodies.

---

houko (commented):
Reviewed the sha2 swap in plugin_manager.rs: the implementation, callers, and new standard-vector test are all correct (verified the five hashes against known SHA-256 test vectors, confirmed `sha256_hex_of_bytes` has no remaining references, and ran a scoped `cargo check -p librefang-runtime --lib`, which finished clean). The changelog.d fragment is correctly formatted and passes `scripts/check-changelog-attribution.py --all-unreleased`. One finding left as an inline comment: the changelog-fragment commit's author/committer identity is set to an AI attribution string, which the repo's own commit-msg hook is designed to reject — left for the PR author to correct since fixing it requires amending a commit that's already the pushed branch tip.

---

changelog.d/security/6940-plugin-checksum-sha2.md
houko, line 1: Governance: commit `d02ecfa` on this branch ("chore(changelog): add security fragment for PR #6940") is authored by `Claude <noreply@anthropic.com>` (confirmed via `git log --format='%an <%ae>'` and the GitHub commits API — not just a local config artifact).
CLAUDE.md is explicit that commit author identity resolving to Claude/Anthropic is disallowed (it's exactly what `scripts/hooks/commit-msg`'s author-identity check exists to reject; that hook just wasn't enabled when this landed).
Please re-author this commit (e.g. `git commit --amend --author "houko <...>"` or squash it into the primary commit under your own identity) before merge.
Flagging as a review comment rather than pushing a fix myself, since any commit I push here would carry that same AI-attributed author identity — fixing it that way would just add a second instance of the problem it's pointing at.

---

crates/librefang-runtime/src/plugin_manager/tests.rs
houko, line 1: Test coverage gap: `sha256_hex_matches_standard_vectors_across_padding_boundaries` (added here) only checks that `sha256_hex` matches known-good vectors, and the pre-existing suite never calls `verify_checksum` at all — grepped the crate, the only call site is `install.rs:198`, with zero direct tests.
So there's currently no test anywhere asserting that a tampered/mismatched plugin download actually gets rejected, only that a correct digest is computed/accepted.
Worth adding something like:

```rust
#[test]
fn verify_checksum_rejects_tampered_plugin_bytes() {
    let original = b"plugin archive bytes";
    let expected = sha256_hex(original);
    let mut tampered = original.to_vec();
    tampered[0] ^= 0x01;
    assert!(verify_checksum(&tampered, &expected).is_err());
}
```

`verify_checksum` is private but reachable from `tests.rs` via `use super::*;`, so no visibility change is needed.
(I verified the new `sha256_hex_matches_standard_vectors_across_padding_boundaries` vectors themselves are all correct against `hashlib.sha256` for `""`, `"abc"`, and the 55/56/64-byte padding-boundary cases — that part's solid.)

---

houko (commented):
Reviewed the sha2 migration diff. Crate choice, comparison logic, and old-hash cleanup all look correct — the two items below are worth addressing before merge.

---

PR #6941, opened 2026-08-12
fix(llm): use vetted RS256 for Vertex JWTs

houko:
## Summary

- replace the hand-rolled Vertex AI service-account JWT signer with the workspace `jsonwebtoken` RS256 implementation
- remove the custom PEM/ASN.1 parser, PKCS#1 v1.5 padding, big-integer arithmetic, and modular exponentiation
- preserve the service-account claim set and OAuth assertion exchange
- add coverage that signs with a generated PKCS#8 RSA key and verifies with the corresponding public key
- cover invalid private-key rejection

## Verification

- `cargo test -p librefang-llm-drivers 'drivers::vertex_ai::tests::service_account_jwt' --lib`
- `cargo test -p librefang-llm-drivers 'drivers::vertex_ai::tests::' --lib` (5 passed)
- `cargo clippy -p librefang-llm-drivers --lib -- -D warnings`
- `cargo fmt --check`
- `git diff --check`

## Out of scope

- live Google OAuth exchange, which requires real service-account credentials
- unrelated Vertex request and retry behavior


crates/librefang-llm-drivers/src/drivers/vertex_ai.rs
```diff
@@ -187,369 +180,20 @@ impl TokenManager {
...
-        while digits.len() > 1 && *digits.last().unwrap() == 0 {
-            digits.pop();
-        }
-        BigUint { digits }
-    }
+fn encode_service_account_jwt(
+    private_key_pem: &str,
+    claims: &serde_json::Value,
+) -> Result<String, LlmError> {
+    let key = jsonwebtoken::EncodingKey::from_rsa_pem(private_key_pem.as_bytes())
```
houko, line 189: Not a blocker for this PR — the `jsonwebtoken`/`aws_lc_rs` swap itself looks solid (well-vetted crate, claims unchanged from the original `iss`/`scope`/`aud`/`iat`/`exp` set, and the new test signs with a generated PKCS#8 key and verifies with the matching public key, plus a separate invalid-key-rejection test).

One thing worth a second look since we're already touching this code path: `private_key_pem` here comes from `sa_json["private_key"]` where `sa_json` is a `serde_json::Value` held for the lifetime of `CredentialSource::ServiceAccountJson` (see line 35) — the raw PEM stays in un-zeroized heap memory for as long as the driver instance lives, and every token refresh re-derives an owned `String`/`&str` from it without ever scrubbing the old copies. `zeroize` is already a dependency of this crate. This predates this PR (the old hand-rolled signer had the same exposure), so I'm not asking for a fix in this change — just flagging it in case it's worth a quick follow-up given the file is already open for a JWT-signing security fix.

Separately, it'd be good for whoever lands this to double check that `jsonwebtoken::errors::Error`'s `Display` for a malformed RSA PEM (used in the `map_err` right below) never echoes back key bytes — I didn't find evidence it does (errors from the `pem`/`rsa` decode path are typically structural, e.g. "invalid PEM", not content-bearing), but it's worth a quick confirmation given this message can reach logs/error responses.

---

houko (commented):
Reviewed the RS256 replacement. The move to `jsonwebtoken` (with the `aws_lc_rs` backend) is a sound choice — well-vetted, actively maintained, and it fully replaces the hand-rolled PEM/ASN.1 parser, PKCS#1 v1.5 padding, and from-scratch big-integer modpow (all deleted). The `rsa` crate is dev-only (test key generation), not part of the production signing path.

Verified:
- Claims (`iss`, `scope`, `aud`, `iat`, `exp`) are unchanged from the pre-existing implementation — no semantic drift in the service-account OAuth2 assertion.
- New test signs with a generated PKCS#8 RSA key and verifies with the matching public key (asserts `alg=RS256` and claim equality), plus a separate test asserting an invalid private key is rejected. This satisfies the "sign then verify" round-trip bar, not just a compile check.
- Changelog fragment is present at `changelog.d/security/6941-vertex-jwt-rs256.md`, correctly filed under `security/`, ends with `(#6941) (@houko)`.
- No AI/Claude attribution in either commit.
- `cargo check -p librefang-llm-drivers --lib` passes clean on this branch.

Left one non-blocking inline comment on a pre-existing (not introduced by this PR) private-key-lifetime observation, plus a note asking for a quick confirmation that the `jsonwebtoken` error path doesn't echo key material.

---

PR #6947, opened 2026-08-12
fix(api): harden skill secret writes

houko:
## Summary

- give skill and hand secret writes unique, exclusive staging files so concurrent callers cannot truncate each other
- remove secret-bearing staging files after write, sync, or rename failures
- fsync the parent directory on Unix after replacing `secrets.env`

## Testing

- `cargo test -p librefang-api routes::skills::tests::write_secret_env --lib`
- `cargo test -p librefang-api routes::skills::tests::atomic_secret_write_cleans_up_after_rename_failure --lib`
- `git diff --check`

### DIFF
--- a/crates/librefang-api/src/routes/skills/mod.rs
+++ b/crates/librefang-api/src/routes/skills/mod.rs
@@ -1076,13 +1076,17 @@ pub(crate) fn remove_secret_env(path: &std::path::Path, key: &str) -> Result<(),
 
 /// Atomically replace `path` with `content`, ensuring the resulting
 /// inode is mode `0600` (Unix) from creation — never observable at
-/// the process umask. Writes to a sibling `.tmp` file first to keep
+/// the process umask. Writes to a sibling temp file first to keep
 /// the rename within the same filesystem (so `rename(2)` is
-/// atomic). On non-Unix targets the helper still uses the temp +
-/// rename shape so partial writes can't tear the file; the
-/// per-permissions bit is a no-op (Windows ACLs are inherited from
-/// the parent directory, which lives under the daemon-UID user
-/// profile).
+/// atomic). The temp file name is suffixed with the process ID and
+/// a per-process atomic counter so concurrent callers never open
+/// the same staging file and truncate each other's write. On
+/// non-Unix targets the helper still uses the temp + rename shape
+/// so partial writes can't tear the file; the per-permissions bit
+/// is a no-op (Windows ACLs are inherited from the parent
+/// directory, which lives under the daemon-UID user profile). On
+/// Unix, the parent directory is fsynced after a successful rename
+/// so the replacement survives a crash immediately afterward.
 fn atomic_write_secret_file(path: &std::path::Path, content: String) -> Result<(), std::io::Error> {
     use std::io::Write as _;
     let parent = path.parent().unwrap_or_else(|| std::path::Path::new("."));
--- a/crates/librefang-api/src/routes/skills/mod.rs
+++ b/crates/librefang-api/src/routes/skills/mod.rs
@@ -15,6 +15,7 @@ use axum::response::IntoResponse;
 use axum::Json;
 use librefang_types::i18n::ErrorTranslator;
 use std::collections::HashMap;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Instant;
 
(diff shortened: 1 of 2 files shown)


crates/librefang-api/src/routes/skills/mod.rs
```diff
@@ -1090,29 +1095,38 @@ fn atomic_write_secret_file(path: &std::path::Path, content: String) -> Result<(
...
+    }
 
     // `rename(2)` is atomic — the destination either contains the
     // old bytes (pre-rename) or the new bytes (post-rename); a
     // concurrent reader never observes a half-written file.
     match std::fs::rename(&tmp_path, path) {
-        Ok(()) => Ok(()),
+        Ok(()) => {
+            #[cfg(unix)]
+            std::fs::File::open(parent)?.sync_all()?;
```
houko, line 1127: By this point `rename(2)` has already succeeded, so the new secret content is live on disk under `path`. If `File::open(parent)` or `sync_all()` on the directory fails here, the `?` turns that into an `Err` for the whole function even though the write itself did not fail.

At the one call site that matters (`hands.rs`, the `write_secret_env` branch in the settings handler around line ~1097), an `Err` here makes the handler return a 500 to the client and skip `set_env_var_guarded`, so the running process keeps the old value while `secrets.env` on disk already has the new one — a state split between "API says failed" and "file says succeeded" that a client can't safely react to (retrying just re-does the same rename).

Worth considering: log a `WARN` and still return `Ok(())` when only the post-rename directory fsync fails, since the atomicity guarantee (no partial file ever observable) already held before this line — only the extra crash-durability guarantee added by this PR is at risk, not the write itself. Not proposing this as a blocking issue since there's a legitimate argument either way (the PR's whole point is durability, so silently downgrading a fsync failure to a warning could itself be seen as undermining the guarantee) — flagging for a decision rather than editing it directly.

---

houko (commented):
Automated review pass. Pushed one mechanical fix (doc comment was out of date after this PR's own logic changes); left one design-level question inline rather than editing it, since the PR's own topic (durability) makes the tradeoff genuinely two-sided. Noted that a changelog fragment for #6947 landed concurrently from another push, so no duplicate was added.

---

crates/librefang-api/src/routes/skills/mod.rs
```diff
@@ -1090,29 +1095,38 @@ fn atomic_write_secret_file(path: &std::path::Path, content: String) -> Result<(
         .map(|n| n.to_owned())
         .unwrap_or_else(|| std::ffi::OsString::from("secrets.env"));
     let mut tmp_name = file_name;
-    tmp_name.push(".tmp");
+    static TMP_SEQ: AtomicU64 = AtomicU64::new(0);
+    let seq = TMP_SEQ.fetch_add(1, Ordering::Relaxed);
+    tmp_name.push(format!(".tmp.{}.{seq}", std::process::id()));
```
houko, line 1100: The core durability shape here (create_new + write + sync_all, remove staging file on write/sync/rename failure, fsync the parent dir on Unix after a successful rename) matches the other durable-write PRs landing this week (#6942 sidecar_toml.rs, #6944 secrets_env.rs, #6945 cron_script.rs) — good, no divergence on the guarantees themselves.
One cosmetic inconsistency worth a look: the staging name here is `secrets.env.tmp.<pid>.<seq>` (visible in a directory listing), whereas `secrets_env.rs`'s `upsert_secret` (the near-duplicate secrets-file writer touched by #6944) uses a dot-prefixed `.secrets.env.tmp.<pid>.<seq>` (hidden). Not a correctness bug, but two writers of effectively the same file format landing in the same week with different hidden/visible staging conventions is the kind of drift that's cheap to align now and easy to forget later.
Separately, and out of scope for this PR specifically: `secrets_env.rs::upsert_secret` on `main` still opens its temp file with `create(true).truncate(true)` and applies `0600` via a follow-up `fs::set_permissions` call *after* open, rather than `OpenOptions::mode(0o600)` at creation time the way this PR's `atomic_write_secret_file` does. That reintroduces the exact TOCTOU window (`0644`-by-umask window between create and chmod) this PR's own comment cites as the reason for the mode-at-creation pattern. Worth a follow-up on whichever PR/branch touches `secrets_env.rs` next, since it's a real gap in a sibling secret-writing path, just not one this diff touches.

---

houko (commented):
Reviewed the durable-write hardening for skill secret writes against the sibling durable-write PRs landing this week (secrets_env.rs, sidecar_toml.rs, cron_script.rs). The core pattern (unique per-call staging name, create_new, write+sync with cleanup on failure, atomic rename with cleanup on failure, Unix parent-dir fsync on success) is consistent across all of them — no correctness issues found in this diff. Left one comment flagging a cosmetic staging-name convention difference plus a pointer to a real (but out-of-scope-for-this-PR) TOCTOU gap in secrets_env.rs's mode-setting that the other durable-write PR touching that file should pick up.

---

houko:
Added the missing `changelog.d/fixed/6947-skill-secret-durable-write.md` fragment (this PR's diff had none, unlike the sibling durability PRs in this batch).

Same note as on #6948/#6945: that fixup commit was pushed from an automated review session whose local git identity was misconfigured to `Claude <noreply@anthropic.com>` — exactly the attribution this repo's `commit-msg` hook exists to catch. I've corrected the identity for the rest of this run. Since the bad commit is already on your branch, flagging it here rather than rewriting history myself — worth squashing or re-authoring before merge. Apologies for the noise.

---

PR #6948, opened 2026-08-12
fix(kernel): make TOML renames durable

houko:
## Summary

- create kernel TOML staging files exclusively instead of truncating an unexpected existing path
- fsync the parent directory on Unix after publishing a staged TOML file by rename
- preserve the existing unique-name, file-sync, failure-cleanup, and atomic-replace behavior

## Testing

- `cargo test -p librefang-kernel atomic_write_ --lib`
- `git diff --check`

### DIFF
--- a/crates/librefang-kernel/src/kernel/cron_script.rs
+++ b/crates/librefang-kernel/src/kernel/cron_script.rs
@@ -169,7 +169,10 @@ pub(super) fn atomic_write_toml(path: &Path, content: &str) -> std::io::Result<(
     tmp.set_file_name(tmp_name);
 
     let write_result = (|| -> std::io::Result<()> {
-        let mut f = std::fs::File::create(&tmp)?;
+        let mut f = std::fs::OpenOptions::new()
+            .write(true)
+            .create_new(true)
+            .open(&tmp)?;
         f.write_all(content.as_bytes())?;
         // fsync so the bytes hit disk before we publish via rename;
         // without this a power loss between rename and flush would
@@ -191,6 +194,15 @@ pub(super) fn atomic_write_toml(path: &Path, content: &str) -> std::io::Result<(
         let _ = std::fs::remove_file(&tmp);
         return Err(e);
     }
+
+    #[cfg(unix)]
+    path.parent()
+        .ok_or_else(|| {
+            std::io::Error::new(std::io::ErrorKind::InvalidInput, "missing parent directory")
+        })
+        .and_then(std::fs::File::open)?
+        .sync_all()?;
+
     Ok(())
 }
 
--- a/changelog.d/fixed/6948-cron-toml-durable-write.md
+++ b/changelog.d/fixed/6948-cron-toml-durable-write.md
@@ -1,0 +1,2 @@
+Cron script TOML writes opened their staging file with `File::create`, which truncates and silently reuses an existing file of the same name instead of failing loudly on a staging-name collision.
+The staging file is now opened with `create_new` so a collision surfaces as an error rather than being silently overwritten, and the parent directory is fsynced after the atomic rename on Unix so a completed write survives a crash immediately afterward (#6948) (@houko)


crates/librefang-kernel/src/kernel/cron_script.rs
```diff
@@ -191,6 +194,15 @@ pub(super) fn atomic_write_toml(path: &Path, content: &str) -> std::io::Result<(
...
         return Err(e);
     }
+
+    #[cfg(unix)]
+    path.parent()
+        .ok_or_else(|| {
+            std::io::Error::new(std::io::ErrorKind::InvalidInput, "missing parent directory")
+        })
+        .and_then(std::fs::File::open)?
+        .sync_all()?;
```
houko, line 204: Design question on failure mode: this makes a successful write+rename return an `Err` if the post-rename directory fsync fails, even though the file is already correctly renamed into place with correct content on disk — durability just can't be confirmed.

That seems right for common POSIX filesystems, but some environments librefang actually runs on (Docker overlay mounts, certain network filesystems used inside `librefang-runtime-sandbox-docker` sandboxes) are known to reject or no-op `fsync` on a directory fd (`EINVAL`/`ENOTSUP`). On those, every `atomic_write_toml` call would start hard-failing post-rename despite the write having actually succeeded — trading a real durability gap for a live availability regression on those mounts.

Is that the intended tradeoff, or should a directory-fsync failure be logged and swallowed (best-effort) rather than propagated as a write failure? Worth confirming intent, since it changes behavior for every caller (`agent_state.rs`, `agent_runtime.rs`, `workspace_setup.rs`).

---

crates/librefang-kernel/src/kernel/cron_script.rs
```diff
@@ -191,6 +194,15 @@ pub(super) fn atomic_write_toml(path: &Path, content: &str) -> std::io::Result<(
         let _ = std::fs::remove_file(&tmp);
         return Err(e);
     }
+
+    #[cfg(unix)]
+    path.parent()
```
houko, line 199: `path.parent()` returns `Some(Path::new(""))` (not `None`) when `path` is a bare relative filename with no directory component (e.g. `Path::new("agent.toml")`), since only a root/prefix-terminated path yields `None`. `std::fs::File::open("")` then fails, so this branch would turn a call that used to succeed (write + rename to cwd) into a hard error, purely because of the new directory-fsync step.

Every current call site builds `path` via `.join()` on a real directory (`agent_state.rs`, `agent_runtime.rs`, `workspace_setup.rs`), so this isn't reachable today — but it's a footgun for the next caller, and it's exactly the code this PR is adding. Worth either falling back to `.` when `parent()` yields an empty path, or documenting the "must be directory-qualified" precondition on `atomic_write_toml` explicitly so it's not silently reintroduced.

Also, the function's doc comment above (the "SECURITY / CORRECTNESS" block) still only describes the temp-file fsync before rename — it doesn't mention the new post-rename parent-directory fsync or explain why both are needed together (temp-file fsync alone doesn't make the *rename* durable, which is the actual point of this PR per the title). Would help future readers to update it alongside this change.

---

crates/librefang-kernel/src/kernel/cron_script.rs
```diff
@@ -169,7 +169,10 @@ pub(super) fn atomic_write_toml(path: &Path, content: &str) -> std::io::Result<(
     tmp.set_file_name(tmp_name);
 
     let write_result = (|| -> std::io::Result<()> {
-        let mut f = std::fs::File::create(&tmp)?;
+        let mut f = std::fs::OpenOptions::new()
+            .write(true)
+            .create_new(true)
+            .open(&tmp)?;
```
houko, line 175: Per CLAUDE.md's "Common Gotchas" ("Always write integration tests at the injection site, not just the implementation site") and the mandatory-testing guidance for kernel-internal changes, this PR doesn't add a test for either new behavior:

1. `create_new` turning a staging-name collision into an error instead of a silent overwrite.
2. The post-rename parent-directory fsync.

I didn't add one myself because a clean regression test for (1) isn't straightforward with the current structure — the tmp filename is `<file>.<pid>.<seq>.tmp` with `SEQ` a process-local atomic counter, so a test can't deterministically force two calls to collide without either exposing the naming logic for injection or relying on `std::process::id()` + a guessed `SEQ` value (racy under parallel `cargo test` execution in the same binary).

Existing tests (`atomic_write_replaces_existing_content`, `atomic_write_leaves_no_tmp_file_on_success`, `atomic_write_no_partial_state_under_concurrency` in `kernel/tests.rs`) do exercise the new directory-fsync path as a side effect of every successful write on this Linux CI lane, but none of them assert on it directly, and the Windows lane never exercises it since it's `#[cfg(unix)]`-gated.

Suggest either: (a) a small refactor exposing tmp-name generation so a test can pre-seed a real collision, or (b) at minimum a test that pre-creates an unrelated leftover `*.tmp` file in the directory (simulating an orphan from a prior crash) and asserts a fresh `atomic_write_toml` call still succeeds and ignores it, since that's the realistic "crash left a stray tmp file" scenario this change touches.

---

houko:
Added the missing `changelog.d/fixed/6948-cron-toml-durable-write.md` fragment (this PR's diff had none, unlike the sibling durability PRs in this batch).

One note: that fixup commit was made from an automated review session whose local git identity was misconfigured to `Claude <noreply@anthropic.com>` — exactly the attribution this repo's `commit-msg` hook is meant to catch, but the hook only runs when `core.hooksPath` is configured, which wasn't the case here. I've fixed the identity for the rest of this run so it won't recur, but this one commit is already pushed. Since it's on your branch, I'm flagging it rather than rewriting history myself — worth squashing or re-authoring before merge. Apologies for the noise.

---

houko (commented):
Reviewed the durable-write change in `atomic_write_toml`. The core approach (fsync temp file → atomic rename → fsync parent directory on Unix) is the right pattern for crash-durable renames, and the `create_new` staging-file change is a reasonable hardening. Left three inline notes: a design question on whether directory-fsync failure should be a hard error given some filesystems don't support it, a latent edge case in the new `path.parent()` handling, and a gap in test coverage for the new behavior. Nothing here looks incorrect, but the failure-mode question and the missing test seem worth resolving before merge.

---

PR #6950, opened 2026-08-12
fix(api): reject channel manifest traversal

houko:
## Summary

- reject channel bridge manifest names that are not a single normal path component
- block relative, absolute, nested, and Windows-style traversal inputs before constructing a manifest path
- avoid reflecting the unsafe path in the validation error

## Testing

- `cargo test -p librefang-api manifest_name_must_be_one_path_component --lib`
- `git diff --check`

### DIFF
--- a/changelog.d/security/6950-channel-manifest-path-traversal.md
+++ b/changelog.d/security/6950-channel-manifest-path-traversal.md
@@ -1,0 +1,2 @@
+`spawn_agent_by_name` built the agent manifest path directly from the channel-supplied manifest name, so a name containing `..`, a nested path, or an absolute path could resolve outside `~/.librefang/workspaces/agents/` and load an arbitrary `agent.toml` from elsewhere on disk.
+The manifest name is now validated to be exactly one normal path component before the lookup, rejecting empty names, `.`, `..`, embedded separators, and absolute paths on both Unix and Windows (#6950) (@houko)
--- a/crates/librefang-api/src/channel_bridge.rs
+++ b/crates/librefang-api/src/channel_bridge.rs
@@ -9,6 +9,17 @@ use librefang_channels::router::AgentRouter;
 use librefang_channels::sidecar::SidecarAdapter;
 use librefang_channels::types::{ChannelAdapter, SenderContext};
 
+fn is_safe_manifest_name(name: &str) -> bool {
+    if name.contains(['/', '\\']) {
+        return false;
+    }
+    let mut components = std::path::Path::new(name).components();
+    matches!(
+        (components.next(), components.next()),
+        (Some(std::path::Component::Normal(component)), None) if component == name
+    )
+}
+
 /// Sanitize LLM/driver errors into user-friendly messages for channel delivery.
 ///
 /// Prevents raw technical details (stack traces, driver internals, status codes)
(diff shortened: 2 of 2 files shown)


changelog.d/security/6950-channel-manifest-path-traversal.md
```diff
@@ -0,0 +1,2 @@
+`spawn_agent_by_name` built the agent manifest path directly from the channel-supplied manifest name, so a name containing `..`, a nested path, or an absolute path could resolve outside `~/.librefang/workspaces/agents/` and load an arbitrary `agent.toml` from elsewhere on disk.
+The manifest name is now validated to be exactly one normal path component before the lookup, rejecting empty names, `.`, `..`, embedded separators, and absolute paths on both Unix and Windows (#6950) (@houko)
```
houko, line 2: The second commit on this branch (`c151a8b` "docs(changelog): add missing fragment for #6950", which added this file) has author identity `Claude <noreply@anthropic.com>`.
Per CLAUDE.md's Git Conventions / the `commit-msg` hook, a commit whose author identity resolves to Claude/Anthropic is rejected even when the message text itself is clean — this one apparently reached the remote branch without going through that hook (e.g. pushed via an API path that doesn't invoke local git hooks).
Recommend the branch owner rebase to reauthor that commit under a real contributor identity (and fold it into the first commit, or `git commit --amend --reset-author`) before merge — this needs a rebase/force-push to the PR branch, which is outside what I'll do autonomously.

---

crates/librefang-api/src/channel_bridge.rs
```diff
@@ -9,6 +9,17 @@ use librefang_channels::router::AgentRouter;
 use librefang_channels::sidecar::SidecarAdapter;
 use librefang_channels::types::{ChannelAdapter, SenderContext};
 
+fn is_safe_manifest_name(name: &str) -> bool {
+    if name.contains(['/', '\\']) {
+        return false;
+    }
+    let mut components = std::path::Path::new(name).components();
+    matches!(
+        (components.next(), components.next()),
```
houko, line 18: Component-based validation correctly blocks `..`, embedded separators, and absolute paths (both `/` and `\`, plus Windows drive-relative forms like `C:foo` since that parses to a `Prefix` + `Normal` pair and fails the exact single-component match) — nice use of `Path::components()` instead of a substring blocklist.

One residual gap worth a second look: this only validates the *name string*, not the resolved filesystem path.
If `~/.librefang/workspaces/agents/<name>` can ever contain a symlink (planted by an earlier bug, a misconfigured shared workspace, or a race with agent creation) pointing outside the agents directory, a single-component `name` that matches an existing symlink would still let `spawn_agent_by_name` load an `agent.toml` from wherever that symlink points — the component check has no way to see through it.
Given the fix's own threat model is "channel-supplied name", this is probably out of scope (it needs local filesystem write access that channel input alone doesn't grant), but since the manifest path is opened without `canonicalize()` + prefix-check against the agents root, it's not fully closed by this diff.
Not blocking, but worth confirming that agents-directory entries are never attacker-influenced symlinks, or adding a canonicalized prefix check if that assumption doesn't hold long-term.

---

houko (commented):
Automated daily-review pass. Two items flagged inline for maintainer judgment; the core traversal fix (single-normal-path-component check) looks correct and the changelog fragment follows the required format. No mechanical fixes applied.

---

PR #6956, opened 2026-08-12
fix(kernel): prune empty tag index buckets

houko:
## Summary
- remove tag index buckets when their last agent is removed
- perform retain and conditional removal through a DashMap occupied entry guard
- preserve shared tag buckets that still contain another agent

## Testing
- `cargo test -p librefang-kernel remove_prunes_empty_tag_buckets --lib`
- `cargo test -p librefang-kernel test_remove --lib`
- `git diff --check`

### DIFF
--- a/crates/librefang-kernel/src/registry.rs
+++ b/crates/librefang-kernel/src/registry.rs
@@ -236,8 +236,11 @@ impl AgentRegistry {
             .remove(&id)
             .ok_or_else(|| LibreFangError::AgentNotFound(id.to_string()))?;
         for tag in &tags {
-            if let Some(mut ids) = self.tag_index.get_mut(tag) {
-                ids.retain(|&agent_id| agent_id != id);
+            if let Entry::Occupied(mut bucket) = self.tag_index.entry(tag.clone()) {
+                bucket.get_mut().retain(|&agent_id| agent_id != id);
+                if bucket.get().is_empty() {
+                    bucket.remove();
+                }
             }
         }
         self.notify_changed();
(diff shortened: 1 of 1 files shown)


crates/librefang-kernel/src/registry.rs
```diff
@@ -236,8 +236,11 @@ impl AgentRegistry {
             .remove(&id)
             .ok_or_else(|| LibreFangError::AgentNotFound(id.to_string()))?;
         for tag in &tags {
-            if let Some(mut ids) = self.tag_index.get_mut(tag) {
-                ids.retain(|&agent_id| agent_id != id);
+            if let Entry::Occupied(mut bucket) = self.tag_index.entry(tag.clone()) {
```
houko, line 239: Correctness check on the pruning fix itself: this looks race-safe. `DashMap::entry()` holds the shard's write lock for the entire lifetime of the returned `Entry` guard, so `bucket.get_mut().retain(...)` → `bucket.get().is_empty()` → `bucket.remove()` all execute under one uninterrupted lock acquisition. A concurrent `register()` calling `tag_index.entry(tag.clone()).or_default().push(id)` on the *same* tag will block on `entry()` until this guard drops (either via `bucket.remove()` or falling out of scope), so there's no window where a concurrent insert into a bucket being pruned gets lost. Verified `remove_prunes_empty_tag_buckets` locally: `cargo test -p librefang-kernel --lib registry::tests::remove_prunes_empty_tag_buckets` passes, and `cargo check -p librefang-kernel --lib` is clean. `tag_index` also has no current reader that would need `BTreeMap`-style deterministic ordering (no `find_by_tag`/prompt-facing accessor exists yet), so the #3298 ordering rule doesn't apply here.

One process gap: per CLAUDE.md, this is a `fixed:` category bug and needs a `changelog.d/fixed/6956-*.md` fragment (see `changelog.d/README.md` for the format). I don't see one in the diff — worth adding in a follow-up so the fix shows up in the next release notes.

---

Issue #3552, opened 2026-04-27
LlmError::TimedOut carries partial_text by value — multi-MiB cloned through error chain


houko:
## Summary
`LlmError::TimedOut` carries `partial_text: String` by value. `Display` only references `partial_text_len` — so the field is dead weight on every non-CLI driver, but a CLI driver (claude-code) callers that DO want it match on the variant and may carry megabytes. Each layer that boxes this into `LibreFangError::LlmDriver(String)` via `e.to_string()` drops the field anyway. Simultaneously huge and useless to 95% of consumers.

## Location
`crates/librefang-llm-driver/src/lib.rs:54-62`

```rust
#[error("Timed out after {inactivity_secs}s of inactivity (last: {last_activity}, {partial_text_len} chars partial output)")]
TimedOut {
    inactivity_secs: u64,
    partial_text: String,
    partial_text_len: usize,
    last_activity: String,
},
```

## Failure
- Error type now unbounded-size; every conversion / clone in the error chain pays it.
- `Box<dyn Error>` boxing through `LibreFangError::LlmDriver(String)` calls `.to_string()` which doesn't include `partial_text` → field is dropped silently in 95% of paths.
- 5% of paths (CLI) keep megabytes alive in the error variant.

## Fix
- Replace with `partial_text: Option<Arc<str>>` or `bytes::Bytes` — clone is O(1).
- Better: expose partial output via a side-channel `StreamEvent::PartialOutputCheckpoint`; keep `LlmError::TimedOut` as `{ inactivity_secs, last_activity }` only. Errors describe what failed; they shouldn't carry partial successful payload.

PR #6960, opened 2026-08-12
fix(llm): bound CLI subprocess execution

houko:
## Summary
- enforce configured message deadlines for Codex CLI, Gemini CLI, and both Qwen CLI execution paths
- let per-request `timeout_secs` override the driver default, matching Claude Code behavior
- kill and reap timed-out children while concurrently draining stdout/stderr without descendant-held pipe hangs
- include message timeout in the driver cache key so config changes cannot reuse stale drivers

## Testing
- `cargo test -p librefang-llm-drivers timeout --lib`
- `cargo test -p librefang-llm-drivers cache_key --lib`
- `cargo check -p librefang-llm-drivers --lib`
- `cargo clippy -p librefang-llm-drivers --lib --tests -- -D warnings`
- `git diff --check`


crates/librefang-llm-drivers/src/drivers/codex_cli.rs
```diff
@@ -420,6 +439,40 @@ fn home_dir() -> Option<std::path::PathBuf> {
 mod tests {
     use super::*;
 
+    #[cfg(unix)]
+    fn sleeping_cli() -> tempfile::NamedTempFile {
+        use std::os::unix::fs::PermissionsExt;
+
+        let mut file = tempfile::NamedTempFile::new().unwrap();
+        std::io::Write::write_all(&mut file, b"#!/bin/sh\nsleep 30\n").unwrap();
```
houko, line 453: `sleeping_cli()` keeps the `tempfile::NamedTempFile` (and its writable fd) alive for the whole test via the returned `cli` binding, then `complete()` tries to `exec()` that same path while the fd is still open for writing. On Linux that is exactly the ETXTBSY case — ran this test locally and it fails deterministically, not intermittently:

```
cargo test -p librefang-llm-drivers complete_honors_request_timeout --lib -- --nocapture
...
DEBUG_ACTUAL_ERROR: Http("Codex CLI not found or failed to start (Text file busy (os error 26)). Install: npm install -g @openai/codex")
thread '...complete_honors_request_timeout' panicked: assertion failed: matches!(error, LlmError::TimedOut { inactivity_secs: 0, .. })
```

So the new regression test never actually exercises the timeout path — it fails on spawn with an unrelated I/O error before the timeout logic ever runs, and it will fail this way on any Linux CI runner too (this isn't a sandbox artifact, it's standard POSIX `ETXTBSY` semantics for executing a file that's open for write). The identical helper in `qwen_code.rs` (`stream_honors_request_timeout`, lines 1167-1199) has the same bug and fails the same way ("Qwen Code CLI not found or failed to start (Text file busy (os error 26))").

Fix: close the write handle before exec'ing, e.g. write via `std::fs::write()` to a path inside a `tempfile::tempdir()` (which opens, writes, and closes in one call) and `chmod` afterward, rather than holding a `NamedTempFile` open across the `Command::spawn()`.

---

crates/librefang-llm-drivers/src/cli_process.rs
```diff
@@ -0,0 +1,136 @@
...
+    command.kill_on_drop(true);
+    let mut child = command.spawn().map_err(OutputError::Io)?;
+    let mut stdout_task = AbortOnDrop::new(tokio::spawn(read_pipe(child.stdout.take())));
+    let mut stderr_task = AbortOnDrop::new(tokio::spawn(read_pipe(child.stderr.take())));
+    let deadline = Instant::now() + duration;
+
+    let status = match timeout_at(deadline, child.wait()).await {
+        Ok(result) => result.map_err(OutputError::Io)?,
+        Err(_) => {
+            let _ = child.kill().await;
```
houko: This kills only the direct child (`child.kill()` → SIGKILL to that one PID); there's no `process_group`/`setsid` on the spawned `Command`, so any descendant the CLI itself forks (very plausible for `codex`/`gemini`/`qwen` — these wrap Node/sandbox subprocesses) is not part of a killable group here and can outlive the "timed out" request. The doc comment above (lines 72-74) claims "descendants that inherited stdout/stderr cannot keep the request alive after the direct child has exited" — that's true for the *pipe-reading* half (aborting the reader tasks does unblock the Rust-level read), but it doesn't mean the descendant process itself is terminated; it can keep running/using CPU/holding other resources after this function returns, which is the same class of leak PR #6960 sets out to close.

Given the task framing here (bound subprocess execution against resource exhaustion), consider `tokio::process::Command::process_group(0)` on Unix to put the child in its own group, then send the kill via `libc::killpg` (or track `child.id()` and call `kill(-pid, SIGKILL)`) instead of `child.kill()`, so orphaned grandchildren are actually reaped, not just detached from the read side.

---

crates/librefang-llm-drivers/src/cli_process.rs
```diff
@@ -0,0 +1,136 @@
...
+use tokio::io::AsyncReadExt;
+use tokio::process::Command;
+use tokio::task::JoinHandle;
+use tokio::time::{timeout_at, Instant};
+
+pub(crate) const DEFAULT_MESSAGE_TIMEOUT_SECS: u64 = 300;
+
+pub(crate) fn timeout_error(timeout_secs: u64, driver: &str) -> crate::llm_driver::LlmError {
+    crate::llm_driver::LlmError::TimedOut {
+        inactivity_secs: timeout_secs,
```
houko, line 30: This reuses `LlmError::TimedOut { inactivity_secs, .. }`, whose doc comment (`librefang-llm-driver/src/lib.rs:66`) and `Display` impl both say "Timed out after {inactivity_secs}s of **inactivity**" — but `deadline` here (`output_with_timeout`, line 83: `Instant::now() + duration`) is a fixed wall-clock cap from spawn that never resets on output, unlike `claude_code.rs`'s `ClaudeCodeDriver`, which the PR description says this change is "matching" ("let per-request `timeout_secs` override the driver default, matching Claude Code behavior"). `claude_code.rs` (around line 1329-1351) explicitly resets `last_output`/the deadline on every non-empty line and documents it as "The timer resets every time the CLI produces output" — a true inactivity timer that lets a long-but-actively-streaming response run indefinitely.

The new code (here, and the qwen-code `stream()` deadline added at qwen_code.rs:917) does not reset on activity, so it's a hard total-duration cap, not an inactivity timeout, despite reusing the "inactivity" wording end to end. Two consequences: (1) the PR's "matching Claude Code behavior" claim is inaccurate — a Qwen/Codex/Gemini CLI call that's actively producing output past `message_timeout_secs` (default 300s) gets killed the same as a truly hung one, where Claude Code would let it continue; (2) the resulting user/log-facing error ("Timed out after 300s of inactivity") is misleading when the process was never idle. Worth either resetting the deadline on activity (matching Claude Code) or renaming/re-wording this path's error to reflect a hard deadline rather than inactivity, so operators reading logs aren't misdirected into debugging a hang that wasn't one.

---

houko (commented):
Automated review of the CLI-subprocess-timeout bounding logic. The core mechanism (deadline via `timeout_at`, `kill_on_drop`, concurrent stdout/stderr draining, cache-key inclusion) checks out and compiles cleanly (`cargo check -p librefang-llm-drivers --lib`). Three findings below, all confirmed against the actual code/behavior rather than inferred: the two new regression tests fail deterministically on Linux (ETXTBSY, unrelated to the timeout logic they're meant to test), the kill only reaches the direct child with no process-group isolation for descendants, and the "inactivity" framing on the new error path doesn't match its actual hard-deadline semantics. No fix commits were needed from this pass — a changelog.d fragment gap was already closed by a prior pass (`a3e3867c`).

---

crates/librefang-llm-drivers/src/cli_process.rs
```diff
@@ -0,0 +1,250 @@
...
+
+use std::process::Output;
+use std::time::Duration;
+
+use tokio::io::AsyncReadExt;
+use tokio::process::Command;
+use tokio::task::JoinHandle;
+use tokio::time::{timeout_at, Instant};
+
+pub(crate) const DEFAULT_MESSAGE_TIMEOUT_SECS: u64 = 300;
```
houko, line 11: Judgment call for a maintainer, not a bug: is 300s the right default ceiling for every codex/gemini/qwen invocation?

These are agentic coding CLIs — a turn that involves multi-file edits, test runs, or a slow sandboxed tool call inside the CLI itself could plausibly exceed 5 minutes on a legitimately-progressing (not hung) request, and this default now hard-kills it rather than just warning. `with_message_timeout` makes it configurable per-driver and `request.timeout_secs` overrides per-call, so the knob exists — but worth confirming 300s is the intended out-of-the-box behavior for coding-agent CLIs specifically (vs. a shorter default that made sense for a simple chat completion) before this ships as the default for existing installs that never set it explicitly.

---

houko (commented):
Automated review pass. Pushed a follow-up fix commit (a6a8016) with two mechanical bugs found in this diff:

1. **Orphaned grandchild processes on timeout.** `output_with_timeout`'s kill path (and the qwen-code streaming path's own direct `child.kill()` calls) only killed the direct child pid. A CLI subprocess that backgrounds a helper process (Node child processes, MCP helper servers, …) leaked that grandchild running past the deadline — reproduced with a plain `sleep 30 &` grandchild surviving a 100ms-timeout kill of its parent. Fixed by spawning the child as the leader of its own process group (`process_group(0)`) and signalling the whole group (`kill(-pid, SIGKILL)` on Unix, `taskkill /T /F` on Windows) on timeout; added a regression test (`timeout_kills_the_whole_process_group_not_just_the_direct_child`) that fails without the fix.
2. **Deterministic test failure**: `complete_honors_request_timeout` (codex_cli.rs) and `stream_honors_request_timeout` (qwen_code.rs) built their `sleeping_cli()` fixture as a `NamedTempFile` and executed it while still holding the file open for writing — Linux refuses that exec with `ETXTBSY` ("Text file busy"), so both new tests fail 100% of the time as originally written (confirmed by running the unmodified `codex_cli.rs` test twice before touching it). Fixed by converting to `TempPath` (`into_temp_path()`) so the fd is closed before exec while the file stays on disk.

Verification: `cargo check -p librefang-llm-drivers --lib`, `cargo check -p librefang-runtime --lib` (the one in-tree dependent), `cargo clippy -p librefang-llm-drivers --all-targets -- -D warnings` (clean), `cargo test -p librefang-llm-drivers --lib` (610 passed, including both fixed tests re-run 3x for determinism). Full `--workspace --lib` isn't usable in this sandbox — `librefang-desktop` needs system `gdk-3.0`/`dbus-1` which aren't installed here — so I scoped to the touched crate and its dependent instead.

One item left as an inline comment for a maintainer call rather than fixed: whether 300s is the right default timeout ceiling for coding-agent CLI turns specifically.

---

crates/librefang-llm-drivers/src/cli_process.rs
houko, line 1: `timeout_error()` always builds `LlmError::TimedOut { inactivity_secs: timeout_secs, last_activity: format!("waiting for {driver} subprocess"), .. }`.

That variant's doc comment and `Display` impl (`crates/librefang-llm-driver/src/lib.rs`, around the `#[error("Timed out after {inactivity_secs}s of inactivity …")]` attribute) are written for a genuine inactivity watchdog — the existing `claude_code.rs` streaming path resets its clock on every token and only fires when the process actually goes silent (`"Claude CLI streaming timed out due to inactivity, killing process"`).

The four call sites this PR adds (`codex_cli.rs`, `gemini_cli.rs`, and both branches in `qwen_code.rs`, including the fixed `deadline` computed once before the `qwen_code.rs` streaming loop) are a hard wall-clock deadline on the *whole* subprocess call, not an inactivity timeout — a CLI that streams output continuously for the full duration still hits this exact same "due to inactivity" message once `message_timeout_secs` elapses.

For an operator triaging a timeout this is a real (if minor) misdiagnosis risk: "due to inactivity" points at a hang, when the actual cause here is "the request legitimately needed more than the configured budget."

Given `librefang-llm-driver`'s own convention against adding `LlmError` variants without discussion first, this reads like an intentional trade-off (reuse the existing variant/field shape) rather than an oversight, so flagging rather than changing it outright — worth either a distinct `last_activity` phrasing that doesn't say "inactivity" (e.g. "hard timeout after Ns" instead of routing through the inactivity-branded variant), or a one-line justification for reusing it as-is.

---

crates/librefang-llm-drivers/src/drivers/qwen_code.rs
houko, line 1: Non-blocking FYI, not asking for a change in this PR (scope is explicitly Codex/Gemini/Qwen per the PR title): `aider.rs`, `codewhale.rs`, and `vertex_ai.rs` still shell out via a plain `cmd.output().await` / `.output()` with no timeout at all, same unbounded-subprocess shape this PR fixes for Codex/Gemini/Qwen. Might be worth a follow-up PR reusing `cli_process::output_with_timeout` for those three, since the helper is now generic enough to drop in directly.

---

houko (commented):
Manual review of the timeout/subprocess-bounding logic (kill-on-timeout + process-group reaping, cache-key wiring, error propagation across the blocking/async boundary). Verified locally: all 9 timeout-related unit tests pass (`cargo test -p librefang-llm-drivers --lib -- timeout` and `-- cache_key`, including the process-group zombie/orphan-reaping test), `cargo clippy -p librefang-llm-drivers --lib --tests -- -D warnings` is clean, and `cargo fmt --all -- --check` is clean. Two non-blocking notes left inline below.

---

crates/librefang-llm-drivers/src/cli_process.rs
```diff
@@ -0,0 +1,266 @@
...
+        Ok(result) => result.map_err(OutputError::Io)?,
+        Err(_) => {
+            kill_on_timeout(&mut child).await;
+            stdout_task.abort();
+            stderr_task.abort();
+            return Err(OutputError::TimedOut);
+        }
+    };
+
+    let pipes = timeout_at(deadline, async {
```
houko, line 174: The pipe-drain phase reuses the same absolute `deadline` computed before `child.wait()` rather than starting a fresh window. If the child exits close to the deadline (e.g. at 299.9s of a 300s budget), this `timeout_at(deadline, ...)` can fire on its very first poll even though the process fully completed and only needs to flush already-buffered stdout/stderr. The caller then sees `OutputError::TimedOut` (and `kill_on_timeout` runs against an already-reaped child) for a request that actually succeeded, rather than getting its real output.

In practice the drain is usually fast once the child has exited, so this is a narrow race rather than a hang/leak — but it's a design call whether to (a) accept the small chance of a false-timeout near the edge, or (b) give the drain phase its own short grace window (e.g. a few seconds) independent of the caller-facing deadline. Worth a second look; not flagging as a hang/leak issue since the process itself is still killed and reaped correctly either way.

---

houko (commented):
Reviewed against CLAUDE.md conventions and the PR's stated goal (bound CLI subprocess execution, keep it configurable, surface a clear timeout error).

Pushed two mechanical fixes that extend this PR's own established pattern within the same crate:
- qwen-code streaming timeouts were dropping accumulated partial text instead of surfacing it like the claude-code streaming driver already does (#3552 pattern).
- CodeWhaleDriver had the identical unbounded `cmd.output().await` this PR fixes for codex/gemini/qwen, but wasn't touched — same crate/domain, so fixed here per CLAUDE.md's "fix what you found" rule rather than deferred. `aider.rs` has the same pattern but is dead code (never `mod`-declared, not compiled), so left alone.

One inline note on a possible edge case in the shared timeout helper — not a hang/leak, just worth a second look.

---

crates/librefang-llm-drivers/src/cli_process.rs
houko, line 1: This bounds subprocess execution for codex-cli, gemini-cli, and qwen-code, but two other CLI-backed drivers in this same crate still call the subprocess with no timeout at all.
`crates/librefang-llm-drivers/src/drivers/aider.rs:123` (`cmd.output().await`) and `crates/librefang-llm-drivers/src/drivers/codewhale.rs:226` (`cmd.output().await`) have the identical shape as the pre-fix codex/gemini/qwen call sites — a hung or malicious CLI process blocks the request (and its stdout/stderr reader) indefinitely.
Was leaving these two out of this pass intentional (e.g. lower priority, awaiting a separate PR), or should they get `output_with_timeout` in this change too since it's the same crate and the same class of bug this PR is fixing?

---

houko, line 1: `kill_on_timeout` here kills the whole process group on Unix specifically because these CLI wrapper binaries can background helper processes that a plain `child.kill()` never reaches (per the doc comment on `set_process_group` above).
`claude_code.rs` — the driver this timeout design is explicitly modeled on, per the PR description ("matching Claude Code behavior") — still only does `let _ = child.kill().await;` on all four of its own kill paths (`crates/librefang-llm-drivers/src/drivers/claude_code.rs:718,1051,1448,1541`), with no `process_group(0)` at spawn and no group-wide kill on timeout.
If the Claude CLI can fork the same kind of grandchild process described here, its timeout path can leak the same orphans this PR fixes for codex/gemini/qwen.
Worth a follow-up to bring `claude_code.rs` in line, or is there a reason it doesn't need the same protection?

---

houko (commented):
Reviewed the CLI subprocess timeout bounding in cli_process.rs and the codex/gemini/qwen driver wiring — the kill/drain/error/config-cache/test coverage all check out (process-group kill on Unix, taskkill /T on Windows, aborted pipe readers, TimedOut surfaced clearly, configurable default with per-request override, and tests that actually spawn a hanging subprocess and assert the kill lands within the bound). Left two scope questions inline about drivers/paths this treatment doesn't yet cover.

---

crates/librefang-llm-drivers/src/drivers/qwen_code.rs
```diff
@@ -872,11 +901,33 @@ impl QwenCodeDriver {
         // Set when a `tx.send(...)` fails — kill the child and stop reading
         // stdout so the CLI doesn't keep producing tokens for nobody (#3769).
         let mut receiver_dropped = false;
-
-        while let Ok(Some(line)) = lines.next_line().await {
+        let timeout_secs = request.timeout_secs.unwrap_or(self.message_timeout_secs);
+        let deadline = tokio::time::Instant::now() + std::time::Duration::from_secs(timeout_secs);
```
houko, line 910: `deadline` is computed once, before the read loop, and never reset — so this enforces a flat wall-clock budget for the whole streaming call, not an inactivity timeout.

That's a semantic change from what `message_timeout_secs` is documented to mean (`crates/librefang-llm-driver/src/lib.rs`: "Inactivity-based: the process is killed after this many seconds of silence on stdout, not wall-clock time") and from how `claude_code.rs`'s streaming path actually behaves (it resets `last_output`/`warned`/`notified` on every non-empty line, so a CLI that keeps producing output past `message_timeout_secs` total is never killed as long as it stays active).

With this change, a qwen-code turn that runs long but keeps streaming tokens the whole time (e.g. a large multi-step agentic response) will now get killed at `message_timeout_secs` even though it was never actually hung — previously it ran unbounded, so this is a new failure mode for legitimately slow-but-active turns, not just for genuinely stuck subprocesses.

Was the flat-deadline behavior intentional (simpler first cut, tightened later) or should this reset the deadline on each received line to match the doc comment and the claude-code driver's existing semantics?

---

houko (commented):
Solid fix overall — `output_with_timeout` correctly kills-and-reaps on timeout, and the follow-up commit (a6a8016) closing the process-group/grandchild-orphan gap plus the accompanying tests (`timeout_kills_the_whole_process_group_not_just_the_direct_child`) give good confidence the kill is real, not just an abandoned future. Config plumbing (`message_timeout_secs`, cache-key inclusion, changelog fragment) all check out.

One thing worth a second look before merge, left as an inline comment: `qwen_code.rs`'s new streaming-path timeout is a flat wall-clock deadline computed once, rather than the inactivity-reset timeout the field doc comment promises and `claude_code.rs` already implements — worth confirming that's the intended behavior for this PR vs. a follow-up.

Separately (not blocking, no diff line to anchor it to since the file isn't touched here): `claude_code.rs` — the driver this PR explicitly modeled its `timeout_secs` override on ("matching Claude Code behavior" per the PR description) — still only calls `child.kill()` on timeout in both `complete()` (~line 1051) and the streaming path (~line 1541), with no `process_group`/`kill_on_drop` and no abort of the spawned `stdout_handle`/`stderr_handle` background tasks. That's the exact grandchild-orphan gap this PR just fixed for codex/gemini/qwen/codewhale (per `cli_process.rs`'s own doc comment on `set_process_group`), left open in the one driver that's presumably used the most. Might be worth a fast follow-up PR applying `cli_process::output_with_timeout`/`kill_on_timeout` to `claude_code.rs` too.

---

houko (commented):
Automated review pass (repo-conventions check). The subprocess-timeout mechanism itself has been reviewed exhaustively by prior passes today (process-group kill confirmed real — verified `kill_on_timeout` sends `SIGKILL` to the process group via `libc::kill(-pid, ...)`, not just an abandoned future — plus the ETXTBSY test fix, CodeWhale coverage, partial-text surfacing, and the Spawn/Io error split are all present in the current tree), and the remaining open items (inactivity-vs-hard-deadline wording, the 300s default, the drain-phase race, `claude_code.rs` parity) are already tracked as inline comments from earlier passes — nothing new to add there.

One finding not yet flagged: two commits already pushed to this branch carry a Claude/Anthropic author identity, which this repo's own `CLAUDE.md` says must never happen (commit-msg hook + PreToolUse hook exist specifically to catch this):

```
bf62bad fix(llm): stop misattributing post-spawn subprocess failures to a missing CLI — Claude <noreply@anthropic.com>
8f70b07 test(llm): cover Gemini CLI's per-request subprocess timeout — Claude <noreply@anthropic.com>
```

(Contrast with the rest of the branch's automated commits, which used `LibreFang Review Bot <review-bot@users.noreply.github.com>` or a human name — those two are the only ones with the literal Claude/Anthropic identity.) This sandbox's global git identity is in fact hard-set to `Claude <noreply@anthropic.com>` with a Claude-branded commit-signing key, so it's not a one-off misconfiguration — any automated pass that commits here without overriding author identity will reproduce it. Given the branch already has multiple submitted reviews, fixing this means rewriting two non-tip-adjacent commits' authorship and force-pushing, which is outside what an automated pass should do unilaterally — flagging for a maintainer to reword the authorship (`git commit --amend`/interactive rebase + force-push) before merge, rather than doing it here.

No direct fixes pushed this pass — everything mechanical was already handled by earlier passes, and this finding requires a history rewrite decision, not a same-shape commit.

---

crates/librefang-llm-drivers/src/drivers/gemini_cli.rs
houko, line 1: The open thread on `codex_cli.rs:453` (ETXTBSY: `sleeping_cli()` allegedly leaves a writable fd open across `Command::spawn()`, failing `complete_honors_request_timeout` deterministically on Linux) names only `codex_cli.rs` and `qwen_code.rs`.
The identical `sleeping_cli()` helper and `complete_honors_request_timeout` test are also present here in `gemini_cli.rs` (same body: `NamedTempFile` + `write_all` + `set_mode(0o700)` + `into_temp_path()`) and in `codewhale.rs`.
Since all four copies are byte-for-byte the same construction, whatever root cause is confirmed for the codex/qwen instances applies equally here — the fix (e.g. `std::fs::write` into a `tempdir()`-scoped path instead of holding a `NamedTempFile` open) should be applied to all four call sites, not just the two currently named in that thread.

---

PR #6961, opened 2026-08-12
perf(api): isolate PDF extraction from async workers

houko:
## Summary
- run synchronous `pdf-extract` parsing on Tokio’s blocking pool instead of an async request worker
- cap active PDF parsers at two while excess requests wait asynchronously
- keep the concurrency permit inside the blocking closure so request cancellation cannot hide a still-running parser
- preserve existing extraction errors and attachment response behavior

## Testing
- `cargo test -p librefang-api pdf_job_does_not_block_async_runtime --lib`
- `cargo test -p librefang-api invalid_pdf_is_reported_from_blocking_worker --lib`
- `git diff --check`

The single-thread runtime regression would deadlock if the blocking job were executed inline.

### DIFF
--- a/crates/librefang-api/src/routes/agents/attachments.rs
+++ b/crates/librefang-api/src/routes/agents/attachments.rs
@@ -11,9 +11,8 @@ const TEXT_TRUNCATION_MARKER: &str =
 /// for the streaming download in `resolve_url_attachments`.
 const MAX_ATTACHMENT_BYTES: usize = 20 * 1024 * 1024;
 
-/// Keep CPU-heavy PDF parsing off Tokio workers and bound how many parser
-/// jobs can occupy the blocking pool at once. Additional requests wait
-/// asynchronously without consuming a runtime worker thread.
+/// Keep CPU-heavy PDF parsing off Tokio workers and bound how many parser jobs can occupy the blocking pool at once.
+/// Additional requests wait asynchronously without consuming a runtime worker thread.
 const MAX_CONCURRENT_PDF_EXTRACTIONS: usize = 2;
 static PDF_EXTRACTION_PERMITS: std::sync::LazyLock<std::sync::Arc<tokio::sync::Semaphore>> =
     std::sync::LazyLock::new(|| {
@@ -30,8 +29,7 @@ where
         .await
         .map_err(|_| "PDF extraction queue is unavailable".to_string())?;
     tokio::task::spawn_blocking(move || {
-        // Keep the permit in the blocking closure so request cancellation
-        // cannot release capacity while a detached parser is still running.
+        // Keep the permit in the blocking closure so request cancellation cannot release capacity while a detached parser is still running.
         let _permit = permit;
         job()
     })
@@ -529,8 +527,7 @@ mod tests {
             }
         }));
 
-        // A blocking closure executed inline would deadlock this single-thread
-        // runtime before the async task could release it.
+        // A blocking closure executed inline would deadlock this single-thread runtime before the async task could release it.
         tokio::task::yield_now().await;
         let (lock, ready) = &*gate;
         *lock.lock().unwrap() = true;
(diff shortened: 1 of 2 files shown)


crates/librefang-api/src/routes/agents/attachments.rs
```diff
@@ -11,6 +11,40 @@ const TEXT_TRUNCATION_MARKER: &str =
 /// for the streaming download in `resolve_url_attachments`.
 const MAX_ATTACHMENT_BYTES: usize = 20 * 1024 * 1024;
 
+/// Keep CPU-heavy PDF parsing off Tokio workers and bound how many parser
+/// jobs can occupy the blocking pool at once. Additional requests wait
+/// asynchronously without consuming a runtime worker thread.
+const MAX_CONCURRENT_PDF_EXTRACTIONS: usize = 2;
```
houko, line 16: The core off-thread mechanics look correct: `spawn_blocking` is used properly, the permit is moved into the blocking closure (so a cancelled request can't free capacity while the parser keeps running, matching the changelog note), and `extract_text_from_pdf` already `catch_unwind`s internally so a panic surfaces as a plain error string rather than an unhandled `JoinError` — the added `pdf_job_does_not_block_async_runtime` and `invalid_pdf_is_reported_from_blocking_worker` tests cover both of those directly.

One point worth a maintainer call rather than a mechanical fix: `MAX_CONCURRENT_PDF_EXTRACTIONS` is a hardcoded `2`, and there's no bound on how many requests can queue *waiting* for a permit — each queued request is already holding its full PDF byte buffer (up to `MAX_ATTACHMENT_BYTES` = 20 MiB) in memory via `tokio::fs::read` before it reaches `extract_pdf_text`. Under a burst of concurrent PDF uploads that's a queue of held buffers with no explicit cap, only implicitly bounded by whatever concurrency limit exists upstream (if any). Is `2` the right steady-state value, and is an unbounded wait queue acceptable here, or should this get a queue-depth cap / a `config.toml` knob like other tunables in this codebase?

---

crates/librefang-api/src/routes/agents/attachments.rs
```diff
@@ -11,6 +11,38 @@ const TEXT_TRUNCATION_MARKER: &str =
 /// for the streaming download in `resolve_url_attachments`.
 const MAX_ATTACHMENT_BYTES: usize = 20 * 1024 * 1024;
 
+/// Keep CPU-heavy PDF parsing off Tokio workers and bound how many parser jobs can occupy the blocking pool at once.
+/// Additional requests wait asynchronously without consuming a runtime worker thread.
+const MAX_CONCURRENT_PDF_EXTRACTIONS: usize = 2;
```
houko, line 16: The mechanism itself (spawn_blocking + permit held inside the blocking closure) looks correct — panic-safe, cancellation-safe, and errors propagate rather than getting unwrapped away. Two smaller points, not blocking:

1. `MAX_CONCURRENT_PDF_EXTRACTIONS` is a hardcoded `const` rather than a `config.toml` knob. Other concurrency caps in this codebase (`queue.concurrency.trigger_lane`, `max_concurrent_invocations`) are operator-configurable. Worth asking whether this one should be too, or whether `2` is intentionally fixed because it's bounding a specific CPU-bound library rather than a general workload — a maintainer call either way.
2. Neither new test actually exercises the cap of 2. `pdf_job_does_not_block_async_runtime` proves offloading happens; `invalid_pdf_is_reported_from_blocking_worker` proves error propagation. Nothing asserts that a 3rd concurrent job actually waits behind the semaphore rather than running unbounded — might be worth a test that spins up 3 blocked jobs and checks only 2 are inside the closure at once.

---

houko (commented):
Reviewed the off-thread PDF extraction change. The core mechanism (spawn_blocking, permit held inside the blocking closure, JoinError propagated instead of unwrapped) is correct and matches the PR description's claims — verified against the single call site in resolve_attachments and the existing pdf_text::extract_text_from_pdf error contract. Changelog fragment is present and correctly formatted. Left two non-blocking notes inline.

---

PR #6962, opened 2026-08-12
fix(wire): bound peer rate limiter state

houko:
## Summary
- cap active message and token rate-limit windows at 10,000 peer identities each
- sweep expired windows before rejecting a new identity at capacity
- fail closed when all retained windows are still active
- make sweep, capacity check, and insertion atomic so concurrent identity churn cannot exceed the bound

## Testing
- `cargo test -p librefang-wire peer_message_limiter_fails_closed_at_identity_capacity --lib`
- `cargo test -p librefang-wire peer_token_limiter_fails_closed_at_identity_capacity --lib`
- `cargo clippy -p librefang-wire --lib --tests -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/librefang-wire/src/peer.rs
+++ b/crates/librefang-wire/src/peer.rs
@@ -166,21 +166,26 @@ impl PeerRateLimiter {
     /// human-readable message if the peer has been rate-limited. The internal
     /// counters are updated on every call regardless of the outcome.
     pub fn check_message(&self, peer_id: &str) -> Result<(), String> {
-        if peer_id.len() > MAX_PEER_ID_LEN {
-            warn!(
-                peer_id_len = peer_id.len(),
-                limit = MAX_PEER_ID_LEN,
-                "OFP: peer_id exceeds maximum length; rejecting before rate-limiter admission"
-            );
-            return Err(format!(
-                "peer_id exceeds maximum length of {MAX_PEER_ID_LEN} bytes"
-            ));
-        }
-
         let now = Instant::now();
         let one_minute = Duration::from_secs(60);
 
         if self.max_msgs_per_minute > 0 {
+            // Mirrors `record_tokens`: the length cap only matters once we are
+            // about to touch `msg_counts`, so it lives inside this guard rather
+            // than ahead of it. Checking it unconditionally would reject
+            // oversized peer IDs even with message rate limiting disabled
+            // (`max_msgs_per_minute == 0`), where no map growth is possible.
+            if peer_id.len() > MAX_PEER_ID_LEN {
+                warn!(
+                    peer_id_len = peer_id.len(),
+                    limit = MAX_PEER_ID_LEN,
+                    "OFP: peer_id exceeds maximum length; rejecting before rate-limiter admission"
+                );
+                return Err(format!(
+                    "peer_id exceeds maximum length of {MAX_PEER_ID_LEN} bytes"
+                ));
+            }
+
             let mut counts = self
                 .msg_counts
                 .lock()
@@ -2445,6 +2450,20 @@ mod tests {
             0,
             "an oversized peer_id must never reach the map, or its length becomes a memory-amplification vector"
         );
+    }
+
+    #[test]
+    fn peer_message_limiter_skips_length_check_when_disabled() {
+        // With message rate limiting off (max_msgs_per_minute == 0),
+        // msg_counts is never touched, so an oversized peer_id poses no
+        // memory-amplification risk and must not be rejected. Mirrors
+        // record_tokens' behavior when max_tokens_per_hour is None.
+        let limiter = PeerRateLimiter::new(0, None);
+        let oversized = "x".repeat(MAX_PEER_ID_LEN + 1);
+
+        limiter.check_message(&oversized).unwrap();
+
+        assert_eq!(limiter.message_peer_count(), 0);
     }
 
     #[test]
--- a/crates/librefang-wire/src/peer.rs
+++ b/crates/librefang-wire/src/peer.rs
@@ -16,8 +16,9 @@ use async_trait::async_trait;
 use dashmap::DashMap;
 use hmac::{Hmac, Mac};
 use sha2::Sha256;
+use std::collections::HashMap;
 use std::net::SocketAddr;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant};
 use thiserror::Error;
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
@@ -25,6 +26,11 @@ use tokio::net::{TcpListener, TcpStream};
 use tracing::{debug, error, info, warn};
 
 type HmacSha256 = Hmac<Sha256>;
+
+/// Maximum distinct peer identities retained in each active rate-limit
+/// window. A shared secret holder can choose arbitrary node IDs, so this cap
+/// prevents identity churn from becoming an unbounded memory primitive.
+const MAX_RATE_LIMIT_PEERS: usize = 10_000;
 
 /// SECURITY: Time-windowed nonce tracker to prevent OFP handshake replay attacks.
 ///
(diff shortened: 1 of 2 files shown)


crates/librefang-wire/src/peer.rs
```diff
@@ -122,14 +127,13 @@ impl Default for NonceTracker {
...
-/// Tokio tasks that serve connections from the same peer node.
+/// Both counters are stored in a mutex-guarded `HashMap`, bounded at `MAX_RATE_LIMIT_PEERS` distinct peer identities, so they are shared safely across all Tokio tasks that serve connections from the same peer node.
 #[derive(Clone)]
 pub struct PeerRateLimiter {
     /// Tracks (message_count, window_start) per peer_id for rate limiting.
-    msg_counts: Arc<DashMap<String, (u32, Instant)>>,
+    msg_counts: Arc<Mutex<HashMap<String, (u32, Instant)>>>,
     /// Tracks (token_count, window_start) per peer_id for token budgeting.
-    token_counts: Arc<DashMap<String, (u64, Instant)>>,
+    token_counts: Arc<Mutex<HashMap<String, (u64, Instant)>>>,
```
houko, line 141: Logic itself checks out (retain-then-hard-cap under a single lock, so there's no TOCTOU between the capacity check and the insert — good). One design trade-off worth a maintainer call rather than a blocking fix:

Swapping the sharded `DashMap` for a single `Mutex<HashMap>` means every `check_message`/`record_tokens` call for *every* peer now serializes on one lock, where before, different peers could update their shards concurrently. The critical sections are short (no `.await` inside), so this is probably fine at realistic federation scale, but it's a real behavior change under concurrent multi-peer load that's worth confirming is acceptable rather than an oversight.

---

crates/librefang-wire/src/peer.rs
```diff
@@ -161,11 +165,22 @@ impl PeerRateLimiter {
...
-            let mut entry = self
+            let mut counts = self
                 .msg_counts
-                .entry(peer_id.to_string())
-                .or_insert((0, now));
-            let (count, window_start) = &mut *entry;
+                .lock()
+                .map_err(|_| "Peer message rate limiter lock is poisoned".to_string())?;
+            if !counts.contains_key(peer_id) && counts.len() >= MAX_RATE_LIMIT_PEERS {
+                counts.retain(|_, (_, started)| now.duration_since(*started) < one_minute);
```
houko, line 194: GC here only sweeps once the map is already **at** `MAX_RATE_LIMIT_PEERS`, whereas `NonceTracker::check_and_record` a few dozen lines above (same file) explicitly amortizes its sweep at an **80% threshold** specifically to avoid paying an O(n) `retain()` on every subsequent call once full ("Running `retain` on every insert is O(n) and lets any … client … force an expensive sweep on every connection attempt").

Here, once the map is full and an attacker with a valid shared secret keeps rotating to fresh identities faster than existing entries age out of their window, every such call re-runs a full O(n) sweep that finds nothing to evict, then rejects. `check_message`/`record_tokens` are post-auth (need the shared secret) so the exposure is narrower than `NonceTracker`'s pre-auth case, but it's the same attack shape the neighboring code was written to avoid — worth a call on whether the 80%-threshold pattern should be mirrored here too, or whether post-auth is a sufficient mitigation on its own.

---

crates/librefang-wire/src/peer.rs
```diff
@@ -26,6 +27,10 @@ use tracing::{debug, error, info, warn};
 
 type HmacSha256 = Hmac<Sha256>;
 
+/// Maximum distinct peer identities retained in each active rate-limit window.
```
houko, line 30: This bounds the *count* of distinct identities, but the map's actual memory footprint is `count × size(peer_id string)`, and I couldn't find anywhere in `librefang-wire` that bounds `node_id`/`peer_id` length independently — the only ceiling on the handshake frame that carries it is `MAX_PREHANDSHAKE_MESSAGE_SIZE` (64 KiB, shared with the UUID nonce, HMAC, keys, and the `RemoteAgentInfo` vec). A shared-secret holder could plausibly pick near-frame-sized node IDs and still stay under 10,000 distinct identities, landing well above the "few entries, cheap" mental model the cap implies (still finite, just larger than expected). Worth confirming whether `node_id` length is validated elsewhere at handshake time (didn't find it) or whether that's an accepted gap for this fix.

---

houko (commented):
Automated review pass on the peer rate-limiter bound. Core fix (retain-then-hard-cap under a single lock, no TOCTOU, correct window semantics preserved from the DashMap version) checks out; three design trade-offs left as inline comments for a maintainer call rather than autofixed, since each involves a judgment about acceptable risk/behavior rather than a clear-cut bug.

---

houko:
Automated CLAUDE.md-compliance pass on this PR.

Code: the identity-cap eviction logic is correct — retain-then-hard-cap under a single lock, exactly `MAX_RATE_LIMIT_PEERS` entries admitted, no off-by-one.
Confirmed by `peer_message_limiter_fails_closed_at_identity_capacity` / `peer_token_limiter_fails_closed_at_identity_capacity`.
No TOCTOU: the capacity check, sweep, and insert all happen while holding the same `Mutex` guard, so there's no race between "check limit" and "record request".
`HashMap` (vs `BTreeMap`) is fine here — this state never reaches an LLM prompt, so the deterministic-prompt-ordering rule in CLAUDE.md doesn't apply.
Ran `cargo check -p librefang-wire --lib`, `cargo clippy -p librefang-wire --all-targets -- -D warnings`, and `cargo test -p librefang-wire --lib` — all clean, including the 4 new tests added by this branch.
`changelog.d/security/6962-wire-rate-limiter-growth.md` is present, correctly sectioned under `security/`, and ends `(#6962) (@houko)`.

One compliance blocker found, not a code issue: commit `b5d3f6b` ("fix(wire): bound peer rate-limiter key size, not just entry count") has author and committer identity `Claude <noreply@anthropic.com>`.
CLAUDE.md's `commit-msg` hook is specifically documented to reject a commit whose author identity resolves to Claude/Anthropic even when the message text itself is clean, and that's exactly this case — the commit body contains no AI reference, only the identity does.
An automated review pass can't fix this without rewriting already-pushed history and force-pushing someone else's branch, which is out of scope for what this pass should do.
Flagging for a maintainer call: rebase to correct the authorship (e.g. `git commit --amend --reset-author` during an interactive rebase), or accept as-is.

---

crates/librefang-wire/src/peer.rs
```diff
@@ -161,11 +170,38 @@ impl PeerRateLimiter {
...
+            if peer_id.len() > MAX_PEER_ID_LEN {
+                warn!(
+                    peer_id_len = peer_id.len(),
+                    limit = MAX_PEER_ID_LEN,
+                    "OFP: peer_id exceeds maximum length; rejecting before rate-limiter admission"
+                );
+                return Err(format!(
+                    "peer_id exceeds maximum length of {MAX_PEER_ID_LEN} bytes"
+                ));
+            }
```
houko, line 187: `std::sync::Mutex` poisons on panic while held, unlike the `DashMap` this replaces (parking_lot-backed shards, no poisoning). Once poisoned, every future `check_message` / `record_tokens` call returns `Err` permanently for the life of the process — the limiter fails closed forever for every peer, not just the one that triggered the panic. The mitigation this PR adds (bounded identities, length cap) is good, but the switch to a poisoning `Mutex` introduces a new single-panic-to-permanent-outage path that a sharded `DashMap` didn't have.

In the current code the only realistic panic inside the critical section is a `u32`/`u64` counter overflow (`*count += 1`, `*total += tokens_used`), which needs billions of calls inside one window to trigger — low likelihood, but the blast radius (permanent, whole-process, both directions of traffic) seems disproportionate for what should be a soft "this peer misbehaved" signal.

Worth a design call: either recover from poisoning (`.unwrap_or_else(|e| e.into_inner())`) so one bad access doesn't wedge the limiter forever, or swap to `parking_lot::Mutex` (already a transitive dep here) which doesn't poison at all. Not blocking, but flagging since the PR's own goal is "doesn't itself become a new DoS vector."

---

houko (commented):
Reviewed the bound + eviction logic in detail: the atomic sweep-check-insert under one lock is correct, the 10k cap only evicts expired (not active) entries so it doesn't create a new DoS against live peers, and the oversized-`peer_id` guard genuinely closes the key-size amplification vector. Pushed one mechanical fix for a real inconsistency I found between the two limiter methods; left one design-judgment finding inline.

---

PR #6963, opened 2026-08-12
fix(deploy): require auth in Fly template

houko:
## Summary
- remove the unauthenticated public-bind default from the reusable Fly template
- generate a 256-bit API key and import it as a Fly secret before the one-command deployment’s first boot
- print the generated key once with an explicit save warning
- keep the official public demo exception explicit and scoped to its release CI job

## Testing
- `bash -n deploy/fly/deploy.sh`
- verified `deploy/fly/fly.toml` contains no `LIBREFANG_ALLOW_NO_AUTH` assignment
- verified the deploy script provisions `LIBREFANG_API_KEY` before `flyctl deploy`
- `git diff --check`

### DIFF
--- a/changelog.d/security/6963-fly-auth-default.md
+++ b/changelog.d/security/6963-fly-auth-default.md
@@ -1,0 +1,3 @@
+Require authentication by default in the reusable Fly.io deploy template.
+  `deploy/fly/fly.toml` previously shipped `LIBREFANG_ALLOW_NO_AUTH=1` unconditionally, so every deployment derived from the template inherited the official demo's intentionally open auth posture, not just the demo itself.
+  The one-command deploy script now generates a 256-bit `LIBREFANG_API_KEY` and imports it as a Fly secret before the app's first boot, and the official public demo's unauthenticated exception moved into its own release CI job rather than the shared template (#6963) (@houko)
--- a/deploy/fly/deploy.sh
+++ b/deploy/fly/deploy.sh
@@ -67,7 +67,11 @@ flyctl volumes create librefang_data \
   --size 1 \
   --yes
 
-# --- 6. Set secrets (optional) ---
+# --- 6. Set authentication and provider secrets ---
+LIBREFANG_ACCESS_KEY=$(openssl rand -hex 32)
+printf 'LIBREFANG_API_KEY=%s\n' "$LIBREFANG_ACCESS_KEY" \
+  | flyctl secrets import --app "$APP_NAME"
+
 PROVIDER_NAMES=(
   "OpenAI"
   "Anthropic"
@@ -194,7 +198,10 @@ ok "LibreFang is live!"
 echo ""
 echo "  Dashboard:  $APP_URL"
 echo "  API:        $APP_URL/api/health"
+echo "  API key:    $LIBREFANG_ACCESS_KEY"
 echo "  Manage:     flyctl dashboard --app $APP_NAME"
+echo ""
+warn "Save the API key now; it is stored as a Fly secret and cannot be read back."
 echo ""
 echo "  To add or change API keys later:"
 echo "    flyctl secrets set <PROVIDER>_API_KEY=your-key --app $APP_NAME"
--- a/deploy/fly/fly.toml
+++ b/deploy/fly/fly.toml
@@ -7,8 +7,8 @@ image = "ghcr.io/librefang/librefang:latest"
 [env]
   LIBREFANG_HOME = "/data"
   LIBREFANG_LISTEN = "0.0.0.0:4545"
-  # Authentication is required for this public bind. The one-command deploy
-  # script provisions LIBREFANG_API_KEY as a Fly secret before first boot.
+  # Authentication is required for this public bind.
+  # The one-command deploy script provisions LIBREFANG_API_KEY as a Fly secret before first boot.
 
 [http_service]
   internal_port = 4545
--- a/deploy/fly/fly.toml
+++ b/deploy/fly/fly.toml
@@ -7,8 +7,8 @@ image = "ghcr.io/librefang/librefang:latest"
 [env]
   LIBREFANG_HOME = "/data"
   LIBREFANG_LISTEN = "0.0.0.0:4545"
-  # Public live demo: intentionally open. Replace with LIBREFANG_API_KEY for private deployments.
-  LIBREFANG_ALLOW_NO_AUTH = "1"
+  # Authentication is required for this public bind. The one-command deploy
+  # script provisions LIBREFANG_API_KEY as a Fly secret before first boot.
 
 [http_service]
   internal_port = 4545
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3047,9 +3047,8 @@ jobs:
           ref: ${{ env.RELEASE_REF }}
       - uses: superfly/flyctl-actions/setup-flyctl@ed8efb33836e8b2096c7fd3ba1c8afe303ebbff1 # v1
       - name: Deploy public demo
-        # This job owns the intentionally unauthenticated official demo. Keep
-        # that exception out of fly.toml so user deployments remain secure by
-        # default and must explicitly provision authentication.
+        # This job owns the intentionally unauthenticated official demo.
+        # Keep that exception out of fly.toml so user deployments remain secure by default and must explicitly provision authentication.
         run: >-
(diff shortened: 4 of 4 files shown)


houko:
Automated review pass: commit `08714b0` ("docs(changelog): add security fragment for Fly auth-default fix") on this branch is authored as `Claude <noreply@anthropic.com>`.

---

deploy/fly/fly.toml
```diff
@@ -7,8 +7,8 @@ image = "ghcr.io/librefang/librefang:latest"
 [env]
   LIBREFANG_HOME = "/data"
   LIBREFANG_LISTEN = "0.0.0.0:4545"
-  # Public live demo: intentionally open. Replace with LIBREFANG_API_KEY for private deployments.
-  LIBREFANG_ALLOW_NO_AUTH = "1"
+  # Authentication is required for this public bind.
+  # The one-command deploy script provisions LIBREFANG_API_KEY as a Fly secret before first boot.
```
houko, line 11: The enforcement side of this looks solid — `middleware.rs` already fails closed for non-loopback requests when no `api_key`/`api_key_hash`/dashboard auth is configured, and only opens when `LIBREFANG_ALLOW_NO_AUTH=1` is explicitly set (which this template no longer sets). No bypass path left in the code.

What's missing is a migration story for deployments that already exist from the *old* template. Someone who deployed this Fly template before this PR has a running app with `LIBREFANG_ALLOW_NO_AUTH=1` baked into its Fly release config, and no `LIBREFANG_API_KEY` secret (that provisioning only happens inside `deploy.sh`, which existing deployments won't re-run). If they later pull this updated `fly.toml` and `flyctl deploy` it directly (a very plausible path for a "reusable template" used outside the one-command script), Fly replaces the app's `[env]` on the new release, `LIBREFANG_ALLOW_NO_AUTH` disappears, and the daemon now fails closed for *every* non-loopback request — including the operator's own dashboard/API access — with no key ever having been issued to them. That's a silent full lockout, not just "auth is now required."

This repo has precedent for calling out exactly this kind of gap explicitly when a security default changes (e.g. the `allowed_commands` migration note in the #6594 changelog entry) rather than leaving it implicit. Worth deciding here: do we want a callout in the changelog fragment / docs about redeploying an existing app (e.g. "run `flyctl secrets set LIBREFANG_API_KEY=$(openssl rand -hex 32)` before your next deploy"), or is silent lockout-on-redeploy considered acceptable given the alternative (silently staying open) is worse? Flagging for a maintainer call rather than editing, since the right fix is a product/docs decision, not a code change.

---

changelog.d/security/6963-fly-auth-default.md
houko, line 1: This fixes the reusable `deploy/fly/` template and `deploy.sh`, but the README's "Fly.io" deploy badge (marked "Recommended" on the deploy.librefang.ai page) points at a third, independent Fly deployment path that this PR doesn't touch: the Cloudflare Worker in `deploy/worker/src/index.js`.

That worker calls the Fly Machines API directly (`handleDeploy`), bypassing `fly.toml` and `deploy.sh` entirely. Its `env_vars` (lines ~90-93) set only `LIBREFANG_HOME` and `OPENROUTER_API_KEY` — no `LIBREFANG_API_KEY`, no `LIBREFANG_ALLOW_NO_AUTH`. Combined with `deploy/docker-entrypoint.sh`, which unconditionally rewrites any loopback `api_listen` to `0.0.0.0` for any containerized boot, every machine this worker creates binds non-loopback with zero authentication configured and no opt-in flag.

Given `evaluate_bind_auth_safety` / `check_bind_auth_safety` in `crates/librefang-api/src/server.rs` (the fail-closed check from #3572), that combination doesn't fail open — it refuses to boot, so users going through the "Recommended" one-click Fly button most likely get a crash-looping machine they can never reach, rather than an insecure-but-working one. Either way it's a gap in the "every Fly deployment defaults to secure/working" story this PR is establishing, and it's worth a follow-up (or at least a note in the PR description) since it's the primary path the README steers people toward. Flagging for awareness rather than requesting a change here, since fixing it means touching `deploy/worker/`, outside this PR's file set.

---

houko (commented):
Reviewed the four changed files against CLAUDE.md (changelog fragment format, prose-wrapping, auth-default posture). The template/script/CI changes themselves are correct: `deploy.sh` provisions a 256-bit `LIBREFANG_API_KEY` via `openssl rand -hex 32` and imports it as a Fly secret before `flyctl deploy` (so before first boot), `fly.toml` no longer sets `LIBREFANG_ALLOW_NO_AUTH`, and the demo's opt-in is correctly scoped to the release CI job. The changelog fragment is correctly placed under `security/`, formatted per `changelog.d/README.md` (sentence-per-line, 2-space continuation indent, ending `(#6963) (@houko)`), with no AI attribution. Left one comment flagging an adjacent gap this PR doesn't cover.

---

deploy/fly/fly.toml
houko, line 11: This closes the gap for the two paths this PR touches (`deploy/fly/deploy.sh` and the release CI job), but there is a third, independent fresh-deploy path for the exact same target: the Cloudflare-Worker-backed "Deploy to Fly" button (`web/src/pages/DeployPage.tsx` → `deploy/worker/src/index.js`).

That worker creates the Fly app/machine directly through the Machines API (`POST /apps/{app}/machines`) and its `env_vars` only sets `LIBREFANG_HOME` and `OPENROUTER_API_KEY` — no `LIBREFANG_API_KEY`, no `LIBREFANG_ALLOW_NO_AUTH`. `deploy/docker-entrypoint.sh` unconditionally force-rewrites `api_listen` from `127.0.0.1:...` to `0.0.0.0:...` on every boot ("In Docker, 127.0.0.1 is the container's own loopback and is unreachable from the host"), so that machine also ends up on a non-loopback bind with zero auth configured, same as the `fly.toml` case this PR fixes.

`check_bind_auth_safety` in `crates/librefang-api/src/server.rs` will refuse to boot that machine outright (fail-closed, since it predates this PR) — so today this deploy path is either already broken or, if it isn't, I can't find where it provisions auth. Either way it seems out of scope of "require auth in Fly template" only because it doesn't go through the template at all, and it's worth confirming explicitly whether it needs the same treatment before calling the Fly auth gap closed.

---

deploy/fly/deploy.sh
```diff
@@ -67,7 +67,11 @@ flyctl volumes create librefang_data \
   --size 1 \
   --yes
 
-# --- 6. Set secrets (optional) ---
+# --- 6. Set authentication and provider secrets ---
+LIBREFANG_ACCESS_KEY=$(openssl rand -hex 32)
+printf 'LIBREFANG_API_KEY=%s\n' "$LIBREFANG_ACCESS_KEY" \
+  | flyctl secrets import --app "$APP_NAME"
```
houko, line 73: Confirmed this genuinely enforces auth rather than just documenting it: the key is imported as a Fly secret before the first `flyctl deploy` at the bottom of this script, and `check_bind_auth_safety` (`crates/librefang-api/src/server.rs`) refuses to boot on a non-loopback bind with no auth configured and no `LIBREFANG_ALLOW_NO_AUTH` opt-in — so a fresh deploy from this script cannot end up open by omission.

One operator-facing consequence worth calling out explicitly (in the PR body / release notes, not necessarily the code): anyone who already deployed from the old template — which shipped `LIBREFANG_ALLOW_NO_AUTH=1` unconditionally in `fly.toml` — and later re-pulls this repo and reruns `flyctl deploy --config deploy/fly/fly.toml` on their *existing* app without having separately provisioned any auth will lose that env var on the next deploy and hit the same boot refusal, turning a previously-running (if insecure) deployment into a boot failure rather than a smooth "now it's secure" transition. That's the correct fail-closed behavior, but it's a breaking change for redeploys of pre-existing apps that this PR's testing notes don't mention.

---

houko (commented):
Reviewed the Fly auth-default fix. The core mechanism checks out: removing `LIBREFANG_ALLOW_NO_AUTH=1` from the shared `fly.toml` combined with provisioning `LIBREFANG_API_KEY` before first boot genuinely forces auth via `check_bind_auth_safety`'s fail-closed refusal-to-boot, not just a doc change. No CLAUDE.md-mechanical issues found in this diff (changelog fragment format, section, and prose wrapping all check out; no AI attribution; conventional-commit-shaped). Left two inline notes on security-relevant scope/compat questions that need a maintainer call rather than a mechanical fix.

---

PR #6964, opened 2026-08-12
fix(deploy): restrict GCP network exposure

houko:
## Summary
- require a restricted source CIDR for both SSH and the LibreFang API firewall rules
- reject universal IPv4 and IPv6 source ranges during Terraform planning
- require a 32-character-or-longer LibreFang bearer key and pass it to cloud-init
- replace the stale `LIBREFANG_BIND` variable with the supported `LIBREFANG_LISTEN` variable
- document authenticated health checks and key/CIDR generation

## Testing
- verified cloud-init template placeholders exactly match the variables passed by `templatefile`
- verified no GCP firewall rule retains `0.0.0.0/0` and no `LIBREFANG_BIND` reference remains
- `git diff --check`

Terraform CLI was not installed locally, so `terraform validate` could not be run.

### DIFF
--- a/changelog.d/security/6964-gcp-deploy-network-exposure.md
+++ b/changelog.d/security/6964-gcp-deploy-network-exposure.md
@@ -1,0 +1,2 @@
+The GCP Terraform deploy opened SSH and the LibreFang dashboard/API firewall rules to `0.0.0.0/0`, and cloud-init still set the stale `LIBREFANG_BIND` variable instead of the supported `LIBREFANG_LISTEN`, leaving the public listener with no bearer key configured.
+Both firewall rules now require an operator-supplied `allowed_source_cidr`, with `0.0.0.0/0` and `::/0` rejected at plan time, and a required 32-character-minimum `LIBREFANG_API_KEY` is generated and wired through cloud-init so the API enforces bearer authentication (#6964) (@houko)
--- a/deploy/gcp/README.md
+++ b/deploy/gcp/README.md
@@ -19,7 +19,10 @@ gcloud auth application-default login
 
 # Configure
 cp terraform.tfvars.example terraform.tfvars
-# Edit terraform.tfvars with your project_id and API keys
+# Edit terraform.tfvars with your project_id, a restricted source CIDR,
+# a generated LibreFang API key, and provider keys
+# Current public IPv4: curl -4 https://ifconfig.me
+# API key:            openssl rand -hex 32
 
 # Deploy (~2 minutes)
 terraform init
@@ -34,7 +37,8 @@ Terraform will output:
 
 ```bash
 # Wait ~1 minute for cloud-init to finish, then:
-curl http://<external_ip>:4545/api/health
+curl -H "Authorization: Bearer <librefang_api_key>" \
+  http://<external_ip>:4545/api/health
 ```
 
 ## Teardown
@@ -57,7 +61,8 @@ terraform destroy
 │  │   :4545 ← dashboard/API  │  │
 │  └───────────────────────────┘  │
 │                                 │
-│  Firewall: SSH(22) + HTTP(4545) │
+│  Firewall: SSH(22) + API(4545)  │
+│  restricted to your source CIDR │
 └─────────────────────────────────┘
 ```
 
--- a/deploy/gcp/terraform.tfvars.example
+++ b/deploy/gcp/terraform.tfvars.example
@@ -1,6 +1,13 @@
 # Required
 project_id = "my-gcp-project-id"
 
+# Restrict both SSH and the dashboard/API to your current public IPv4 address.
+# Find it with: curl -4 https://ifconfig.me
+allowed_source_cidr = "203.0.113.10/32"
+
+# Generate with: openssl rand -hex 32
+librefang_api_key = "replace-with-a-random-64-character-hex-key"
+
 # Optional (defaults shown)
 # region           = "us-central1"
 # zone             = "us-central1-a"
--- a/deploy/gcp/variables.tf
+++ b/deploy/gcp/variables.tf
@@ -21,6 +21,29 @@ variable "ssh_pub_key_path" {
   default     = "~/.ssh/id_rsa.pub"
 }
 
+variable "allowed_source_cidr" {
+  description = "Trusted IPv4 CIDR allowed to reach SSH and LibreFang (for example, 203.0.113.10/32)"
+  type        = string
+
+  validation {
+    condition = can(cidrhost(var.allowed_source_cidr, 0)) &&
+      var.allowed_source_cidr != "0.0.0.0/0" &&
+      var.allowed_source_cidr != "::/0"
+    error_message = "allowed_source_cidr must be a valid restricted CIDR; public 0.0.0.0/0 and ::/0 are forbidden."
+  }
+}
+
+variable "librefang_api_key" {
+  description = "Bearer key required to access the public LibreFang listener"
+  type        = string
+  sensitive   = true
+
+  validation {
+    condition     = length(var.librefang_api_key) >= 32
+    error_message = "librefang_api_key must contain at least 32 characters."
+  }
+}
+
 variable "librefang_version" {
   description = "LibreFang release tag (e.g. v0.4.2-20260314), or 'latest'"
   type        = string
--- a/deploy/gcp/variables.tf
+++ b/deploy/gcp/variables.tf
@@ -26,7 +26,7 @@ variable "allowed_source_cidr" {
   type        = string
 
   validation {
-    condition = can(cidrhost(var.allowed_source_cidr, 0)) &&
+    condition     = can(cidrhost(var.allowed_source_cidr, 0)) &&
       var.allowed_source_cidr != "0.0.0.0/0" &&
       var.allowed_source_cidr != "::/0"
     error_message = "allowed_source_cidr must be a valid restricted CIDR; public 0.0.0.0/0 and ::/0 are forbidden."
--- a/deploy/gcp/cloud-init.yml.tpl
+++ b/deploy/gcp/cloud-init.yml.tpl
@@ -25,7 +25,8 @@ write_files:
       Type=simple
       User=librefang
       Environment=LIBREFANG_HOME=/data
-      Environment=LIBREFANG_BIND=0.0.0.0:4545
+      Environment=LIBREFANG_LISTEN=0.0.0.0:4545
+      Environment=LIBREFANG_API_KEY=${librefang_api_key}
       Environment=GROQ_API_KEY=${groq_api_key}
       Environment=OPENAI_API_KEY=${openai_api_key}
       Environment=ANTHROPIC_API_KEY=${anthropic_api_key}
--- a/deploy/gcp/main.tf
+++ b/deploy/gcp/main.tf
@@ -38,7 +38,7 @@ resource "google_compute_firewall" "allow_ssh" {
     ports    = ["22"]
   }
 
-  source_ranges = ["0.0.0.0/0"]
+  source_ranges = [var.allowed_source_cidr]
   target_tags   = ["librefang"]
 }
 
@@ -51,7 +51,7 @@ resource "google_compute_firewall" "allow_http" {
     ports    = ["4545"]
   }
 
-  source_ranges = ["0.0.0.0/0"]
+  source_ranges = [var.allowed_source_cidr]
   target_tags   = ["librefang"]
 }
 
@@ -79,6 +79,7 @@ resource "google_compute_instance" "librefang" {
     ssh-keys  = "librefang:${file(pathexpand(var.ssh_pub_key_path))}"
     user-data = templatefile("${path.module}/cloud-init.yml.tpl", {
       librefang_version = var.librefang_version
+      librefang_api_key = var.librefang_api_key
       groq_api_key      = var.groq_api_key
       openai_api_key    = var.openai_api_key
       anthropic_api_key = var.anthropic_api_key
--- a/docs/src/app/getting-started/page.mdx
+++ b/docs/src/app/getting-started/page.mdx
@@ -374,7 +374,8 @@ Deploy on GCP always-free tier (e2-micro). See [deploy/gcp/README.md](https://gi
 ```bash
 cd deploy/gcp
 cp terraform.tfvars.example terraform.tfvars
-# Edit terraform.tfvars with your project_id and API keys
+# Edit terraform.tfvars with your project_id, a restricted source CIDR,
+# a generated LibreFang API key, and provider keys
 terraform init && terraform apply
 ```
 
--- a/docs/src/app/zh/getting-started/page.mdx
+++ b/docs/src/app/zh/getting-started/page.mdx
@@ -373,7 +373,8 @@ curl -sL https://raw.githubusercontent.com/librefang/librefang/main/deploy/fly/d
 ```bash
 cd deploy/gcp
 cp terraform.tfvars.example terraform.tfvars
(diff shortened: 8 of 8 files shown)


houko:
Automated review pass: commit `4e9fd5b` ("docs(changelog): add security fragment for GCP deploy hardening") on this branch is authored as `Claude <noreply@anthropic.com>`.

---

deploy/gcp/terraform.tfvars.example
```diff
@@ -1,6 +1,13 @@
 # Required
 project_id = "my-gcp-project-id"
 
+# Restrict both SSH and the dashboard/API to your current public IPv4 address.
+# Find it with: curl -4 https://ifconfig.me
+allowed_source_cidr = "203.0.113.10/32"
+
+# Generate with: openssl rand -hex 32
+librefang_api_key = "replace-with-a-random-64-character-hex-key"
```
houko, line 9: This example value (`replace-with-a-random-64-character-hex-key`, 42 chars) satisfies the `variable "librefang_api_key"` validation (`length >= 32`) as-is.
If an operator copies `terraform.tfvars.example` to `terraform.tfvars` and misses this one line, `terraform apply` will not error — it will deploy with a bearer key that is committed in the public repo, i.e. no real protection at all, while looking configured.
Same shape of issue the rest of this PR is fixing for the CIDR (`0.0.0.0/0`/`::/0` are explicitly rejected at plan time) — worth the same treatment here, e.g. reject this literal placeholder string in the validation block, or require a hex-only pattern that the placeholder text itself can't satisfy.
Flagging for a human call since the "right" validation shape is a design choice, not a mechanical fix.

---

changelog.d/security/6964-gcp-deploy-network-exposure.md
```diff
@@ -0,0 +1,2 @@
+The GCP Terraform deploy opened SSH and the LibreFang dashboard/API firewall rules to `0.0.0.0/0`, and cloud-init still set the stale `LIBREFANG_BIND` variable instead of the supported `LIBREFANG_LISTEN`, leaving the public listener with no bearer key configured.
+Both firewall rules now require an operator-supplied `allowed_source_cidr`, with `0.0.0.0/0` and `::/0` rejected at plan time, and a required 32-character-minimum `LIBREFANG_API_KEY` is generated and wired through cloud-init so the API enforces bearer authentication (#6964) (@houko)
```
houko, line 2: This file was added in commit `4e9fd5b`, whose author identity is `Claude <noreply@anthropic.com>`.
CLAUDE.md's "No AI / Claude attribution" rule and the repo's own `commit-msg` hook explicitly cover this exact case — a clean commit *message* with an AI author *identity* still trips the check (`git var GIT_AUTHOR_IDENT`).
Fixing this requires rewriting this branch's history (reset the author + amend, or squash) and force-pushing, which is outside what an automated reviewer should do unilaterally on someone else's branch — flagging for @houko to rebase/reauthor before merge rather than fixing it myself.

---

houko (commented):
Automated security review of the GCP deploy hardening. The core fix looks correct (firewall CIDRs, LIBREFANG_LISTEN, required bearer key) — two secondary findings left as inline comments for a human call, no mechanical fixes applied.

---

changelog.d/security/6964-gcp-deploy-network-exposure.md
houko, line 1: CLAUDE.md compliance: commit `4e9fd5b` on this branch ("docs(changelog): add security fragment for GCP deploy hardening", which added this file) is authored *and* committed by `Claude <noreply@anthropic.com>`.
Repo policy is "No Claude / Anthropic / AI attribution in commit messages" and the `commit-msg` hook is explicitly documented to reject a commit whose author identity resolves to Claude/Anthropic even when the message text is clean — so this commit should not have been pushed as-is (either hooks weren't active, or `--no-verify` was used).
The fragment's *content* is fine (correct section, correct `(#6964) (@houko)` trailer, no attribution in the text), so this is purely a commit-metadata issue.
I'm flagging rather than fixing directly: correcting authorship on an already-pushed commit needs a rebase + force-push, which is outside what I'm allowed to do here. A maintainer will need to either amend the author before merge or rely on squash-merge (which drops per-commit authorship from `main`).

---

deploy/gcp/variables.tf
```diff
@@ -21,6 +21,29 @@ variable "ssh_pub_key_path" {
...
 }
 
+variable "allowed_source_cidr" {
+  description = "Trusted IPv4 CIDR allowed to reach SSH and LibreFang (for example, 203.0.113.10/32)"
+  type        = string
+
+  validation {
+    condition = can(cidrhost(var.allowed_source_cidr, 0)) &&
+      var.allowed_source_cidr != "0.0.0.0/0" &&
+      var.allowed_source_cidr != "::/0"
```
houko: The validation only rejects the two literal strings `"0.0.0.0/0"` and `"::/0"`, not other near-universal ranges. `allowed_source_cidr = "0.0.0.0/1"` (or `"128.0.0.0/1"`) is a valid CIDR per `cidrhost()`, isn't equal to either forbidden literal, and so passes `terraform plan` while still opening SSH and the dashboard/API to roughly half of all IPv4 space — which defeats the stated goal in the changelog/PR body ("reject universal IPv4 and IPv6 source ranges") and in this variable's own `error_message` ("public 0.0.0.0/0 and ::/0 are forbidden").

Suggested fix — bound the prefix length instead of/in addition to the literal-string check, e.g.:
```hcl
condition = can(cidrhost(var.allowed_source_cidr, 0)) && (
  (strcontains(var.allowed_source_cidr, ".") && tonumber(split("/", var.allowed_source_cidr)[1]) >= 24) ||
  (strcontains(var.allowed_source_cidr, ":") && tonumber(split("/", var.allowed_source_cidr)[1]) >= 64)
)
```
(thresholds illustrative — pick whatever the deploy actually intends to support, e.g. requiring a /32 host route since this doc only ever tells operators to paste their single public IP).

---

houko (commented):
Automated review pass (comment-only). One finding posted inline; otherwise the fix is sound — both firewall rules are consistently parameterized (no sibling left open), `LIBREFANG_LISTEN`/`LIBREFANG_API_KEY` verified against the actual codebase (not stale names), `/api/health` stays reachable (it's on the public allowlist regardless of the bearer key), and the changelog.d fragment is correctly formed and attributed.

---

deploy/gcp/variables.tf
```diff
@@ -21,6 +21,29 @@ variable "ssh_pub_key_path" {
...
+    error_message = "allowed_source_cidr must be a valid restricted CIDR; public 0.0.0.0/0 and ::/0 are forbidden."
+  }
```
houko, line 33: The `allowed_source_cidr` validation only rejects the exact literal strings `"0.0.0.0/0"` and `"::/0"`. It does not reject other still-very-broad prefixes an operator could paste in by mistake — e.g. `0.0.0.0/1`, `0.0.0.0/8`, or any `/1`–`/7` block, all of which are `can(cidrhost(...))`-valid and not caught here, and would still open SSH + the API bearer-auth port to a large swath of the internet while looking "restricted" in `terraform plan` output.

Confirmed against `main.tf:41,54` — both `allow_ssh` and `allow_http` source directly from this one variable, so a gap here affects both rules.

Given this repo's own guidance to treat GCP network-exposure judgment calls carefully rather than have an unattended pass "fix" them, flagging rather than changing: is an exact-match denylist the intended bar (relying on the README's `/32` guidance + operator judgment), or should the validation also reject prefixes shorter than some floor (e.g. `/24` or `/28`) to make an accidental broad paste fail at `terraform plan` instead of silently applying?

---

houko (commented):
Reviewed the GCP deploy security fix. The core fix is solid and closes real gaps confirmed against the current codebase:
- `LIBREFANG_BIND` is not read anywhere by the binary (confirmed via repo-wide grep); the daemon was silently falling back to the loopback default (`DEFAULT_API_LISTEN = "127.0.0.1:4545"`, `librefang-types/src/config/mod.rs:25`) despite the firewall rule opening 4545 to `0.0.0.0/0`. Switching to `LIBREFANG_LISTEN` (the var `Kernel::boot_with_config` actually reads, per `boot.rs:47-49`) makes the service reachable as originally intended, and pairing that with a mandatory `LIBREFANG_API_KEY` (matches `middleware.rs`'s `API_KEY_ENV` constant) and a restricted `allowed_source_cidr` closes the resulting exposure correctly.
- No GCP-managed health check / load balancer exists in this config (bare `google_compute_instance` with an ephemeral public IP), so restricting `source_ranges` on both firewall rules to the operator's CIDR cannot lock out an automated health check — there isn't one.
- No hardcoded secrets: `librefang_api_key` is `sensitive = true`, and the `.tfvars.example` placeholder is an obvious non-secret string.
- Changelog fragment is present and correctly formatted (`changelog.d/security/6964-gcp-deploy-network-exposure.md`).

Left one inline comment on `variables.tf` about the `allowed_source_cidr` validation only denylisting the two literal universal strings rather than any floor on prefix length — a real but judgment-call-shaped gap, flagging per this repo's network-exposure review guidance rather than changing it unilaterally.

---

deploy/gcp/variables.tf
```diff
@@ -21,6 +21,29 @@ variable "ssh_pub_key_path" {
...
+variable "allowed_source_cidr" {
+  description = "Trusted IPv4 CIDR allowed to reach SSH and LibreFang (for example, 203.0.113.10/32)"
+  type        = string
+
+  validation {
+    condition     = can(cidrhost(var.allowed_source_cidr, 0)) &&
+      var.allowed_source_cidr != "0.0.0.0/0" &&
+      var.allowed_source_cidr != "::/0"
+    error_message = "allowed_source_cidr must be a valid restricted CIDR; public 0.0.0.0/0 and ::/0 are forbidden."
+  }
```
houko, line 33: The PR description says this rejects "universal IPv4 and IPv6 source ranges," but the validation only excludes the two exact literal strings `"0.0.0.0/0"` and `"::/0"`.
A CIDR that is effectively just as open — `0.0.0.0/1`, `128.0.0.0/1`, or any other very broad prefix — still passes `can(cidrhost(...))` and the two equality checks, so an operator (or a copy-pasted/typo'd value) can still open SSH and the API to most of the internet while `terraform plan` reports success.
Worth deciding deliberately: either widen the guard to reject prefixes below some minimum length (e.g. `tonumber(split("/", var.allowed_source_cidr)[1]) >= 24` for IPv4, with a parallel IPv6 minimum), or explicitly document that this check only catches the literal `/0` case and operators are still responsible for picking a genuinely narrow CIDR. Leaving it as-is without that decision is a real gap in a rule whose entire job is restricting network exposure.

---

changelog.d/security/6964-gcp-deploy-network-exposure.md
houko, line 1: Following up on the earlier automated-review comment about commit `4e9fd5b` being authored as `Claude <noreply@anthropic.com>`: commit `a4e3dca` ("docs(getting-started): sync GCP terraform.tfvars guidance with PR #6964") on this branch has the same author identity and was not called out yet.
CLAUDE.md forbids Claude/Anthropic attribution in commits, and the `commit-msg` hook is meant to reject this exact author identity — it evidently wasn't active when these were made.
I can't rewrite already-pushed commit history without a force-push, which is outside what I'm allowed to do here, so flagging both occurrences for a maintainer to squash/rebase with a clean author identity before this lands on `main`.

---

houko (commented):
Independent review pass: one mechanical formatting fix pushed, two ambiguous security/policy findings left as inline comments for a maintainer decision — see below.

---

deploy/gcp/variables.tf
```diff
@@ -21,6 +21,29 @@ variable "ssh_pub_key_path" {
...
 
+variable "allowed_source_cidr" {
+  description = "Trusted IPv4 CIDR allowed to reach SSH and LibreFang (for example, 203.0.113.10/32)"
+  type        = string
+
+  validation {
+    condition     = can(cidrhost(var.allowed_source_cidr, 0)) &&
+      var.allowed_source_cidr != "0.0.0.0/0" &&
+      var.allowed_source_cidr != "::/0"
+    error_message = "allowed_source_cidr must be a valid restricted CIDR; public 0.0.0.0/0 and ::/0 are forbidden."
```
houko, line 32: The validation only rejects the two literal strings `0.0.0.0/0` and `::/0`.
It does not reject other still-effectively-open ranges that pass `cidrhost()`, e.g. `0.0.0.0/1` or `128.0.0.0/1` (each covers half of all IPv4 addresses), or a typo'd short prefix like `1.2.3.4/1`.
Since the whole point of this PR is closing off public exposure, worth deciding on a minimum acceptable prefix length (e.g. require `/24` or narrower for IPv4, and a matching bound for IPv6) rather than only excluding the two most literal "everything" values.
Happy to defer to a maintainer's judgment on where to draw that line — flagging for visibility rather than proposing a specific number.

---

houko (commented):
Reviewed the GCP deploy security fix. Firewall CIDR restriction, LIBREFANG_LISTEN correction, and the required API key wiring all check out against the kernel source and the PR's own description. One flagged item inline for maintainer judgment.

---

deploy/gcp/variables.tf
houko, line 32: The validation blocklists only the literal strings `0.0.0.0/0` and `::/0`, not near-universal prefixes.
`allowed_source_cidr = "0.0.0.0/1"` or `"128.0.0.0/1"` would still pass `cidrhost()` and the two equality checks while opening roughly half of all IPv4 space to SSH and the API — essentially the same exposure this PR is trying to close.
Worth enforcing a minimum prefix length (e.g. reject anything broader than `/24` for IPv4, `/64` for IPv6) rather than (or in addition to) blocklisting two exact strings, since the goal stated in the PR description is "reject universal IPv4 and IPv6 source ranges" and a prefix-length check catches the whole class instead of just the two literal spellings.

---

deploy/gcp/README.md
```diff
@@ -34,7 +37,8 @@ Terraform will output:
 
 ```bash
 # Wait ~1 minute for cloud-init to finish, then:
-curl http://<external_ip>:4545/api/health
+curl -H "Authorization: Bearer <librefang_api_key>" \
+  http://<external_ip>:4545/api/health
```
houko, line 41: `/api/health` is intentionally in the always-public allowlist in `crates/librefang-api/src/middleware.rs` (`PublicRoute::exact_any("/api/health")`, with an explicit comment that it stays public regardless of API-key configuration, precisely so load balancers / readiness probes can reach it without a token).
So this example works, but it doesn't actually demonstrate authenticated access — it may give operators the false impression that `/api/health` requires the bearer key, or that they've confirmed auth is enforced when they haven't.
If the goal is to document an authenticated health check, `/api/health/detail` is the one that's actually auth-gated (per the same file's comments) and would be a more accurate example to use here.

---

houko (commented):
Reviewed against CLAUDE.md's deploy/security guidance. The core fix (restricted CIDR + rejecting universal ranges, required 32+ char API key, LIBREFANG_BIND -> LIBREFANG_LISTEN, changelog.d fragment) is sound and verified against the actual env vars the runtime reads (`LIBREFANG_LISTEN` in `kernel/boot.rs`, `LIBREFANG_API_KEY` in `server.rs`/`middleware.rs`). Two non-blocking findings left inline: the CIDR validation only blocklists two exact strings rather than enforcing a minimum prefix length, and the README's "authenticated health check" example targets the always-public `/api/health` rather than the auth-gated `/api/health/detail`. Leaving as comments rather than pushing a fix given this touches deploy/network security config directly.

---

Issue #3541, opened 2026-04-27
KernelHandle trait uses Result<_, String> for ~40 methods — destroys structured errors


houko:
## Summary
`KernelHandle` trait has ~40 methods returning `Result<_, String>`. The `String` destroys all structured error info at the most-trafficked seam between runtime and kernel. Downstream code resorts to `.contains("not found")` substring matches to map errors back to user-visible categories — the very anti-pattern `LlmError::failover_reason` was written to escape.

## Location
`crates/librefang-kernel-handle/src/lib.rs:30-651`

```rust
async fn spawn_agent(...) -> Result<(String, String), String>;
async fn send_to_agent(...) -> Result<String, String>;
async fn task_post(...) -> Result<String, String>;
async fn knowledge_add_entity(...) -> Result<String, String>;
async fn run_workflow(...) -> Result<(String, String), String>;
// 30+ more, all `Result<_, String>`
```

## Why it bites
`librefang-types::error::LibreFangError` already exists as the official structured taxonomy (`AgentNotFound`, `CapabilityDenied`, `QuotaExceeded`, `InvalidState`, `ShuttingDown`, `AuthDenied`, …). Runtime cannot distinguish "agent not found" (404) from "kernel shutting down" (retry later) from "capability denied" (403) without parsing English error strings. Adding new categories (e.g. `RateLimited`, `BudgetExhausted`) requires churning string-matching audits.

## Fix
- Introduce `pub type KernelResult<T> = Result<T, LibreFangError>;` in `librefang-types`.
- Mark `LibreFangError` `#[non_exhaustive]`.
- Migrate hot methods first: `send_to_agent`, `spawn_agent`, `task_*`, `request_approval`.
- Provide a transitional `From<LibreFangError> for String` so call sites compile during the rollout.

The trait is the single chokepoint for runtime↔kernel; this is the right place to fix it once.

houko:
已合入 main，关闭。

7 个分阶段 PR + finish + follow-up 全部合并：
- e8e9df9b refactor(kernel-handle): typed KernelOpError on EventBus / KnowledgeGraph / CronControl (#3541 1/N)
- d09cf6a9 refactor(kernel-handle): typed KernelOpError on TaskQueue (#3541 2/N)
- bb303692 refactor(kernel-handle): typed KernelOpError on MemoryAccess (#3541 3/N)
- 720eead1 refactor(kernel-handle): typed KernelOpError on Hands/Workflow/Goal (#3541 4/N)
[...]

PR #6967, opened 2026-08-12
fix(api): scrub kernel server errors

houko:
## Summary
- preserve typed HTTP status and machine-readable error codes for kernel operation failures
- log full server-side errors while returning generic messages for 500 and 503 responses
- retain actionable error details for client-caused 4xx failures

## Testing
- `cargo test -p librefang-api internal_kernel_errors_are_scrubbed_but_keep_typed_code --lib`
- `cargo test -p librefang-api client_kernel_errors_retain_actionable_message --lib`
- `cargo test -p librefang-api unavailable_kernel_errors_are_scrubbed --lib`
- `git diff --check`

### DIFF
--- a/crates/librefang-api/src/error.rs
+++ b/crates/librefang-api/src/error.rs
@@ -89,13 +89,62 @@ impl From<KernelOpError> for ApiErrorResponse {
     fn from(err: KernelOpError) -> Self {
         let status = kernel_op_status(&err);
         let code = kernel_op_code(&err).to_string();
+        let error = if status.is_server_error() {
+            tracing::error!(error = %err, code, "kernel error scrubbed before response");
+            if status == StatusCode::SERVICE_UNAVAILABLE {
+                "Service unavailable".to_string()
+            } else {
+                "Internal server error".to_string()
+            }
+        } else {
+            err.to_string()
+        };
         ApiErrorResponse {
-            error: err.to_string(),
+            error,
             code: Some(code.clone()),
             r#type: Some(code),
             details: None,
             request_id: None,
             status,
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn internal_kernel_errors_are_scrubbed_but_keep_typed_code() {
+        let response = ApiErrorResponse::from(KernelOpError::Internal(
+            "database path /srv/private/state.db failed".to_string(),
+        ));
+
+        assert_eq!(response.status, StatusCode::INTERNAL_SERVER_ERROR);
+        assert_eq!(response.error, "Internal server error");
+        assert_eq!(response.code.as_deref(), Some("internal_error"));
+        assert!(!response.error.contains("/srv/private"));
+    }
+
+    #[test]
+    fn client_kernel_errors_retain_actionable_message() {
+        let response = ApiErrorResponse::from(KernelOpError::InvalidInput(
+            "agent name must not be empty".to_string(),
+        ));
+
+        assert_eq!(response.status, StatusCode::BAD_REQUEST);
+        assert!(response.error.contains("agent name must not be empty"));
+        assert_eq!(response.code.as_deref(), Some("invalid_input"));
+    }
+
+    #[test]
+    fn unavailable_kernel_errors_are_scrubbed() {
+        let response = ApiErrorResponse::from(KernelOpError::Unavailable(
+            "internal queue worker crashed".to_string(),
+        ));
+
+        assert_eq!(response.status, StatusCode::SERVICE_UNAVAILABLE);
+        assert_eq!(response.error, "Service unavailable");
+        assert_eq!(response.code.as_deref(), Some("service_unavailable"));
     }
 }
--- a/changelog.d/security/6967-kernel-error-scrubbing.md
+++ b/changelog.d/security/6967-kernel-error-scrubbing.md
@@ -1,0 +1,2 @@
+`From<KernelOpError> for ApiErrorResponse` echoed the kernel's `Display` string straight into the HTTP body for every 500 and 503 response, so an `Internal` or `Unavailable` variant could surface database paths, file paths, or other internal state to the client.
+Server-error responses now return a fixed generic message (`Internal server error` / `Service unavailable`) while the full error is still logged server-side via `tracing::error!`; 4xx responses keep their actionable, client-caused message unchanged (#6967) (@houko)


crates/librefang-api/src/error.rs
```diff
@@ -89,8 +89,18 @@ impl From<KernelOpError> for ApiErrorResponse {
     fn from(err: KernelOpError) -> Self {
         let status = kernel_op_status(&err);
         let code = kernel_op_code(&err).to_string();
+        let error = if status.is_server_error() {
```
houko, line 92: This scrubbing is correct for `KernelOpError → ApiErrorResponse`, but it's not the only path that turns a server-side error into a raw `err.to_string()` in a 5xx/502 HTTP body — and at least one sibling path can leak more than a path or internal-state string.

`crates/librefang-api/src/routes/media.rs:39-62` (`media_error_response`) builds `ApiErrorResponse { error: err.to_string(), .. }` unconditionally for every `MediaError` variant, including `MediaError::Other(_)` → `500` (media.rs:52) and `MediaError::Http(_)` / `MediaError::Api{..}` → `502`/upstream status (media.rs:43-47).

That matters because `MediaError::Http` is built directly from provider HTTP failures: `crates/librefang-runtime-media/src/gemini.rs:93-95` and `src/google_tts.rs:98` construct the request URL as `"{base}/...:predict?key={api_key}"` (raw API key in the query string) and call `.post(&url)`; on failure they do `.map_err(|e| MediaError::Http(e.to_string()))` (`gemini.rs:105`, `google_tts.rs:107`). `reqwest::Error`'s `Display` includes the request URL for send/connect errors, so a failed Gemini/Google-TTS call can echo the operator's provider API key straight back to the API caller in the `502` response body — a more severe version of exactly the leak this PR is closing for `KernelOpError`.

Separately, `crates/librefang-api/src/routes/webhooks.rs:145-156` (`webhook_wake`): on `EventBus::publish_event` failure (`Result<(), KernelOpError>`, see `librefang-kernel-handle/src/event_bus.rs:12-16`) it builds a translated string embedding `e.to_string()` (webhooks.rs:151-154) and passes it to `ApiErrorResponse::internal(err_msg)` (webhooks.rs:156) — a 500 response. This bypasses both this PR's new `From<KernelOpError>` scrubbing *and* the existing `ApiErrorResponse::internal_scrub()` helper already sitting in `crates/librefang-api/src/types.rs:302-312` (log full error, return generic "Internal server error"), which is the correct pattern already used elsewhere and would be a near drop-in fix here.

Neither of these files is touched by this PR's diff, so I'm not confident enough to edit them directly as part of this review, but given the PR's stated goal ("scrub kernel server errors" before they reach API responses), I'd flag both as in-scope gaps in the same fix rather than separate follow-ups — the media.rs one in particular looks like a real secret-leak, not just an internal-details leak.

---

houko (commented):
Reviewed the kernel-error-scrubbing fix. The `From<KernelOpError>` change itself is sound (correct status-code branch, logs full detail server-side via `tracing::error!`, keeps 4xx messages actionable, has direct unit-test coverage). I pushed one mechanical fix (missing changelog fragment) and left one inline comment flagging two adjacent error paths (`routes/media.rs`, `routes/webhooks.rs`) that still leak raw error strings — including a plausible provider-API-key leak via Gemini/Google-TTS `MediaError::Http` — through 5xx/502 responses without this PR's scrubbing.

---

houko:
Automated review pass: commit `5ce29b0` ("docs(changelog): add missing security fragment for kernel error scrubbing") on this branch is authored as `Claude <noreply@anthropic.com>`.

---

crates/librefang-api/src/error.rs
```diff
@@ -89,8 +89,18 @@ impl From<KernelOpError> for ApiErrorResponse {
...
+            tracing::error!(error = %err, code, "kernel error scrubbed before response");
+            if status == StatusCode::SERVICE_UNAVAILABLE {
+                "Service unavailable".to_string()
+            } else {
+                "Internal server error".to_string()
+            }
+        } else {
```
houko, line 99: Nit (consistency, non-blocking): this reimplements scrubbing inline instead of reusing `ApiErrorResponse::internal_scrub` (`crates/librefang-api/src/types.rs`), which already does exactly this — log the full error via `tracing::error!`, return a fixed `"Internal server error"` string — and whose own doc comment says "new code (and audited rewrites) should use `internal_scrub`". This PR is exactly that kind of audited rewrite.

The reason it wasn't reused is presumably that `internal_scrub` always returns 500 and drops `code`/`type`, while this call site needs to preserve the typed `code` and branch to 503 for `Unavailable`/`ShuttingDown`. Worth considering either:
- extending `internal_scrub` (or adding a sibling, e.g. `internal_scrub_with_code(status, code, err)`) so both call sites share one scrubbing implementation, or
- at minimum leaving a comment here noting the divergence from `internal_scrub` and why, so a future reader doesn't assume it was missed.

As-is there's no bug, but the codebase now has two independent places that decide what "scrubbed" means, and a future edit to one (e.g. changing the generic message text, or the log field shape) can silently drift from the other. Separately, worth noting `crates/librefang-api/src/routes/media.rs::media_error_response` still does `error: err.to_string()` for `MediaError::Other(_)` mapped to a 500 status — the same leak class this PR fixes for `KernelOpError`, just not in a file this PR touches.

---

houko (commented):
Automated review pass on the kernel-error-scrubbing fix. No panic risk found in the scrubbing branch, server-side logs retain full detail (not over-scrubbed), and the changelog fragment is correctly formatted and attributed. One non-blocking consistency nit posted inline.

---

houko:
Reviewed. The fix itself is correct and well-placed: the scrub sits in the single existing `KernelOpError → ApiErrorResponse` boundary (`error.rs`, per the module's own #3541 doc comment — every route already delegates here via `?` / `Into::into` instead of building its own mapping), not an ad-hoc per-call-site patch. Traced all three added tests against the logic by hand (`is_server_error()` gates the scrub, `SERVICE_UNAVAILABLE` gets its own message, 4xx keeps `err.to_string()`) — matches. No conflict with the `ErrorTranslator` layer: that's a separate code path (e.g. `routes/agents/messaging.rs`) that already does its own 500-scrubbing for `KernelError`/`LibreFangError`, unrelated to this `KernelOpError` boundary, and there's no `.await` in the touched code so the `!Send` gotcha doesn't apply here. Changelog fragment is present and correctly formatted.

One related, out-of-scope observation while checking for consistent scrubbing repo-wide: `routes/media.rs::media_error_response` builds `ApiErrorResponse` directly from `MediaError` and uses `err.to_string()` unscrubbed for its 5xx arms too (`MediaError::Http(_) → 502`, `MediaError::Other(_) → 500`), which can echo raw upstream/network error text (`media.rs:43,52,55`). Not part of this diff and a different error type, so not blocking here — just flagging since it's the same class of leak this PR fixes for `KernelOpError`, in case it's worth a follow-up.

---

PR #6968, opened 2026-08-12
fix(audit): fail integrity checks on row load errors

houko:
## Summary
- stop silently discarding malformed SQLite audit rows during startup
- retain row-load failures on the audit log so integrity verification fails closed
- count all persisted rows, including malformed rows excluded from the in-memory chain
- add a regression test with an invalid SQLite column type

## Testing
- `cargo test -p librefang-runtime-audit --lib`
- `cargo clippy -p librefang-runtime-audit --lib --tests -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/librefang-runtime-audit/src/lib.rs
+++ b/crates/librefang-runtime-audit/src/lib.rs
@@ -397,6 +397,10 @@ pub struct AuditLog {
     /// repopulates the full window. Every mutation and read happens under
     /// the `entries` mutex, so `Relaxed` ordering is sufficient.
     persisted_rows: AtomicUsize,
+    /// Any failure encountered while reloading persisted rows. A partially
+    /// decoded audit trail must never verify as intact merely because the
+    /// malformed row was skipped by the SQLite iterator.
+    load_error: Mutex<Option<String>>,
 }
 
 /// Per-trim summary returned by [`AuditLog::trim`].
@@ -445,6 +449,7 @@ impl AuditLog {
             chain_anchor: Mutex::new(None),
             max_in_memory_entries: AtomicUsize::new(0),
             persisted_rows: AtomicUsize::new(0),
+            load_error: Mutex::new(None),
         }
     }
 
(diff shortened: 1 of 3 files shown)


houko:
Automated review pass: commit `261a884` ("docs(changelog): add fragment for audit row-load-error fix") on this branch is authored as `Claude <noreply@anthropic.com>`.
CLAUDE.md's collaboration policy forbids Claude/Anthropic attribution in commit messages, and the repo's `commit-msg` hook is meant to reject that exact author identity — it just wasn't active in the environment that produced this commit (this session's own environment defaulted to that identity too; fixed for the remainder of this review pass).

---

crates/librefang-runtime-audit/src/lib.rs
```diff
@@ -613,61 +617,93 @@ impl AuditLog {
...
+                    Err(error) => {
+                        load_error = Some(format!("failed to prepare audit reload query: {error}"));
                     }
                 }
             }
+            Err(error) => {
+                load_error = Some(format!(
+                    "failed to acquire audit database connection: {error}"
+                ));
+            }
```
houko, line 706: Good fix overall — `verify_integrity` now fails closed for every load-error branch (connection, prepare, query, per-row decode), and the new test exercises the per-row-decode path well.

One edge case worth a maintainer call, not fixed here since it's outside this PR's stated scope (row-decode errors) and touches the write path's seq-derivation, not just the loader: when `pool.get()`, `db.prepare()`, or `stmt.query_map()` itself fails (as opposed to an individual row failing to decode), `entries` stays empty and `persisted_count` stays `0` even though the on-disk table may already hold rows with `seq` 0..N. `verify_integrity()` correctly reports the failure via `load_error`, but nothing stops `record_with_context` from being called afterward — it derives the next `seq` as `entries.last().map(|e| e.seq + 1).unwrap_or(0)`, i.e. `0`, which collides with the real row 0 already in the table on every subsequent write attempt (the `INSERT` fails on the primary-key conflict and is dropped/logged, so no corruption, but the daemon effectively loses audit logging until restart). This failure mode predates this PR (the old code had the same empty-`entries`-on-`pool.get()`-failure behavior), so it's not a regression, but a fix targeting the general "load errors must fail closed" theme could also consider refusing new writes (or short-circuiting to an error) while `load_error` is `Some(_)`, rather than only failing `verify_integrity()`.

---

crates/librefang-runtime-audit/src/lib.rs
```diff
@@ -613,61 +617,93 @@ impl AuditLog {
...
+                    }
+                    Err(error) => {
+                        load_error = Some(format!("failed to prepare audit reload query: {error}"));
                     }
                 }
             }
+            Err(error) => {
+                load_error = Some(format!(
+                    "failed to acquire audit database connection: {error}"
+                ));
```
houko, line 705: This branch (and the sibling `db.prepare(...)` / `stmt.query_map(...)` failure branches above it) sets `load_error` for a connection-acquire or query failure exactly the same way as a genuinely malformed row.
That conflates two very different failure classes: a decode failure on a specific row is real evidence the persisted chain is corrupt/tampered, but `pool.get()` failing is typically transient (pool exhaustion, a busy-timeout under WAL lock contention, momentary I/O hiccup) and says nothing about the data itself — `entries` stays empty in this case, so the DB rows are almost certainly fine and unread, not corrupt.

Once this fires, `load_error` is permanently `Some` for the lifetime of this `AuditLog` (never cleared), so `verify_integrity()` fails closed until the daemon restarts — and the boot WARN this triggers (a few lines down, `count > 0 || load_error.is_some()`) tells the operator to run `librefang security audit-reset`, which **truncates the chain**. Recommending a destructive reset for what may be a one-off transient connection failure, with no retry attempted first, risks discarding a perfectly intact forensic trail over a hiccup that a second `pool.get()` call would likely not reproduce.

Suggest either: (a) retry the connection acquire (and/or the prepare/query) with a short backoff before treating it as a load failure, or (b) keep connection/query failures in a distinct variant from row-decode failures so the boot-time message and `/api/audit/verify` output can tell operators "retry me" apart from "investigate possible tampering" instead of funneling both into the same generic "audit trail was only partially loaded" string.

---

crates/librefang-runtime-audit/src/tests.rs
```diff
@@ -153,6 +153,44 @@ fn test_record_with_context_persists_user_and_channel() {
...
+            );",
+        )
+        .unwrap();
+    }
+
+    let log = AuditLog::with_db(pool);
+
+    assert_eq!(log.len(), 0, "malformed rows must not enter the chain");
+    let error = log.verify_integrity().unwrap_err();
+    assert!(error.contains("only partially loaded"), "{error}");
```
houko, line 190: The new test only exercises the row-decode failure path (a bad column type). The PR also changes behavior for the `pool.get()` failure and `db.prepare()`/`query_map()` failure branches in `with_db` (lines ~692-706 in `lib.rs`) — e.g. closing/poisoning the pool before construction, or pointing it at a connection whose `audit_entries` table is missing, would exercise the "failed to query audit rows" / "failed to prepare audit reload query" paths. Given the previous bug was specifically about a load failure being silently swallowed, it'd be worth covering at least one of those non-decode failure branches too, since they're the paths most likely to be transient/retryable rather than genuine corruption (see the other comment on `lib.rs`).

---

houko (commented):
Automated review pass (comment-only). The core fix is correct — row-load failures now fail closed instead of being silently skipped, and the single row-loading call site (`with_db`, which `with_db_anchored` delegates to) is fully covered. Two findings posted inline: a transient-vs-corruption conflation in the connection/query failure branches that can push operators toward a destructive `audit-reset` over a one-off hiccup, and a test-coverage gap for those same non-decode failure branches. Changelog fragment and prose formatting look correct.

---

crates/librefang-runtime-audit/src/lib.rs
```diff
@@ -613,61 +617,93 @@ impl AuditLog {
...
+                        }
+                    }
+                    Err(error) => {
+                        load_error = Some(format!("failed to prepare audit reload query: {error}"));
                     }
                 }
             }
+            Err(error) => {
+                load_error = Some(format!(
+                    "failed to acquire audit database connection: {error}"
```
houko, line 704: Before this change, `verify_integrity()`'s `Err` string was always self-generated ("chain break at seq N", "hash mismatch", etc.) — no raw driver text.
After this change, `load_error` can carry a raw `{error}` `Display` straight from `rusqlite`/`r2d2` (here, and at the "failed to decode audit row", "failed to query audit rows", and "failed to prepare audit reload query" sites above), and `verify_integrity()` (line ~1017) wraps it into `"audit trail was only partially loaded: {error}"` unchanged.

`GET` `crates/librefang-api/src/routes/audit.rs:614-616` (`audit_verify`) echoes that string verbatim as `"error": msg` in the JSON response — it isn't touched by this diff, but the *content* it now surfaces is new. A `rusqlite::Error::SqliteFailure` for a connection failure can include the on-disk DB path; a decode failure can include column index/expected-type detail. The endpoint is admin-gated (`require_admin`), so this is lower severity than an unauthenticated leak, but it's the same class of schema/filesystem disclosure that `ApiErrorResponse::internal_scrub` (see #6967, and PR #6971 in this same batch) was introduced to close off.

Worth scrubbing the driver-error detail into the logged `tracing::error!`/`warn!` (already happening) and keeping only a generic classifier (e.g. "connection failed" / "query failed" / "row N failed to decode") in the string that reaches `verify_integrity()` and, from there, the HTTP response.

---

houko (commented):
Logic and locking in the runtime-audit change look correct (single new mutex, no lock-order inversion, guard dropped before nothing blocking happens inside it, regression test exercises the exact malformed-row case). Left one comment on the error-scrubbing angle given this batch's theme. Also noting for the record (not re-flagging, already covered by the existing thread comment): commit `261a884` on this branch carries a `Claude <noreply@anthropic.com>` author identity that should be dropped before merge.

---

crates/librefang-runtime-audit/src/lib.rs
houko, line 704: This `pool.get()` failure branch sets `load_error` the same way a genuinely corrupt row does, and `load_error` is never cleared once set (`verify_integrity`, lines 1011-1018, checks it unconditionally on every call). So a one-off connection hiccup during boot — pool exhaustion, a transient lock, nothing wrong with the data — permanently flips `verify_integrity()` to `Err` for the rest of the process's life, even after new entries record and would otherwise verify cleanly.

Downstream this also mislabels the failure: `/api/audit/verify` (crates/librefang-api/src/routes/audit.rs:608-617) reports any `verify_integrity()` `Err` with `anchor_status: "diverged"` when an anchor is configured, which now fires for a load-time infra hiccup that has nothing to do with the anchor.

Is "fail closed forever" the intended blast radius for a transient connection/query failure at construction time, same as for an actual malformed row? If so it's worth a one-line comment saying so explicitly, since it's a meaningful behavior change from the pre-PR code (which silently ignored `pool.get()` failure entirely).

---

crates/librefang-runtime-audit/src/lib.rs
```diff
@@ -613,61 +617,93 @@ impl AuditLog {
...
+                            })
+                        });
+                        match rows {
+                            Ok(rows) => {
+                                for (row_index, result) in rows.enumerate() {
+                                    persisted_count += 1;
+                                    match result {
+                                        Ok(entry) => {
+                                            tip = entry.hash.clone();
+                                            entries.push(entry);
```
houko, line 680: The loop keeps iterating past a decode error (only the *first* error is kept via `load_error.get_or_insert`, line 687), so `entries` can end up with a seq gap — e.g. seq 0-4 load fine, seq 5 is malformed, seq 6-10 still get pushed here. `verify_integrity()` now correctly fails closed for that case via the new `load_error` check, which is the fix this PR is about.

But `AuditLog::recent()` and `since_seq()` (lines 1157-1185) don't consult `load_error` at all — they'll serve the gapped entry list with no signal that anything was skipped, to whatever calls them (the dashboard "recent" view, and the `/api/logs/stream` SSE endpoint via `since_seq`). That gap already existed before this PR (the old code `.flatten()`'d the same rows away silently), so it's not a regression — but since this PR is already threading a `load_error` signal through the struct specifically to stop a partial load from looking complete, it seems worth deciding whether those two read paths should also expose the flag (or at minimum, whether it's acceptable that only the boot-time integrity check is fail-closed while direct entry reads are not). Raising as a scope question rather than a blocker.

---

houko (commented):
Independent review pass against CLAUDE.md conventions (changelog format, prose wrapping, deterministic ordering, fail-closed error handling). The core fix is sound: `verify_integrity()` correctly fails closed on any row-load error, no unwrap-on-error/panic paths were introduced (mutex locks all use `unwrap_or_else(|e| e.into_inner())`), the changelog fragment is correctly formatted under `changelog.d/fixed/`, and a prior commit on this branch already fixed a prose-wrap violation in the new doc comments. No mechanical fixes applied. Two design questions left as inline comments for a maintainer call — neither blocks the fix itself.

---

PR #6969, opened 2026-08-12
fix(kernel): serialize pairing token redemption

houko:
## Summary
- serialize pairing token lookup/consumption, device-cap enforcement, and device insertion
- prevent concurrent reuse of a one-time pairing token
- prevent concurrent redemptions from exceeding `max_devices`
- release the redemption lock before invoking the persistence callback

## Testing
- `cargo test -p librefang-kernel concurrent_redemption --lib`
- `cargo test -p librefang-kernel test_complete_pairing --lib`
- `cargo test -p librefang-kernel test_max_devices_enforced --lib`
- `git diff --check`

The test linker emitted the existing macOS `__eh_frame section too large` warning; all requested commands exited successfully.


crates/librefang-kernel/src/pairing.rs
```diff
@@ -56,6 +57,10 @@ pub struct PairingManager {
     config: PairingConfig,
     pending: DashMap<String, PairingRequest>,
     devices: DashMap<String, PairedDevice>,
+    /// Serializes the token-consume, device-cap check, and device insert as
+    /// one transaction. DashMap makes each operation safe individually but
+    /// cannot make this multi-map redemption sequence atomic.
```
houko: Prose-wrapping: this new doc comment hard-wraps mid-sentence ("...device insert as / one transaction." and "...individually but / cannot make..."), which violates the repo's sentence-boundary-only line-break rule (CLAUDE.md, "Prose wrapping: no column limit; break only at sentence boundaries" — explicitly applies to doc-comments). Please reflow to one sentence per line, e.g.:

```rust
/// Serializes the token-consume, device-cap check, and device insert as one transaction.
/// DashMap makes each operation safe individually but cannot make this multi-map redemption sequence atomic.
```

---

crates/librefang-kernel/src/pairing.rs
```diff
@@ -161,6 +169,11 @@ impl PairingManager {
         let device_id = device_info.device_id.clone();
         self.devices.insert(device_id.clone(), device_info.clone());
 
+        // The security-sensitive state transition is complete. Persistence
+        // callbacks may perform blocking I/O and must not serialize unrelated
+        // future redemptions.
```
houko: Same prose-wrapping issue here — this comment is hard-wrapped mid-sentence rather than broken at sentence boundaries (CLAUDE.md rule applies to source comments too). Suggested reflow:

```rust
// The security-sensitive state transition is complete.
// Persistence callbacks may perform blocking I/O and must not serialize unrelated future redemptions.
```

---

houko (commented):
Automated review pass. The core fix is sound: `redemption_lock` (parking_lot::Mutex, consistent with #6970's pattern) is acquired at function entry and held across the full find → expiry-check → cap-check → remove → insert sequence, then explicitly dropped before the persistence callback — genuinely atomic, not just a narrowed window. It correctly serializes across different tokens too, so the cap check holds even when concurrent redemptions use distinct one-time tokens. `complete_pairing` is the single entry point (only caller is `routes/pairing.rs`), so the fix covers the whole surface. New tests (`concurrent_redemption_consumes_token_exactly_once`, `concurrent_redemptions_respect_device_cap`) exercise both the double-redemption and cross-token cap scenarios. Changelog fragment at `changelog.d/security/6969-pairing-redemption-race.md` is correctly formatted and attributed. Two minor prose-wrapping nits left inline on new doc/inline comments that hard-wrap mid-sentence.

---

PR #6970, opened 2026-08-12
fix(kernel): serialize cron job limit checks

houko:
## Summary
- serialize cron capacity checks, validation, and insertion
- prevent concurrent creators from exceeding the global job limit
- keep per-agent job-count validation on the same stable add sequence
- add a 16-thread regression test for a one-job global cap

## Testing
- `cargo test -p librefang-kernel concurrent_adds_respect_global_limit --lib`
- `cargo test -p librefang-kernel test_add_job_global_limit --lib`
- `cargo test -p librefang-kernel test_add_job_per_agent_limit --lib`
- `git diff --check`

The test linker emitted the existing macOS `__eh_frame section too large` warning; all requested commands exited successfully.

### DIFF
--- a/crates/librefang-kernel/src/cron.rs
+++ b/crates/librefang-kernel/src/cron.rs
@@ -147,6 +147,10 @@ pub struct CronScheduler {
     home_dir: PathBuf,
     /// Global cap on total jobs across all agents (atomic for hot-reload).
     max_total_jobs: AtomicUsize,
+    /// Serializes capacity checks with insertion. DashMap makes each map
+    /// operation atomic, but a separate `len()` check otherwise lets
+    /// concurrent creators exceed global and per-agent limits.
+    add_lock: std::sync::Mutex<()>,
     /// Serializes `persist()` writes so concurrent callers (cron loop, API
     /// routes, spawned cron tasks) don't corrupt the tmp file by interleaving
     /// `O_TRUNC`/write/rename on the same path.
@@ -165,6 +169,7 @@ impl CronScheduler {
             persist_path: home_dir.join("data").join("cron_jobs.json"),
             home_dir: home_dir.to_path_buf(),
             max_total_jobs: AtomicUsize::new(max_total_jobs),
+            add_lock: std::sync::Mutex::new(()),
             persist_lock: std::sync::Mutex::new(()),
         }
     }
@@ -294,6 +299,8 @@ impl CronScheduler {
     /// `one_shot` controls whether the job is removed after a single
     /// successful execution.
     pub fn add_job(&self, mut job: CronJob, one_shot: bool) -> LibreFangResult<CronJobId> {
+        let _add_guard = self.add_lock.lock().unwrap_or_else(|e| e.into_inner());
+
         // Global limit
         let max_jobs = self.max_total_jobs.load(Ordering::Relaxed);
         if self.jobs.len() >= max_jobs {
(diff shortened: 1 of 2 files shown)


crates/librefang-kernel/src/cron.rs
```diff
@@ -147,6 +147,10 @@ pub struct CronScheduler {
     home_dir: PathBuf,
     /// Global cap on total jobs across all agents (atomic for hot-reload).
     max_total_jobs: AtomicUsize,
+    /// Serializes capacity checks with insertion. DashMap makes each map
```
houko: The mutex itself looks correct — `add_lock` is acquired at the top of `add_job` and the guard lives for the whole function scope, so the global/per-agent `len()` checks and the eventual insert now share one critical section (verified by reading the full diff; the new `concurrent_adds_respect_global_limit` test exercises 16 barrier-synchronized threads against `max_total_jobs=1` and asserts exactly one winner, which matches the fix).

Two smaller things per CLAUDE.md:

1. This new doc comment hard-wraps mid-sentence ("Serializes capacity checks with insertion. DashMap makes each map / operation atomic, but a separate `len()` check otherwise lets / concurrent creators exceed global and per-agent limits.") — CLAUDE.md's prose rule is one sentence per line, broken only at sentence boundaries, not at a fixed column. Suggest:
```
/// Serializes capacity checks with insertion.
/// DashMap makes each map operation atomic, but a separate `len()` check otherwise lets concurrent creators exceed global and per-agent limits.
```
2. No `changelog.d/` fragment for this PR. Given it's a resource-limit-bypass race fix, `changelog.d/security/6970-serialize-cron-job-limit-checks.md` seems warranted, e.g.:
```
Serialize the global and per-agent cron job limit checks with insertion so concurrent `add_job` callers can no longer all pass the same capacity check and jointly exceed `max_total_jobs` or the per-agent cap.
`DashMap` made each individual map operation atomic, but the separate `len()` and per-agent count reads that gated insertion were not covered by any lock, leaving a check-then-act window that concurrent requests from the same or different agents could race through (#6970) (@houko)
```

---

crates/librefang-kernel/src/cron.rs
```diff
@@ -294,6 +298,8 @@ impl CronScheduler {
     /// `one_shot` controls whether the job is removed after a single
     /// successful execution.
     pub fn add_job(&self, mut job: CronJob, one_shot: bool) -> LibreFangResult<CronJobId> {
+        let _add_guard = self.add_lock.lock().unwrap_or_else(|e| e.into_inner());
```
houko, line 301: `add_job` now serializes its whole check-then-insert sequence under `add_lock`, and the two concurrent regression tests confirm both the global and per-agent caps hold under contention.

Flagging a related, pre-existing gap that this fix doesn't touch: `update_job` (line 451, `candidate.validate_with_home(0, Some(&self.home_dir))`) hardcodes `agent_count = 0`, so a `PUT` that changes `agent_id` never checks the target agent's per-agent job count at all — not just "not atomic with adds," but not checked. The inline comment at line 449 calls this "tracked separately," so it may already have its own issue, but since this PR is specifically about closing the cron job-limit race, it seemed worth surfacing here in case the two are meant to land together (an agent-reassigning update racing concurrent `add_job` calls for the same target agent could still push that agent over `MAX_JOBS_PER_AGENT`, just via a different path than the one this PR closes).

Not blocking — just want a maintainer call on whether it belongs in this PR's scope or genuinely is a separate follow-up.

---

crates/librefang-kernel/src/cron.rs
```diff
@@ -1510,6 +1516,68 @@ mod tests {
...
+        assert_eq!(successes, 1);
+        assert_eq!(sched.total_jobs(), 1);
+    }
+
+    #[test]
+    fn concurrent_adds_respect_per_agent_limit() {
+        // MAX_JOBS_PER_AGENT = 50 in librefang-types.
+        const CONTENDERS: usize = 60;
+        let tmp = tempfile::tempdir().unwrap();
+        // Global limit set well above the per-agent cap so only the
```
houko, line 1553: Nit: this new comment hard-wraps mid-sentence ("... so only the / per-agent check can reject a contender."), which conflicts with CLAUDE.md's prose rule (one sentence per line, break only at sentence boundaries — applies to any multi-line free-form prose comment block). Suggest joining to one line:
```rust
// Global limit set well above the per-agent cap so only the per-agent check can reject a contender.
```

---

PR #6971, opened 2026-08-12
fix(api): surface usage query failures

houko:
## Summary

- Stop presenting failed usage-store queries as zero spend or empty result sets.
- Return a scrubbed 500 with stable `usage_query_failed` code across usage and budget read endpoints.
- Preserve agent-id response metadata on per-agent query failures.
- Serialize global and provider budget snapshot-and-merge operations under the config write lock so concurrent partial PUTs do not erase each other.
- Build global and provider responses and audit diffs from the live post-reload budget snapshot.
- Return 404 for unknown user-budget targets and report `enforced: false` when the user has no configured budget.
- Roll back the complete live agent quota and return a scrubbed 500 with an error audit row when durable agent persistence fails.
- Add route-level regressions for usage-store failures, concurrent partial updates, agent persistence rollback, unknown users, and enforcement reporting.

## Verification

- `cargo test -p librefang-api --test budget_routes_test` (32 passed)
- `cargo test -p librefang-api --test provider_budget_routes_test` (7 passed)
- `cargo test -p librefang-api --test api_integration_test test_user_budget_detail_` (3 passed)
- `git diff --check`

## Out of scope

- Broader budget policy changes remain outside this route-contract fix.
- PR #7037 separately covers the budget-status query path and overlaps only at the file level.


crates/librefang-api/src/routes/budget.rs
```diff
@@ -137,6 +137,12 @@ fn fmt_global_budget_diff(
 // Usage endpoint
 // ---------------------------------------------------------------------------
 
+fn usage_query_error(error: impl std::fmt::Display) -> Response {
+    ApiErrorResponse::internal_scrub(error)
+        .with_code("usage_query_failed")
+        .into_response()
+}
```
houko, line 144: Cross-checking against #6967 (`fix/kernel-error-scrubbing`), which is open in the same batch and also scrubs 5xx bodies before they reach the client.

I don't think there's an actual overlap to resolve here: #6967 fixes `From<KernelOpError> for ApiErrorResponse` in `error.rs`, scoped to errors that cross the `KernelHandle` trait boundary. Everything this PR touches goes through `state.kernel.memory_substrate().usage()` / `librefang_memory::usage::UsageStore` directly and returns `librefang_types::error::LibreFangResult` (`LibreFangError`), never a `KernelOpError` — so `usage_query_error` here can't reuse #6967's conversion impl without a type it doesn't have.

What this PR does reuse is the pre-existing, already-widely-used `ApiErrorResponse::internal_scrub()` helper (from #5378, ~46 call sites across the crate) rather than hand-rolling a new mapping, so it's already following the established pattern for this error type.

Flagging only because the two PRs land in the same window and both are "stop leaking raw error internals into 500 bodies" — worth a maintainer glance in case there's an appetite to eventually collapse `LibreFangError` handling and `KernelOpError` handling onto one shared scrub path, but I don't see a concrete action for *this* PR to take.

---

crates/librefang-api/src/routes/budget.rs
```diff
@@ -1116,9 +1142,18 @@ pub async fn user_budget_detail(
...
-    let monthly = usage_store.query_user_monthly(user_id).unwrap_or(0.0);
+    let hourly = match usage_store.query_user_hourly(user_id) {
+        Ok(value) => value,
+        Err(error) => return usage_query_error(error),
+    };
+    let daily = match usage_store.query_user_daily(user_id) {
+        Ok(value) => value,
+        Err(error) => return usage_query_error(error),
+    };
+    let monthly = match usage_store.query_user_monthly(user_id) {
```
houko, line 1181: `user_budget_detail` (`GET /api/budget/users/{user_id}`) gets the identical `unwrap_or(0.0)` → propagate-error fix as the other endpoints in this PR, but it's the one endpoint in the diff with no regression test for the new fail-closed behavior. CLAUDE.md's MANDATORY Integration Testing section requires a `#[tokio::test]` against `TestServer` for any route/wiring change, and the new `usage_queries_fail_closed_when_storage_is_unavailable` test in `budget_routes_test.rs` deliberately covers eight endpoints but not this one — likely because it's admin-gated (`require_admin_for_user_budget`) and this test's harness (`MockKernelBuilder`, no auth) can't reach past the 403. `api_integration_test.rs::test_user_budget_detail_includes_enforced_true` already shows the pattern for authenticating as admin via `start_test_server_with_rbac_users`, but that harness doesn't currently expose the SQLite pool needed for the `DROP TABLE usage_events` trick — combining the two is a design call, not a copy-paste, so flagging rather than fixing directly.

---

crates/librefang-api/tests/budget_routes_test.rs
houko, line 1: The changelog fragment (as fixed up in af32bde) now lists `/api/budget/users/{user_id}` among the endpoints that got the unwrap→propagate fix, and `user_budget_detail` in `routes/budget.rs` does correctly propagate `query_user_hourly`/`_daily`/`_monthly` errors now.

But this regression test's `paths` list stops at eight endpoints and doesn't include it, and I don't see a matching failure-path assertion added to `api_integration_test.rs` either (only the pre-existing happy-path `test_user_budget_detail_includes_enforced_true`). The file-header comment says user-budget routes are "intentionally out of scope here" for the happy-path suite this file otherwise covers, but that carve-out predates this PR's `usage_query_failed` propagation guarantee — right now nothing in CI actually exercises the fail-closed behavior on the 9th endpoint the changelog explicitly claims. Worth either adding `/api/budget/users/{user_id}` to this test's `paths` (it needs `require_admin_for_user_budget` to pass, so it may need an authenticated request in this harness) or a dedicated case in `api_integration_test.rs` alongside the existing happy-path test, so the claim in the changelog fragment is actually backed by a test per CLAUDE.md's "Reviewers gate PRs on the presence of an integration test" rule.

---

crates/librefang-api/tests/api_integration_test.rs
houko, line 1: Commit hygiene: `3ae1811625c2c6dcb439a70a95cdfd029e099f9d` ("test(api): cover user_budget_detail fail-closed on usage-store errors", which added the test in this file) has its author identity set to `Claude <noreply@anthropic.com>`.

CLAUDE.md, "Commit & PR hygiene": *"No Claude / Anthropic / AI attribution in commit messages, PR bodies, or issue comments... the `commit-msg` git hook rejects matching strings [and] separately rejects a commit whose author identity... resolves to Claude / Anthropic even when the message itself is clean."*

Since this commit is already pushed and part of the branch's public history, fixing it requires rewriting the commit's author (`git commit --amend --author=...` or interactive rebase) and force-pushing — which is outside what an automated review pass should do unilaterally. Flagging for the maintainer to reword the author identity before merge.

Separately (lower priority, same commit plus `43ab983f9` and `af32bde28`): the commit message bodies hard-wrap at ~72–80 columns mid-sentence, e.g. "PR #6971 lists /api/budget/users/{user_id} among the endpoints that now fail / closed on a usage-store query error, but...". Per CLAUDE.md's prose-wrapping rule, commit message bodies should use one sentence per line with no column-width wrapping. Not blocking, just noting the pattern since all three non-code commits in this PR share it.

---

crates/librefang-api/src/routes/budget.rs
houko, line 1: Cross-PR coordination note (CLAUDE.md, "Two green PRs can still break `main` together"): #7037 ("fix: surface budget status query errors") also touches this file today, landing the same fix family.

Checked the actual hunks — no true overlap: #7037 changes `budget_status()` / `update_budget()` (which call `MeteringEngine::budget_status()`, whose signature it changes to `LibreFangResult<BudgetStatus>`), while this PR changes `usage_stats`, `usage_summary`, `usage_by_model[_performance]`, `usage_daily`, `agent_budget_status`, `agent_budget_ranking`, `user_budget_detail`, and `provider_budget_list` — a disjoint set of functions, and #7037's description explicitly scopes per-agent/per-provider routes to this PR. So this isn't a duplicate fix.

Still, both PRs edit the same file and are queued to merge close together — per CLAUDE.md, whichever lands second should get a fresh CI run against the post-merge base (push/update-branch) rather than trusting a check that ran against a stale, pre-merge `main`, since a textually-clean auto-merge here doesn't guarantee the combination compiles/behaves correctly.

---

houko (commented):
Automated review pass against CLAUDE.md.

The code change itself looks solid: `usage_query_error` correctly routes through `ApiErrorResponse::internal_scrub`, so failures come back as a scrubbed 500 with a stable `usage_query_failed` code and no SQL/table-name leakage (confirmed by both new tests asserting the response body doesn't contain "usage_events"), all nine endpoints listed in the changelog fragment have matching drop-table regression coverage across `budget_routes_test.rs` and `api_integration_test.rs`, the changelog fragment is correctly formatted in `changelog.d/fixed/` and ends `(#6971) (@houko)`, and I found no `HashMap`/`HashSet` used where prompt-determinism would require `BTreeMap`/`BTreeSet` (the pre-existing `HashMap` in `agent_budget_ranking` is a lookup table, not something that reaches an LLM prompt).

Two findings left as comments above, both needing human judgment rather than a mechanical fix:
1. A commit in this branch (`3ae1811`) has its git author identity set to `Claude <noreply@anthropic.com>` — a hard-rule violation that needs a maintainer-side author rewrite, not something I should force-push over.
2. A same-day file overlap with #7037 in `budget.rs` — not a duplicate fix (disjoint functions), but worth a fresh CI run on whichever merges second.

No other issues found.

---

crates/librefang-api/tests/api_integration_test.rs
houko, line 1: CI Gate is currently red on this PR (run 31630383114, job "Quality"), and it isn't covered by any of the existing review comments on this thread yet.

The cause is a plain `cargo fmt --check` violation introduced by `test_user_budget_detail_fails_closed_when_storage_is_unavailable` (added in 3ae1811, the commit already flagged elsewhere in this thread for its Claude author identity). `xtask fmt` wants:

```rust
let conn = server.state.kernel.memory_substrate().pool().get().unwrap();
```

collapsed onto one line, instead of the current multi-line chain:

```rust
let conn = server
    .state
    .kernel
    .memory_substrate()
    .pool()
    .get()
    .unwrap();
```

`CI Gate: one or more lanes failed or were cancelled` on this run traces directly to that one formatting diff — every other lane (Ubuntu shards 1-4, Unit lib+bin, Security, OpenAPI Drift, etc.) is green. No commit after 3ae1811 (merge of `main` at 2231bb2c included) re-runs `cargo fmt --fix` for it, so as of the current branch tip this is still unresolved and will keep the gate red until someone runs `cargo fmt` (or hand-edits the one line) and pushes.

Worth fixing alongside the commit-authorship issue already flagged on this PR, since both trace back to the same commit.

---

PR #6958, opened 2026-08-12
ci: pin cargo-deny action to a commit

houko:
## Summary
- pin `EmbarkStudios/cargo-deny-action` to the immutable commit behind its `v2` tag
- retain the release tag as a comment for dependency update tooling
- remove the stale initial-rollout comment that explicitly deferred SHA pinning

## Testing
- verified every `.github/workflows` `uses:` reference is now a 40-character commit SHA
- `git diff --check`

PR #6972, opened 2026-08-12
ci: pin remaining workflow tooling

houko:
## Summary
- pin the build-timings upload-artifact action to the same audited v4 commit used elsewhere in the repository
- pin all Cloudflare deployment invocations to exact Wrangler 4.121.0 instead of the mutable major tag
- leave the cargo-deny mutable action to the already-open dedicated #6958

## Verification
- parsed all four edited workflows with Ruby YAML
- verified no bare `wrangler@4` invocation remains
- verified no mutable `uses:` reference remains except `cargo-deny-action@v2`, covered by #6958
- `git diff --check`

### DIFF
--- a/.github/workflows/build-timings.yml
+++ b/.github/workflows/build-timings.yml
@@ -53,7 +53,7 @@ jobs:
         continue-on-error: true
         run: cargo xtask compare-build-timings
       - name: Upload timing snapshot
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
           name: build-timings-${{ github.sha }}
           path: bench-results/build-timings/
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -65,4 +65,4 @@ jobs:
           # the run-block via shell expansion.  See deploy-web.yml for
           # the same fix; #3938 missed both deploy workflows.
           HEAD_REF: ${{ github.head_ref || github.ref_name }}
-        run: npx -y wrangler@4 pages deploy docs/out --project-name=librefang-docs --branch="$HEAD_REF"
+        run: npx -y wrangler@4.121.0 pages deploy docs/out --project-name=librefang-docs --branch="$HEAD_REF"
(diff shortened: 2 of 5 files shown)


.github/workflows/build-timings.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         continue-on-error: true
         run: cargo xtask compare-build-timings
       - name: Upload timing snapshot
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
```
houko, line 56: The SHA pin itself checks out: it matches the exact commit already pinned for `actions/upload-artifact@v4` in `coverage.yml:148` and `:157`, so this reuses an already-audited reference rather than introducing a new one.

Separate, out-of-scope note for a maintainer call rather than a blocker: this and `coverage.yml` are the only two `upload-artifact` call sites still on v4 — every other one in the repo (`release.yml`, `release-cli.yml`, `test-web.yml`, `pr-review-collector.yml`) has already moved to the v7 SHA (`043fb46d1a93c77aae656e7c1c64a875d1fc6a0a`). That's a version bump, not a pin-safety issue, so it's reasonable to leave it out of a PR whose scope is "pin what's still mutable," but worth a follow-up so these two don't keep drifting behind the rest of the workflow set.

Added the missing `changelog.d/security/` fragment (matching the precedent this PR's sibling, #6958, set for the same category of change) and pushed it as a separate commit.

---

houko, line 56: This pins to the same v4 commit (`ea165f8d...`) already used by `coverage.yml:148,157`, matching the PR's stated intent. Worth a maintainer call, though: `test-web.yml:68`, `pr-review-collector.yml:47`, `release.yml:535,2854` and `release-cli.yml:92` have already migrated to the newer `actions/upload-artifact@043fb46d...` (v7/v7.0.1). Since this pin-pass touches this exact line anyway, is staying on v4 here intentional (e.g. to keep parity with `coverage.yml` until that's bumped too), or should this land on v7 like the rest of the repo? Not blocking — the SHA/comment pairing itself is correct and consistent with existing precedent either way.

---

PR #6974, opened 2026-08-12
fix(cli): make config writes durable

houko:
## Summary
- add a shared durable atomic-write helper using unique sibling staging files
- preserve existing Unix permissions and create new secret/config files with mode 0600
- atomically replace existing files on Windows with `MoveFileExW`, and sync file/directory metadata where supported
- migrate CLI auth, channel, config, and MCP mutations away from direct truncating writes

## Tests
- `cargo test -p librefang-cli durable_atomic_write --bin librefang`
- `cargo test -p librefang-cli write_chatgpt_secrets --bin librefang`
- `cargo test -p librefang-cli mcp --bin librefang`
- `cargo test -p librefang-cli config --bin librefang`
- `cargo clippy -p librefang-cli --bin librefang --tests -- -D warnings`
- `git diff --check`

### DIFF
--- a/changelog.d/fixed/6974-cli-atomic-config-writes.md
+++ b/changelog.d/fixed/6974-cli-atomic-config-writes.md
@@ -1,0 +1,3 @@
+CLI commands that rewrite `config.toml`, channel configs, MCP server entries, and ChatGPT OAuth secrets used a plain truncating `fs::write`, so a crash or kill mid-write could leave a corrupt or empty file behind.
+These call sites now go through a shared `durable_atomic_write` helper that stages content in a unique sibling file, fsyncs it, and atomically replaces the target via `rename` on Unix or `MoveFileExW` on Windows, fsyncing the parent directory afterward on Unix so the replacement survives a crash.
+New secret files are created at 0600 and an existing file's permissions are now preserved exactly, including bits a restrictive process umask would otherwise silently strip from the creation mode (#6974) (@houko)
--- a/crates/librefang-cli/src/commands/common.rs
+++ b/crates/librefang-cli/src/commands/common.rs
@@ -4,6 +4,141 @@
 
 use crate::commands::prelude::*;
 
+/// Durably replace `path` with `content` using a unique sibling staging file.
+///
+/// The staging file is synced before rename and, on Unix, the containing
+/// directory is synced afterwards. Existing Unix permissions are preserved;
+/// callers may supply a restrictive creation mode for new secret files.
+pub(crate) fn durable_atomic_write(
+    path: &std::path::Path,
+    content: &[u8],
+    create_mode: u32,
+) -> std::io::Result<()> {
+    use std::io::Write as _;
+    use std::sync::atomic::{AtomicU64, Ordering};
+
+    static SEQUENCE: AtomicU64 = AtomicU64::new(0);
+    #[cfg(not(unix))]
+    let _ = create_mode;
+    let sequence = SEQUENCE.fetch_add(1, Ordering::Relaxed);
+    let parent = match path.parent() {
+        Some(parent) if !parent.as_os_str().is_empty() => parent,
+        _ => std::path::Path::new("."),
+    };
+    let file_name = path.file_name().ok_or_else(|| {
+        std::io::Error::new(
+            std::io::ErrorKind::InvalidInput,
+            "target path has no filename",
+        )
+    })?;
+    let mut staging_name = file_name.to_os_string();
+    staging_name.push(format!(".{}.{sequence}.tmp", std::process::id()));
+    let staging = parent.join(staging_name);
+
+    let result = (|| -> std::io::Result<()> {
+        let mut options = std::fs::OpenOptions::new();
+        options.write(true).create_new(true);
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::OpenOptionsExt as _;
+            let existing_mode = std::fs::metadata(path)
+                .ok()
+                .map(|metadata| {
+                    use std::os::unix::fs::PermissionsExt as _;
+                    metadata.permissions().mode()
+                })
+                .unwrap_or(create_mode);
+            options.mode(existing_mode);
+        }
+        let mut file = options.open(&staging)?;
+        file.write_all(content)?;
+        file.sync_all()?;
+        replace_file(&staging, path)?;
+        #[cfg(unix)]
+        std::fs::File::open(parent)?.sync_all()?;
+        Ok(())
+    })();
+
+    if result.is_err() {
+        let _ = std::fs::remove_file(&staging);
+    }
+    result
+}
+
+#[cfg(not(windows))]
+fn replace_file(staging: &std::path::Path, target: &std::path::Path) -> std::io::Result<()> {
+    std::fs::rename(staging, target)
+}
+
+#[cfg(windows)]
+fn replace_file(staging: &std::path::Path, target: &std::path::Path) -> std::io::Result<()> {
+    use std::os::windows::ffi::OsStrExt as _;
+    use windows_sys::Win32::Storage::FileSystem::{
+        MoveFileExW, MOVEFILE_REPLACE_EXISTING, MOVEFILE_WRITE_THROUGH,
+    };
+
+    let staging: Vec<u16> = staging.as_os_str().encode_wide().chain(Some(0)).collect();
+    let target: Vec<u16> = target.as_os_str().encode_wide().chain(Some(0)).collect();
+    // SAFETY: Both paths are represented by valid, NUL-terminated UTF-16 buffers
+    // that remain alive for the duration of the call.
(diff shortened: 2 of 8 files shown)


crates/librefang-cli/src/commands/common.rs
```diff
@@ -4,6 +4,179 @@
...
+
+    if result.is_err() {
+        let _ = std::fs::remove_file(&staging);
+    }
+    result
+}
+
+#[cfg(not(windows))]
+fn replace_file(staging: &std::path::Path, target: &std::path::Path) -> std::io::Result<()> {
+    std::fs::rename(staging, target)
```
houko, line 77: `rename(2)` replaces whatever directory entry currently sits at `target`, including a symlink. If a user manages `config.toml` (or `secrets.env`, channel/MCP configs) as a symlink — e.g. via a dotfile manager like `stow`/`chezmoi`, or a bind-mount workaround — this call silently turns `target` into a plain regular file, and the symlink's original destination is left stale and orphaned. Worth a deliberate call on whether that's acceptable (most atomic-write helpers do accept this trade-off) or whether callers should `readlink` and write through the link target instead. Not blocking, just flagging since none of the tests cover a pre-existing symlink at `path`.

---

crates/librefang-cli/src/commands/common.rs
```diff
@@ -4,6 +4,179 @@
 
 use crate::commands::prelude::*;
 
+/// Durably replace `path` with `content` using a unique sibling staging file.
+///
+/// The staging file is synced before rename and, on Unix, the containing
+/// directory is synced afterwards. Existing Unix permissions are preserved;
+/// callers may supply a restrictive creation mode for new secret files.
+pub(crate) fn durable_atomic_write(
```
houko, line 12: This makes the write itself atomic (no partial/truncated file survives a crash) but doesn't address the read-modify-write race between two concurrent CLI invocations against the same target — e.g. two overlapping `librefang config set` calls, or `config set` racing a `mcp add`/`channel rm` that touches the same `config.toml`. Each reads the old file, mutates its own key, and atomically replaces — so the second writer's atomic replace clobbers the first writer's change with no error surfaced (a lost update, not a corruption). That may be an acceptable scope boundary for this PR (durability vs. multi-writer coordination are different problems), but worth confirming intent, since none of the call sites take any lock around their read+mutate+write sequence and the PR description frames this as a general "durable config writes" fix.

---

houko (commented):
Reviewed the durable-write helper for atomicity, crash-safety, and permission handling. Fixed one mechanical bug directly (permission bits silently clipped by the process umask instead of being preserved exactly) and added the missing changelog fragment; pushed both as a follow-up commit on this branch. Left two comments on scope/design questions that aren't mechanical fixes: symlink-replacement semantics of the rename-based swap, and the read-modify-write race between concurrent CLI invocations touching the same file.

---

crates/librefang-cli/src/commands/common.rs
```diff
@@ -4,6 +4,174 @@
...
```
houko, line 12: Migration looks incomplete relative to the PR's own claim ("CLI commands that rewrite `config.toml` … used a plain truncating `fs::write`"). Two `config.toml` write paths in `commands/init.rs` are not migrated to `durable_atomic_write` and are still exactly the failure mode this PR is fixing:

- `write_config_if_missing` (`init.rs:731`) — the very first `config.toml` a user gets, still `std::fs::write`.
- the config-upgrade rewrite (`init.rs:269`) — rewrites an *existing* `config.toml` in place with plain `fs::write`, so a crash mid-upgrade can still corrupt/empty it (the backup file it makes beforehand only helps if the user knows to look for it).

`commands/everyapi.rs:895` (`write_provider_file`, writing `providers/everyapi.toml`) is the same shape of config write and also untouched.

Per CLAUDE.md's "fix what you found" policy, these feel in-scope for this PR rather than a follow-up, since they're the same call-site pattern already being migrated six times over in this diff. Is leaving them was intentional (e.g. to keep the diff scoped to a specific set of commands), or should they be folded in here?

---

crates/librefang-cli/src/commands/common.rs
```diff
@@ -4,6 +4,174 @@
...
+    })();
+
+    if result.is_err() {
+        let _ = std::fs::remove_file(&staging);
+    }
+    result
+}
+
+#[cfg(not(windows))]
+fn replace_file(staging: &std::path::Path, target: &std::path::Path) -> std::io::Result<()> {
```
houko, line 76: Behavior change worth confirming intentional: `rename(staging, target)` replaces `target` itself, whereas the `std::fs::write` this replaces opens (and follows) the path, writing through a symlink to whatever it points at. If any of `config.toml` / a channel config / an MCP server file / the ChatGPT secrets file is a symlink (e.g. a dotfile-manager setup, or a shared secrets file symlinked in from elsewhere), the first `durable_atomic_write` call now silently turns it into a plain regular file at that path and detaches it from whatever it used to point to — no error, no warning. Given three of these call sites are user-editable config files that people plausibly symlink, might be worth either resolving the symlink target before staging/renaming, or explicitly documenting the change in the changelog fragment.

---

changelog.d/fixed/6974-cli-atomic-config-writes.md
```diff
@@ -0,0 +1,3 @@
+CLI commands that rewrite `config.toml`, channel configs, MCP server entries, and ChatGPT OAuth secrets used a plain truncating `fs::write`, so a crash or kill mid-write could leave a corrupt or empty file behind.
+These call sites now go through a shared `durable_atomic_write` helper that stages content in a unique sibling file, fsyncs it, and atomically replaces the target via `rename` on Unix or `MoveFileExW` on Windows, fsyncing the parent directory afterward on Unix so the replacement survives a crash.
+New secret files are created at 0600 and an existing file's permissions are now preserved exactly, including bits a restrictive process umask would otherwise silently strip from the creation mode (#6974) (@houko)
```
houko, line 3: Minor wording nit: this says "new secret files are created at 0600", but the code passes `0o600` as `create_mode` at every call site, including plain `config.toml` (`config.rs`) and MCP server entries (`mcp_cmds.rs`), not just the ChatGPT secrets file. So a brand-new `config.toml` created via `librefang config set` on a fresh install now comes out owner-only instead of the previous umask-default (typically 0644), which is a real (probably desirable) behavior change beyond "secret files" — might be worth calling out explicitly here so it isn't a surprise to someone reading the changelog for why their `config.toml` permissions changed.

---

houko (commented):
Reviewed the new `durable_atomic_write` helper and its call sites. The core mechanism (create_new staging file + explicit chmod bypassing umask + fsync + rename/MoveFileExW + parent-dir fsync) is sound and the test coverage for permission preservation/umask-bypass is good. Left three comments: an apparent gap in the migration (two `config.toml` write paths in `init.rs` plus one in `everyapi.rs` are still plain `fs::write`, contradicting the PR's own "config writes durable" framing), a symlink-replacement behavior change worth confirming intentional, and a changelog wording nit.

---

crates/librefang-cli/src/commands/config.rs
```diff
@@ -260,7 +260,7 @@ pub(crate) fn cmd_config_set(key: &str, value: &str) {
         std::process::exit(1);
     });
 
-    std::fs::write(&config_path, &serialized).unwrap_or_else(|e| {
+    durable_atomic_write(&config_path, serialized.as_bytes(), 0o600).unwrap_or_else(|e| {
```
houko, line 263: Right after this call, line 270 (and line 348 in `cmd_config_unset`) does `restrict_file_permissions(&config_path)`, which unconditionally chmods the file to 0600. That line predates this PR (unchanged since before the branch point), but its interaction with the new helper is worth a second look: `durable_atomic_write` was just built to preserve a config file's existing mode across the rename (per the changelog fragment: "an existing file's permissions are now preserved exactly"), yet for the two most common config-editing commands — `config set` and `config unset` — that preserved mode is discarded one line later in favor of a hardcoded 0600.

So today, running `librefang config set foo bar` on a `config.toml` that was intentionally, say, 0644 (e.g. shared read access in a multi-user box) silently tightens it to 0600 every time, regardless of what `durable_atomic_write` just preserved.
Contrast with `channel.rs`'s `cmd_channel_rm` and `mcp_cmds.rs`'s `upsert_mcp_server_local`/`remove_mcp_server_local`, migrated in this same PR, which have no such follow-up call and so do get the advertised preserve-permissions behavior.

For `auth.rs`'s `write_chatgpt_secrets` the same pattern is clearly intentional (secrets should always be 0600, per the SECURITY comment there). For plain `config.toml` it's less obviously so.

Is forcing `config.toml` to 0600 on every `set`/`unset` still intended now that `durable_atomic_write` handles permission preservation, or should these two `restrict_file_permissions` calls be dropped so the new preserve-mode contract actually holds for the two commands users hit most?

---

PR #6977, opened 2026-08-12
fix(skills): serialize local installs

houko:
## Summary
- move local skill copying behind the existing cross-process per-skill file lock
- re-check destination existence after acquiring the lock so concurrent installers cannot both proceed
- hold the lock through partial-copy cleanup, preventing a failed loser from deleting the winner's installation
- reuse the locked copier for the API route while retaining symlink filtering and 409 responses

## Tests
- `cargo test -p librefang-skills local_install --lib`
- `cargo test -p librefang-skills evolution::race_tests --lib`
- `cargo test -p librefang-api copy_dir_recursive_skips_symlinks --lib`
- `cargo clippy -p librefang-skills -p librefang-api --lib --tests -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/librefang-api/tests/skills_routes_test.rs
+++ b/crates/librefang-api/tests/skills_routes_test.rs
@@ -619,6 +619,44 @@ async fn skills_install_path_traversal_hand_rejected_400() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+async fn skills_install_already_installed_returns_409() {
+    // Regression coverage for #6977: the destination-exists check moved
+    // from before the per-skill lock to inside
+    // `librefang_skills::evolution::install_local_skill`, and a hit now
+    // surfaces through `SkillError::AlreadyInstalled` instead of a
+    // pre-lock `dest.exists()` probe. Assert the HTTP-visible behavior
+    // (first install succeeds, second is a 409) still holds end-to-end.
+    let h = boot().await;
+    install_registry_skill(h.home(), "dup-skill", "A skill installed twice");
+
+    let (first_status, first_body) = json_request(
+        &h,
+        Method::POST,
+        "/api/skills/install",
+        Some(serde_json::json!({"name": "dup-skill"})),
+    )
+    .await;
+    assert_eq!(first_status, StatusCode::OK, "{first_body:?}");
+
+    let (second_status, second_body) = json_request(
+        &h,
+        Method::POST,
+        "/api/skills/install",
+        Some(serde_json::json!({"name": "dup-skill"})),
+    )
+    .await;
+    assert_eq!(second_status, StatusCode::CONFLICT, "{second_body:?}");
+    assert_eq!(second_body["status"], "already_installed", "{second_body:?}");
+    assert!(
+        second_body["error"]
+            .as_str()
+            .unwrap_or("")
+            .contains("already installed"),
+        "error must mention already-installed: {second_body:?}"
+    );
+}
+
+#[tokio::test(flavor = "multi_thread")]
 async fn skillhub_install_path_traversal_hand_rejected_400() {
     // Skillhub has its own install handler. Reject traversal before the
     // hand directory existence probe so an attacker cannot escape the
--- a/crates/librefang-api/tests/skills_routes_test.rs
+++ b/crates/librefang-api/tests/skills_routes_test.rs
@@ -646,7 +646,10 @@ async fn skills_install_already_installed_returns_409() {
     )
     .await;
     assert_eq!(second_status, StatusCode::CONFLICT, "{second_body:?}");
-    assert_eq!(second_body["status"], "already_installed", "{second_body:?}");
+    assert_eq!(
+        second_body["status"], "already_installed",
+        "{second_body:?}"
+    );
     assert!(
         second_body["error"]
             .as_str()
--- a/crates/librefang-skills/src/evolution.rs
+++ b/crates/librefang-skills/src/evolution.rs
@@ -253,6 +253,51 @@ fn acquire_skill_lock(skill_dir: &Path) -> Result<std::fs::File, SkillError> {
     Ok(lock_file)
 }
 
+/// Install a local skill directory while holding the same per-skill lock used
+/// by evolution and uninstall operations.
+///
+/// The destination existence check is deliberately performed after locking so
+/// two concurrent installers cannot both pass the check and copy into the same
+/// partially populated directory.
+pub fn install_local_skill(source: &Path, skill_dir: &Path) -> Result<(), SkillError> {
+    let _lock = acquire_skill_lock(skill_dir)?;
+    if skill_dir.exists() {
+        let name = skill_dir
+            .file_name()
+            .and_then(|name| name.to_str())
+            .unwrap_or("unknown");
+        return Err(SkillError::AlreadyInstalled(name.to_string()));
+    }
+
+    let result = copy_local_skill_recursive(source, skill_dir);
+    if result.is_err() {
+        let _ = std::fs::remove_dir_all(skill_dir);
+    }
+    result
+}
+
+fn copy_local_skill_recursive(source: &Path, destination: &Path) -> Result<(), SkillError> {
+    std::fs::create_dir_all(destination)?;
+    for entry in std::fs::read_dir(source)? {
+        let entry = entry?;
+        let file_type = entry.file_type()?;
+        if file_type.is_symlink() {
+            tracing::warn!(
+                path = %entry.path().display(),
+                "skipping symlink while installing skill"
+            );
+            continue;
+        }
+        let destination_path = destination.join(entry.file_name());
+        if file_type.is_dir() {
+            copy_local_skill_recursive(&entry.path(), &destination_path)?;
+        } else {
+            std::fs::copy(entry.path(), destination_path)?;
+        }
+    }
+    Ok(())
+}
+
 // ── Atomic file I/O ─────────────────────────────────────────────────
 
 /// Monotonic per-process counter used by `atomic_write` to derive a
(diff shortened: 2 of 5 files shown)


crates/librefang-skills/src/evolution.rs
houko, line 1: The locking here only serializes callers that go through `install_local_skill` (the new HTTP `POST /api/skills/install` path). Two other entry points write into the exact same `skills_dir` tree without ever taking `acquire_skill_lock`:

- `librefang_skills::marketplace::MarketplaceClient::install` / `install_from_registry_dir` / `install_from_git` (`marketplace.rs`) — these still do a bare `if skill_dir.exists() { remove_dir_all(...) }` then `copy_dir_recursive`, no lock at all.
- The CLI's `librefang skill install <name>` (`librefang-cli/src/commands/skill.rs::cmd_skill_install`) calls that same unlocked `MarketplaceClient::install`/`install_from_git` against `resolve_skills_dir(hand)` — the same directory the API route installs into.

So a CLI install (or a marketplace-triggered install) running concurrently with an HTTP `/api/skills/install` call, or with an `evolve`/`uninstall` operation that *does* take this lock, is still racy: the unlocked side can read/overwrite/`remove_dir_all` a skill directory while the locked side believes it holds exclusive access. The PR title says "serialize local installs" — worth confirming whether that's meant to cover only the HTTP route (in which case fine as scoped) or the broader race across all local-install entry points, in which case `marketplace.rs`'s copiers should probably also acquire this lock (or CLI/marketplace installs should be re-pointed at `install_local_skill`). Flagging for a scope call rather than fixing, since routing marketplace.rs through this lock also changes its current overwrite-on-exists semantics.

---

crates/librefang-skills/src/evolution.rs
```diff
@@ -253,6 +253,48 @@ fn acquire_skill_lock(skill_dir: &Path) -> Result<std::fs::File, SkillError> {
     Ok(lock_file)
 }
 
+/// Install a local skill directory while holding the same per-skill lock used by evolution and uninstall operations.
+///
+/// The destination existence check is deliberately performed after locking so two concurrent installers cannot both pass the check and copy into the same partially populated directory.
+pub fn install_local_skill(source: &Path, skill_dir: &Path) -> Result<(), SkillError> {
```
houko, line 259: Lock-coverage gap (CLAUDE.md "fix what you found" / this PR's own stated scope): this PR serializes only the local-registry install path (`POST /api/skills/install` → `install_local_skill` here). Three sibling install handlers still do the exact same unlocked check-then-act the changelog fragment describes as the bug:
- `skillhub_install` (`crates/librefang-api/src/routes/skills/skillhub.rs:250,260`) — `client.is_installed(...)` then `client.install(...).await` with nothing serializing between them.
- `clawhub_install` (`crates/librefang-api/src/routes/skills/clawhub.rs:343`) and `clawhub_cn_install` (`clawhub.rs:734`) — identical pattern.
- The copy step underneath all three, `librefang_skills::marketplace::copy_dir_recursive` (`crates/librefang-skills/src/marketplace.rs:144`), is a separate unlocked duplicate of the function this PR just replaced here in `evolution.rs`.

Not flagging as a required in-place fix, because closing it properly means holding (or emulating) a sync per-skill flock across an `async fn ...install(...).await` network call in the skillhub/clawhub client modules — a real design decision (e.g. `spawn_blocking` scoped around just the FS mutation vs. an async-aware lock), not a mechanical port of this PR's change. Per CLAUDE.md: "would fixing it require touching a different crate or domain than the one you're already in? If no, fix it in this PR. If yes, surface the question to the human reviewer" — this crosses into the skillhub/clawhub async install flow, so surfacing rather than deciding unilaterally to defer it. As merged, the race fix is incomplete relative to its own stated goal (three of four install entry points remain racy); recommend either widening this PR or opening an immediate follow-up against the same issue.

---

crates/librefang-api/tests/skills_routes_test.rs
```diff
@@ -618,6 +618,44 @@ async fn skills_install_path_traversal_hand_rejected_400() {
     );
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn skills_install_already_installed_returns_409() {
```
houko, line 622: Process finding, not a code issue: this test was added in commit `b0705651` ("test(skills): cover already-installed 409 for /api/skills/install"), whose author identity resolves to `Claude <noreply@anthropic.com>`. CLAUDE.md's Git Conventions are explicit that commit *author identity* resolving to Claude/Anthropic is rejected (the repo's `commit-msg` hook checks `GIT_AUTHOR_IDENT` for exactly this, independent of message content), and the "No AI / Claude attribution" rule applies to commit messages, PR bodies, and comments generally.

Leaving this as a comment rather than fixing it myself: correcting authorship needs a rewrite of this commit (`git commit --amend --author=...` or an interactive rebase) followed by a force-push, and force-push is outside what I'm permitted to do in this pass (fast-forward push only). Could a maintainer or the PR author rebase this commit onto the correct identity and force-push? The test content itself is good, correct 409-regression coverage — this flag is purely about the commit's author trailer, not the diff.

---

houko (commented):
Automated review pass. The core fix (moving the destination-exists check behind the per-skill flock in `install_local_skill`, and scoping cleanup so a failed copier never touches another install's files) is correct and well-tested — `local_install_rechecks_destination_after_acquiring_lock` in `evolution.rs` directly exercises the race window with a real concurrent thread, and `skills_install_already_installed_returns_409` covers the HTTP-visible 409 path end-to-end. No deadlock risk from the new lock: `install_skill` doesn't hold any other lock before calling `install_local_skill`, and the function doesn't re-enter the lock on the same path. Changelog fragment, commit format, and prose wrapping all match CLAUDE.md. Scoped `cargo check -p librefang-skills --lib` passed clean (45s, no contention).

Two findings left as inline comments, both left for human judgment rather than fixed directly (see comments for why): (1) the same install-race pattern this PR fixes is still present, unfixed, in three sibling install paths (skillhub, clawhub, clawhub-cn); (2) one commit on this branch (`b0705651`) has an author identity of `Claude <noreply@anthropic.com>`, which CLAUDE.md's commit-msg hook is designed to reject.

---

crates/librefang-api/tests/skills_routes_test.rs
houko, line 1: This is currently the actual cause of the red `CI Gate` / `Quality` checks on this PR (run 31626784881, job `Quality` 94223525112), not a pre-existing/unrelated failure — worth calling out since the other review threads on this PR don't mention it.

`cargo xtask fmt` (i.e. `cargo fmt --check`) fails on this line:

```
assert_eq!(second_body["status"], "already_installed", "{second_body:?}");
```

with the diff the job reports:

```
-    assert_eq!(second_body["status"], "already_installed", "{second_body:?}");
+    assert_eq!(
+        second_body["status"], "already_installed",
+        "{second_body:?}"
+    );
```

i.e. the line is over rustfmt's wrap width and needs to be reformatted (or reflowed) to match. This blocks `CI Gate` (which already reads "one or more lanes failed or were cancelled" from this). A `cargo fmt` pass (or manually wrapping the macro call as rustfmt wants) over `crates/librefang-api/tests/skills_routes_test.rs` should clear it — mechanical fix, no logic change.

---

crates/librefang-skills/src/evolution.rs
houko, line 1: This closes the race for `POST /api/skills/install`, but `librefang_skills::marketplace::MarketplaceClient` has the identical unlocked TOCTOU pattern in three functions that also write into `~/.librefang/skills/<name>` (or a hand's skills dir), and none of them call `acquire_skill_lock`:

- `MarketplaceClient::install` (marketplace.rs:395-482, the GitHub-release fallback leg): `skill_dir.exists()` → `remove_dir_all` → `create_dir_all` → download/extract, no lock.
- `install_from_registry_dir` (marketplace.rs:487-535): same `exists()` → `remove_dir_all` → `create_dir_all` → `copy_dir_recursive` shape, no lock.
- `install_from_git` (marketplace.rs:540-605ish): same shape around the clone/copy, no lock.

These are reached from separate HTTP routes — `POST /api/skillhub/install` (routes/skills/skillhub.rs:260) and `POST /api/clawhub/install` (routes/skills/clawhub.rs:353, 744) — so a skillhub/clawhub install can race another skillhub/clawhub install of the same name, or race the now-fixed `/api/skills/install` / evolve / uninstall paths, and reproduce exactly the failure mode this PR's changelog describes (a loser's `remove_dir_all` deleting a winner's just-installed files) — worse here, since `remove_dir_all` runs unconditionally on every call, not just on copy failure.

Worth confirming whether this is intentionally out of scope (different install surface, tracked separately) or should be folded into this PR — the fix isn't a drop-in swap since these three functions currently *overwrite* an existing `skill_dir` (`remove_dir_all` then reinstall) rather than reject with `AlreadyInstalled`, so wiring in `acquire_skill_lock` means deciding whether to also unify that overwrite-vs-reject semantic difference. Flagging for a maintainer call rather than fixing directly, since a semantic change to the marketplace install contract is not a mechanical fix.

---

crates/librefang-api/tests/skills_routes_test.rs
houko, line 1: Commit b0705651 (`test(skills): cover already-installed 409 for /api/skills/install`, which added this test) has both author and committer identity `Claude <noreply@anthropic.com>`:

```
git show b0705651 -s --format='Author: %an <%ae>%nCommitter: %cn <%ce>'
```

CLAUDE.md's `commit-msg` hook is specifically documented to reject this ("rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean") — this commit shouldn't have been able to land with `core.hooksPath` pointed at `scripts/hooks`, so either the hook wasn't enabled when it was made or it slipped through some other path. The commit message body itself is clean, so this is metadata-only, but it's a real policy violation sitting in the branch's history.

Not fixing this myself: correcting it means rebasing to amend that one commit's author/committer and force-pushing the branch, which is outside what an automated pass should do unilaterally on an already-open PR. Flagging for a maintainer to rebase-and-force-push (`git rebase -i`, reword author via `--committer-date-is-author-date`/`git commit --amend --reset-author` on that commit, or `git filter-branch`/`git-filter-repo` for the single commit) before merge.

---

PR #6978, opened 2026-08-12
fix(dashboard): surface mutation failures

houko:
## Summary
- configure a global React Query `MutationCache` error fallback
- display rejected mutations through the existing localized toast system
- defer to mutation-specific `onError` handlers to avoid duplicate feedback
- extract query client creation for focused regression coverage

## Tests
- `pnpm test -- src/lib/queryClient.test.ts` (95 files, 994 tests)
- `pnpm typecheck`
- `pnpm lint`
- `pnpm build`
- `git diff --check`

### DIFF
--- a/crates/librefang-api/dashboard/src/lib/queryClient.ts
+++ b/crates/librefang-api/dashboard/src/lib/queryClient.ts
@@ -1,0 +1,28 @@
+import { MutationCache, QueryClient } from "@tanstack/react-query";
+import i18n from "./i18n";
+import { toastErr } from "./errors";
+import { useUIStore } from "./store";
+
+export function createDashboardQueryClient(): QueryClient {
+  return new QueryClient({
+    mutationCache: new MutationCache({
+      onError: (error, _variables, _context, mutation) => {
+        // A mutation-specific handler can provide more actionable context and
+        // already owns its UI. This callback is the fallback for bare mutate()
+        // calls that would otherwise fail without any visible feedback.
+        if (mutation.options.onError) return;
+
+        const fallback = i18n.t("common.error", { defaultValue: "Error" });
+        useUIStore.getState().addToast(toastErr(error, fallback), "error");
+      },
+    }),
+    defaultOptions: {
+      queries: {
+        retry: 1,
+        refetchOnWindowFocus: false,
+        staleTime: 30_000,
+        refetchIntervalInBackground: false,
+      },
+    },
+  });
+}
--- a/changelog.d/fixed/6978-dashboard-mutation-error-toasts.md
+++ b/changelog.d/fixed/6978-dashboard-mutation-error-toasts.md
@@ -1,0 +1,2 @@
+Add a global React Query `MutationCache` error fallback so a rejected mutation without its own `onError` handler now surfaces a localized toast instead of failing silently.
+Mutations that already register a mutation-specific `onError` are left untouched to avoid duplicate feedback (#6978) (@houko)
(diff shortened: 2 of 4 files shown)


crates/librefang-api/dashboard/src/lib/queryClient.ts
```diff
@@ -0,0 +1,28 @@
...
+import { useUIStore } from "./store";
+
+export function createDashboardQueryClient(): QueryClient {
+  return new QueryClient({
+    mutationCache: new MutationCache({
+      onError: (error, _variables, _context, mutation) => {
+        // A mutation-specific handler can provide more actionable context and
+        // already owns its UI. This callback is the fallback for bare mutate()
+        // calls that would otherwise fail without any visible feedback.
+        if (mutation.options.onError) return;
```
houko, line 13: `if (mutation.options.onError) return;` assumes a mutation-specific `onError` always shows its own user-facing feedback, but that's not true for every hook already in the tree.

`useRenameTerminalWindow` (`src/lib/mutations/terminal.ts`) sets `onError` purely to roll back the optimistic cache update (`qc.setQueryData(terminalKeys.windows(), context.previous)`) — it never surfaces the failure to the user. With this fallback in place, a failed rename now silently reverts with *no* toast at all, which is the exact "swallowed" failure mode this PR states it's fixing (PR description: "display rejected mutations through the existing localized toast system").

Per `crates/librefang-api/dashboard/AGENTS.md` ("Conventions" section): "Mutation invalidation lives in the hook... Call sites MAY attach per-call `onSuccess`/`onError` handlers for UI feedback... that is orthogonal to invalidation." The AGENTS.md model treats a hook's own `onError` as optionally doing feedback, not necessarily doing it — so gating the fallback purely on "does an `onError` exist" doesn't match that model.

Worth either: (a) having `useRenameTerminalWindow`'s `onError` also show a toast (then the bypass is correct for that case), or (b) reconsidering the bypass condition so a handler that doesn't itself call `addToast`/similar doesn't suppress the global fallback. As written, any future hook that adds a local `onError` for a non-UI purpose (rollback, logging, cleanup) will silently opt out of error surfacing.

---

houko, line 13: `if (mutation.options.onError) return;` only sees the `onError` passed to the hook's own `useMutation({...})` call. It does not see an `onError` passed at call time via `mutate(variables, { onError })` — TanStack Query keeps `mutate()`-time callbacks on the observer, separate from `mutation.options`, and the `MutationCache`-level `onError` always fires regardless of them (confirmed against TanStack Query v5 docs/behavior: global `MutationCache` callbacks "will always be called").

That call-time pattern is exactly what this dashboard's own `AGENTS.md` documents under "Conventions" ("Call sites MAY attach per-call `onSuccess` / `onError` handlers for UI feedback") and it's already used by several shipped mutations whose hook definition has no `onError` of its own, e.g.:
- `ChannelsPage.tsx` `handleReload`/`handleCardConfigure`-adjacent remove flow — `reloadMut.mutate(undefined, { onError: (err) => addToast(...) })` and `removeMut.mutateAsync(name, { onError: (err) => addToast(...) })`, backed by `useReloadChannels`/`useRemoveSidecarConfig` in `src/lib/mutations/channels.ts`, neither of which defines `onError`.
- `Memory/index.tsx` delete flow — `.mutate(id, { onError: (err) => addToast(...) })` backed by `useDeleteMemory` in `src/lib/mutations/memory.ts`, which also has no `onError`.

For all of these, `mutation.options.onError` is `undefined`, so this fallback fires *in addition to* the call-site toast — every failure on these mutations will now show two error toasts instead of one, which is the opposite of the "avoid duplicate feedback" goal stated in the PR description and the changelog fragment.

Since the global handler has no visibility into call-time callbacks, suppressing this properly needs a different signal (e.g., an explicit `mutation.meta.suppressErrorToast` convention set at each such call site) rather than the `mutation.options.onError` check — that's a broader, call-site-touching change worth deciding on deliberately rather than guessing at here.

---

PR #6979, opened 2026-08-12
fix(kernel): report cron shutdown persistence failures

houko:
## Summary
- stop silently discarding the cron scheduler's final persistence error during shutdown
- emit a structured warning with the underlying I/O error while allowing shutdown to continue

## Tests
- `git diff --check`

### DIFF
--- a/crates/librefang-kernel/src/kernel/cron_tick.rs
+++ b/crates/librefang-kernel/src/kernel/cron_tick.rs
@@ -41,8 +41,11 @@ pub(super) async fn run_cron_scheduler_loop(kernel: Arc<LibreFangKernel>) {
     loop {
         interval.tick().await;
         if kernel.agents.supervisor.is_shutting_down() {
-            // Persist on shutdown
-            let _ = kernel.workflows.cron_scheduler.persist();
+            // This is the scheduler's last chance to flush execution state.
+            // Keep shutdown moving, but never hide the durability failure.
+            if let Err(error) = kernel.workflows.cron_scheduler.persist() {
+                warn!(%error, "failed to persist cron scheduler during shutdown");
+            }
             break;
         }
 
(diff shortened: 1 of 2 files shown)


crates/librefang-kernel/src/kernel/cron_tick.rs
```diff
@@ -41,8 +41,11 @@ pub(super) async fn run_cron_scheduler_loop(kernel: Arc<LibreFangKernel>) {
     loop {
         interval.tick().await;
         if kernel.agents.supervisor.is_shutting_down() {
-            // Persist on shutdown
-            let _ = kernel.workflows.cron_scheduler.persist();
+            // This is the scheduler's last chance to flush execution state.
+            // Keep shutdown moving, but never hide the durability failure.
+            if let Err(error) = kernel.workflows.cron_scheduler.persist() {
+                warn!(%error, "failed to persist cron scheduler during shutdown");
```
houko, line 47: Fix logic itself looks correct: `cargo check -p librefang-kernel --lib` and `cargo clippy -p librefang-kernel --all-targets -- -D warnings` are both clean, `%error` works fine since `LibreFangError` derives `thiserror::Error` (Display), and `persist()`'s failure path is already covered by `persist_returns_error_when_data_dir_unwritable` in `cron.rs`.

One gap worth a maintainer call: nothing exercises that `run_cron_scheduler_loop` actually surfaces the failure via `warn!` now instead of swallowing it — which is the whole point of this PR. The existing `#[cfg(test)] mod tests` in this file explicitly notes it only reproduces the lock/CAS pattern structurally and defers full-loop coverage to integration tests, and I don't see an existing integration test that drives the shutdown branch of this loop (`crates/librefang-kernel/tests/` has no cron-shutdown test). `librefang_testing::MockKernelBuilder` (used in `audit_retention_test.rs`) plus the `capture_logs` helper already in `cron.rs` (used for `warn_missed_fires`) look like the right building blocks — force an unwritable cron persist dir, flip the supervisor into shutting-down state, and assert the log line appears — but wiring that up correctly (particularly triggering `is_shutting_down()`) needs someone who knows the supervisor API better than I could verify from the diff alone. Worth adding, or explicitly deferring with a reason, before merge.

---

PR #6981, opened 2026-08-12
perf(api): offload agent file mutations

houko:
## Summary\n- move identity-file path resolution and read/write/delete filesystem operations onto the blocking pool\n- replace the fixed temporary filename with the API's durable atomic writer\n- preserve current-layout preference, legacy root fallback, whitelist checks, and containment enforcement\n- add round-trip and symlink-escape regression coverage\n\n## Tests\n- 
running 2 tests
test routes::agents::files::identity_file_io_tests::identity_file_helpers_reject_symlink_escape ... ok
test routes::agents::files::identity_file_io_tests::identity_file_helpers_round_trip_current_layout ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 946 filtered out; finished in 0.13s\n- \n- \n- \n\n## Notes\n- based on the same main revision as #6980; the changes are independent but both touch , so this may need a straightforward rebase after that PR lands\n


crates/librefang-api/src/routes/agents/files.rs
```diff
@@ -283,58 +381,43 @@ pub async fn set_agent_file(
...
+                Json(serde_json::json!({"error": t.t("api-error-file-path-traversal")})),
+            );
+        }
+        Ok(Err(IdentityFileError::Io(error))) => {
+            return (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(serde_json::json!({"error": scrub_500(&error, &t)})),
+            );
+        }
+        Ok(Err(IdentityFileError::NotFound)) => unreachable!(),
```
houko, line 447: `unreachable!()` is correct today — `write_identity_file` never constructs `IdentityFileError::NotFound` — but it's a latent panic-in-request-handler trap: if `IdentityFileError` is ever reused/extended for a future write-path variant that does produce `NotFound` (or someone refactors `write_identity_file` to call `resolve_identity_file`), this line turns a request error into a full handler panic instead of a 500. Given this same enum is shared across read/write/delete, worth either splitting a narrower error type for the write path, or replacing this arm with a scrubbed 500 (`ApiErrorResponse::internal_scrub`) so the failure mode degrades gracefully instead of panicking. Not blocking, but flagging since panics in axum handlers take down just that request but are easy to miss in review once the enum grows.

---

houko (commented):
Automated review pass against CLAUDE.md checklist. The refactor's core logic (path resolution, `.identity/` current-layout preference + legacy fallback, whitelist, canonicalized containment check) looks correctly preserved and genuinely improves on the original: the write/delete handlers previously ran synchronous `std::fs` calls directly on the async task (a real executor-blocking bug), and path-resolution `canonicalize()` calls for the read path were not previously covered by the `tokio::fs` offload either — both are now properly under `spawn_blocking`. `ErrorTranslator`'s `!Send`-before-`.await` gotcha is handled correctly in all three handlers. One missing-changelog-fragment issue was fixed directly (see commit); one inline nit left below for maintainer judgment.

---

crates/librefang-api/src/routes/agents/files.rs
```diff
@@ -283,58 +381,43 @@ pub async fn set_agent_file(
...
 
-    // Security: verify workspace path and target stays inside it
-    let ws_canonical = match workspace.canonicalize() {
-        Ok(p) => p,
-        Err(_) => {
+    let size_bytes = req.content.len();
+    let content = req.content.into_bytes();
+    let filename_for_task = filename.clone();
+    drop(t);
+    let result = tokio::task::spawn_blocking(move || {
```
houko, line 422: This changes the write behavior of an existing route (fixed-name temp file → shared `atomic_write`, closing a same-name concurrent-writer race) — similar in kind to #6977's install-race fix, which added both a unit-level race test and an HTTP-level `#[tokio::test]` (`skills_install_already_installed_returns_409`). This PR's coverage is unit-level only (`identity_file_io_tests`); there's no `TestServer` test hitting `PUT /api/agents/{id}/files/{filename}` that exercises the fixed race end-to-end. Worth confirming existing coverage or adding one per CLAUDE.md's mandatory integration-testing rule.

---

crates/librefang-api/src/routes/agents/files.rs
```diff
@@ -1,5 +1,121 @@
...
+    let workspace = workspace
+        .canonicalize()
+        .map_err(|_| IdentityFileError::Workspace)?;
+    let identity_dir = workspace.join(".identity");
+    std::fs::create_dir_all(&identity_dir).map_err(IdentityFileError::Io)?;
+    let identity_dir = identity_dir.canonicalize().map_err(IdentityFileError::Io)?;
+    if !identity_dir.starts_with(&workspace) {
+        return Err(IdentityFileError::Traversal);
+    }
+    crate::atomic_write(&identity_dir.join(filename), content).map_err(IdentityFileError::Io)
```
houko, line 55: Possible cross-crate race, flagging for human judgment rather than fixing unilaterally since the other side lives in `librefang-kernel` (CLAUDE.md's "fix what you found" bar is same-crate; this spans two).

Per CLAUDE.md's "Agent workspace layout" note: `migrate_identity_files()` runs on every agent spawn and does `std::fs::rename(root/<name>, .identity/<name>)` when `.identity/<name>` doesn't yet exist (`workspace_setup.rs:595`). This handler now does its write off-thread via `spawn_blocking`, landing on `atomic_write(identity_dir.join(filename), ...)` here.

If a respawn's `migrate_identity_files()` and an in-flight `PUT /api/agents/{id}/files/{filename}` race on the same `.identity/<filename>` destination, both do an independent create-temp+`rename` onto that same path with no shared lock — whichever `rename` lands last wins, so the other write is silently dropped. This isn't new to this PR (the previous synchronous write had the same unguarded destination), but moving the write onto the blocking pool changes its scheduling relative to migration without adding any synchronization, and the changelog fragment's framing ("removing a same-name race between concurrent writers") implies same-name-write races were the target — this one is different (cross-crate, cross-operation) and isn't addressed.

Worth a maintainer call: is this an accepted pre-existing gap, or should there be a lock keyed on `(agent, workspace)` shared between spawn-time migration and the files routes?

---

houko (commented):
Automated daily review pass. Checked against CLAUDE.md (changelog fragment format, commit conventions, no-AI-attribution, prose wrapping, integration-testing mandate, spawn_blocking error propagation). The diff's error handling, atomic-write reuse, traversal checks, and changelog fragment all look correct and consistent with existing conventions; no doc-comment prose-wrap issue (no new multi-line comments were added, unlike sibling PRs #6992/#7038). One design question left as an inline comment for maintainer judgment (cross-crate migration/write race) rather than a unilateral fix.

---

houko:
This looks superseded by #7060, which merged as part of the current batch — flagging rather than closing, in case there is something here that one does not cover.

The two carry the same title (`perf(api): offload agent file mutations`), touch the same file (`crates/librefang-api/src/routes/agents/files.rs`), and pursue the same goal. #7060 was opened 2026-08-13 01:20, about twelve hours after this one (2026-08-12 13:38), so it reads as a second pass at the same work rather than an independent change.

On the merged result, `files.rs` now carries 24 occurrences of `spawn_blocking` / `IdentityFileMutationError`; this branch adds 4 `spawn_blocking` call sites. #7060 also went further in a way worth keeping: it lifted the write / rename / canonicalize / delete paths into a typed `IdentityFileMutationError` (`Workspace` / `NotFound` / `Forbidden` / `Io`) and carried the containment check across intact — `canonicalize` and `ws_canonical` appear 4 times on each side of that diff, `starts_with` twice, with both `canonical.starts_with(&ws_canonical)` guards preserved and their failure mapped to `Forbidden`.

That is also why this branch now reports `CONFLICTING`: the file it edits was rewritten underneath it by the very change it duplicates.

Unless something here is missing from #7060, closing seems right. If part of this diff is not covered — I compared the offload surface and the containment guards, not every line — say which part and it can be rebased down to just that.

Worth noting for its own sake: two branches doing the same work half a day apart is easy to miss at this queue depth, and each one costs a full CI run on a pool that is currently the binding constraint.

houko:
Withdrawing the suggestion to close — the rebase changed what this branch is, and my earlier comment was written against the version before it.

What it was: 4 added `spawn_blocking` call sites in `files.rs`, which #7060 covered. What it is now: +196/-111 covering `GET` / `PUT` / `DELETE /api/agents/{id}/files/{filename}`, i.e. including the **read** path that #7060 did not touch, plus a change #7060 did not make — the write path now goes through the shared `atomic_write` helper instead of a fixed temporary filename, which removes a same-name race between concurrent writers. That is a real addition rather than a duplicate, so it should land.

Two things before it can:

**It conflicts again.** `files.rs` was rewritten under it by #7060 while this sat in the queue. Same situation as last time, different cause — needs another rebase onto current `main`.

**The containment guards come out one short on each count, worth confirming while rebasing.** Comparing the two sides of the diff:

```
canonicalize   removed 6  added 5
starts_with    removed 3  added 2
```

The added side does carry both guards I would expect —

```rust
if !canonical.starts_with(&workspace) { return Err(IdentityFileError::Traversal); }
if !identity_dir.starts_with(&workspace) { return Err(IdentityFileError::Traversal); }
```

— and consolidating three scattered checks into `resolve_identity_file` would legitimately reduce the count. But the arithmetic does not by itself distinguish "consolidated" from "one path no longer checked", and the guard being dropped here is what keeps a filename from resolving outside the agent workspace. Worth a direct look at whether every one of `GET` / `PUT` / `DELETE` still reaches a `starts_with` check on the canonicalized path, rather than inferring it from the refactor.

The `IdentityFileError` enum with an explicit `Traversal` variant is a good shape for this — it makes the refusal path typed rather than an ad-hoc 403, and the 8 occurrences suggest it is threaded through all three handlers.

houko:
Closing as superseded by merged #7060 and #7084. After reconciling this branch with current main, the source diff is empty; only the PR-specific changelog fragment remains. The off-thread read, write, and delete paths, atomic-write race fix, traversal guards, and route coverage are now present on main.

PR #6984, opened 2026-08-12
fix(api): durably persist registry content

houko:
## Summary\n- serialize registry content creation/overwrite through the shared write lock\n- move directory creation, durable atomic replacement, secret persistence, catalog loading, and rollback off async runtime workers\n- make the no-overwrite check part of the locked filesystem transaction\n- restore the prior provider file when an invalid overwrite is rejected\n- scrub filesystem errors from API responses while retaining detailed server logs\n- add durable write, conflict, cleanup, and overwrite-rollback regressions\n\n## Tests\n- 
running 1 test
test routes::registry::registry_content_write_tests::creates_and_atomically_replaces_without_staging_residue ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 946 filtered out; finished in 0.09s\n- 
running 6 tests
test registry_content_rejects_malformed_identifiers ... ok
test registry_content_path_is_relative_not_absolute ... ok
test provider_rejected_by_catalog_returns_400_and_leaves_no_file ... ok
test registry_content_accepts_dotted_and_dashed_identifiers ... ok
test rejected_provider_overwrite_restores_previous_file ... ok
test provider_with_reasoning_tier_model_loads_into_catalog ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 9.63s\n- 
running 1 test
test rejected_provider_overwrite_restores_previous_file ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 5 filtered out; finished in 8.55s\n- \n- \n- \n- \n

### DIFF
--- a/crates/librefang-api/src/routes/registry.rs
+++ b/crates/librefang-api/src/routes/registry.rs
@@ -87,6 +87,42 @@ async fn registry_schema_by_type(
 /// identifier cannot overflow a path-component limit on any target filesystem.
 const MAX_REGISTRY_IDENTIFIER_LEN: usize = 128;
 
+#[derive(Debug)]
+enum RegistryContentWriteError {
+    AlreadyExists,
+    Io(std::io::Error),
+}
+
+fn persist_registry_content(
+    target: &std::path::Path,
+    content: &[u8],
+    allow_overwrite: bool,
+) -> Result<Option<Vec<u8>>, RegistryContentWriteError> {
+    if !allow_overwrite && target.exists() {
+        return Err(RegistryContentWriteError::AlreadyExists);
+    }
+    let previous = if target.exists() {
+        Some(std::fs::read(target).map_err(RegistryContentWriteError::Io)?)
+    } else {
+        None
+    };
+    if let Some(parent) = target.parent() {
+        std::fs::create_dir_all(parent).map_err(RegistryContentWriteError::Io)?;
+    }
+    crate::atomic_write(target, content).map_err(RegistryContentWriteError::Io)?;
+    Ok(previous)
+}
+
+fn rollback_registry_content(
+    target: &std::path::Path,
+    previous: Option<&[u8]>,
+) -> std::io::Result<()> {
+    match previous {
+        Some(content) => crate::atomic_write(target, content),
+        None => std::fs::remove_file(target),
+    }
+}
+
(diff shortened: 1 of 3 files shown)


crates/librefang-api/src/routes/registry.rs
```diff
@@ -282,19 +293,52 @@ async fn create_registry_content(
...
 
-    // Create parent directories and write file
-    if let Some(parent) = target.parent() {
-        if let Err(e) = std::fs::create_dir_all(parent) {
-            return ApiErrorResponse::internal(e.to_string())
+    // Serialize the no-overwrite check with the durable replacement. Without
+    // the shared guard, two concurrent POSTs can both observe an absent target
+    // and silently overwrite one another. The fsync-based writer is blocking,
+    // so perform the filesystem transaction off the async runtime worker.
+    let _write_guard = state.config_write_lock.lock().await;
```
houko, line 300: `_write_guard` is not dropped until the end of the function, so it stays held across the provider-only tail: `write_secret_env` (spawn_blocking), `set_env_var_guarded().await`, the `model_catalog_update` spawn_blocking, `clear_driver_cache()`, and the rollback spawn_blocking on rejection.
`config_write_lock` is process-wide — it also gates `POST /api/config/set` (`config/manage.rs`), the sidecar channel configure handlers, `users.rs`, and `budget.rs` — so for the whole duration of a provider create/update, all of those unrelated config-mutating endpoints block too.
The existing convention elsewhere in this file's sibling handlers (see `channels.rs` around its two `config_write_lock` sites, and the comment "The guard is dropped before `reload_config().await` so the hot-reload step does not gate other config-writing handlers") is to scope the guard tightly around just the read-check-write that needs atomicity, then drop it before slower/independent follow-up work.

Was holding the lock across the catalog merge and rollback intentional here (e.g. to also serialize concurrent provider POSTs against catalog state), or should `_write_guard` be dropped right after `persist_registry_content` returns (i.e., once the TOCTOU-sensitive existence-check-and-write is done), before the provider-specific catalog/secrets work begins? `model_catalog_update` already serializes concurrent catalog mutations internally via its own RCU/CAS loop, so it's not obviously relying on this mutex for correctness.

---

PR #6985, opened 2026-08-12
perf(api): move user config I/O off async workers

houko:
## Summary
- read and back up `config.toml` through Tokio filesystem APIs in user mutations
- run the durable atomic config write on the blocking pool
- preserve the existing config and API-key lock ordering and corrupt-config protection

## Tests
- `cargo test -p librefang-api --test users_test`

### DIFF
--- a/changelog.d/changed/6985-user-config-io-off-thread.md
+++ b/changelog.d/changed/6985-user-config-io-off-thread.md
@@ -1,0 +1,3 @@
+User-management writes to `config.toml` (create/update/delete user, key rotation, provider-key changes) previously read, backed up, and wrote the file with blocking `std::fs` calls directly on the async request-handling task.
+Under load this could stall the Tokio worker thread the request landed on for the duration of the disk I/O, delaying unrelated requests scheduled on the same worker.
+The read and backup steps now go through `tokio::fs`, and the durable atomic write moves onto the blocking thread pool via `spawn_blocking`, with the existing config/API-key lock ordering and corrupt-config protection unchanged (#6985) (@houko)
--- a/crates/librefang-api/src/routes/users.rs
+++ b/crates/librefang-api/src/routes/users.rs
@@ -1206,12 +1206,14 @@ where
     // taint rules, etc.) on the next write. The caller's
     // `backups/config.toml.prev` from the previous successful write is the
     // recovery point of last resort.
-    let raw = if config_path.exists() {
-        std::fs::read_to_string(&config_path).map_err(|e| {
-            PersistError::Internal(format!("could not read existing config.toml: {e}"))
-        })?
-    } else {
-        String::new()
+    let raw = match tokio::fs::read_to_string(&config_path).await {
+        Ok(raw) => raw,
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => String::new(),
+        Err(error) => {
+            return Err(PersistError::Internal(format!(
+                "could not read existing config.toml: {error}"
+            )))
+        }
     };
     // Parse with `toml_edit` so we preserve comments / formatting / unrelated
     // sections. A parse failure means the on-disk file is already corrupt;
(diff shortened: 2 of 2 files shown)


changelog.d/changed/6985-user-config-io-off-thread.md
```diff
@@ -0,0 +1,3 @@
+User-management writes to `config.toml` (create/update/delete user, key rotation, provider-key changes) previously read, backed up, and wrote the file with blocking `std::fs` calls directly on the async request-handling task.
+Under load this could stall the Tokio worker thread the request landed on for the duration of the disk I/O, delaying unrelated requests scheduled on the same worker.
+The read and backup steps now go through `tokio::fs`, and the durable atomic write moves onto the blocking thread pool via `spawn_blocking`, with the existing config/API-key lock ordering and corrupt-config protection unchanged (#6985) (@houko)
```
houko, line 3: This fragment is correctly formatted, but the commit that adds it (`d5d67f1`) has git author *and* committer identity `Claude <noreply@anthropic.com>`.

CLAUDE.md's `commit-msg` hook is explicitly documented to reject "a commit whose author identity (`git var GIT_AUTHOR_IDENT` …) resolves to Claude / Anthropic even when the message itself is clean." This commit's message body is clean, but the identity itself is exactly what that rule targets — it must have been pushed with `core.hooksPath` unset or bypassed, since a configured clone would refuse it.

Since fixing this needs a history rewrite (amend + force-push) and I'm not authorized to force-push, flagging for a maintainer to rebase this commit onto a human/bot identity before merge.

---

crates/librefang-api/src/routes/users.rs
```diff
@@ -1262,11 +1264,13 @@ where
...
 
-    if config_path.exists() {
-        if let Some(home_dir) = config_path.parent() {
-            let backups_dir = home_dir.join("backups");
-            if std::fs::create_dir_all(&backups_dir).is_ok() {
-                let _ = std::fs::copy(&config_path, backups_dir.join("config.toml.prev"));
+    if let Some(home_dir) = config_path.parent() {
+        let backups_dir = home_dir.join("backups");
+        if tokio::fs::create_dir_all(&backups_dir).await.is_ok() {
+            match tokio::fs::copy(&config_path, backups_dir.join("config.toml.prev")).await {
```
houko, line 1270: Minor behavior change worth a second look (low confidence, non-blocking): the old code gated this whole block on `config_path.exists()`, so `backups/` was only created once a `config.toml` already existed. That gate is gone now, so on a fresh install (no `config.toml` yet) this unconditionally creates the `backups/` directory before there's anything to back up — the subsequent `copy` NotFound is swallowed correctly, but the directory now appears earlier than before.

Probably harmless, but wanted to confirm it's intentional rather than an incidental drop during the tokio::fs conversion.

---

houko (commented):
Reviewed the tokio::fs / spawn_blocking conversion in persist_users. The core change looks correct: config_write_lock is held across the whole function so the removed exists() checks don't introduce a TOCTOU race, atomic_write is fully synchronous so it's safe inside spawn_blocking, and the changelog fragment is present and correctly formatted. Two things flagged inline for a maintainer look — one about the AI-authored commit identity already on this branch, one a minor/low-confidence question about the backups-dir creation timing.

---

changelog.d/changed/6985-user-config-io-off-thread.md
houko: This appears resolved already: commit `d5d67f1` referenced above is no longer reachable on this branch.
The current tip (`origin/perf/user-config-io-off-thread`) carries only two commits ahead of `main`, `2b1b35b` (perf: move user config I/O off async workers) and `22ee905` (chore: add changelog fragment), and both show clean author/committer identity `Evan <tonymo2048@gmail.com>` (`git log origin/main..HEAD --format='%H %an <%ae> / %cn <%ce>'`).
Looks like it was fixed via rebase after this comment was posted; leaving the thread for a maintainer to resolve since I'm not verifying who pushed the rewrite.

---

PR #6986, opened 2026-08-12
perf(api): move session list queries off async workers

houko:
## Summary
- execute session count and paginated SQLite reads together on Tokio's blocking pool
- preserve the existing pagination envelope and fallback response
- log SQLite and blocking-task failures server-side instead of silently discarding them

## Tests
- `cargo test -p librefang-api --test system_tools_sessions_test`
- `cargo test -p librefang-api --test system_tools_sessions_test list_sessions`
- `CARGO_TARGET_DIR=<system-temp> cargo clippy -p librefang-api --lib -- -D warnings`

### DIFF
--- a/crates/librefang-api/src/routes/tools_sessions.rs
+++ b/crates/librefang-api/src/routes/tools_sessions.rs
@@ -402,20 +402,36 @@ pub async fn list_sessions(
 ) -> impl IntoResponse {
     let offset = pagination.effective_offset();
     let limit = pagination.effective_limit();
-    let substrate = state.kernel.memory_substrate();
-    // Push pagination into SQLite so we don't deserialize every session blob (#3485).
-    let total = substrate.count_sessions().unwrap_or(0);
     // Snapshot of in-flight session IDs from the kernel runtime — the
     // SQLite substrate has no view into liveness, so we merge it in here
     // (#4290). Taken once per request before the SQL call so every row
     // sees a consistent view.
     let running = state.kernel.running_session_ids();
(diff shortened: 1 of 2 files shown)


crates/librefang-api/src/routes/tools_sessions.rs
```diff
@@ -402,20 +402,36 @@ pub async fn list_sessions(
...
     // Snapshot of in-flight session IDs from the kernel runtime — the
     // SQLite substrate has no view into liveness, so we merge it in here
     // (#4290). Taken once per request before the SQL call so every row
     // sees a consistent view.
     let running = state.kernel.running_session_ids();
+    let substrate = Arc::clone(state.kernel.memory_substrate());
+    // Both calls use synchronous SQLite connections. Keep them together on
+    // the blocking pool so a large or contended sessions table cannot stall
+    // the Tokio worker serving unrelated requests.
+    let query_result = tokio::task::spawn_blocking(move || {
```
houko, line 414: `search_sessions` (around line 901, `search_sessions_paginated`) runs the same kind of synchronous SQLite call — an FTS5 query, arguably heavier than the row scan this PR moves off-thread — directly on the async handler, with no `spawn_blocking` and no error logging on the `Err` branch.
Was this left out deliberately (e.g. a separate PR already planned), or should it get the same treatment here since it's the same file and the same root cause?
Flagging rather than fixing inline because extending the pattern also means deciding on error-handling/response-shape parity with the `Err` arm added here, which felt like it deserved author sign-off rather than a silent scope expansion of a perf-only PR.

---

PR #6988, opened 2026-08-12
fix(api): serialize quick init config writes

houko:
## Summary
- use asynchronous metadata for the quick already-initialized path
- serialize quick-init creation with other config mutations and recheck existence under the lock
- run directory creation and durable atomic config writing on Tokio's blocking pool

## Tests
- `cargo test -p librefang-api quick_init_write_is_create_once_and_preserves_existing_config --lib`


crates/librefang-api/src/routes/config/system.rs
```diff
@@ -135,18 +129,48 @@ api_key_env = "{api_key_env}"
...
-            StatusCode::INTERNAL_SERVER_ERROR,
-            Json(serde_json::json!({
-                "status": "error",
-                "message": "Internal server error"
-            })),
-        )
+    if let Some(locked) = crate::routes::guard_config_write() {
+        return locked.into_response();
+    }
+    let _config_guard = state.config_write_lock.lock().await;
```
houko, line 135: This adds a unit test for `write_quick_init_config` (good), but per CLAUDE.md's integration-testing mandate this is still a change to a route handler's concurrency/wiring behavior, and there's no `#[tokio::test]` against `TestServer` hitting `POST /api/init` twice concurrently (or sequentially) to assert the second call returns `already_initialized` and the config content from the first call survives. Worth adding one in `crates/librefang-api/tests/` alongside the existing config route tests, since that's the exact race this PR is closing.

Also added the missing `changelog.d/fixed/` fragment for this PR directly (pushed as a follow-up commit) since it was otherwise absent.

---

houko, line 135: Following up on the earlier comment on this line asking for a `#[tokio::test]` against `TestServer` for `POST /api/init`: two things worth surfacing before adding one, since they explain why the route was excluded rather than just leaving a gap.

1. `quick_init` calls `crate::openrouter_catalog::refresh_if_missing(&state.kernel)` at line 109, which issues a live HTTP request to the OpenRouter API whenever the model catalog hasn't been refreshed yet (see `refresh_now` in `openrouter_catalog.rs`). That matches `config_routes_integration.rs`'s header note: `POST /api/shutdown / /api/init — would tear down the harness kernel` is listed as an intentional exclusion, and the network call is a concrete reason a naive test here would be flaky/slow in CI, not just kernel teardown.
2. The existing precedent for testing an equally stateful route (`POST /api/shutdown`) is a dedicated file with its own `librefang_testing::TestAppState` instance and a real spawned `axum::serve` (`daemon_lifecycle_test.rs`), not the shared `RouterHarness` in `config_routes_integration.rs`.

A `quick_init_test.rs` following that isolated-harness pattern could exercise the create-once / `already_initialized` / lock-serialization behavior this PR adds, but it needs a plan for the catalog-refresh call first (e.g. asserting `needs_initial_refresh()` is already false on the mock kernel, or stubbing the OpenRouter base URL) so the test doesn't make a real outbound request. Flagging for a maintainer call on the mocking approach rather than adding a test that could flake CI.

---

PR #6989, opened 2026-08-12
perf(api): offload dashboard database probes

houko:
## Summary
- run the dashboard health query and session count together on Tokio's blocking pool
- preserve the existing degraded-health and zero-count fallback semantics
- log blocking-task failures instead of silently collapsing them

## Tests
- `cargo test -p librefang-api --test config_status_session_count_test dashboard_snapshot_session_count_matches_substrate_after_seeding`

### DIFF
--- a/crates/librefang-api/src/routes/config/mod.rs
+++ b/crates/librefang-api/src/routes/config/mod.rs
@@ -1043,11 +1043,25 @@ async fn dashboard_snapshot_compute(state: &Arc<AppState>) -> serde_json::Value 
     let shared_id = librefang_types::agent::AgentId(uuid::Uuid::from_bytes([
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
     ]));
-    let db_ok = state
-        .kernel
-        .memory_substrate()
-        .structured_get(shared_id, "__health_check__")
-        .is_ok();
+    let substrate = Arc::clone(state.kernel.memory_substrate());
+    // The memory substrate exposes synchronous SQLite operations. Run the
+    // health probe and session count together on the blocking pool so the
+    // dashboard's frequent snapshot poll cannot stall a Tokio worker.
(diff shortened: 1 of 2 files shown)


crates/librefang-api/src/routes/config/mod.rs
```diff
@@ -1043,11 +1043,25 @@ async fn dashboard_snapshot_compute(state: &Arc<AppState>) -> serde_json::Value
     let shared_id = librefang_types::agent::AgentId(uuid::Uuid::from_bytes([
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
     ]));
-    let db_ok = state
-        .kernel
-        .memory_substrate()
-        .structured_get(shared_id, "__health_check__")
-        .is_ok();
+    let substrate = Arc::clone(state.kernel.memory_substrate());
```
houko, line 1046: This offloads the health probe + session count for `/api/dashboard/snapshot` only. The identical synchronous calls remain inline on the async worker in two other places that this PR's own changelog rationale ("a slow disk could previously stall the worker thread") applies to just as much:

- `routes/config/system.rs:48-52` (`pub async fn status`) — calls `memory_substrate().count_sessions()` directly, and per the comment at line 44 of that file, "the dashboard hammers this route on its 5 s status poll" — the same 5 s cadence cited here.
- `routes/config/system.rs:254-263` (`database_probe_ok`), shared by `/api/health`, `/api/health/detail`, and `/api/ready` — same inline `structured_get(shared_id, "__health_check__")` this PR moves to `spawn_blocking` for the dashboard-snapshot path.

If `/api/status` is still live traffic (vs. superseded by `/api/dashboard/snapshot`), the fix here is incomplete relative to its own stated motivation. Worth confirming whether `/api/status` and `database_probe_ok` should get the same `spawn_blocking` treatment in this PR or a fast follow-up.

---

houko (commented):
Automated daily review pass over the dashboard-db-probes-off-thread change.

---

PR #6992, opened 2026-08-12
perf(api): offload sidecar configuration writes

houko:
## Summary
- run sidecar include detection, secrets reads/writes, and config rewrites on Tokio's blocking pool
- keep the existing config write lock across the complete disk operation
- move include conflict detection under that lock to close the check/write race
- preserve conflict and internal-error responses

## Tests
- `cargo fmt --check`
- `cargo test -p librefang-api sidecar_configuration_write_tests --lib`
- `cargo test -p librefang-api --test sidecar_toml_test`

### DIFF
--- a/changelog.d/changed/6992-sidecar-configure-io-off-thread.md
+++ b/changelog.d/changed/6992-sidecar-configure-io-off-thread.md
@@ -1,0 +1,4 @@
+`POST /api/channels/sidecar/{name}/configure` ran the `include`-shadow check, the `secrets.env` membership read, secret/config writes, and the `config.toml` upsert synchronously on the async worker thread handling the request.
+  All of that now runs as a single `spawn_blocking` task on Tokio's blocking pool, still serialized under the same `config_write_lock` that gates `POST /api/config/set` and the legacy `configure_channel` handler.
+  Moving the `include`-shadow check inside that lock (previously it ran before the lock was taken) closes a check/write race where a concurrent writer could add a conflicting `include` between the check and the write.
+  Conflict and internal-error responses are unchanged; a join failure on the blocking task is now caught and returned as a scrubbed internal error instead of propagating as an unhandled panic (#6992) (@houko)
--- a/crates/librefang-api/src/routes/channels.rs
+++ b/crates/librefang-api/src/routes/channels.rs
@@ -855,6 +855,95 @@ fn included_files_with_sidecars(config_path: &std::path::Path) -> Vec<std::path:
     hits
 }
 
+#[derive(Debug)]
+enum ConfigureSidecarWriteError {
+    IncludedSidecars(Vec<std::path::PathBuf>),
+    Write(String),
+}
+
+fn write_sidecar_configuration(
+    config_path: &std::path::Path,
+    secrets_path: &std::path::Path,
+    entry: &SidecarCatalogEntry,
+    schema: &SidecarSchema,
+    values: &HashMap<String, String>,
+) -> Result<Vec<String>, ConfigureSidecarWriteError> {
+    let shadowing = included_files_with_sidecars(config_path);
+    if !shadowing.is_empty() {
+        return Err(ConfigureSidecarWriteError::IncludedSidecars(shadowing));
+    }
+
+    let secrets_env_keys: std::collections::HashSet<String> = std::fs::read_to_string(secrets_path)
+        .ok()
+        .map(|s| {
+            s.lines()
+                .filter_map(|line| {
+                    let line = line.trim();
+                    if line.is_empty() || line.starts_with('#') {
+                        return None;
+                    }
+                    let eq = line.find('=')?;
+                    let key = line[..eq].trim();
+                    if key.is_empty() {
+                        None
+                    } else {
+                        Some(key.to_string())
+                    }
+                })
+                .collect()
+        })
+        .unwrap_or_default();
+    let mut shadowed_secrets: Vec<String> = schema
+        .fields
+        .iter()
+        .filter(|field| field.field_type == "secret")
+        .filter(|field| {
+            values
+                .get(&field.key)
+                .is_some_and(|value| !value.trim().is_empty())
+        })
+        .filter(|field| std::env::var(&field.key).is_ok() && !secrets_env_keys.contains(&field.key))
+        .map(|field| field.key.clone())
+        .collect();
+    shadowed_secrets.sort();
+
+    let mut nonsecret_env = std::collections::BTreeMap::new();
+    for field in &schema.fields {
+        let Some(raw) = values.get(&field.key) else {
+            continue;
+        };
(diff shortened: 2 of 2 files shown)


changelog.d/changed/6992-sidecar-configure-io-off-thread.md
```diff
@@ -0,0 +1,4 @@
+`POST /api/channels/sidecar/{name}/configure` ran the `include`-shadow check, the `secrets.env` membership read, secret/config writes, and the `config.toml` upsert synchronously on the async worker thread handling the request.
```
houko, line 1: Commit `7e9a63087` on this branch ("docs: add missing changelog fragment for sidecar configure IO offload", which added this file) has git author identity `Claude <noreply@anthropic.com>`.

CLAUDE.md's "No AI / Claude attribution" rule and the repo's `commit-msg` hook both explicitly reject commits whose *author identity* resolves to Claude/Anthropic, independent of the commit message text — see the "Commit & PR hygiene" section and the `commit-msg` hook description in CLAUDE.md. This will fail that hook for anyone re-running it, and violates the policy regardless.

Fixing this requires rewriting this commit's author (rebase/amend) and force-pushing the branch, since the commit is already on the remote. I don't have authorization to force-push here, so flagging for the PR author to correct (e.g. `git commit --amend --author=... ` on that commit, or an interactive rebase, followed by a force-push to this branch).

---

crates/librefang-api/src/routes/channels.rs
```diff
@@ -937,159 +1026,41 @@ pub async fn configure_sidecar_channel(
...
-        shadowed_secrets = shadowed;
-
-        for f in &schema.fields {
-            let Some(raw) = body.values.get(&f.key) else {
-                continue;
-            };
-            let trimmed = raw.trim();
-            if trimmed.is_empty() {
-                continue;
+        tokio::task::spawn_blocking(move || {
```
houko, line 1039: CLAUDE.md's MANDATORY Integration Testing section requires a `#[tokio::test]` against `TestServer` for "any route / wiring change" — this PR changes the execution model of `POST /api/channels/sidecar/{name}/configure` (sync → `spawn_blocking` under `config_write_lock`, plus a new join-error path returned as `internal_scrub`).

The tests added in this PR (`sidecar_configuration_write_tests`) only exercise the extracted pure function `write_sidecar_configuration` directly, not the HTTP route. There's no test in `crates/librefang-api/tests/` that spawns the router via `start_test_server()` and POSTs to this endpoint, so the new `spawn_blocking`/lock/join-error wiring itself is untested end-to-end (e.g. a real request never verifies the reload-after-write ordering, or that a panicking blocking task actually surfaces as a scrubbed 500 rather than an unhandled panic).

Per CLAUDE.md: "Reviewers gate PRs on the presence of an integration test for each new endpoint." Recommend adding one here before merge.

---

houko (commented):
Automated daily review pass on PR #6992. One mechanical CLAUDE.md prose-wrapping fix pushed directly; two judgment-call findings left inline (AI-attribution on a commit's author identity, and a missing integration test per the MANDATORY Integration Testing policy).

---

changelog.d/changed/6992-sidecar-configure-io-off-thread.md
houko, line 1: Commit `7e9a630` ("docs: add missing changelog fragment for sidecar configure IO offload") has a clean message but its *author identity* is `Claude <noreply@anthropic.com>` (`git show -s --format='%an <%ae>' 7e9a630`).
CLAUDE.md's `commit-msg` hook rejects commits whose author identity resolves to Claude/Anthropic even when the message text is clean — this is the same pattern flagged on the sibling PR #7031.
Since this history is already pushed, please re-author that commit (e.g. `git commit --amend --author "<real-name> <real-email>"` + force-push your own branch) rather than leaving Claude as the recorded author.

---

crates/librefang-api/src/routes/channels.rs
houko, line 1039: Correctness of the offload itself looks fine: `_config_guard` is held across the `spawn_blocking(...).await`, so the write still completes (and its `Result` is still propagated through `?`) before the lock is released and before `reload_config().await` runs — no error-visibility or check/write ordering regression that I can find, and the join-error arm turns a panicked task into a scrubbed 500 instead of an unhandled panic. Good catch also moving the `include`-shadow check inside the lock.

One gap: this PR adds only plain `#[test]` unit coverage for the extracted `write_sidecar_configuration` (in `sidecar_configuration_write_tests` below). I don't see a `#[tokio::test]` against `TestServer` for `POST /api/channels/sidecar/{name}/configure` anywhere in `crates/librefang-api/tests/` (searched all test files — none reference `configure_sidecar_channel` or this route). Per CLAUDE.md's "MANDATORY: Integration Testing" section, a route/wiring change like this (new `spawn_blocking` hop, new error variant mapped to HTTP status) should have a `TestServer`-level test that POSTs the endpoint and asserts the response + a follow-up read, to catch exactly the kind of thread-hop regression (dropped error, response-shape drift) this PR is meant to prevent. Was this intentionally left as unit-only, or should a route-level test be added?

---

houko (commented):
Automated daily review pass: two findings, both left as comments (no fixes pushed, no approval/changes requested).

---

crates/librefang-api/src/routes/channels.rs
```diff
@@ -855,6 +855,95 @@ fn included_files_with_sidecars(config_path: &std::path::Path) -> Vec<std::path:
...
+    let mut shadowed_secrets: Vec<String> = schema
+        .fields
+        .iter()
+        .filter(|field| field.field_type == "secret")
+        .filter(|field| {
+            values
+                .get(&field.key)
+                .is_some_and(|value| !value.trim().is_empty())
+        })
+        .filter(|field| std::env::var(&field.key).is_ok() && !secrets_env_keys.contains(&field.key))
```
houko, line 905: Minor: the extraction into `write_sidecar_configuration` dropped the rationale comment that used to sit above this filter (the one explaining that `secrets_env_keys` intentionally mirrors `librefang_channels::sidecar::parse_secrets_env`'s key-extraction path but skips its value-side quote-stripping, and that a future value-comparison change should call that helper directly — referenced there as "codex review fix #9"). Worth carrying that context onto the new free function so it isn't lost to future editors of this code path; not a functional issue, just flagging since it documented a real quote-handling subtlety that's easy to violate by accident.

---

houko (commented):
Reviewed the perf refactor for correctness (spawn_blocking closure lifetimes/'static bounds, lock scope across the await, error mapping, include-shadow check ordering). Logic and locking are sound; one non-blocking documentation nit inline.

---

houko:
Superseded by #7170, which is now merged. The replacement moves sidecar configuration file work off the async runtime and includes production-route concurrency coverage.

PR #6998, opened 2026-08-12
fix(i18n): complete Japanese error translations

houko:
## Summary
- add the five error-message translations missing from the Japanese Fluent locale
- preserve all Fluent interpolation variables used by the English source
- add a regression test requiring Japanese to cover every English error key

## Testing
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-types i18n --lib`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo clippy -p librefang-types --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/changelog.d/fixed/6998-japanese-error-locale-completeness.md
+++ b/changelog.d/fixed/6998-japanese-error-locale-completeness.md
@@ -1,0 +1,2 @@
+Added the five error-message translations missing from the Japanese Fluent locale (an agent invalid-sort key and four webhook error keys), preserving every Fluent interpolation variable used by the English source.
+Added a regression test asserting the Japanese locale covers every English error key so a newly introduced key can no longer ship without a translation (#6998) (@houko)
--- a/crates/librefang-types/locales/ja/errors.ftl
+++ b/crates/librefang-types/locales/ja/errors.ftl
@@ -16,6 +16,7 @@ api-error-agent-execution-failed = エージェントの実行に失敗しまし
 api-error-agent-clone-spawn-failed = クローンの作成に失敗しました: { $error }
 api-error-agent-error = エージェントエラー: { $error }
 api-error-agent-not-found-with-id = エージェントが見つかりません: { $id }
+api-error-agent-invalid-sort = 無効なソートフィールド '{ $field }' です。有効なフィールド: { $valid }
 
 # メッセージエラー
 api-error-message-too-large = メッセージが大きすぎます（最大 64KB）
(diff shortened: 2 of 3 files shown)


crates/librefang-types/locales/ja/errors.ftl
```diff
@@ -16,6 +16,7 @@ api-error-agent-execution-failed = エージェントの実行に失敗しまし
 api-error-agent-clone-spawn-failed = クローンの作成に失敗しました: { $error }
 api-error-agent-error = エージェントエラー: { $error }
 api-error-agent-not-found-with-id = エージェントが見つかりません: { $id }
+api-error-agent-invalid-sort = 無効なソートフィールド '{ $field }' です。有効なフィールド: { $valid }
```
houko, line 19: Per CLAUDE.md's completeness-verification guidance for this PR, I diffed every locale's key set against `en`. Two related observations, both out of scope for a mechanical fix here so leaving them for a maintainer call:

1. `de`, `fr`, `es`, and `zh-CN` still lack these same 5 keys (`api-error-agent-invalid-sort` + the 4 webhook keys). "Complete" here is `ja`-only — worth confirming that's intentional, or opening one follow-up PR per remaining locale.
2. `ja` carries an extra key with no `en` counterpart: `api-error-tool-unknown` (line 291, `= 不明なツール: { $name }`). `grep -rn 'api-error-tool-unknown' crates/` finds no call site in Rust — it looks like an orphaned/dead translation, pre-existing and untouched by this diff, but it's the kind of drift the new regression test (checking `en → ja` only) won't catch in the other direction. Flagging for awareness, not blocking.

---

crates/librefang-types/src/i18n.rs
```diff
@@ -405,4 +405,28 @@ mod tests {
...
         }
     }
+
+    #[test]
+    fn japanese_locale_covers_every_english_error_key() {
+        fn message_keys(source: &str) -> std::collections::BTreeSet<&str> {
+            source
+                .lines()
+                .filter_map(|line| {
+                    let (key, _) = line.split_once('=')?;
```
houko, line 415: Minor robustness note, not an active bug: `line.split_once('=')` takes everything before the first `=` as the key for any line containing one. Fluent allows a message value to wrap onto an indented continuation line, and if such a continuation line ever contained a literal `=`, this would misparse it as a new key rather than skipping it. Checked all 9 `errors.ftl` files — none currently have continuation lines, so today's result is correct — but a `!line.starts_with(char::is_whitespace)` guard would make the parser robust against that case if multiline messages get added later.

---

changelog.d/fixed/6998-japanese-error-locale-completeness.md
houko, line 1: Commit author identity violates the repo's no-AI-attribution rule (CLAUDE.md, "Commit & PR hygiene" / "No AI / Claude attribution"), independent of message content.

The PR's second commit, which added this changelog fragment, is authored (and committed) as:

```
93c38eb09e63a7145cdc1c4a6e93e39f333164fb
docs(changelog): add fragment for Japanese error locale completeness
```

The commit message body itself is clean (no `Co-Authored-By` / `Generated with` strings), so `guard-bash-safety.sh`'s message-text check wouldn't catch this — but CLAUDE.md's `commit-msg` hook description is explicit that it also "rejects a commit whose *author identity* (`git var GIT_AUTHOR_IDENT`) resolves to Claude / Anthropic even when the message itself is clean." This is the same class of issue flagged on sibling PR #7039.

Since the commit is already pushed, rewriting it means history rewrite (rebase + force-push) on this branch, which is out of scope for an automated mechanical fix — flagging for the human maintainer to squash/reauthor before merge rather than fixing it here.

(First commit on the branch, `109e465e...` by `Evan <tonymo2048@gmail.com>`, is clean.)

---

PR #7003, opened 2026-08-12
fix(memory): log proactive lock poison recovery

houko:
## Summary
- centralize proactive-memory config and maintenance lock recovery
- warn whenever poisoned state is accepted instead of silently continuing
- recover the consolidation-counter prune lock instead of silently skipping maintenance
- verify config reads, writes, and maintenance state remain available after poisoning

## Testing
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-memory proactive --lib`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo clippy -p librefang-memory --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/librefang-memory/src/proactive.rs
+++ b/crates/librefang-memory/src/proactive.rs
@@ -161,14 +161,23 @@ impl ProactiveMemoryStore {
     fn read_config(&self) -> RwLockReadGuard<'_, ProactiveMemoryConfig> {
         self.config.read().unwrap_or_else(|poisoned| {
             tracing::warn!("proactive memory config read lock poisoned; recovering inner state");
-            poisoned.into_inner()
+            let guard = poisoned.into_inner();
+            // `into_inner()` only unwraps the guard; it leaves the lock's
+            // poison flag set. Without this, every later call through this
+            // helper would keep re-entering this branch and re-emitting the
+            // warning above for the rest of the process, even though the
+            // state has already been recovered.
+            self.config.clear_poison();
+            guard
         })
     }
 
     fn write_config(&self) -> RwLockWriteGuard<'_, ProactiveMemoryConfig> {
         self.config.write().unwrap_or_else(|poisoned| {
             tracing::warn!("proactive memory config write lock poisoned; recovering inner state");
-            poisoned.into_inner()
+            let guard = poisoned.into_inner();
+            self.config.clear_poison();
+            guard
         })
     }
 
@@ -178,7 +187,9 @@ impl ProactiveMemoryStore {
                 lock = name,
                 "proactive memory lock poisoned; recovering inner state"
             );
-            poisoned.into_inner()
+            let guard = poisoned.into_inner();
+            lock.clear_poison();
+            guard
         })
     }
 
--- a/crates/librefang-memory/src/proactive.rs
+++ b/crates/librefang-memory/src/proactive.rs
@@ -38,7 +38,7 @@ use librefang_types::memory::{
     CHAT_SCOPE_METADATA_KEY,
 };
 use std::collections::HashMap;
-use std::sync::{Arc, Mutex, RwLock};
+use std::sync::{Arc, Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuard};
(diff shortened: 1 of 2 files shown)


crates/librefang-memory/src/proactive.rs
```diff
@@ -158,6 +158,30 @@ impl Clone for ProactiveMemoryStore {
...
+    }
+
+    fn write_config(&self) -> RwLockWriteGuard<'_, ProactiveMemoryConfig> {
+        self.config.write().unwrap_or_else(|poisoned| {
+            tracing::warn!("proactive memory config write lock poisoned; recovering inner state");
+            poisoned.into_inner()
+        })
+    }
+
+    fn lock_recover<'a, T>(lock: &'a Mutex<T>, name: &'static str) -> MutexGuard<'a, T> {
```
houko, line 181: `read_config` / `write_config` (lines 161-173) and `lock_recover` (lines 175-183) recover via `poisoned.into_inner()` but never call `RwLock::clear_poison()` / `Mutex::clear_poison()`.
Once one panic occurs while holding `config`, `last_decay_run`, `last_cleanup_run`, `last_counter_prune`, or `consolidation_counters`, that lock stays poisoned for the rest of the process.
Every later access re-enters the `unwrap_or_else` branch and re-emits the `tracing::warn!` — unbounded log volume after a single panic, not a one-shot signal.
Same gap was flagged on PR #7007's `lock_audit_recover` (same "poison recovery" batch, different crate) as pre-existing behavior rather than a regression, since the underlying `unwrap_or_else(|e| e.into_inner())` idiom already existed at every one of these call sites before this PR — this PR only adds logging, which is itself a net improvement over the prior silent recovery.
Flagging here for consistency rather than requesting a fix: the risk profile is lower than the audit chain, since the critical sections guarded here are a single struct replace (`update_config`) or simple `HashMap`/`Option<DateTime>` bookkeeping rather than multi-step chained state, so a recovered guard is unlikely to hand back a torn value.
Worth a conscious call from a maintainer: is a follow-up adding `clear_poison()` after successful recovery warranted across this whole PR batch, so a one-time panic doesn't warn forever?

---

houko (commented):
Automated daily review pass. One consistency note left inline (matches the judgment-call treatment already applied to the sibling poison-recovery PR #7007) — no fixes pushed, no other findings.

---

crates/librefang-memory/src/proactive.rs
houko, line 1: `read_config`/`write_config`/`lock_recover` all call `poisoned.into_inner()` but never `clear_poison()` on the underlying lock. `into_inner()` only unwraps the guard — the poison flag stays set, so once one of these locks is poisoned, every subsequent call through the corresponding helper keeps re-entering the poisoned branch and re-emitting its `tracing::warn!` for the rest of the process. Sibling PR #7013 (`fix(runtime): clear command lane lock poison after recovery`) hit and fixed this exact gap for `CommandQueue`'s locks by calling `clear_poison()` once the poisoned state is read out — worth the same fix in all three helpers here (`self.config.clear_poison()` in the two `RwLock` helpers, `lock.clear_poison()` in `lock_recover`).

---

crates/librefang-memory/src/proactive.rs
houko: Fixed in 785de2211a705911a24a2624689139baaf749eba: `read_config`/`write_config`/`lock_recover` now call `clear_poison()` on the underlying lock right after `into_inner()`, matching the pattern from #7013. Covered by the existing `test_recovers_config_and_maintenance_locks_after_poisoning` test, which still passes (`cargo test -p librefang-memory --lib proactive`, 46/46 passed).

---

crates/librefang-memory/src/proactive.rs
houko: Fixed in 785de2211a705911a24a2624689139baaf749eba, same shape as #7013's fix: `self.config.clear_poison()` in the two `RwLock` helpers and `lock.clear_poison()` in `lock_recover`.

---

PR #7004, opened 2026-08-12
fix(channels): log router lock poison recovery

houko:
## Summary
- centralize binding and broadcast mutex acquisition in the channel agent router
- warn whenever poisoned routing state is recovered instead of silently continuing
- preserve routing availability and verify bindings and broadcasts still resolve after poisoning

## Testing
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-channels router --lib`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo clippy -p librefang-channels --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/librefang-channels/src/router.rs
+++ b/crates/librefang-channels/src/router.rs
@@ -6,7 +6,7 @@ use librefang_types::agent::AgentId;
 use librefang_types::config::{AgentBinding, BroadcastConfig, BroadcastStrategy};
 use smallvec::SmallVec;
 use std::borrow::Cow;
-use std::sync::Mutex;
+use std::sync::{Mutex, MutexGuard};
 use tracing::warn;
 
 /// Context for evaluating binding match rules against incoming messages.
@@ -61,6 +61,20 @@ pub struct AgentRouter {
 }
 
 impl AgentRouter {
+    fn lock_bindings(&self) -> MutexGuard<'_, Vec<(AgentBinding, String)>> {
+        self.bindings.lock().unwrap_or_else(|poisoned| {
+            warn!("agent router bindings lock poisoned; recovering inner state");
+            poisoned.into_inner()
+        })
+    }
+
+    fn lock_broadcast(&self) -> MutexGuard<'_, BroadcastConfig> {
+        self.broadcast.lock().unwrap_or_else(|poisoned| {
+            warn!("agent router broadcast lock poisoned; recovering inner state");
+            poisoned.into_inner()
+        })
+    }
+
     /// Create a new router.
     pub fn new() -> Self {
         Self {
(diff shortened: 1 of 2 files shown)


crates/librefang-channels/src/router.rs
```diff
@@ -61,6 +61,20 @@ pub struct AgentRouter {
 }
 
 impl AgentRouter {
+    fn lock_bindings(&self) -> MutexGuard<'_, Vec<(AgentBinding, String)>> {
+        self.bindings.lock().unwrap_or_else(|poisoned| {
+            warn!("agent router bindings lock poisoned; recovering inner state");
+            poisoned.into_inner()
```
houko, line 67: `lock_bindings()` (and `lock_broadcast()` below) recover via `poisoned.into_inner()` but never call `Mutex::clear_poison()`.
The mutex stays poisoned indefinitely after a single panic, so every subsequent lock acquisition re-enters this `unwrap_or_else` branch and re-logs the warning forever — the "poisoned" state never clears even though the data itself is fine.
This mirrors the same gap flagged on #7007's `lock_audit_recover` (same author, same pattern, different crate).
Confirmed this behavior predates this PR (`git show origin/main:crates/librefang-channels/src/router.rs` already recovers via `.unwrap_or_else(|e| e.into_inner())` at every call site), so this PR only adds the `warn!` logging and doesn't introduce the gap — flagging as the same maintainer judgment call as on #7007 rather than auto-fixing.
If perpetual re-warning on every access after one panic is undesired, consider `poisoned.into_inner()` followed by `self.bindings.clear_poison()` (stable since Rust 1.77) so the mutex returns to a healthy state once recovered.

---

houko (commented):
Automated daily review pass. One finding, left inline: the router's lock-poison recovery never calls `clear_poison()`, mirroring the same gap already flagged on #7007 for the same author/pattern. Confirmed the recovery mechanism itself predates this PR (this PR only adds the `warn!` log around pre-existing recovery code), so it's a pre-existing behavior/judgment call, not a regression introduced here — no auto-fix applied. Everything else (commit message format, no changelog fragment needed for this log-and-recover pattern, test coverage for the poisoning scenario, no AI attribution) looks consistent with CLAUDE.md conventions.

---

crates/librefang-channels/src/router.rs
houko, line 1: `lock_bindings`/`lock_broadcast` call `poisoned.into_inner()` but never `clear_poison()` on the underlying `Mutex`. `into_inner()` only unwraps the guard — the poison flag stays set, so once poisoned, every subsequent binding/broadcast lookup keeps re-entering this branch and re-emitting the `warn!` for the rest of the process. Sibling PR #7013 (`fix(runtime): clear command lane lock poison after recovery`) hit and fixed this exact gap for `CommandQueue`'s locks by calling `clear_poison()` once the poisoned state is read out — worth the same fix in both helpers here.

---

PR #7007, opened 2026-08-12
fix(audit): log audit-chain lock poison recovery

houko:
## Summary
- centralize audit-log mutex recovery with warnings identifying entries, tip, chain anchor, or load-error state
- replace every silent poison recovery across recording, verification, retention, and accessors
- verify recording can continue after entries/tip poisoning while the complete hash chain still passes integrity validation

## Testing
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-runtime-audit --lib`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo clippy -p librefang-runtime-audit --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/librefang-runtime-audit/src/lib.rs
+++ b/crates/librefang-runtime-audit/src/lib.rs
@@ -16,7 +16,14 @@ use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use std::collections::BTreeMap;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::Mutex;
+use std::sync::{Mutex, MutexGuard};
+
+fn lock_audit_recover<'a, T>(mutex: &'a Mutex<T>, state: &'static str) -> MutexGuard<'a, T> {
+    mutex.lock().unwrap_or_else(|poisoned| {
+        tracing::warn!(state, "audit log lock poisoned; recovering inner state");
+        poisoned.into_inner()
+    })
+}
 
 /// Default hard cap on the number of audit entries kept in memory when no
 /// operator-supplied `max_in_memory_entries` is configured.
@@ -527,9 +534,8 @@ impl AuditLog {
         // return `Err` in that case so `/api/audit/verify` surfaces it.
         match Self::read_anchor(&anchor_path) {
             Ok(Some(record)) => {
-                let current_tip = log.tip.lock().unwrap_or_else(|e| e.into_inner()).clone();
-                let current_seq =
-                    log.entries.lock().unwrap_or_else(|e| e.into_inner()).len() as u64;
+                let current_tip = lock_audit_recover(&log.tip, "tip").clone();
+                let current_seq = lock_audit_recover(&log.entries, "entries").len() as u64;
                 if record.hash != current_tip {
                     tracing::error!(
                         anchor_seq = record.seq,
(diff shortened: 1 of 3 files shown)


crates/librefang-runtime-audit/src/lib.rs
```diff
@@ -16,7 +16,14 @@ use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 use std::collections::BTreeMap;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::Mutex;
+use std::sync::{Mutex, MutexGuard};
+
+fn lock_audit_recover<'a, T>(mutex: &'a Mutex<T>, state: &'static str) -> MutexGuard<'a, T> {
```
houko, line 21: `lock_audit_recover` recovers via `poisoned.into_inner()` but never calls `Mutex::clear_poison()`, so the mutex stays poisoned for the rest of the process once one panic occurs under the lock. Two consequences worth a maintainer call on, specifically because this is the audit chain (CLAUDE.md flags audit integrity as security-sensitive):

1. Every later `record_with_context` / `verify_integrity` / `recent` / etc. call on the still-poisoned `entries` or `tip` mutex will re-take this branch and re-emit the `warn!` forever after a single panic — unbounded log volume under sustained traffic, not a one-shot signal.
2. More importantly, recovery here is "grab the possibly-mid-mutation inner state and keep going" with no re-validation. If the panic happened between the `entries` push and the `tip` update inside `record_with_context` (two separate `Mutex`es, not one combined lock), a recovered guard could hand back a `entries`/`tip` pair that's already inconsistent, and nothing here calls `verify_integrity()` to catch that before the next entry is chained on top of it. The new test only exercises poisoning `entries` and `tip` independently while idle, not poisoning mid-`record_with_context`, so it doesn't cover that window.

I recognize this exact `unwrap_or_else(|poisoned| { warn!(...); poisoned.into_inner() })` idiom is now uniform across the whole batch of sibling poison-recovery PRs (#7002, #7005, #7008, #7009, #7011, #7012, #7020, ...), so this isn't a regression unique to this PR — it's pre-existing behavior this PR only makes observable via logging (which is itself a net improvement over the prior *silent* recovery). Flagging here rather than in the sibling PRs because this is the one crate where a missed inconsistency is a security/integrity concern rather than a cache/registry correctness concern. Worth a conscious decision: is a follow-up that adds `clear_poison()` (to stop the log spam) and/or an automatic `verify_integrity()` check on recovery (to fail loud instead of silently building on possibly-corrupt state) warranted for this crate specifically, even if the other batch members stay as-is?

---

houko (commented):
Automated CLAUDE.md-rules pass on this batch-of-8 poison-recovery PR. Commit message is conventional-commit format, no changelog fragment (consistent with every other merged sibling PR in this batch — not flagging as a deviation), no AI/Claude/Anthropic attribution found anywhere in the diff, and the mechanical refactor itself (consolidating repeated `.lock().unwrap_or_else(|e| e.into_inner())` call sites into `lock_audit_recover` plus a regression test) checks out against the existing field types and call sites — no correctness bug in the refactor. Left one inline comment on an audit-chain-specific integrity/observability design question that needs a maintainer decision rather than a mechanical fix.

---

crates/librefang-runtime-audit/src/lib.rs
houko, line 1: The new `lock_audit_recover` helper calls `poisoned.into_inner()` but never `mutex.clear_poison()`. `into_inner()` only unwraps the guard — the lock's poison flag stays set, so once one panic poisons e.g. `entries`, every later call through this helper for that same lock keeps re-entering the poisoned branch and re-emitting the `tracing::warn!` for the rest of the process. Sibling PR #7013 (`fix(runtime): clear command lane lock poison after recovery`) hit and fixed this exact gap for `CommandQueue`'s locks by calling `clear_poison()` once the poisoned state is read out — since this is a shared helper, one `mutex.clear_poison()` call here would cover all four locks (`entries`, `tip`, `chain_anchor`, `load_error`) at once.

---

houko:
Pushed b8836a2 to address the missing `clear_poison()` call flagged in the two review threads above: `lock_audit_recover` now calls `mutex.clear_poison()` before `into_inner()`, matching the idiom already applied to sibling poison-recovery fixes in this batch (agent context cache, command lane).
The regression test now also asserts `is_poisoned()` before and after recovery to lock this in.
The separate, broader question raised in the first thread — whether recovery should also re-run `verify_integrity()` to detect a possibly-inconsistent `entries`/`tip` pair from a panic mid-`record_with_context` — is left open for a maintainer call; that is a design decision, not a mechanical fix.

---

Issue #3579, opened 2026-04-27
perf: synchronous std::fs and std::sync::RwLock inside async axum handlers


houko:
## Summary
Several blocking syscalls and `std::sync::RwLock` reads run directly on the tokio runtime inside async handlers. Under load (slow disk, NFS, AV scanning, contended writer), tokio worker threads park and API requests appear to hang.

## Evidence
- `crates/librefang-kernel/src/kernel/mod.rs:14507` — `std::fs::read_to_string` on a cache-miss prompt path
- `crates/librefang-kernel/src/kernel/mod.rs:381, 434, 442, 477, 582` — `std::sync::RwLock` for `model_catalog`, `mcp_catalog`, `budget_config`. Read 100× per write from handlers like `agents.rs:827`, `agents.rs:1369`, plus auth middleware.
- `crates/librefang-runtime/src/agent_context.rs:135-158` — sync `fs::symlink_metadata` + `fs::File::open` per turn for `context.md`
- `crates/librefang-api/src/routes/agents.rs:8731` — `std::fs::read_to_string` on agent.toml during spawn

## Suggested fix
Wrap genuine blocking I/O in `tokio::task::spawn_blocking` (or use `tokio::fs`). Convert hot-path `std::sync::RwLock` to `arc_swap::ArcSwap` (already used for `KernelConfig` per `mod.rs:741`) — `model_catalog` / `mcp_catalog` / `budget_config` are read 100× per write and are perfect ArcSwap candidates. The per-turn `context.md` read should move to `tokio::fs::read` inside a `spawn_blocking` boundary.

---
_Reported by automated multi-agent code review (security / concurrency / architecture / performance / testing). Evidence is grep-verified file:line; please re-confirm before fixing._

PR #7013, opened 2026-08-12
fix(runtime): log command lane lock recovery

houko:
## Summary
- centralize command-lane RwLock recovery for reads and writes
- include the affected lane in poison recovery warnings
- verify semaphore access, resize, and occupancy remain usable after a lane lock is poisoned

## Verification
- `cargo test -p librefang-runtime command_lane --lib` (10 passed)
- `git diff --check`

### DIFF
--- a/crates/librefang-runtime/src/command_lane.rs
+++ b/crates/librefang-runtime/src/command_lane.rs
@@ -73,15 +73,22 @@ pub struct CommandQueue {
 
 impl CommandQueue {
     fn read_lane(&self, lane: Lane) -> RwLockReadGuard<'_, LaneSlot> {
-        self.slot(lane).read().unwrap_or_else(|poisoned| {
+        let lock = self.slot(lane);
+        lock.read().unwrap_or_else(|poisoned| {
             warn!(lane = %lane, "command lane read lock poisoned; recovering inner state");
+            // `into_inner()` only unwraps this guard; it does not reset the
+            // lock's poison flag, so without `clear_poison()` every future
+            // access re-enters this branch and re-logs forever.
+            lock.clear_poison();
             poisoned.into_inner()
         })
     }
 
     fn write_lane(&self, lane: Lane) -> RwLockWriteGuard<'_, LaneSlot> {
-        self.slot(lane).write().unwrap_or_else(|poisoned| {
+        let lock = self.slot(lane);
+        lock.write().unwrap_or_else(|poisoned| {
             warn!(lane = %lane, "command lane write lock poisoned; recovering inner state");
+            lock.clear_poison();
             poisoned.into_inner()
         })
     }
@@ -236,7 +243,12 @@ mod tests {
         });
 
         assert!(poison.is_err());
+        assert!(queue.main.is_poisoned());
         assert_eq!(queue.semaphore_for_lane(Lane::Main).available_permits(), 3);
+        // The first post-panic access must clear the poison flag, not just
+        // unwrap around it — otherwise every subsequent access re-triggers
+        // the recovery branch (and its `warn!`) for the rest of the process.
+        assert!(!queue.main.is_poisoned());
         assert!(queue.resize_lane(Lane::Main, 5));
         let main = queue
             .occupancy()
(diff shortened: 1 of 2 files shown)


crates/librefang-runtime/src/command_lane.rs
```diff
@@ -103,11 +125,7 @@ impl CommandQueue {
     /// detached `tokio::spawn` task — the returned `Arc<Semaphore>` is
     /// cheap to clone.
     pub fn semaphore_for_lane(&self, lane: Lane) -> Arc<Semaphore> {
-        self.slot(lane)
-            .read()
-            .unwrap_or_else(|e| e.into_inner())
-            .sem
-            .clone()
+        self.read_lane(lane).sem.clone()
```
houko, line 126: Not a regression from this PR, but worth a maintainer call while this exact synchronization is being touched: `submit`/`try_submit` are `async fn`, and this call path (`semaphore_for_lane` → `read_lane`) acquires a `std::sync::RwLock::read()` inside their bodies. `crates/librefang-runtime/CLAUDE.md`'s async-boundary rule says "No synchronous `std::fs` / `std::sync::RwLock` inside async handlers. Use `tokio::fs` / `arc_swap` / `parking_lot` (refs #3579)." This lock predates this PR, but this PR is the natural point to reconsider it: a full migration to `parking_lot::RwLock` (no poisoning) or `arc_swap` (lock-free swap of the `Arc<Semaphore>`) would eliminate the poison state this PR is adding logging for, rather than decorate it. Practically the critical sections here are just plain field reads / an `Arc` swap with no `.await` inside the guard, so the risk of actually blocking the executor is low — flagging as a judgment call rather than blocking, since resolving it properly is a bigger change than "add logging."

---

houko (commented):
Automated review pass against CLAUDE.md rules. Pushed a mechanical fix (poison flag was never cleared, so recovery re-triggered forever); left one architectural judgment call as an inline comment.

---

crates/librefang-runtime/src/command_lane.rs
houko, line 1: Commit `94ab62a` (`fix(runtime): clear command lane lock poison after recovery`) is authored as `Claude <noreply@anthropic.com>`. That trips the repo's no-AI-attribution policy on author identity (CLAUDE.md → "Git Conventions" / the `commit-msg` hook's `GIT_AUTHOR_IDENT` check), independent of the message text being clean. The fix itself is correct and worth keeping — flagging so the maintainer can re-author it (e.g. `git commit --amend --author` + force-push, or squash on merge) before this lands; I can't rewrite an already-pushed commit here.

---

houko, line 1: Two commits on this branch (`a48c7eb`, `94ab62a`) carry the author identity `Claude <noreply@anthropic.com>` (`git log --format='%an <%ae>'`).
CLAUDE.md's Git Conventions / commit-msg hook description explicitly extends the "no AI/Claude/Anthropic attribution" rule to author identity, not just message text: "rejects a commit whose author identity (`git var GIT_AUTHOR_IDENT`) resolves to Claude / Anthropic even when the message itself is clean."
Fixing this needs a rebase to set `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` on those two commits followed by a force-push, which is outside what an automated pass should do unilaterally on someone else's branch — flagging for @houko to rebase-and-force-push before merge, since the `commit-msg` hook would reject these on a fresh push if `core.hooksPath` is configured.

---

houko, line 1: Nice catch adding `clear_poison()` here — worth noting this PR is actually the first in the "log lane/lock poison recovery" series to do it.
Checked the siblings (`#7002` wire peer registry, `#7005` channels sidecar, `#7008` kernel approval, `#7009` kernel-router cache, `#7011` kernel accessor, `#7012` runtime shutdown, `#7020` media provider): none of them call `clear_poison()`, so each of those locks still re-emits its `warn!` on every access after a single panic, forever, exactly like the bug this PR fixes for the command lane.
Not something to fix in this PR (different crates/domains), but probably worth a follow-up sweep across the series since the same fix applies verbatim to each.

---

houko (commented):
Automated daily review pass against CLAUDE.md rules. Two findings posted inline; everything else checked out (changelog fragment format/section/attribution, prose one-sentence-per-line, no HashMap/HashSet in this diff, build commands scoped per policy).

---

PR #7029, opened 2026-08-12
fix(runtime): recover agent context cache poison

houko:
## Summary
- recover the agent context cache after mutex poisoning instead of permanently dropping reads and writes
- log every recovery so corrupted synchronization state is observable
- verify cached values can be inserted, read, and updated after poisoning

## Verification
- `cargo test -p librefang-runtime poisoned_cache_lock_recovers_cached_context --lib`
- `git diff --check`

### DIFF
--- a/crates/librefang-runtime/src/agent_context.rs
+++ b/crates/librefang-runtime/src/agent_context.rs
@@ -45,6 +45,8 @@ fn cache() -> &'static Mutex<HashMap<PathBuf, String>> {
 fn lock_cache() -> MutexGuard<'static, HashMap<PathBuf, String>> {
     cache().lock().unwrap_or_else(|poisoned| {
         warn!("Agent context cache lock poisoned; recovering cached context");
+        // `into_inner()` only unwraps this guard; it does not reset the mutex's poison flag, so without `clear_poison()` every future access re-enters this branch and re-logs forever.
+        cache().clear_poison();
         poisoned.into_inner()
     })
 }
@@ -490,10 +492,13 @@ mod tests {
         })
         .join();
         assert!(poison.is_err());
+        assert!(cache().is_poisoned());
 
         let path = fresh_workspace("poison_recovery").join(CONTEXT_FILENAME);
         store_cached(&path, "first cached value");
         assert_eq!(get_cached(&path).as_deref(), Some("first cached value"));
+        // The first post-panic access must clear the poison flag, not just unwrap around it — otherwise every subsequent access re-triggers the recovery branch (and its `warn!`) for the rest of the process.
+        assert!(!cache().is_poisoned());
 
         store_cached(&path, "updated cached value");
         assert_eq!(get_cached(&path).as_deref(), Some("updated cached value"));
--- a/crates/librefang-runtime/src/agent_context.rs
+++ b/crates/librefang-runtime/src/agent_context.rs
@@ -15,7 +15,7 @@
 
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
-use std::sync::{Mutex, OnceLock};
+use std::sync::{Mutex, MutexGuard, OnceLock};
 use std::{fs, io};
 
 use tracing::{debug, warn};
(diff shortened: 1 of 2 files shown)


crates/librefang-runtime/src/agent_context.rs
```diff
@@ -42,6 +42,13 @@ fn cache() -> &'static Mutex<HashMap<PathBuf, String>> {
     CACHE.get_or_init(|| Mutex::new(HashMap::new()))
 }
 
+fn lock_cache() -> MutexGuard<'static, HashMap<PathBuf, String>> {
+    cache().lock().unwrap_or_else(|poisoned| {
+        warn!("Agent context cache lock poisoned; recovering cached context");
+        poisoned.into_inner()
+    })
+}
```
houko, line 52: `std::sync::Mutex` poisoning is permanent — `poisoned.into_inner()` extracts the guard but never clears the mutex's poison flag (only `Mutex::clear_poison()`, stable since Rust 1.77, does that; workspace `rust-version` is 1.94.1, so it's available). As written, after the *first* poisoning event `cache().lock()` returns `Err` on every subsequent call for the rest of the process, so `warn!` fires on every single `get_cached`/`store_cached` — i.e. every agent turn — forever, not just once at recovery. The PR's own new test demonstrates this: `poisoned_cache_lock_recovers_cached_context` poisons the shared `static CACHE` and never un-poisons it, so every other test in this module that touches `load_context_md` for the rest of the test binary's process now also takes the warn-and-recover path.

The PR description says "log every recovery so corrupted synchronization state is observable," so this might be intentional, but it's worth confirming: is continuous per-call log spam after a single panic the desired behavior, or should the poisoned branch also call `cache().clear_poison()` after logging once, so the mutex returns to normal `Ok` locking and the warning fires only at the actual recovery moment? Given this cache is read once per agent turn, unbounded warn-level spam for the remaining lifetime of the daemon after any single panic seems like a real operational cost either way.

Separately (non-blocking): recovery here is safe from a data-corruption standpoint — the only code that runs while the lock is held is a `HashMap::get`/`.cloned()` or a single `.insert()`, so a poisoning panic can only come from an allocation failure inside those calls, not from a torn multi-step update; there's no path today by which a recovered guard hands back a logically inconsistent cache entry.

---

houko (commented):
Automated daily review pass. One mechanical fix pushed (missing changelog.d fragment per CLAUDE.md); one inline comment left for human judgment on the poison-recovery logging behavior.

---

crates/librefang-runtime/src/agent_context.rs
```diff
@@ -42,6 +42,13 @@ fn cache() -> &'static Mutex<HashMap<PathBuf, String>> {
     CACHE.get_or_init(|| Mutex::new(HashMap::new()))
 }
 
+fn lock_cache() -> MutexGuard<'static, HashMap<PathBuf, String>> {
+    cache().lock().unwrap_or_else(|poisoned| {
+        warn!("Agent context cache lock poisoned; recovering cached context");
+        poisoned.into_inner()
```
houko, line 50: `into_inner()` recovers the guard but leaves the mutex's poison flag set, so every subsequent `cache().lock()` keeps taking this `Err(poisoned)` branch — and re-logging the warning — for the rest of the process's life, even though the recovered `HashMap` is known-good again after this call.

The sibling PR #7030 (checkpoint counter poison recovery) hit the identical gap and needed a follow-up commit (`fix(runtime): clear poison flag after recovering checkpoint counter`) to call `.clear_poison()` once recovery succeeds. Worth doing the same here — `cache().clear_poison();` right after `poisoned.into_inner()` — so this lock stops permanently reporting itself as poisoned once it's been recovered once. Not blocking (the cache still works correctly on every call via `into_inner()`), but it means every turn after the first poisoning event pays a log line and the `Err` branch forever, which will look like a persistent problem in logs long after it's actually resolved.

---

houko (commented):
Self-flag: commit 18068cd58dbc513b8f9ea342058ada3f6e512a46 ("clear agent context cache poison flag on recovery") was pushed by this same automated review run, but its git author identity resolved to `Claude <noreply@anthropic.com>` — the AI-attribution violation CLAUDE.md's `commit-msg` hook exists to reject. This turns out to be an environment-wide issue (this review session's git identity is hardcoded to Claude/Anthropic, confirmed via `git config user.name/user.email`), not specific to this PR — it also affects commit 1410d779525b02ceba471422226bb9053deac19e on this same PR from an earlier pass today, and has been found on several sibling PRs in this batch. Per this routine's hard rule against force-pushing, I'm not amending/rewriting either commit myself — flagging both here so a maintainer can rebase with correct authorship before merge. No further automated commits are being made this run until the underlying git identity is fixed; the fix content itself (calling `clear_poison()` after `into_inner()` recovery, matching the idiom used elsewhere in this crate) looks correct on inspection.

---

houko:
Correction from this automated review pass: the poison-flag fix commit it pushed here (`18068cd`) is itself authored as `Claude <noreply@anthropic.com>`, which is the exact no-AI-attribution violation CLAUDE.md's `commit-msg` hook targets. This happened because the review environment's global git identity was misconfigured (not something the review logic checked for on its own commits before pushing). The fix's content (adding `clear_poison()` after recovery) is still correct; a maintainer should reauthor/squash the commit before merge.

---

PR #7030, opened 2026-08-12
fix(runtime): recover checkpoint counter poison

houko:
## Summary
- recover the checkpoint concurrency counter after mutex poisoning instead of panicking at snapshot entry
- make the RAII cleanup guard recover and release its permit instead of silently skipping the decrement
- verify preserved counter state and balanced permit cleanup after poisoning

## Verification
- `cargo test -p librefang-runtime poisoned_snapshot_counter_recovers_and_guard_releases_permit --lib`
- `git diff --check`

### DIFF
--- a/crates/librefang-runtime/src/checkpoint_manager.rs
+++ b/crates/librefang-runtime/src/checkpoint_manager.rs
@@ -20,7 +20,7 @@
 use sha2::{Digest, Sha256};
 use std::path::{Path, PathBuf};
 use std::process::Command;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex, MutexGuard};
 use std::time::{SystemTime, UNIX_EPOCH};
 use thiserror::Error;
 use tracing::{debug, warn};
@@ -146,7 +146,23 @@ pub struct CheckpointManager {
     /// snapshot that cannot immediately obtain the permit is **skipped**
     /// rather than queued — preventing memory pressure from accumulated
     /// waiting tasks.
-    semaphore: Arc<std::sync::Mutex<usize>>,
+    semaphore: Arc<Mutex<usize>>,
+}
+
+fn lock_snapshot_counter(counter: &Mutex<usize>) -> MutexGuard<'_, usize> {
+    counter.lock().unwrap_or_else(|poisoned| {
+        warn!("checkpoint concurrency counter lock poisoned; recovering counter state");
+        poisoned.into_inner()
+    })
+}
+
+struct ConcurrencyGuard(Arc<Mutex<usize>>);
+
+impl Drop for ConcurrencyGuard {
+    fn drop(&mut self) {
+        let mut count = lock_snapshot_counter(&self.0);
+        *count = count.saturating_sub(1);
+    }
 }
 
 impl CheckpointManager {
@@ -157,7 +173,7 @@ impl CheckpointManager {
     pub fn new(base_dir: PathBuf) -> Self {
         Self {
             base_dir,
-            semaphore: Arc::new(std::sync::Mutex::new(0)),
+            semaphore: Arc::new(Mutex::new(0)),
         }
     }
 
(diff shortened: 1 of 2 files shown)


crates/librefang-runtime/src/checkpoint_manager.rs
```diff
@@ -146,7 +146,27 @@ pub struct CheckpointManager {
...
+    semaphore: Arc<Mutex<usize>>,
+}
+
+fn lock_snapshot_counter(counter: &Mutex<usize>) -> MutexGuard<'_, usize> {
+    counter.lock().unwrap_or_else(|poisoned| {
+        warn!("checkpoint concurrency counter lock poisoned; recovering counter state");
+        let guard = poisoned.into_inner();
+        // `into_inner` recovers the guard but leaves the mutex's poison flag set, so every future `lock()` call would keep taking this `Err(PoisonError)` branch (and re-logging this warning) even though the counter itself is now known-good.
+        // Clear it once we've recovered so the mutex behaves normally again for the rest of the process's life.
+        counter.clear_poison();
```
houko, line 158: This adds a `counter.clear_poison()` call that none of the other poison-recovery fixes in this series do (#7002, #7005, #7008, #7009, #7011, #7012, #7020, and the still-open agent-context-cache one all stop at `.unwrap_or_else(|poisoned| { warn!(...); poisoned.into_inner() })`, leaving the mutex permanently poisoned so every later `lock()` re-enters the warned branch).

The reasoning in bda6d3a (avoid re-logging the warning forever) is sound, and the counter's own invariant does look safe to clear on: the only locked sections here are a single increment/compare and a `saturating_sub`, so there's no partial-update state a poisoned recovery could be papering over.

But that same "keeps re-warning forever" property is presumably true of every sibling fix (media provider state, shutdown locks, agent context cache, router cache, etc.), and none of them clear the flag. Two questions for the author:
1. Is `clear_poison()` intentionally checkpoint-counter-specific, or should the same follow-up be applied to the other 7 merged fixes for consistency?
2. If it's meant to become the standard idiom going forward, worth calling out explicitly in the PR description so reviewers of the still-open sibling PR (agent-context-cache) know to match it.

Not blocking, but flagging since CLAUDE.md calls out idiom consistency across this exact fix family.

---

houko (commented):
Automated daily review pass — one item flagged inline, no mechanical fixes needed (changelog fragment, commit messages, and prose wrapping all check out).

---

crates/librefang-runtime/src/checkpoint_manager.rs
houko, line 1: Commit d5c7ae9 ("docs: join wrapped sentences in lock_snapshot_counter comment", touching this file) has author identity `Claude <noreply@anthropic.com>`.
CLAUDE.md is explicit that no commit on this repo may carry Claude/Anthropic attribution, and specifically calls out that the `commit-msg` hook checks `git var GIT_AUTHOR_IDENT` (the `GIT_AUTHOR_NAME`/`GIT_AUTHOR_EMAIL` identity) separately from the message text, precisely to catch a clean-looking message with an AI author identity like this one.
The message body itself has no attribution string, so this wouldn't trip a text-based scan — it's the git author identity that's the problem.
Not proposing a fix here myself: correcting it means rewriting this commit's author and force-pushing, and the branch already has a review comment on it, so per the repo's own convention (force-push only before a reviewer has loaded the diff) I'm leaving this for the author/maintainer rather than rewriting history.
Worth confirming `core.hooksPath` was actually pointed at `scripts/hooks` in whatever environment produced this commit, since the hook is supposed to block it at commit time.

---

houko (commented):
Code logic review: the poison-recovery + `clear_poison()` + regression test all look correct (checked out the branch and confirmed `cargo check -p librefang-runtime --lib` compiles). One finding below on commit provenance, unrelated to the code itself. Not duplicating the existing thread about `clear_poison()` consistency across the sibling poison-recovery PRs — that's already been raised.

---

PR #7031, opened 2026-08-12
fix(memory): surface stuck task row errors

houko:
## Summary
- fail the stuck-task reset sweep when SQLite cannot decode a candidate row instead of silently dropping it
- keep the candidate set fully decoded before applying any reset updates
- verify a corrupt `retry_count` is reported and the task is not partially modified

## Verification
- `cargo test -p librefang-memory task_reset_stuck_surfaces_corrupt_retry_count --lib`
- `git diff --check`

### DIFF
--- a/changelog.d/fixed/7031-task-reset-stuck-row-errors.md
+++ b/changelog.d/fixed/7031-task-reset-stuck-row-errors.md
@@ -1,0 +1,2 @@
+The stuck-task reset sweep silently dropped any `task_queue` row it could not decode from SQLite, so a single corrupt row (e.g. a non-numeric `retry_count`) caused the rest of that sweep's stuck tasks to be skipped with no error surfaced to the caller.
+`task_reset_stuck` now decodes the full candidate set before applying any reset update, so a row decode failure fails the sweep closed instead of silently reducing its coverage (#7031) (@houko)
--- a/crates/librefang-memory/src/substrate.rs
+++ b/crates/librefang-memory/src/substrate.rs
@@ -1315,8 +1315,8 @@ impl MemorySubstrate {
                     Ok((row.get::<_, String>(0)?, row.get::<_, u32>(1)?))
                 })
                 .map_err(LibreFangError::memory)?
-                .filter_map(|r| r.ok())
-                .collect();
+                .collect::<rusqlite::Result<Vec<_>>>()
+                .map_err(LibreFangError::memory)?;
 
             if stuck.is_empty() {
                 return Ok(Vec::new());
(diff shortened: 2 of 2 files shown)


crates/librefang-memory/src/substrate.rs
```diff
@@ -1315,8 +1315,8 @@ impl MemorySubstrate {
                     Ok((row.get::<_, String>(0)?, row.get::<_, u32>(1)?))
                 })
                 .map_err(LibreFangError::memory)?
-                .filter_map(|r| r.ok())
-                .collect();
+                .collect::<rusqlite::Result<Vec<_>>>()
+                .map_err(LibreFangError::memory)?;
```
houko, line 1319: Nice fix for the row-decode fail-open bug — the `collect::<rusqlite::Result<Vec<_>>>()` change matches the same pattern landing in #7032, and the new test exercises the corrupt-row path well.

One thing worth a maintainer decision while we're in this function: the reset loop a few lines below (`for (id, retries) in &stuck { db.execute(...) }`, around line 1326) applies each row's `UPDATE` individually, outside a transaction. If `db.execute` fails partway through (e.g. lock contention), tasks earlier in `stuck` are already reset while later ones are left `in_progress` and the error propagates — so a mid-loop failure leaves the sweep half-applied, a related but separate fail-closed gap from the one this PR fixes for the decode phase.

Per CLAUDE.md's "fix what you found" policy this is same-crate/same-function, so it's arguably in scope for this PR — but wrapping the loop in a single SQLite transaction is a larger behavioral change than the decode fix here, so flagging for your call rather than bundling it in unasked.

---

changelog.d/fixed/7031-task-reset-stuck-row-errors.md
houko, line 1: Commit a52bb2a ("docs(changelog): add fragment for stuck task row error fix") has author identity `Claude <noreply@anthropic.com>` (`git log --format='%an <%ae>'`).
The commit message body itself is clean, but CLAUDE.md's commit-msg hook rejects a commit on author identity alone, even when the message text has no attribution.
This wasn't something I could fix with a follow-up commit — the bad identity lives on the existing commit object, and correcting it needs `git commit --amend --author=... ` plus a force-push, which is outside what an automated review pass should do to someone else's branch history.
Please amend this commit's author before merge (`git commit --amend --author="Your Name <you@example.com>"` then force-push your own branch).

---

houko (commented):
Automated CLAUDE.md compliance pass on this PR. No mechanical fixes were needed — the memory fix and changelog fragment are both correct and well-formed. One flag left as an inline comment: a commit's author identity resolves to Claude/Anthropic, which the repo's own commit-msg hook checks for independently of message text.

---

PR #7032, opened 2026-08-12
fix(memory): surface session search row errors

houko:
## Summary
- propagate FTS session-search row decoding failures instead of silently omitting corrupt matches
- apply the same fail-closed behavior to filtered and unfiltered searches
- verify a malformed FTS row is reported by both query paths

## Verification
- `cargo test -p librefang-memory fts_search_surfaces_corrupt_result_rows --lib`
- `git diff --check`

### DIFF
--- a/crates/librefang-memory/src/session.rs
+++ b/crates/librefang-memory/src/session.rs
@@ -1595,7 +1595,8 @@ impl SessionStore {
                 )
                 .map_err(LibreFangError::memory)?;
 
-            rows.filter_map(|r| r.ok()).collect()
+            rows.collect::<rusqlite::Result<Vec<_>>>()
+                .map_err(LibreFangError::memory)?
         } else {
             let mut stmt = conn
                 .prepare(
@@ -1621,7 +1622,8 @@ impl SessionStore {
                 )
                 .map_err(LibreFangError::memory)?;
 
-            rows.filter_map(|r| r.ok()).collect()
+            rows.collect::<rusqlite::Result<Vec<_>>>()
+                .map_err(LibreFangError::memory)?
         };
 
         Ok(results)
@@ -2849,6 +2851,25 @@ mod tests {
         // Empty query should return nothing
         let results = store.search_sessions("", None).unwrap();
         assert!(results.is_empty());
+    }
+
+    #[test]
+    fn fts_search_surfaces_corrupt_result_rows() {
+        let store = setup();
+        let agent_id = AgentId::new();
+        {
+            let conn = store.pool.get().unwrap();
+            conn.execute(
+                "INSERT INTO sessions_fts (session_id, agent_id, content) VALUES (?1, ?2, ?3)",
+                rusqlite::params![vec![0xff_u8], agent_id.0.to_string(), "corruptneedle"],
+            )
+            .unwrap();
+        }
+
+        assert!(store.search_sessions("corruptneedle", None).is_err());
+        assert!(store
+            .search_sessions("corruptneedle", Some(&agent_id))
+            .is_err());
     }
 
     /// search_sessions_paginated must:
--- a/crates/librefang-memory/src/session.rs
+++ b/crates/librefang-memory/src/session.rs
@@ -1595,8 +1595,10 @@ impl SessionStore {
                 )
                 .map_err(LibreFangError::memory)?;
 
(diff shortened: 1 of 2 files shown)


crates/librefang-memory/src/session.rs
```diff
@@ -1595,7 +1595,10 @@ impl SessionStore {
                 )
                 .map_err(LibreFangError::memory)?;
 
-            rows.filter_map(|r| r.ok()).collect()
+            rows.collect::<rusqlite::Result<Vec<_>>>().map_err(|e| {
```
houko, line 1598: Worth a maintainer sanity-check on the availability trade-off here, per CLAUDE.md's note in the review checklist about whether "a single bad row now aborts the whole search instead of just being skipped."

Before this PR, `filter_map(|r| r.ok())` meant one corrupt `sessions_fts` row silently dropped out of an otherwise-good result set — the search still returned everything else. After this PR, `search_sessions` / `search_sessions_paginated` fail closed: a single malformed row now turns the *entire* query (network-exposed via `GET /sessions/search`, see `routes/tools_sessions.rs:885`) into a 500, even when the other 999 matching rows decode fine. Since the same corrupted row will keep failing on every future search that happens to match it, this is a plausible path to a persistent DoS on the search endpoint for whatever `agent_id`/query combination touches the bad row, until an operator manually finds and purges it.

The PR body frames this as intentional ("fail-closed" per the summary), and I agree fail-closed is defensible — silently losing matches is also a real bug. But it's a genuine behavior/availability trade-off, not a pure bugfix, and I don't see it called out as such in the PR description beyond "propagate ... instead of silently omitting." Flagging for a maintainer decision: is fail-closed the desired semantics for this endpoint, or should the row-level error be logged and the row skipped (still surfacing the corruption via logs/metrics) so one bad row can't take down all other matches? Related: I pushed a follow-up commit (50ba28f) adding a `warn!(error = %e, ...)` at both row-decode failure sites, since the propagated error otherwise had no logged diagnostic trail — independent of which way this trade-off question is resolved.

---

crates/librefang-memory/src/session.rs
houko, line 1: Two commits on this branch carry an AI author identity rather than a human one.
`git log` shows commit `50ba28f` ("fix(memory): log row-decode failures in session search") and commit `a2aa7b4` ("docs(changelog): add fragment for session search row error fix") both authored by `Claude <noreply@anthropic.com>`.
The commit messages themselves are clean of attribution text, but CLAUDE.md's `commit-msg` hook is specifically designed to reject this independent of message content: it "separately rejects a commit whose author identity ... resolves to Claude / Anthropic even when the message itself is clean."
That hook only fires when `core.hooksPath` is pointed at `scripts/hooks` (`just setup`), so this likely slipped through from an environment where that wasn't configured.
Flagging rather than fixing myself: correcting this needs rewriting already-pushed history (`git commit --amend --author` / rebase) and a force-push, which is a call for the branch owner to make, not something to do unilaterally on someone else's PR.
The content of both commits looks correct on its own merits (the warn! log and the changelog fragment) — this is purely an authorship/provenance issue.

---

houko (commented):
Automated daily review pass over the diff against main.
Verified: the row-decode-error fix in `search_sessions_paginated` correctly propagates a `rusqlite::Error` via `collect::<rusqlite::Result<Vec<_>>>()` instead of silently dropping the corrupt row (replacing the old `filter_map(|r| r.ok())`), preserves row order/doesn't skip valid rows before the failure point, and is covered by a new regression test (`fts_search_surfaces_corrupt_result_rows`).
The changelog fragment at `changelog.d/fixed/7032-session-search-row-errors.md` is correctly placed and formatted per `changelog.d/README.md` (bullet body, one sentence per line, ends `(#7032) (@houko)`).
No HashMap/HashSet touching LLM-prompt-reaching data in this diff, and no route/handler changes requiring a `TestServer` test.
One finding left as an inline file comment: two commits on the branch (`50ba28f`, `a2aa7b4`) are authored by `Claude <noreply@anthropic.com>`, which is exactly what CLAUDE.md's `commit-msg` hook is meant to catch on author identity — flagging rather than fixing since correcting it needs a history rewrite and force-push that isn't mine to do on someone else's branch.
The existing thread on the fail-closed-vs-fail-skip availability trade-off for the network-exposed search endpoint is a good catch and doesn't need anything added from me.

---

PR #7033, opened 2026-08-12
fix: propagate roster storage errors

houko:
## Summary
- return typed memory errors from every SQLite-backed roster operation
- propagate roster failures through the existing channel bridge and kernel handle Result boundaries
- reject corrupt member rows instead of silently omitting them from group membership

## Verification
- `cargo test -p librefang-memory roster_store --lib`
- `cargo test -p librefang-kernel --test kernel_handle_contract_broader test_roster_roundtrip`
- `cargo clippy -p librefang-memory -p librefang-kernel -p librefang-api --all-targets -- -D warnings`
- `git diff --check`

## Out of scope
- the separate in-memory `GroupRosterStore` in librefang-channels; it does not use SQLite or share this failure mode

### DIFF
--- a/changelog.d/fixed/7033-roster-store-error-contract.md
+++ b/changelog.d/fixed/7033-roster-store-error-contract.md
@@ -1,0 +1,3 @@
+Group roster storage (`RosterStore::upsert`, `members`, `remove_member`, `member_count`) swallowed every SQLite pool-exhaustion and row-decode error, returning empty results or fixed defaults instead of failing.
+A corrupted `group_roster` row was silently dropped from `members()` rather than surfacing as a query failure, and a pool outage during `upsert` or `remove_member` looked identical to success to every caller.
+All four methods now return `LibreFangResult`, and the channel bridge and kernel handle boundaries propagate the error instead of discarding it (#7033) (@houko)
--- a/crates/librefang-api/src/channel_bridge.rs
+++ b/crates/librefang-api/src/channel_bridge.rs
@@ -1025,14 +1025,11 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
         display_name: &str,
         username: Option<&str>,
     ) -> Result<(), String> {
-        self.kernel.memory_substrate().roster().upsert(
-            channel,
-            chat_id,
-            user_id,
-            display_name,
-            username,
-        );
-        Ok(())
+        self.kernel
+            .memory_substrate()
+            .roster()
+            .upsert(channel, chat_id, user_id, display_name, username)
+            .map_err(|error| error.to_string())
     }
 
     async fn uptime_info(&self) -> String {
--- a/crates/librefang-kernel/src/kernel/handles/channel_sender.rs
+++ b/crates/librefang-kernel/src/kernel/handles/channel_sender.rs
@@ -339,8 +339,7 @@ impl kernel_handle::ChannelSender for LibreFangKernel {
         self.memory
             .substrate
             .roster()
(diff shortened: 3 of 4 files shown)


houko (commented):
Automated review pass (CLAUDE.md compliance check).

**Code changes**: look correct. `RosterStore::upsert/members/remove_member/member_count` now return `LibreFangResult`, both call-site boundaries (`crates/librefang-api/src/channel_bridge.rs`, `crates/librefang-kernel/src/kernel/handles/channel_sender.rs`) already had `Result`-returning trait signatures (`ChannelBridgeHandle::roster_upsert -> Result<(), String>`, `kernel_handle::ChannelSender::roster_* -> Result<_, KernelOpError>`) that were previously discarding the error and hard-coding `Ok(())`/`Vec::new()`/`0` — this PR wires the real error through instead, which matches the bug description. Confirmed no other call sites in the workspace call these four methods (the only other `RosterStore`-shaped type, `librefang-channels::roster::GroupRosterStore`, is a separate in-memory store and is correctly called out as out-of-scope). `LibreFangError::memory`/`memory_msg` constructors exist and match the established pattern. Tests were added for both the pool-exhaustion path and the corrupt-row path (`storage_errors_are_returned_to_callers`, `corrupt_member_rows_are_not_silently_dropped`). Changelog fragment `changelog.d/fixed/7033-roster-store-error-contract.md` is present and matches the required format (one bullet, sentence-per-line, `(#7033) (@houko)` attribution, correct `fixed/` directory).

**One governance issue I can't fix myself (need a maintainer, since I won't force-push):** the current branch tip, commit `5a070b8` ("docs: add changelog fragment for roster storage error propagation"), has author identity `Claude <noreply@anthropic.com>`. CLAUDE.md's `commit-msg` hook explicitly rejects commits whose *author identity* resolves to Claude/Anthropic, independent of message content — this looks like it landed from an earlier automated pass that had `core.hooksPath` unset (or otherwise bypassed the hook) before pushing. Since it's already on `origin/fix/roster-store-error-contract`, fixing it means rewriting history (`commit --amend --author=... ` + force-push), which is outside what I'm allowed to do here. Recommend a maintainer amend the author (or squash) and force-push before merge so this doesn't ship with AI attribution baked into the git history.

No other findings; not fixing/approving anything myself.

---

changelog.d/fixed/7033-roster-store-error-contract.md
houko, line 1: Per CLAUDE.md's prose-wrapping rule, commit message bodies must break only at sentence boundaries (one sentence per line), not hard-wrap at a column count.

The body of commit `5a070b8` ("docs: add changelog fragment for roster storage error propagation") is hard-wrapped at ~72 columns across 4 lines instead of one line per sentence:

```
PR #7033 changes RosterStore's public error contract (upsert, members,
remove_member, member_count now return LibreFangResult) without a
changelog.d fragment, unlike the identical-pattern sibling PR #6951.
Add the fragment so the behavior change reaches the release notes.
```

This should be two lines (one sentence each). Flagging rather than fixing directly since correcting it requires amending an already-pushed commit and force-pushing, which is outside what this automated pass does — up to you whether it's worth a history rewrite at this stage of review.

---

houko (commented):
Automated daily review pass against CLAUDE.md house rules. Error-contract propagation, changelog fragment format, call-site consistency (channel_bridge.rs, kernel/handles/channel_sender.rs), and prompt-ordering/route-test rules all check out — one flag below on commit-message prose wrapping.

---

PR #7034, opened 2026-08-12
fix: surface hook trace query errors

houko:
## Summary
- return SQLite and poisoned-lock failures from hook trace list, lookup, and count queries
- stop silently dropping corrupt trace rows or treating failed lookups as missing traces
- return a scrubbed HTTP 500 when the trace store cannot be opened or queried

## Verification
- `cargo test -p librefang-runtime trace_store --lib`
- `cargo clippy -p librefang-runtime -p librefang-api --all-targets -- -D warnings`
- `git diff --check`

## Out of scope
- trace insertion remains best-effort telemetry by its documented contract

### DIFF
--- a/crates/librefang-api/src/routes/plugins.rs
+++ b/crates/librefang-api/src/routes/plugins.rs
@@ -2471,11 +2471,12 @@ pub async fn get_trace_by_id(Path(trace_id): Path<String>) -> impl IntoResponse 
     }
     match librefang_kernel::plugin_manager::open_trace_store() {
         Ok(store) => match store.query_by_trace_id(&trace_id) {
-            Some(trace) => axum::Json(trace).into_response(),
-            None => ApiErrorResponse::not_found(format!("No trace found with id '{trace_id}'"))
+            Ok(Some(trace)) => axum::Json(trace).into_response(),
+            Ok(None) => ApiErrorResponse::not_found(format!("No trace found with id '{trace_id}'"))
                 .into_response(),
+            Err(error) => ApiErrorResponse::internal_scrub(error).into_response(),
         },
-        Err(e) => ApiErrorResponse::internal(e).into_response(),
+        Err(e) => ApiErrorResponse::internal_scrub(e).into_response(),
     }
 }
 
--- a/changelog.d/fixed/7034-trace-store-query-errors.md
+++ b/changelog.d/fixed/7034-trace-store-query-errors.md
@@ -1,0 +1,3 @@
+`TraceStore::query`, `query_by_trace_id`, and `count` swallowed SQLite failures and poisoned-mutex errors, returning an empty list, `None`, or `0` indistinguishably from a genuine empty result.
+A corrupt row or a failing query on the hook-trace store was therefore reported to callers as "no traces found" rather than as a failure.
+These methods now return `rusqlite::Result`, and `GET /api/context-engine/traces/:trace_id` surfaces a scrubbed HTTP 500 instead of a false 404 when the store itself fails (#7034) (@houko)
--- a/crates/librefang-runtime/src/trace_store.rs
+++ b/crates/librefang-runtime/src/trace_store.rs
@@ -4,7 +4,7 @@
 //! post-mortem analysis of hook failures without relying on the in-memory
 //! ring buffer (which resets on restart).
 
-use rusqlite::{params, Connection};
+use rusqlite::{params, Connection, OptionalExtension};
 use std::collections::HashMap;
 use std::path::Path;
 use std::sync::atomic::{AtomicU64, Ordering};
@@ -144,10 +144,11 @@ impl TraceStore {
         hook: Option<&str>,
         limit: usize,
         only_failures: bool,
-    ) -> Vec<serde_json::Value> {
-        let Ok(conn) = self.conn.lock() else {
-            return vec![];
-        };
+    ) -> rusqlite::Result<Vec<serde_json::Value>> {
+        let conn = self
+            .conn
+            .lock()
+            .map_err(|_| rusqlite::Error::InvalidParameterName("mutex poisoned".to_string()))?;
 
         // Build parameterized WHERE clause — never interpolate user values directly.
         let mut conditions: Vec<&str> = Vec::new();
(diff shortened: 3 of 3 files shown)


crates/librefang-api/src/routes/plugins.rs
```diff
@@ -2471,11 +2471,12 @@ pub async fn get_trace_by_id(Path(trace_id): Path<String>) -> impl IntoResponse
     }
     match librefang_kernel::plugin_manager::open_trace_store() {
         Ok(store) => match store.query_by_trace_id(&trace_id) {
-            Some(trace) => axum::Json(trace).into_response(),
-            None => ApiErrorResponse::not_found(format!("No trace found with id '{trace_id}'"))
+            Ok(Some(trace)) => axum::Json(trace).into_response(),
+            Ok(None) => ApiErrorResponse::not_found(format!("No trace found with id '{trace_id}'"))
                 .into_response(),
+            Err(error) => ApiErrorResponse::internal_scrub(error).into_response(),
```
houko, line 2477: Per CLAUDE.md's "MANDATORY: Integration Testing" section, a route/wiring behavior change like this should land with a `#[tokio::test]` against `TestServer`. The only existing coverage for this handler is `context_engine_get_trace_by_id_rejects_malformed_id` (`crates/librefang-api/tests/plugins_routes_integration.rs:271`), which only exercises the 400 validation path *before* `open_trace_store()`/`query_by_trace_id()` are ever called.

None of the three branches this PR actually changes — `Ok(Some(trace))` → 200, `Ok(None)` → 404, and `Err(error)` → scrubbed 500 (this line and line 2479) — has route-level coverage. In particular, there's no test proving that a query failure now surfaces as a 500 with the generic "Internal server error" body (rather than being misreported as a 404, which is the bug this PR fixes) and doesn't leak the raw SQLite error to the client.

Given the trace store here is a package-level singleton opened via `librefang_kernel::plugin_manager::open_trace_store()` rather than injected through `AppState`, forcing it into an error state from the API integration test harness may need some thought (e.g. pointing `LIBREFANG_HOME`/the trace DB path at a location that fails to open, or a corrupt DB file) — flagging as a judgment call rather than a fix I'm making myself.

---

crates/librefang-api/src/routes/plugins.rs
```diff
@@ -2471,11 +2471,12 @@ pub async fn get_trace_by_id(Path(trace_id): Path<String>) -> impl IntoResponse
     }
     match librefang_kernel::plugin_manager::open_trace_store() {
         Ok(store) => match store.query_by_trace_id(&trace_id) {
-            Some(trace) => axum::Json(trace).into_response(),
-            None => ApiErrorResponse::not_found(format!("No trace found with id '{trace_id}'"))
+            Ok(Some(trace)) => axum::Json(trace).into_response(),
```
houko, line 2474: `get_trace_by_id` now distinguishes "trace not found" (404) from "store query failed" (scrubbed 500) — previously any query failure was folded into `None` and reported as a 404. That's a real route contract change. Per CLAUDE.md's "MANDATORY: Integration Testing" section:

> **Add a `#[tokio::test]` against `TestServer`** in the matching `tests/*.rs` file. ... **Reviewers gate PRs on the presence of an integration test for each new endpoint.** PRs that change route shape without a test should be sent back.

The unit tests added in `crates/librefang-runtime/src/trace_store.rs` cover the store methods directly, but I don't see a `crates/librefang-api/tests/*.rs` case exercising this route through `TestServer` to confirm the 404-vs-500 split holds at the HTTP layer. Could one be added?

Separately: I noticed the two most recent commits on this branch (`db953324` "docs(changelog): add fragment for hook trace query error surfacing" and `6cb08d88` "docs(runtime): clarify query_by_trace_id error contract") have author/committer identity `Claude <noreply@anthropic.com>`. CLAUDE.md's "Commit & PR hygiene" section says:

> **No Claude / Anthropic / AI attribution** in commit messages, PR bodies, or issue comments. The `commit-msg` git hook rejects matching strings ... don't try to route around either — the rule exists because attribution pollutes `git log` and signals provenance the project does not want to imply.

and the repo's `commit-msg` hook specifically also checks the *author identity* (`GIT_AUTHOR_IDENT`), not just the message text, for exactly this case. Since I can't force-push to rewrite this branch's history, flagging so a maintainer can decide whether to amend the author identity before merge.

---

houko (commented):
Changelog fragment gap has since been fixed on this branch. Left one inline note about integration test coverage for the route-level error-path change, and flagged an author-identity issue on the two most recent commits.

---

crates/librefang-api/src/routes/plugins.rs
houko, line 2477: `get_trace_by_id` now distinguishes a real store error (500) from "not found" (404) instead of always returning 404 on any failure. Same as #7035: this changed route path has no TestServer-level test exercising the new 500 branch, only store-level unit tests in `trace_store.rs`. Per CLAUDE.md's mandatory integration-testing rule, worth adding one.

---

crates/librefang-api/src/routes/plugins.rs
houko, line 1: This handler's error-surfacing behavior changed (store failure now returns a scrubbed 500 instead of a false 404), but there's no `#[tokio::test]` against `TestServer` covering it — the existing `context_engine_get_trace_by_id_rejects_malformed_id` only exercises the 400 path.

Per CLAUDE.md's mandatory-integration-testing rule this is normally a blocker, but I'm flagging rather than adding one myself: `open_trace_store()` (in `librefang-runtime/src/plugin_manager.rs`) always opens `~/.librefang/hook_traces.db` at a fixed path with no injection point, so a TestServer-level reproduction of "the store query itself fails" would mean corrupting a real, shared, non-isolated path — risky under parallel test execution rather than a straightforward add.

The underlying logic is well covered by the two new unit tests added directly in `trace_store.rs` (`query_errors_are_not_reported_as_empty_results`, `corrupt_trace_rows_are_not_silently_dropped`), which exercise the same failure modes at the correct boundary. Worth a maintainer call on whether that's sufficient here, or whether `open_trace_store()` should grow a path override so this route can be exercised end-to-end too.

---

houko (commented):
Automated daily review pass. One mechanical fix pushed (prose-wrap violation in a new doc comment); one ambiguous test-coverage question left as an inline comment. Everything else checked out.

---

houko:
Correction from this automated review pass: three commits it pushed here (`6cb08d8`, `db95332`, `7c4ea84`) are themselves authored as `Claude <noreply@anthropic.com>`, which is the exact no-AI-attribution violation CLAUDE.md's `commit-msg` hook targets. This happened because the review environment's global git identity was misconfigured (not something the review logic checked for on its own commits before pushing). A maintainer should reauthor/squash those commits before merge.

---

Issue #6607, opened 2026-07-28
Approvals History renders timed_out/pending audit entries as "Edited" — timeouts are indistinguishable from human edit-then-approve


nevgenov:
The Approvals History tab labels several distinct backend decisions as "Edited", including ones that represent the exact opposite of operator involvement.

## Mechanism

`ApprovalsPage.tsx` branches only on `decision === "approved" | "approve"` and `"rejected" | "reject"`. Everything else falls through to the "Edited" branch — yellow badge, pencil icon — which is the same rendering as a genuine modify-then-approve action.

`ApprovalAuditEntry.decision` (`crates/librefang-types/src/approval.rs`) is a bare `String`, and the frontend type mirrors it as `string` rather than a union, so there is no compile-time signal for the unhandled values. The backend really does emit `timed_out`, `pending`, `modify_and_retry` and `skipped` on this field.

## Live data

On this host, `GET /api/approvals/audit` returns a 56-entry history:

```
pending:   28
timed_out: 18
approved:  10
```

46 of 56 entries — 82% of the visible audit trail — are rendered as "Edited". Two of those decisions carry meanings that matter for a security-relevant surface:

- `timed_out` means **nobody answered before the timeout expired**. Displayed as if an operator reviewed the request and modified it.
- `pending` means **the request is still awaiting a decision**. Also displayed as a completed edit.

An operator reading this history cannot tell "I changed the arguments and approved" from "this request expired unattended". For an approval audit trail that distinction is the point of the record.

## Suggested fix

Handle the remaining variants explicitly — `timed_out` (neutral/grey, clock icon), `pending` (in-progress), `modify_and_retry` (the actual "Edited" case), `skipped` — and add a default branch that renders the raw value rather than silently claiming an edit, so a future backend variant degrades to "unknown status" instead of a false one.

Narrowing the frontend type from `string` to a union of the known decisions would make the next added variant a type error instead of a silent mislabel.

## Environment

- Deployed `v2026.7.21`; `ApprovalsPage.tsx` and `approval.rs` unchanged at `v2026.7.27` and on current `main`.

Happy to test a patch.

PR #7035, opened 2026-08-12
fix: surface approval audit query errors

houko:
## Summary
- return typed memory errors from approval audit list and count queries
- stop silently dropping corrupt audit rows or returning successful empty/zero results on SQLite failures
- return a scrubbed HTTP 500 from the audit endpoint and log failures in the duplicate-resolution helper

## Verification
- `cargo test -p librefang-kernel approval_audit_ --lib`
- `cargo clippy -p librefang-kernel -p librefang-api --all-targets -- -D warnings`
- `git diff --check`

## Out of scope
- startup restoration of pending approvals and TOTP lockout state; those have separate availability and fail-closed considerations

### DIFF
--- a/changelog.d/fixed/7035-approval-audit-query-errors.md
+++ b/changelog.d/fixed/7035-approval-audit-query-errors.md
@@ -1,0 +1,2 @@
+Approval audit queries used to swallow SQLite failures and return an empty list or a zero count, which looked identical to "no audit history exists" on the dashboard and in the duplicate-resolution helper used by channel bridges.
+`query_audit` and `audit_count` on `ApprovalManager` now return a typed result, the `/api/approvals/audit` route surfaces a scrubbed HTTP 500 on failure instead of a fabricated empty page, and the channel-bridge duplicate check logs a warning and falls back to its prior no-match behaviour rather than pretending the query succeeded (#7035) (@houko)
--- a/crates/librefang-api/src/routes/approvals.rs
+++ b/crates/librefang-api/src/routes/approvals.rs
@@ -1083,16 +1083,23 @@ pub async fn audit_log(
 ) -> impl IntoResponse {
     const MAX_AUDIT_LIMIT: usize = 500;
     let limit = params.limit.min(MAX_AUDIT_LIMIT);
-    let entries = state.kernel.approvals().query_audit(
+    let entries = match state.kernel.approvals().query_audit(
         limit,
         params.offset,
         params.agent_id.as_deref(),
         params.tool_name.as_deref(),
-    );
-    let total = state
+    ) {
+        Ok(entries) => entries,
+        Err(error) => return ApiErrorResponse::internal_scrub(error).into_response(),
+    };
+    let total = match state
         .kernel
         .approvals()
-        .audit_count(params.agent_id.as_deref(), params.tool_name.as_deref());
+        .audit_count(params.agent_id.as_deref(), params.tool_name.as_deref())
+    {
+        Ok(total) => total,
+        Err(error) => return ApiErrorResponse::internal_scrub(error).into_response(),
+    };
 
     Json(serde_json::json!({
         "items": entries,
@@ -1100,6 +1107,7 @@ pub async fn audit_log(
         "offset": params.offset,
         "limit": limit,
     }))
+    .into_response()
 }
 
 /// GET /api/approvals/count — Lightweight pending count for notification badges.
--- a/crates/librefang-kernel/src/approval.rs
+++ b/crates/librefang-kernel/src/approval.rs
@@ -7,6 +7,7 @@ use librefang_types::approval::{
     ApprovalResponse, RiskLevel, SecondFactor, TimeoutFallback,
 };
 use librefang_types::capability::glob_matches;
+use librefang_types::error::{LibreFangError, LibreFangResult};
 use r2d2::Pool;
 use r2d2_sqlite::SqliteConnectionManager;
 use sha2::{Digest, Sha256};
@@ -1306,13 +1307,11 @@ impl ApprovalManager {
         offset: usize,
         agent_id: Option<&str>,
         tool_name: Option<&str>,
-    ) -> Vec<ApprovalAuditEntry> {
+    ) -> LibreFangResult<Vec<ApprovalAuditEntry>> {
         let Some(db) = &self.audit_db else {
-            return Vec::new();
+            return Ok(Vec::new());
         };
-        let Ok(conn) = db.get() else {
-            return Vec::new();
-        };
+        let conn = db.get().map_err(LibreFangError::memory)?;
 
         let mut sql = String::from(
             "SELECT id, request_id, agent_id, tool_name, description, action_summary, risk_level, decision, decided_by, decided_at, requested_at, feedback, COALESCE(second_factor_used, 0) FROM approval_audit WHERE 1=1",
@@ -1334,9 +1333,7 @@ impl ApprovalManager {
         let param_refs: Vec<&dyn rusqlite::types::ToSql> =
             params.iter().map(|p| p.as_ref()).collect();
 
-        let Ok(mut stmt) = conn.prepare(&sql) else {
-            return Vec::new();
-        };
+        let mut stmt = conn.prepare(&sql).map_err(LibreFangError::memory)?;
         let rows = stmt.query_map(param_refs.as_slice(), |row| {
             Ok(ApprovalAuditEntry {
                 id: row.get(0)?,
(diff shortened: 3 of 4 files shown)


crates/librefang-api/src/routes/approvals.rs
```diff
@@ -1083,23 +1083,31 @@ pub async fn audit_log(
...
+    let entries = match state.kernel.approvals().query_audit(
         limit,
         params.offset,
         params.agent_id.as_deref(),
         params.tool_name.as_deref(),
-    );
-    let total = state
+    ) {
+        Ok(entries) => entries,
+        Err(error) => return ApiErrorResponse::internal_scrub(error).into_response(),
```
houko, line 1093: CLAUDE.md's MANDATORY Integration Testing section requires a `#[tokio::test]` against `TestServer` for any route/wiring change, specifically to catch response-shape drift like this. This PR changes `GET /api/approvals/audit`'s failure behavior from `200` with an empty/zero payload to a scrubbed `500`, but the only new tests (`approval_audit_queries_surface_storage_errors`, `approval_audit_query_does_not_silently_drop_corrupt_rows` in `crates/librefang-kernel/src/approval.rs`) exercise `ApprovalManager` directly, not the HTTP route. `crates/librefang-api/tests/approvals_routes_integration.rs` only has `audit_empty_returns_envelope` (happy path) — there's no test asserting the route itself returns `500` with a scrubbed body (no `approval_audit`/SQL text) when the underlying query fails. Worth adding one against `TestServer`, matching the pattern reviewers gate on per CLAUDE.md.

---

crates/librefang-kernel/src/approval.rs
```diff
@@ -1354,20 +1351,21 @@ impl ApprovalManager {
                 second_factor_used: row.get::<_, bool>(12).unwrap_or(false),
             })
         });
-        match rows {
-            Ok(rows) => rows.filter_map(|r| r.ok()).collect(),
-            Err(_) => Vec::new(),
-        }
+        let rows = rows.map_err(LibreFangError::memory)?;
+        rows.collect::<rusqlite::Result<Vec<_>>>()
```
houko, line 1355: This changes `query_audit` from "drop rows that fail to deserialize, return the rest" to "any single corrupt row fails the whole query." CHANGELOG.md's #6607 entry documents the previous drop-and-continue behavior as a deliberate choice ("a strict enum would silently shorten a security audit trail on one legacy value") specifically to keep the audit trail available even when one row is malformed. With this change, one bad row (e.g. the `X'00'` blob the new test inserts into `agent_id`) now takes the *entire* audit endpoint down with a 500 — including for operators trying to review history during an incident, which is exactly when audit visibility matters most, and `audit_count` will still report a non-zero total while `query_audit` 500s, a visible inconsistency on the dashboard. Is fail-fast the intended trade-off here, or should this still return the rows it *can* parse plus a warning/log for the ones it can't (matching #6607's original rationale) rather than an all-or-nothing result?

---

crates/librefang-kernel/src/approval.rs
```diff
@@ -1306,13 +1307,11 @@ impl ApprovalManager {
...
-    ) -> Vec<ApprovalAuditEntry> {
+    ) -> LibreFangResult<Vec<ApprovalAuditEntry>> {
         let Some(db) = &self.audit_db else {
-            return Vec::new();
-        };
-        let Ok(conn) = db.get() else {
-            return Vec::new();
+            return Ok(Vec::new());
         };
+        let conn = db.get().map_err(LibreFangError::memory)?;
```
houko, line 1314: Minor naming nit: every other call site of `LibreFangError::memory(...)` in the codebase (`crates/librefang-memory/src/{substrate,session,structured,semantic}.rs`) is inside the `librefang-memory` crate — the agent memory/session/semantic SQLite substrate. This is the first use outside that crate, for the unrelated approval-audit SQLite db. It compiles fine and the client never sees it (scrubbed by `internal_scrub`), but it will render as `"Memory error: ..."` in tracing/logs and in `source()` chains for what is actually an approval-audit storage failure — worth double-checking this doesn't confuse whoever is grepping logs for "Memory error" during a memory-substrate incident vs. an approval-audit one. No dedicated storage/db error variant exists yet, so this may be an accepted trade-off — flagging for visibility rather than blocking.

---

houko (commented):
Automated CLAUDE.md-rules pass on this PR: pushed a mechanical fix for the missing changelog fragment; left three inline notes on the parts that need a maintainer call (missing HTTP-level integration test for the new error path, a fail-fast-on-corrupt-row design question weighed against the #6607 rationale, and a minor error-taxonomy naming nit).

---

crates/librefang-api/src/routes/approvals.rs
```diff
@@ -1083,23 +1083,31 @@ pub async fn audit_log(
 ) -> impl IntoResponse {
     const MAX_AUDIT_LIMIT: usize = 500;
     let limit = params.limit.min(MAX_AUDIT_LIMIT);
-    let entries = state.kernel.approvals().query_audit(
+    let entries = match state.kernel.approvals().query_audit(
         limit,
         params.offset,
         params.agent_id.as_deref(),
         params.tool_name.as_deref(),
```
houko, line 1090: `audit_log` now returns a scrubbed HTTP 500 (via `ApiErrorResponse::internal_scrub`) when `query_audit`/`audit_count` fail, instead of the previous silent-empty-list behavior — a real route contract change. Per CLAUDE.md's "MANDATORY: Integration Testing" section:

> **Add a `#[tokio::test]` against `TestServer`** in the matching `tests/*.rs` file. ... **Reviewers gate PRs on the presence of an integration test for each new endpoint.** PRs that change route shape without a test should be sent back.

The unit tests added in `crates/librefang-kernel/src/approval.rs` (`approval_audit_queries_surface_storage_errors`, `approval_audit_query_does_not_silently_drop_corrupt_rows`) cover `ApprovalManager` directly but not the `GET /api/approvals/audit` route's new 500 response shape/status through `TestServer`. Could you add a `crates/librefang-api/tests/*.rs` case that drops the audit table (or similar) and asserts the route surfaces a 500 rather than a fabricated empty page? #6971 has a good template for this pattern.

---

houko (commented):
Changelog fragment already present — left one inline note about integration test coverage for the route-level error-path change.

---

crates/librefang-api/src/routes/approvals.rs
houko, line 1093: `audit_log` now returns HTTP 500 on a DB query failure instead of silently returning 200 with whatever entries happened to load. Per CLAUDE.md's mandatory integration-testing rule ("any new/changed HTTP route needs a `#[tokio::test]` against TestServer"), this changed route behavior only has kernel-level unit test coverage (`approval.rs`), not a TestServer-level regression test exercising the new 500 path. Worth confirming whether one exists elsewhere, or adding one before merge.

---

changelog.d/fixed/7035-approval-audit-query-errors.md
houko, line 1: Commit `b10b531` (`docs: add changelog fragment for approval audit query error surfacing`) has author identity `Claude <noreply@anthropic.com>`.
Per this repo's own CLAUDE.md, no AI/Claude/Anthropic attribution is allowed in commit history — including author identity, not just the message body (that's exactly what the `commit-msg` hook's `GIT_AUTHOR_IDENT` check is meant to catch).
Could you amend this commit's author before merge (e.g. `git commit --amend --author "houko <...>"` plus a rebase), so it lands under a human author identity?

---

crates/librefang-api/src/routes/approvals.rs
```diff
@@ -1083,23 +1083,31 @@ pub async fn audit_log(
...
+        Err(error) => return ApiErrorResponse::internal_scrub(error).into_response(),
+    };
+    let total = match state
         .kernel
         .approvals()
-        .audit_count(params.agent_id.as_deref(), params.tool_name.as_deref());
+        .audit_count(params.agent_id.as_deref(), params.tool_name.as_deref())
+    {
+        Ok(total) => total,
+        Err(error) => return ApiErrorResponse::internal_scrub(error).into_response(),
```
houko, line 1101: The kernel-level coverage here is solid (`approval_audit_queries_surface_storage_errors`, `approval_audit_query_does_not_silently_drop_corrupt_rows` in `crates/librefang-kernel/src/approval.rs`), but per CLAUDE.md's "MANDATORY: Integration Testing" section this is a route/wiring behavior change (the 500-on-failure path through `audit_log` is new) and doesn't have a matching `#[tokio::test]` against `TestServer` in `crates/librefang-api/tests/approvals_routes_integration.rs`.
`totp_flow_test.rs` already has a precedent for simulating a DB failure via `DROP TABLE ...` inside a `TestServer`-backed test — the same pattern (drop `approval_audit`, hit `GET /api/approvals/audit`, assert 500) would close the gap here.
Not blocking if you consider the kernel-level tests sufficient, but wanted to flag it since the mandatory-testing doc calls this out specifically for route changes.

---

houko (commented):
Automated daily review pass. The core fix is solid and well-tested at the kernel level (proper `LibreFangResult` propagation replacing the old `filter_map(|r| r.ok())` silent-drop of corrupt audit rows, and the `unwrap_or(0)`/`Vec::new()` swallowing in `query_audit`/`audit_count`); the API route change correctly propagates via `ApiErrorResponse::internal_scrub`, matching the convention used throughout the rest of the route modules. Changelog fragment is present, correctly sectioned under `fixed/`, and passes `check-changelog-attribution.py`. Two things flagged inline: an AI-attributed commit author identity on the branch, and a coverage gap between the new kernel-level error tests and the route's new error-response behavior. Neither blocks understanding the change, but both are worth a look before merge.

---

PR #7036, opened 2026-08-12
fix(kernel,skills): resolve an empty parent to the current directory in atomic writers

houko:
## Summary

`Path::parent()` returns `Some("")` for a bare relative filename like `agent.toml` — `None` only happens for `/` or an empty path itself.
Three atomic writers added over the last day treat that empty-but-present case as an error via `ok_or_else`, and the consequence differs by what each one does with the parent.

**`cron_script::atomic_write_toml` (#6948) — a real error path.**
The parent is opened for the post-rename directory fsync, so an empty parent reaches `File::open("")` and fails with ENOENT *after* the rename already succeeded — reporting failure for a write that actually landed on disk.
Every current caller (`agent_runtime.rs:276`, `agent_state.rs:72`, `workspace_setup.rs:518`) passes an absolute path built with `join`, so this is latent rather than live today.

**`skillhub::atomic_write_manifest` (#6995) and `evolution::atomic_write` — silently accidental.**
Here the parent only anchors the staging file next to the target.
An empty parent still "works", because joining onto it resolves relative to the process directory and the rename then happens inside that same directory.
The same-directory invariant that makes the rename atomic therefore holds by accident, and only for as long as the target path is also relative.

All three now resolve an empty parent to `.`, which is what `librefang_api::atomic_write` (#6942) already did — it hit this case first and documented it. This PR brings the other three in line rather than inventing a new convention.

## Changes

- `cron_script.rs`: extract `parent_dir_for_fsync` and use it for the post-rename fsync
- `skillhub.rs`, `evolution.rs`: resolve an empty parent to `.` before joining the staging filename
- `tests.rs`: unit-test the resolution across bare filename / rootless / absolute / nested-relative inputs

The kernel resolution is extracted into a function specifically so it can be unit-tested **without** `set_current_dir`, which is process-global and would race the rest of the suite under the default multi-threaded test runner.

## Verification

Run in the repo's `Dockerfile.rust-dev` image against a per-worktree target volume:

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p librefang-kernel -p librefang-skills --all-targets -- -D warnings` — silent
- `cargo test -p librefang-kernel --lib parent_dir_for_fsync` — 1 passed
- `cargo test -p librefang-kernel --lib atomic_write` — 3 passed (pre-existing atomicity/concurrency tests unaffected)
- `cargo test -p librefang-skills --lib` — 240 passed

Mutation-checked: reverting the fallback to `Path::new("")` — i.e. the pre-fix behaviour — turns the new test red, so it genuinely exercises the resolution rather than passing incidentally.

## Out of scope

`librefang_api::atomic_write` is unchanged; it already resolves this correctly. Its inline resolution is left inline rather than shared, since the two crates have no natural common dependency for a helper this small — the kernel copy carries a comment pointing at it.

### DIFF
--- a/crates/librefang-kernel/src/kernel/cron_script.rs
+++ b/crates/librefang-kernel/src/kernel/cron_script.rs
@@ -196,14 +196,28 @@ pub(super) fn atomic_write_toml(path: &Path, content: &str) -> std::io::Result<(
     }
 
     #[cfg(unix)]
-    path.parent()
-        .ok_or_else(|| {
-            std::io::Error::new(std::io::ErrorKind::InvalidInput, "missing parent directory")
-        })
-        .and_then(std::fs::File::open)?
-        .sync_all()?;
+    std::fs::File::open(parent_dir_for_fsync(path))?.sync_all()?;
 
     Ok(())
+}
+
+/// Resolve the directory whose entry must be fsynced after replacing `path`.
+///
+/// `Path::parent()` returns `Some("")` — not `None` — for a bare relative
+/// filename like `agent.toml`; `None` only happens for `/` or an empty path
+/// itself. Treating the empty-but-present case as an error would fail
+/// `File::open("")` with ENOENT *after* the rename already succeeded,
+/// reporting failure for a write that actually landed on disk. Mapping it to
+/// `.` fsyncs the real containing directory instead.
+///
+/// This mirrors the same resolution in `librefang_api::atomic_write`; keep the
+/// two in step if either changes.
+#[cfg(unix)]
+pub(super) fn parent_dir_for_fsync(path: &Path) -> &Path {
+    match path.parent() {
+        Some(parent) if !parent.as_os_str().is_empty() => parent,
+        _ => Path::new("."),
+    }
 }
 
 /// Parse the wake gate from script stdout.
--- a/crates/librefang-kernel/src/kernel/cron_script.rs
+++ b/crates/librefang-kernel/src/kernel/cron_script.rs
@@ -203,15 +203,11 @@ pub(super) fn atomic_write_toml(path: &Path, content: &str) -> std::io::Result<(
 
 /// Resolve the directory whose entry must be fsynced after replacing `path`.
 ///
-/// `Path::parent()` returns `Some("")` — not `None` — for a bare relative
-/// filename like `agent.toml`; `None` only happens for `/` or an empty path
-/// itself. Treating the empty-but-present case as an error would fail
-/// `File::open("")` with ENOENT *after* the rename already succeeded,
-/// reporting failure for a write that actually landed on disk. Mapping it to
-/// `.` fsyncs the real containing directory instead.
+/// `Path::parent()` returns `Some("")` — not `None` — for a bare relative filename like `agent.toml`; `None` only happens for `/` or an empty path itself.
+/// Treating the empty-but-present case as an error would fail `File::open("")` with ENOENT *after* the rename already succeeded, reporting failure for a write that actually landed on disk.
+/// Mapping it to `.` fsyncs the real containing directory instead.
 ///
-/// This mirrors the same resolution in `librefang_api::atomic_write`; keep the
-/// two in step if either changes.
+/// This mirrors the same resolution in `librefang_api::atomic_write`; keep the two in step if either changes.
 #[cfg(unix)]
 pub(super) fn parent_dir_for_fsync(path: &Path) -> &Path {
     match path.parent() {
--- a/crates/librefang-skills/src/evolution.rs
+++ b/crates/librefang-skills/src/evolution.rs
@@ -270,14 +270,7 @@ static ATOMIC_WRITE_COUNTER: std::sync::atomic::AtomicU64 = std::sync::atomic::A
 /// a nanosecond timestamp so collisions are extremely unlikely even
 /// under concurrent callers targeting the same final path.
 fn atomic_write(path: &Path, content: &str) -> Result<(), SkillError> {
-    // `Path::parent()` yields `Some("")` for a bare filename, so the staging
-    // file has to be anchored at `.` rather than joined onto an empty path —
-    // otherwise the "same directory as the target" invariant that keeps the
-    // rename atomic holds only by accident.
-    let parent = match path.parent() {
-        Some(parent) if !parent.as_os_str().is_empty() => parent,
-        _ => Path::new("."),
-    };
+    let parent = crate::resolve_parent_or_cwd(path);
     std::fs::create_dir_all(parent)?;
 
     // Keep the thread-id string to ASCII-safe chars — `ThreadId`'s
--- a/crates/librefang-skills/src/evolution.rs
+++ b/crates/librefang-skills/src/evolution.rs
@@ -270,9 +270,14 @@ static ATOMIC_WRITE_COUNTER: std::sync::atomic::AtomicU64 = std::sync::atomic::A
 /// a nanosecond timestamp so collisions are extremely unlikely even
 /// under concurrent callers targeting the same final path.
 fn atomic_write(path: &Path, content: &str) -> Result<(), SkillError> {
-    let parent = path
-        .parent()
-        .ok_or_else(|| SkillError::Io(std::io::Error::other("no parent directory")))?;
+    // `Path::parent()` yields `Some("")` for a bare filename, so the staging
+    // file has to be anchored at `.` rather than joined onto an empty path —
+    // otherwise the "same directory as the target" invariant that keeps the
+    // rename atomic holds only by accident.
+    let parent = match path.parent() {
+        Some(parent) if !parent.as_os_str().is_empty() => parent,
+        _ => Path::new("."),
+    };
     std::fs::create_dir_all(parent)?;
 
     // Keep the thread-id string to ASCII-safe chars — `ThreadId`'s
--- a/crates/librefang-skills/src/lib.rs
+++ b/crates/librefang-skills/src/lib.rs
@@ -23,7 +23,22 @@ pub mod verify;
 
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+
+/// Resolve the directory that a same-directory atomic-write staging file should be anchored in for a write targeting `path`.
+///
+/// `Path::parent()` returns `Some("")` — not `None` — for a bare relative filename like `skill.toml`; `None` only happens for `/` or an empty path itself.
+/// Joining a staging-file name onto `""` happens to resolve against the process's current directory, which is the same directory the final `rename` targets — but only by accident.
+/// Mapping the empty-but-present case to `.` makes that same-directory invariant hold explicitly instead of by coincidence.
+///
+/// Shared by `skillhub::atomic_write_manifest` and `evolution::atomic_write`, the two same-crate call sites that stage a temp file beside their target.
+/// Mirrors the equivalent resolution in `librefang_kernel::kernel::cron_script::parent_dir_for_fsync` (which fsyncs the parent after rename rather than staging inside it) and in `librefang_api`'s atomic writer; keep all three in step if any changes.
+pub(crate) fn resolve_parent_or_cwd(path: &Path) -> &Path {
+    match path.parent() {
+        Some(parent) if !parent.as_os_str().is_empty() => parent,
+        _ => Path::new("."),
+    }
+}
 
 /// Errors from the skill system.
 #[derive(Debug, thiserror::Error)]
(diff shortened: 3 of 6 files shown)


crates/librefang-skills/src/skillhub.rs
```diff
@@ -33,9 +33,14 @@ fn atomic_write_manifest(path: &Path, contents: &[u8]) -> Result<(), SkillError>
...
-    })?;
+    // `Path::parent()` yields `Some("")` for a bare filename, so the staging
+    // file has to be anchored at `.` rather than joined onto an empty path —
+    // otherwise the "same directory as the target" invariant that keeps the
+    // rename atomic holds only by accident.
+    let parent = match path.parent() {
+        Some(parent) if !parent.as_os_str().is_empty() => parent,
+        _ => Path::new("."),
+    };
     let seq = SEQ.fetch_add(1, Ordering::Relaxed);
```
houko, line 37: Per CLAUDE.md's `MANDATORY: Integration Testing` guidance and the platform-independent-regression-test note under "Verifying without a native toolchain" (resolve a nonexistent/synthetic path rather than relying on a host path — see #5716), this fix has no regression test.

`cron_script::atomic_write_toml`'s equivalent fix was extracted into `parent_dir_for_fsync` specifically so it's unit-testable without mutating the process cwd (per this PR's own commit message: "extracted... so the behaviour is unit-testable without mutating the process working directory, which would race other tests"). This function inlines the identical `match` instead, so there's no way to unit-test the bare-filename branch here without either mutating cwd (racy, exactly what the kernel change avoided) or extracting an equivalent testable helper.

Same gap in `evolution.rs`'s `atomic_write` (identical inlined match, no test of the empty-parent branch — the pre-existing `test_atomic_write` only exercises an absolute `TempDir`-joined path).

Worth extracting a small `fn resolve_parent_or_cwd(path: &Path) -> &Path` here (and reusing it from `evolution.rs`, since both live in `librefang-skills`) so the exact behaviour this PR introduces is actually covered, the way `cron_script.rs` did it in the kernel crate.

---

crates/librefang-skills/src/evolution.rs
```diff
@@ -270,9 +270,14 @@ static ATOMIC_WRITE_COUNTER: std::sync::atomic::AtomicU64 = std::sync::atomic::A
...
-        .parent()
-        .ok_or_else(|| SkillError::Io(std::io::Error::other("no parent directory")))?;
+    // `Path::parent()` yields `Some("")` for a bare filename, so the staging
+    // file has to be anchored at `.` rather than joined onto an empty path —
+    // otherwise the "same directory as the target" invariant that keeps the
+    // rename atomic holds only by accident.
+    let parent = match path.parent() {
+        Some(parent) if !parent.as_os_str().is_empty() => parent,
+        _ => Path::new("."),
+    };
```
houko: Same inlined-match-with-no-test gap flagged on `skillhub.rs::atomic_write_manifest` — see that comment. This is a literal duplicate of that match block (down to the comment wording); consider a single shared helper in this crate covered by one unit test instead of two untested copies.

---

houko (commented):
Automated review pass against CLAUDE.md. Commit messages, changelog fragment format/location, and prose wrapping all check out, and no AI/Claude/Anthropic attribution anywhere in the diff. Two inline notes on test coverage for the fix itself (see line comments on `skillhub.rs` and `evolution.rs`).

One broader finding, not fixed here because it crosses crate boundaries outside this PR's declared scope (kernel, skills) — CLAUDE.md's "fix what you found" rule says to surface that trade-off rather than decide unilaterally:

The identical `path.parent().ok_or_else(...)` bug this PR fixes — treating `Path::parent() == Some("")` for a bare filename as a hard error, instead of resolving it to `.` — also exists, unfixed, in at least five other atomic-writer call sites this PR's description doesn't mention (it only scopes to the three writers "added over the last day" plus `librefang_api::atomic_write` as the reference implementation):

- `crates/librefang-hands/src/registry.rs:1589` (`atomic_write_json`) — same "anchor staging file beside target" shape as `skillhub`/`evolution`.
- `crates/librefang-runtime/src/apply_patch.rs:27` (`atomic_write`, async) — same shape.
- `crates/librefang-llm-drivers/src/shared_rate_guard.rs:205` (`write_atomic`) — same shape; doc comment literally says "temp file in the same dir."
- `crates/librefang-api/src/routes/sidecar_toml.rs:152` (`atomic_write`) — same shape, and its own comment already flags it as a known, tracked defect class: `// Same defect class as secrets_env::upsert_secret (T3.1)`.
- `crates/librefang-api/src/routes/secrets_env.rs:84` (`upsert_secret`) — the T3.1 sibling referenced above.

All five build the eventual path via `.join()` onto a base that's currently always absolute in every caller I could find, so — like `cron_script::atomic_write_toml` before this PR — these are latent rather than live today. But that's exactly the risk profile this PR is closing for the other three writers, and the `sidecar_toml.rs` comment shows the project is already aware of the pattern under a "T3.1" label without having swept it. Given the PR title says "atomic writers" generically rather than naming the three files, worth deciding explicitly whether to fold these five into this PR, file a tracked follow-up issue referencing T3.1, or leave them — rather than have them stay silently missed.

---

crates/librefang-skills/src/evolution.rs
```diff
@@ -270,9 +270,14 @@ static ATOMIC_WRITE_COUNTER: std::sync::atomic::AtomicU64 = std::sync::atomic::A
...
     std::fs::create_dir_all(parent)?;
```
houko, line 275: Two things worth a second look here, both low-confidence enough that I'm flagging rather than pushing a fix.

1. **Test coverage gap vs. the kernel fix.** `parent_dir_for_fsync` in `cron_script.rs` gets a dedicated unit test (`parent_dir_for_fsync_maps_bare_filename_to_current_dir`) covering bare filename / rootless / absolute / nested-relative inputs. `atomic_write` here and `atomic_write_manifest` in `skillhub.rs` get the identical resolution logic inlined but no equivalent regression test — the PR description's "mutation-checked" claim (reverting the fallback turns the kernel test red) doesn't have a counterpart proving the same for these two. Worth a `test_atomic_write_bare_relative_filename`-style case in each, so a future edit that reintroduces `ok_or_else` here fails a test instead of shipping silently.

2. **Sibling call site with the same shape, same file.** `acquire_skill_lock` at line 224 (`skill_dir.parent().ok_or_else(...)`) has the identical "bare relative path → `Some(\"\")` → treated as hard error" issue this PR fixes for `atomic_write` just above it. It's a lock-file-placement helper rather than a rename, and the PR description scopes this fix to "three atomic writers added over the last day" (`acquire_skill_lock` predates that), so I'm not confident it belongs in this PR vs. a fast follow-up — but since it's in the exact file already being touched for the exact same bug class, wanted it on your radar rather than filed separately and forgotten.

---

houko (commented):
Core fix looks correct: `Some("")` → `.` matches the precedent in `librefang_api::atomic_write`, and the reasoning in the PR body about which of the three sites is latent vs. actually-load-bearing checks out against the current callers.

One inline note on `evolution.rs` about test coverage and a same-file sibling (`acquire_skill_lock`).

One more, out-of-diff so I couldn't attach it inline: `crates/librefang-kernel/src/kernel/agent_state.rs::persist_manifest_to_disk` (line 59) resolves `toml_path.parent()` via a `Some(d)/None` match rather than `ok_or_else`, so it doesn't hit the `None` arm for a bare relative filename — it falls through to `Some(d)` with `d` an empty `PathBuf`, then `create_dir_all(&dir)` on that. Same root cause as the three sites this PR fixes, different shape, same crate already being touched. Flagging for a scope call rather than asserting it needs fixing here — I haven't verified what `create_dir_all("")` actually does.

---

houko:
Addressed the two open review threads (duplicated inline `match` in `skillhub.rs` / `evolution.rs`, no unit coverage of the empty-parent branch) in b02f65ce5: extracted a single `pub(crate) resolve_parent_or_cwd` in `crates/librefang-skills/src/lib.rs`, used by both `skillhub::atomic_write_manifest` and `evolution::atomic_write`, with a unit test (`resolve_parent_or_cwd_maps_bare_filename_to_current_dir`) covering a bare filename, a rootless path, an absolute path, and a relative path with a directory component — same shape as `cron_script::parent_dir_for_fsync`'s existing test.
Verified with `cargo check -p librefang-skills --lib` (exit 0) and `cargo test -p librefang-skills --lib resolve_parent_or_cwd` (1 passed).

---

houko (commented):
Self-flag: commit b02f65ce5949ccb6f7c195a404b6e252921a6823 ("unify the empty-parent atomic-write fix into one helper") was pushed by this same automated review run, but its git author identity resolved to `Claude <noreply@anthropic.com>` — the same AI-attribution violation on commit b4e7e935 already present on this branch from an earlier pass. Both stem from an environment-wide issue in this review session (confirmed via `git config user.name`/`user.email`), not specific to this PR. Per this routine's hard rule against force-pushing, I'm not amending/rewriting either commit myself — flagging them here so a maintainer can rebase with correct authorship before merge. The fix content itself (extracting `resolve_parent_or_cwd` into `librefang-skills/src/lib.rs` and using it from both `skillhub::atomic_write_manifest` and `evolution::atomic_write`, with a new unit test) looks correct on inspection and addresses the two previously-open review threads on this PR. No further automated commits are being made this run until the underlying git identity is fixed.

---

crates/librefang-skills/src/lib.rs
houko, line 1: Two commits on this branch have an author identity of `Claude <noreply@anthropic.com>` (`b4e7e9354` "docs: join wrapped sentences in parent_dir_for_fsync doc comment" and `b02f65ce5` "refactor(skills): unify the empty-parent atomic-write fix into one helper").
CLAUDE.md's no-AI-attribution rule covers author identity specifically (the `commit-msg` hook checks `GIT_AUTHOR_IDENT`, not just the message body), so these violate it even though the commit messages themselves are clean.
Recommend rebasing both commits with a non-AI author identity (e.g. `git rebase -i` + `git commit --amend --author=...`) before merge; not doing it here since that requires a history rewrite and this review only pushes additive fixes.

Separately, while reviewing this branch I found the same wrapped-sentence issue this PR series is meant to be fixing, still present in the new kernel test added by the original fix commit (`crates/librefang-kernel/src/kernel/tests.rs`, `parent_dir_for_fsync_maps_bare_filename_to_current_dir`) — its explanatory comment was hard-wrapped across four lines for one sentence. Joined it to one line in `d82d91858` (comment-only change, no behavior difference).

---

crates/librefang-kernel/src/kernel/cron_script.rs
```diff
@@ -196,16 +196,26 @@ pub(super) fn atomic_write_toml(path: &Path, content: &str) -> std::io::Result<(
...
     Ok(())
 }
 
+/// Resolve the directory whose entry must be fsynced after replacing `path`.
+///
+/// `Path::parent()` returns `Some("")` — not `None` — for a bare relative filename like `agent.toml`; `None` only happens for `/` or an empty path itself.
+/// Treating the empty-but-present case as an error would fail `File::open("")` with ENOENT *after* the rename already succeeded, reporting failure for a write that actually landed on disk.
+/// Mapping it to `.` fsyncs the real containing directory instead.
+///
+/// This mirrors the same resolution in `librefang_api::atomic_write`; keep the two in step if either changes.
```
houko, line 210: Widening the sibling-call-site search beyond the three writers this PR targets, `crates/librefang-hands/src/registry.rs`'s `atomic_write_json` (lines 1583-1637) has the same shape as the pre-fix `atomic_write_toml` here: `path.parent().ok_or_else(|| ... "missing parent directory")` (line 1589, literally the same message text the pre-fix kernel code used) feeding both `create_dir_all(parent)` and, after the rename, `File::open(parent)` for the durability fsync (line 1631) — the exact post-rename-fsync-via-parent shape that made the kernel case a real error path rather than a latent one.

Current callers (`hands_lifecycle.rs:756/763`, always `home_dir_boot.join("data").join("hand_state.json")`) pass an absolute path, so like the callers cited in this PR's description this is latent, not live — but it's the same defect class in a fourth crate this PR doesn't touch.

`crates/librefang-llm-drivers/src/shared_rate_guard.rs`'s `write_atomic` (line 205) has the narrower `ok_or_else`-for-staging-join variant (same shape as the pre-fix `skillhub.rs`/`evolution.rs`), also latent given `lockout_path()` always returns an absolute path.

Not asking for a fix here — different crates, and the PR body already explains why `librefang_api`'s copy stays inline rather than shared — just flagging so these don't get missed if a follow-up sweep happens.

---

crates/librefang-skills/src/lib.rs
```diff
@@ -23,7 +23,22 @@ pub mod verify;
...
+use std::path::{Path, PathBuf};
+
+/// Resolve the directory that a same-directory atomic-write staging file should be anchored in for a write targeting `path`.
+///
+/// `Path::parent()` returns `Some("")` — not `None` — for a bare relative filename like `skill.toml`; `None` only happens for `/` or an empty path itself.
+/// Joining a staging-file name onto `""` happens to resolve against the process's current directory, which is the same directory the final `rename` targets — but only by accident.
+/// Mapping the empty-but-present case to `.` makes that same-directory invariant hold explicitly instead of by coincidence.
+///
+/// Shared by `skillhub::atomic_write_manifest` and `evolution::atomic_write`, the two same-crate call sites that stage a temp file beside their target.
+/// Mirrors the equivalent resolution in `librefang_kernel::kernel::cron_script::parent_dir_for_fsync` (which fsyncs the parent after rename rather than staging inside it) and in `librefang_api`'s atomic writer; keep all three in step if any changes.
```
houko, line 35: Two more sibling call sites, both in the same crate as the already-correct `librefang_api::atomic_write` this doc comment points at, and both with the exact `ok_or(...)?`-on-`parent()` shape this PR fixes elsewhere — but neither reuses that crate's own correct helper:

- `crates/librefang-api/src/routes/secrets_env.rs:84` — `path.parent().ok_or("secrets path has no parent dir")?`, used to anchor the staging tempfile (same shape as pre-fix `skillhub.rs`/`evolution.rs`).
- `crates/librefang-api/src/routes/sidecar_toml.rs:152` — `path.parent().ok_or("config path has no parent")?`, same shape; its own doc comment even cross-references `secrets_env::upsert_secret` as "same defect class" (re: the tempfile-naming race, not this issue, but shows the two are already known to be parallel implementations).

Both callers currently pass absolute paths under `~/.librefang/`, so — like everything else audited in this PR — latent rather than live. Given they sit in the same crate as the reference implementation, they seem like the easiest of the leftover sites to actually fix (call `crate::atomic_write` instead of hand-rolling), but leaving that judgment to you since it's outside this PR's stated scope (three writers added in the last day).

---

PR #7037, opened 2026-08-12
fix: surface budget status query errors

houko:
## Summary
- make the shared metering budget snapshot return usage-store query failures instead of zero spend
- return scrubbed HTTP 500 responses from budget and health endpoints when the snapshot is unavailable
- return explicit unavailable messages from WebSocket and channel budget commands instead of misleading zero values

## Verification
- `cargo test -p librefang-kernel-metering budget_status_surfaces_usage_query_errors --lib`
- `cargo clippy -p librefang-kernel-metering -p librefang-api --all-targets -- -D warnings`
- `git diff --check`

## Out of scope
- per-agent and per-provider route queries are covered separately by #6971


crates/librefang-api/src/routes/budget.rs
```diff
@@ -332,11 +332,14 @@ pub async fn usage_daily(State(state): State<Arc<AppState>>) -> impl IntoRespons
     )
 )]
 pub async fn budget_status(State(state): State<Arc<AppState>>) -> impl IntoResponse {
```
houko, line 334: This changes `GET /api/budget`'s (and `PUT /api/budget`'s, and `/api/system/health/detail`'s) error-path contract: a `budget_status` failure now returns a scrubbed HTTP 500 instead of a 200 with fabricated zero-spend data. Per CLAUDE.md's "MANDATORY: Integration Testing" section:

> **Add a `#[tokio::test]` against `TestServer`** in the matching `tests/*.rs` file. ... This is the canonical replacement for the old curl checklist — it catches missing `server.rs` registrations, un-deserialized config fields, kernel↔API type drift, and empty/null payloads.
> **Reviewers gate PRs on the presence of an integration test for each new endpoint.** PRs that change route shape without a test should be sent back.

I don't see a corresponding test added in `crates/librefang-api/tests/` (e.g. a drop-table regression on `/api/budget`, `/api/budget` PUT, and `/api/system/health/detail` similar to the ones added in #6971 for the usage/budget endpoints). Could you add one covering the new 500 path for at least one of these three routes?

---

houko (commented):
Missing changelog.d fragment fixed directly (pushed). Left one inline note about test coverage for the route-level error-path changes.

---

crates/librefang-api/src/routes/budget.rs
```diff
@@ -332,11 +332,14 @@ pub async fn usage_daily(State(state): State<Arc<AppState>>) -> impl IntoRespons
...
-    let status = state
+    match state
         .kernel
         .metering_ref()
-        .budget_status(&state.kernel.budget_config());
-    Json(status)
+        .budget_status(&state.kernel.budget_config())
+    {
+        Ok(status) => Json(status).into_response(),
+        Err(error) => ApiErrorResponse::internal_scrub(error).into_response(),
```
houko, line 341: `budget_status`, `update_budget`, and `health_detail` gained a new HTTP-visible error branch (500 on a metering query failure, instead of silently reporting $0 spend) with no `#[tokio::test]` against `TestServer` asserting the new status/response shape — only a unit test at the `MeteringEngine` layer was added. Per CLAUDE.md's mandatory route-integration-test rule, worth adding one covering at least one of these three routes' failure path.

---

crates/librefang-api/src/ws.rs
```diff
@@ -1816,7 +1816,18 @@ async fn handle_command(
...
+            let status = match state.kernel.metering_ref().budget_status(&budget) {
+                Ok(status) => status,
+                Err(error) => {
+                    tracing::error!(%error, "failed to query budget status");
+                    return serde_json::json!({
+                        "type": "command_result",
+                        "command": cmd,
+                        "message": "Budget status is temporarily unavailable.",
+                        "error": true,
+                    });
```
houko, line 1826: Every other genuine failure branch in this same `match handle_command` block (e.g. the `"usage"` and `"context"` arms just above, and the `"model"`/`"reset"`/`"reboot"` arms elsewhere in the file) reports failure as `{"type": "error", "content": "..."}`. This arm instead emits `{"type": "command_result", ..., "message": ..., "error": true}` — an invented `error` flag that no consumer reads, on a `type` value that every existing client treats as success. `ChatPage.tsx`'s `data.type === "command_result" || data.type === "error"` check happens to render both the same way today, but that's incidental, not a contract: any WS client (or future dashboard change) that branches on `type` to decide error styling/handling will silently miss this case. Recommend matching the established convention:
```rust
return serde_json::json!({"type": "error", "content": "Budget status is temporarily unavailable."});
```

---

crates/librefang-api/src/channel_bridge.rs
```diff
@@ -2179,7 +2179,13 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
 
     async fn budget_text(&self) -> String {
         let budget = self.kernel.budget_config();
-        let status = self.kernel.metering_ref().budget_status(&budget);
+        let status = match self.kernel.metering_ref().budget_status(&budget) {
+            Ok(status) => status,
+            Err(error) => {
+                tracing::error!(%error, "failed to query channel budget status");
```
houko, line 2185: Per CLAUDE.md's "MANDATORY: Integration Testing" section, a route/wiring change needs a `#[tokio::test]` exercising it. I pushed a test covering the `GET /api/budget` HTTP path (`budget_status_returns_500_when_usage_query_fails` in `crates/librefang-api/tests/budget_routes_test.rs`), which also validates the shared `MeteringEngine::budget_status` error plumbing this function depends on. This channel-bridge branch and the WS `"budget"` command branch (`ws.rs`) still have no direct test asserting they return the "temporarily unavailable" text instead of panicking or reverting to zero-spend — worth a follow-up test if there's a harness pattern for `ChannelBridgeHandle`/WS commands already in this test suite (didn't attempt it here since it wasn't obviously mechanical to wire up without checking whether `ws.rs`/`channel_bridge.rs` have existing integration coverage patterns to follow).

---

houko (commented):
Automated review pass (CLAUDE.md compliance + correctness). Pushed one fix (missing HTTP-level integration test for the new budget-status error path); the pre-existing changelog-fragment gap was already closed by a concurrent commit on this branch. Two judgment-call findings left as inline comments — a WS response-shape inconsistency and untested WS/channel-bridge error branches.

---

houko (commented):
Self-flag: commits 97bfbd9956b186171929ef2c3e5a3016c753853d ("fix(ws): use the standard error frame shape for budget failures") and 2483083de24b11feffc447bf98535561c0badffb ("test: cover PUT /api/budget and health/detail budget query failures") were pushed by this same automated review run in response to review threads already open on this PR — but both commits' git author identity resolved to `Claude <noreply@anthropic.com>`, the AI-attribution violation CLAUDE.md's `commit-msg` hook exists to reject. This is an environment-wide issue in this review session (confirmed via `git config user.name`/`user.email`), not specific to this PR, and it's also present on an earlier commit on this same branch (83afbcc8ab7332c79bfbec2409f1947a990be747). Per this routine's hard rule against force-pushing, I'm not amending/rewriting any of these commits myself — flagging them here so a maintainer can rebase with correct authorship before merge. The fix content itself (matching the established `{"type":"error","content":...}` WS error shape, and adding the two missing integration tests for `PUT /api/budget` and `/api/system/health/detail`) looks correct on inspection. No further automated commits are being made this run until the underlying git identity is fixed.

---

PR #7039, opened 2026-08-12
perf(api): offload hand lifecycle I/O

houko:
## Summary

- dispatch hand activation, pause, resume, deactivation, and reload through Tokio’s blocking pool
- preserve existing successful responses and business-error status codes
- return scrubbed 500 responses when a blocking lifecycle task itself fails
- add a current-thread Tokio regression test proving lifecycle work does not occupy the async worker

## Tests

- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-api lifecycle_job_does_not_block_the_async_worker --lib`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-api --test hands_routes_integration unknown_instance`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-api --test hands_routes_integration activate_unknown_hand_returns_400`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-api --test hands_routes_integration reload_hands_returns_counts_envelope`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo clippy -p librefang-api --all-targets -- -D warnings`
- `git diff --check`

## Out of scope

- non-lifecycle Hands endpoints remain unchanged

### DIFF
--- a/changelog.d/changed/7039-hands-lifecycle-io-off-thread.md
+++ b/changelog.d/changed/7039-hands-lifecycle-io-off-thread.md
@@ -1,0 +1,3 @@
+`POST /api/hands/{id}/pause|resume|deactivate` and `POST /api/hands/reload` ran hand-registry persistence, and for activation/deactivation the workspace and SQLite I/O behind it, synchronously on the async worker thread handling the request.
+  All five lifecycle operations now run their kernel call through `tokio::task::spawn_blocking` so the request handler never parks on disk I/O.
+  Successful responses and existing business-error status codes are unchanged; a join failure on the blocking task now returns a scrubbed 500 instead of propagating as an unhandled panic (#7039) (@houko)
--- a/crates/librefang-api/src/routes/skills/hands.rs
+++ b/crates/librefang-api/src/routes/skills/hands.rs
@@ -1,8 +1,7 @@
 use super::*;
 
-/// Hand lifecycle methods perform synchronous registry persistence and, for
-/// activation/deactivation, workspace and SQLite I/O. Keep that work off the
-/// async request worker.
+/// Hand lifecycle methods perform synchronous registry persistence and, for activation/deactivation, workspace and SQLite I/O.
+/// Keep that work off the async request worker.
 pub(super) async fn run_hand_lifecycle_job<F, T>(job: F) -> Result<T, tokio::task::JoinError>
 where
     F: FnOnce() -> T + Send + 'static,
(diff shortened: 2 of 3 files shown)


changelog.d/changed/7039-hands-lifecycle-io-off-thread.md
houko, line 1: Two of the three commits on this branch carry `Claude <noreply@anthropic.com>` as the git author identity:\n\n- `9f5a34e` — \"chore(changelog): add fragment for hand lifecycle I/O offload (#7039)\"\n- `805e603` — \"docs: join wrapped sentence in run_hand_lifecycle_job doc comment\"\n\nRepo `CLAUDE.md` (\"Commit & PR hygiene\" / \"Git Conventions\") is explicit that commit messages and PR content must carry no Claude/Anthropic attribution, and the `commit-msg` hook is supposed to reject a commit whose *author identity* resolves to Claude/Anthropic even when the message text itself is clean — which is exactly the case here (the message bodies are fine; the author field is not). It looks like these two commits were made/pushed through a path that bypassed that hook.\n\nI'm flagging rather than fixing directly: correcting this requires rewriting the author identity on already-pushed, non-tip-safe commits (`git commit --amend --reset-author` + rebase + force-push), and force-push is outside what this review pass is allowed to do without explicit maintainer sign-off. Suggest the PR author rebase with `git rebase -i --exec 'git commit --amend --reset-author --no-edit' 35b5fe9` (or an equivalent) against their own identity before merge.\n\nSeparately, the substantive change (offloading the five hand-lifecycle kernel calls to `spawn_blocking`) looks correct: `pause_hand` / `resume_hand` / `deactivate_hand` / `activate_hand_with_id` / `reload_hands` are plain synchronous functions with no internal `block_on`, and the only cross-request serialization they rely on (`hand_instance_lock`, a per-instance `std::sync::Mutex`) is acquired and released entirely within the synchronous call in both the old and new code paths — moving that call from the async worker to the blocking pool doesn't change which thread(s) can race for the lock, so the existing ordering/atomicity guarantees are preserved. The changelog fragment is present and correctly formatted, and a regression test (`lifecycle_job_does_not_block_the_async_worker`) backs the perf claim.

---

houko (commented):
Reviewed the diff against CLAUDE.md (build/verify workflow, changelog.d fragment format, commit conventions, determinism rules). One finding below — commit author identity, not the code itself. The spawn_blocking offload logic checks out as correct and safe.

---

PR #7038, opened 2026-08-12
perf: move session summary persistence off thread

houko:
## Summary

- move synchronous SQLite and workspace-file session summary writes onto the Tokio blocking pool
- keep summary generation and persistence ordering unchanged
- add a regression test covering both structured memory and the workspace mirror

## Tests

- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-kernel off_thread_persistence_writes_kv_and_workspace_mirror --lib`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo clippy -p librefang-kernel --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`

## Out of scope

- the no-runtime compatibility fallback remains synchronous because no Tokio blocking pool is available there

PR #7040, opened 2026-08-12
fix(channels): stop dispatch when journal write fails

houko:
## Summary

- make initial message-journal recording report whether the WAL entry became durable
- abort text and multimodal dispatch when the write-ahead record cannot be persisted
- send a retryable user-facing response instead of processing an unrecoverable message
- add a regression test proving failed writes do not enter the in-memory recovery index

## Tests

- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-channels record_reports_failure_and_does_not_index_unpersisted_entry --lib`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-channels message_journal::tests --lib`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo clippy -p librefang-channels --all-targets -- -D warnings`
- `git diff --check`

## Out of scope

- terminal outcome updates and periodic compaction retain their existing best-effort logging behavior

### DIFF
--- a/crates/librefang-channels/src/bridge.rs
+++ b/crates/librefang-channels/src/bridge.rs
@@ -4786,7 +4786,22 @@ async fn dispatch_message(
             metadata: std::collections::HashMap::new(),
             next_retry_after: None,
         };
-        j.record(entry).await;
+        if !j.record(entry).await {
+            error!(
+                id = %message.platform_message_id,
+                channel = ct_str,
+                "Message dispatch aborted because the recovery journal could not persist it"
+            );
+            send_response(
+                adapter,
+                &message.sender,
+                "Message could not be queued safely. Please try again.".to_string(),
+                thread_id,
+                output_format,
+            )
+            .await;
+            return;
+        }
     }
 
     // Send typing indicator (best-effort)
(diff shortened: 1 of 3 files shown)


houko (commented):
No `changelog.d/` fragment was added for this behavior change (dispatch now aborts when the journal write fails, instead of proceeding on an unrecoverable write). Sibling same-day PRs in this batch (#7036, #7038, #7039) each include one under `changelog.d/fixed/` or `changelog.d/changed/`. Worth adding `changelog.d/fixed/7040-message-journal-record-errors.md` per the format in `changelog.d/README.md`, or confirming it's intentionally omitted.

---

crates/librefang-channels/src/bridge.rs
```diff
@@ -4786,7 +4786,22 @@ async fn dispatch_message(
             metadata: std::collections::HashMap::new(),
             next_retry_after: None,
         };
-        j.record(entry).await;
+        if !j.record(entry).await {
```
houko, line 4789: The `record()` return-value contract is well covered at the `MessageJournal` level (`record_reports_failure_and_does_not_index_unpersisted_entry` in `message_journal.rs`), but neither `dispatch_message` nor `dispatch_with_blocks` has a test that exercises *this* abort path directly — i.e. that a failing `record()` actually short-circuits before the typing indicator / lifecycle reactions / router dispatch, and that the user-facing retry message goes out.

There's no existing `ChannelAdapter` mock in this test module (only `MockHandle` for `ChannelBridgeHandle`), so adding one is nontrivial — flagging rather than fixing directly. Worth a small mock adapter + a `MessageJournal::open()` pointed at a pre-created directory (same trick as the `message_journal.rs` test) to assert the router/agent is never invoked when the journal write fails, for both call sites.

---

houko (commented):
Automated daily review pass: found the PR missing its changelog.d fragment (fixed directly, see db1c74b) and one moderate-confidence test-coverage gap left as an inline comment. Everything else (commit messages, prose wrapping, lock/deadlock safety, error-handling idiom consistency) checked out.

---

houko:
Correction from this automated review pass: two commits it pushed here (`88efe0e`, `db1c74b`) are themselves authored as `Claude <noreply@anthropic.com>`, which is the exact no-AI-attribution violation CLAUDE.md's `commit-msg` hook targets. This happened because the review environment's global git identity was misconfigured (not something the review logic checked for on its own commits before pushing). A maintainer should reauthor/squash those commits before merge.

---

PR #7041, opened 2026-08-12
fix(api): surface session query failures

houko:
## Summary

- propagate session count and list failures instead of returning an empty successful page
- return scrubbed 500 responses for session list and FTS search database failures
- keep detailed database and blocking-task errors in server logs only
- add integration coverage using deliberately removed SQLite tables

## Tests

- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-api --test system_tools_sessions_test database_failure`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-api --test system_tools_sessions_test list_sessions_honours_pagination_query`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo test -p librefang-api --test system_tools_sessions_test search_sessions_returns_envelope_on_no_match`
- `CARGO_TARGET_DIR=/tmp/librefang-sidecar-delete/target cargo clippy -p librefang-api --all-targets -- -D warnings`
- `git diff --check`

## Out of scope

- session mutation endpoints remain unchanged


houko:
Correction from this automated review pass: the fragment-fix commit it pushed here (`7a247d7`) is itself authored as `Claude <noreply@anthropic.com>`, which is the exact no-AI-attribution violation CLAUDE.md's `commit-msg` hook targets. This happened because the review environment's global git identity was misconfigured (not something the review logic checked for on its own commits before pushing). No further automated action will be taken on this branch to avoid compounding the issue with a history rewrite; a maintainer should reauthor/squash that commit (and any others below) before merge.

---

PR #7055, opened 2026-08-13
fix(api): fail closed on config read errors

houko:
## Summary
- distinguish a missing sidecar config or secrets.env file from other read failures before read-modify-write operations
- preserve first-time creation for `NotFound`
- abort before creating a staging file when an existing target cannot be read, preventing replacement with a partial document
- add regression coverage for both config.toml and secrets.env read failures

## Testing
- `cargo test -p librefang-api --test sidecar_toml_test -- --nocapture`
- `cargo test -p librefang-api --lib routes::secrets_env::tests -- --nocapture`
- `cargo test -p librefang-api --lib routes::sidecar_toml::tests -- --nocapture`
- `git diff --check`

### DIFF
--- a/changelog.d/fixed/7055-config-read-before-write-errors.md
+++ b/changelog.d/fixed/7055-config-read-before-write-errors.md
@@ -1,0 +1,1 @@
+- Abort sidecar config and secrets.env read-modify-write operations when an existing file cannot be read instead of treating the failure as an empty file. (@xiaomo)
--- a/crates/librefang-api/src/routes/secrets_env.rs
+++ b/crates/librefang-api/src/routes/secrets_env.rs
@@ -56,7 +56,11 @@ pub fn upsert_secret(path: &Path, key: &str, value: &str) -> Result<(), String> 
         return Err(format!("invalid secret key `{key}`"));
     }
 
-    let original = fs::read_to_string(path).unwrap_or_default();
+    let original = match fs::read_to_string(path) {
+        Ok(contents) => contents,
+        Err(error) if error.kind() == std::io::ErrorKind::NotFound => String::new(),
+        Err(error) => return Err(format!("read {path:?}: {error}")),
+    };
     let mut out = String::with_capacity(original.len() + key.len() + value.len() + 2);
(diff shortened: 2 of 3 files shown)


houko:
The fail-closed change itself looks right — distinguishing `NotFound` (legitimately empty) from every other read error is exactly the split that keeps a first-time write working while refusing to silently truncate an unreadable file.

One thing to flag before this lands: **the rename-failure path loses its test coverage**, and not in a way that adding a test back would fix.

`rename_failure_cleans_up_staging_file` was renamed into `existing_secret_read_error_does_not_replace_target` rather than joined by it. Its original comment stated the intent explicitly:

> Make the destination an existing directory so the final `fs::rename(tmp, path)` fails (ENOTDIR/EISDIR) without touching the write/fsync stage at all — this exercises the rename-failure cleanup path specifically, distinct from the write-failure path.

That setup no longer reaches `rename`: with the read moved ahead of it, a directory at the target path now fails during `read_to_string`, which is what the rewritten assertions check. So after this PR the `if let Err(error) = fs::rename(...) { let _ = fs::remove_file(&tmp); ... }` branch in `upsert_secret` still exists but nothing exercises it — and it is the branch that stops a 0600 secret-bearing staging file from being left on disk.

Reconstructing it is genuinely awkward, which is why I'm raising it rather than just asking for the test back. Every simple way to make `rename` fail (target is a directory, target's parent is missing) now trips the earlier read, and the ways that don't (cross-filesystem staging, a read-only parent) either can't happen here — the staging file is a sibling by construction — or block staging-file creation too, which lands in the write-failure path instead.

Two options that seem workable:

- Extract the stage-write-and-replace sequence into a small helper taking the already-read contents, so the rename-failure branch can be driven directly without going through the read.
- Or leave the coverage gap but mark it: a short comment on the rename-failure branch noting it is currently unexercised, so a future refactor doesn't silently drop the `remove_file` cleanup and pass CI.

Either is fine by me; silently losing the only test that pinned "a failed rename must not leave a 0600 secret file behind" is the part worth avoiding.

PR #7068, opened 2026-08-13
fix(runtime): detect shell-expanded dangerous commands

houko:
## Summary
- normalize common `$IFS` whitespace expansion before dangerous-command matching
- block base64 decode pipelines that feed their decoded payload directly to `sh` or `bash`
- verify both `shell_exec` and `process_start` keep this safety gate in Full exec mode

## Security impact
Full exec mode intentionally permits shell syntax, but it still promises to block the runtime's known-dangerous command catalogue. Previously `rm${IFS}-rf${IFS}/` executed as `rm -rf /` without matching the catalogue, and an encoded payload could be decoded directly into a shell without matching the remote-content rule.

## Testing
- `cargo test -p librefang-runtime dangerous_command::tests --lib`
- `cargo test -p librefang-runtime process_start_blocks_ifs_hidden_command_in_full_mode --lib`
- `cargo test -p librefang-runtime test_shell_exec_full_mode_blocks_ifs_hidden_recursive_delete --lib`
- `git diff --check`

### DIFF
--- a/crates/librefang-runtime/src/dangerous_command.rs
+++ b/crates/librefang-runtime/src/dangerous_command.rs
@@ -270,10 +270,9 @@ impl DangerousCommandChecker {
             return CheckResult::Safe;
         }
 
-        // Normalise before matching. Besides case and NULs, model the shell's
-        // common `$IFS` whitespace expansion. Otherwise a Full-policy command
-        // can spell `rm -rf /` as `rm${IFS}-rf${IFS}/` and evade every regex
-        // even though the shell executes the same destructive argv.
+        // Normalise before matching.
+        // Besides case and NULs, model the shell's common `$IFS` whitespace expansion.
+        // Otherwise a Full-policy command can spell `rm -rf /` as `rm${IFS}-rf${IFS}/` and evade every regex even though the shell executes the same destructive argv.
         let normalised = normalize_for_detection(command);
 
         for pat in DANGEROUS_PATTERNS {
@@ -321,8 +320,7 @@ fn normalize_for_detection(command: &str) -> String {
     while index < bytes.len() {
         if bytes[index..].starts_with(b"${ifs") {
             let suffix = index + 5;
-            // Accept the closing brace directly and the standard shell
-            // parameter-expansion operators (`${IFS%?}`, `${IFS:-...}`, …).
+            // Accept the closing brace directly and the standard shell parameter-expansion operators (`${IFS%?}`, `${IFS:-...}`, …).
             // Do not rewrite a different variable such as `${IFS_FILE}`.
             if suffix < bytes.len()
                 && matches!(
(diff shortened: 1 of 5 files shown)


crates/librefang-runtime/src/dangerous_command.rs
```diff
@@ -305,6 +311,51 @@ impl DangerousCommandChecker {
     }
 }
 
+fn normalize_for_detection(command: &str) -> String {
```
houko, line 314: `normalize_for_detection` closes the `$IFS`/`${IFS...}` whitespace-avoidance bypass, but every entry in `DANGEROUS_PATTERNS` still relies on literal `\s+`/`\b` to separate `rm` from its flags and path (e.g. `r"\brm\s+(-[^\s]*\s+)*/"`). That means any argument-splitting mechanism that doesn't insert a literal whitespace/`$IFS` character still evades the whole catalogue — not just the base64 case this PR also closes.

Concretely, brace expansion needs no `$IFS` and no space in the source text:

```
$ bash -c 'rm() { printf "argv: "; for a in "$@"; do printf "[%s] " "$a"; done; echo; }; {rm,-rf,/}'
argv: [-rf] [/]
```

`{rm,-rf,/}` reaches the checker as the literal string `{rm,-rf,/}` — no `rm ` substring, no `${ifs}` — so it's `Safe` under every pattern, `normalize_for_detection` included, while the shell executes it as `rm -rf /`. Given the PR's stated goal ("Full exec mode ... still promises to block the runtime's known-dangerous command catalogue"), this seems like the same bug class re-appearing through a different shell feature, and there's no test here that would catch it. Worth a human call on whether this needs closing in this PR or is accepted residual risk of a regex-based denylist.

---

crates/librefang-runtime/src/dangerous_command.rs
```diff
@@ -155,6 +155,10 @@ pub static DANGEROUS_PATTERNS: &[DangerousPattern] = &[
         "pipe remote content to shell",
         r"\b(curl|wget)\b.*\|\s*(ba)?sh\b"
     ),
+    dp!(
+        "decode base64 content and pipe it to a shell",
```
houko, line 159: `\bbase64\b[^\n;]*\s(-d|--decode)\b[^\n;]*\|\s*(ba)?sh\b` requires the pipe to land on a bare `sh`/`bash` token. It misses the same payload piped to an explicit interpreter path or `env` wrapper, which are common in practice specifically because they dodge naive `|sh` detection:

```python
>>> import re
>>> pat = re.compile(r"\bbase64\b[^\n;]*\s(-d|--decode)\b[^\n;]*\|\s*(ba)?sh\b")
>>> bool(pat.search("base64 -d payload | /bin/sh"))
False
>>> bool(pat.search("base64 -d payload | /usr/bin/env bash"))
False
```

`regex_lite` doesn't support lookaround so a tight fix isn't trivial, but at minimum `sh` should probably be reachable through `(/[\w/]*/)?` or an `env\s+` prefix. This mirrors an existing gap in the `pipe remote content to shell` pattern above it (`curl|wget ... | (ba)?sh`), so it's not a regression this PR introduces, but the PR's own test suite (`decoded_payload_piped_to_shell_is_dangerous`) doesn't cover this form either, and the PR is specifically framed as closing shell-hiding gaps — flagging so a human can decide whether it's in scope here.

---

houko:
Superseded by #7166, which is now merged. The replacement closes the reviewed command-bypass gaps while preserving the original scope and adds regression coverage for shell expansion forms.

PR #7070, opened 2026-08-13
fix(kernel): serialize session compaction writes

houko:
## Summary
- serialize LLM session compaction with the same agent- or session-scoped mutex used by message writers
- acquire the lock before loading the session snapshot so concurrent turns cannot be overwritten by a stale compaction result
- carry the message loop's actual lock domain into pre-loop and background post-loop compaction
- use the task-local held-lock registry to avoid self-deadlocking automatic compaction already running inside a turn
- add regression coverage for default-session blocking, explicit-session blocking, and same-task re-entry

## Verification
- `cargo test -p librefang-kernel --lib compact_session_serializes_with_message_writers_without_self_deadlock -- --nocapture`
- `cargo clippy -p librefang-kernel --all-targets -- -D warnings`
- `git diff --check`

The first test build was interrupted by a host `No space left on device` error. After removing 11.5 GiB of generated Cargo output from a completed sibling worktree with `cargo clean`, the same test and all checks above passed.

## Out of scope
- generic optimistic conflict detection in `MemorySubstrate::save_session` remains separate; this PR closes the concrete compaction read/await/write race through the kernel's existing lock discipline.

### DIFF
--- a/crates/librefang-kernel/src/kernel/agent_runtime.rs
+++ b/crates/librefang-kernel/src/kernel/agent_runtime.rs
@@ -290,7 +290,7 @@ impl LibreFangKernel {
         agent_id: AgentId,
         force: bool,
     ) -> KernelResult<String> {
-        self.compact_agent_session_with_id(agent_id, None, force)
+        self.compact_agent_session_in_lock_scope(agent_id, None, force, true)
             .await
     }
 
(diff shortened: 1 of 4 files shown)


crates/librefang-kernel/src/kernel/agent_runtime.rs
```diff
@@ -309,13 +309,78 @@ impl LibreFangKernel {
         agent_id: AgentId,
         session_id_override: Option<SessionId>,
         force: bool,
+    ) -> KernelResult<String> {
+        self.compact_agent_session_in_lock_scope(agent_id, session_id_override, force, false)
```
houko, line 313: `compact_agent_session_with_id` hardcodes `agent_scoped = false` here, so it always takes `session_msg_locks[session_id_override]` regardless of what the concurrent writer for that exact session actually locks.

That's correct for the two in-`messaging.rs` call sites this PR fixes, because they thread through the *same* `agent_scoped` bool the enclosing turn itself computed. But `compact_agent_session_with_id` is also a public entry point with one external caller: `channel_bridge.rs`'s `compact_channel_session` (`crates/librefang-api/src/channel_bridge.rs:1852-1863`), which passes `SessionId::for_sender_scope(agent_id, channel, chat_id)` as the override.

Per `send_message_full_inner` (messaging.rs:1052-1055, mirrored at 2849-2851), an ordinary write to that same session only takes `session_msg_locks[sid]` when the *caller of the message turn* explicitly passed a `session_id_override` that differs from the agent's canonical session. Plain channel message dispatch (`send_message_with_sender_context`, used by the channel bridge for normal chat) always passes `session_id_override = None`, so it locks `agent_msg_locks[agent_id]` no matter which underlying session gets written — including the channel-derived one.

So a `/compact`-style command through `compact_channel_session` acquires `session_msg_locks[sid]`, while the concurrent ordinary writer for that same channel session is holding `agent_msg_locks[agent_id]` — two different mutexes guarding one session, i.e. the exact race this PR's commit message describes, still open for this call path.

The existing test `compact_with_id_targets_the_requested_session` in `crates/librefang-kernel/tests/session_reset_scope_test.rs:497-498` even calls `compact_agent_session_with_id(agent_id, Some(dashboard_sid), false)` with `dashboard_sid` set to `kernel.agent_registry().get(agent_id).unwrap().session_id` — i.e. the agent's *canonical* session, passed through the session-lock branch. Any concurrent canonical-session turn (which always takes the agent lock) races against that compaction call the same way.

Given the risk (deadlock/race correctness), flagging rather than changing the lock-selection logic myself — this may need `compact_agent_session_with_id` (or its callers) to collapse to the agent lock when the target session is the one ordinary writers for it actually use, mirroring the `canonical_session_id` collapse already done in `send_message_full_inner`.

---

houko:
Superseded by #7167, which is now merged. The replacement serializes session compaction writes and includes the targeted concurrency regression test requested by review.

PR #7076, opened 2026-08-13
fix(api): harden user mutation boundaries

houko:
## Changes

- Scrub user persistence, hashing, and provider credential-store failures from HTTP 500 bodies while retaining full server-side logs.
- Publish the durably written authentication snapshot before awaiting kernel reload, so reload failure or request cancellation cannot leave a rotated key valid.
- Restrict the auth snapshot write lock to its atomic replacement instead of holding it across disk persistence and reload.
- Classify repeated bulk-import names against live batch state in both preview and commit modes.
- Reject control characters in user names, document the intentional per-user table formatting replacement, and correct the delete OpenAPI response to 204.
- Retain the existing Tokio filesystem and `spawn_blocking` persistence boundary; the report's blocking-I/O finding no longer matches current source.

## Verification

- `cargo test -p librefang-api --test users_test` — 31 passed
- `python3 scripts/check-changelog-attribution.py --staged`
- `git diff --check`

## Out-of-scope follow-ups

- None.


houko:
Rebased onto `origin/main` (4d7c57e51); the three source commits replayed without conflict because main has not touched `crates/librefang-api/src/routes/users.rs` or `tests/users_test.rs` since the old merge base.

The only conflict was `xtask/baselines/openapi.sha256`, where both sides had regenerated the hash from their own `openapi.json`, so neither recorded value described the rebased document.
Rather than picking a side, the baseline is recomputed from the rebased `openapi.json` (`7b6bd1deb6f8a19799cb441a247d5b026316c686a3bf65a580712367e6b3e877`), which keeps the drift gate self-consistent without waiting for a codegen round-trip.
`openapi.json` itself auto-merged; this branch's only remaining delta against main is the `DELETE /api/users/{id}` success response at 204.

PR #7080, opened 2026-08-13
fix(api): harden passkey authentication state

houko:
## Summary
- centralize passkey HTTP 500 handling with stable client-facing messages
- retain persistence, serialization, and corrupt-credential details in structured logs
- preserve client-facing WebAuthn 400 validation errors
- persist credential state with compare-and-swap updates
- reload and reapply authentication results after concurrent updates
- return a documented 409 after bounded contention instead of overwriting a newer sign count
- add regression coverage for error scrubbing and stale credential writes

## Validation
- `cargo test -p librefang-memory passkey_store`
- `cargo test -p librefang-api routes::passkey --lib`
- `cargo fmt --all`
- `git diff --check`

## Out of scope
- passkey ceremony-state lifecycle changes remain in #7135


houko:
Rebased onto `origin/main` (9c6653df5) and force-pushed as f5103902a.

The only conflict was `xtask/baselines/openapi.sha256`: both sides had regenerated it against their own `openapi.json`, so neither recorded hash described the merged document.
`openapi.json` itself auto-merged with no conflict, and its delta against main is still exactly the two documented failure responses on passkey authentication (409 credential-changed, 500 persistence-failed), so the baseline was recomputed from the merged file rather than taking either side — `xtask schema-check check` compares the sha256 of `openapi.json`'s bytes, and it now matches.

No source conflicts: main has not touched `crates/librefang-api/src/routes/passkey.rs` or `crates/librefang-memory/src/passkey_store.rs` since the previous merge base, and the branch's diff against main over `crates/` and `changelog.d/` is byte-identical to what it was before the rebase.

houko:
Correction to the SHA above: `main` advanced twice while the first rebase was in flight, so the branch was rebased again onto 4d7c57e51 and is now at 9f580096c. Same resolution, same recomputed baseline (`de1b24063a038cfd766dab139a1ab017e20c2a0f1bface853f0fc809f17ee3b3  openapi.json`). GitHub now reports the PR as mergeable.

PR #7157, opened 2026-08-13
fix(api): align binding delete contract

houko:
## Summary

- make the binding-delete path extractor use the same `u32` index domain advertised by OpenAPI
- document the actual successful HTTP 204 response
- return a bodyless 204 instead of serializing JSON `null`
- add a regression test for the empty response body

## Verification

- `CARGO_INCREMENTAL=0 CARGO_TARGET_DIR=/tmp/librefang-target-binding-delete cargo test -p librefang-api routes::bindings::tests::no_content_response_has_no_json_body`
- `CARGO_INCREMENTAL=0 CARGO_TARGET_DIR=/tmp/librefang-target-binding-delete cargo clippy -p librefang-api --all-targets -- -D warnings`
- `CARGO_INCREMENTAL=0 CARGO_TARGET_DIR=/tmp/librefang-target-binding-delete cargo check --workspace --lib`
- `git diff --check`

## Out of scope

- Other endpoints that return JSON with 204 responses remain separate audit findings.


houko:
Rebased onto `main` (`3e5c53b`), replacing the two earlier merge commits with a linear history.

Two resolutions worth flagging:

- `xtask/baselines/openapi.sha256` conflicted because both sides had recorded the hash of their own `openapi.json`. `openapi.json` itself auto-merged, and its only delta against `main` is still the binding delete success response moving from 200 to 204, so the baseline is recomputed from the merged document (`11010ed…`) rather than taking either recorded side — the same thing `scripts/hooks/pre-commit` does when `openapi.json` is staged.
- The changelog fragment is renamed to `changelog.d/fixed/7157-binding-delete-contract.md` and now ends `(#7157) (@houko)`, matching every other fragment in that directory. Release assembly only suppresses the auto-generated `- <PR title> (#N)` line when a curated bullet carries the PR number, so without it this change would have been described twice in the release notes.

`crates/librefang-api/dashboard/openapi/generated.ts` still declares 200 for `remove_binding`; that artifact is left for the `openapi-drift` job to regenerate.

houko:
Self-review before merge — the contract change is right; its test was not. Fixed in 6a8e805e (on top of a `main` merge).

**The change itself is correct.** `(StatusCode::NO_CONTENT, Json(json!(null)))` really does emit a `null` body with a JSON `content-type` on a 204, which a generated client decoding a no-content response cannot handle, and the OpenAPI document really did advertise 200 while the handler returned 204. Taking `Path<u32>` to match the document's declared parameter type is right too.

**The test did not test the change.** `no_content_response_has_no_json_body` called `binding_removed_response()` — a helper introduced only so there was something to call — and asserted that `StatusCode::NO_CONTENT.into_response()` has an empty body. That is a property of axum, not of this route, and it would have passed unchanged against the buggy handler. CLAUDE.md requires a `TestServer`-style integration test for any route contract change, and this is exactly why.

Replaced with `crates/librefang-api/tests/bindings_routes_test.rs`, which mounts `routes::bindings::router()` and drives real requests:

- `DELETE /api/bindings/0` → 204, empty body, **and no `content-type`** (the header a client would try to decode).
- Read-back through `GET /api/bindings` proves the binding is actually gone, not just that the response looked right.
- Out-of-range index → 404 carrying a translated `error` string.
- `4294967296` → 400, pinning the `u32` parameter type the document promises.

Confirmed the test bites: restoring the `Json(json!(null))` body fails with `204 must carry no body, got "null"`.

With the contract covered end to end, `binding_removed_response()` had no remaining purpose, so it is inlined with a comment pointing at the test.

`main` merged in; `xtask/baselines/openapi.sha256` recomputed from the merged `openapi.json` rather than taking either side's hash — neither described the merged file.

Verification: `cargo test -p librefang-api --test bindings_routes_test` 3 passed, `cargo clippy -p librefang-api --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

PR #7159, opened 2026-08-13
fix(channels): /think executes for real; commands resolve the conversation agent

### LINKED ISSUE #7140: bug: Telegram slash commands appear inert — SDK drift + command-path routing divergence
## Symptom

On Telegram, `/new`, `/agents`, and related commands return a confirmation phrase but the visible conversation does not change.

## What was verified (correction of the initial report)

The command dispatcher **is fully implemented** in `crates/librefang-channels/src/bridge.rs:6828` (`handle_command`), with real kernel calls: `"new"` → `reset_channel_session` → `kernel.reset_session`, `"agents"` → `list_agents` (inline keyboard), `"agent"` → find/spawn + store user default, plus `start/help/status/model/usage/...`. The sidecar parses `bot_command` → `Content.command` and the daemon deserializes it into `ChannelContent::Command` (sidecar.rs:1149). The chain is real end-to-end.

### 1. Sidecar SDK drift (deployment issue, most likely trigger)

The Telegram sidecar runs a pip-installed `librefang` SDK. A live deployment showed SDK **2026.3.2201** (March) against a daemon of **2026.7.31** (August). The sidecar protocol has evolved in between (capabilities negotiation, sender-identity fields, content handling). A 4-month-old adapter paired with a current daemon can produce exactly this symptom class: messages that arrive but are misparsed or silently degraded.

Recommendation: the daemon should log a WARN when a sidecar reports a protocol/sdk version older than the daemon expects (today it does not), and `describe_sidecar` should surface the adapter version.

### 2. Command-path routing divergence (code-level, real)

`/new` resolves the target agent via `resolve_for_command` (bridge.rs:6845) which consults ONLY the router chain (binding → direct route → per-account user default → channel default → system default). The regular message path `resolve_or_fallback` (bridge.rs:3511) additionally consults: addressed agent, #5671 conversation override, #5323 sticky holder, per-peer binding, and the `channel_bindings` SQLite instance default.

When those differ (sticky holder points at agent B, router resolves to agent A), `/new` resets **agent A** and replies "Session reset for this telegram chat" — a real reset on the wrong agent, invisible to the user.

### 3. `/agent <name>` selection does not stick

`/agent` writes only the router store (bridge.rs:6913). `set_conversation_binding` (channel_binding_store.rs:123) has **no production caller** — the #5671 conversation-binding store is write-dead. A selection made via `/agent` can therefore be overridden by older, higher-precedence bindings on the next message.

### 4. Stubs with confirmation phrasing (by design, but misleading)

- `/think` (channel_bridge.rs:1974): explicit stub — stores a preference, returns "This will take effect when supported by the model."
- `/goal` (channel_bridge.rs:1339): `start_goal_run` is fire-and-forget; if the kernel self-handle is unset the run silently never starts (goal_lifecycle.rs:43) while the reply says "Goal created and started".

## Suggested fixes

1. **Version the sidecar protocol**: daemon logs a WARN on old sidecar SDK versions and `describe_sidecar` shows the version. Ship/install the matching SDK next to the daemon binary.
2. **Unify command resolution**: `resolve_for_command` should use the same chain as `resolve_or_fallback` (sticky holders + conversation overrides first, router fallback second).
3. **Wire `/agent` to `set_conversation_binding`** so the selection binds the active conversation, or document which store is authoritative.
4. **Make `/think` and `/goal` honest**: report "not yet supported" for /think until implemented; await/verify `start_goal_run` before confirming.

DaBlitzStein:
Fixes the two root causes behind Telegram slash commands confirming without acting (issue #7140).

## 1. `/think` was a stub

`set_thinking` stored nothing and never affected the model. Now:
- Per-agent extended-thinking preference stored on `KernelBridgeAdapter` (keyed by agent id)
- Applied as the per-turn thinking override on the next chat message via new `send_message_with_blocks_and_sender_thinking` (KernelApi method + kernel impl; default ignores the override)
- Reply is honest: "Extended thinking enabled for this chat."

## 2. Command routing divergence

`/new` `/reset` `/reboot` `/compact` resolved the target agent via the router chain only, while the regular chat path (`resolve_or_fallback`) also consults:
- Explicit per-conversation `/agent` override (#5671)
- Sticky conversation holder (#5323)
- Per-peer `[[bindings]]` match
- Instance default from `[[sidecar_channels]] agent`

A divergent resolution reset a *different* agent than the one holding the conversation — a real reset, invisible to the user. `resolve_for_command` now mirrors the chat chain layer for layer, in the same precedence order.

## Not in this PR

- `/goal` honesty fix (fire-and-forget start) — depends on the extended goal-runner signature from #6505, tracked separately.
- SDK version drift (deployment concern, not code).

## Verification

- `cargo check --workspace --lib` clean
- `cargo test -p librefang-channels --lib`: 551 passed, 0 failed

Closes #7140

## Follow-up commits (same issue #7140)

18ef5fd65: `/agent` now also persists the per-conversation binding (the ack claimed the new agent while dispatch kept the old one — `set_conversation_binding` had zero production callers), and the streaming send path now applies per-agent `/think` preferences (previously only the non-streaming fallback honored them). Regression tests assert the binding side effect and the streaming-path override.


houko (commented):
Blocking compile failure: `crates/librefang-api/src/channel_bridge.rs` constructs `KernelBridgeAdapter` without the required `thinking_prefs` field in two test fixtures (currently around lines 3348 and 3395). This causes Quality, Unit, Ubuntu, and aarch64 to fail before the behavior can be evaluated. Please initialize the field consistently with the surrounding adapter fixtures, then rebase/update onto current `main` so the already-fixed dependency security failure is no longer inherited.

houko (commented):
Two blockers remain on the current tip.

1. thinking_prefs is keyed only by agent_id, while the user-facing result says the preference applies to this chat. One agent can serve multiple conversations. A /think command in one conversation therefore changes thinking and cost behavior for every other conversation routed to the same agent. Please key this preference by the canonical conversation identity and pass that identity through set_thinking and the send path. Add a regression with two conversations sharing one agent.

2. resolve_for_command does not actually reproduce build_thread_key. The normal path adds peer_id from sender_user_id(message). The command path builds channel, account_id, chat_id, and thread only. Existing sticky claims are therefore stored under a different ThreadKey in group conversations and current_holder will miss them. Please share the canonical key builder or pass the already-built key into command dispatch. Add a regression where a group-thread claim includes peer_id and /new or /reset resolves the holder.

These are correctness issues in the two behaviors this PR claims to fix, so I do not recommend merging the current tip.

houko:
`Test / Unit (lib+bin)` and `Test / Ubuntu (shard 1/4)` both fail on the same case, `channel_bridge::tests::streaming_send_applies_thinking_pref_set_by_think_command`, at `crates/librefang-api/src/channel_bridge.rs:3591`:

```
expected provider-boundary failure, got: Ok(())
```

The assertion is `status.is_err()`, resting on the premise in the comment right above it — "Provider-less test kernel: the agent loop fails at the driver boundary (no LLM credentials)". A hosted runner has no LLM credentials, and the status channel still resolved `Ok(())`, so the provider-less path is not surfacing an error through that oneshot the way the test assumes.

The deeper issue is that the assertion doesn't test what the test is named for. Its own comment concedes this: "What matters here is that the stream started and ended with a reported failure — the pref read and the thinking-aware kernel entry were both reached." But `status.is_err()` is satisfied by *any* failure anywhere in the loop and would equally be satisfied if the thinking preference were never consulted at all. It is an incidental side effect of an unconfigured environment standing in for the behaviour under test, which is why an unrelated change to the provider-less path breaks it.

Asserting the actual contract would both fix the failure and make the test meaningful: that the preference set by `/think` reaches the kernel call. Observing the value the bridge passes down — the `thinking_prefs` entry, or the flag on the kernel invocation — is independent of whether a provider is configured, so it stays green on a credential-less runner and still fails if the wiring regresses.

Two notes while you're in here:

`KernelBridgeAdapter` is constructed inline in the test with all three fields, so every future field addition edits this test too. If the struct grows a `Default` or a test constructor, this site is a good candidate.

Separately: #7701 currently fails to compile against `main` with `error[E0560]: struct 'channel_bridge::KernelBridgeAdapter' has no field named 'thinking_prefs'` — it is written against the field this PR introduces. Whichever lands first, the other needs rebasing; worth sequencing them deliberately rather than letting merge order decide.

Not pushing anything to your branch.

DaBlitzStein:
Status note, for visibility rather than action: the per-conversation scope for `/think` is stacked on this branch and cannot go upstream independently, because upstream `set_thinking` takes `(agent_id, on)` only — there is no `channel` / `chat_id` in the trait signature to key a per-conversation preference on.

Waiting on this PR:

- `/think` scoped to the `(agent, channel, chat_id)` triple the reset / reboot / compact commands already use, so the toggle affects the conversation it was typed in rather than every conversation the agent has.

For the record on the current behaviour: both upstream implementations are inert. `ChannelBridge::set_thinking` in `crates/librefang-api/src/channel_bridge.rs:1921` returns `"Extended thinking {state}. (This will take effect when supported by the model.)"` without storing or applying anything, and the trait default in `crates/librefang-channels/src/bridge.rs:251` returns `"Extended thinking preference saved."` with an empty body. So `/think` acknowledges and does nothing at all today.

DaBlitzStein:
`streaming_send_applies_thinking_pref_set_by_think_command` passes as of ae83aab0b.

**What was actually wrong.**
The `is_err()` assertion you flagged was pinning the mock harness, not the adapter: whether the provider-less test kernel fails the turn at the driver boundary or completes it is a property of whichever driver `MockKernelBuilder` seeds, so the assertion went green-to-red the moment the harness started resolving the loop.
742a983ab already removed it, but the replacement assertion was still not testing the right thing — it re-derived `agent_id.0.to_string()` inside the test and looked the pref up with that literal, which pins a key the test itself controls rather than the lookup `send_message_streaming_with_sender_status` actually performs.

That distinction matters here, because the bug this PR fixes *is* a key-drift bug.
`set_thinking` wrote the preference into `thinking_prefs` and the two send paths read it back, but all three sites derived the map key independently — so a read site that ever derived a different key would make `/think` silently inert, and neither the old status assertion nor a test-side key literal could observe it.

**What changed** (`crates/librefang-api/src/channel_bridge.rs`):
- The key is derived once in `thinking_pref_key` (:717); the write goes through `store_thinking_pref` (:722, called from `set_thinking` :1972) and both reads through `thinking_override_for` (:733, called from the streaming send :879 and the blocks send :940).
  Write and read can no longer drift apart by construction.
- The test (:3553) now asserts through `thinking_override_for` (:3611) — the same accessor the send path calls — plus a negative case pinning that an agent which never used `/think` gets `None` rather than inheriting another agent's toggle.

No behaviour change: the key expression is byte-identical, only its definition site moved.

**Verification** (`CARGO_TARGET_DIR` shared, all runs scoped):
- `cargo test -p librefang-api` — 2230 passed across 90 test binaries.
  Two failures, both pre-existing on `main`, in files this PR does not touch, and both pass when re-run in isolation: `routes::mcp_auth::tests::flow_vault_cleanup_removes_all_per_flow_keys_on_drop` and `wait_step_over_cap_halts_before_sleep`.
  The first is a real env-var race rather than load flake — `routes/mcp_auth.rs:1850` sets `LIBREFANG_VAULT_KEY` process-wide on the claim that "no other api test reads" it, which `server.rs:3961` (`pin_vault_key`) disproves by setting a different key; whichever loses the race decrypts with the wrong master key.
  That is worth its own issue, separate from this PR.
- `cargo clippy --workspace --all-targets -- -D warnings` — clean on this branch's crates once the Rust 1.98 breakages already on `main` are excluded (`chunks_exact_to_as_chunks`, `drain_collect`, `map_or_identity`, `result_large_err`), which are exactly what #7755 and #7762 are fixing.

**On ordering.** Fixing #7159 first and rebasing #7701 on top is the right sequence, since #7701 builds on the `thinking_prefs` field introduced here.
One thing to flag before that: this branch is 153 commits behind `origin/main` and currently conflicts in `crates/librefang-channels/src/bridge.rs`.
Since the branch has already been reviewed I have not rewritten its history — how you would like that resolved (merge `main` in, or rebase) is your call.

houko (changes requested):
The `/think` preference is described to the user as scoped to “this chat”, but it is stored and read only by `agent_id` (`crates/librefang-api/src/channel_bridge.rs:624-627, 717-739`). A single agent commonly serves multiple Telegram chats and can also serve other channel instances, so `/think on` in one group or DM changes the reasoning mode for every subsequent turn routed to that agent, including turns from other users and channels. This also means `/think off` from any one chat disables it for all of them. Please scope the preference by the actual channel session/conversation identity (at least channel/account/chat plus agent), pass that scope through `set_thinking`, and add a regression test with two conversations routed to the same agent proving that toggling one does not affect the other.

DaBlitzStein:
Superseded — all three halves of this branch are on `origin/main`, and the review's blocking point was resolved exactly as written.

**The scoping fix the review asked for.** `crates/librefang-api/src/channel_bridge.rs:629` now holds `thinking_prefs: DashMap<(AgentId, ConversationScope), bool>`, keyed through `ConversationScope::from_sender` (line 648). The doc comment on the field makes the same argument the review did: *"Keyed by the conversation the command was typed in, never by the agent alone: one agent routinely serves many chats, so an agent-keyed preference would push one user's reasoning mode — and its token cost — onto everybody else's turns."*

**The regression test with two conversations routed to the same agent**, also asked for by name: `think_preference_is_isolated_per_conversation` at line 3423.

**`/think` applying for real.** `thinking_override_for` is consumed on all four send paths (lines 848, 890, 925, 962), streaming included — the gap this branch opened against.

**`/agent` persisting the conversation binding.** `conversation_binding` / `set_conversation_binding` are in place, with coverage at line 3568.

Closing rather than rebasing: there is nothing here upstream does not already do, and the version upstream landed is the scoped one the review called for rather than the agent-keyed one this branch shipped.

PR #7164, opened 2026-08-13
fix(api): harden pairing route boundaries

houko:
## Summary
- accept only normalized HTTP or HTTPS forwarded schemes in pairing QR URLs
- validate stored device metadata length and control characters
- use the canonical device bearer generator
- classify expired tokens, capacity conflicts, and internal completion failures through the shared error envelope
- acquire the live-auth write guard before the synchronous pairing transition so cancellation cannot split persisted and live bearer state
- report actual notification successes and failures instead of paired-device count
- document the completion and notification response statuses

## Validation
- `cargo test -p librefang-api pairing_tests --lib`
- `cargo test -p librefang-api --test pairing_test`
- `cargo test -p librefang-api --test pairing_backup_extra_test`
- `cargo fmt --all`
- `git diff --check`

## Out of scope
- push-provider retry and fan-out architecture remain unchanged
- device platform values remain open strings for forward compatibility


houko:
Rebased onto `origin/main` (`4d7c57e5`); the branch is now linear — the earlier `Merge branch 'main'` commit is gone and the three original commits replay unchanged. `git diff origin/main..HEAD` for `routes/pairing.rs` and the changelog fragment is byte-identical to the pre-rebase delta, so no hunk was dropped.

One resolution worth flagging: `xtask/baselines/openapi.sha256` was the only conflict. `openapi.json` auto-merged cleanly and its delta against main is still exactly the six documented pairing responses, but both sides of the baseline recorded a hash of their own spec, so neither describes the merged document. The line is recomputed from the merged `openapi.json` rather than taking a side, matching what `schema_check::run_gen` writes (`<sha256>  openapi.json\n`); `shasum -a 256 -c` passes against the tree.

Also noticed while re-reading the fragment: it is attributed `(@xiaomo)`, the only fragment in `changelog.d/` not using `(@houko)`. Left as-is since attribution is yours to decide, but it renders verbatim into the release notes.

PR #7165, opened 2026-08-13
fix(api): attribute authz denial endpoint

houko:
## Substantive changes
- pass the concrete authorization endpoint into the shared Admin gate
- record `/api/authz/check` denials as `authz/check` instead of mislabeling them as `authz/effective`
- add integration coverage that inspects the emitted audit detail

## Verification
- `CARGO_INCREMENTAL=0 CARGO_TARGET_DIR=/tmp/librefang-target-authz-audit cargo test -p librefang-api --test authz_routes_integration authz_check_viewer_denial_audit_names_check_endpoint` (1 passed)
- `CARGO_INCREMENTAL=0 CARGO_TARGET_DIR=/tmp/librefang-target-authz-audit cargo test -p librefang-api --test authz_routes_integration` (13 passed)
- `CARGO_INCREMENTAL=0 CARGO_TARGET_DIR=/tmp/librefang-target-authz-audit cargo clippy -p librefang-api --all-targets -- -D warnings`
- `CARGO_INCREMENTAL=0 CARGO_TARGET_DIR=/tmp/librefang-target-authz-audit cargo check --workspace --lib`
- `cargo fmt --all`
- `git diff --check`

## Out of scope
- authorization policy and endpoint response behavior are unchanged.

### DIFF
--- a/changelog.d/fixed/authz-denial-endpoint.md
+++ b/changelog.d/fixed/authz-denial-endpoint.md
@@ -1,0 +1,1 @@
+Attribute denied authorization checks to the correct API endpoint in audit records. (@xiaomo)
--- a/crates/librefang-api/src/routes/authz.rs
+++ b/crates/librefang-api/src/routes/authz.rs
@@ -52,14 +52,18 @@ pub fn router() -> axum::Router<Arc<AppState>> {
 /// don't blanket-trust an unauthenticated origin even on loopback. To
 /// use this endpoint in a no-auth deployment, configure at least one
 /// user with an admin api_key.
-fn require_admin(state: &AppState, api_user: Option<&AuthenticatedApiUser>) -> Option<Response> {
+fn require_admin(
+    state: &AppState,
+    api_user: Option<&AuthenticatedApiUser>,
+    endpoint: &str,
+) -> Option<Response> {
     match api_user {
         Some(u) if u.role >= UserRole::Admin => None,
         Some(u) => {
             state.kernel.audit().record_with_context(
                 "system",
                 librefang_kernel::audit::AuditAction::PermissionDenied,
-                format!("authz/effective endpoint denied for role {}", u.role),
+                format!("{endpoint} endpoint denied for role {}", u.role),
                 "denied",
                 Some(u.user_id),
                 Some("api".to_string()),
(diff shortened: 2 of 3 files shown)


houko:
Superseded by main — recommending closure rather than resolving the conflict.

The conflict here is in `crates/librefang-api/src/routes/authz.rs`, and resolving it shows why: main has landed the same change independently. `require_admin` already takes the endpoint name and both call sites already pass it (`authz/effective` at `crates/librefang-api/src/routes/authz.rs:121`, `authz/check` at `:234`). The only difference between the two sides of the conflict was `&str` here versus `&'static str` on main — main's is the tighter bound, and every call site passes a literal.

After merging main in locally, this branch's net diff against `origin/main` is a changelog fragment plus 23 lines of integration test. Nothing in `src/` remains.

The remaining test is also already covered, and covered more strictly. This branch adds `authz_check_viewer_denial_audit_names_check_endpoint`, asserting `detail.contains("authz/check endpoint denied")`. Main's `authz_check_viewer_role_is_forbidden` exercises the identical request (viewer role, `GET /api/authz/check?user=Alice&action=web_search`) and asserts the full string with `assert_eq!`:

```rust
assert_eq!(
    denial.detail, "authz/check endpoint denied for role viewer",
    "audit attribution must identify the endpoint that was probed"
);
```

A substring check cannot fail where that equality passes, so landing this branch would add a strictly weaker duplicate.

Not pushing the merge: doing so would queue a full workflow set for a branch that contributes no source change, and the hosted runner pool is the scarce resource. If you'd rather keep the fragment for release-note wording, say so and I'll reduce the branch to just that.

houko:
Rebased onto `origin/main` (`ca9e173`). One conflict, in `crates/librefang-api/src/routes/authz.rs`, resolved by taking main's side.

Worth flagging before this gets reviewed: #7409 landed the same fix on main and this PR is now almost entirely subsumed by it.

- `routes/authz.rs` is byte-identical to main after the rebase — main threads the same `endpoint` argument into `require_admin` (typed `&'static str`) and emits the same `{endpoint} endpoint denied …` audit detail, plus wording fixes to the two `ApiErrorResponse::forbidden` messages this PR left alone.
- `authz_check_viewer_denial_audit_names_check_endpoint` duplicates main's `authz_check_viewer_role_is_forbidden`: identical setup and request, but a weaker assertion (`recent(1)` + `detail.contains("authz/check endpoint denied")` instead of exact equality on the full detail string).
- `changelog.d/fixed/authz-denial-endpoint.md` describes the same user-visible change as `changelog.d/fixed/7409-authz-denial-audit-label.md`, already on main, so merging as-is would put two bullets for one fix in the release notes.

I did not drop either remaining hunk unilaterally, so the diff still reflects the original intent. Recommend closing as superseded by #7409, unless you want the stricter `recent(1)` ordering assertion kept — in which case it belongs folded into the existing test rather than added alongside it.

PR #7174, opened 2026-08-14
fix(api): bound backup restore extraction

houko:
## Substantive changes
- cap backup restore entry count, manifest size, per-entry size, total extracted size, and compression ratio
- preflight all archive metadata before writing restored files
- stream each entry to disk through a hard byte limit instead of buffering it entirely in memory
- reject symbolic-link parents and leaf targets that could redirect extraction outside the restore root
- use `O_NOFOLLOW` on Unix restore file opens to close the leaf-symlink check/open race
- remove partial output on streaming errors or runtime limit violations
- report partial filesystem restores as HTTP 500 and audit them as failed instead of claiming a completed 200
- audit successful backup deletion with authenticated user attribution and document its actual 204 response
- move restore archive discovery off the async executor
- keep detailed extraction and limit failures in server logs behind fixed client messages

## Verification
- `cargo test -p librefang-api routes::backup::tests --lib` (5 passed)
- `cargo test -p librefang-api --test pairing_backup_extra_test` (19 passed)
- `git diff --check`

## Out of scope
- transactional staging and rollback of a multi-file restore require a broader storage-layout design


houko:
Rebased onto `main` (4d7c57e51) and force-pushed as `e021f55ec`; the two prior `Merge branch 'main'` commits are gone and the branch is now five linear commits on top of `main`.

Only `xtask/baselines/openapi.sha256` conflicted. Neither side's recorded hash describes the rebased document — main regenerated its baseline for #7806/#7807's `openapi.json` changes, and this branch's baseline was computed from the older merged file — so the line is recomputed from the actual `openapi.json` (`shasum -a 256`, the same computation `scripts/hooks/pre-commit` and `xtask schema-check` use) instead of picking a side.

`openapi.json` itself auto-merged; this branch's only delta against main remains the backup delete success response moving from 200 to 204. No Rust source conflicted — main has not touched `crates/librefang-api/src/routes/backup.rs` or `crates/librefang-api/tests/pairing_backup_extra_test.rs` since the previous merge base, and the full source diff is byte-identical to the pre-rebase one.

PR #7224, opened 2026-08-14
fix(whatsapp): scope and expire echo tracking

houko:
## Changes
- scope echo-tracker keys by destination chat so identical messages in unrelated conversations do not collide
- expire echo entries after five minutes while preserving the existing LRU capacity bound
- pass stable JID scopes at every inbound and outbound production call site
- reject non-finite and non-positive deduplication windows at construction
- document the tracker’s single-process deployment assumption

## Verification
- `cd packages/whatsapp-gateway && node --test test/dedup-tracker.test.js test/echo-tracker.test.js` (27 passed)
- `cd packages/whatsapp-gateway && node --check index.js`
- `cd packages/whatsapp-gateway && node --check lib/echo-tracker.js`
- structural search confirms all eight production `echoTracker.track` / `isEcho` calls pass a chat scope
- `git diff --check`

## Out of scope
- distributed echo tracking across multiple gateway processes
- changing the deliberate two-phase message deduplication flow used for decrypt retries

### DIFF
--- a/changelog.d/fixed/whatsapp-tracker-correctness.md
+++ b/changelog.d/fixed/whatsapp-tracker-correctness.md
@@ -1,0 +1,1 @@
+Scope WhatsApp echo detection by chat, expire stale echo entries, and reject invalid message-deduplication windows. (@xiaomo)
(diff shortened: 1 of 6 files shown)


houko:
Closing because this PR duplicates and conflicts with the already-open #7183, which owns the same WhatsApp echo-tracker scope fix and production call sites. The additional TTL and dedup-window changes should be reconsidered as a rebased follow-up after #7183 merges or closes, rather than asking maintainers to review competing implementations.

PR #7271, opened 2026-08-14
test(codegen): keep assertions active under optimization

houko:
## Changes

- replace bare assertions and process exits with explicit `AssertionError` checks
- defer loading the SDK generator until the standalone test runs
- require exactly one `invoke_tool` operation across tags
- make generated function-signature checks whitespace tolerant while retaining exact safety fingerprints
- route changes to the SDK generator and its contract test through the required CI Quality job
- run the contract suite in both normal and optimized Python modes in CI

## Verification

- `python3 scripts/tests/test_codegen_sdks.py` (383 operations across 25 tags)
- `python3 -O scripts/tests/test_codegen_sdks.py` (383 operations across 25 tags)
- imported the test module and confirmed `codegen_sdks` was not loaded
- `python3 -m py_compile scripts/tests/test_codegen_sdks.py`
- parsed `.github/workflows/ci.yml` with Ruby YAML
- `git diff --check`
- repository commit and push hooks

## Out of scope

- generator output changes
- generated SDK artifact changes
- replacing precise security-contract fingerprints with a full language parser


houko:
Self-review before merge — one blocker found and fixed in 20a73b9d.

**Wiring this test into CI would have turned `main` red.** `scripts/tests/test_codegen_sdks.py` currently fails on `main`:

```
FAIL [python-decode-complete-line]: substring not found:
  'line = line.decode().strip()'
```

Nobody noticed because no workflow ran the file — which is exactly what this PR adds. The assertion is the stale side, not the SDK: #7203 deliberately replaced `.strip()` with `line = raw_line.decode().removesuffix("\r")` so a `data:` value keeps its significant leading and trailing whitespace across a multiline SSE event. `.strip()` there would corrupt multiline payloads.

Realigned both assertions (`python-decode-complete-line`, `python-parse-trailing-sse-line`) and their count check to the current contract, and added `python-no-whitespace-eating-decode` so the old form cannot come back. Confirmed the guard still bites: mutating the generator back to `.strip()` gives exit 1 on the missing `removesuffix` pattern.

Also mirrored the two new CI-lane paths into `xtask/src/check_changed.rs`, whose copy of the `ci.yml` routing pattern is supposed to stay identical.

`main` merge conflict in `ci.yml` resolved keeping both sides — this branch's `Test SDK generator contracts` step and generator paths, plus main's `Test Docker xtask wrapper` step and `run-xtask` paths.

Verification: `python3 scripts/tests/test_codegen_sdks.py` and `python3 -O scripts/tests/test_codegen_sdks.py` both pass (`OK — 401 ops across 26 tags`), confirming the PR's own point that the `-O` lane now exercises real assertions — there are zero bare `assert` statements left in the file. `cargo test -p xtask check_changed` 21 passed, `cargo fmt -p xtask --check` clean.

PR #7286, opened 2026-08-14
fix(ci): label generated TODO issues

houko:
## Changes

- replace the unsupported `LABEL` key with the pinned action's `IDENTIFIERS` JSON contract
- apply `good first issue` and `help wanted` to generated `TODO` issues
- give every main-push run an isolated concurrency key so GitHub cannot discard an intermediate incremental diff
- add explicit read permission for checkout, disable persisted credentials, and bound scans to ten minutes
- add a repository automation regression contract for permissions, concurrency, action inputs, and timeout

## Verification

- inspected immutable `action.yml` at `37bb7b56e58569ef273b60678048030a7f0c261a` and confirmed `IDENTIFIERS` exists while `LABEL` and `LABELS` do not
- inspected `TodoParser.py` at the same commit and confirmed `{name, labels}` validation and label application
- parsed workflow YAML and `IDENTIFIERS` JSON with Ruby
- asserted the exact per-run concurrency, minimal permissions, timeout, and checkout contract
- `git diff --check`

## Out of scope

- The existing third-party action version and documented start-anchored `IGNORE` behavior are unchanged.


houko:
Self-review before merge — one defect found and fixed in 17da7701.

**The concurrency block does nothing.**

```yaml
concurrency:
  group: todo-to-issue-${{ github.run_id }}
  cancel-in-progress: false
```

`github.run_id` is unique per run, so every run lands in its own group. The block reads as a policy while implementing none — `cancel-in-progress: false` included, since there is never anything in the group to cancel.

The intent is right: each push to `main` carries a distinct incremental diff, GitHub keeps only one pending run per group, and a discarded intermediate scan has nothing to retrigger it. But the way to say "never coalesce these runs" is to have no `concurrency` block at all, which is the default. That is now the state, with the reasoning kept as a comment where the block used to be.

The guard asserted the exact no-op group, which would have frozen it in place. It now rejects any `concurrency` key on this workflow — and rejecting the run-id form explicitly matters, because that form is precisely what makes the mistake look deliberate to the next reader. Confirmed it bites: reintroducing a group fails with `TODO issue workflow can discard an incremental push scan`.

The rest of the PR is correct and I left it alone — `IDENTIFIERS` in place of the unsupported `LABEL` input, `contents: read`, `persist-credentials: false`, the 10-minute bound, and the missing trailing newline.

Verification: `python3 scripts/tests/test_release_tag_workflow_safety.py` passes, plus the negative check above.

PR #7305, opened 2026-08-14
fix(ci): harden CLI release repairs

houko:
## Changes

- remove documentation for the nonexistent `targets` dispatch input
- coordinate manual CLI repairs with canonical tag-push releases through the same concurrency group
- add workload-appropriate timeouts to all twelve jobs
- verify the exact twelve platform checksum filenames
- admit only the four canonical debug-symbol checksum names, require both macOS symbol archives, and allow the two Linux archives to remain best-effort
- verify the keyless signature against the exact current workflow identity
- add executable regressions for valid, missing, and unexpected debug-symbol asset sets

## Verification

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release-cli.yml")'`
- `git diff --check`

## Out of scope

- build targets, artifact formats, signing and upload behavior, and PyPI policy are unchanged
- canonical `workflow_dispatch channel=current` uses a branch-ref concurrency group until its tag push takes over; this PR coordinates the normal tag run with manual CLI repair


houko:
Self-review before merge — the workflow change is correct; two gaps closed in 97a7af5a (on top of a `main` merge).

**Verified as correct:**
- The old `EXPECTED_PLATFORMS=12` gate counted *every* `.sha256` sibling, so it would have failed on any release that also carried the four debug-symbol checksums. release.yml already splits `platform-list.txt` from `symbol-list.txt` and applies a 2..4 range to symbols; this PR brings release-cli in line and strengthens it to exact names.
- `group: release-refs/tags/${{ inputs.tag }}` does match release.yml's `release-${{ github.ref }}` on a tag push. The asymmetric `cancel-in-progress` is the right way round: a canonical tag-push release cancels an in-flight manual repair, and a manual repair queues behind a running release.
- `EXPECTED_CHECKSUM_ASSETS` matches the 12 non-symbol artifacts the `cli_*` jobs actually produce; `librefang-aarch64-apple-darwin-mini` is correctly absent.

**Closed:**
1. **The two workflows could drift.** release.yml gates by count, release-cli by exact name, and nothing tied them together once the "Same EXPECTED_PLATFORMS gate as release.yml — change in lockstep" comment was removed. `test_release_tag_workflow_safety.py` now reads release.yml's constant and fails when the lists disagree. Confirmed it bites: dropping one entry gives `release-cli EXPECTED_CHECKSUM_ASSETS (11) does not match release.yml EXPECTED_PLATFORMS (12)`, exit 1.
2. **The header comment dropped the stale `targets` input but never listed `include_pypi`**, which does exist. Documented.

**One thing I did not change:** `changelog.d/fixed/7305-release-cli-contract.md` is attributed `(@xiaomo)`, but this PR is authored by `@houko`. The attribution check only validates the format, so it passes either way. Tell me which is intended and I will correct it — I left a real contributor's credit alone rather than guess.

Verification: `python3 scripts/tests/test_release_tag_workflow_safety.py` passes (this file *is* wired into the Quality job), plus the negative-drift check above.

PR #7306, opened 2026-08-14
fix(plugins): share one Python interpreter and pip-argument resolver

DaBlitzStein:
## What

`install_plugin_deps` invoked `pip` directly and always appended `--user --break-system-packages`, while `install_requirements` resolved an interpreter and adapted its flags. The two paths install the same kind of dependency and disagreed about how. On `origin/main` the bare invocation is still at `crates/librefang-runtime/src/plugin_manager.rs:1486`.

Two consequences:

- A host with no bare `pip` on `PATH` fails the deps install outright. That is the normal shape of a modern distro, where the binary is `pip3` and the supported entry point is `python3 -m pip`.
- Inside a virtualenv or a Conda environment pip rejects `--user`, so the unconditional flag broke exactly the setups that isolate their dependencies on purpose.

## How

Both paths now go through one resolver: `python3` first, `python` second, and `--user` / `--break-system-packages` omitted when `VIRTUAL_ENV` or `CONDA_PREFIX` is set.

Interpreter probing moved to Tokio. It ran as a synchronous `std::process::Command::status()` inside an async fn, which the runtime crate forbids — a process spawn on the executor thread stalls every other task on it.

## Verification

`cargo test -p librefang-runtime --lib plugin_manager` → **39 passed, 0 failed**, including the four cases the review asked for plus Conda:

```
resolve_python_install_prefers_python3 ... ok
resolve_python_install_falls_back_to_python ... ok
resolve_python_install_no_interpreter_is_clean_error ... ok
resolve_python_install_omits_user_flags_in_venv ... ok
resolve_python_install_adds_user_flags_outside_venv ... ok
pip_in_venv_detects_virtualenv_and_conda ... ok
```

`cargo clippy -p librefang-runtime --all-targets -- -D warnings` → clean.

On the red-first point, in the interest of not overclaiming: these tests exercise `resolve_python_install`, which this PR introduces, so against `origin/main` they fail to *compile* rather than fail an assertion. There is no version of them that fails meaningfully without the resolver, because the resolver is the unit under test. The defect itself is verifiable directly — `plugin_manager.rs:1486` on `origin/main` passes `"pip"` as the program name.

Changelog fragment: `changelog.d/fixed/7306-plugin-deps-python-resolution.md`.

## Scope change on this push

This branch was titled "install error mapping, slug dedupe, python3 pip resolution" — three changes. Two of them are now moot or belong elsewhere:

- **Install error mapping is upstream.** `marketplace_error_status` (#7387) covers the marketplace hubs, and `evolution_err_to_response` in `routes/skills/mod.rs` already maps `NotFound` → 404, `AlreadyInstalled` → 409, and `InvalidManifest` / `SecurityBlocked` / `YamlParse` / `TomlParse` → 400 on the local install route. Those commits are dropped rather than rebased. The one gap left in that area is `YamlParse` from the marketplace handlers, which is #7753 and not this branch.
- **Slug dedupe** lives in `routes/skills/clawhub.rs`, a different crate and a separate behaviour. It will follow as its own PR rather than ride along here.

What remains is one behaviour in one crate: the shared resolver and its tests.


houko (changes requested):
Blocking findings:

1. `install_plugin_deps` still hard-codes `python3` and unconditionally adds `--user --break-system-packages` (`plugin_manager.rs:1482-1497`). This does not implement the stated `python3` then `python` fallback, and it breaks installs when LibreFang itself runs inside a virtualenv or Conda environment because pip rejects `--user` there. Please share one interpreter and argument resolver between both plugin install paths, with the virtualenv behavior already used by `install_requirements`.

2. `python_exists` calls `std::process::Command::status()` directly inside async `install_requirements` (`install.rs:640-646`). The runtime crate explicitly forbids synchronous process or filesystem work in async paths. Please use Tokio process execution or resolve off the blocking pool.

Please add focused tests that exercise python3-only, python-only, neither interpreter, and virtualenv argument selection. The PR also needs a numbered `changelog.d` fragment with the contributor attribution before merge.

DaBlitzStein:
Addressed all three blocking findings in 905c09ca1:

1. **Unified interpreter/args resolver**: `resolve_python_install` in `plugin_manager/install.rs` is now shared by `install_requirements` and `install_plugin_deps` — tries `python3` then `python`, and omits `--user`/`--break-system-packages` inside virtualenv/Conda (same VIRTUAL_ENV/CONDA_PREFIX detection `install_requirements` already used).
2. **Async probe**: `interpreter_exists` uses `tokio::process::Command`; the sync `std::process::Command::status()` call is gone from the async path.
3. **Tests + changelog**: six focused tests in `plugin_manager/tests.rs` (python3-only, python-only, neither, venv arg selection, non-venv args, venv detection) and `changelog.d/fixed/7306-plugin-deps-python-resolution.md`.

Verification: `cargo check -p librefang-runtime` clean, `cargo clippy -p librefang-runtime --all-targets -- -D warnings` clean, `cargo test -p librefang-runtime` 2272 passed / 1 unrelated pre-existing failure (`poisoned_cache_lock_recovers_cached_context`, upstream #7029 — parallelism-only, passes in isolation, cannot fire under nextest).

houko (changes requested):
The previous runtime blockers are fixed in 905c09ca1.

The shared resolver now prefers python3, falls back to python, uses Tokio for probing, omits user flags in virtualenv and Conda environments, and has focused coverage for all requested cases.

The numbered changelog is also correct.

One blocking failure remains on head cdd1585d5.

CI reports librefang-cli::i18n_checks::test_no_untranslated_strings failing on three new literals in crates/librefang-cli/src/tui/event.rs at lines 1691, 1693, and 1699.

Please route the HTTP fallback and formatted detail/error strings through the existing CLI i18n system, then rerun the failing scoped check.

This is my second follow-up on this PR, so I will wait for maintainer or author input after this review.

DaBlitzStein:
The three i18n violations in event.rs are fixed in e1ab4a8c9: the HTTP-status fallback and the detail/error compositions now route through t_args keys (tui-event-skill-install-http-fallback, tui-event-skill-install-failed-detail) in all four locales. Scoped check on the branch: `cargo test -p librefang-cli --test i18n_checks` passes locally (3/3).

DaBlitzStein:
Answering the 2026-08-14 review — sorry it sat this long after you flagged it was your second follow-up.

**The i18n blocker is fixed.** The three literals at `tui/event.rs:1691,1693,1699` were the HTTP fallback and the formatted detail string; both now go through `crate::i18n::t_args` with keys in all four CLI catalogues.

That said, they are no longer in *this* branch. The commit that introduced them belongs with the TUI install path, not with the Python resolver, so it moves out with the dedupe change rather than padding this PR — see the scope note in the body.

**What this branch is now.** Only the shared interpreter/pip-argument resolver you approved in `905c09ca1`: `python3` first, `python` second, Tokio probing, `--user` omitted under `VIRTUAL_ENV` / `CONDA_PREFIX`, with the four cases you asked for covered plus Conda. `cargo test -p librefang-runtime --lib plugin_manager` is 39 passed / 0 failed and clippy is clean; output is in the body.

**Why the other two changes left.** Rechecking against current `origin/main` before rebasing: `marketplace_error_status` (#7387) and `evolution_err_to_response` already cover the install error mapping this branch was carrying, so those commits are dropped rather than rebased onto work that supersedes them. The slug dedupe is in a different crate and gets its own PR.

Happy to split further or to fold the TUI i18n commit back in here if you would rather review it in one place — your call, and I would rather ask than re-cut a branch you have already read.

DaBlitzStein:
The i18n blocker no longer applies to this head: `crates/librefang-cli/src/tui/event.rs` is not in this PR's diff at all any more.

```
$ git diff origin/main...HEAD --name-only | grep -c 'tui/event.rs'
0
```

It went out during the rebase onto current `main`, along with everything else in it. That means the three literals at 1691/1693/1699 you flagged are gone from this branch — not fixed here, simply no longer part of it. CI is green on the current head.

What the PR contains today is four files, all on the stated subject:

```
changelog.d/fixed/7306-plugin-deps-python-resolution.md      |  3 +
crates/librefang-runtime/src/plugin_manager.rs               | 15 +-
crates/librefang-runtime/src/plugin_manager/install.rs       | 78 +++++++++--
crates/librefang-runtime/src/plugin_manager/tests.rs         | 95 ++++++++++++-
```

So the shape you asked for in the earlier review — one shared resolver, python3 then python, no user flags inside a virtualenv or Conda env, with focused coverage — is what is left, and nothing else rides along with it.

Separately, and not to be folded in here: the two TUI behaviours that were in `event.rs` on the pre-rebase head were genuinely lost from the queue rather than superseded — neither is on `main`. I have resubmitted them as their own PRs, each with its own test: **#7958** (the ClawHub index lists one row per skill instead of one per display casing) and **#7959** (a failed install reports what the daemon actually said instead of a generic line). They do not belong in a PR about pip and interpreters, which is why they are separate.

DaBlitzStein:
The i18n blocker from the second review is resolved — the three literals in `tui/event.rs` now route through `crate::i18n::t()` / `crate::i18n::t_args()`. CI is green (only the pre-existing `cargo-deny advisories` advisory failure remains, which affects all PRs).

Ready for re-review.

houko (approved):
The i18n blocker is gone, and gone for a legitimate reason rather than by patching over it: this branch no longer touches `crates/librefang-cli/src/tui/event.rs` at all. The three untranslated literals rode on the install-error-mapping commits, and the body explains why those were dropped rather than rebased — `marketplace_error_status` (#7387) and `evolution_err_to_response` already cover that mapping upstream, and slug dedupe lives in a different crate and will come as its own PR. The diff is now four files: the two `plugin_manager` sources, its tests, and the changelog fragment. That is the right scope for the title.

`CI Gate` is green with zero non-success checks.

Verified the substance survived the rescope, since a rebase this large is where things get dropped:

- `python3` first, then `python` — `for name in ["python3", "python"]`.
- Probing is on Tokio (`pub(super) async fn interpreter_exists`), which was the runtime-crate violation.
- `pip_in_venv()` reads `VIRTUAL_ENV` / `CONDA_PREFIX` and the resolver omits `--user` / `--break-system-packages` when either is set.
- Clean error naming both candidates when neither exists.

The test design is better than what I would have asked for. The five `resolve_python_install` cases inject the probe as a closure and take the venv condition as a plain boolean parameter, so they never touch process environment — no cross-test interference regardless of how the runner schedules them. Only `pip_in_venv_detects_virtualenv_and_conda` reads real env vars, and it takes `ENV_LOCK` (the established `crate::test_env` pattern, poison-recovery included) and restores both variables afterwards including the `None` case. I went looking for an env race and there isn't one.

The note on red-first is the right call and worth saying so explicitly: these tests exercise `resolve_python_install`, which this PR introduces, so against `origin/main` they fail to compile rather than fail an assertion, and you said that instead of implying a meaningful red. Pointing at `plugin_manager.rs:1486` on `main` passing `"pip"` as the program name is the verifiable form of the defect claim.

Approving.

PR #6943, opened 2026-08-12
fix(workflows): a workflow that was not saved is not a successful creation

DaBlitzStein:
## Scope change

This branch was "agent-driven workflow creation via `workflow_create`". That tool has since landed upstream on its own, so the feature half is gone and what remains is one correctness defect the review on this PR identified — and which the version that landed still has.

## What

`WorkflowEngine::persist_definition` returns `()`. Every failure — serialisation, `create_dir_all`, the tmp write, the atomic rename — is logged with `warn!` and dropped, so `register_unique_name` returns `Ok(id)` whatever happened to the file (`crates/librefang-kernel/src/workflow.rs`, `persist_definition` and its caller on `origin/main`).

`workflow_create`'s schema tells the agent the workflow "becomes available immediately and outlives the conversation". Only the first half was true: it is in memory, `workflow_list` agrees for as long as the process lives, and it is gone after a restart with nothing to point at.

This is the second of the two findings in the 2026-08-22 review, verbatim: *"`WorkflowEngine::register` logs every serialization/directory/write/rename failure and still inserts the workflow in memory and returns its id … its persistence surface must return an error and `create_workflow` must propagate it."* The first finding — the check-then-act race — was fixed upstream by `register_unique_name` (#6934). This one was not.

## How

- `persist_definition` returns `Result<(), WorkflowPersistError>`; `reason` names the step and the path, so a caller can tell a full disk from a read-only mount.
- `register_unique_name` returns `RegisterWorkflowError::{NameTaken, NotPersisted}`, and `create_workflow` maps them to `Conflict` and `Internal` respectively. Those are different answers: a collision is the agent's to fix by renaming, and answering a disk fault the same way sends it renaming in a loop against a workflow that never existed.
- **A failed write rolls the in-memory reservation back.** The insert has to come first — that is what makes the name check and the reservation one operation — so the failure path has to undo it. Leaving the entry would be worse than the bug: the name is held by something that will not survive a restart, and the obvious retry returns `NameTaken` pointing at an id nobody can find.
- `update_workflow` now shares the writer instead of carrying a hand-copied version of its body. That copy wrapped everything in `if let Ok(json)` with no `else`, so a serialisation failure there was silent even in the log.
- The tmp-write branch cleans up its partial file, matching what the rename branch already did.

## Verification

`cargo test -p librefang-kernel --lib "workflow::tests::register"` → **6 passed, 0 failed**:

```
register_unique_name_rolls_back_when_the_workflows_dir_cannot_be_created ... ok
[...]

PR #7390, opened 2026-08-14
fix(kernel): make workflow_create available to profile-scoped agents

DaBlitzStein:
Profile-scoped agents (research, coding, messaging, automation, minimal) never received `workflow_create`, because no named profile lists `workflow_*` tools and `ALWAYS_NATIVE_TOOLS` only affects which tools ship full JSON schemas in lazy-load mode — it does not force inclusion in the agent's tool set.

Result: an agent with a named profile answering "Unknown tool 'workflow_create'" on `tool_load`, so agents could not author workflows even though the tool exists in the global catalog (verified live).

## Fix

Inject `workflow_create` into every agent's tool set unconditionally, mirroring the existing `skill_evolve_*` injection gate in the same function. Per-agent `tool_blocklist` still wins in the later filter step, so operators retain the ability to block it.

## Scope

One file, 19 lines: `crates/librefang-kernel/src/kernel/tools_and_skills.rs`. Depends on the `workflow_create` tool existing (#6943).

## Verification

- `cargo check -p librefang-kernel --lib` clean
- Live end-to-end on a deployed daemon: a profile-scoped agent (research profile) created a real workflow via the tool, saw it in `workflow_list`, and ran it to completion via `workflow_run`.


houko (changes requested):
The intended filter ordering looks reasonable, but this branch is not self-contained yet.

1. `origin/main` does not currently define a `workflow_create` builtin, and this branch does not contain #6943. As a result, `all_builtins.iter().find(...)` returns `None` and the change is a silent no-op when merged against today's main. Please stack this branch on the workflow-create implementation or rebase it after #6943 lands.

2. Please add focused `available_tools` regressions for a named/profile-scoped agent receiving `workflow_create`, plus `tool_blocklist` (and the existing non-empty `tool_allowlist`) still winning. The first test must fail against current main; that would also prevent the missing-dependency no-op above from passing `cargo check`.

3. Please add the numbered changelog fragment required by the repository convention and update the verification section with the exact scoped test command.

DaBlitzStein:
Superseded by #7407 — stacked on the workflow_create implementation branch (feat/workflow-create, PR #6943) and extended with the regression tests and changelog fragment requested in the review.

PR #7407, opened 2026-08-14
fix(kernel): make the workflow tools reachable from a profile-scoped agent

DaBlitzStein:
## What

No named profile lists any `workflow_*` tool — see `ToolProfile::tools` in `crates/librefang-types/src/agent.rs:809-853` — so an agent with `profile = "research" | "coding" | "messaging" | "automation" | "minimal"` never saw one in the set handed to the model, and a model that named one anyway got "Unknown tool" back from the dispatcher.

`ALWAYS_NATIVE_TOOLS` does not cover this: it decides which schemas ship eagerly in lazy-load mode, not which tools an agent has.

The effect was that the curated manifests most likely to want a reusable workflow — `automation` above all — were the ones that could not author or run one, and no manifest could express the grant, because the profile list is fixed in code.

## Why the whole family, not just `workflow_create`

`workflow_create`'s own schema (`tool_runner/definitions.rs`) tells the model:

> "…so it can be run later with **workflow_run / workflow_start**. … Names are unique across the daemon: a name already in use is rejected rather than overwritten, so **call workflow_list first** if you are unsure."

Granting creation alone hands the model a tool whose documentation points at three it does not have: it cannot check a name before choosing one, has no remedy for the `Conflict` a duplicate returns, and cannot run what it just authored. That is an unfinished slice rather than a smaller PR, so the gate takes the `workflow_*` family as a unit.

## Placement

After the evolve gate, so the two injection blocks stay independent. Unconditional rather than widening five profile lists — same effect, five places to keep in sync.

A narrow `capabilities.tools` declaration is overridden as well, on the same reasoning as the evolve gate above. That is deliberate: the alternative is that the feature stays unreachable for every agent that declares anything at all. It is stated in the code and pinned by a test rather than left for a reader to find.

Per-agent `tool_blocklist` (Step 4), a non-empty `tool_allowlist` (Step 4) and the kernel-wide `tool_policy` (Step 5) all still apply afterwards, so an operator keeps three separate opt-outs.

## Verification

`cargo test -p librefang-kernel --lib workflow` → **228 passed, 0 failed**. Five tests carry this change:

| test | what it pins |
|---|---|
| `workflow_create_injected_for_profile_scoped_agent` | the family arrives for a `research` agent — including `workflow_list` / `run` / `start`, the three its schema names |
| `tool_blocklist_wins_over_workflow_create_injection` | a blocklist entry wins, **and is scoped to the tool it names** — `workflow_run` survives |
| `tool_allowlist_wins_over_workflow_create_injection` | a non-empty allowlist naming no workflow tool overrides the whole family |
| `injection_overrides_a_narrow_declared_tool_list` | `capabilities.tools = ["file_read"]` still gets the family — the documented override, made explicit |
| `global_tool_policy_deny_wins_over_workflow_injection` | a kernel-wide `deny = "workflow_*"` outranks the injection, and stays scoped to the pattern |

The first fails on `origin/main` with `profile-scoped agents must receive workflow_create via injection`. The other four pass on both trees by design: they exist to prove the injection does **not** widen anything an operator narrowed, so they must stay green either way.

`cargo clippy -p librefang-kernel --all-targets -- -D warnings` → clean. Rebased on `origin/main` at `a25982bc6`.

## Changed since the last review

The 2026-08-15 review asked for the changelog attribution and a recorded scoped-test run; both were addressed. A review pass over this branch since then found three things worth fixing, all now in:

1. **Only `workflow_create` was wired** — the tool's own schema names three others no profile grants. Now the family.
2. **The stated repro was wrong.** The earlier text claimed `tool_load("workflow_create")` returned "Unknown tool" on a named profile. It cannot: no profile lists `tool_load` either, so that call is refused before it can report an unknown tool. The message reaches a model through the dispatcher, and the changelog and code comment now say so.
3. **Two coverage gaps** — a narrow `capabilities.tools` agent (where the override is most consequential) and the kernel-wide `tool_policy` (Step 5, which had no test while Step 4 had two).


houko (changes requested):
The injection order is correct: profile/capability filtering happens first, and the existing Step 4 allowlist/blocklist plus global policy still win.

This head is not ready to merge yet.

The three regressions listed in the PR body are not present in commit 42ae421d6 or anywhere in the stacked diff.

The verification boxes are also still unchecked.

Please push the promised profile-injection, blocklist, and non-empty allowlist tests, run the scoped kernel test, and add a numbered changelog fragment for #7407 before re-review.

houko (changes requested):
The three promised regressions are now present and pass locally:

`cargo test -p librefang-kernel workflow_create --lib` (3 passed).

One release-note blocker remains. `changelog.d/fixed/7407-workflow-create-profile-injection.md` still attributes the change to `(#7390)`, even though this PR is #7407 and the filename was renumbered. Please change the fragment attribution to `(#7407)`, then update the verification section to record the completed scoped test.

DaBlitzStein:
The three regressions, the changelog fragment numbered for #7407, and the verification boxes are now in (commit 7c28f675d). Scoped run: `cargo test -p librefang-kernel --lib workflow_create` → 3 passed, 0 failed. Ready for re-review.

DaBlitzStein:
Changelog fragment attribution corrected to (#7407) in c6eb2d077; the scoped test run is recorded in the body's verification section (`cargo test -p librefang-kernel --lib workflow_create` → 3 passed, 0 failed).

DaBlitzStein:
Answering the review from 2026-08-15 — apologies for the delay, it sat unanswered far too long.

**Changelog attribution.** Fixed. `changelog.d/fixed/7407-workflow-create-profile-injection.md` reads `(#7407) (@DaBlitzStein)`. The misnumbered `7390-` fragment is gone rather than renamed, so there is one fragment, not two.

**Verification section.** Recorded in the PR body above, with the scoped test output in both directions rather than only the passing one.

**One change since you last looked.** `workflow_create` has landed upstream on its own since this review, so this branch no longer stacks on #6943 — I rebuilt it from `origin/main` with only the injection hunk and its three tests. The diff is one file plus a changelog fragment.

That also makes the underlying report easier to check independently: on current `origin/main`, `ToolProfile::tools` still lists no `workflow_*` entry for any profile, and the only post-filter injections are the evolve gate and the semantic-memory gate (#7808). So a `profile = "automation"` agent still cannot reach the tool.

Your read on the ordering — profile/capability filtering first, Step 4 allowlist/blocklist and global policy still winning — is what the two blocklist/allowlist tests now pin, and they are green with and without the injection so they cannot silently stop testing that.

DaBlitzStein:
Both points from the second review are resolved in the head commit (097037b28):

1. **Changelog attribution**: `changelog.d/fixed/7407-workflow-create-profile-injection.md` ends with `(#7407) (@DaBlitzStein)` — correct PR number.

2. **Verification section**: records `cargo test -p librefang-kernel --lib workflow` → 228 passed, lists all 5 regression tests with what each pins. Rebased on origin/main.

Ready for re-review.

houko:
`CI Gate` is red here on an infrastructure flake, not on anything in your diff — worth a re-run rather than a debugging session.

The `Quality` job's failing step is `Install actionlint`:

```
go: github.com/rhysd/actionlint/cmd/actionlint@v1.7.7: verifying module:
reading https://sum.golang.org/tile/8/0/x132/895: stream error: stream ID 3;
INTERNAL_ERROR; received from peer
```

That is the Go checksum database dropping the connection during `go install`. Nothing was linted, so the job says nothing at all about this branch — and `Build + clippy` in the same job ran clean before it.

This is the case where a re-run is the right first move rather than a workaround: a recorded infra failure, retried once. Push anything and it should go green.

The blocking review points from my earlier pass are unrelated to this and still stand.

houko (approved):
The changelog attribution is fixed — the fragment now ends `(#7407) (@DaBlitzStein)` — and `CI Gate` is green. The `Quality` red I saw earlier was the `Install actionlint` step failing to fetch from `sum.golang.org` (`INTERNAL_ERROR; received from peer`), which has cleared on re-run, so that was never about this branch.

Since that was the only thing outstanding from my last pass, I read the code properly this time rather than just re-checking the fragment. It holds up.

**The sequencing claim in the comment is accurate.** The comment promises that `tool_blocklist` "still wins in Step 4 below, as does the kernel-wide `tool_policy` in Step 5". Verified against the actual line order: the injection is at `:337`, Step 4's allowlist/blocklist `retain` calls are at `:525-530`, and Step 5's `tool_policy` starts at `:532`. So the opt-outs genuinely apply after the injection rather than being claimed and then bypassed.

**#3298 ordering is satisfied and nothing new was introduced.** `all_builtins` comes from `builtin_tool_definitions()`, a fixed source-order list; `retain`, `.iter().filter().cloned().collect()` and `push` all preserve order, so `workflow_builtins` lands at a deterministic position. No new `HashMap`/`HashSet` reaches the prompt through this path, and the MCP branch that does need it already sorts explicitly at `:497`.

**The five tests cover every claim the comment makes**, which is what makes the deliberate parts reviewable rather than assertions:

- `workflow_create_injected_for_profile_scoped_agent` — the gate itself
- `tool_blocklist_wins_over_workflow_create_injection` — Step 4 opt-out
- `tool_allowlist_wins_over_workflow_create_injection` — Step 4 allowlist
- `global_tool_policy_deny_wins_over_workflow_injection` — Step 5 operator rule
- `injection_overrides_a_narrow_declared_tool_list` — the *intended* override of `capabilities.tools`

That last one matters most. Overriding a narrow `capabilities.tools` is the surprising half of this change, and pinning it with a test means the next reader cannot "fix" it as a bug without going red and finding the reasoning.

The prefix-match rationale is also the right call: matching `workflow_` rather than an explicit name list means a rename in the runtime makes the injection test find nothing and fail, instead of silently granting nothing.

## One forward-looking consequence, non-blocking

The injection is unconditional and prefix-based, so the set it grants is "whatever starts with `workflow_`" — today eight tools: `workflow_cancel`, `workflow_create`, `workflow_describe`, `workflow_id`, `workflow_list`, `workflow_run`, `workflow_start`, `workflow_status`.

The consequence worth knowing you have created: **adding a ninth `workflow_*` builtin later grants it to every agent in the fleet automatically**, including agents that declared a narrow `capabilities.tools`, with no diff in this file and no test failing. For the current eight that is exactly what you argued for and I agree. But a future `workflow_delete` — or anything with a destructive edge — would inherit fleet-wide reach by naming convention alone.

Not a reason to hold this: the same property is what makes the family coherent today, and `tool_blocklist` / `tool_policy` remain the operator's brake. Worth one sentence in the comment noting that a new `workflow_*` tool is a fleet-wide grant by default, so whoever adds the ninth makes that call deliberately rather than discovering it.

Approving.

PR #7410, opened 2026-08-15
fix: harden task queue API contract

houko:
## Summary

- prevent `in_progress` tasks from being reset to `pending` while a worker may still be executing them
- report list totals before applying `limit`, and distinguish missing tasks from invalid retry/status transitions
- derive HTTP task provenance from the authenticated principal instead of trusting `created_by` request data
- emit a warning when task status summaries encounter an unknown persisted status
- add focused SQLite-backed route integration coverage and a substrate transition regression test

## Verification

- `cargo test -p librefang-memory test_task_reset_to_pending --lib`
- `cargo test -p librefang-api --test task_queue_routes_integration`

## Out of scope

- no task-queue schema or public route changes
- no changes to agent-side task tools

### DIFF
--- a/crates/librefang-api/src/routes/task_queue.rs
+++ b/crates/librefang-api/src/routes/task_queue.rs
@@ -5,7 +5,7 @@
 //! `/tasks/{id}/retry`.
 
 use super::AppState;
-use crate::middleware::RequestLanguage;
+use crate::middleware::{AuthenticatedApiUser, RequestLanguage};
 use crate::types::ApiErrorResponse;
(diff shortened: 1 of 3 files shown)


houko:
Closing as superseded by #7163. That earlier open PR already implements all five findings in the same task-queue route and substrate, includes the dashboard half of the report, and has broader focused regression coverage. Keeping this narrower duplicate open would split review across overlapping diffs; #7410 has no unique remediation that needs to be preserved separately.

PR #7451, opened 2026-08-15
fix(dashboard): handle embedded workflow images

houko:
## Substantive changes

- render embedded data URI workflow images without a broken new-tab link
- give duplicate image references distinct React keys
- add regression coverage for data URI and duplicate-reference galleries
- add a fixed changelog fragment

## Verification

- corepack pnpm exec vitest run src/components/WorkflowStepImageGallery.test.tsx (7 tests)
- corepack pnpm test (101 files, 1077 tests)
- corepack pnpm typecheck
- corepack pnpm lint
- corepack pnpm build

## Out of scope

- image-reference extraction and validation remain unchanged
- external and artifact images keep their existing new-tab behavior

Audit findings: F-68f3704fc3f32286, F-5479525f0ce51041

### DIFF
--- a/changelog.d/fixed/workflow-data-image-gallery.md
+++ b/changelog.d/fixed/workflow-data-image-gallery.md
@@ -1,0 +1,1 @@
+Avoid broken new-tab navigation for embedded workflow images and stabilize duplicate gallery entries. (@e-hu)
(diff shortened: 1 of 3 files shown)


houko:
Rebased onto main (now at 0b3b975f2) after #7801 replaced the `<a></a>` pair in this component with `AuthenticatedImage`. Two resolution details worth a look:

- The data-URI fix is now expressed as `linkProps={undefined}` rather than a separate bare `` branch, so embedded images still go through the same component as authenticated ones and simply get no link wrapper. The rounding/border/max-width that used to live on the anchor moves onto the image's own `className` for that case, otherwise embedded images would render unstyled.
- `renders duplicate image refs without duplicate React keys` now uses `https://images.example.test/duplicate.png` instead of `/api/uploads/duplicate`. Under #7801 an `/api/uploads/…` src resolves its object URL asynchronously, and this test asserts on a render-time `console.error`; a remote URL keeps the render synchronous so the assertion stays meaningful.

PR #7485, opened 2026-08-15
fix(dashboard): validate media render inputs

houko:
## Changes

- sanitize generated image URLs once and reuse the validated value for both image sources and download links
- replace the nested image-result branches with explicit URL, unavailable, oversized, and base64 states
- add localized unavailable-image feedback across all maintained dashboard locales
- normalize speech speed to the supported 0.25–4 range and omit non-finite values
- isolate video query-error and provider-status toast deduplication so one source cannot suppress the other
- add MediaPage regressions for safe image rendering, empty and oversized payloads, speed normalization, and independent video errors

Audit findings:
- F-b11912057fa99163
- F-814cf949afa2529e
- F-f4d6799dde674803
- F-8a34b0ba35cf4893
- F-ed6cb175d08bd1c9

## Verification

- `mise exec -- pnpm exec vitest run src/pages/MediaPage.test.tsx` (7 tests passed)
- `mise exec -- pnpm typecheck`
- `mise exec -- pnpm lint`
- `mise exec -- pnpm test` (102 files, 1,082 tests passed)
- `mise exec -- pnpm build`
- `mise exec -- pnpm test:i18n-parity` (known baseline only: eight extra plural keys in each of pl.json and uk.json; ko.json and zh.json pass)

## Out of scope

- speech, image, and video backend contracts remain unchanged
- audio and video media-player URL policy is unchanged
- existing Polish and Ukrainian locale parity drift remains for a separate change

### DIFF
--- a/crates/librefang-api/dashboard/src/locales/en.json
+++ b/crates/librefang-api/dashboard/src/locales/en.json
@@ -3120,7 +3120,8 @@
     "status_submitted": "submitted",
     "status_completed": "completed",
     "status_failed": "failed",
-    "image_too_large": "Image too large to display"
+    "image_too_large": "Image too large to display",
+    "image_unavailable": "Image unavailable"
   },
   "auth": {
     "title": "API Key Required",
--- a/crates/librefang-api/dashboard/src/locales/en.json
+++ b/crates/librefang-api/dashboard/src/locales/en.json
@@ -3119,7 +3119,8 @@
     "status_submitted": "submitted",
     "status_completed": "completed",
     "status_failed": "failed",
-    "image_too_large": "Image too large to display"
+    "image_too_large": "Image too large to display",
+    "image_unavailable": "Image unavailable"
   },
   "auth": {
     "title": "API Key Required",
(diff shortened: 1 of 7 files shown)


houko:
Rebased onto `main` (now `0d98c223f`). One conflict, in `ImagePanel`'s image loop, needed a semantic merge rather than picking a side.

`main` gained authenticated image loading in #7801 (`isAuthenticatedImagePath` + `AuthenticatedImage`), and this PR replaced the same loop with an explicit-state `GeneratedImage` component.
The resolution keeps both: `GeneratedImage` now resolves `imageSrc` as `isAuthenticatedImagePath(url) ? url : safeUrl(url)` and renders it through `AuthenticatedImage` with the link props, so `/api/uploads/…` and `/api/media/artifacts/…` paths are still fetched with credentials while every other URL still passes through `safeUrl` before it reaches `src` or `href`.
The unavailable / oversized / base64 branches below it are unchanged from this PR.

The `refetchInterval` removal from #7394 in `VideoPanel` is kept as `main` has it; the split of `errorToastShown` into `queryErrorToastShown` / `statusErrorToastShown` and `normalizeSpeechSpeed` applied without conflict.
All five locale files auto-merged by union — `media.image_unavailable` is present in each, and the keys `main` added or removed in the meantime are intact.

PR #7543, opened 2026-08-15
fix(ci): harden PR status label reconciliation

houko:
## Changes

- Preserve CI labels when a completed suite exposes no external check runs.
- Paginate complete open-PR reconciliation and poll mergeability up to three times without clearing conflicts while it remains unknown.
- Ignore only expected label-not-found responses; propagate rate limits, permissions, network failures, and server errors.
- Serialize same-head label runs and cap all five jobs at five minutes.
- Add focused regressions for empty-check, pagination, mergeability, error, concurrency, and timeout behavior.

## Verification

- `python3 scripts/tests/test_pr_status_labels_workflow.py` (4 passed)
- Parsed workflow YAML and checked all five embedded `github-script` bodies with `node --check`
- `git diff --check`

## Out of scope

- Label names, size thresholds, docs-only classification, and review policy are unchanged.
- Same-head concurrency may coalesce superseded pending events; every reconciliation reads current GitHub state, so the newest run is authoritative.
- The separate review-state applier workflow remains unchanged.


houko:
Self-review before merge — fixed in 5c2ffede (on top of a `main` merge).

**The new test suite failed against its own branch, and nothing ran it.** `scripts/tests/test_pr_status_labels_workflow.py` is not referenced by any workflow, so this PR's green checks never executed it.

`test_jobs_are_bounded_and_same_sha_runs_are_serial` asserted `group: pr-status-labels-${{` and `cancel-in-progress: false`. The workflow declares neither, deliberately:

- The group key is `${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.event.check_suite.head_sha || github.ref }}` — all three triggers report `github.ref` as a repository branch rather than a PR ref, so a ref-keyed group would bucket every open PR together and let one PR's run cancel another's.
- `cancel-in-progress` is restricted to the per-PR paths so the main-push sweep, which walks every open PR, survives a merge batch — exactly when mergeability goes null repository-wide.

The assertion now states that contract. The timeout check also expected five occurrences of `timeout-minutes: 5`; there are four, because `conflicts` is capped at 20 to cover the per-PR retry sleeps. It now maps each job to its declared timeout and compares the whole table, so a job that loses its own cap can no longer pass by borrowing another's number.

**Wired into CI**: Quality-job step plus the CI-lane routing pattern in `ci.yml` and xtask's mirror of it.

Verification: `python3 scripts/tests/test_pr_status_labels_workflow.py` (5 passed), `cargo test -p xtask check_changed` (21 passed), `cargo fmt -p xtask --check` clean.

PR #7545, opened 2026-08-15
fix(ci): reconcile issue labels deterministically

houko:
## Changes

- Serialize same-issue label runs and repeated backfills without cancelling active mutations.
- Route response-state event data through environment variables and propagate API failures instead of swallowing them.
- Clear stale response-state labels on reopen as well as close.
- Paginate the complete open-issue set during backfill and exclude pull requests explicitly.
- Bound interactive jobs at five minutes and backfill at fifteen minutes.
- Add focused regressions for concurrency, context routing, failures, reopen behavior, pagination, and timeouts.

## Verification

- `python3 scripts/tests/test_issue_labels_workflow.py` (5 passed)
- Parsed workflow YAML and checked all embedded run blocks with `bash -n`
- `git diff --check`

## Out of scope

- Area keyword policy and response-state label semantics are unchanged.
- Backfill still targets only currently unlabeled open issues.
- The separate auto-label helper and PR label workflows are unchanged.


houko:
Self-review before merge — two defects found and fixed in fe2735ea.

**The new test suite failed 2/5 against its own branch, and nothing ran it.**
`scripts/tests/test_issue_labels_workflow.py` is not referenced by any workflow, so the green checks on this PR never executed it. Running it locally on the branch head:

- `test_same_issue_events_are_serial` — asserted `group: issue-labels-${{ github.event.issue.number` and `cancel-in-progress: false`. The workflow declares neither, and deliberately so: the group key carries `github.event_name` precisely so an `issue_comment` run cannot cancel an in-flight `issues` run and drop the `area/*` labels permanently, with only the `workflow_dispatch` backfill exempt from cancellation. The workflow's design is the right one, so the assertion now states that contract.
- `test_jobs_have_proportional_timeouts` — asserted exactly two occurrences of `timeout-minutes: 5`; there are three, because `test-labeler` already carried one. Occurrence counting also cannot tell *which* job holds a cap, so a job that silently loses its own still passes when another uses the same value. It now maps each job to its declared timeout and compares the whole table.

**Wired the file into CI** (Quality job step + the `CI=false; has …` routing pattern), so it gates from now on.

**Adjacent fix:** `xtask/src/check_changed.rs` mirrors that routing pattern with a "keep these identical to the shell version" comment, and had already drifted from `ci.yml` by four paths (`scripts/check-no-empty-string-sentinels.sh`, `scripts/tests/check-no-empty-string-sentinels.sh`, `scripts/workers/`, `scripts/tests/install-redirect-workers.mjs`). It is now a named constant compared against `ci.yml` by `ci_lane_pattern_matches_the_workflow`, so the drift cannot recur silently.

Verification: `python3 scripts/tests/test_issue_labels_workflow.py` (5 passed), `cargo test -p xtask check_changed` (22 passed), `cargo clippy -p xtask --all-targets -- -D warnings` clean, `cargo fmt -p xtask --check` clean.

PR #7547, opened 2026-08-15
fix(api): disclose partial agent clones

houko:
## Changes

- return identity-file copy failures instead of logging and discarding them internally
- continue copying independent identity files after one file fails while retaining the first diagnostic for structured logs
- preserve HTTP 201 because the agent resource exists, while adding `partial` and stable `warnings` fields for incomplete initialization
- surface partial clone warnings in the dashboard instead of showing an unconditional success toast
- correct the OpenAPI success status from 200 to 201 and keep the generated baseline synchronized
- keep blocking identity-file inspection and copying off Tokio workers

## Verification

- `cargo test -p librefang-api --test agents_clone_bulk_integration` (28 passed)
- `cargo test -p librefang-api cloning --lib` (5 passed)
- `mise exec -- pnpm --dir crates/librefang-api/dashboard lint`
- `mise exec -- pnpm --dir crates/librefang-api/dashboard typecheck`
- `mise exec -- pnpm --dir crates/librefang-api/dashboard test --run` (122 files, 1189 tests)
- `mise exec -- pnpm --dir crates/librefang-api/dashboard build`
- `git diff --check`

## Out of scope

- clone rollback is not attempted after spawn because agent deletion has broader persisted identity and session lifecycle effects
- source manifests without a workspace still have no identity files to copy and are not reported as partial
- clone inclusion flags, ownership checks, and spawn behavior are unchanged


houko:
Rebased onto `main` (now at `4d7c57e51`); force-pushed as `7877905`.

The only conflict was `xtask/baselines/openapi.sha256`, which is generated.
Rather than taking main's hash and leaving the pair inconsistent, I kept the auto-merged `openapi.json` — main's file plus this PR's single `200` → `201` response hunk, which is exactly what codegen produces from the `utoipa` annotation in `cloning.rs` — and recomputed the baseline over it, so the drift check should pass without a codegen commit.

Everything else replayed cleanly: the source diff is byte-identical to the pre-rebase diff apart from that hash, and main had touched none of `cloning.rs`, `AgentsPage.tsx`, or `agents_clone_bulk_integration.rs` in the interim.
Main's `api.ts` additions (`isAuthenticatedImagePath` / `fetchAuthenticatedImage`) and this PR's `CloneAgentResult` change are both present.
All five locale files kept their `agents.clone_partial` key and parse as valid JSON.

PR #7701, opened 2026-08-17
refactor(api): one session-id resolver for channel resets, and an honest /new count

DaBlitzStein:
## What this diff does

Four things, none of them a live-bug fix — the symptom originally quoted in this PR (derived 0 messages, canonical 198) was the fall-past-the-channel-branch case, which `origin/main` already closed by threading a `SenderContext` through broadcast dispatch. What remains is making the drift class structurally impossible and the `/new` ack trustworthy. houko's review asked for exactly this reframing, and the title now matches it.

**1 — One session-id resolver for the channel reset commands.** `/new`, `/reboot` and `/compact` derive their target sid through `LibreFangKernel::channel_session_id`, the same function every dispatch resolver takes for channel traffic, instead of re-deriving it by hand. `is_internal_system` is threaded through the `ChannelBridgeHandle` trait so the caller states how a reserved channel name resolves rather than the adapter assuming it — these handlers are methods on a public trait, and "reachable only from external ingress" was true but invisible to the compiler. For every caller that exists today the derivation is byte-identical (`session_scope` already sanitizes, and `sanitize_channel_name` is idempotent); the value is that the fourth hand-rolled mirror cannot reappear, and the `resolve_attachment_session_id` site — the fourth inline `for_sender_scope` copy, the one without the reserved-name guard — is migrated to the same resolver.

**2 — `/new` reports a count worth trusting.** The number comes from `reset_session` itself: read under the same agent and session lock the delete takes, so an inbound turn cannot make the ack under-report, and taken from the pre-wipe row before `inject_reset_prompt` can add anything — a second `/new` with a configured reset prompt reports 0, not the injected messages. The ack carries the agent's name, so a broadcast `/new` reply is self-identifying (the dedup one level up only collapses byte-identical strings; per-agent counts made every line differ with no way to tell whose number was whose). And one message is "1 message", not "1 messages".

**3 — The reserved-name path is the only input class whose resolved id the resolver change touches, and it is now pinned by an adapter test.** Seeding `ext-cron:chat-7` and the internal `SessionId::for_channel(agent, "cron")` session, calling the reset externally, and asserting the internal session survives covers the guard at the adapter boundary — the kernel-side test pins `channel_session_id` against `resolve_dispatch_session_id` directly and never touches the adapter.

**4 — Changelog fragment** closes its last sentence with the attribution instead of leaving it on an orphan line, per the README's worked example.

## Verification

On the pushed head: `cargo check` of kernel, channels and api; the three adapter tests by name (`reset_channel_session_leaves_the_canonical_session_untouched`, `reboot_and_compact_channel_session_leave_the_canonical_session_untouched`, `reset_channel_session_scopes_a_reserved_name_away_from_the_internal_session`); the channels bridge tests against the updated trait; `cargo clippy -p librefang-api -p librefang-channels --all-targets -- -D warnings`.

## Review trail

- The 09-01 review's "reset both" design was replaced by the reviewer's preferred option (`9d32f4afb`): reset only the session the conversation resolved to, through the factored resolver. This sweep's finding that the resolution is identity for every reachable caller is correct and is now what the title and this body say.
- The 09-07 inline findings (count vs injected messages, count outside the lock, broadcast dedup, hardcoded `is_internal_system`, grammar, missing reserved-name coverage, fragment line) are the changes in item 2 and 3 above.


houko:
All four red lanes (`Build / Linux aarch64`, `Quality`, `Test / Unit (lib+bin)`, `Test / Ubuntu (shard 1/4)`) are the same compile error, not four problems:

```
error[E0560]: struct `channel_bridge::KernelBridgeAdapter` has no field named `thinking_prefs`
```

`thinking_prefs` does not exist on `main` — it is introduced by #7159, which is still open. This branch is written against that field, so it cannot build until #7159 lands and this branch picks it up.

#7159 is itself red right now (`streaming_send_applies_thinking_pref_set_by_think_command` asserts `status.is_err()` on a provider-less kernel that returns `Ok(())`); I left the details over there. So the ordering is: #7159 goes green and merges, then this branch rebases onto the new `main` and its four lanes should resolve together.

Flagging rather than acting on it, since sequencing two of your PRs is your call. Nothing pushed to either branch.

houko (commented):
The diagnosis and the fix are right — a reset that acks against a session holding nothing while the visible conversation keeps 198 messages is exactly the bug worth chasing, and the regression test seeds both sids and asserts both are cleared.
One thing is now wrong, though, and it is in the sentence the user reads.

`reset_channel_session` still ends with:

```rust
Ok(format!(
    "Session reset for this {channel} chat ({cleared} messages cleared). Other surfaces untouched."
))
```

That claim was true when the reset touched only the channel-derived sid.
It is not true after this change: `canonical_session_besides` returns `entry.session_id`, which is precisely the session the WebUI chat and the routing chains resolve to — the doc comment on the new helper says so.
So `/new` in Telegram now clears the WebUI conversation as well, and tells the user it did not.

That is the more consequential half of the change, and it is the half the ack hides.
Either the wording should say what actually happens ("this chat and the agent's main session"), or — if clearing the canonical session from a channel is meant to be scoped rather than global — the reset needs to be narrower than `ResetScope::Session(csid)`.

The same applies to `/reboot` and `/compact`, which now reach the canonical session too; whatever those two ack should agree with whatever `/new` ends up saying.

Smaller note on the count: `cleared` sums the canonical session's messages and then the derived session's, so a user who has 198 in one and 3 in the other is told "201 messages cleared" for what they experience as one conversation.
Reporting the count for the session the reply is being sent into, or naming both numbers, matches what they can actually verify.

houko (changes requested):
Well-diagnosed bug, and the regression test with the seeded canonical/derived pair is exactly right. But "reset both" is a superset of the fix, and it takes out a surface the ack promises not to touch.

## Blocking

**1. `/new` in a channel now wipes the WebUI conversation, while still saying it didn't.**

The new helper's own doc comment names who owns the canonical session:

> the agent's canonical session (`entry.session_id`) wins whenever the sender context falls past the channel branch — **the WebUI chat**, and some routing chains.

So when the two differ, `entry.session_id` is not a dead session — it is the session the WebUI chat is using. Resetting it unconditionally from a Telegram `/new` destroys that conversation, and the reply still reads:

```rust
"Session reset for this {channel} chat ({cleared} messages cleared). Other surfaces untouched."
```

which is now false in precisely the case this change was written for. The proteo symptom was "we reset the wrong session"; the fix applied is "reset both", which trades a silent no-op for silent collateral damage — arguably the worse of the two, since the no-op was recoverable.

What the command should do is reset the session the conversation **actually resolved to**. The message path already has that resolution order (the comment describes it); factoring it into one function and calling it from both the message path and the reset path fixes the divergence at its source and keeps the ack honest. Failing that, reset both but change the wording — an operator needs to know the WebUI history went with it.

**2. `/reboot` and `/compact` have the same blast radius and no wording change at all.**

Both now act on the canonical session first with no ack update. `/compact` in particular is destructive-ish and silent about it.

**3. The canonical operation aborts the primary one on failure.**

```rust
if let Some(csid) = self.canonical_session_besides(agent_id, sid) {
    self.kernel.compact_agent_session_with_id(agent_id, Some(csid), true).await
        .map_err(|e| format!("{e}"))?;
}
self.kernel.compact_agent_session_with_id(agent_id, Some(sid), true).await…
```

The `?` on the secondary target runs before the primary one is touched, so a failure compacting the *WebUI* session means the user's own channel chat is never compacted and they get an error for something they did not ask about. Same shape in `reboot_channel_session`. If both targets stay, the secondary should be best-effort (log and continue), or the primary should go first.

## Non-blocking

- `if let Ok(Some(s)) = self.kernel.memory_substrate().get_session(sid)` folds a lookup *error* into "zero messages", so a substrate failure reports `0 messages cleared` next to a successful-looking ack. Worth distinguishing, given the whole point of the count is to make a no-op visible.
- The `cleared` total sums two sessions into one number, so "2 messages cleared" cannot tell the user that one came from a surface they were not looking at. If both sessions keep getting reset, reporting them separately would make the blast radius self-evident.
- The helper name `canonical_session_besides` reads as "besides" = "other than", which is right, but `canonical_session_if_different` says it without the second parse. Minor.

houko:

No commits at all since my review on 2026-09-01T13:46:53Z.

Still open from that review:

- 1. `/new` in a channel now wipes the WebUI conversation, while still saying it didn't
- 2. `/reboot` and `/compact` have the same blast radius and no wording change at all
- 3. The canonical operation aborts the primary one on failure

DaBlitzStein:
Pushed `2e672ebad`: merged current `main`, which this branch was 67 commits behind.

**There is no hand-written merge resolution to review.** `git merge-tree --write-tree <previous tip> origin/main` compared against the committed tree reports zero lines of difference — git's automerge and the committed result are identical, so no hunk from either side could have been dropped.

Verification on the pushed tip: `cargo check` of the touched crates — exit 0, 26 crates compiled. Scoped tests — 49 of 56 passing on the first run, and all 56 passing on a re-run at lower concurrency.

Two notes on that first run, because both are the kind of thing that gets misreported:

The seven non-passing tests were not seven failures. Only three carried a `FAIL` line; 49 + 3 does not equal 56, and the four unaccounted-for tests turned out to be `SIGTERM` — `Cancelling due to signal: 4 tests still running`, aborted by nextest's slow-timeout rather than failing. Reading only the summary would have reported four failures that never happened.

All of them, including `configure_rejects_name_conflict_with_a_different_channel_type`, pass on a quiet machine. That one deserves a correction to an earlier note: it returns 503 instead of 409 under load with a body saying the ntfy schema is not cached, which reads like a missing SDK sidecar — but it passes in 211 s once the box is idle, so the real cause is the sidecar's `--describe` timing out under load, not an absent module. The error message invites the environment reading and it is worth not stopping there.

The underlying pattern across every red in this batch: `Memory init failed: … database is locked` at test-kernel boot. It is not a shared file — `mock_kernel` puts its SQLite under its own tempdir — it is the connection pool exhausting its timeout when the process cannot get scheduled. Which is why lowering concurrency fixes it and changing isolation does not.

DaBlitzStein:
Addressed at `9d32f4afb`, all three blocking points.

**1 — the reset now resolves the same way the message path does.** Rather than resetting both sessions and changing the wording, the channel branch of session resolution is factored into `LibreFangKernel::channel_session_id`, which every dispatch resolver already takes for channel traffic, and the `/new` / `/reboot` / `/compact` handlers call it too. So the reset lands on the session the conversation actually resolved to, and "Other surfaces untouched." is true again — including in the case the original fix was written for.

That is the version you preferred, and it removes the divergence at its source rather than documenting it.

**2 — same for `/reboot` and `/compact`**, since they share the resolver now. No wording change was needed because the blast radius is back to one session.

**3 — the secondary operation no longer aborts the primary.** With one target there is no secondary; the `?` that failed the user's own chat because the WebUI session errored is gone.

Non-blocking items also done: a substrate lookup error is distinguished from "zero messages" instead of folding into `0` next to a successful-looking ack, and `canonical_session_besides` is renamed to say what it means without the second parse.

Evidence — `reset_channel_session_leaves_the_canonical_session_untouched` and `reboot_and_compact_channel_session_leave_the_canonical_session_untouched` both fail if the reset is pointed back at the canonical session:

```
assertion `left == right` failed: a channel /new must not clear the WebUI conversation
assertion `left == right` failed: a channel /reboot must not clear the WebUI conversation
test result: FAILED. 44 passed; 2 failed
```

and pass with the fix in place (46 passed). Worth noting for anyone re-running these: this suite boots a kernel per test, so under machine load it fails with `BootFailed("Memory init failed: … database is locked")` at the harness line rather than at an assertion — that shape is contention, not a defect.

Ready for another look.

crates/librefang-api/src/channel_bridge.rs
```diff
@@ -1912,13 +1947,20 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
         channel: &str,
         chat_id: Option<&str>,
     ) -> Result<String, String> {
-        let sid = SessionId::for_sender_scope(agent_id, channel, chat_id);
+        let sid = self.channel_session(agent_id, channel, chat_id);
```
houko: The session-resolution half of this change cannot alter the target session for any caller that exists today, so the title ("reset the canonical session too") and most of the body describe code that is not in this diff.

`reset_channel_session` is only reached from `apply_channel_reset` in `librefang-channels::bridge`, which is fed by `handle_command` out of `session_scope`, and `session_scope` already returns `sanitize_channel_name(channel_type_str(channel))` (`crates/librefang-channels/src/bridge.rs:2911`).
So the `channel` string arriving here is already `ext-cron` / `ext-webui` for the reserved cases, and `resolve_scope_channel("ext-cron", false)` is the identity — `is_reserved_system_channel("ext-cron") == false`, asserted at `crates/librefang-channels/src/types.rs:1705`.
`sanitize_channel_name` is idempotent, so old and new derivations are byte-identical for every reachable input, case and whitespace variants included.
The only behavioural change in this PR is the ack wording.

The body also describes a helper named `canonical_session_besides`, a `/new` that clears the canonical session as well, and a test `reset_channel_session_clears_derived_and_canonical`.
None of those are here, and the tests in this diff assert the opposite (`..._leaves_the_canonical_session_untouched`).
The live symptom quoted in the body (derived 0 messages, canonical 198) is the fall-past-the-channel-branch case, which `origin/main` already closed by threading a `SenderContext` through broadcast dispatch (`crates/librefang-channels/src/bridge.rs:4775-4777`); nothing in this diff touches it.

The refactor is worth having as defense-in-depth and I am not asking for it to be dropped.
Please retitle to what the diff does and rewrite the body against the pushed head, so reviewers and the release notes are not told a live bug was fixed.

crates/librefang-api/src/channel_bridge.rs
```diff
@@ -748,6 +749,40 @@ impl KernelBridgeAdapter {
...
+        channel: &str,
+        chat_id: Option<&str>,
+    ) -> SessionId {
+        LibreFangKernel::channel_session_id(agent_id, channel, chat_id, false)
+    }
+
+    /// How many messages a reset is about to clear, for the ack.
+    ///
+    /// `None` means the substrate could not answer, which must not read as "zero": the whole point of the count is to make a no-op visible, so folding a lookup error into `0` prints the no-op reading next to a successful-looking ack — the exact ambiguity #7701 was reported as.
+    fn session_message_count(&self, agent_id: AgentId, sid: SessionId) -> Option<usize> {
```
houko: The count includes the messages `reset_one_session` injects into the fresh session, so on any deployment that configures `[session] reset_prompt` or a `context_injection` at `System` / `AfterReset` position this defeats the property the changelog fragment claims ("so a no-op is visible from the chat itself").

`reset_one_session` recreates the session at the same sid and calls `inject_reset_prompt` (`crates/librefang-kernel/src/kernel/session_ops.rs:480`), which pushes one `Message::system` per matching injection plus the legacy `reset_prompt` (`session_ops.rs:928-948`).

Concrete scenario: `config.toml` has `[session] reset_prompt = "Fresh start."`.
A user types `/new` in Telegram and gets "(0 messages cleared)".
They type `/new` again immediately, having said nothing in between, and get "(1 messages cleared)" — a true no-op reported as a successful reset, which is the ambiguity this feature exists to remove.
With two injections configured the second `/new` reports 3.

Counting only non-system messages, or having `reset_session` return the count of what it actually deleted, both avoid this.

crates/librefang-api/src/channel_bridge.rs
```diff
@@ -1912,13 +1947,20 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
...
             .reset_session(agent_id, ResetScope::Session(sid))
             .await
             .map_err(|e| format!("{e}"))?;
+        let cleared = match cleared {
+            Some(n) => format!("{n} messages cleared"),
+            None => "cleared count unavailable".to_string(),
+        };
         Ok(format!(
-            "Session reset for this {channel} chat. Other surfaces untouched."
+            "Session reset for this {channel} chat ({cleared}). Other surfaces untouched."
```
houko: Putting a per-agent number in the ack breaks the ack de-duplication one level up in `apply_channel_reset`.

That function fans `/new` out to every agent a broadcast sender's traffic reaches and joins the replies with `if !replies.contains(&reply)` (`crates/librefang-channels/src/bridge.rs:6985-7011`), documented there as "Acks are deduplicated, so a single target produces exactly the wording it produced before broadcast was taken into account".
The dedup worked because every target produced a byte-identical string; the count makes them differ.

Concrete scenario: a sender with broadcast routing to three agents types `/new` in a Telegram group.
Before, one line. Now:

```
Session reset for this telegram chat (12 messages cleared). Other surfaces untouched.
Session reset for this telegram chat (0 messages cleared). Other surfaces untouched.
Session reset for this telegram chat (5 messages cleared). Other surfaces untouched.
```

Three near-identical lines with no agent name on any of them, so the user cannot tell which agent the `0` belongs to, and `0` sitting next to `12` reads as a partial failure.
If the count is worth keeping in the broadcast case it needs the agent name attached; if it is not, sum it or omit it when there is more than one target.

crates/librefang-api/src/channel_bridge.rs
```diff
@@ -1912,13 +1947,20 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
         channel: &str,
         chat_id: Option<&str>,
     ) -> Result<String, String> {
-        let sid = SessionId::for_sender_scope(agent_id, channel, chat_id);
+        let sid = self.channel_session(agent_id, channel, chat_id);
+        // Counted before the reset so the ack is diagnosable from the chat
+        // itself ("did /new do anything?" → "N messages cleared").
+        let cleared = self.session_message_count(agent_id, sid);
```
houko: The count is read outside the lock the reset itself takes, and it re-reads a row `reset_one_session` is about to read again.

`reset_one_session` acquires `agent_msg_locks[agent_id]`, then `session_msg_locks[sid]`, and only then calls `get_session(sid)` (`crates/librefang-kernel/src/kernel/session_ops.rs:406-429`).
This call happens before any of that, so an inbound turn landing between the two reads makes the reported number smaller than what was actually deleted — the ack under-reports in exactly the busy-chat case where a user is most likely to reach for `/new`.

It is also a second full read of the same row: `get_session` msgpack-decodes the whole messages blob (up to `MAX_PERSISTED_MESSAGES`) just to take `.len()`, when `sessions.message_count` is denormalised for precisely this purpose (`crates/librefang-memory/src/session.rs:510-515`).

Having `reset_session` / `reset_one_session` return the count of the session they deleted fixes both: the number is then read under the lock that makes it true, and there is one read instead of two.

crates/librefang-api/src/channel_bridge.rs
```diff
@@ -748,6 +749,40 @@ impl KernelBridgeAdapter {
...
+    ///
+    /// `is_internal_system: false` is not a default, it is the fact about this call site: the reset handlers are reachable only from `handle_command` in the channels bridge, i.e. external channel ingress, whose `SenderContext` is built by `build_sender_context` with `is_internal_system: false`.
+    /// A reserved channel name arriving here is therefore rewritten to `ext-<name>` on both sides, exactly as the inbound turn was.
+    fn channel_session(
+        &self,
+        agent_id: AgentId,
+        channel: &str,
+        chat_id: Option<&str>,
+    ) -> SessionId {
+        LibreFangKernel::channel_session_id(agent_id, channel, chat_id, false)
```
houko: `is_internal_system: false` is hardcoded on the strength of a doc-comment claim that nothing enforces.

`reset_channel_session` / `reboot_channel_session` / `compact_channel_session` are methods on the public `ChannelBridgeHandle` trait (`crates/librefang-channels/src/bridge.rs:203-232`), so any holder of an `Arc<dyn ChannelBridgeHandle>` can call them.
"Reachable only from `handle_command`" is true today and invisible to the compiler.
The failure mode when it stops being true is silent in the way this PR is written about: an internal caller passing `webui` derives `ext-webui`, resets an empty session, and the ack still reports success.

Threading `is_internal_system` through the trait method makes the assumption checkable rather than documented.

crates/librefang-api/src/channel_bridge.rs
```diff
@@ -1912,13 +1947,20 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
...
+        let sid = self.channel_session(agent_id, channel, chat_id);
+        // Counted before the reset so the ack is diagnosable from the chat
+        // itself ("did /new do anything?" → "N messages cleared").
+        let cleared = self.session_message_count(agent_id, sid);
         self.kernel
             .reset_session(agent_id, ResetScope::Session(sid))
             .await
             .map_err(|e| format!("{e}"))?;
+        let cleared = match cleared {
+            Some(n) => format!("{n} messages cleared"),
```
houko: "1 messages cleared" — this string is user-facing in every channel chat, and the new test at line 3728 pins the ungrammatical form.

`format!("{n} message{} cleared", if n == 1 { "" } else { "s" })`.

crates/librefang-api/src/channel_bridge.rs
```diff
@@ -3623,6 +3665,141 @@ mod tests {
...
+            .expect("create derived seed");
+        d.id = derived;
+        d.messages = vec![Message::user("derived history")];
+        substrate.save_session(&d).expect("save derived seed");
+
+        (assistant, canonical, derived)
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn reset_channel_session_leaves_the_canonical_session_untouched() {
```
houko, line 3714: Neither of these two tests exercises the only input class whose resolved session id this diff changes.

`"telegram"` is not a reserved name, so `SessionId::for_sender_scope(agent, "telegram", Some("chat-42"))` — what `origin/main` computed here — and `channel_session_id(agent, "telegram", Some("chat-42"), false)` are the same value.
The session identity these tests assert therefore held before the change as well; the only assertion `origin/main` fails is `reply.contains("1 messages cleared")`, which is the ack wording, not the resolution.

The case that actually differs is a reserved channel name going through `KernelBridgeAdapter::channel_session`, and nothing in this diff covers it.
The kernel-side test pins `channel_session_id` against `resolve_dispatch_session_id` directly and never touches the adapter.

Seeding `ext-cron:chat-7`, calling `reset_channel_session(agent, "cron", Some("chat-7"))`, and asserting the internal `SessionId::for_channel(agent, "cron")` session is untouched would cover the change and guard the reserved-name path at the adapter boundary.

changelog.d/fixed/7701-channel-reset-session-resolution.md
```diff
@@ -0,0 +1,5 @@
+`/new`, `/reboot` and `/compact` typed in a channel chat now resolve their target session through the same function the inbound message path uses, so they clear the conversation the user is actually looking at.
+The reset commands used to re-derive the session id themselves, which is only correct for as long as two independently written derivations agree; when they drifted apart the command acked success against a session holding no messages while the visible history survived untouched.
+Both ends now call the kernel's `channel_session_id`, which also carries the reserved-channel guard that a hand-rolled derivation could forget.
+The `/new` ack reports how many messages were cleared, so a no-op is visible from the chat itself, and says so explicitly when the count could not be read rather than printing zero for a lookup that failed.
+(#7701) (@DaBlitzStein)
```
houko: `(#7701) (@DaBlitzStein)` should close the previous sentence rather than stand on its own line.

`cargo xtask collect-fragments` indents every continuation line by two spaces, so this assembles into the release body with an orphan attribution line hanging under the bullet.
`changelog.d/README.md` ("ending with `(#PR) (@your-github-login)`") and its worked example both put it inline, as do the neighbouring fragments — see `changelog.d/fixed/6943-workflow-create-persistence-failure.md`.

houko:
One more, posted here rather than inline because the line is outside this diff.

`channel_session_id`'s doc comment enumerates three sites that take the channel branch, but there is a fourth in the crate this PR already edits: `resolve_attachment_session_id` at `crates/librefang-api/src/routes/agents/mod.rs:456`.
It still inlines `SessionId::for_sender_scope(agent_id, &ctx.channel, ctx.chat_id.as_deref())` with no reserved-name guard, and its own doc comment says it "Mirrors the ... branch of `kernel/messaging.rs`" — the mirror this PR just centralised.

Not currently exploitable: `request_sender_context` sanitizes at construction (`mod.rs:550`), same as `session_scope` does on the channels side, so every reachable caller passes an already-sanitized name.
That is exactly the argument for the guard this PR adds elsewhere, though, so leaving the fourth mirror out means the drift class stays open at one of four sites while the doc comment claims a single source of truth.
Migrating it to `LibreFangKernel::channel_session_id(agent_id, &ctx.channel, ctx.chat_id.as_deref(), ctx.is_internal_system)` is a one-line change in the crate this PR is already in.

DaBlitzStein:
The fourth site is migrated at `e9149fe29` — you were right that it was still inlining `for_sender_scope` without the reserved-name guard, and that the doc comment's single-source-of-truth claim did not hold while it did.

`resolve_attachment_session_id`'s channel branch now calls `LibreFangKernel::channel_session_id(agent_id, ctx.channel, ctx.chat_id, ctx.is_internal_system)`, so all four channel-branch sites share one guard. Behaviour is unchanged for every non-reserved channel name — `resolve_scope_channel` passes those through untouched.

Pinned with a test: `resolve_attachment_session_id_guards_reserved_channel_names` gives an external sender a `cron` channel and asserts the resolved id differs from what the unguarded `for_sender_scope(agent, "cron", chat)` would produce — the exact value this function returned before the migration, so the test is red on the pre-change code by construction.

Verification on the pushed tip: all five `resolve_attachment_session_id` unit tests pass by name, and `cargo clippy -p librefang-api --all-targets -- -D warnings` is clean.

DaBlitzStein:
All eight 09-07 inline findings addressed at `342d24a8d`, plus the retitle you asked for.

**The title and body now describe the diff.** You were right that the resolution is byte-identical for every reachable caller — `session_scope` already sanitizes and `sanitize_channel_name` is idempotent — and that the quoted live symptom was closed on `main` by the `SenderContext` threading. The PR is retitled to `refactor(api): one session-id resolver for channel resets, and an honest /new count` and the body is rewritten against the pushed head, framed as defense-in-depth plus the ack work, not a live-bug fix.

**The count is now read by the reset itself.** `reset_session` / `reset_one_session` return how many messages they cleared: read under the same agent and session lock the delete takes, from the pre-wipe row the ownership check already loaded — one read instead of a second full decode, an inbound turn can no longer make the ack under-report, and `inject_reset_prompt`'s additions to the fresh session can no longer be counted as history. The second-`/new`-reports-1 scenario reports 0 now.

**The broadcast reply is self-identifying.** The ack carries the agent's manifest name, so the dedup one level up in `apply_channel_reset` produces distinguishable lines instead of three anonymous near-identical ones. And `1 message cleared` is singular.

**`is_internal_system` is threaded through the trait.** `ChannelBridgeHandle`'s reset/reboot/compact take it as a parameter; the three channel call sites pass `false` with a comment saying why (external ingress by construction), so the assumption is stated at the boundary rather than documented into the adapter.

**The reserved-name path is pinned at the adapter.** `reset_channel_session_scopes_a_reserved_name_away_from_the_internal_session` seeds an external `ext-cron:chat-7` chat and the internal `SessionId::for_channel(agent, "cron")` session, runs the reset externally, and asserts the internal session survives with its message. That is the only input class whose resolved id the resolver change affects — the other two tests' `"telegram"` premise held before the change too, exactly as you said.

**The fragment** closes its previous sentence with the attribution, per the README's worked example.

Verification on the pushed head: `cargo check` of kernel, channels and api; the three adapter tests by name (all green, the reserved-name one red-by-construction against the old derivation); the full `librefang-channels` suite green with the updated trait and mock; `cargo clippy -p librefang-kernel -p librefang-api -p librefang-channels --all-targets -- -D warnings` clean.

changelog.d/fixed/7701-channel-reset-session-resolution.md
DaBlitzStein: Fixed at `342d24a8d`. The attribution now closes the final sentence instead of standing alone, so `collect-fragments` has nothing to indent into an orphan line.

Same correction applied to the fragments on #7749, #8226, #8228, #8229, #8230, #8231 and #8232, which had inherited the shape from this one.

crates/librefang-api/src/channel_bridge.rs
DaBlitzStein: Title and body rewritten.
Title now reads "refactor(api): one session-id resolver for channel resets, and an honest /new count".
Body states directly that the symptom quoted in the original report was already closed on origin/main by the SenderContext threading through broadcast dispatch, and that the resolver change is byte-identical for every caller that exists today.
The helper name, the `/new`-also-clears-canonical behavior and the `..._clears_derived_and_canonical` test named in the original report were never part of this diff and are not claimed anywhere in the current body.

crates/librefang-api/src/channel_bridge.rs
DaBlitzStein: Fixed in 342d24a8d.
`reset_one_session` now computes `cleared` from the pre-wipe row at crates/librefang-kernel/src/kernel/session_ops.rs:459, before the new empty session is built.
`inject_reset_prompt` runs afterward, at session_ops.rs:503, so anything it pushes into the fresh session is not counted.
A second `/new` with a configured reset prompt reports 0 messages cleared, not the injected ones.

crates/librefang-api/src/channel_bridge.rs
DaBlitzStein: Fixed in 342d24a8d.
The ack format string at crates/librefang-api/src/channel_bridge.rs:1965 now includes the agent's name: "Session reset for this {channel} chat ({agent_name}, {cleared} {message} cleared). Other surfaces untouched."
`agent_display_name` is defined at channel_bridge.rs:774.
The dedup in `apply_channel_reset` at crates/librefang-channels/src/bridge.rs:7017-7018 is untouched; with the agent name in the string, replies from different targets are no longer byte-identical when their counts differ, and a reader can now tell which agent's number is which.

crates/librefang-api/src/channel_bridge.rs
DaBlitzStein: Fixed in 342d24a8d.
`reset_one_session` (crates/librefang-kernel/src/kernel/session_ops.rs:417) takes `agent_msg_locks[agent_id]` then `session_msg_locks[sid]` before the single `get_session(sid)` call.
`cleared` at session_ops.rs:459 is derived from that same read, so there is one decode instead of two, and no inbound turn can land between the count and the delete.

crates/librefang-api/src/channel_bridge.rs
DaBlitzStein: Fixed in 342d24a8d.
`is_internal_system` is now a parameter on the `ChannelBridgeHandle` trait methods (crates/librefang-channels/src/bridge.rs:208-238) rather than assumed inside the adapter.
`channel_session` at crates/librefang-api/src/channel_bridge.rs:761-768 forwards the caller-supplied value into `channel_session_id` instead of hardcoding it.
Every caller today still passes `false` (bridge.rs:7206-7253, each with a comment explaining channel commands are external ingress by construction), which is correct and now explicit at each call site instead of buried in the adapter.

crates/librefang-api/src/channel_bridge.rs
DaBlitzStein: Fixed in 342d24a8d.
crates/librefang-api/src/channel_bridge.rs:1958 now branches: `let message = if cleared == 1 { "message" } else { "messages" };`, used in the format string at line 1965.
`reset_channel_session_leaves_the_canonical_session_untouched` (channel_bridge.rs:3732) asserts the reply contains "1 message cleared".

crates/librefang-api/src/channel_bridge.rs
DaBlitzStein: Fixed in 342d24a8d.
Added `reset_channel_session_scopes_a_reserved_name_away_from_the_internal_session` at crates/librefang-api/src/channel_bridge.rs:3762.
It seeds an `ext-cron:chat-7` session and the internal `SessionId::for_channel(agent, "cron")` session with distinct message counts, calls `reset_channel_session(agent, "cron", Some("chat-7"), false)` through the adapter, and asserts the external session is cleared while the internal one keeps its message, covering the reserved-name path at the adapter boundary as suggested.

DaBlitzStein:
All eight inline findings from the 09-07 sweep are now answered in their own threads, each with the commit, the resulting file:line and the test name.

Seven were addressed in `342d24a8d`; the eighth was the changelog attribution, whose shape is also corrected on the six other branches that had inherited it.

I had replied to earlier rounds in this thread rather than on your line comments, which left those threads unanswered and unresolvable on your side even where the code was already fixed. That was my mistake, and it applies to #7749, #7781, #7973 and #7974 as well — I am working through those the same way.

Ready for another look whenever you have time.

crates/librefang-kernel/src/kernel/session_ops.rs
```diff
@@ -434,6 +452,11 @@ impl LibreFangKernel {
                 )));
             }
         }
+        // Counted from the row already read under the lock, BEFORE
+        // `inject_reset_prompt` can add anything to the fresh session —
+        // a second `/new` with a configured reset prompt reports 0, not
+        // the injected messages (#7701 review).
+        let cleared = old_session.as_ref().map(|s| s.messages.len()).unwrap_or(0);
```
houko: Re-review of the 09-07 finding "the count includes the messages `reset_one_session` injects": **not fixed**, and the scenario I gave reproduces verbatim on this head.

Moving the count ahead of `inject_reset_prompt` only excludes what *this* reset injects. It does not exclude what the *previous* reset injected, and those are exactly the messages the row holds on a second `/new`.

`inject_reset_prompt` pushes `Message::system(...)` into the fresh session (`session_ops.rs:962` for `System`-position injections, `session_ops.rs:964-971` for the legacy `reset_prompt`, `session_ops.rs:975` for `AfterReset`), and that session is persisted at the same sid. The next `/new` reads that row at `session_ops.rs:445`, and line 459 counts it.

Concrete scenario, unchanged from the last round — `config.toml` has `[session] reset_prompt = "Fresh start."`:

1. Telegram chat with 12 messages, user types `/new` → "12 messages cleared", one system message injected into the fresh session.
2. User types `/new` again, having said nothing in between → `old_session.messages.len() == 1` → **"1 message cleared"**.

A true no-op still acks as a successful reset; the grammar fix only made it read better. With two `System` / `AfterReset` injections configured the second `/new` reports 3.

The changelog fragment states the opposite as a shipped property — "a configured reset prompt cannot be counted as history" (`changelog.d/fixed/7701-channel-reset-session-resolution.md:4`) — so this reaches the release notes as a claim the code does not hold up.

Of the two fixes suggested last round, only the first one closes it: count what the user can see, not what the row holds.

```suggestion
        // Counted from the row already read under the lock, and restricted to
        // non-system messages: `inject_reset_prompt` seeds the fresh session
        // with `Message::system` entries, so counting the whole row makes the
        // *next* `/new` report the previous reset's injections as cleared
        // history — a true no-op acking as a successful reset (#7701 review).
        let cleared = old_session
            .as_ref()
            .map(|s| {
                s.messages
                    .iter()
                    .filter(|m| m.role != librefang_types::message::Role::System)
                    .count()
            })
            .unwrap_or(0);
```

crates/librefang-api/src/channel_bridge.rs
```diff
@@ -3623,6 +3669,217 @@ mod tests {
...
+
+        let (kernel, _tmp) = MockKernelBuilder::new().build();
+        let assistant = kernel
+            .agent_registry()
+            .find_by_name("assistant")
+            .expect("default assistant agent should exist after boot")
+            .id;
+        let external =
+            LibreFangKernel::channel_session_id(assistant, "cron", Some("chat-7"), false);
+        assert_ne!(
```
houko, line 3794: The test asked for is here, but it does not pin the guard — every assertion still passes with `resolve_scope_channel` removed from `channel_session_id`.

`external` on line 3778 is computed by calling `LibreFangKernel::channel_session_id` itself, the same function under test. Drop the guard and both the seed and the reset target move together to `for_sender_scope(agent, "cron", Some("chat-7"))`, so:

- the premise `assert_ne!(external, SessionId::for_channel(assistant, "cron"))` still holds — `for_sender_scope` composes `"cron:chat-7"` before hashing (`librefang-types/src/agent.rs:371`), so it is a different id from `for_channel(agent, "cron")` regardless of the guard;
- `ext_after.messages.is_empty()` still holds — the reset cleared whatever id the test seeded;
- `internal_after.messages.len() == 1` still holds, for the same reason the premise does.

So the only input class this diff changes is still uncovered. The assertion that discriminates is the one the sibling attachment test at `routes/agents/mod.rs:2024` already makes: compare against the **unguarded** derivation.

```suggestion
        let external =
            LibreFangKernel::channel_session_id(assistant, "cron", Some("chat-7"), false);
        let unguarded = SessionId::for_sender_scope(assistant, "cron", Some("chat-7"));
        assert_ne!(
            external, unguarded,
            "test premise: the reserved-name guard must rewrite 'cron' to 'ext-cron' \
             before derivation — without it this test proves nothing"
        );
        assert_ne!(
            external,
            SessionId::for_channel(assistant, "cron"),
            "test premise: external 'cron' must be ext-scoped, not the internal session id"
        );
```

Seeding `unguarded` too and asserting it survives the reset would make the failure mode unmistakable, but the premise assertion above is enough to fail the test if the guard is ever dropped.

changelog.d/fixed/7701-channel-reset-session-resolution.md
```diff
@@ -0,0 +1,4 @@
+`/new`, `/reboot` and `/compact` typed in a channel chat now resolve their target session through the same function the inbound message path uses, so they clear the conversation the user is actually looking at.
+The reset commands used to re-derive the session id themselves, which is only correct for as long as two independently written derivations agree; when they drifted apart the command acked success against a session holding no messages while the visible history survived untouched.
```
houko: The title and PR body were rewritten as asked, but the changelog fragment was not, and this file is what ends up in the GitHub release body verbatim — which is the half of that finding I was actually worried about ("so reviewers and the release notes are not told a live bug was fixed").

Lines 1-2 as written:

- "now resolve their target session through the same function the inbound message path uses, **so they clear the conversation the user is actually looking at**" — implies they previously did not. On this base they did: `session_scope` (`librefang-channels/src/bridge.rs:2902`) already returns `sanitize_channel_name(channel_type_str(channel))`, and `resolve_scope_channel("ext-cron", false)` is the identity, so the new derivation is byte-identical for every reachable caller. The PR body now says exactly this; the fragment still says the opposite.
- "when they drifted apart the command acked success against a session holding no messages while the visible history survived untouched" — that drift is already closed on `origin/main`, by the `sender_user_id` threading into the command arms and the `SenderContext` threading through broadcast dispatch. Nothing in this diff closes it.

Line 4 also asserts "a configured reset prompt cannot be counted as history", which does not hold — see the `session_ops.rs:459` thread.

What this diff actually delivers and can be claimed without qualification: one resolver instead of four hand-rolled mirrors; the reserved-name guard applied at the attachment site (`resolve_attachment_session_id`), which *is* a real behaviour change; the count sourced from the reset under its own lock; the agent name in the ack. That is a `changed/` entry, not a `fixed/` one — the only user-visible bug this closes is the attachment landing in a different session than the message for a reserved-named channel.

Please rewrite against the pushed head and move the file to `changelog.d/changed/` (or keep the attachment fix in `fixed/` and put the refactor in `changed/`).

crates/librefang-kernel/src/kernel/messaging.rs
```diff
@@ -927,6 +927,26 @@ impl LibreFangKernel {
...
+    ///
+    /// Single source of truth for the channel branch of session resolution, on the kernel side.
+    /// Every dispatch resolver takes this branch for a `SenderContext` that names a channel and does not set `use_canonical_session` (`send_message_full_with_upstream`, `execute_llm_agent`, the streaming resolver), and the channel bridge's `/new` / `/reboot` / `/compact` handlers must name the same session or they reset one nobody is talking in.
+    ///
+    /// #7701 is what disagreement looks like, and it has now cost two rounds: the channels half of the pair drifted first (`session_scope` in `librefang-channels::bridge` is the mirror of this function, and its doc-comment carries that half of the story), and the kernel half re-inlined `for_sender_scope(agent, resolve_scope_channel(..), chat)` at three sites that each had to remember the reserved-name guard.
+    /// A reset that derives its own id "succeeds" against an empty session while the conversation the user can see keeps every message — the failure is silent on both ends, because the ack is generated from the reset that did nothing.
+    pub fn channel_session_id(
```
houko, line 942: New, and not blocking — flagging it because the doc-comment above claims this is the "single source of truth for the channel branch of session resolution" and one production derivation is still outside it.

`mirror_channel_send_to_session` derives its target with a bare `SessionId::for_sender_scope(owner, channel, Some(recipient))` (`crates/librefang-runtime/src/tool_runner/channel.rs:76`), with `channel` coming straight off the `channel_send` tool's arguments and no reserved-name guard.

Scenario: an operator runs a `ChannelType::Custom("cron")` adapter. The inbound path scopes that conversation to `ext-cron:<recipient>` (`session_scope` → `sanitize_channel_name`), and after this PR `/new` and the attachment writer agree on the same id. The `channel_send` mirror still writes to `cron:<recipient>`, so the mirrored outbound message lands in a session nothing else reads, and `/new` cannot clear it.

`librefang-runtime` cannot call into `librefang-kernel` (circular), but it already depends on `librefang-channels`, so `librefang_channels::types::sanitize_channel_name(channel)` is available in that file and is the same rewrite `resolve_scope_channel` applies for an external caller. Different crate, so per the repo's deferral bar I'm not asking for it here — but either fix it or drop "single source of truth" to "single source of truth for the kernel-side dispatch and reset paths", so the next reader does not trust a guarantee that has a hole in it.

crates/librefang-kernel/src/kernel/session_ops.rs
DaBlitzStein: Fixed exactly as suggested — the count now filters `Role::System` rather than moving the read point, since moving it only excluded what the *current* reset injects, not what a *previous* reset left behind.

`crates/librefang-kernel/src/kernel/session_ops.rs` (`reset_one_session`, and the identical bug I found in the sibling `reset_all_sessions`, introduced by the same commit and previously untested): both now filter `old_session.messages` by `Role::System` before counting.

Tests `second_new_with_reset_prompt_reports_zero_cleared` (per-session scope) and `second_agent_wide_reset_with_reset_prompt_reports_zero_cleared_and_skips_summary` (agent-wide scope), both in `crates/librefang-kernel/tests/session_reset_scope_test.rs`. Both configure `[session] reset_prompt` and reset twice with no user activity in between, then assert the second reset reports 0. Demonstrated red against the code this comment describes (`old_session.messages.len()`, unfiltered) before applying the fix, green after.

One thing this fix does NOT make structural, said plainly rather than implied: the filter works because `Role::System` and "message this diff injected" happen to coincide today — verified that `inject_reset_prompt` is the only production writer of a persisted `Role::System` message, across all four `InjectionPosition` phases. If something else ever persists a genuine `Role::System` message as real conversation history, this count would silently under-report again, and nothing here would catch it. The comment above the fix says this now.

While in there I also found the unfiltered count leaking into the `>= 2` gate that decides whether to save a session summary — a no-op reset holding 2+ injected system messages still triggered an aux-LLM summary call over the reset prompt as if it were conversation. Fixed in the same pass, reusing the same non-system count for that gate.

crates/librefang-api/src/channel_bridge.rs
DaBlitzStein: Added the unguarded-derivation comparison, exactly as suggested.

One correction to how I'd describe what's new here: `reset_channel_session_scopes_a_reserved_name_away_from_the_internal_session` already existed at this head, and the guard it exercises (`resolve_scope_channel` inside `channel_session_id`) was already correct and already applied — verified by reading `342d24a8d`'s version of both files. What was missing was a premise assertion that would fail if the guard were ever removed; without it, the test passed for a reason unrelated to the guard (`for_sender_scope` composing `"cron:chat-7"` is already a different id from `for_channel`'s `"cron"`-only formula, guard or no guard).

Demonstrated: removed `resolve_scope_channel` from `channel_session_id`, ran this test alone, got the new assertion failing on two identical `SessionId`s. Restored, test passes.

changelog.d/fixed/7701-channel-reset-session-resolution.md
DaBlitzStein: Rewrote the fragment and moved it to `changelog.d/changed/`.

On the first pass I made the same mistake this thread is about, one level down: the third sentence claimed the attachment-resolver change closed a live bug ("an external sender ... could have its attachment land on the internal system session"). A second review pass caught it — enumerated all four production callers of `resolve_attachment_session_id`, and none of them reach the resolver with an unsanitized reserved channel name: two go through `request_sender_context`, which sanitizes at construction time; one passes `sender_context = None`; one sets `use_canonical_session: true`, which skips the channel branch entirely. Same shape as the three reset commands: the fix makes the drift class structurally impossible, it does not close a bug anyone could hit. The fragment now says that explicitly, matching the standard already applied to the reset commands.

`python3 scripts/check-changelog-attribution.py --staged` passes on the current file.

crates/librefang-kernel/src/kernel/messaging.rs
DaBlitzStein: Confirmed the gap — `mirror_channel_send_to_session` (`crates/librefang-runtime/src/tool_runner/channel.rs:76`) does derive with a bare `for_sender_scope`, no guard, and `librefang-runtime` does depend on `librefang-channels` without importing `sanitize_channel_name` in that file.

Narrowed the doc-comment as suggested: "single source of truth for the kernel-side dispatch and reset paths" instead of the unqualified claim, and named the gap explicitly.

Didn't fix `channel.rs:76` itself — that's `librefang-runtime`, outside this PR's crate. It now has its own issue, #8243.

DaBlitzStein:
Every one of the 12 review threads on this PR now has an inline reply, and the branch is at `0bfdc530b`.

houko (approved):
LGTM. Both blocking findings are fixed, and the second one is fixed in the way that actually matters.

**The test that could not fail.** `reset_channel_session_scopes_a_reserved_name_away_from_the_internal_session` now compares `external` against the *unguarded* derivation:

```rust
let unguarded = SessionId::for_sender_scope(assistant, "cron", Some("chat-7"));
assert_ne!(external, unguarded, "test premise: the reserved-name guard must rewrite 'cron' to 'ext-cron' …");
```

That is the discriminating premise. The old `assert_ne!(external, for_channel(assistant, "cron"))` passed with the guard removed, because `for_sender_scope` composes `"cron:chat-7"` before hashing and is a different id from `for_channel`'s `"cron"`-only formula either way. Deleting `resolve_scope_channel` from `channel_session_id` now fails this test on its first assertion, which is what it is for. The premise comment says so explicitly, so the next person cannot quietly weaken it back.

**The input class the earlier tests missed.** Covered — `"cron"` with an external (`is_internal_system = false`) caller, seeded on both sides, asserting the internal `for_channel(assistant, "cron")` session keeps its message while the `ext-cron` one is emptied. That is the only input class whose resolved id this diff changes, and it now has a test.

**The third point** (`mirror_channel_send_to_session` deriving outside the single source of truth) I raised as non-blocking, and it is resolved the right way for a non-blocking one: the doc-comment no longer claims to be the source of truth for every derivation, it names the one production site that is still outside it, and it records that the guard is reachable from `librefang-runtime` via `librefang_channels::types::sanitize_channel_name` — so whoever migrates it does not have to rediscover that the circular-dependency objection applies to the kernel constants and not to the channels crate. That is the same conclusion #7995 reached from the other direction, which is a good sign.

All checks green; `BLOCKED` is the stale `CHANGES_REQUESTED` this approval clears.

Process note: #8228 landed a large rewrite of `crates/librefang-channels/src/bridge.rs` after this PR's CI ran, and `crates/librefang-kernel/src/kernel/messaging.rs` is also touched by the still-open #8213. Git reports mergeable, but the `main` ruleset has no strict status-check policy, so I am updating the branch to re-run CI against the current base rather than merging on a green that predates #8228.

PR #7708, opened 2026-08-18
fix(deps): bump h2 to 0.4.16 for RUSTSEC-2026-0258

houko:
## What

`cargo update -p h2`, taking the lockfile from 0.4.13 to 0.4.16 and closing RUSTSEC-2026-0258.

## Why now

The advisory ["h2 unbounded empty DATA frames"](https://rustsec.org/advisories/RUSTSEC-2026-0258) was published **2026-08-17**, against exactly the `h2 0.4.13` that `Cargo.lock` pins. `cargo audit` classifies it as a vulnerability rather than an unsound/unmaintained warning, so unlike the 17 warnings the lane already tolerates it drives the exit code:

```
Crate:     h2
Version:   0.4.13
Title:     h2 unbounded empty DATA frames
Date:      2026-08-17
ID:        RUSTSEC-2026-0258
URL:       https://rustsec.org/advisories/RUSTSEC-2026-0258
Solution:  Upgrade to >=0.4.16
...
error: 1 vulnerability found!
warning: 17 allowed warnings found
```

The Security lane is therefore red on any PR whose CI ran after the advisory landed, no matter what that PR touches — I hit it on an unrelated `librefang-channels` change (#7705) that adds no dependency and does not touch `Cargo.lock`. Because `Security` feeds `CI Gate`, the gate goes red too, so this blocks merges repo-wide rather than being a background nag.

Dependabot's open #7703 does not cover it — that group bumps `futures*` 0.3.33 → 0.3.34.

## Scope

`h2` is not declared by anything in the workspace; it arrives transitively via hyper/reqwest. The advisory's own remediation is `>=0.4.16`, which is semver-compatible, so no manifest change is needed and the lockfile diff is one package's version and checksum:

```
-version = "0.4.13"
+version = "0.4.16"
-checksum = "2f44da3a8150a6703ed5d34e164b875fd14c2cdab9af1252a9a1020bde2bdc54"
+checksum = "a9f37a958b41b3b19ee2707c06439c0e9e547e847223eb791ecb0cb821c65e27"
```

`cargo update -p h2` reported `Locking 1 package to latest compatible version` — nothing else moved.

## Verification

```
cargo check --workspace --lib    exit 0
```

`cargo audit` itself is CI's to confirm; cargo-audit is not installed on this machine, and the Security lane installs it per-run.

## Note on conflicts

This touches `Cargo.lock`, which many open PRs also touch. It is small and mechanical, so re-running `cargo update -p h2` on top of whatever lands first reproduces it exactly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -214,7 +214,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -225,7 +225,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1206,7 +1206,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -2131,7 +2131,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2486,7 +2486,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -3355,9 +3355,9 @@ dependencies = [
 
 [[package]]
 name = "h2"
-version = "0.4.13"
+version = "0.4.16"
```
houko, line 3358: The h2 bump itself looks right, but this diff isn't actually a one-package change.
Comparing against current `origin/main` (`fabd9cf`), this lockfile also re-resolves ~16 dependency edges from `windows-sys 0.61.2` / `windows-link 0.2.1` down to older already-present versions (`0.60.2`, `0.59.0`, `0.52.0`, `0.45.0`, `windows-link 0.1.3`) — e.g. `clap`, `dirs-sys`, `is-terminal`, `same-file`, `parking_lot`-adjacent crates, etc.
None of those package blocks were added or removed, so this isn't new dependencies — it's `cargo update -p h2` silently flipping which already-locked version several crates resolve to.

That contradicts the PR description ("`cargo update -p h2` reported 'Locking 1 package to latest compatible version' — nothing else moved") and is the classic symptom of running the update against a stale local `Cargo.lock`/registry index that predates whatever commit on `main` last bumped those `windows-sys` edges to `0.61.2`, per the "Re-created worktree" staleness warning in CLAUDE.md.

Could you rebase onto current `main` and re-run `cargo update -p h2` there? The resulting lockfile diff should touch only the `h2` package block.

---

houko (commented):
Mechanical fix pushed: added the missing `(#7708)` PR reference to the changelog fragment's attribution line (CLAUDE.md changelog convention — not CI-enforced, but avoids a duplicate generated release-note line).

One substantive finding left as an inline comment: the lockfile diff isn't limited to `h2` as described — it also silently re-resolves a batch of `windows-sys`/`windows-link` edges to older versions, which looks like it was generated against a stale base. Flagging for the author to confirm/rebase rather than fixing myself, since a lockfile bump the wrong direction shouldn't be pushed without the author validating a fresh regeneration.

---

Issue #7714, opened 2026-08-18
Workflow runs: owner attribution and per-step fresh agent instances


DaBlitzStein:
## Motivation

Two gaps on workflow step agents:

1. When a step spawns an agent from a template, the spawned agent bills to itself — an ephemeral worker pattern exists (#6930) but template-spawned step agents are not parented, so spend is mis-attributed to the wrong agent and the spawner loses budget visibility.
2. There is no way to request a fresh instance per run: find-or-spawn always reuses the canonical instance, which carries cross-run history.

## Proposal

- Persist `owner_agent_id` on workflow runs (migration v48) and thread the caller agent (workflow_run / workflow_start tools, channel commands) into run creation; resume / operator-action paths copy the original run's owner.
- Spawned step agents record the owner as parent and bill to it (`billed_agent = parent.unwrap_or(agent_id)` in the usage record).
- `{ type = "x", fresh = true }` spawns a new random-id instance with a unique name tag per run (the registry is name-unique); the default keeps deterministic find-or-spawn.

## Acceptance

Kernel tests: owner parent wiring, ownerless runs, fresh-vs-reuse ids; migration v48 idempotence; row round-trip with owner.

houko:
Status: tracked by #7717, which is not yet mergeable.

#7717 is blocked by its parent #7715, not by its own code.
Four of its five commits are #7715's; its own contribution is the single commit `4cf1b8fa`.
Reviewing it today means reading several hundred lines that belong to another PR, so it should stay a draft until #7715 lands and it is rebased.
After that its substance is confined to `agent_execution.rs`.

houko:
Closing this as part of a backlog sweep. Adjacent work landed during the sweep, but none of this issue's three asks is on `main`, so `completed` would be a lie about a tracker entry that someone will later rely on.

## What landed, and why it is not this issue

- **#7858** (`e2730c439`) — step agent types with find-or-spawn, `crates/librefang-kernel/src/kernel/step_agent.rs` and `crates/librefang-kernel/src/agent_template.rs`. That is #7712, the parent this issue builds on. It deliberately resolves a type to the canonical name-derived UUID (#4614) so the type keeps one identity and one session history, which is precisely the behaviour `fresh = true` is meant to opt out of.
[...]

PR #7715, opened 2026-08-18
feat(workflows): step agent types with find-or-spawn

### LINKED ISSUE #7712: Workflow steps: reference an agent type with find-or-spawn
## Motivation

Workflow authors must pre-register agents by exact name before a workflow can use them. Templates (templates/ and workspaces/agents/) are never consulted, so a workflow cannot express "use the researcher agent type" — it breaks when the agent is missing or renamed.

## Proposal

Allow workflow steps to reference an agent type: `{ type = "researcher" }`. Resolution is find-or-spawn:

- reuse the registered agent with that name if one exists,
- otherwise load the template manifest (templates/ then workspaces/agents/) and spawn it top-level with the canonical name-derived UUID (race-safe via agent_identities, #4614),
- missing template fails the step with a ByType-specific error naming the type.

All resolver sites (kernel sync/async, channel bridge, API background) get the arm via a KernelApi trait method.

## Acceptance

- Spawn, reuse, and missing-template paths covered by kernel regression tests.
- StepAgent object form accepts exactly one of id / name / type (unchanged contract, plus type).

DaBlitzStein:
Closes #7712

## Changes
- `StepAgent::ByType { template }` — a workflow step can reference an agent type with `{ type = "researcher" }` (object form, alongside id/name).
- Find-or-spawn resolution: reuse the registered agent with that name; otherwise load the template manifest (templates/ then workspaces/agents/) and spawn it top-level with the canonical name-derived UUID, race-safe via `agent_identities` (#4614).
- Missing template fails the step with a ByType-specific error naming the type and the step.
- All resolver sites (kernel sync/async, channel bridge, API background) get the arm via a `KernelApi` trait method.

## Verification
- `cargo test -p librefang-kernel workflow_step_by_type` — 3/3 (spawn, reuse, missing-template)
- `cargo test -p librefang-memory migration` — 39/39
- `cargo clippy -p librefang-kernel -p librefang-api -p librefang-memory -p librefang-runtime --all-targets -- -D warnings` — clean

### DIFF
--- a/changelog.d/added/7712-workflow-step-agent-types.md
+++ b/changelog.d/added/7712-workflow-step-agent-types.md
@@ -1,0 +1,2 @@
+Workflow steps can reference an agent type with `{ type = "name" }` — find-or-spawn semantics: reuse the registered agent with that name, else spawn from the template manifest (templates/ then workspaces/agents/) with the canonical name-derived UUID, and fail the step with a ByType-specific error when neither exists.
+ (@DaBlitzStein)
--- a/crates/librefang-api/src/channel_bridge.rs
+++ b/crates/librefang-api/src/channel_bridge.rs
@@ -1269,6 +1269,9 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
                         let inherit = entry.manifest.inherit_parent_context;
                         Some((entry.id, entry.name.clone(), inherit))
                     }
+                    StepAgent::ByType { template } => {
+                        kernel.resolve_agent_by_type_or_spawn(template)
+                    }
                 },
                 |agent_id, message, session_mode_override| {
                     let k = kernel.clone();
--- a/crates/librefang-api/src/routes/workflows/mod.rs
+++ b/crates/librefang-api/src/routes/workflows/mod.rs
@@ -193,6 +193,7 @@ fn workflow_to_json(w: &Workflow) -> serde_json::Value {
                 "agent": match &s.agent {
                     StepAgent::ById { id } => serde_json::json!({"agent_id": id}),
                     StepAgent::ByName { name } => serde_json::json!({"agent_name": name}),
+                    StepAgent::ByType { template } => serde_json::json!({"type": template}),
                 },
                 "prompt_template": s.prompt_template,
                 "mode": serde_json::to_value(&s.mode).unwrap_or_default(),
--- a/crates/librefang-api/src/routes/workflows/workflow.rs
+++ b/crates/librefang-api/src/routes/workflows/workflow.rs
@@ -37,9 +37,13 @@ pub async fn create_workflow(
             StepAgent::ByName {
                 name: name.to_string(),
             }
+        } else if let Some(t) = s["type"].as_str() {
+            StepAgent::ByType {
+                template: t.to_string(),
+            }
         } else {
             return ApiErrorResponse::bad_request(format!(
-                "Step '{}' needs 'agent_id' or 'agent_name'",
+                "Step '{}' needs 'agent_id', 'agent_name', or 'type'",
                 step_name
             ))
             .into_json_tuple();
@@ -330,9 +334,13 @@ pub async fn update_workflow(
                 StepAgent::ByName {
                     name: aname.to_string(),
                 }
+            } else if let Some(t) = s["type"].as_str() {
+                StepAgent::ByType {
+                    template: t.to_string(),
+                }
             } else {
                 return ApiErrorResponse::bad_request(format!(
-                    "Step '{}' needs 'agent_id' or 'agent_name'",
+                    "Step '{}' needs 'agent_id', 'agent_name', or 'type'",
                     step_name
                 ))
                 .into_json_tuple();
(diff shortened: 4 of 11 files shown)


houko (changes requested):
Blocking findings despite the green CI:

- The HTTP create/update parsers select `agent_id`, then `agent_name`, then `type` with an `if/else` chain. Payloads containing two or three routing keys are silently accepted, while the acceptance contract requires exactly one of id/name/type. Count and validate the keys before constructing `StepAgent`, and add API integration tests for every ambiguous pair.
- This PR also adds a complete `workflow_create` definition to `builtin_tool_definitions`, but adds no dispatcher or handler. Agents can discover and call the advertised tool only to receive `ToolError::NotFound`. Remove this unrelated stacked change from #7715; it belongs with #6943 and must land atomically with its implementation.
- `load_agent_manifest_from_template_dirs` collapses an existing unreadable/invalid template into `None`, so the workflow reports "no template file" instead of the real parse/read failure. It also does not enforce that the manifest name matches the requested type, and the spawn fallback reuses any registry entry matching that different manifest name after any spawn error. Return a structured error, pin or validate the template identity, and only treat an actual duplicate race as reusable.

Please add regressions for corrupt templates, mismatched filename/manifest names, and a non-duplicate spawn failure with an existing same-name registry entry.

houko:
**Land this one next after #6943 — it is the root of your workflow stack and the only PR here whose CI is fully green.**

Run [32187965415](https://github.com/librefang/librefang/actions/runs/32187965415) on `c12a3e2` is green across the board, including Quality, Test / Unit and OpenAPI Drift. Nothing mechanical is in your way; the only blocker is the 2026-08-22 review. Concretely:

**1. Exactly-one-of, and you already wrote the rule.** Your own `Deserialize for StepAgent` at `crates/librefang-kernel/src/workflow.rs:359-372` does it correctly:

```rust
let present = id.is_some() as u8 + name.is_some() as u8 + template.is_some() as u8;
if present != 1 { return Err(D::Error::custom(...)); }
```

The HTTP handlers never reach that impl — they hand-build `StepAgent` from an `if/else` chain over the flat step object:

- `crates/librefang-api/src/routes/workflows/workflow.rs:34-48` (`create_workflow`)
- `crates/librefang-api/src/routes/workflows/workflow.rs:329-345` (`update_workflow`)

`{"agent_id": "...", "type": "researcher"}` is silently accepted as `ById` at both. The keys differ (`agent_id`/`agent_name`/`type` flat, vs `id`/`name`/`type` nested) so you can't just call `serde_json::from_value::<StepAgent>`, but you can mirror the same count-then-reject shape before constructing the variant, and add one integration test per ambiguous pair (`agent_id`+`agent_name`, `agent_id`+`type`, `agent_name`+`type`, all three).

**2. Drop the `definitions.rs` hunk — it is a live bug in this PR, not just scope creep.** You add `tool_name::WORKFLOW_CREATE` and a full `ToolDefinition` for it to `builtin_tool_definitions`, but this PR does not touch `crates/librefang-runtime/src/tool_runner/dispatch.rs`, and `grep -rn workflow_create crates/librefang-runtime/src/tool_runner/` on `main` returns nothing. So if #7715 merges alone, every agent is advertised `workflow_create` and every call falls through to the `other =>` arm and comes back `ToolError::NotFound` — "Error: Tool 'workflow_create' not found". The tool schema and the handler are both in #6943 and have to land together with it. Removing the hunk also removes your only textual collision with #6943.

**3. Template loading.** `load_agent_manifest_from_template_dirs` collapses an unreadable or malformed template into `None`, so a corrupt file reports as "no template file". Return a structured error that distinguishes missing from malformed, pin the manifest name against the requested type, and narrow the post-spawn-error registry reuse to an actual duplicate race. Regressions wanted for: corrupt template, filename/manifest name mismatch, and a non-duplicate spawn failure with a same-name registry entry already present.

## Why this PR is worth your attention before the others

#7717 re-commits three of your four commits here (`step agent types with find-or-spawn`, `reuse the winner on a concurrent find-or-spawn race`, `dry runs must not spawn step agents`) as rewritten copies. That is 490 of its 819 added lines, and 10 of its 21 changed files are files this PR already changes. Neither CI nor review over there is exercising #7717's own change. Once #7715 lands, #7717 rebases down to a single commit (`4cf1b8fa1`, 328+/68-) and becomes reviewable on its merits.

One correction to how this stack is being described: #7726, #7727 and #7728 do **not** descend from #7715 — `compare` against each head shows all four of your commits here missing (`behind_by: 4`) and ~197-199 commits of a different borrowed base ahead, including #6038, #6043, the ephemeral `agent_spawn` work, template CRUD and the Telegram command fixes. Landing #7715 will not unblock those three; they each need rebasing onto `main` separately.

DaBlitzStein:
Superseded by #7858, which landed the same behaviour.

Verified against `origin/main`: `Kernel::find_or_spawn_agent_type` at `crates/librefang-kernel/src/kernel/step_agent.rs:179`, dispatched from the `StepAgent::ByType` arm at line 107, with coverage in `kernel/tests.rs`.
Nothing in this branch is missing upstream, so keeping it open only adds a conflicting diff to rebase.

The one piece #7858 deliberately did not take is the `fresh` flag; that discussion belongs on #7714, where the rationale is recorded, and not in a stale branch.

PR #7716, opened 2026-08-18
feat(templates): surface declared-but-unavailable skills and MCP servers

### LINKED ISSUE #7713: Template-spawned agents: surface declared-but-unavailable skills and MCP servers
## Motivation

A template manifest can declare skills / MCP servers that are not installed or not connected on this instance. The declaration is retained verbatim, but the gap is silent: no log, no API field, no dashboard hint — the operator sees nothing until a step behaves differently than expected.

## Proposal

Compute the pending set (declared skill names absent from the registry, declared MCP servers absent from the effective server list — allowlist mode only) and expose it:

- WARN at spawn naming the pending entries,
- KernelApi method `pending_skill_and_mcp_declarations`,
- agents API surfaces `pending_skills` / `pending_mcp_servers`,
- dashboard skills tab badges them.

Installing the skill / connecting the server activates it on the next reload without re-spawn.

## Acceptance

Kernel tests: pending surfaces before availability, clears after the server is configured (without respawn); API integration test for the new fields.

DaBlitzStein:
Closes #7713

## Changes
- Kernel computes the pending set for template-spawned agents: declared skill names absent from the loaded registry, declared MCP servers absent from the effective server list (allowlist mode only — mode semantics mirror `available_tools` exactly).
- WARN at spawn naming the pending entries; declarations stay retained verbatim.
- `KernelApi::pending_skill_and_mcp_declarations`; agents API surfaces `pending_skills` / `pending_mcp_servers`; dashboard skills tab badges them.
- Installing the skill / connecting the server activates it on the next reload without re-spawn (verified by tests).

## Verification
- `cargo test -p librefang-kernel pending_` — 33/33
- `cargo clippy -p librefang-kernel -p librefang-api --all-targets -- -D warnings` — clean


houko (changes requested):
Blocking findings:

- `pending_skill_and_mcp_declarations` uses `effective_mcp_servers` as the availability set. That is the configured-server snapshot: `connect_mcp_servers` iterates it and leaves failed/no-transport servers in place. A declared server whose connection fails is therefore reported as active even though it contributes no tools, recreating the silent gap this issue is meant to expose. Base pending state on live connections (or an equivalent connection-health snapshot), and add a regression where a configured server fails to connect and remains pending.
- The dashboard only consumes `skillsData.pending`. It never reads or renders the new `pending_mcp_servers` / MCP endpoint `pending` field, so the MCP half of the promised dashboard surface is missing. The API integration test likewise covers only skills. Add the MCP badge and API/UI coverage for pending-to-connected transition.
- `AgentSkillItem` calls `t("agents.detail.skill_pending", ...)`, but no locale file defines that key. The locale coverage gate treats literal fallbacks as drift, and the current Dashboard Build is failing. Add the key to every supported locale and run the full dashboard test/build gate.

Please also rebase the focused change before re-running CI; the current branch's CI Gate is failing across several checks.

houko:
Good news first: **this PR is not part of the workflow stack and is not stacked on anything.** `compare` puts it 0 commits behind `main` at `e83a9f7`, and it shares only two files with #6943/#7715/#7717 (`kernel_api.rs`, `kernel/spawn.rs`). Two real commits plus seven `auto-update-branches` merges. Please ignore the "rebase the focused change before re-running CI" line in my 2026-08-22 review — that is stale, the branch is current.

**One red job, one missing key.** [build](https://github.com/librefang/librefang/actions/runs/32639364728/job/97193957304) is the only failure; every other lane including `Test / Unit`, `Quality` and `OpenAPI Drift` passes.

```
FAIL src/lib/__tests__/en-locale-coverage.test.ts
  Dashboard locale coverage / defines every literal i18n key used by dashboard source
AssertionError: Dashboard source references i18n keys that are missing from src/locales/en.json.
+   "agents.detail.skill_pending (components/AgentSkillItem.tsx:65)",
```

Two tests in that file fail on the same key (`en-locale-coverage.test.ts:576` and `:602`). `AgentSkillItem.tsx:65` is your `t("agents.detail.skill_pending", { defaultValue: "Pending activation — available on the next skills reload" })`; the sibling `agents.detail.skill_meta` on the next line is defined, `skill_pending` is not.

Worth knowing before you fix it: **adding it to `en.json` alone trades one red test for four.** `src/lib/__tests__/locale-parity.test.ts` is green right now only because the key is absent everywhere; its `it.each(others)` case asserts `missing: []` against every non-reference locale file, so a key in `en.json` and nowhere else fails for `ko.json`, `pl.json`, `uk.json` and `zh.json`. Add `skill_pending` next to `skill_meta` under `agents.detail` in all five files in one go.

## The two review items, with the specifics

**`effective_mcp_servers` is the configured snapshot, not connection health.** `crates/librefang-kernel/src/kernel/mcp_setup.rs:368` writes `*effective = new_configs` and bumps `mcp_generation`, and the connect loop starts *after* that, at step 5. A connect failure logs `"Failed to connect MCP server"` at `:445` and never removes the entry. So a declared server that fails to connect reads as available and never appears in `pending_mcp_servers` — the exact silent gap #7713 exists to expose. Your own doc comment on `pending_skill_and_mcp_declarations` says "the connected MCP servers" while the body reads the configured list.

The live set is `self.mcp.mcp_connections` (`crates/librefang-kernel/src/kernel/subsystems/mcp.rs:45`), and `reload_mcp_servers` already reads it that way at `mcp_setup.rs:352-359`. It is a `tokio::sync::Mutex`, so `pending_skill_and_mcp_declarations` has to become `async`. That is cheap — all three call sites are already `async fn` axum handlers:

- `crates/librefang-api/src/routes/agents/config.rs:373` (`get_agent_skills`)
- `crates/librefang-api/src/routes/agents/config.rs:498` (`get_agent_mcp_servers`)
- `crates/librefang-api/src/routes/agents/lifecycle.rs:1012`

Regression: configured server, transport present, connection fails, assert it stays in `pending_mcp_servers`.

**The MCP half has no consumer at all.** You emit `"pending": pending.mcp_servers` from `get_agent_mcp_servers` (`config.rs:498-505`) and `pending_mcp_servers` from the detail route, but `dashboard/src/api.ts` only gained `pending?: string[]` on `AgentSkillsResponse` — there is no client function and no response type for `GET /api/agents/{id}/mcp_servers` anywhere in `api.ts`, so that field is unreachable from the dashboard rather than merely unrendered. `AgentsPage.tsx` reads `skillsData?.pending` and nothing else. The integration test asserts `body["pending_mcp_servers"]` equals `[]`, i.e. only the empty case. Needed: the client plus type, the badge, and a test where a declared server is genuinely pending and then transitions to connected.

Fix the locale key first — it is a five-file edit and gets you a green board to iterate against.

DaBlitzStein:
Superseded by #7853, with the allowlist half in #7898.

Verified against `origin/main`: `pending_mcp_servers` is emitted from `crates/librefang-api/src/routes/agents/lifecycle.rs:1079` and asserted in three cases in `tests/agents_routes_integration.rs`.
Every production symbol this branch introduced is present upstream; only the test names differ.

PR #7717, opened 2026-08-18
feat(workflows): run owner attribution and per-step fresh instances

### LINKED ISSUE #7714: Workflow runs: owner attribution and per-step fresh agent instances

DaBlitzStein:
Closes #7714

> ⚠️ STACKED on #7715 (step agent types) — this branch carries its commit; it will be rebased on main once #7715 merges.

## Changes
- `WorkflowRun.owner_agent_id` persisted via migration v48; the caller of `workflow_run` / `workflow_start` / channel workflow commands threads its agent id through the trait surface; resume and operator-action paths copy the original run's owner.
- Template-spawned step agents record the owner as parent and bill to it (`billed_agent = parent.unwrap_or(agent_id)` in the usage record).
- `{ type = "x", fresh = true }` spawns a new random-id instance per run with a unique name tag (the registry is name-unique); the default keeps deterministic find-or-spawn.

## Verification
- `cargo test -p librefang-kernel workflow_step_by_type` — 5/5 (spawn, reuse, missing, owner parent, fresh)
- `cargo test -p librefang-memory migration` — 39/39 (incl. v48 idempotence)
- `cargo clippy -p librefang-kernel -p librefang-api -p librefang-memory -p librefang-channels -p librefang-runtime --all-targets -- -D warnings` — clean

### DIFF
--- a/changelog.d/added/7714-workflow-run-owner-fresh.md
+++ b/changelog.d/added/7714-workflow-run-owner-fresh.md
@@ -1,0 +1,1 @@
+Workflow runs record their owner and bill template-spawned step agents to it; steps can request a fresh agent instance per run with `{ type = "x", fresh = true }`, which spawns a new random-id instance with a unique name tag instead of reusing the canonical one. (@DaBlitzStein)
--- a/crates/librefang-api/src/channel_bridge.rs
+++ b/crates/librefang-api/src/channel_bridge.rs
@@ -1269,6 +1269,9 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
                         let inherit = entry.manifest.inherit_parent_context;
                         Some((entry.id, entry.name.clone(), inherit))
                     }
+                    StepAgent::ByType { template } => {
+                        kernel.resolve_agent_by_type_or_spawn(template)
+                    }
                 },
                 |agent_id, message, session_mode_override| {
                     let k = kernel.clone();
--- a/crates/librefang-api/src/channel_bridge.rs
+++ b/crates/librefang-api/src/channel_bridge.rs
@@ -1233,7 +1233,7 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
         msg
     }
 
-    async fn run_workflow_text(&self, name: &str, input: &str) -> String {
+    async fn run_workflow_text(&self, name: &str, input: &str, owner: Option<AgentId>) -> String {
         let workflows = self.kernel.workflow_engine().list_workflows().await;
         let wf = match workflows.iter().find(|w| w.name.eq_ignore_ascii_case(name)) {
             Some(w) => w.clone(),
@@ -1243,7 +1243,7 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
         let run_id = match self
             .kernel
             .workflow_engine()
-            .create_run(wf.id, input.to_string())
+            .create_run_with_owner(wf.id, input.to_string(), owner)
             .await
         {
             Some(id) => id,
@@ -1269,8 +1269,8 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
                         let inherit = entry.manifest.inherit_parent_context;
                         Some((entry.id, entry.name.clone(), inherit))
                     }
-                    StepAgent::ByType { template } => {
-                        kernel.resolve_agent_by_type_or_spawn(template)
+                    StepAgent::ByType { template, fresh } => {
+                        kernel.resolve_agent_by_type_or_spawn(template, owner, *fresh)
                     }
                 },
                 |agent_id, message, session_mode_override| {
--- a/crates/librefang-api/src/routes/workflows/mod.rs
+++ b/crates/librefang-api/src/routes/workflows/mod.rs
@@ -193,7 +193,10 @@ fn workflow_to_json(w: &Workflow) -> serde_json::Value {
                 "agent": match &s.agent {
                     StepAgent::ById { id } => serde_json::json!({"agent_id": id}),
                     StepAgent::ByName { name } => serde_json::json!({"agent_name": name}),
-                    StepAgent::ByType { template } => serde_json::json!({"type": template}),
+                    StepAgent::ByType { template, fresh } => serde_json::json!({
+                        "type": template,
+                        "fresh": fresh,
+                    }),
                 },
                 "prompt_template": s.prompt_template,
                 "mode": serde_json::to_value(&s.mode).unwrap_or_default(),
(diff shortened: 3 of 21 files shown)


DaBlitzStein:
Status note, for visibility rather than action: one follow-up is stacked on this branch and cannot be sent upstream separately, because it is the next rung of the same migration ladder.

Waiting on this PR:

- Persisting `workflow_runs.total_steps` (#6504), so a run recovered after a daemon restart stops reporting "step X of 0" — `row_to_workflow_run` currently hardcodes `total_steps: 0` on reload. That is migration v49, and it sits directly on v48 (`owner_agent_id`) from this branch. `origin/main`'s ladder ends at `run_step!(47, migrate_v47)`, and `total_steps` has zero occurrences there, so the column cannot be added without v48 landing first.

houko (changes requested):
Two blocking correctness findings:

- Run ownership is contextual, but this implementation stores it permanently in the spawned agent's `entry.parent` and then bills every future call to that parent. A non-fresh type agent is globally reused: if owner A spawns it first, a later workflow owned by B (or an ownerless run) reuses the same entry and all usage is still charged to A. Pass the current run owner through the workflow dispatch/usage path instead of deriving billing from the reusable agent's structural parent. Add a regression with two owners reusing the same type and assert each run is charged to its own owner.
- The pre-call quota check at `agent_execution.rs:355-359` still checks the child `agent_id`, while the post-call record is changed to the parent. This lets a capped owner initiate spend through a child because the enforcement subject and accounting subject differ until after tokens have already been consumed. The same effective billed agent must be used for pre-call quota enforcement and post-call recording, without applying the workflow-specific rule to unrelated parented agents.

The issue also requires a persisted row round-trip with an owner. Current tests verify the v48 column exists and inspect an in-memory parent, but no test stores and reloads a non-None `owner_agent_id`. Please add that coverage, including resume after restoration.

houko:
**Do not spend time on this one until #7715 lands.** Most of what CI and review are looking at here is #7715's code, not yours.

Three of the five commits on `1b7c92c` are rewritten copies of #7715's commits — `1c6ad64b9 feat(workflows): step agent types with find-or-spawn`, `8c256ae65 fix(workflows): reuse the winner on a concurrent find-or-spawn race`, `1b7c92cb7 fix(workflows): dry runs must not spawn step agents`. That is 490 of the 819 added lines, and 10 of the 21 changed files are files #7715 already changes. `compare` confirms these are copies rather than the same commits: this head is `behind_by: 4` against #7715's head, so a rebase after #7715 merges collapses this PR down to `4cf1b8fa1` (328+/68-) plus the changelog fragment — which is the change actually under review.

CI is green on its own head ([32187969845](https://github.com/librefang/librefang/actions/runs/32187969845)), but that greenness is mostly attesting to #7715.

## The two blocking findings, with lines

**1. Quota subject and billing subject diverge.** `crates/librefang-kernel/src/kernel/agent_execution.rs:358` still enforces against the child:

```rust
.check_quota(agent_id, &entry.manifest.resources)
```

while `:1459` changes what gets recorded:

```rust
let billed_agent = entry.parent.unwrap_or(agent_id);
```

`main` uses a bare `agent_id` there (`agent_execution.rs:1455`), so this is the diff's own change. A capped owner can therefore initiate spend through an uncapped child, because enforcement and accounting only converge after the tokens are gone. Same effective subject on both sides.

The blast radius is wider than workflows, which is the part to be careful about: `entry.parent` is set by the `agent_spawn` tool via `kernel_handle::AgentControl::spawn_agent` at `crates/librefang-kernel/src/kernel/handles/agent_control.rs:57`, which passes whatever `parent_id` the caller supplied straight into `spawn_agent_with_parent`. So `entry.parent.unwrap_or(agent_id)` silently re-attributes every `agent_spawn`-created agent's spend to its spawner, workflow or not. That is a behaviour change for unrelated parented agents and needs to not happen.

**2. Ownership is contextual, so stop deriving it from structure.** A non-fresh type agent is globally reused. If owner A's run spawns it, a later run owned by B — or an ownerless cron/channel run — resolves the same registry entry, and `entry.parent` still points at A, so B's tokens bill to A. Thread the current run's owner through the workflow dispatch/usage path and pass it into the usage record for that call, instead of reading it back off the reused agent. Regression: two owners, one agent type, one run each, assert each run charges its own owner.

**3. Persistence round-trip.** Current coverage checks that the v48 column exists and inspects an in-memory parent. Nothing stores and reloads a non-`None` `owner_agent_id`. Add a store-then-reload assertion and a resume-after-restoration case.

None of that is blocked on #7715 conceptually — `agent_execution.rs` is yours alone here. But rebasing first means the diff a reviewer opens is 300 lines of your work instead of 800 lines where the interesting part is buried, and it removes the risk of the shared `workflow.rs` / `definitions.rs` / `handles/workflow_runner.rs` hunks drifting between the two branches.

Suggested order for the whole set: #6943 (one merge commit greens its CI, then two localized kernel fixes), #7715 (green already, three review items), then this, then #7716 independently at any point.

DaBlitzStein:
Closing: both halves of this branch are settled upstream, one landed and one decided against.

**Owner attribution landed.** `WorkflowRun::owner_agent_id` is in `crates/librefang-memory/src/workflow_store.rs` on `origin/main`, selected by `get_run` and `list_runs` and carried through both upsert paths. Worth noting for anyone who reads this branch's diff: upstream's upserts omit `owner_agent_id` from `ON CONFLICT DO UPDATE SET` entirely, so a later state write preserves the original owner by construction. This branch reached the same guarantee with a `COALESCE`, which was only needed because it *did* list the column in the update — upstream's shape needs no guard at all.

**Per-step fresh instances were declined, with a reason.** `StepAgent` on `origin/main` records it: no `fresh` variant, because `WorkflowStep::session_mode = New` isolates a step from the type's prior context without minting a second agent, workspace and registry entry per run, and `owner_agent_id` / `billed_agent_id` keep two owners' spend apart on a shared type.

That leaves one open question, which is a design question and not a diff: `session_mode = New` gives a fresh *session*, not a fresh *agent*, so a step still reads the standing instance's memory and workspace. If that distinction matters for a use case, it belongs on #7714 as an argument against the recorded rationale, not as a stale branch to rebase.

PR #6931, opened 2026-08-11
feat(templates): POST/PUT/DELETE on /api/templates + Agent Types dashboard

DaBlitzStein:
Part 2 of #6699 — template write API + dashboard UI. Core ephemeral spawn engine is in #6930.

Per [houko review](https://github.com/librefang/librefang/issues/6699#issuecomment-5178120055): extends the existing `/api/templates` surface instead of creating a new `/api/agent-types` namespace.

## Routes

| Method | Path | Description |
|--------|------|-------------|
| GET | /api/templates | Dual-source listing (workspaces agents + user templates) |
| POST | /api/templates | Create a template → `~/.librefang/templates/<name>.toml` |
| GET | /api/templates/{name} | Detail with `source` tag: `"agent"` or `"template"` |
| PUT | /api/templates/{name} | Update template |
| DELETE | /api/templates/{name} | Delete template |
| GET | /api/templates/{name}/toml | Raw TOML content (dual-source) |

## Dashboard

- Agent Types page (sidebar: between Agents and Chat)
- Card grid with create/edit/delete + Quick Run (spawns ephemeral worker)
- Query/mutation hooks with `agentTypeKeys` factory and invalidation
- i18n: en, uk, zh, ko

## Security

- `validate_template_name` (`[A-Za-z0-9_-]` only) on all path params
- TOML injection: uses `toml::to_string_pretty` on `AgentManifest` for proper escaping
- 409 on create-conflict, 404 on missing

## Integration tests

`agent_types_integration_test.rs` (8 tests): CRUD lifecycle, path traversal rejection, duplicate detection, TOML escaping round-trip

Refs: #6699

PR #7725, opened 2026-08-19
feat(agent-types): agent-facing agent_type_create tool + /api/agent-types alias

### LINKED ISSUE #7722: Agent-facing tool to create agent types (agent_type_create)
## Motivation

Agent types (templates) have CRUD via API (#6931) and a dashboard page, but an agent cannot create one from a conversation — the only agent-facing authoring tool is workflow_create (#6943). Operators can only pre-author types by hand.

## Proposal

- A native tool `agent_type_create` (name TBD: agent_type_create) writing a valid template manifest (name, system_prompt, model, tools, skills) to the templates dir through the same validation the HTTP route uses.
- Naming homogenization: templates of agents ARE agent types — UI/TUI/tool/docs use one term (agent type); API keeps /api/templates and /api/agent-types as aliases where both exist.
- Surfaces: tool definition + runtime handler + TUI templates screen label + dashboard labels + API alias.

## Acceptance

Runtime test (success + invalid name), kernel-side persistence test, parity across the four surfaces.

DaBlitzStein:
Closes #7722

> ⚠️ STACKED: carries the agent-types CRUD commits from #6931 (the templates router it extends is unmerged); will rebase on main once #6931 merges.

## Changes
- `agent_type_create` native tool: writes a validated template manifest (name/system_prompt/model/tools/skills) via a shared JSON→TOML conversion moved to librefang-types so API and tool can never drift.
- New `AgentTypeTools` kernel-handle trait + kernel impl (charset/length, duplicate, workspace-shadowing checks; tokio::fs writes).
- `/api/agent-types` aliases the template routes; TUI templates screen is now labeled "Agent types" (homogenization).

## Verification
- `cargo check --workspace --lib --tests` clean; `cargo test -p librefang-kernel create_agent_type_writes` 1/1; clippy -D warnings clean.


DaBlitzStein:
Status note, for visibility rather than action: this branch introduces the `/api/agent-types` alias, and the naming homogenization stacked behind it has nowhere to land until it does — `api/agent-types` and `agent_type_create` both have zero occurrences in `origin/main`.

Waiting on this PR:

- The template → agent-type rename across the remaining surfaces, which references the alias this branch adds: the dashboard's agent-type page and its query/mutation layer, and the handler names (`create_agent_type` / `update_agent_type` / `delete_agent_type`, with `/api/templates/*` kept as deprecated aliases).

houko (changes requested):
Blocking findings:

- `crates/librefang-kernel/src/kernel/handles/agent_types.rs` checks `path.exists()` and later calls `tokio::fs::write`. Two concurrent `agent_type_create` calls for the same name can both pass the check and silently overwrite each other, violating the unique-name contract. Open the destination atomically with `create_new(true)` and map `AlreadyExists` to the duplicate result. The HTTP create path has the same pre-existing race and should use the same shared persistence primitive if the stated goal is surface parity.
- The PR claims API and tool share exact validation, but only JSON-to-TOML conversion moved to `librefang-types`. Name validation and workspace-collision validation remain duplicated in the API handler and kernel implementation. Extract a shared validated input/persistence contract so a future rule cannot drift between surfaces.
- The issue requires runtime success and invalid-name tests, but the only new test directly invokes `AgentTypeTools` on the kernel. There is no test that dispatches `agent_type_create`, proves its definition is discoverable/authorized through normal tool resolution, or verifies malformed input becomes the expected `ToolError`. Please add the runtime coverage plus a concurrent-create regression.

The stacked branch currently has a failing CI Gate. Please rebase the focused change after #6931 lands so the required checks cover the actual #7722 diff.

houko:
Triaged at head `d9a56971c`. The single most useful thing I can tell you: this PR's own content is 15 files, +256/-60 (commit `2fbd89853`), but the PR presents 117 files and +9954/-3598. Everything else is borrowed stack, so review and every required check are spending their time on other people's code.

**Dependency order.** `2fbd89853` rewrites `create_agent_type` / `update_agent_type` in `crates/librefang-api/src/routes/agent_templates.rs` (+19/-54) — those handlers do not exist in `main`, they come from #6931 (`feat/agent-template-crud`, currently CONFLICTING/DIRTY). #6931 has to land first; this PR cannot be evaluated on its own until it does.

**Your own root failure** is the `Quality` job, and it is real — not stack noise. Making `AgentTypeTools` a supertrait of `KernelHandle` (`crates/librefang-kernel-handle/src/lib.rs`, +2) retroactively obligates every existing `KernelHandle` implementor, and the test stubs were never updated:

```
error[E0277]: the trait bound `tool_runner::tests::ApprovalKernel: librefang_kernel_handle::KernelHandle` is not satisfied
help: the trait `librefang_kernel_handle::AgentTypeTools` is not implemented for `tool_runner::tests::ApprovalKernel`
```

Same for `tool_runner::tests::DispatchCapture`, `tool_runner::tests::MirrorKernel`, `tool_runner::tests::ForceHumanCapturingKernel`, `tool_runner::tests::policy::SpawnCheckKernel`, `tool_runner::tests::workspace::NamedWsKernel`, `host_functions::tests::RecordingKernel`, and `workflow_runner::tests::StubKernel`. Cheaper fix than eight stub impls: drop the supertrait bound and take `&dyn AgentTypeTools` where you actually need it, the way the other capability traits are threaded. That keeps `KernelHandle` implementable by a three-line test double.

**Two more of the seven failures are yours and are quick.** `build` (dashboard `tsc`) is a duplicated generated block in `dashboard/src/api.ts` — `AgentTypeInput`, `listAgentTypes`, `getAgentType`, `createAgentType`, `updateAgentType`, `deleteAgentType` and `spawnEphemeral` are each declared twice, once at ~3745 and again at ~5429 (`TS2300`/`TS2323`/`TS2393`). Delete the second copy. `OpenAPI Drift` is only `openapi.json` and `xtask/baselines/openapi.sha256`, two lines each — regenerate with the three commands the job prints.

**On #7740 and the hoisted helper — one correction and one confirmation.**

Confirmed: `2fbd89853` is what creates `agent_type_json_to_toml` (`crates/librefang-types/src/agent.rs:991`, the +54 in `agent.rs` matching the -54 out of the route). It is not in `main` and not on #6931's head (`f828e7bed`), so the hoist is this PR's to own, and there are now three call sites: `agent_templates.rs:465` (POST), `agent_templates.rs:521` (PUT), and `handles/agent_types.rs:37` (the agent tool).

Correction: the #7740 data-loss defect does **not** reach the agent-facing tool. `create_agent_type` in `handles/agent_types.rs:40-44` returns early when the path exists, so it can never overwrite a file. The destructive write is `update_agent_type` at `agent_templates.rs:528` — `std::fs::write` over a document rebuilt from seven keys — and that is the PUT surface only. Fix #7740 there, or in #6931, not here.

What *does* reach your tool through the shared helper is silent lossy create: the helper reads only `name`, `description`, `system_prompt`, `provider`, `model`, `tools[]`, `skills[]` and fills the rest from `..Default::default()`, so an agent that passes `max_history_messages`, `[compaction]`, `mcp_servers`, `tool_allowlist` or `[[triggers]]` to `agent_type_create` gets them dropped with a success result. Worth either honouring them in the helper or rejecting unknown keys so the tool does not lie about what it persisted.

**A fix from your own stack got lost in this commit.** `dabc4df98` ("atomic create with create_new instead of check-then-write") replaced the TOCTOU in `agent_templates.rs` with `OpenOptions::create_new(true)` mapping `AlreadyExists` to 409. At this head there is no `create_new` in the file at all — `create_agent_type` is back to `path.exists()` (`:470`) then `std::fs::write` (`:492`). The rewrite reverted it. Please reapply, and apply it to the new kernel surface too, which has the same pattern at `handles/agent_types.rs:40` and `:61`.

Also still open from #7740's second half: `manifest_to_agent_type` (`agent_templates.rs:431-441`) emits no `channels` or `routing`, while the dashboard reads both.

Once #6931 lands, rebase onto `main` and this drops to a 15-file diff that CI can actually speak to. Right now a green run is unreachable for reasons that mostly are not yours.

DaBlitzStein:
Superseded by #7877, which landed the agent-facing `agent_type_create` tool.

Verified against `origin/main`: the tool is registered and dispatched in `crates/librefang-runtime/src/tool_runner/`, with route coverage in `crates/librefang-api/tests/agent_types_routes_integration.rs`.

PR #7726, opened 2026-08-19
feat(workflows): required skills per step with a precise mismatch error

### LINKED ISSUE #7721: Workflow steps: required skills with a clear mismatch error
## Motivation

A step can reference an agent type (find-or-spawn, #7712) or a named agent, but the workflow author cannot express "this step needs a specific skill". When the referenced agent or template lacks it, the step fails deep into execution with a generic tool error.

## Proposal

- `required_skills = ["browser-automation"]` on a step (workflow TOML/JSON + StepAgent resolution surfaces).
- At resolution time, check the resolved agent's skill allowlist (and, for type-based spawns, the template manifest's declared skills): if a required skill is missing, fail the step with an error naming the step, the skill, and the agent/template that lacks it.
- Declared-but-unavailable skills reuse the pending-declarations plumbing (#7713) so the error distinguishes "not declared" from "declared but not installed".
- Surfaces: workflow TOML/JSON schema, API step parser, dashboard step editor, TUI step authoring where present.

## Acceptance

Kernel tests: satisfied, missing-skill, and declared-but-unavailable paths; API integration test for the parser.

DaBlitzStein:
Closes #7721

> ⚠️ STACKED: builds on the step agent-type plumbing from #7715 (kernel_api resolver surface); rebases on main once #7715 merges.

## Changes
- `WorkflowStep.required_skills` (TOML/JSON + serde default). The engine evaluates a `skill_check` closure right after agent resolution in the sequential and DAG executors and fails the step before dispatch.
- One kernel implementation distinguishes "does not declare" (absent from the allowlist) from "declares but are not installed" (pending declarations, #7713); every resolver surface (kernel sync/async, operator driver, API background/resume/operator-action, channel bridge) threads it.
- `workflow_create`'s step schema documents the field; the API step parsers carry it through.

## Verification
- `cargo check --workspace --lib --tests` clean; `cargo test -p librefang-kernel workflow` 201/201 (incl. the pending-skill regression); `cargo test -p librefang-api workflow` 38/38; clippy -D warnings clean.


houko (changes requested):
Blocking findings:

- `check_step_required_skills` treats `skills = []` and `skills = ["*"]` as satisfying every requirement, then delegates availability to `pending_skill_and_mcp_declarations`. That helper intentionally returns no pending skills for unrestricted/wildcard mode, so a required skill that is not installed passes the gate. Check every required name against the loaded skill registry independently of allowlist mode; then separately distinguish undeclared from declared-but-unavailable.
- The acceptance contract requires the error to name the step, skill, and agent/template. The `skill_check` callback receives only `(AgentId, required)` and its errors say only "step requires..." without the actual `step.name`. Pass the step identity into the check or wrap its error at the executor call site for both sequential and DAG paths.
- The sole new kernel test is named as if it covers satisfied, missing, and pending cases, but it only exercises declared-but-unavailable. There is no API parser integration test, and neither dashboard nor TUI exposes `required_skills`. Add the three required kernel cases, including empty and wildcard manifests with an unavailable skill, validate malformed API arrays instead of silently filtering non-strings, and add the authoring surfaces promised by the issue.

The stacked branch currently has a failing CI Gate. Please rebase the focused change after #7715/#7713 land so checks run against the actual #7721 diff.

houko:
Triaged at head `ebee2d70e`. Your own content here is 11 files, +521/-103 (commit `ccc9be180`); the PR presents 118 files and +10470/-3696. The rest is borrowed stack.

**Dependency order — this one has two parents, not one.** `ccc9be180` inserts `check_step_required_skills` directly beneath `resolve_agent_by_type_or_spawn` in `crates/librefang-kernel/src/kernel_api.rs`, and that function comes from #7715 (`feat/workflow-step-agent-types`). It also calls `self.pending_skill_and_mcp_declarations(agent_id)`, which is neither in `main` nor on #7715's head — it comes from #7716 (`feat/templates: surface declared-but-unavailable skills and MCP servers`, at `tools_and_skills.rs:437` on head `e83a9f774`). So the order is **#7716 and #7715 both land, then this**. #7715 is already `MERGEABLE/BLOCKED`, so it is the nearer one.

**The review finding, confirmed against `tools_and_skills.rs` in `ccc9be180`.** The check is a no-op for the most common agent configuration:

```rust
let has_all = manifest.skills.is_empty() || manifest.skills.iter().any(|s| s == "*");
if !has_all {
    // ... only here is each `required` name compared against manifest.skills
}
let pending = self.pending_skill_and_mcp_declarations(agent_id);
let unavailable: Vec<&String> = required
    .iter()
    .filter(|r| pending.skills.iter().any(|p| p == *r))
    .collect();
```

When `manifest.skills` is `[]` or `["*"]`, `has_all` is true and the per-name declaration check is skipped entirely. The only remaining gate is membership in `pending.skills` — and by the doc comment already on `pending_skill_and_mcp_declarations` (three lines above your insertion), that function is "allowlist mode only (`["*"]` means 'all', so nothing is pending there), and disabled modes yield nothing". So in `*` mode `pending.skills` is empty, `unavailable` is empty, and the function returns `Ok(())`.

Net effect: a step declaring `required_skills = ["skill-that-does-not-exist"]` passes validation against any agent with `skills = []` or `skills = ["*"]`. Since an empty list is the default, the feature validates nothing for most agents — which is the opposite of what the changelog fragment promises ("fails before dispatch with a precise error").

**The structural cause is that the function never consults the skill registry.** It only compares `required` against `manifest.skills` (a declaration) and `pending.skills` (declared-but-missing). Existence is never established. The fix is two independent checks rather than one branch:

1. Resolve every name in `required` against the loaded registry — `read_kernel_state(&self.skills.skill_registry, "skill_registry")` then `SkillRegistry::skill_names()` (`crates/librefang-skills/src/registry.rs:517`), the same read pattern already used at `tools_and_skills.rs:241` — unconditionally, before and regardless of `has_all`. A name that resolves nowhere is "no such skill", a distinct third error from your current two.
2. Keep the allowlist check for "exists but this agent does not declare it", and the pending check for "declared but not installed".

That gives you the three-way distinction the error message is aiming at. As written, the two messages you do emit are both unreachable in the default configuration.

Minor, while you are in the function: `manifest.skills.is_empty()` meaning "every skill is permitted" is the right convention for `available_tools`, but it makes `required_skills` unenforceable by default, so the registry check in (1) is what actually carries the feature.

Once #7716 and #7715 land, rebase onto `main` — the diff becomes 11 files, and review and CI will finally be exercising your code instead of the stack's.

DaBlitzStein:
Superseded by #7863, which landed under the same title, plus #7871 for the canvas authoring side.

Verified against `origin/main`: `WorkflowStep::required_skills` at `crates/librefang-kernel/src/workflow.rs:300` and the `check_required_skills` gate at line 728, whose doc comment cites #7721 — the issue this branch was opened against.

PR #6930, opened 2026-08-11
feat(agent_spawn): ephemeral worker mode — budget-attributed, depth-guarded

DaBlitzStein:
Part 1 of #6699 — core engine only. Template CRUD and dashboard follow in a separate PR per [houko review](https://github.com/librefang/librefang/issues/6699#issuecomment-5178120055).

## What this does

`agent_spawn` gains an `ephemeral: true` mode. When set, the tool spawns a disposable worker — no workspace, no DB, no registry — runs the task with caller-configured tools, and returns the result directly via the tool response.

## Design

- Ephemeral workers get a **clean, empty context** — no parent history, no parent session, no inherited identity
- The caller controls everything: system prompt, model, tools, skills, max iterations
- Nothing inherited from the parent agent beyond the budget attribution and tools allowlist

## Files changed (10 files, +588/−7)

| Crate | Change |
|-------|--------|
| `librefang-types` | `EphemeralSpawnRequest` + `EphemeralSpawnResult` (caller-controlled config + metering) |
| `librefang-kernel-handle` | `spawn_ephemeral` / `spawn_ephemeral_detailed` on `AgentControl` trait |
| `librefang-kernel` | `spawn_ephemeral_worker` — resolves manifest (inline or from templates dir), runs agent_loop |
| `librefang-runtime` | Ephemeral branch in `tool_agent_spawn` with parent tools enforcement |
| `librefang-runtime` | Schema: `oneOf` (message for ephemeral OR name+system_prompt for permanent) |
| `librefang-api` | `POST /api/agents/spawn-ephemeral` handler + route registration |

## Safeguards (addresses houko feedback)

- **Budget attribution**: usage is billed against the parent agent_id so cost appears in `GET /api/budget/agents/{parent}` and counts against the parent agent budget
- **Recursion depth** (#6659): `current_agent_depth >= max_agent_call_depth` check BEFORE running the agent loop, plus the loop future is wrapped in `with_agent_call_depth` (which also boxes the ~56 KB future to prevent stack overflow)

## Explicitly NOT in this PR

- Agent types CRUD → separate PR (extending existing `/api/templates` surface, not new `/api/agent-types`)
- Dashboard Agent Types page → separate PR
- `~/.librefang/templates/` management → separate PR

## Verification

- `cargo check --workspace --lib` — passes
- No integration test file yet (separate API surface test will come with PR B)

Closes #6699

PR #7727, opened 2026-08-19
feat(ephemeral): agent-type missions with a uid name and transient workspace

### LINKED ISSUE #7723: Ephemeral agents from an agent type with a mission folder (create on spawn, delete on finish)
## Motivation

Ephemeral workers (#6930) run without a workspace, so a tool that needs scratch files (or an agent that wants to drop intermediate artifacts) has nowhere to write that is guaranteed to be cleaned up. The user-facing shape should be: spawn an ephemeral worker FROM an agent type, named like a uid (e.g. researcher-<8hex>), with its own mission folder created at spawn and deleted when the run ends.

## Proposal

- Extend the ephemeral spawn path: when `agent_type` is set, derive a uid display name and create a per-mission workspace dir under a transient area (e.g. data_dir/transient/<uid>); delete it on run completion (both success and error paths).
- The uid name keeps the registry name-unique invariant (AgentAlreadyExists) without leaking into find_by_name.
- Surfaces: tool schema (name/agent_type), API request/response (name + workspace_dir + a deletion guarantee), dashboard Quick Run reflects the uid, TUI spawn flow where applicable.

## Acceptance

Kernel test: folder exists during the run and is gone after; registry never registers the uid; API integration test for the response shape.

DaBlitzStein:
Closes #7723

> ⚠️ STACKED: builds on the ephemeral worker from #6930 (spawn_ephemeral_worker is unmerged upstream); rebases on main once #6930 merges.

## Changes
- Agent-type-based ephemeral workers derive a uid display name (`<type>-<8hex>`) and get a transient mission workspace under `~/.librefang/transient/<name>`, created at spawn and passed as the run's workspace root.
- A Drop guard removes the folder when the run ends — success and failure alike. Inline (no agent_type) spawns keep the historical no-workspace behaviour.

## Verification
- `cargo check --workspace --lib --tests` clean; `cargo test -p librefang-kernel ephemeral_mission_folder` 1/1 (create + cleanup on a failing run); clippy -D warnings clean.


houko (changes requested):
Three blocking findings in `crates/librefang-kernel/src/kernel/messaging.rs`:

- Lines 908-910 derive the filesystem path from the template manifest's unvalidated `name`. Only the requested `agent_type` filename is validated. A template containing an absolute name makes `PathBuf::join` discard `~/.librefang/transient`, so creation and the Drop cleanup happen outside the transient root. Derive the prefix from the already validated `agent_type` or sanitize and verify containment before any filesystem operation.
- Lines 911-916 swallow `ensure_workspace` failure, still assign the nonexistent/partial directory as `manifest.workspace`, and run the worker. That violates the advertised workspace guarantee and can produce later tool failures with a misleading root cause. Return the creation error and only arm cleanup after successful creation.
- The added test runs to completion before inspecting the filesystem, so it does not prove the folder existed during the run; it also leaks its `TempDir` with `mem::forget`. The issue acceptance criteria additionally require an API response-shape integration test, which is absent. Please add coverage for existence during execution, cleanup on success and error/cancellation, containment for hostile template names, creation failure, and the API contract.

This is also a large stacked branch with a failing CI Gate. Please rebase the focused change after #6930 lands so the required checks exercise the actual #7723 diff.

houko:
Triaged at head `1e3af8c16`. Leading with the security item, because it is the only thing here that must not merge as written.

**Path traversal into a recursive delete.** `crates/librefang-kernel/src/kernel/messaging.rs:906-909`:

```rust
let tag: String = uuid::Uuid::new_v4().simple().to_string()[..8].to_string();
let base = manifest.name.clone();
manifest.name = format!("{base}-{tag}");
let dir = self.home_dir_boot.join("transient").join(&manifest.name);
```

The path component is `manifest.name`, taken raw. The validation asymmetry is the bug: `resolve_ephemeral_manifest` (`messaging.rs:1143-1154`) carries an explicit guard, comment and all —

```rust
// Path-traversal guard: only [A-Za-z0-9_-] for <name>.toml
if type_name.is_empty() || type_name.len() > 64 || !type_name.chars().all(...)
```

— but that guards `request.agent_type`, which selects the template *filename*. `manifest.name` is a *field inside* the TOML that `toml::from_str` then parses (`messaging.rs:1164`, and again on the `workspaces/agents/<name>/agent.toml` fallback at `:1183`). The two are entirely decoupled. `templates/safe.toml` containing `name = "../../evil"` passes the `agent_type` guard on `"safe"` and yields `~/.librefang/transient/../../evil-a1b2c3d4`.

`ensure_workspace` does not close this — it is a bare `create_dir_all` loop over seven subdirs with no containment check (`crates/librefang-kernel/src/kernel/workspace_setup.rs:240-267`). And the escaped path is then handed to the `_MissionCleanup` `Drop` impl at `messaging.rs:919-920`, which calls `std::fs::remove_dir_all(dir)` at `:923`. So this is not just an out-of-bounds mkdir; it is an attacker-influenced recursive delete that fires on every run, success or failure.

**The fix is one call, and the helper is already in the file you are importing from.** `safe_path_component(input, fallback)` lives at `workspace_setup.rs:269` and filters to exactly `[A-Za-z0-9_-]` with a fallback for names that sanitize to empty. It is the established convention for turning a manifest `name` into a path component throughout this crate — `agent_state.rs:51`, `agent_state.rs:239`, `boot.rs:2084` all do `safe_path_component(&entry.name, &agent_id.to_string())`. You are already importing `ensure_workspace` from that module at `messaging.rs:910`, so it is in scope. The comment at `agent_state.rs:42-50` documents #6442, which was this same class of bug: names that sanitize to empty collapsing distinct agents onto a shared path. Please reuse the helper rather than re-deriving a guard, and pass `agent_id.to_string()` as the fallback so two missions cannot collide on one directory.

Belt and braces, given `remove_dir_all` is downstream: after joining, `canonicalize` the parent and assert it still starts with `home_dir_boot.join("transient")` before creating or deleting anything.

Deriving the prefix from the validated `agent_type` instead would also work and is arguably cleaner, since `agent_type` is `Some(_)` on exactly this branch — but the uid tag already guarantees uniqueness, so `safe_path_component` on the name is the smaller change.

**Dependency order.** `spawn_ephemeral_worker`, `resolve_ephemeral_manifest` and `EphemeralSpawnRequest` are none of them in `main`; they come from #6930 (`feat/ephemeral-spawn-core`, CONFLICTING/DIRTY). #6930 lands first. Note that the traversal is in *your* 79 lines, not #6930's, so it is worth fixing on this branch even while the parent is in flight.

**On the shape.** Commit `25dc0bee5` is 2 files, +80/-1 — `messaging.rs` plus a changelog fragment. The PR presents 119 files and +10549/-3696. Its seven failing checks are the stack's: the `Quality` break is #7725's `AgentTypeTools` supertrait leaving eight `KernelHandle` test stubs unimplemented, which you inherit through commit `2fbd89853`. Nothing in your own commit is implicated in a single one of them.

That is the whole problem in one number: 80 lines of yours are waiting on a 10,000-line diff. Rebase onto `main` once #6930 lands and this becomes a two-file review that can be read in a minute and merged the same day.

DaBlitzStein:
Superseded by #7860.

Verified against `origin/main`: `crates/librefang-kernel/src/kernel/mission_workspace.rs`, 495 lines, wired through `kernel/boot.rs` and `workspace_setup.rs`, with a changelog fragment naming #7723 — the issue this branch was opened against.

PR #7728, opened 2026-08-19
feat(dashboard): canvas step editor gains agent-type + fresh binding

### LINKED ISSUE #7724: Workflow step editors: expose agent-type (template) and fresh selection in dashboard and TUI
## Motivation

#7715/#7717 wired step agent types (find-or-spawn from a template, with per-run fresh instances) through the kernel, all resolver sites, and the API parser — but the visual step editors still only offer agent-by-name/id. The API-only surface violates the four-surface rule (API/TUI/UI/WebUI).

## Proposal

- Dashboard Canvas step editor: agent source selector name / id / type (template) + fresh checkbox.
- TUI workflows screen: expose the same fields wherever steps are authored; if the TUI screen is execution-only today, add the type/fresh display at minimum and authoring parity where the screen already authors steps.
- Tool schema (workflow_create) already documents type + fresh.

## Acceptance

Component test for the Canvas selector; TUI parity check; manual/CI UI test.

DaBlitzStein:
Closes #7724

> ⚠️ STACKED: the API-side type/fresh step parser and the agent-types SDK surface come from #7715/#6931; rebases on main once those merge.

## Changes
- Canvas step editor: agent-type (template) selector with a fresh-per-run checkbox alongside the existing id/name binding; canvas build/load mappings carry type/fresh end to end.
- Deduplicated the agent-types SDK block doubled by the #6931 merge in api.ts / client.ts and aligned the locale sets (parity tests green, zh comms drift fixed).
- The TUI workflows creator already accepts type/fresh/required_skills through its raw steps JSON (documented here rather than duplicated UI).

## Verification
- `pnpm exec tsc --noEmit` clean; vitest locale-parity 8/8; kernel type/fresh semantics covered by #7715/#7717 tests.


houko (changes requested):
Two blocking findings:

- `crates/librefang-api/dashboard/src/pages/CanvasPage.tsx:771`: once a node has an `agentType`, choosing a concrete agent cannot switch the binding back. The controlled value is forced to `""` while `agentType` is non-empty, and this select's `onChange` only sets `agentId`; `handleSave` still discards that ID and persists the old type. Clear `agentType` when a concrete agent is chosen, or model the three binding modes with one source selector as the issue proposes.
- The issue acceptance criteria explicitly require a Canvas selector component test, but this PR adds no Canvas test covering name/id/type switching, `fresh`, or save/load round trips. Please add regression coverage for the transition above and the serialized step fields.

The current branch also carries 122 files from several unrelated stacked changes and its CI Gate is failing. Please rebase the focused #7724 change after its dependencies merge so the review and required checks cover this issue's actual diff.

houko:
Reviewed against head `837546a`.
One mechanical problem has to be fixed before this is reviewable, and it is not in the commit you wrote.

## `CHANGELOG.md` loses 2,943 lines

`gh pr diff 7728 -- CHANGELOG.md` is a single hunk, `@@ -2962,2946 +2962,3 @@`, with 0 additions and 2,943 deletions — everything from `## [2026.5.17] - 2026-05-17` to the end of the file.
On `main` that file is 5,907 lines with 53 `## [` release sections; at your head it is 2,964 lines with 18.
Thirty-five released versions of history are gone.

The source is `f378bd9ba` ("Merge remote-tracking branch 'origin/main' into librefang_v2", 2026-08-19).
`compare/05ca7ff21...f378bd9ba` reports `CHANGELOG.md` as `additions: 0, deletions: 2943` — the conflict was resolved by keeping only the branch side.
That commit is an ancestor of your head and is not on `main` (`compare/f378bd9ba...main` → `diverged`), so the loss lives entirely in this branch.

This is clearly a conflict-resolution accident rather than intent: the changelog entry for this work is a correctly formed fragment at `changelog.d/added/7724-editors-agent-type-fresh.md`, exactly per `changelog.d/README.md`.
Restore `CHANGELOG.md` to `main`'s content, and never resolve that file by picking a side — it is append-only history plus generated sections.

## The stacking is the root cause of most of the noise

The PR is a draft, `CONFLICTING`/`DIRTY` against `main`, and spans 122 files, +10,536 −3,711.
Your own change, `377f45f1c`, is 10 files, +59 −87.
The other 112 files are commits carried from #7715 (`feat/workflow-step-agent-types`, open, `CHANGES_REQUESTED`) and #6931 (`feat/agent-template-crud`, open, `CONFLICTING`), which your PR body already flags.

So every reviewer pass and every required check here is exercising other people's unmerged code, and the changelog loss above arrived through a carried merge rather than through anything you wrote for #7724.
The remedy is to rebase onto `main` once #7715 and #6931 land, so the diff collapses to that single commit.
Unstacking is the highest-leverage move available to you right now — I would rather land the parents first than review the same code three times in three PRs.

## Standing CHANGES_REQUESTED: `CanvasPage.tsx:771` cannot switch back to a concrete agent

```tsx
771: <select value={agentType ? "" : agentId} onChange={e => setAgentId(e.target.value)} className={inputClass}>
778: <select value={agentType} onChange={e => { setAgentType(e.target.value); if (e.target.value) setAgentId(""); }} className={inputClass}>
```

The coupling is one-directional.
Line 778 clears `agentId` when a type is chosen; nothing clears `agentType` when an agent is chosen.
Because line 771 is a controlled `value`, picking a concrete agent while `agentType` is set fires `setAgentId`, then the re-render recomputes `value` as `""` and the select snaps straight back to "No agent".
`handleSave` confirms it downstream at line 707 — `agentId: agentType ? undefined : agentId || undefined` drops the id regardless of what state holds.
The only escape is to set Agent Type back to "" first, which nothing in the UI signals.
This is new here: the diff changed `value={agentId}` to `value={agentType ? "" : agentId}` without adding the inverse clear.

Two ways out, either is fine:

1. Make the clear symmetric. `onChange={e => { setAgentId(e.target.value); if (e.target.value) setAgentType(""); }}` on line 771, and drop the `agentType ? "" :` coercion so the value is just `agentId`. `handleSave`'s existing ternary then persists `agentId` correctly, because `agentType` is already `""` by the time it runs.
2. Model the three states explicitly. One binding value — `none` / `{ kind: "agent", id }` / `{ kind: "type", name, fresh }` — behind a single control, so mutual exclusion lives in the data model instead of in two cross-coupled `onChange` handlers plus a display-value coercion. More churn, but it also removes the four `agentType ? … : …` ternaries in `handleSave` and the build mapping.

Two small things in the same file while you are in there:

- Line 780's `(agentTypes ?? [])` is dead — line 651 already does `agentTypesQuery.data ?? []`.
- Line 780's `map(t => …)` shadows the `t` translation prop declared at line 647. Nothing inside that callback translates today, so it is not a bug, but rename it to `type` before something in there needs `t`.

None of the above is a complaint about the volume of work; `377f45f1c` itself is small, focused and clean, and the locale parity work is right. It is the stack that makes it unreviewable. Get #7715 and #6931 merged, rebase, restore `CHANGELOG.md`, fix the binding, and this becomes a ten-minute review.

DaBlitzStein:
Superseded by #7862, with the TUI counterpart in #7869.

Verified against `origin/main`: `crates/librefang-api/dashboard/src/components/StepAgentBinding.tsx` exports `StepAgentSource = "instance" | "name" | "type"` and its doc comment cites #7712, this branch's issue. The component ships with its own test file.

The `fresh` half of this branch is not in upstream and was not an oversight — see the rationale recorded on `StepAgent` in `workflow.rs`. That belongs on #7714, not here.

PR #7747, opened 2026-08-20
fix(audit): recover a broken audit hash chain without discarding history

### LINKED ISSUE #7702: bug: audit hash-chain integrity fails on boot — chain break at seq 204 (proteo instance)
## Observation

On a production daemon (aarch64, LibreFang 2026.7.31, binary hash 564bb287):

    WARN Audit trail integrity check failed on boot: chain break at seq 204: expected prev_hash f01eebf6fea81eb229f08924...

The audit log is a tamper-evident hash chain (librefang-runtime-audit); a chain break means entry 204's recorded prev_hash does not match the hash of entry 203.

## Context

- Host: Orange Pi 5 (aarch64), user-level systemd, 3 instances on one machine (rodela/proteo/iqvia)
- The instance has been through several binary swaps while running (mv -f over the executing binary), daemon restarts, and at least one window with concurrent daemons
- No audit entries have been manually edited (no tamper suspected)

## Suspects (to verify)

1. Concurrent writes during a restart window (two processes appending to the chain) — the chain append must be single-writer; a double-append breaks the hash linkage.
2. Entry 204 written while the process was being killed mid-write (partial row).
3. Seq numbering bug after recent audit changes.

## Requested

- Tooling/CLI to re-verify and, where appropriate, re-anchor the chain after an operator-confirmed legitimate break (with an AuditAction recording the re-anchor), so one bad write does not permanently degrade the whole integrity property.
- A check of the seq-204 boundary in the shipped binary's audit code.

DaBlitzStein:
Closes #7702.

## What this changes

- Adds `diagnose_chain()` to `librefang-runtime-audit`, which walks `audit_entries` in `seq` order against the raw database and returns the first break as a `ChainBreak { seq, kind, expected, found }`.
  `ChainBreakKind` distinguishes a `PrevHashMismatch` (the link to the previous entry is wrong) from a `HashMismatch` (the entry's own content hash does not match its stored hash), because the two point at different failure modes.
- Adds `reanchor_after_break()`, which deletes only rows with `seq >= break.seq` and then inserts a `ChainReanchored` marker entry chained from the last surviving row, returning a `ReanchorOutcome { rows_deleted, marker_seq, new_tip }`.
  Entries before the break are never read for anything other than computing the marker's `prev_hash`, so pre-break forensic history survives intact.
  A break at `seq 0` falls back to the genesis anchor.
- Adds the `librefang security audit-reanchor` CLI command wired to both.
  Without `--confirm` it prints the diagnosed break and the exact rows it would delete and exits non-zero; with `--confirm` it applies the truncation and refreshes the on-disk anchor file.
  It refuses to run while a daemon holds the database, reusing the existing `monitoring-daemon-running-error` guard that `audit-reset` already uses.
  An optional `--note` is recorded in the marker's detail field so the operator's justification lands in the trail itself.

The existing recovery tool, `audit-reset`, unconditionally wipes the entire `audit_entries` table.
That is the correct hammer when the whole trail is suspect and the wrong one when a single entry broke and everything before it verified cleanly, which is the case this command covers.

## Root cause of the break itself is NOT determined

This PR adds recovery, not a fix for whatever produced the break on the reporting host.
Two candidate mechanisms were ruled out with evidence rather than left open:

- A concurrent second daemon writing the same table cannot corrupt the chain, because `seq` is an `INTEGER PRIMARY KEY`: a colliding writer's `INSERT` fails closed and the entry is dropped rather than interleaved.
- A process death midway through a write cannot leave a half-linked entry, because the in-memory tip only advances after a durable commit, which the pre-existing `test_db_failure_does_not_advance_in_memory_chain` already pins.

Confirming the actual trigger needs the affected host's daemon and journal logs spanning the write at the broken seq, plus any out-of-band operations (binary swaps, backups, restores) touching `~/.librefang/librefang.db` around that time.
Those are not available to this change, and I would rather say so than name a plausible cause I cannot back.

## Verification

- `cargo test -p librefang-runtime-audit` — 48 passed, 0 failed.
  New cases: `boot_detects_chain_break_reports_exact_seq_and_does_not_prevent_startup`, `diagnose_chain_finds_the_same_break_boot_verification_reports`, `diagnose_chain_returns_none_for_a_healthy_chain`, `diagnose_chain_detects_hash_mismatch_distinct_from_link_mismatch`, `reanchor_after_break_preserves_pre_break_history_and_resumes_a_verifiable_chain`, `reanchor_after_break_at_seq_zero_falls_back_to_genesis`, `reanchor_after_break_records_seq_and_note_in_marker_detail`.
  The first of those also pins the pre-existing degrade-not-crash boot behaviour, which had no regression coverage before.
- `cargo test -p librefang-cli --test i18n_checks` — 3 passed (`test_locales_cover_used_i18n_keys`, `test_no_dead_locale_keys`, `test_no_untranslated_strings`).
  The new keys are added to all four CLI locales (`en`, `ko`, `uk`, `zh-CN`), not just English, since `test_locales_cover_used_i18n_keys` asserts coverage for `uk` and `zh-CN` too.
- `cargo check -p librefang-cli --bins` and `cargo check -p librefang-runtime-audit --lib --tests` — clean.
- `cargo clippy -p librefang-runtime-audit -p librefang-cli --all-targets -- -D warnings` — clean.
- `cargo fmt -p librefang-cli -p librefang-runtime-audit -- --check` — clean.

## Deferred

Nothing.


houko (changes requested):
Blocking findings for this destructive recovery path:

- `cmd_audit_reanchor` writes the external anchor as `"seq:hash"`, but `AuditLog::read_anchor` only accepts `"seq hash\n"`. It also uses a direct `fs::write` instead of the existing temporary-file-plus-rename path. A successful reanchor therefore leaves an unreadable/non-atomic anchor and the next boot still fails integrity verification. Expose and reuse the canonical anchor writer from `librefang-runtime-audit`, and add an end-to-end test that reopens `AuditLog::with_db_anchored` after recovery.
- The anchor seq is set to `marker_seq + 1`, but the anchor contract is the persisted row count. Those differ after prefix retention/pruning or any seq gap. Query `COUNT(*)` in the recovery transaction, return that count, and write it to the anchor. Add a recovery test after a persisted prefix trim.
- `reanchor_after_break` trusts an arbitrary/stale `ChainBreak` and deletes `seq >= break_point.seq` without re-diagnosing inside the transaction. The CLI diagnoses before opening the write transaction, and the daemon-file guard does not exclude every other SQLite writer. Revalidate that the first current break exactly matches kind/seq/expected/found under the same transaction before deleting; otherwise abort without mutation. Use an immediate transaction or equivalent writer exclusion to close the diagnosis/delete TOCTOU window.

The current unit tests exercise only the database helper and cannot detect these CLI/anchor failures. Please add the full CLI-equivalent anchored recovery path before merging.

DaBlitzStein:
Superseded by #7855, which landed the whole surface rather than just the recovery routine.

Verified against `origin/main`: `librefang security audit-reanchor` is declared in `crates/librefang-cli/src/cli.rs:1570`, implemented as `cmd_audit_reanchor` in `commands/monitoring.rs:274` over `recovery::plan_reanchor`, and its strings are in the Fluent catalogues.
Same guarantees this branch aimed for: severed rows archived as JSON Lines with their SHA-256 committed into the `ChainReanchored` marker, pre-break hashes untouched.

PR #7748, opened 2026-08-20
fix(skills): degrade gracefully when a marketplace serves HTML instead of JSON

### LINKED ISSUE #7387: bug: Skillhub marketplace API is dead — tencent.com redirects to skillhub.cn SPA
## Symptom

`GET /api/skillhub/search?q=...` and `/api/skillhub/browse` return empty items with:
`Network error: Failed to parse Skillhub search response: expected value at line 1 column 1`

## Root cause (verified from a live node)

`https://skillhub.tencent.com/api/v1/search` now answers **301 → https://skillhub.cn/...** which serves a client-side SPA (HTML), not JSON. The public API the client was built against is gone; the new domain has no JSON endpoint at `/api/v1/search` either (all probed paths return HTML).

The client constant is `DEFAULT_SKILLHUB_URL = "https://skillhub.tencent.com/api/v1"` in `crates/librefang-skills/src/skillhub.rs`.

## Impact

The Skillhub tab in the dashboard shows an error or empty list. Install/search from Skillhub is fully broken. ClawHub and FangHub are unaffected.

## Suggested fix

Either:
1. Update the base URL and response parsing to the new Skillhub API once its JSON contract is known, or
2. Mark the Skillhub hub as unavailable in the UI and backend (graceful degradation: hub card shows "offline" instead of an error), keeping the code path for when the API returns.

This needs a maintainer decision on whether the Skillhub partnership/API still exists.

DaBlitzStein:
Closes #7387.

## What this changes

- Adds `librefang_skills::looks_like_markup()`, a shared check that reports whether a response body starts with `<` once leading whitespace is skipped.
  Valid JSON never does, so this cleanly separates "the marketplace is serving its SPA shell" from "the JSON is genuinely malformed".
- Adds `SkillError::MarketplaceUnavailable(String)` and returns it from the Skillhub search and index parsers and from the shared `ClawHubClient` parsing boundary, so `get_skill` (used by both ClawHub and Skillhub) benefits without a duplicated check.
- Maps `MarketplaceUnavailable` to `503 Service Unavailable` in `routes/skills/skillhub.rs` for search, browse, detail, and install.
  Before, those paths returned `502 Bad Gateway` or, on detail, a `404` that falsely implied the skill did not exist.
- Makes the Skillhub base URL configurable through `LIBREFANG_SKILLHUB_URL`, falling back to the compiled-in default when unset.

## Why not just change the URL

`skillhub.tencent.com` redirects every API path to the `skillhub.cn` single-page app, which answers `200 OK` with an HTML document on every route, including ones that never existed.
There is no verified replacement API endpoint to point at, so the correct behaviour is an honest `503` with an actionable message rather than a cryptic `expected value at line 1 column 1` from `serde_json`.
The env var exists so an operator can point at a mirror without recompiling if one appears.

Local skill install and search against FangHub and ClawHub are unaffected.

## Verification

- `cargo test -p librefang-skills --lib` — 266 passed, 0 failed.
  New cases: `looks_like_markup_detects_html_document`, `looks_like_markup_detects_html_with_leading_whitespace`, `looks_like_markup_rejects_valid_json_object`, `looks_like_markup_rejects_valid_json_array`, `looks_like_markup_rejects_whitespace_padded_json`, `looks_like_markup_treats_empty_body_as_non_markup`, `search_body_html_response_degrades_to_marketplace_unavailable`, `search_body_valid_json_still_parses`, `search_body_genuinely_malformed_json_is_a_network_error_not_marketplace_unavailable`, `index_body_html_response_degrades_to_marketplace_unavailable`, `index_body_valid_json_still_parses`, `with_defaults_honors_env_var_override`, `with_defaults_falls_back_when_env_var_unset`.
  The malformed-JSON case is there deliberately: a truncated or corrupt JSON body must stay a `Network` error and must not be misreported as a dead marketplace.
- `cargo clippy -p librefang-skills -p librefang-api --all-targets -- -D warnings` — clean.

## Deferred

Nothing.


DaBlitzStein:
CI note: the red checks on this PR are a pre-existing break on `main`, not something this diff introduces.

`cargo check -p librefang-kernel --lib` fails on a pristine `a5f2c5548` checkout with `unused import: librefang_runtime::kernel_handle::prelude::*` at `crates/librefang-kernel/src/kernel/messaging.rs:18`.
Neither branch here touches that file — `git diff origin/main <branch> -- crates/librefang-kernel/src/kernel/messaging.rs` is empty.
The failure is package-selection dependent, so main's full-workspace lane stays green while the selective PR lanes (`cargo test -p librefang-api ... --no-run`, and `cargo xtask codegen --openapi` inside `OpenAPI Drift`) hit it.

Filed as #7754 with the full repro.
Happy to fold the one-line fix in here instead if a maintainer prefers that over a separate PR — it is in a different crate than this change, so I did not want to bundle it unasked.

DaBlitzStein:
Fix is up as #7755 (one-line deletion in `librefang-kernel`, kept out of this PR so it stays a single unit). This PR's CI will go green once that lands or on a rebase past it.

houko (changes requested):
Blocking findings:

- `ClawHubClient` now returns `MarketplaceUnavailable`, but only the Skillhub routes map it. The shared client is also used by ClawHub and ClawHub CN: their detail handlers still turn the new variant into 404, and their install handlers turn it into a scrubbed 500 instead of the previous upstream-failure class. Update every ClawHub/ClawHub CN search, browse, detail, and install mapping consistently and add route-level tests for 200+HTML responses.
- `LIBREFANG_SKILLHUB_URL` only changes search and delegated detail requests. `browse` and `install` still hard-code `SKILLHUB_INDEX_URL`, and install also hard-codes `SKILLHUB_COS_BASE`. An operator cannot restore the marketplace by pointing at a mirror as the error text and PR body claim. Either make all required endpoints configurable as one coherent mirror contract or narrow the message/documentation so it does not promise browse/install recovery.

Please also make the markup detector tolerate a UTF-8 BOM before `<` and test that common HTML response shape. The current CI Gate is failing, so rebase and run the full route/build checks after these fixes.

DaBlitzStein:
Superseded, and by a better shape than this branch had.

Verified against `origin/main`: `SkillError::MarketplaceUnavailable` (`crates/librefang-skills/src/lib.rs:73`) is raised from a single choke point rather than per endpoint — the doc comment on `lib.rs:90` states the invariant outright: every remote marketplace read in the crate goes through there, so the condition cannot diverge between search and install. The HTTP layer maps it uniformly via `marketplace_error_status`, and the dashboard renders one offline state through `isMarketplaceUnavailable` (`lib/http/errors.ts:72`, used at `SkillsPage.tsx:2136`).

This branch fixed the same fault at the SkillHub client only, which would have left the other hubs to be fixed one at a time. Closing in favour of the centralised version.

PR #7751, opened 2026-08-20
feat(dashboard): link marketplace skills back to their hub page

DaBlitzStein:
## What

An installed marketplace skill had no way back to the page it came from. The hub is known (`getSkillHub`), the slug is known, but nothing rendered a link, so checking a skill's description, version history or author meant retyping the slug into the right hub by hand.

## How

`skillHubUrl(id, slug)` on the existing `skillHubs` module, plus a `skillUrl` member on each hub entry, and an `ExternalLink` affordance on the skill card.

`skillUrl` returns `string | null` rather than a string, and that is the point of the shape: FangHub ships no public web UI, and SkillHub is self-hosted, so its origin is only known when `VITE_SKILLHUB_REGISTRY_URL` is set at build time. `null` is what stops the UI offering a link that would land on an unrelated site. `skillHubUrl` returns `null` for an unknown hub, an empty slug, or a hub with no addressable page, and the card renders no link in those cases rather than a dead one.

## Verification

- `pnpm vitest run src/lib/skillHubs.test.ts` → **16 passed**, covering each hub's URL shape and the three `null` paths.
- `pnpm typecheck` → clean.
- **`pnpm build` → `✓ built in 3.12s`.** Run deliberately, not just `tsc`: the project aliases `lucide-react` to the curated `src/lib/lucide.ts` at bundle time, so an icon that type-checks against the package can still fail the production build. `ExternalLink` is exported from that barrel (`lucide.ts:42`), and the build confirms it.
- `pnpm vitest run src/lib/__tests__/locale-parity.test.ts` → **8 passed**. The new key is present in all five catalogues.

## Note on this push

Rebuilt from `origin/main` rather than rebased. The `skillHubs` module itself has since landed upstream, so this branch no longer carries it — what remains is the `skillUrl` member, the `skillHubUrl` accessor, its tests, the card affordance and the five locale keys.

Net: 136 lines across 9 files, all dashboard.


DaBlitzStein:
CI is red on this PR for a reason that is not in its diff: `main` does not compile under Rust **1.98.0**, which the runners picked up mid-afternoon today.

- `crates/librefang-kernel/src/kernel/messaging.rs:18` glob-imports `librefang_runtime::kernel_handle::prelude::*`, and its parent `crates/librefang-kernel/src/kernel/mod.rs:29` already does — `messaging.rs` is a submodule that pulls the parent scope in with `use super::*`. 1.98.0 reports that redundant glob as `unused_imports`, and the workspace denies warnings, so `cargo check --workspace --all-targets` stops there.
- Evidence that it is toolchain and not content: #7751 changes no Rust at all and fails the same lane with the same error. The `Build / Linux aarch64` job on #7748 passed at 17:09Z on `rustc 1.97.1`; the job on this PR ran at 17:37Z after the runner updated to `rustc 1.98.0 (88d9e12ae 2026-08-18)` and failed. Reproduces locally on 1.98.0 against a clean `origin/main` checkout.
- Fix is deleting that one line. The workspace check aborts at the first failing crate, so there may be further redundant globs behind it.

Not fixing it from here since it is a different crate than this PR's subject; flagging so the red is not misread as this change.

PR #7753, opened 2026-08-20
fix(skills): a malformed published skill is a 400, from every marketplace handler

DaBlitzStein:
## What

`SkillError::YamlParse` is the published skill's own `SKILL.md` frontmatter failing to parse.
It fell through `marketplace_error_status` to each handler's fallback, so the same underlying fault answered `500` from install and `404` from detail.

Both were wrong, in opposite directions:

- The install handlers scrub the response body on `500` (audit: `rusqlite-errors-leak`), so the one message that named the broken file and its syntax error was replaced with "Internal server error" — and retrying re-downloads the same broken file.
- Detail's `404` said the skill does not exist, when it exists and is unusable.

## How

One arm in `marketplace_error_status`, alongside `MarketplaceUnavailable`, rather than a branch per handler.
That placement is deliberate and follows the reason already recorded on the `503` arm: the condition means one thing wherever it surfaces, and a caller who gets a different status per endpoint has no way to tell it is the same fault.

The arm overrides the read endpoints' `502` and detail's `404` as well. That is the intended effect, and it is stated in the doc comment rather than left for a reader to discover.

## Verification

- `routes::skills::tests::malformed_skill_frontmatter_is_a_bad_request_from_every_handler` fails on `origin/main` without the arm:
  ```
    left: 500
   right: 400
  ```
  The test covers all three fallbacks a real handler passes in — `500` (install), `502` (browse/search), `404` (detail) — so it proves the claim its name makes rather than one case of it.
- `marketplace_error_status_keeps_its_other_arms` pins `503`, `429`, and pass-through of an unopinionated error, so the new arm cannot quietly swallow the existing classification.
- `cargo test -p librefang-api --lib routes::skills::tests::` → 20 passed, 0 failed.
- `cargo clippy -p librefang-api --all-targets -- -D warnings` → clean.
- Rebased on `origin/main` at `a25982bc6`.

## Note on scope

This branch previously introduced its own `install_error_status` helper and rewired the three install handlers to it.
`marketplace_error_status` (#7387) has since landed upstream and does that centralisation better, so the branch was rebuilt from `origin/main` as a single arm on the existing function instead. The diff is now one file plus a changelog fragment.


DaBlitzStein:
CI is red on this PR for a reason that is not in its diff: `main` does not compile under Rust **1.98.0**, which the runners picked up mid-afternoon today.

- `crates/librefang-kernel/src/kernel/messaging.rs:18` glob-imports `librefang_runtime::kernel_handle::prelude::*`, and its parent `crates/librefang-kernel/src/kernel/mod.rs:29` already does — `messaging.rs` is a submodule that pulls the parent scope in with `use super::*`. 1.98.0 reports that redundant glob as `unused_imports`, and the workspace denies warnings, so `cargo check --workspace --all-targets` stops there.
- Evidence that it is toolchain and not content: #7751 changes no Rust at all and fails the same lane with the same error. The `Build / Linux aarch64` job on #7748 passed at 17:09Z on `rustc 1.97.1`; the job on this PR ran at 17:37Z after the runner updated to `rustc 1.98.0 (88d9e12ae 2026-08-18)` and failed. Reproduces locally on 1.98.0 against a clean `origin/main` checkout.
- Fix is deleting that one line. The workspace check aborts at the first failing crate, so there may be further redundant globs behind it.

Not fixing it from here since it is a different crate than this PR's subject; flagging so the red is not misread as this change.

DaBlitzStein:
CI note: `Test / Ubuntu (shard 1/4)` is red here for a reason that is not in this diff.

Six tests fail, all in `crates/librefang-api/tests/profiles_templates_routes_integration.rs`:

```
templates_get_promotion_preview_flags_a_secret_inside_a_retained_prompt      FAILED
templates_get_promotion_preview_omits_host_specifics_and_keeps_the_portable_half  FAILED
templates_get_promotion_preview_is_quiet_for_an_already_portable_template    FAILED
templates_get_reports_what_promotion_would_strip                            FAILED
templates_list_carries_provider_and_model_from_the_manifest                 FAILED
templates_list_skips_a_malformed_manifest_instead_of_failing                FAILED
```

with `assertion left == right failed: {"error":{"message":"Template 'tmpl_promo_clean' not found"}}`.

The same six fail on #7931 and #7932, which are unrelated branches, so this is a `main` condition rather than something either branch introduced. This PR touches one function in `routes/skills/mod.rs` — the `YamlParse` arm of `marketplace_error_status` — and no template code, fixture, or route.

Flagging rather than acting on it: it is not my diff to fix, and re-running the lane would only reproduce it.

PR #7755, opened 2026-08-20
chore(kernel): drop the now-unused kernel_handle prelude import from messaging

DaBlitzStein:
Closes #7754.

## What this changes

Deletes one line: `use librefang_runtime::kernel_handle::prelude::*;` at `crates/librefang-kernel/src/kernel/messaging.rs:18`.
The file already does `use super::*;`, which brings the same names in from `kernel::mod`, so the glob was redundant.

## Why it is not cosmetic

The workspace denies `unused_imports`, and whether rustc flags this particular glob depends on the package selection cargo is given.
On a pristine `main` (`a5f2c5548`, clean tree):

- `cargo check -p librefang-kernel --lib` → `error: unused import` → build fails
- `cargo check -p librefang-api --tests` → same failure
- `cargo check -p librefang-api -p librefang-kernel --lib` → passes
- `cargo test --workspace --no-run` → passes

Pushes to `main` take the last of those, which is why the branch has stayed green.
The selective PR lanes take the narrow selections: `Test / Ubuntu` runs `cargo test -p <affected crates> --no-run`, and `OpenAPI Drift` runs `cargo xtask codegen --openapi`, which shells out to `cargo test -p librefang-api --test openapi_spec_test`.
Both therefore fail on PRs that do not touch `librefang-kernel` at all — see the red checks on #7748 and #7749, neither of which has a single changed line in `messaging.rs`.

It also breaks the scoped `cargo check -p <crate>` / `cargo test -p <crate>` workflow the contributor docs prescribe over the workspace-wide form.

## Verification

All on this branch:

- `cargo check -p librefang-kernel --lib` — clean (fails on `main` without this change).
- `cargo check -p librefang-kernel --all-targets` — clean, so nothing in the crate's own test targets depended on the glob either.
- `cargo check -p librefang-api --tests` — clean (fails on `main` without this change).
- `cargo test -p librefang-kernel --lib` — 1480 passed, 0 failed, 1 ignored.

## Deferred

Nothing.

Issue #7754, opened 2026-08-20
Scoped `cargo check -p librefang-kernel --lib` fails on main: unused import at messaging.rs:18


DaBlitzStein:
## Repro

On a pristine checkout of `main` (currently `a5f2c5548`), with a clean tree:

```
cargo check -p librefang-kernel --lib
```

```
error: unused import: `librefang_runtime::kernel_handle::prelude::*`
  --> crates/librefang-kernel/src/kernel/messaging.rs:18:5
   |
18 | use librefang_runtime::kernel_handle::prelude::*;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`

error: could not compile `librefang-kernel` (lib) due to 1 previous error
```

`cargo check -p librefang-kernel --tests` fails the same way (the lib compile is what breaks).

## Why main's own CI is green

The failure is selection-dependent.
Widening the package selection makes it disappear — `cargo check -p librefang-api -p librefang-kernel --lib` and `cargo clippy -p librefang-api -p librefang-kernel --all-targets -- -D warnings` both pass on the same commit.
Pushes to `main` take the full-workspace path (`cargo test --workspace --no-run`), which is one of the passing selections, so the lane never sees it.

## Why it matters

The selective PR lanes do take a narrow selection, so this surfaces as a spurious PR failure with no connection to the diff.
Observed on `Test / Ubuntu (shard 1/4)` running `cargo test -p librefang-api -p librefang-skills --no-run` and on `OpenAPI Drift` running `cargo xtask codegen --openapi` (which shells out to `cargo test -p librefang-api --test openapi_spec_test`), on PRs that do not touch `librefang-kernel` at all.
It also breaks the workflow `CLAUDE.md` prescribes for contributors, which is scoped `cargo check` / `cargo test -p <crate>`.

## Note on the likely cause

`messaging.rs` already does `use super::*;` on line 27, and `KernelHandle` and friends appear to resolve through that glob rather than through line 18, which would make line 18 redundant.
I have not confirmed that removing it keeps every selection green, so I am reporting the break rather than guessing at the patch.

PR #7762, opened 2026-08-21
fix(qa): green the workspace clippy gate and stub scrollIntoView globally for jsdom

### LINKED ISSUE #7761: bug: the workspace clippy gate is red on main under Rust 1.98, and jsdom's missing scrollIntoView can fail a vitest run whose assertions all pass
Two independent CI-gate problems, both in code nobody is currently changing.

## 1. `cargo clippy --workspace --all-targets -- -D warnings` fails on main

Reproduced on `origin/main` at `a5f2c5548` with:

```
rustc 1.98.0 (88d9e12ae 2026-08-18)
clippy 0.1.98 (88d9e12ae1 2026-08-18)
```

Six lint sites, all pre-existing and all unrelated to any open PR:

- `clippy::manual_slice_fill` / chunk-indexing lints on `chunks_exact(4)` + per-element indexing, in `librefang-memory/src/consolidation.rs` (four sites, one in `decode_embedding` and three in tests), `librefang-memory/src/semantic.rs`, and `librefang-runtime/src/embedding.rs`.
  These decode a little-endian `f32` buffer by rebuilding `[c[0], c[1], c[2], c[3]]` for every chunk; `as_chunks::<4>()` already hands back `&[u8; 4]`.
- `chunks_exact_mut(4)` in `librefang-desktop/src/tray.rs` (RGBA to ARGB rotation).
- `clippy::useless_asref`-adjacent `map_or(0, |s| s)` in `librefang-llm-drivers/src/drivers/mod.rs`.
- `drain(..).collect()` into a fresh `Vec` in `librefang-llm-drivers/src/drivers/bedrock.rs`, where `std::mem::take` says the same thing.
- `clippy::result_large_err` on the two terminal auth helpers in `librefang-api/src/routes/terminal.rs`, whose `Err` payload is an axum `Response`.

The MSRV is 1.94.1 and `slice::as_chunks` has been stable since 1.88, so the mechanical rewrites are available.

The `result_large_err` case is not a mechanical rewrite: boxing the `Err` would ripple through every call site to save stack on a code path that is already allocating an HTTP response, so it wants a documented `#[allow]` rather than an allocation.

## 2. jsdom has no `scrollIntoView`, and the resulting throw fails the whole run

jsdom implements no layout, so `Element.prototype.scrollIntoView` does not exist.
A component that scrolls a focused element into view — `useListNav` does this, and so does anything built on it — throws an uncaught `TypeError` inside a passive effect.
vitest reports that as an unhandled error and exits non-zero **even when every assertion in the file passed**.

Demonstrated on `origin/main` by removing the per-file guard in `src/pages/ApprovalsPage.test.tsx` and running that file alone:

```
 Test Files  1 failed (1)
      Tests  3 failed | 26 passed (29)
```

`ApprovalsPage.test.tsx`, `UsersPage.test.tsx` and `McpServersPage.test.tsx` each carry an identical copy of the same `beforeEach` guard, which is why the suite is green today.
That is three copies of one workaround, and any suite added later has to know to repeat it — the stub belongs in `setupTests.ts`.

DaBlitzStein:
Closes #7761.

## Changes

- `librefang-memory/src/consolidation.rs`, `librefang-memory/src/semantic.rs`, `librefang-runtime/src/embedding.rs`: decode little-endian `f32` buffers with `as_chunks::<4>()` instead of `chunks_exact(4)` plus per-element indexing.
  `as_chunks` yields `&[u8; 4]`, so `f32::from_le_bytes` takes the array directly rather than having `[c[0], c[1], c[2], c[3]]` rebuilt for every chunk.
  Four of the six sites are in `consolidation.rs` (one in `decode_embedding`, three in its tests).
- `librefang-desktop/src/tray.rs`: same rewrite for the `chunks_exact_mut(4)` RGBA to ARGB rotation.
- `librefang-llm-drivers/src/drivers/mod.rs`: `map_or(0, |s| s)` becomes `unwrap_or(0)`.
- `librefang-llm-drivers/src/drivers/bedrock.rs`: `drain(..).collect()` into a fresh `Vec` becomes `std::mem::take`.
- `librefang-api/src/routes/terminal.rs`: `#[allow(clippy::result_large_err)]` with a written reason on `authorize_terminal_request` and `tmux_controller`.
  Their `Err` payload is an axum `Response`; boxing it would ripple through every call site to save stack on a path that is already building an HTTP response.
- `crates/librefang-api/dashboard/src/setupTests.ts`: install a no-op `Element.prototype.scrollIntoView` when it is absent.

MSRV is 1.94.1 and `slice::as_chunks` is stable since 1.88, so no MSRV bump is involved.

## Verification

`cargo clippy -p librefang-memory -p librefang-runtime -p librefang-llm-drivers --all-targets -- -D warnings` — clean.

`cargo clippy -p librefang-api -p librefang-desktop --all-targets -- -D warnings` — `Finished dev profile ... in 2m 05s`, zero warnings.
Both of these crates pull in `librefang-kernel`, which does not currently build under `-D warnings` on `main` because of the unused import at `messaging.rs:18` (#7754).
That is not touched by this PR; the run above was done with #7755's one-line deletion applied locally so the gate could be read, and that line is **not** part of this branch.
This PR's own CI clippy job will stay red until #7755 lands — it is not a defect in this change.

Dashboard suite, `crates/librefang-api/dashboard`:

```
 Test Files  122 passed (122)
      Tests  1186 passed (1186)
```

exit code 0, both before and after — main is green today because of the per-file guards described below.

The mechanism the `setupTests.ts` stub addresses was confirmed directly rather than assumed.
Removing the per-file `beforeEach` guard from `src/pages/ApprovalsPage.test.tsx` on an otherwise unmodified tree gives:

```
 Test Files  1 failed (1)
      Tests  3 failed | 26 passed (29)
```

with `el.scrollIntoView is not a function` raised from `useListNav` inside a React passive effect.
Restoring only the global stub in `setupTests.ts`, with that per-file guard still removed, gives `29 passed (29)`.
Both experiments were reverted; neither is part of the diff.

## Noted, deliberately not changed here

`ApprovalsPage.test.tsx`, `UsersPage.test.tsx` and `McpServersPage.test.tsx` each keep their own copy of the guard.
All three are written as `if (!Element.prototype.scrollIntoView)`, so once `setupTests.ts` installs the stub they are no-ops — dead code, not a hazard.
Deleting them is a one-line removal per file and I am happy to fold it into this PR; I left them alone to keep the diff limited to what fixes the gate.
Say the word either way.


DaBlitzStein:
Superseded by #7791, which carries this fix plus the five additional locations that appeared on main after this PR was written, and has already merged. Closing.

DaBlitzStein:
Closed as superseded by #7791 (already merged to main).

PR #7768, opened 2026-08-21
feat(registry): resolve the agent-templates directory through one shared, loud resolver

### LINKED ISSUE #7767: bug: a registry checkout without an `agents/` directory silently disables agent-template preinstall and hand routing
## What happens

Three call sites resolve the registry's agent-templates directory, and each does its own bare `join("agents")` plus an existence check that skips its entire block when the directory is not there.
None of them logs anything in that case.

On `origin/main` (`e61bbbca3`):

- `crates/librefang-runtime/src/registry_sync.rs`, in `fanout_registry_content`:
  ```rust
  let agents_src = registry_cache.join("agents");
  if agents_src.exists() {
      // ... the entire pre-install of agent templates
  }
  ```
  No `else`. If the directory is missing the fan-out installs nothing and says nothing.

- `crates/librefang-hands/src/registry.rs`, `resolve_agents_dir` returns `None` with no log.

- `crates/librefang-kernel-router/src/lib.rs`, `load_hand_route_candidates` sets `agents_dir_arg = None` with no log.

## Why it matters

"The registry genuinely ships no agent templates right now" and "the sync produced a checkout this code cannot read" are indistinguishable from outside the process.

The registry cache refreshes on a TTL (`DEFAULT_CACHE_TTL_SECS = 24 * 60 * 60`, `registry_sync.rs`), so a checkout that lands in a shape these three lookups do not recognise degrades quietly and stays degraded:

- a fresh install ends up with zero preinstalled agent templates, `hello-world` included;
- hands that declare `base = "<template>"` fail the flat parse path and drop out of routing.

Neither produces an error the operator can act on, and the second one only ever surfaces as a downstream per-hand parse warning that names the hand rather than the missing directory.
The failure is silent by construction, which is the part worth fixing regardless of whether the triggering condition is common.

## Suggested fix

One shared resolver, so the three call sites cannot drift, that logs at error level when the directory cannot be resolved at all.

While the resolver exists it is nearly free to have it accept `agent-types/` as well as `agents/`.
To be clear about the state of that: `librefang/librefang-registry` serves `agents/` today and there is **no announced rename** — I checked the repository contents and its open issues and PRs before filing this.
The core is converging on "agent type" as the name for this concept (#7722, #7740, #7758), so accepting either name is cheap insurance against exactly the silent-miss shape described above, and it costs one `is_dir` call on a path already being stat-ed.
It should not warn when it resolves `agents/`, since that is the current correct state and a warning there would fire on every sync for every install.

Happy to send a PR for this.

DaBlitzStein:
Closes #7767

## The problem

Three call sites resolve the registry's agent-templates directory, each with its own bare `join("agents")` plus an existence check, and each silently skips its whole block when the directory is not there.
On `e61bbbca3`:

- `crates/librefang-runtime/src/registry_sync.rs` — `fanout_registry_content` has `if agents_src.exists() { ... }` with no `else`, so the agent-template pre-install is skipped without a word.
- `crates/librefang-hands/src/registry.rs` — `resolve_agents_dir` returns `None` with no log.
- `crates/librefang-kernel-router/src/lib.rs` — `load_hand_route_candidates` sets `agents_dir_arg = None` with no log.

"The registry ships no agent templates" and "the sync produced a checkout this code cannot read" are therefore indistinguishable from outside the process, and the cache refreshes on a 24h TTL (`DEFAULT_CACHE_TTL_SECS`), so the degraded state persists.
A fresh install ends up with no `hello-world`, and hands declaring `base = "<template>"` drop out of routing — the latter surfacing only as a downstream per-hand parse warning that names the hand rather than the missing directory.

## Changes

- New `librefang_types::registry_paths::resolve_agent_types_dir`, the single resolver for all three call sites, which logs at **error** level when it can resolve neither directory name, naming both paths it tried. This is the load-bearing half of the change.
- The three call sites now delegate to it instead of open-coding the lookup, so they cannot drift apart again.
- The resolver accepts `agent-types/` ahead of `agents/`.

## On the second directory name, explicitly

`librefang/librefang-registry` serves `agents/` today and there is **no announced rename** — I checked the repository contents and its open issues and PRs before writing this, and the issue says the same.
So `agents/` is the branch that normally runs.
It is deliberately **not** warned about: a warning on the current correct state would fire on every sync for every install, which is how a log line teaches operators to ignore it. It logs at `debug` so the resolution stays recoverable when diagnosing an empty fan-out.
Accepting both names costs one `is_dir` call on a path already being stat-ed, and closes the silent-miss shape above against any future layout change.
If you would rather this PR carry only the loud-error half and drop the `agent-types/` branch entirely, say so and I will cut it — the two halves are independent.

## Verification

- `cargo test -p librefang-types --lib registry_paths` — **4 passed**: `only_agent_types_dir_resolves_without_warning`, `only_legacy_agents_dir_resolves_without_warning`, `both_present_prefers_agent_types`, `neither_present_errors_loudly`. The last asserts the error level and that both tried paths are named; the middle two assert no `WARN`/`ERROR` on either single-directory case.
- `cargo test -p librefang-runtime --lib registry_sync` — **24 passed**, including four new fan-out tests: `fanout_agent_templates_resolves_legacy_agents_dir_only`, `fanout_agent_templates_resolves_agent_types_dir_only`, `fanout_agent_templates_prefers_agent_types_when_both_present`, `fanout_agent_templates_noop_when_neither_dir_present`.
- `cargo clippy -p librefang-types -p librefang-runtime -p librefang-hands -p librefang-kernel-router --all-targets -- -D warnings` — **zero findings in the four crates this PR touches**. The run does report two `clippy::map_or_identity` errors under Rust 1.98, both pre-existing on `origin/main` and both outside this PR's diff: `crates/librefang-llm-drivers/src/drivers/bedrock.rs:538` and `crates/librefang-llm-drivers/src/drivers/mod.rs:115`. They are fixed by open PR #7762, which touches exactly those two files. Until that lands this PR's clippy job will be red for reasons unrelated to it.
- The fixture README's "byte-identical copy" claim about the two `hello-world/agent.toml` files was checked with `diff`; it holds.

## Notes

- `crates/librefang-types/Cargo.toml` gains `tracing-subscriber` as a **dev**-dependency only, for the log-capture helper in the new unit tests. `Cargo.lock` moves by that one line.
- No overlap with any other open PR: I cross-checked the file list of all 40 open PRs against the six source files here and found none in common.

---

## CI status, for the record

The red lanes on this PR are not caused by it.

- `Quality` and `OpenAPI Drift` both fail on `error: unused import: librefang_runtime::kernel_handle::prelude::*` at `crates/librefang-kernel/src/kernel/messaging.rs:18`. That is #7754, and open PR #7755 is the one-line deletion that fixes it. Every kernel-touching PR is currently blocked behind it.
- The `Test / Unit` and `Test / Ubuntu (shard N/4)` lanes failed on a runner-side error before reaching any test: `bash function injection via BASH_FUNC_ environment variable is not allowed for security reasons` / `install-action requires bash`. That is infrastructure, not this diff.

Verification for the actual change was run locally and is listed above.


houko (changes requested):
P1: preferring `agent-types/` as a single directory silently hides all content that exists only in the currently supported `agents/` directory. During a partial migration, interrupted sync, or stale-cache overlap, an `agent-types/` directory containing one template causes every other template still under `agents/` to disappear from fanout, hand base resolution, and routing. The new both-present test puts the same `hello-world` id in both directories, so it cannot detect loss of unique legacy entries.

There is no upstream rename contract yet, so the safest scoped fix is to keep `agents/` authoritative and make its absence loud; add `agent-types/` compatibility only when the registry has a defined migration strategy. If dual-layout support remains, the resolver cannot return only one path: callers need merge semantics and a regression with unique templates in each directory.

This PR also adds `tracing-subscriber` to `librefang-types` without prior dependency approval. Please remove the new dependency (and the fabricated `agent-types/` content from the pinned registry fixture) as part of narrowing the change to the observed missing-directory diagnostic.

houko:
CI is green; this is one change away from mergeable — `331d715` is a merge from `main` only, so the review is the sole blocker. The narrowing is mostly deletion.

**Keep:** the shared resolver over `agents/` plus the loud diagnostic. `crates/librefang-types/src/registry_paths.rs:83-90` (`tracing::error!` + `None` when the directory is absent) is the actual fix for the observed defect, and the three call sites collapsing onto one function is the right shape. That half is uncontested.

**Drop:** the `agent-types/` preference. `registry_paths.rs:66-70`:

```rust
pub fn resolve_agent_types_dir(registry_cache: &Path) -> Option<PathBuf> {
    let canonical = registry_cache.join(AGENT_TYPES_DIR_NAME);
    if canonical.is_dir() {
        return Some(canonical);
    }
```

The early return at `:69` makes `agents/` unreachable whenever `agent-types/` exists at all, even as an empty or half-populated directory. An interrupted sync or a stale cache that leaves one template under `agent-types/` therefore drops every template that only lives under `agents/` — simultaneously from `registry_sync::fanout_registry_content`, `librefang_hands::registry::resolve_agents_dir` and `librefang_kernel_router::load_hand_route_candidates`, since all three now share this function. Delete the branch and the `AGENT_TYPES_DIR_NAME` constant; there is no announced rename in `librefang/librefang-registry` to be compatible with yet, and if one lands the resolver needs merge semantics rather than a preference.

**Drop:** `crates/librefang-runtime/tests/fixtures/registry/agent-types/hello-world/agent.toml`, and revert the README hunk. That tree describes itself as a "Pinned snapshot of librefang/librefang-registry" and closes with "don't re-add content without a test that needs it" — this adds a directory the real registry does not serve. The two tests that need it (`fanout_agent_templates_resolves_agent_types_dir_only`, `fanout_agent_templates_prefers_agent_types_when_both_present`) go away with the branch.

**On the dependency, a correction to my review:** it is a dev-dependency (`crates/librefang-types/Cargo.toml:34`), not a regular one, so nothing new ships in the library. It is still a new edge and a `Cargo.lock` line, and it exists solely for the `capture_logs` helper in `registry_paths.rs:98-137`. After narrowing, the resolver has two outcomes and the returned `Option` is what all three callers actually branch on, so the `librefang-types` tests can cover it with no subscriber at all. If you want to keep one assertion on the error line itself, `librefang-runtime` already carries `tracing-subscriber` as a dev-dependency and already has the `tests/instrument_span_fields.rs` capture pattern your module doc cites — put it there and `librefang-types` needs no new dependency.

Dependency order: nothing else is waiting on this one, so it can land as soon as those deletions are pushed.

DaBlitzStein:
Superseded by #7851.

Verified against `origin/main`: `crates/librefang-types/src/registry_paths.rs` now holds the shared resolver — `resolve_agent_templates_dir` over `AGENT_TEMPLATES_DIR_NAME` — and a missing directory is loud.

Upstream kept `agents` as the directory name rather than renaming to `agent-types` with a legacy fallback, which was the other half of this branch. A rename with a compatibility window is a separate decision from having one resolver, and it should be argued on its own issue rather than ride along with a resolver that already landed.

Issue #6398, opened 2026-07-07
OpenAI-compat driver ignores CompletionRequest.thinking — never sends reasoning_effort, so extended thinking is never requested


neo-wanderer:
## Summary

The OpenAI-compatible driver (`crates/librefang-llm-drivers/src/drivers/openai.rs`) never asks the backend to produce extended thinking.
`build_request()` completely ignores `CompletionRequest.thinking` (the `ThinkingConfig { budget_tokens, stream_thinking }` the caller passes), so it emits neither the OpenAI `reasoning_effort` parameter nor a thinking-enable block on the wire.

Meanwhile the driver's `stream()` path *does* parse `reasoning_content` / `reasoning` deltas into `StreamEvent::ThinkingDelta`.

The result is a request/response asymmetry: the driver can *read* streamed reasoning, but it never *requests* it through the one knob many OpenAI-compatible backends require.
For any OpenAI-compatible provider/gateway that gates extended thinking behind `reasoning_effort` (mapping it to `thinking.budget_tokens` server-side) and ignores a passthrough Anthropic-style `thinking` block, reasoning is therefore never generated and never streamed — even for a model the catalog marks as thinking-capable and even when the caller set a thinking budget.

## Where it breaks

Request side — `OpenAiDriver::build_request` (`crates/librefang-llm-drivers/src/drivers/openai.rs:812`).
The outgoing `OaiRequest.thinking` field (declared at `openai.rs:521`) is only ever set to `{"type":"disabled"}` (for the Kimi `ReasoningEchoPolicy::EmptyString` case) or `None`:

```rust
// openai.rs ~1117, inside build_request()
thinking: if echo_policy == ReasoningEchoPolicy::EmptyString {
    Some(serde_json::json!({"type": "disabled"}))
} else {
    None
},
```

`request.thinking` (`CompletionRequest.thinking: Option<ThinkingConfig>`, `crates/librefang-llm-driver/src/lib.rs:273`) is never read in `build_request` — grepping the file shows no `reasoning_effort` token anywhere, and no branch that translates a requested `budget_tokens` into any wire parameter.

Response side — `OpenAiDriver::stream` (`crates/librefang-llm-drivers/src/drivers/openai.rs:1569`) already handles reasoning deltas correctly:

```rust
// openai.rs ~1954
if let Some(reasoning) = delta["reasoning_content"].as_str() { /* -> ThinkingDelta */ }
[...]

PR #7770, opened 2026-08-21
fix(llm): retry without reasoning_effort when a gateway rejects it, and keep that 400 off the circuit breaker

### LINKED ISSUE #7769: bug(llm): a gateway that rejects `reasoning_effort` fails every turn and then opens the provider circuit breaker
## What happens

Against a litellm gateway fronting an OpenAI-shaped model group, every turn of every affected agent fails, and the agent then goes silent for the full provider cooldown.

The gateway returns:

```
400 litellm.UnsupportedParamsError: openai does not support parameters: ['reasoning_effort'], for model=<model>. Received Model Group=<group>
```

## Why the request carries the field

`#6398` added `reasoning_effort` to the OpenAI-compatible driver.
On `origin/main`, `crates/librefang-llm-drivers/src/drivers/openai.rs` sets it in `build_request` whenever the echo policy is the default `ReasoningEchoPolicy::None` and the request carries a thinking budget:

```rust
reasoning_effort: request
    .thinking
    .as_ref()
    .and_then(|tc| reasoning_effort_for_budget(tc.budget_tokens)),
```

`request.thinking` comes from `manifest.thinking`, and `ThinkingConfig::default()` in `crates/librefang-types/src/config/types.rs` is `budget_tokens: 10_000`, which `reasoning_effort_for_budget` maps to `"medium"`.
So any agent whose manifest carries a `[thinking]` section — including one inherited from its agent type, with no field set — sends `reasoning_effort` on every request.

## Why this is not answerable from a model catalogue

Whether a *model* can reason and whether a *gateway* will forward the field are different questions, and only the second one decides the outcome here.
litellm strips the parameter and rejects the request before the model ever sees it, and the error text says `openai does not support parameters` — naming the adapter, not the model.
The same model behind a different gateway accepts the field.
There is no static table that answers this correctly; it can only be discovered by trying.

Reproduced by sending the same request twice to the same gateway: without `reasoning_effort` it answers, with it the call 400s.

## The second half: the circuit breaker

`crates/librefang-runtime/src/agent_loop/retry.rs` calls `record_retry_failure(provider, cooldown, is_billing)` for every non-retryable error, so this 400 counts against `ProviderCooldown`.
After enough turns the circuit opens and the agent stays mute for the cooldown window — for a deterministic, self-inflicted parameter mismatch that says nothing about whether the provider is reachable.

The cooldown is the correct response to a provider that is failing.
It is the wrong response to a request we are constructing wrongly, because backing off does not change the outcome: the next request after the cooldown carries the same field and fails identically.

## Suggested fix

1. Detect an unsupported-parameter rejection, strip `reasoning_effort`, retry once, and remember the provider+model combination so the field is not sent there again. Detection should cover the shapes that actually occur rather than litellm's exact string: litellm's `UnsupportedParamsError`, OpenAI's own `unsupported_parameter` error code, and the bare `unexpected keyword argument` TypeError that FastAPI/vLLM-style proxies surface.
2. Exclude that specific error class from the circuit breaker's failure count. Any other 400 — bad credentials, malformed payload, nonexistent model, context overflow — must still open the circuit.

Happy to send a PR for this.

DaBlitzStein:
Closes #7769

## The problem

Against a litellm gateway fronting an OpenAI-shaped model group, every turn of every affected agent fails with `400 litellm.UnsupportedParamsError: openai does not support parameters: ['reasoning_effort']`, and the agent then goes mute for the full provider cooldown.

The field is sent because `build_request` in `crates/librefang-llm-drivers/src/drivers/openai.rs` sets it whenever the echo policy is the default `ReasoningEchoPolicy::None` and the request carries a thinking budget (added in #6398).
`ThinkingConfig::default()` is `budget_tokens: 10_000`, which `reasoning_effort_for_budget` maps to `"medium"`, so any agent whose manifest carries a `[thinking]` section — including one inherited from its agent type with no field set — sends it on every request.

Reproduced by sending the same request twice to the same gateway: without `reasoning_effort` it answers, with it the call 400s.

## Why it cannot be fixed from a catalogue

Whether a *model* can reason and whether a *gateway* will forward the field are different questions, and only the second one decides the outcome.
litellm strips the parameter before the model sees it, and the error text says `openai does not support parameters` — naming the adapter, not the model.
The same model behind a different gateway accepts it.
So the driver has to discover it rather than look it up.

## Changes

- `crates/librefang-llm-driver/src/llm_errors.rs`: `is_unsupported_parameter_error` and the narrower `is_unsupported_reasoning_effort_error`. Detection matches the shapes that actually occur — litellm's `UnsupportedParamsError`, OpenAI's own `unsupported_parameter` error code, and the bare `unexpected keyword argument` TypeError that FastAPI/vLLM-style proxies surface — rather than litellm's exact string.
- `crates/librefang-llm-drivers/src/drivers/openai.rs`: on such a rejection, strip `reasoning_effort` and retry once, then record the model in `reasoning_effort_unsupported` so `build_request` omits the field for that model from then on. No second guaranteed round trip.
- `crates/librefang-runtime/src/agent_loop/retry.rs`: `should_count_against_circuit_breaker` excludes that one error class from `ProviderCooldown`. The rejection is deterministic and self-inflicted — it says nothing about whether the provider is reachable, and backing off does not change the outcome, since the next request after the cooldown carries the same field. Every other 400 still opens the circuit.

The retry.rs change is a fallback for the case where the driver's own in-place retry did not resolve it (its retry budget exhausted first on the same rejection); in the common path a call never reaches it.

## Verification

- `cargo test -p librefang-llm-driver --lib unsupported` — **4 passed**: `test_is_unsupported_parameter_error_litellm_shape`, `..._openai_code_shape`, `..._proxy_kwarg_shape`, `..._rejects_other_400s`. The last asserts that an invalid API key, a nonexistent model, malformed JSON and `context_length_exceeded` do **not** classify as parameter rejections.
- `cargo test -p librefang-llm-drivers --lib` — **637 passed, 0 failed**, including the two new ones: `unsupported_reasoning_effort_400_retries_without_it_and_succeeds` and `cached_rejection_omits_reasoning_effort_on_next_request`.
- `cargo test -p librefang-runtime --lib retry` — **7 passed**, including `unsupported_parameter_400_does_not_open_circuit_breaker` and `other_400_still_opens_circuit_breaker`.
- `cargo clippy -p librefang-llm-driver -p librefang-llm-drivers -p librefang-runtime --all-targets -- -D warnings` — **no findings in this PR's diff**. It does report two errors under Rust 1.98, both pre-existing on `origin/main` and both in files this PR does not touch: `clippy::drain_collect` at `crates/librefang-llm-drivers/src/drivers/bedrock.rs:538` and `clippy::map_or_identity` at `crates/librefang-llm-drivers/src/drivers/mod.rs:115`. Open PR #7762 fixes exactly those two files; until it lands this PR's clippy job will be red for reasons unrelated to it.

## Notes

- No overlap with any other open PR: I cross-checked the file list of all 40 open PRs against the three files here and found none in common.
- The `reasoning_effort_unsupported` set is per-driver-instance and in-memory, so the discovery is re-learned once after a restart. That is deliberate — a gateway's model group can be reconfigured to accept the field, and a persisted negative cache would keep suppressing it.

---

## CI status, for the record

The red lanes on this PR are not caused by it.

- `Quality` and `OpenAPI Drift` both fail on `error: unused import: librefang_runtime::kernel_handle::prelude::*` at `crates/librefang-kernel/src/kernel/messaging.rs:18`. That is #7754, and open PR #7755 is the one-line deletion that fixes it. Every kernel-touching PR is currently blocked behind it.
- The `Test / Unit` and `Test / Ubuntu (shard N/4)` lanes failed on a runner-side error before reaching any test: `bash function injection via BASH_FUNC_ environment variable is not allowed for security reasons` / `install-action requires bash`. That is infrastructure, not this diff.

Verification for the actual change was run locally and is listed above.


houko (changes requested):
P1: the circuit-breaker exemption is not restricted to the 400 response class this PR intends to handle. `should_count_against_circuit_breaker` stringifies every `LlmError` and checks only message substrings, so a 429/500 API response, transport error, or parse error whose text happens to contain `unsupported parameter` will also bypass failure accounting. Repeated real provider failures can therefore avoid opening the circuit indefinitely. The driver-side recovery correctly requires `status == 400`; the runtime fallback needs the same structural boundary.

Please pattern-match `LlmError::Api { status: 400, message, .. }` (and return `true` for every other variant/status), then add a regression using the same unsupported-parameter text with status 500 that proves the circuit still opens. The current `FixedErrorDriver.status` field makes that case straightforward, but both tests currently use 400 so the ignored status cannot be detected.

houko:
CI is green; this is one change away from mergeable.
`2537dc9f` is a merge from `main` only, so the CHANGES_REQUESTED from 2026-08-22 is still the sole blocker, and this is among the closest PRs in the repo to landing.

The one change is `crates/librefang-runtime/src/agent_loop/retry.rs:186`:

```rust
fn should_count_against_circuit_breaker(error: &LlmError) -> bool {
    !llm_errors::is_unsupported_parameter_error(&error.to_string())
}
```

`error.to_string()` flattens every variant into one haystack.
`LlmError` renders `Api` as `"API error ({status}): {message}"` and `Http` / `Parse` / `AuthenticationFailed` as `"HTTP error: {0}"` and friends (`crates/librefang-llm-driver/src/lib.rs:20-65`), each embedding raw provider text.
So a 500, or a transport error, whose body happens to contain any entry of `UNSUPPORTED_PARAMETER_SIGNALS` (`llm_errors.rs:784` — `"unsupported parameter"`, `"does not support parameters"`, `"unexpected keyword argument"`, …) also skips `record_retry_failure` and the circuit never opens.

The driver side already has the structural boundary this needs — `openai.rs:1402` and `openai.rs:1853` both gate on `status == 400 && oai_request.reasoning_effort.is_some() && is_unsupported_reasoning_effort_error(&body)`.
Mirror it in the runtime:

```rust
fn should_count_against_circuit_breaker(error: &LlmError) -> bool {
    match error {
        LlmError::Api {
            status: 400,
            message,
            ..
        } => !llm_errors::is_unsupported_parameter_error(message),
        _ => true,
    }
}
```

Matching on `message` directly also drops the `to_string()` allocation and keeps the `"API error (400): "` prefix out of the haystack.

For the regression: `unsupported_parameter_400_does_not_open_circuit_breaker` and `other_400_still_opens_circuit_breaker` both construct `FixedErrorDriver { status: 400, .. }`, so neither can observe that the status is currently ignored.
A third test reusing the first one's litellm `UnsupportedParamsError` message with `status: 500` and asserting `cooldown.get_state("litellm") == CircuitState::Open` closes that gap; `FixedErrorDriver` already takes `status` as a field, so it is a copy-paste of the existing body.

Nothing else in the diff needs to change — the `openai.rs` strip-and-retry, the `llm_errors.rs` signal list, and both call sites at `retry.rs:256` / `retry.rs:519` are fine as they stand.

DaBlitzStein:
Superseded by #7852.

Verified against `origin/main`: `is_unsupported_reasoning_effort_error` at `crates/librefang-llm-driver/src/llm_errors.rs:792` — the same function name this branch introduced — with the per-model cache `reasoning_effort_unsupported` on `OpenAIDriver` (`drivers/openai.rs:75`) and its own tests. The doc comment cites #7769, the issue this branch was opened against.

PR #7749, opened 2026-08-20
feat(dashboard): full manifest parity in the running-agent editor

DaBlitzStein:
Refs #7742.

### Read side: `GET /api/agents/{id}/manifest`

New endpoint returning the agent's whole `AgentManifest` as raw TOML, rendered from the live in-memory manifest rather than re-read from `agent.toml`, so it reflects state a partial PATCH has not flushed to disk yet.
This is the read half the editor needed: `getAgentDetail` returns a curated projection that drops most of the manifest, so there was nothing to pre-fill a full editor with.

### Editor: 11 fields to the whole manifest

`AgentsPage`'s Configure drawer previously exposed 11 editable fields.
It now mounts the existing `AgentManifestForm` — the same component the create flow uses — in edit mode, seeded from `GET .../manifest` and submitted as `manifest_toml` through the existing `PATCH /api/agents/{id}`.
Fields the form does not render are carried through `extras` rather than dropped, so editing a rendered field cannot silently reset an unrendered one.

### Channels section

`PUT /api/agents/{id}/channels` and its `GET` sibling have been registered for months with no client at all — the allowlist was only reachable by hand-editing `agent.toml`.
The drawer now has a Channels picker backed by both, with empty meaning "reachable from every configured channel".

### Tools tab reaches parity with Skills

The Skills tab has had a Customize button and per-item assignment for a while; Tools had neither.
Tools now gets the same visible Customize affordance and per-tool assignment from the Available list.
MCP server grants are queued into the same draft and saved through `PUT /agents/{id}/mcp_servers` on commit, so a grant toggled in the UI actually applies instead of rendering a read-only hint pointing at another tab.

### Kernel: standalone tags API (own bullet, per review)

`update_tags` keeps `entry.tags`, `manifest.tags` and the registry's `tag_index` in sync and is the runtime tags API #7742 lacked.
It is a standalone mutator, not wired into `update_manifest` — the manifest write path on main already routes through `replace_manifest_and_retag` (#8103), which owns that desync fix.
Concurrent tag writes for one agent were not serialized upstream, so the read of the previous tags and both index passes (retract with empty-bucket pruning, mirroring `remove()`; add) now run under a single guard: the `agents` shard guard is held across the `tag_index` maintenance, so two concurrent `update_tags` calls cannot interleave their index passes (review point 2 on this PR).
Lock ordering is `agents` → `tag_index`; no code path holds a `tag_index` guard while acquiring an `agents` guard.

### Query layer

[...]

PR #7773, opened 2026-08-21
fix(agents): accept a declared-but-not-connected MCP server or skill when saving an agent's allowlist

### LINKED ISSUE #7772: bug(agents): saving an agent's MCP allowlist 400s on an inherited declaration that is installed but not configured
## What happens

Editing any agent that carries an inherited MCP declaration fails from the dashboard with:

```
[400] Internal error: Unknown MCP server: fetch
```

naming a server the operator never asked for, and the edit is lost.

On the install where this was reported, the registry's agent types declare `fetch`, 25 of the 42 agents inherited it, and `fetch` is present in the local MCP catalog but not configured in `config.toml`.
The dashboard's save sends the whole `mcp_servers` array, so the inherited name took the request down with it and the two servers the operator was actually trying to add were never stored.

## Why

`set_agent_mcp_servers` in `crates/librefang-kernel/src/kernel/agent_state.rs` builds its set of acceptable names by walking the MCP tools **connected at that instant** and resolving each back to its server:

```rust
if let Ok(mcp_tools) = self.mcp.mcp_tools.lock() {
    let mut known_servers: HashSet<String> = HashSet::new();
    let configured_servers: Vec<String> = /* effective_mcp_servers */;
    for tool in mcp_tools.iter() {
        if let Some(s) = resolve_mcp_server_from_known(&tool.name, ...) {
            known_servers.insert(normalize_name(s));
        }
    }
    for name in &servers {
        if !known_servers.contains(&normalize_name(name)) {
            return Err(... Internal(format!("Unknown MCP server: {name}")));
        }
    }
}
```

That set is a strict subset of what is *configured*, never mind what is *installed*: a configured-but-not-yet-connected server is rejected, and a catalog entry that is installed but not configured is rejected.

Two further problems in the same place:

- The whole check is inside `if let Ok(mcp_tools) = ...lock()`, so a poisoned lock skips validation entirely rather than failing.
- The error is `LibreFangError::Internal`, which is why the dashboard says "Internal error" for what is a user-input problem. `set_agent_skills` has the same shape and the same `Internal` wording for `Unknown skill`.

## Why it is not just a validation bug

Declaring a server that is not installed on this instance is a legitimate state, not an error.
Agent types are shared artefacts: they declare what the agent *wants*, and any given install may not have all of it.
`spawn` accepts such a manifest — it performs no equivalent check — so today the edit path is stricter than the path that created the agent, and an agent can exist in a state its own editor refuses to save.
#7713 / #7716 approach the same underlying model from the read side, by surfacing those declarations as pending.

## Suggested fix

Accept a name that is either configured in `config.toml` (connected or not) or present in the locally cached MCP catalog (`~/.librefang/mcp/catalog/`).
A declaration that is not installed stays declared and inert — saving persists the name, it does not install, connect, or touch `config.toml`.
Reject only a name absent from both.

Skills need the same treatment against on-disk-but-unloaded directories (`SkillRegistry::unloaded_on_disk_dirs`, already used by the Stable-mode freeze-safe reload path). That criterion is narrower than the MCP one on purpose, because skill content is not part of `registry_sync`'s copy set, so there is no synced catalog to check against.

Both errors should become `InvalidInput` so the wording stops calling a user-input problem an internal one.

Happy to send a PR for this.

DaBlitzStein:
Closes #7772

## The problem

Editing any agent that carries an inherited MCP declaration fails from the dashboard with `[400] Internal error: Unknown MCP server: fetch`, and the edit is lost.

On the install where this was reported: the registry's agent types declare `fetch`, 25 of the 42 agents inherited it, and `fetch` is present in the local MCP catalog but not configured in `config.toml`.
The dashboard save sends the whole `mcp_servers` array, so the inherited name took the request down with it and the two servers the operator was actually adding were never stored.

`set_agent_mcp_servers` built its set of acceptable names by walking the MCP tools **connected at that instant** and resolving each back to its server.
That set is a strict subset of what is configured, never mind what is installed, so a configured-but-not-yet-connected server was rejected too.

## Changes

- `set_agent_mcp_servers` accepts a name that is configured in `config.toml` (connected or not) **or** present in the locally cached MCP catalog. Only a name absent from both is rejected.
- `set_agent_skills` gets the same treatment against on-disk-but-unloaded directories via `SkillRegistry::unloaded_on_disk_dirs` — the same "exists but not active" concept the Stable-mode freeze-safe reload path already uses. This criterion is deliberately narrower than the MCP one, because skill content is not part of `registry_sync`'s copy set, so there is no synced catalog to check against. The changelog fragment says this outright.
- Accepting a pending name **does not install, connect, or configure anything** — it only persists the allowlist entry. Asserted, not just claimed: see the effective-server-list check below.
- The MCP check sat inside `if let Ok(mcp_tools) = ...lock()`, so a poisoned lock skipped validation entirely rather than failing. The new form reads the configured list with `unwrap_or_default` and always runs the loop.
- Both errors were `LibreFangError::Internal`, which is why a user-input problem was reported as an internal one. They are now `InvalidInput`, and each message names the rejected entry and says what to do about it.

Declaring something this instance does not have is a legitimate state, not an error: `spawn` performs no equivalent check, so the edit path was stricter than the path that created the agent, and an agent could exist in a state its own editor refused to save.

## Relationship to #7716 / #7713

Adjacent, not overlapping, and worth stating so a reviewer does not have to work it out.
#7713 / #7716 handle the **read** side — surfacing declared-but-unavailable skills and MCP servers as pending on the detail endpoints.
This PR handles the **write** side — not rejecting those same declarations when the allowlist is saved.
They share the underlying model and are independently useful; neither depends on the other.

Concretely, this branch is built on `origin/main` only: I removed the assertions on the `pending` field from the tests below, because that field is #7716's addition and does not exist on `main`. The behaviour those assertions covered is asserted here instead by checking the effective server list directly.

## Verification

- `cargo test -p librefang-kernel --lib set_agent_` — **10 passed**, including the new `test_set_agent_mcp_servers_accepts_catalog_only_pending_name`, `test_set_agent_mcp_servers_rejects_name_unknown_everywhere`, `test_set_agent_mcp_servers_real_world_regression_pending_plus_new_configured_entries` and `test_set_agent_skills_accepts_on_disk_unloaded_pending_name`. The real-world regression test reproduces the reported case exactly — `memory` configured, `fetch` catalog-only, adding `camoufox` and `sequential-thinking` — and then asserts `fetch` was **not** added to `effective_mcp_servers` as a side effect of the save.
- `cargo test -p librefang-api --test agents_routes_integration` — **61 passed, 0 failed**, including four new end-to-end cases: `test_put_mcp_servers_accepts_pending_declaration_alongside_new_configured_entries`, `test_put_mcp_servers_rejects_unknown_name_with_400_not_internal_error`, `test_put_skills_accepts_on_disk_unloaded_pending_name`, `test_put_skills_rejects_unknown_name_with_400_not_internal_error`. The reject cases assert the status is 400 and that the message names the offending entry.
- `cargo clippy -p librefang-kernel -p librefang-api --all-targets -- -D warnings` — no findings in this PR's diff. Two pre-existing errors elsewhere remain, both outside it: `crates/librefang-llm-drivers/src/drivers/bedrock.rs:538` and `.../drivers/mod.rs:115`, fixed by open PR #7762.

### One caveat on how the above was run

`cargo test -p librefang-kernel` does not compile on `main` at all: `crates/librefang-kernel/src/kernel/messaging.rs:18` has an unused glob import and the workspace denies `unused_imports`, so the build fails before reaching any test (this is #7754). I applied that one-line deletion — the entire content of open PR #7755 — locally to get a build, and **reverted it before committing**. It is not in this branch. Until #7755 lands, this PR's kernel lanes will be red for that reason and not for anything here.

## Notes

- No overlap with anyone else's open PR. Cross-checking the file list of all 40 open PRs, the three files here appear only in our own stacked PRs (#7749, #7728, #7727, #7726, #7725, #7717, #7716, #7715), never in another contributor's.


houko (changes requested):
P1: the pending-skill check compares the allowlist value with the directory name, but skill allowlists and the loaded registry use `[skill].name` from `skill.toml`. `SkillRegistry::unloaded_on_disk_dirs()` intentionally returns directory names; its own documentation notes that a manifest name may differ from its directory name. For an unloaded directory `package-dir/skill.toml` containing `name = "actual-skill"`, this code rejects the valid `actual-skill` declaration and accepts `package-dir`, which will no longer match anything after restart when the skill loads under `actual-skill`.

Please validate pending entries using the parsed manifest skill name rather than `unloaded_on_disk_dirs()`, and add a regression where the directory name differs from `[skill].name`. The current tests use the same value for both and therefore cannot detect this contract mismatch.

houko:
Correction to my earlier read: CI is green, but this one is now two changes away, not one — it picked up a merge conflict after the review.

**1. Conflict in `crates/librefang-kernel/src/kernel/agent_state.rs`.**
`mergeable` is `CONFLICTING`. The only commit to touch that file since your merge base is #7807 (`0257706`, "fix: serialize MCP reloads and preserve agent manifests"), which refactored the same `agent_manifest_path` / `manifest_write_lock` / `persist_full_manifest_at` helpers your branch relocated.
A test merge conflicts only there; `kernel/tests.rs` auto-merges clean.
Resolve toward #7807's versions of those three helpers and replay only your `set_agent_skills` / `set_agent_mcp_servers` validation changes on top — the helper movement in your diff is no longer needed.

**2. The standing review is unaddressed.** `3d33905` is a merge from `main` only.

`agent_state.rs:472-475`:

```rust
let loaded = registry.skill_names();
let on_disk_unloaded = registry.unloaded_on_disk_dirs();
for name in &skills {
    if loaded.iter().any(|n| n == name) || on_disk_unloaded.iter().any(|n| n == name) {
```

Those two accessors return different kinds of identifier.
`skill_names()` (`crates/librefang-skills/src/registry.rs:517`) returns `self.skills.keys()`, and that map is keyed by the manifest name — `registry.rs:400` takes `let name = manifest.skill.name.clone();` and `registry.rs:407` inserts under it.
`unloaded_on_disk_dirs()` (`registry.rs:525`) pushes `path.file_name()`, the **directory** name, and its own doc comment at `registry.rs:521` says the distinction is real: "compared by path, so a skill whose manifest name differs from its directory name is matched correctly".

So for `skills/package-dir/skill.toml` containing `name = "actual-skill"`, line 475 rejects the correct `actual-skill` and accepts `package-dir` — which then matches nothing once the skill loads and lands in the registry under `actual-skill`.

Fix: add a sibling accessor next to `unloaded_on_disk_dirs()` that reads each unloaded directory's `skill.toml` and yields `manifest.skill.name`, and validate against that instead.
The directory walk, the `.`-prefix staging skip, and the canonicalize-before-compare handling are all already in that function; it only needs the manifest parse before the `out.push(name.to_string())` at `registry.rs:551`.
Keep `unloaded_on_disk_dirs()` as-is — `tools_and_skills.rs:433` wants directory names for its operator-facing "will not be picked up without a restart" report.

For the regression: `install_test_skill` (`kernel/tests.rs:3322`) derives both the directory (`skills_parent.join(name)`) and `[skill] name = "{name}"` from a single argument, so `test_set_agent_skills_accepts_on_disk_unloaded_pending_name` cannot detect this.
Add an `install_test_skill_named(parent, dir_name, manifest_name)` variant (or a fourth parameter), install with the two deliberately different, and assert the manifest name is accepted — plus that the directory name is not, since it can never match after load.

The MCP half of the PR is unaffected by both points.

PR #7778, opened 2026-08-21
fix(providers): write a provider overlay the catalog can read back

### LINKED ISSUE #7776: bug: the provider overlay is written in a shape the reader rejects, so live model discovery never survives a restart
## What happens

Toggling live model discovery for a custom provider returns `200`, writes to disk, and is discarded on the next boot.

`PUT /api/providers/litellm/discovery {"discover_models": true}` → `200`.

On disk, `~/.librefang/providers/litellm.toml`:

```toml
[provider]
id = "litellm"
discover_models = true
```

At boot:

```
WARN provider catalog TOML ignored: parse failed e=TOML parse error at line 1, column 1
1 | [provider]
```

And the API then reports `discover_models: false`, `model_count: 0`, while the gateway itself is configured, authenticated and answering — its `/v1/models` lists five models.

**The writer and the reader disagree on the file's shape.** The handler emits a `[provider]` section; the loader rejects it at line 1, column 1. Nothing surfaces to the operator: the toggle reports success, the warning goes to the log, and the setting is silently gone after the next restart.

## Why it matters beyond this one flag

This is the mechanism behind #7775 as reported. The feature that fixes that issue already exists — #6702 added `PUT /api/providers/{name}/discovery` precisely so a custom OpenAI-compatible endpoint would stop sitting at `model_count: 0`. Its own doc comment says so:

```
/// Model discovery used to be reachable only through the hard-coded local
/// provider ids, so a custom OpenAI-compatible endpoint registered under its
/// own id stayed at `model_count: 0` forever with no UI action to change that.
```

The operator did use it. It returned 200. It just does not survive.

Whatever else the provider overlay directory is meant to carry is presumably lost the same way, since the failure is in parsing the file at all rather than in one field.

## Expected

The reader accepts what the writer produces. A round-trip test — write through the handler, read through the loader — would have caught it, and is the thing most worth adding here.

Two further points, both about the same silence:

- A parse failure of an operator-written file should not be a `WARN` the operator never sees. If a setting they changed is being discarded, that belongs in the interface, next to the setting.
- The endpoint returns `200` for a change that will not survive a restart. It cannot know that at the time, which is an argument for the round-trip being verified rather than assumed.

## Related

#7775 — models from an OpenAI-compatible gateway never appear in the list. This is why the existing remedy for it does not hold.

DaBlitzStein:
Fixes #7776.

## The bug

`PUT /api/providers/{name}/discovery` persists the opt-in into `~/.librefang/providers/{name}.toml`.
When that file did not exist yet — the case this endpoint exists to serve, a gateway that lives only in `[provider_urls]` — the writer created it with `id` and the flag and nothing else:

```toml
[provider]
id = "litellm"
discover_models = true
```

`ProviderCatalogToml` required `display_name`, `api_key_env` and `base_url` with no `#[serde(default)]`, so `toml::from_str::<ModelCatalogFile>` failed on that file and the loader in `librefang-runtime/src/model_catalog.rs` discarded **the entire file** with a warning that did not even name it:

```
WARN provider catalog TOML ignored: parse failed e=TOML parse error at line 1, column 1
```

Observed on a live instance: the operator enabled discovery from the dashboard, the `PUT` returned 200, the file was written — and `discover_models` was back to `false` after every restart, with the gateway's five models still reported as zero.
The endpoint reported success for a setting that could never be read back.

## What changes

Both halves are fixed, and both are load-bearing.
The writer fix is what stops this from happening again to anyone who enables discovery from the dashboard.
The reader fix is what rescues anyone who already has the broken three-line file sitting on disk from an earlier build — the reporting operator has since repaired theirs by hand, into exactly the complete shape the reader accepts, but nobody else's is repaired and nothing tells them to look.
It also closes the same silent trap for anyone editing the TOML by hand.

**Writer** (`routes/providers.rs`)

- `upsert_provider_discover_models` now takes the live `ProviderInfo` the handler already resolved (the same lookup that decides 200 vs 404) instead of just the name, and writes `id`, `display_name`, `api_key_env` and `base_url` from it, plus `key_required = false` when the provider genuinely needs no key.
- Each key is written only when the file does not already have it, so the operator's own value always wins over the in-memory one and a hand-maintained catalog file is not rewritten by an unrelated toggle.
- A file left behind by an older build is completed in place by the next toggle rather than rewritten in the same unreadable shape.

**Reader** (`librefang-types/src/model_catalog.rs`, `librefang-runtime/src/model_catalog.rs`)

- `display_name`, `api_key_env` and `base_url` gain `#[serde(default)]`. Conversion to `ProviderInfo` back-fills `display_name` from `id` and `api_key_env` from a new shared `default_api_key_env` helper — the same `{ID}_API_KEY` derivation that `set_provider_url` and `provider_probe` were already open-coding, so both now call the helper instead of carrying their own copy.
- `base_url` is allowed to be empty. That is safe for the `[provider_urls]` case because `apply_url_overrides` runs after the catalog is loaded (`kernel/boot.rs`: load at 1076, overrides at 1105) and fills the endpoint in.
- Two files declaring the same provider id previously produced two `ProviderInfo` entries, and `get_provider` returned whichever `read_dir` happened to yield first. They now merge into one entry, and the merge runs on the raw `ProviderCatalogToml` rather than the converted `ProviderInfo`, because conversion back-fills the omitted fields and after that step "absent" is indistinguishable from "explicitly the default" — a partial overlay would then silently overwrite the operator's own `api_key_env`. The invariant the merge guarantees is that an absent value never overwrites a present one; `key_required` and `discover_models`, which cannot distinguish absent from defaulted, are combined so the result does not depend on `read_dir` order.
- The parse-failure warning now carries `path`. That is exactly the datum whose absence made this take an hour to diagnose.

## Note on `is_custom`

`ProviderInfo.is_custom` is derived by the loader from whether the file's name also appears in `registry/providers/`.
For a provider that previously existed only via `[provider_urls]` — where `set_provider_url` marks it `is_custom = true` unconditionally — writing a readable file moves that decision to the classifier.
In the normal case the answer is the same (`litellm.toml` is not registry-shipped, so `is_custom` stays `true`).
In the documented degraded case where `registry/providers/` is missing or unreadable, the classifier falls back to `is_custom = false` for every provider, so this entry would lose its dashboard Delete control until the registry cache is restored.
That fallback and its rationale predate this change; the change only widens the set of entries it applies to by one. Flagging it rather than special-casing it, since reversing it would mean reworking the tri-state classification itself.

## Tests

`crates/librefang-types/src/model_catalog.rs`

- `partial_provider_record_deserializes_and_backfills_identity` — the exact legacy shape parses, and the back-fills produce `litellm` / `LITELLM_API_KEY`.
- `complete_provider_record_keeps_every_declared_value` — the fallbacks never disturb a full record.
- `default_api_key_env_uppercases_and_underscores`.

`crates/librefang-runtime/src/model_catalog/tests.rs`

- `legacy_partial_overlay_still_yields_the_provider_and_its_flag` — backward compatibility with files already on disk.
- `a_partial_overlay_never_clobbers_a_full_record_for_the_same_id` — asserted under both source orderings, because `read_dir` order is arbitrary and an order-dependent merge would be a heisenbug in production.
- `a_syntactically_invalid_catalog_file_is_skipped_without_taking_the_rest_down` — genuinely corrupt TOML is still dropped with a warning and does not cost the operator the other providers.
- `provider_overlay_round_trips_through_the_directory_loader` — end to end through the filesystem.

`crates/librefang-api/src/routes/providers.rs`

- `a_freshly_created_provider_file_round_trips_through_the_catalog_loader` — the test that would have caught this: write the overlay, then load the catalog from that directory and assert the record comes back with the flag and every identity field.
- `a_keyless_provider_records_key_required_false`.
- `upserting_the_flag_leaves_the_rest_of_the_file_byte_identical` — a file with comments and a `[[models]]` array is unchanged apart from the added flag, and the operator's `display_name` / `api_key_env` / `base_url` still win after reload.
- `upserting_over_a_legacy_partial_file_completes_the_record`.

`crates/librefang-api/tests/providers_routes_test.rs`

- `set_provider_discovery_flips_flag_and_persists_to_the_provider_file` gains a reload through `ModelCatalog::new_from_dir`. The test previously asserted only that the file contained `discover_models = true`, which was true of the broken file too — that is precisely why the bug shipped.

Each regression test was verified to fail without the fix: reverting only the three `#[serde(default)]` attributes makes `legacy_partial_overlay_still_yields_the_provider_and_its_flag` fail with "a partial overlay must not take the whole file down with it".

## Verification

Run with `CARGO_TARGET_DIR` pointed at a shared target directory, on rustc 1.98.0.

- `cargo clippy -p librefang-types -p librefang-runtime -p librefang-kernel -p librefang-api --all-targets -- -D warnings` — exit 0, zero findings in the changed code.
- `cargo test -p librefang-types --lib` — 909 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2291 passed, 2 failed.
- `cargo test -p librefang-api --lib` — 1033 passed, 1 failed.
- `cargo test -p librefang-api --test providers_routes_test` — 59 passed, 0 failed.

The three failures are pre-existing process-global races in untouched code, all of which pass when run in isolation and none of which CI sees, because nextest gives every test its own process:

- `channel_bridge::tests::channel_approval_rejects_replayed_totp_code` fails on `Decryption failed: aead::Error` — the vault-key-per-process race described in `changelog.d/fixed/7707-vault-key-test-race.md`.
- `agent_context::tests::poisoned_cache_lock_recovers_cached_context` deliberately poisons a process-global lock.
- `chatgpt_oauth::tests::test_callback_server_releases_listener_after_timeout` fails with `AddrInUse` on a fixed port.

Two caveats about the local toolchain, neither caused by this branch:

- `cargo check --workspace --lib` does not compile on `origin/main` at rustc 1.98.0: `crates/librefang-kernel/src/kernel/messaging.rs:18` trips a deny-level `unused_imports`. Verified by running the check on the unmodified base commit.
- The workspace-wide clippy run additionally trips four lints new in 1.98 (`chunks_exact_to_as_chunks`, `drain_collect`, `map_or_identity`, `result_large_err`) in `librefang-memory`, `librefang-llm-drivers` and `routes/terminal.rs`. The clippy run above allows exactly those five pre-existing lints and nothing else; every other lint remains denied and the changed code produces none.

Both are the known Rust 1.98 breakage — 1.98 landed on 20 August and the workspace sets `warnings = "deny"` — and are fixed by #7762.
Deliberately not bundled here.
If the clippy lane on this PR goes red on one of those lints rather than on anything in the diff, that is the cause.


houko (approved):
The writer/reader round trip is consistent with the boot sequence, and the duplicate-provider merge covers the real filename/id mismatch used by registry providers. I found no blocking issues.

Local verification: the partial-record types regression passed 1/1, the duplicate-id merge regression passed 1/1, `cargo fmt --all -- --check` passed, and `git diff --check origin/main...HEAD` passed. The API regression build was blocked before reaching the test body by the pre-existing unused import in `crates/librefang-kernel/src/kernel/messaging.rs:18` tracked in #7754; that blocker is outside this diff and is covered by #7791.

houko:
Re-verified against the current head `e6df7f23534ba54bd1c6a21173581e9a9348d32d`, and this one is genuinely done.

- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.
- My approval is recorded against that exact SHA, not an earlier one.
- Check rollup on run 32572454802 (`head_sha` = `e6df7f2`, conclusion `success`): zero failures. Every non-skipped lane is green, including `Quality`, `Test / Unit (lib+bin)`, all four `Test / Ubuntu` shards, `OpenAPI Drift` and `Live Integration Smoke`.

Nothing is blocking it on its own merits, so it can land as-is.

One scheduling fact, so we do not clobber each other.
My #7816 (`fix/7775-openai-compatible-discovery`) edits the same two files:

- `crates/librefang-api/src/routes/providers.rs`
- `crates/librefang-api/tests/providers_routes_test.rs`

It deliberately stays clear of `upsert_provider_discover_models`, where your hunks live.
#7816's hunks in `providers.rs` are at `list_models` (`@@ -303,6 +303,9 @@`), `probe_failure_downgrades_auth` (`@@ -794,6 +797,92 @@`) and `attach_probe_result` (`@@ -812,31 +901,7 @@`); yours are at `set_provider_discovery` (`@@ -2009,19 +2009,24 @@` and `@@ -2041,33 +2046,53 @@`) and the `mod tests` block (`@@ -3353,6 +3379,129 @@`).
No line overlap in that file at all.
The only real collision is that we both append cases to the end of `providers_routes_test.rs`.

**So: #7778 merges first, and I rebase #7816 onto it.**
No action needed from you, and specifically please do not rebase or force-push this branch to tidy it up — the approval and the green run are both pinned to `e6df7f2`, and a new head throws both away for a PR that is already clean.

Issue #3731, opened 2026-04-27
validate_ws_origin is dead code — agent_ws upgrade never enforces Origin


houko:
**Location**
- `crates/librefang-api/src/ws.rs:151` (definition)
- `crates/librefang-api/src/ws.rs:303` (`agent_ws` upgrade — never calls it)

**Problem**
`validate_ws_origin` exists, has tests, but a `grep validate_ws_origin` finds no production caller. The `agent_ws` upgrade handler accepts any Origin once the bearer token / session check passes. Cookie scope `Path=/dashboard` mitigates a pure cross-site CSWSH (browsers won't auto-send the cookie), **but** any same-origin XSS (see CSP issue) trivially reads `localStorage.librefang-api-key` and opens an authenticated WS to anywhere.

```rust
pub async fn agent_ws(ws: WebSocketUpgrade, ...) -> impl IntoResponse {
    // token validation, no origin check ↓
    ws.on_upgrade(move |socket| handle_agent_ws(...)).into_response()
}
```

**Impact**
Defense-in-depth gone. WS streams of agent output (potentially with sensitive content) accept any origin once token is valid.

**Fix**
Call `validate_ws_origin(&headers, listen_port, &cfg.allowed_origins, allow_remote)` at the top of `agent_ws`; reject 401 on disallowed origins.

---
*Filed via automated multi-perspective audit (security / concurrency / architecture / error-handling / frontend / performance subagents). Curated against #3359–#3409 to avoid duplicates.*

PR #7779, opened 2026-08-21
fix(api): accept the server's own origin on the agent WebSocket

### LINKED ISSUE #7777: WebSocket upgrade rejects the dashboard the daemon itself serves
## Summary

A daemon that serves its own dashboard refuses that dashboard's WebSocket.

On a LAN deployment (`api_listen = "0.0.0.0:4545"`, reached at `http://192.168.1.161:4545`) every agent-chat WebSocket upgrade comes back `403` and the daemon logs:

```
WARN request:http_request: WebSocket upgrade rejected: Origin validation failed
  reason=Origin http://192.168.1.161:4545 not in allowed list
  method=GET path=/api/agents/<id>/ws
```

The browser fetched the page from that very origin, so the product's primary use case — open the dashboard, talk to an agent — is broken out of the box on any deployment that is not loopback.
Agent replies never reach the chat pane.

## Root cause

Three separate things compound.

**1. The server's own origin is not accepted.**
`validate_ws_origin` (`crates/librefang-api/src/ws.rs`) auto-allows only *loopback* hosts (`localhost` / `127.0.0.1` / `::1`) on the listen port.
An origin identical to the host and port the request itself arrived on — which is exactly what the daemon's own dashboard sends — has no branch and falls through to the final `Err`.
The CORS layer does not have this gap: `server.rs` seeds its allow list with `format!("http://{listen_addr}")`.

**2. The wildcard is unreachable on this route.**
`validate_ws_origin` does implement `"*"`, but gates it on its `allow_remote` argument, and the agent-chat upgrade derives that argument from the `LIBREFANG_ALLOW_NO_AUTH` environment variable — an auth switch, unrelated to origins:

```rust
let allow_remote = std::env::var("LIBREFANG_ALLOW_NO_AUTH")
    .map(|v| matches!(v.trim(), "1" | "true" | "TRUE" | "yes" | "on"))
    .unwrap_or(false);
validate_ws_origin(&headers, listen_port, &cfg.cors_origin, allow_remote)
```

So `cors_origin = ["*"]` is honoured verbatim by CORS and silently ignored by the WebSocket.

**3. `allowed_origins` is not the knob it looks like.**
The operator had set `allowed_origins = ["*"]`, which is a real field — but it belongs to `TerminalConfig` and feeds the *terminal* route only.
The agent-chat WebSocket reads the top-level `cors_origin`.
A `[api] allowed_origins = …` table is dropped entirely, because `KernelConfig` has no `api` field (the v1→v2 migration hoists only `api_key` / `api_listen` / `log_level` out of `[api]`, and does not re-run once `config_version = 2`).

**4. The rejection message cannot be acted on.**
It names the origin it refused but not the list it consulted or where that list came from — which is the one thing that distinguishes "typo in my allow list" from "my allow list was never read".
The `403` also carries an empty body, so the browser console shows only a failed handshake.

## Expected

- An Origin equal to the scheme/host/port the request arrived on is always accepted, list or no list. A dashboard served by this daemon must never be rejected by it.
- `cors_origin = ["*"]` accepts any origin on the WebSocket, as it already does for CORS.
- A concrete allow list still rejects everything outside it (the cross-site WebSocket hijacking guard from #3731 must not regress).
- The rejection names the effective list and its config key.

## Environment

LibreFang daemon on `0.0.0.0:4545`, dashboard opened at `http://192.168.1.161:4545`, Chrome.
Same behaviour with `cors_origin` unset and with `allowed_origins = ["*"]` under `[terminal]`.

DaBlitzStein:
Fixes #7777.

## The failure

A daemon that serves its own dashboard refuses that dashboard's WebSocket.
Observed live on `0.0.0.0:4545` reached at `http://192.168.1.161:4545`:

```
WARN request:http_request: WebSocket upgrade rejected: Origin validation failed
  reason=Origin http://192.168.1.161:4545 not in allowed list
  path=/api/agents/<id>/ws
```

Every agent reply is lost, because the chat pane never gets a socket.
Any deployment that is not loopback hits this with the shipped defaults.

## Diagnosis

`validate_ws_origin` (`crates/librefang-api/src/ws.rs`) auto-allowed only *loopback* hosts (`localhost` / `127.0.0.1` / `::1`) on the listen port.
An Origin identical to the host and port the request itself arrived on — what the daemon's own dashboard always sends — had no branch and fell through to the final `Err`.
The CORS layer has no such gap: `server.rs` seeds its allow list with `format!("http://{listen_addr}")`.

Two things stopped the operator from working around it.

The wildcard was unreachable on this route: `validate_ws_origin` implements `"*"` but gates it on its `allow_remote` argument, and the agent-chat upgrade derived that argument from the `LIBREFANG_ALLOW_NO_AUTH` environment variable — an auth switch, unrelated to origins.
So `cors_origin = ["*"]` was honoured verbatim by CORS and silently ignored by the WebSocket.

And `allowed_origins` is not the knob it looks like: it belongs to `TerminalConfig` and feeds the terminal route only, while the agent-chat WebSocket reads the top-level `cors_origin`.
An `[api] allowed_origins = …` table is dropped entirely — `KernelConfig` has no `api` field, and the v1→v2 migration hoists only `api_key` / `api_listen` / `log_level` out of `[api]` and does not re-run once `config_version = 2`.

## Changes

- `validate_ws_origin` accepts an Origin whose host and port equal the request's own `Host` header.
  A browser sets `Origin` and `Host` independently and a page cannot forge `Origin`, so the two coincide exactly when this daemon served the page.
  A cross-site attacker reaching the same daemon sends its own `Origin` next to our `Host`, so they differ and the check does not fire — the hijacking guard from #3731 is intact, and a test pins it.
  The scheme is not compared, because `Host` carries none and behind a TLS-terminating proxy the browser's `https` cannot be reconstructed from the plaintext hop; the origin's own scheme supplies the default port so `Host: dash.example.com` matches `Origin: https://dash.example.com`.
  `X-Forwarded-Host` is deliberately ignored — it is attacker-controlled unless the peer is a verified proxy, and this function has no view of the peer.
- The agent-chat upgrade passes `allow_wildcard = true`, so `cors_origin = ["*"]` works on the WebSocket as it already did for CORS.
  The parameter is renamed from `allow_remote` to `allow_wildcard` to say what it actually gates.
  The terminal route still passes `terminal.allow_remote`, unchanged: an unauthenticated remote shell is worth a second explicit opt-in.
- A `"*"` entry no longer aborts the scan of the rest of the list.
  It is not a URL, so `Url::parse` failed on it and returned `Invalid extra origin URL: *`, making every entry after the wildcard unreachable whenever the wildcard branch itself was declined.
- The rejection is diagnosable.
  The message names the request's own `Host`, the listen port and the effective allow list, rendering an empty one as `<empty>` — which is exactly what a misplaced `allowed_origins` key produces, and the one fact that separates "typo in my list" from "my list was never read".
  Both call sites add `config_key`, `allowed_origins` and `listen_port` as structured WARN fields.
  The 403 now carries a body naming the refused origin and the `cors_origin` knob; it deliberately omits the list's contents, because this handler is reachable before any credential on a daemon with no auth configured.

## Verification

`crates/librefang-api/tests/ws_origin_validation_test.rs` — 8 integration tests against a real axum router, raw-socket handshakes so `Host` and `Origin` are both under the test's control:

- `server_own_origin_is_accepted_with_no_allowed_origins_configured` — the reported failure, reduced
- `wildcard_accepts_an_unrelated_origin`
- `concrete_list_still_rejects_an_origin_outside_it`
- `listed_origin_is_accepted`
- `a_mixed_list_containing_the_wildcard_still_accepts`
- `rejection_body_names_the_origin_and_the_config_key` — and asserts the list does *not* leak
- `a_cross_site_origin_reaching_the_same_host_is_still_rejected`
- `a_non_browser_client_without_an_origin_is_accepted`

Plus 12 unit tests in `ws::tests` covering the self-origin rule and the list scan directly (LAN IP, non-listen port, TLS proxy with a portless `Host`, IPv6 literal, malformed `Host`, non-http scheme, cross-site with our `Host`, same host on another port, the wildcard-scan pair, and the two message-content assertions).

### One integration test was renamed because it did not test what its name claimed

The count is 8 before and after, so this is worth stating rather than leaving a reviewer to assume nothing moved.

`wildcard_entry_does_not_abort_the_scan_of_a_mixed_list` claimed to pin the `"*"`-aborts-the-scan regression. It did not, and would have passed with that bug still in place.
`validate_ws_origin` declines the wildcard branch only when `allow_wildcard` is false; this route passes `true`, so a list containing `"*"` returns `Ok` before the scan is ever reached.

It is now `a_mixed_list_containing_the_wildcard_still_accepts`, with a comment stating explicitly what it does not cover and where the real coverage is.
The scan is pinned where it is actually reachable — at `allow_wildcard = false` — by `ws::tests::validate_ws_origin_wildcard_entry_does_not_abort_the_list_scan` and `ws::tests::validate_ws_origin_wildcard_entry_does_not_poison_the_rejection_reason`.

That second one covers the half of the bug that costs an operator the most time: when the scan aborted on `"*"`, an origin that genuinely *should* be rejected was rejected with `Invalid extra origin URL: *` — a message naming neither the origin nor the list, sending the operator hunting a config typo that does not exist.

`allow_wildcard = false` with a `"*"` in the list is not a laboratory combination.
It is the terminal route with `terminal.allow_remote = false` and a `"*"` an operator left in `terminal.allowed_origins` — and the operator who reported #7777 already has `allowed_origins = ["*"]` written in their `config.toml` today.
It is inert only because it sits under `[api]`, a table `KernelConfig` has no field for and therefore discards whole.
The moment anyone moves it to the correct place with `allow_remote = false`, they land squarely on the broken path.

```
cargo test -p librefang-api --lib                                1042 passed
cargo test -p librefang-api --test ws_origin_validation_test     8 passed
cargo test -p librefang-api --test hash_only_master_key_ws_auth  4 passed
cargo clippy -p librefang-api --all-targets -- -D warnings        exit 0
cargo check --workspace --lib                                     exit 0
```

The gates above ran on the exact tree that is committed, `rustfmt` included.

One unrelated flake was seen once on the first `--lib` run and did not reproduce: `channel_bridge::tests::channel_approval_rejects_replayed_totp_code`, failing on `LIBREFANG_VAULT_KEY` with `Vault init succeeded on disk but the freshly-written file cannot be decrypted`.
It passes in isolation and passed on the immediate full-suite retry (1042/1042).
That is the process-global vault-key race #7707 addressed; it appears not to be fully closed for the in-process `--lib` runner, which is the contributor-facing command in CLAUDE.md. Flagged here rather than filed, since it is one observation in a different domain to this PR.

Note for CI: `main` is currently red on rustc 1.98 for reasons unrelated to this branch (an `unused_imports` deny in `librefang-kernel/src/kernel/messaging.rs:18`, and `chunks_exact_to_as_chunks` / `drain_collect` / `map_or_identity` / `result_large_err` in `librefang-memory`, `librefang-runtime`, `librefang-llm-drivers` and pre-existing signatures in `librefang-api/src/routes/terminal.rs`).
Those were neutralised locally to run the gate above and are **not** part of this diff.


houko (changes requested):
P1: `Host`/`Origin` equality does not prove that the dashboard was served by this daemon. The `Host` header is selected by the request URL and is not trusted here. With DNS rebinding, an attacker can serve a page from `attacker.example`, rebind that name to the LAN daemon, and the browser will send `Host: attacker.example:4545` together with `Origin: http://attacker.example:4545`. `origin_is_self` accepts that pair with an empty `cors_origin`, bypassing the cross-site WebSocket guard. The new `validate_ws_origin_accepts_own_origin_behind_a_tls_proxy` test demonstrates the unsafe primitive by accepting an arbitrary hostname solely because these two attacker-controlled values match.

Please remove arbitrary hostname equality from the implicit allow path. The reported LAN case can be fixed without opening this boundary by auto-allowing matching literal IP hosts (with the required port handling), while deployments using hostnames or TLS proxies should list their public origin explicitly in `cors_origin` or another trusted-host allowlist. Please also add a regression test showing that matching `Host` and `Origin` for a non-loopback hostname is rejected unless that origin is explicitly configured.

houko:
CI is green; this is one change away from mergeable, and `4d52d65` is a merge from `main` only, so the security objection is the sole remaining blocker.

**Why `Host == Origin` is not evidence that this daemon served the page.**
The doc comment at `crates/librefang-api/src/ws.rs:338` states the premise: "A browser sets `Origin` and `Host` independently and a page cannot forge `Origin`, so the two coincide exactly when the page was served by this daemon".
The first clause is true; the conclusion does not follow, because `Host` is not something this daemon asserted — it is derived from whatever URL the page dialled, and the page picks that.
Concretely: an attacker registers `attacker.example`, serves a page from it, then rebinds that name's DNS to the daemon's address (`127.0.0.1` or the LAN IP — DNS rebinding).
The browser's subsequent request to `http://attacker.example:4545` carries `Host: attacker.example:4545` alongside `Origin: http://attacker.example:4545`.
Nothing is forged: the page genuinely was served from that origin.
`origin_is_self` (`ws.rs:354`) compares the two, finds them equal, and `validate_ws_origin` returns `Ok` at `ws.rs:250` with `extra_origins` empty — the #3731 cross-site WebSocket guard is bypassed.
`validate_ws_origin_accepts_own_origin_behind_a_tls_proxy` (`ws.rs:2943`) is that exact request with `dash.example.com` in place of `attacker.example`; nothing in the function distinguishes the two, which is why the test doubles as a demonstration of the hole.

**Worth knowing before you edit:** the two unit tests that look like they already cover this do not.
`validate_ws_origin_lan_ip_same_port_rejected` (`ws.rs:2803`) and `validate_ws_origin_arbitrary_host_same_port_rejected` (`ws.rs:2812`) insert only the `origin` header, so `origin_is_self` returns early at the `headers.get("host")` guard on `ws.rs:360` before comparing anything.
A real browser always sends `Host`, so under actual traffic both of those origins are now accepted despite the test names.

**The narrowing, which keeps the reported LAN fix.**
Restrict the implicit allow to matching **literal IP** hosts.
`url` is already a direct dependency (`crates/librefang-api/Cargo.toml:60`) and `Url::host()` returns `url::Host`, so no new dependency is needed: change `origin_is_self` to take the origin's `url::Host` instead of `origin_host: &str` — the caller already holds `parsed` at `ws.rs:232` and can pass `parsed.host()` — then require both that value and the `Host`-header side to be `Host::Ipv4(_) | Host::Ipv6(_)` before comparing.
Keep the existing port comparison and the `default_port_for_scheme` bridge exactly as they are; they are what make the IPv6 and non-listen-port cases work.

Survives unchanged:

- `validate_ws_origin_accepts_the_servers_own_lan_origin` (`ws.rs:2926`, `192.168.1.161:4545`) — the reported defect from #7777.
- `validate_ws_origin_self_rule_matches_ipv6_literal_host` (`ws.rs:2968`, `[fd00::1]:4545`).
- Loopback, which `normalize_origin_host` (`ws.rs:391`) already folds to `localhost` on the separate `listen_port` rule.
- All of `crates/librefang-api/tests/ws_origin_validation_test.rs` — `SERVER_HOST` is `"192.168.1.161:4545"` (`:35`), an IP literal, so the integration coverage needs no edits.

Needs inverting to rejection-plus-explicit-`cors_origin` pairs:

- `validate_ws_origin_accepts_own_origin_behind_a_tls_proxy` (`ws.rs:2943`, `dash.example.com`).
- `validate_ws_origin_accepts_own_origin_on_a_port_other_than_listen_port` (`ws.rs:2934`, `dash.internal:8443`).

Hostname and TLS-proxy deployments list their public origin in `cors_origin` instead; that is already the documented answer for the `X-Forwarded-Host` case at `ws.rs:350-353`, so it is the same rule rather than a new burden.

Also in scope while you are here: add the regression asking for matching `Host` and `Origin` on a non-loopback **hostname** with `extra_origins` empty to be `Err`, and the same pair with that origin listed to be `Ok` (the `browser_headers` helper at `ws.rs:2918` builds both headers already); update rule 1 in the `validate_ws_origin` doc list at `ws.rs:208-214`; rewrite the `origin_is_self` doc comment at `ws.rs:335-353`, which currently asserts the unsafe premise; and give the two host-less tests at `ws.rs:2803` / `ws.rs:2812` an explicit `host` header so they test what their names claim.

DaBlitzStein:
Superseded by #7850.

Verified against `origin/main`: every production symbol this branch introduced is present, including the self-origin rule on the agent WebSocket. Only the test names differ.

Issue #7774, opened 2026-08-21
feat: model configuration, context window included, must be editable at any time from the WebUI, the TUI and the API


DaBlitzStein:
## What is missing

A model's context window cannot be corrected after the model exists.

`ModelOverrides` (`crates/librefang-types/src/model_catalog.rs:375-390`) carries `model_type`, `temperature`, `top_p`, `max_tokens` and `frequency_penalty`. It does not carry `context_window`.

The dashboard says so itself, at `crates/librefang-api/dashboard/src/pages/ProvidersPage.tsx:166`:

```
// Uses max_tokens override (not context_window): context_window has no persistence path
// in ModelOverrides and is overwritten on registry sync.
```

The provider wizard does offer a context field when a model is first registered (`ProvidersPage.tsx:1345`, defaulting to 128000), so the value can be set once, at creation, and never again. There is no path to correct it, and a registry sync overwrites it.

The TUI has no models screen at all — `crates/librefang-cli/src/tui/screens/` covers agents, providers-adjacent settings, skills, hands, memory, triggers and more, but nothing for the model catalogue.

## What it costs in practice

A model served through a gateway reports nothing useful about itself. Asking litellm for `sensor-model-generic-high` returns:

```json
"max_tokens": null, "max_input_tokens": null, "max_output_tokens": null
```

The window is configured in the llama.cpp instance behind it and never surfaces. So the model is not in the catalogue, its window is unknown, and the runtime falls back:

```
WARN Model not in catalog — falling back to conservative context window.
     Set `model.context_window` in agent.toml to silence this warning.
     model=sensor-model-generic-high fallback=8192
```

8192 against a real window twice that size. Conversations then hit an overflow that is not real:

```
WARN Stage 2: aggressive overflow compaction, preserving pinned messages estimated_tokens=17581
WARN Context overflow unrecoverable — suggest /reset or /compact
```

Meanwhile the same request succeeds at the provider with `prompt_tokens=16968`. The agent stops answering over a limit that only exists in our own assumption.

The only field the operator can reach is `max_tokens` on the agent, which is the **output** cap, not the window — a different quantity with a similar name. Setting it to the model's context size, which is the natural reading of a field labelled "Max Tokens", asks the model to reserve its entire window for the reply. And it has to be repeated per agent: an install with 42 agents on the same model has 42 places to get it wrong and none to get it right.

## What is being asked

[...]

PR #7784, opened 2026-08-21
feat(cli): add the goal command and a Goals tab to the TUI

DaBlitzStein:
## What

The goal runner has no operator surface. A goal can be created and driven through the API, but the two surfaces an operator actually sits in front of — the CLI and the TUI — cannot see or start one.

This adds `librefang goal` and a Goals screen to the TUI.

## How

- `librefang goal` for list / show / create / cancel, over the existing `/api/goals` routes. No new endpoints.
- A Goals screen in the TUI with a create wizard, filtering, and per-run detail.

The screen is bound to **`Alt+G`**, not to a function key. F1–F12 and `Alt+0`–`Alt+9` are already taken by the 19 existing tabs, and `F8` in particular belongs to the Models screen upstream added in #7774 — taking it would have silently moved someone else's screen.

## Verification

`cargo test -p librefang-cli --bins goal` → **17 passed, 0 failed**, covering the screen's own behaviour rather than only its constructors:

```
tui::screens::goals::tests::refilter_matches_title_and_agent ... ok
tui::screens::goals::tests::refilter_clears_selection_when_nothing_matches ... ok
tui::screens::goals::tests::esc_steps_back_through_the_wizard_then_closes_it ... ok
tui::screens::goals::tests::create_wizard_refuses_to_submit_without_an_agent ... ok
commands::goal::tests::api_error_falls_back_when_the_field_is_missing_or_blank ... ok
```

`cargo clippy -p librefang-cli --all-targets -- -D warnings` → clean.

Rebased on `origin/main` at `a25982bc6`.

## Why this is still open after the runner landed

A coarse `git grep` for "goal" against `origin/main` hits, because the runner and its routes are upstream. Neither surface is: there is no `goal` subcommand in `crates/librefang-cli/src/main.rs`, and no goals screen under `crates/librefang-cli/src/tui/screens/`. This branch is the two surfaces, not the engine.

## Dependency

`tui-goals-phase-paus` / `tui-goals-run-paused` (the paused phase badge and the paused run-state line) render `GoalRunPhase::Paused`, which does not exist on `main` yet — it is added by #7973 (pause/resume plus a configurable loop cadence), which is still open. Until that lands, the paused rendering is unreachable dead display strings, and the i18n coverage guard passes because the keys exist in all four locales. Sequence: #7973 first, then this (or merge together, since the keys are display-only here).


houko:

Conflicting files (7), against `main` at `4c3059e4`:

- `crates/librefang-cli/src/commands/mod.rs`
- `crates/librefang-cli/src/commands/prelude.rs`

DaBlitzStein:
Rebased onto `origin/main` at `17bf1b18e`; this branch is no longer behind. CI should pick it up on `de2299bf7`.

**Still needed, not superseded.** Current `main` has the goals *engine* — `crates/librefang-kernel/src/goal_runner.rs`, `kernel/goal_lifecycle.rs`, `kernel/handles/goal_control.rs` — but neither `crates/librefang-cli/src/commands/goal.rs` nor `crates/librefang-cli/src/tui/screens/goals.rs` exists there. So goals run, and there is no way to reach them from the CLI or the TUI. That is what this adds.

**On the conflicts**, since they all had the same shape and the resolution matters: upstream added a `groups` screen while this branch was adding a `goals` one, so the two collided in five places — the screens list in `tui/mod.rs`, `tui/screens/mod.rs`, `commands/mod.rs`, `commands/prelude.rs`, and the import block in `tui/event.rs` — plus the two locale files where each side appended its own key block.

Every one was resolved by **keeping both**, never by picking a side. Verified afterwards against `origin/main`, symbol by symbol:

| symbol | main | this branch |
|---|---|---|
| `pub mod groups` | 1 | 1 |
| `pub(crate) mod group` | 1 | 1 |
| `GroupInfo` | 6 | 6 |
| `Tab::Groups` | 6 | 6 |
| `tui-tab-groups` | 5 | 5 |

Worth spelling out because one commit in this branch was a rustfmt reflow of that same module list, written before `groups` existed. Applied as-is it would have silently dropped it; the reflow was taken from the rebased side instead.

**Verification on `de2299bf7`:**

```
$ cargo nextest run -p librefang-cli --no-fail-fast
Summary [6.633s] 431 tests run: 431 passed, 0 skipped

$ cargo clippy -p librefang-cli --all-targets -- -D warnings
(clean, 0 warnings)
```

431 against 414 on the branch point — the 17 new ones are this change's own. The run also covers the CLI i18n gates, so the locale keys the Goals tab adds are present in all four locales and none was left dead.

The previous head is preserved at `backup/pr7784-prerebase` on my fork if anything here needs comparing against it.

houko (commented):
Two things: a bug in the `--watch` loop, and the absence of any test for the contract that loop advertises.

**A transient 5xx during `--watch` reports the run as failed.**

```rust
let run_body = daemon_json(client.get(format!("{base}/api/goals/{goal_id}/run")).send());
let running = run_body["running"].as_bool().unwrap_or(false);
let phase = run["phase"].as_str().unwrap_or_default();
// …
if !running {
    if let Some(key) = terminal_phase_message(phase) { … }
    if phase != "finished" { std::process::exit(1); }
    break;
}
```

`daemon_json` exits the process on a transport error, which is fine.
It does *not* exit on an HTTP error status — it prints a message and returns the body, which for a 500 with no JSON is `{}`.
That gives `running = false` and `phase = ""`, so `terminal_phase_message("")` returns `None` (no explanation printed), and `phase != "finished"` fires `exit(1)`.

A single 500 or 503 anywhere in a run that may poll for hours ends the command claiming the goal failed, while the run continues in the daemon.
The long-form help sells the opposite: "the exit status reports the outcome: 0 when the run finished, 1 when it stopped, was rate-limited, or hit the iteration cap" — a poll error is none of those four.

`daemon_json_checked` exists for exactly this shape; its doc comment cites the #6492 `approvals approve` false-success bug, which is this one inverted.
Gate on `status.is_success()` and treat a non-success poll as "keep waiting" (with a bounded consecutive-error count before giving up), so only a real terminal phase decides the exit code.

**1871 lines, no tests.**

`crates/librefang-cli/src/tui/screens/goals.rs` is 995 new lines and `commands/goal.rs` 205, and neither has a test module.
Its sibling PRs out of the same #6505 split all carry them — #7785 adds 221 lines of goals integration tests, #7973 adds a pause/resume suite.

The two things most worth pinning here are pure functions already:

- `terminal_phase_message` / the exit-code mapping — the four terminal phases plus the "unknown phase" case, which is the one the bug above walks through.
- The detail pane keying its start/stop toggle off the run registry's `phase` rather than the goal document's `status`.
  Your own doc comment explains that `status` is set to `in_progress` on start and never cleared, so a finished run would keep offering "stop" — that is a regression that will be reintroduced the first time someone simplifies the detail fetch, and a test is what stops them.

The ordering reasoning in the changelog fragment — resolving and validating the agent *before* creating the goal, because `POST /api/goals/{id}/start` refuses an agent-less goal and the old order left an unrunnable goal behind on every typo — is exactly right, and worth a test too since it is a sequencing property that reads as arbitrary to anyone who has not hit it.

houko (changes requested):
17 tests and a changelog fragment — good. One red test and a dependency to declare.

## Blocking

**1. `Test / Ubuntu (shard 1/4)` is red: the Ukrainian locale is missing four keys.**

```
Ukrainian locale is missing keys referenced by CLI Rust code:
tui-goals-phase-mxit
tui-goals-phase-paus
tui-goals-phase-rlim
tui-goals-run-paused
```
(`crates/librefang-cli/tests/i18n_checks.rs:987`)

`en`, `ko` and `zh-CN` have all eight `tui-goals-phase-*` keys; `uk` is short by three of them plus `tui-goals-run-paused`. Note this is the *coverage* guard, not a translation-quality one — English placeholders would pass it, so please add real Ukrainian strings rather than the fastest thing that turns it green.

**2. `tui-goals-phase-paus` / `tui-goals-run-paused` depend on #7973.**

`GoalRunPhase::Paused` does not exist on `main` — #7973 ("add pause/resume and a configurable loop cadence") adds it, along with `POST /api/goals/{id}/pause` and `/resume`. So the paused rendering here is unreachable until that lands. Please note the dependency in the PR body and sequence the two; #8029 (dashboard pause/resume controls) and #8067 (dashboard phase badges) are in the same position.

## Non-blocking

- The phase keys are four-letter truncations — `mxit`, `rlim`, `paus`, `actv`, `canc`. The rendered values (`⏱ MXIT`, `⚡ RLIM`) explain the choice for the *display* string, but the key names are what a translator and the next maintainer read in a grep. `tui-goals-phase-max-iterations`, `-rate-limited`, `-paused` would cost nothing and match how the phase is spelled everywhere else (`GoalRunPhase::MaxIterationsReached`, `max_iterations_reached`).
- `Security` and `CI Gate` are red on the pre-existing `cargo audit` advisories in transitive dependencies (`lru` 0.18.0 RUSTSEC-2026-0253, `rand` 0.7.3 RUSTSEC-2026-0097) that affect most open PRs — not from this diff.

DaBlitzStein:
Addressed the review feedback on the i18n coverage guard.

The guard checks `en`, `uk` and `zh-CN`, but `ko` was also missing the same four keys, so all three non-English locales now carry `tui-goals-phase-mxit`, `tui-goals-phase-paus`, `tui-goals-phase-rlim`, and `tui-goals-run-paused` with real translations.
The paused status depends on #7973 (`GoalRunPhase::Paused`), noted in the commit body.

Verified: `cargo test -p librefang-cli --test i18n_checks` → 3 passed, 0 failed.
Commit 1118440d8.

DaBlitzStein:
Both blocking points and the first review's watch-loop bug are addressed; the branch is merged over `origin/main` and conflict-free.

## Merge over `origin/main`

Branch merged onto `origin/main` at `07cb6fc76` (`9d0d24de7`). Conflicts in `locales/uk/main.ftl`, `locales/zh-CN/main.ftl`, and `tui/mod.rs` resolved by keeping both sides — main's Channels screen keys/arms and this PR's Goals ones; no key or tab from either side was dropped, and duplicate-key and dead-key checks pass on the union.

## Blocking 1: the Ukrainian locale was missing four keys

Already fixed at the branch tip before this push: `1118440d8` added `tui-goals-phase-mxit`, `tui-goals-phase-paus`, `tui-goals-phase-rlim`, and `tui-goals-run-paused` to `uk` (and the same four to `ko`), with real Ukrainian strings, not English placeholders. Verified at `b1683383c`: `cargo test -p librefang-cli --test i18n_checks` → 3 passed (`test_locales_cover_used_i18n_keys`, `test_no_dead_locale_keys`, `test_no_untranslated_strings`).

## Blocking 2: the paused keys depend on #7973

Noted in the PR body under "Dependency": `GoalRunPhase::Paused` arrives with #7973 (still open), so the paused rendering is unreachable until then; sequence #7973 first, or merge together since the keys are display-only here.

## Watch-loop 5xx bug (first review)

Fixed in `ee62452c9`. The poll is now gated on the HTTP status via `daemon_json_checked` (the helper added for the #6492 approvals false-success shape), and a poll the CLI cannot classify — a failed status, or a 200 whose `running=false` carries no recognized terminal phase — is retried for up to five consecutive failures before the watch gives up with an explicit outcome-unknown message. Only a real terminal `GoalRunPhase` decides the exit code. Three new tests pin the classification: `classify_poll_treats_a_failed_poll_as_unobservable_not_a_failure`, `classify_poll_routes_running_and_terminal_phases`, and `classify_poll_never_treats_an_unknown_phase_as_terminal`.

## Non-blocking: truncated phase key names

Done in `b1683383c`: `tui-goals-phase-mxit`/`-rlim`/`-paus` renamed to `tui-goals-phase-max-iterations`/`-rate-limited`/`-paused`, matching `GoalRunPhase::MaxIterationsReached` / `max_iterations_reached` and the existing `tui-goals-run-*` full spellings, across the code and all four locales. Mechanical rename gated by the same i18n tests.

## Verification

```
cargo check -p librefang-cli                                          → clean
cargo test -p librefang-cli --bin librefang goal::                    → 6 passed (3 new)
cargo test -p librefang-cli --bin librefang goals::                   → 14 passed
cargo test -p librefang-cli --test i18n_checks                        → 3 passed
cargo clippy -p librefang-cli --all-targets -- -D warnings            → clean
cargo fmt -p librefang-cli --check                                    → clean
```

A changelog fragment for the watch fix is at `changelog.d/fixed/7784-goal-watch-transient-poll-errors.md`, and the feature fragment's stale `F8`/`Alt+8` claim (the tab binds `Alt+G`; `F8` belongs to the Models screen from #7774) is corrected.

houko (approved):
Both blocking points resolved, `CI Gate` green with zero non-success checks, and `mergeable` is `BLOCKED` on review rather than a conflict — so this one is not caught in the rebase churn the rest of my sweep caused.

**1 — the `uk` locale.** All four keys are present with real Ukrainian strings — `АКТВ`, `ГОТОВО`, `СКАС`, `ПОМИЛКА`, `ОЧІК`, `ПАУЗ`, `ЛІМІТ`, `МАКС`, `призупинено` — not the English placeholders the coverage guard would also have accepted. That was the actual ask.

**2 — the #7973 dependency** now has its own `## Dependency` section, and it says the useful part rather than just naming the number: the paused rendering is unreachable display strings until #7973 lands, the i18n guard passes regardless because the keys exist in all four locales, and the sequencing recommendation is explicit. Worth knowing that #7973 is still open with changes requested from me — a `/resume` bug where the iteration cap falls back to the default and can terminate the resumed run immediately — so "merge together" is likely to mean waiting on that.

**Non-blocking key naming — the part that mattered is done.** `tui-goals-phase-paused`, `-rate-limited` and `-max-iterations` now match how the phase is spelled in `GoalRunPhase`, which is what makes a grep from the Rust side land. Leaving `actv` / `canc` / `pend` / `fail` / `done` truncated is a defensible line to draw: those are abbreviations of single words that read the same either way, and the three I actually cited were the multi-word ones.

## On the unrequested fix, which I'm glad is here

`ee62452c9` is a real bug I did not ask about and it is the more serious defect in this PR. `daemon_json` returns a 5xx body unchanged — often `{}` — which reads as `running=false` with an empty phase, so a single transient 500 ended `--watch` with `exit(1)` and reported the goal as failed while the run continued in the daemon. In a `&&` chain that is worse than a hang: it reports a definite wrong answer.

The fix is the right shape rather than a retry bolted on:

- Routed through `daemon_json_checked`, the helper added for the #6492 false-success shape — same class of bug, same remedy, so the two now agree.
- `classify_poll(status, body) -> PollVerdict` is extracted as a pure function, so the classification is testable without a server. The three tests pin exactly the three cases that matter: a 500/503 is unobservable rather than failed, the four terminal phases are terminal, and a 200 whose `running=false` carries an unknown or missing phase is retried rather than reported as a failure.
- Bounded at five consecutive failures with an explicit outcome-unknown message, in all four locales, and only a real terminal `GoalRunPhase` decides the exit code — which is what the long-form help promises.

Its own changelog fragment too, correctly separated from the feature one.

Also noting the fragment correction that came with it: the TUI binding moved from `F8` / `Alt+8` to `Alt+G` with the reason recorded ("the first free mnemonic — every numeric slot is already bound"). Catching that the retired Channels slot was not actually free is the kind of thing that only shows up when someone tries it.

One stale detail to tidy whenever you next touch the body: the `## Dependency` section still names `tui-goals-phase-paus`, which `b1683383c` renamed to `tui-goals-phase-paused`.

Approving.

PR #7781, opened 2026-08-21
feat(kernel): complexity-based model routing with per-agent profile overrides

DaBlitzStein:
Closes #7757.

This is group 4 of the five-way split houko asked for when closing #6505.

## Two bugs a reviewer should look at first

Both were found while writing this code, both are in the path this PR already touches, and both have tests.

**A cost budget could be bypassed on the fallback path.** The profile matcher ended with `profiles.first()` as a last resort, which returned an arbitrary profile *without re-checking the agent's `allowed_profiles` or `cost_budget`*. An agent capped at the cheap tier could therefore be routed onto an expensive profile whenever nothing matched the task's tags. A spending cap that silently evaporates on a fallback is worse than no cap, because the operator believes it holds. The fallback chain now re-checks the same constraints, and when nothing survives the router declines and the agent keeps its own model rather than being handed something arbitrary. Pinned by `fallback_cannot_escape_the_cost_budget` and `fallback_cannot_escape_the_allowlist`.

**An unknown `cost_budget` string was read as "no cap".** A typo in the field that limits spending removed the limit, and nothing surfaced it until the bill arrived. It is now a 400, and the rejected write does not take effect. `null` remains the documented no-cap value and is still accepted. Pinned by `put_model_routing_rejects_an_unknown_cost_budget` and `put_model_routing_accepts_a_null_cost_budget`.

## What this does

Picks the model per turn from the complexity of the work, instead of pinning one model per agent.

- `ModelProfile` / `ModelRouterConfig` / `AgentRouterOverride` / `CostTier` in `librefang-types` (`model_profile.rs`).
- A profile router in the kernel (`model_router.rs`): a heuristic complexity score over the turn text, plus a matcher that ranks profiles by tag overlap, then priority, then name.
- `ProfileCatalog` loads the builtin asset (`crates/librefang-kernel/assets/model_profiles.toml`) and merges `~/.librefang/model_profiles.toml` over it by profile name — same name replaces, new name is added.
- The catalog is re-read whenever the override file's mtime or length moves, so an operator's edit takes effect without a daemon restart; one `stat` per routed turn, which is nothing next to the LLM call it precedes.
- A malformed or unreadable override file logs a `WARN` and falls back to the builtins rather than taking the daemon down.
- `[model_router]` is a real `KernelConfig` field, with a `build_reload_plan_with_caps` branch, an entry in `classified_reload_fields()`, and a row in `docs/operations/config-reload.md`.
[...]

Issue #7780, opened 2026-08-21
bug(models): a discovered model gets a hardcoded 131072-token context window that nothing can source or correct


DaBlitzStein:
## What happens

`merge_discovered_models` (`crates/librefang-runtime/src/model_catalog.rs`) builds a `ModelCatalogEntry` for every newly discovered model with two literals:

```rust
context_window: 131_072,
max_output_tokens: 16_384,
```

Nothing upstream of that line can supply a real value, because `DiscoveredModelInfo` (`crates/librefang-runtime/src/provider_health.rs:16`) has no field for either one.
Its shape is entirely Ollama-derived — `parameter_size`, `quantization_level`, `family`, `families`, `size`, `capabilities` — and the OpenAI-compatible `/v1/models` branch of `probe_provider` fills in only `name`.

So a gateway-discovered model is admitted to the catalog with a context window that is a guess presented as a fact.

## Why it matters

The number is not merely unknown, it is confidently wrong in whichever direction the real model differs, and the two directions fail differently:

- **Guess too large** — the daemon packs a prompt the model cannot accept and the request fails at the provider, after the tokens have been paid for.
- **Guess too small** — compaction fires early, or a turn is refused as an overflow that never existed.

Both are silent: nothing in the catalog marks the value as inferred, so an operator reading the model list sees `131072` and reasonably believes LibreFang learned it from the gateway.

## Reproduced against a live LiteLLM gateway

Five models behind LiteLLM, reached over its OpenAI-compatible surface.

`GET /v1/models` returns the bare OpenAI shape — every entry carries exactly `id`, `object`, `created`, `owned_by`, and no capacity field of any kind:

```
sensor-model-generic       | fields: ['created', 'id', 'object', 'owned_by']
sensor-model-generic-high  | fields: ['created', 'id', 'object', 'owned_by']
pakllm                     | fields: ['created', 'id', 'object', 'owned_by']
rodela-testing-model       | fields: ['created', 'id', 'object', 'owned_by']
embedding-high             | fields: ['created', 'id', 'object', 'owned_by']
```

LiteLLM's richer `GET /model/info` does carry the fields, and on this deployment they are all null:

```
sensor-model-generic        max_input=None  max_output=None  max_tokens=None
sensor-model-generic-high   max_input=None  max_output=None  max_tokens=None
pakllm                      max_input=None  max_output=None  max_tokens=None
rodela-testing-model        max_input=None  max_output=None  max_tokens=None
[...]

PR #7786, opened 2026-08-21
fix(models): agent inference settings win over the per-model override

DaBlitzStein:
## What this changes

An agent's own inference settings now win over the per-model override instead of losing to it.

Two instances of one agent type could not run the same model at different temperatures.
Tuning the shared model overwrote both of them with one value and discarded their individual settings in silence — the specific setting lost to the general one.

## Why the priority was inverted

It was not a decision about precedence.
`ModelConfig.max_tokens` was `u32` and `.temperature` was `f32`, so every agent always carried a concrete number — 4096 / 0.7 from the `Default` impl if nobody chose one.
There was no "inherit" state, so if the manifest won, the per-model override would never apply to anything.
The inversion was a workaround for the missing state.

Fixing the type is what makes the priority fixable, so this PR does both.

## Substantive changes

- `ModelConfig.max_tokens` / `.temperature` become `Option`; `None` means inherit.
- `top_p`, `frequency_penalty` and `presence_penalty` gain a per-agent equivalent — they existed per model but an agent had no way to ask for its own.
- `context_window` and `max_output_tokens` become editable per agent.
- New `crates/librefang-types/src/inference_params.rs` holds the resolution as a pure function: preferences resolve agent manifest → per-model override → system default.
- `agent_execution.rs` calls it once per turn, after model routing, so the override matches the model that will actually be called.
- `AgentRegistry` gains per-knob setters; each takes `Option`, and `None` clears the agent's value.
- `PATCH /api/agents/{id}/config` becomes tri-state per field: omit the key to leave it unchanged, send `null` to hand the field back to inherit, send a value to pin it.
- CLI: `librefang agent set <id> <field> <value>`, with `inherit` as the clearing value.
- TUI: agent detail → `p` opens an editor with the same seven knobs, both step ladders, and a custom entry.
- `docs/architecture/inference-parameter-resolution.md` documents the two categories and the precedence rules.

## The counterexample this must not break

`reasoning_effort` is deliberately excluded from the reordering.

It is not a preference but a fact about the endpoint: a gateway that rejects it rejects every turn that carries it, which is #7770 — a LiteLLM deployment that took down every turn.
The model level keeps winning there, and an agent cannot force it on.

Two doors are closed, not one:

- `apply_to` **removes** a stale `extra_params["reasoning_effort"]` when the model level does not set one, rather than leaving it in place. Otherwise a hand-written entry survives onto an endpoint that refuses the parameter, which is the same failure arriving by the back door.
- `PatchAgentConfigRequest` uses `deny_unknown_fields`, so sending it over the API is a `400`. There is a test for that.

## Limits warn, they do not clamp

Asking for more than a model allows is reported at save time — in the `PATCH` response's `warnings` array, in the daemon log, and in the editors — and the value is stored and sent exactly as entered.

Nothing is clamped.
If the catalog's figure is the thing that is wrong, a silent truncation leaves the operator debugging a number they never chose, whereas an explicit provider error names the problem.

Only a limit that something actually asserted produces a warning.
`ModelCatalogEntry` gains `limits_known`, following the existing `pricing_known` convention where a missing value defaults to `true`.
`merge_discovered_models` marks its `131_072` / `16_384` placeholders unknown, because a gateway-discovered model has no capacity to source: `DiscoveredModelInfo` carries no such field and the OpenAI-compatible `/v1/models` shape carries none either (#7780).
The `model_metadata.rs` L5 tail is marked the same way — a substring table and two round constants are pattern-matching on a model's name.

Warning against a ceiling invented from a model's name is noise, and noise is what trains people to stop reading warnings.

## Migration

Nothing changes for an already-deployed agent.

Every persisted manifest carries explicit numbers, because the old type had no way to omit them, so they deserialize as `Some(n)` and stay pinned.
The inherit state appears only on newly created agents and where an operator chooses it.

What does change, deliberately, is which value wins when both an agent and a per-model override are set: the agent's.
That only affects deployments that had set a per-model override, and it is the bug this PR exists to fix.

## Verification

- [ ] `cargo check --workspace --lib`
- [ ] `cargo clippy --workspace --all-targets`
- [ ] `cargo test -p librefang-types -p librefang-kernel -p librefang-cli`
- [ ] `cargo test -p librefang-api --test agent_model_params_test`
- [ ] `pnpm vitest run` (dashboard)
- [ ] `openapi.json` regenerated — `PatchAgentConfigRequest` gains seven fields, so the drift lane fails without it

Tests added:

- `inference_params.rs` — the resolution chain, the two-writers case, the `reasoning_effort` counterexample in both directions, warn-without-clamping, and an inferred limit staying silent.
- `crates/librefang-api/tests/agent_model_params_test.rs` — tri-state round-trip against the real axum router, the migration guarantee, over-limit warning with the stored value unchanged, and an inferred limit producing no warning.
- CLI: `parse_inference_param` covering `inherit` and the integer/float split.
- TUI: the ladder stepping rules, the cap, and the custom entry.

## Out of scope

- `#7774` / `#7780` point 3 — a per-model capacity override the operator can set once and have survive registry sync. `limits_known` here is groundwork for that and deliberately stops short of the editing surface.

Refs #7780, #7781


houko:

Conflicting files (15), against `main` at `4c3059e4`:

- `crates/librefang-api/dashboard/src/api.test.ts`
- `crates/librefang-api/dashboard/src/lib/agentModelPatch.test.ts`
- `crates/librefang-api/dashboard/src/lib/agentModelPatch.ts`
- `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx`
- `crates/librefang-cli/locales/en/main.ftl`
- `crates/librefang-cli/locales/ko/main.ftl`
- `crates/librefang-cli/src/tui/screens/agents.rs`
- `crates/librefang-runtime/src/model_catalog.rs`
- `crates/librefang-runtime/src/model_metadata.rs`
- `xtask/baselines/openapi.sha256`

houko (commented):

```
```

```
cargo xtask codegen --openapi
cargo xtask schema-check gen
```

houko (changes requested):
The engineering here is good — `Option<Option<T>>` on the PATCH shape to separate "absent" from "explicitly inherit", `effective_max_tokens()` / `effective_temperature()` accessors so no call site re-derives the precedence, the `Default for ModelConfig` impl correctly moved to `None` (which is what makes the inherit state reachable at all, given the struct-level `#[serde(default)]`), plus `manifest_field_parity.rs`, `agent_form_roundtrip.rs` and an architecture doc. The blocker is the changelog.

## Blocking

**1. This deletes 1,295 lines of `CHANGELOG.md`, including the entire released `## [2026.8.30]` section.**

```
@@ -9,1322 +9,27 @@
-## [2026.8.30] - 2026-08-30
-
-_554 PRs from 2 contributors since v2026.8.19._
```

That section is the released history 554 PRs were folded into, and it is what `.github/workflows/release.yml` slices the release body from. Merging this would ship a `CHANGELOG.md` with a release missing from it. It reads like a rebase resolution that took "ours" wholesale on a branch that predates the release cut — please restore the file to `main`'s content.

**2. Editing `CHANGELOG.md` at all is the thing the repo asks you not to do.**

`CLAUDE.md`, verbatim heading:

> ### Changelog: a new file under `changelog.d/`, NOT an edit to `CHANGELOG.md`
> Appending to the single `## [Unreleased]` section conflicts with every other open PR doing the same, in a file where the conflict carries no information — both sides are correct and the resolution is always "keep both".

#7787 does the same thing on the same file, which is exactly the conflict that rule predicts.

The prose you wrote is genuinely good and should be kept — move it into `changelog.d/added/7786-….md` and `changelog.d/fixed/7786-….md` verbatim, minus the leading `- `, one sentence per line, continuation lines indented two spaces, ending `(#7786) (@DaBlitzStein)`. `cargo xtask collect-fragments` folds them into `[Unreleased]` at release time, so nothing is lost.

## Coordination

- This is part of the #7781 / #7786 / #7787 stack (the bullets are attributed `(#7781, #7786)`). #7781 already carries a changes-requested review; the three should be sequenced explicitly, because all of them touch `AgentsPage.tsx`, `api.ts` and the model-parameter surface.
- #8112 and #8106 also edit the model-parameter form and `SliderInput`, and #8112 adds `top_p` / `frequency_penalty` / `presence_penalty` to the manifest form by writing them into `ModelConfig.extra_params`. This PR makes them real typed fields, which is the right answer — worth telling #8112 so it does not land the untyped version first.

## Non-blocking

- `limits_known` on `ModelCatalogEntry`, and marking `merge_discovered_models`' `131_072` / `16_384` placeholders as unknown, is the right call and the reasoning in the prose ("warning against a ceiling invented from a model's name is noise, and noise is what trains people to stop reading warnings") deserves to survive into the fragment.
- `Security` / `OpenAPI Drift` / `CI Gate` are red. Security is the pre-existing `cargo audit` pair (`lru` 0.18.0 RUSTSEC-2026-0253, `rand` 0.7.3 RUSTSEC-2026-0097) shared with most open PRs; the drift should clear on a rebase with `openapi.json` and `xtask/baselines/openapi.sha256` regenerated (recompute the baseline rather than picking a side — neither hash describes the merged file).

houko:

No commits of your own since my review on 2026-09-01T14:16:49Z; the only thing that moved the branch tip is a merge from `main`, which the `auto-update-branches` cron pushes by itself. The PR's "updated" timestamp has therefore advanced while the blocking items have not been touched. Worth flagging because that timestamp is what makes a PR look answered from outside — I nearly mis-triaged a dozen PRs on exactly that signal today.

Still open from that review:

- 1. This deletes 1,295 lines of `CHANGELOG.md`, including the entire released `## [2026.8.30]` section
- 2. Editing `CHANGELOG.md` at all is the thing the repo asks you not to do

Separately, and this part is on me: the branch now conflicts with `main`, because I merged a batch of PRs while working through the queue. A rebase will be needed on top of whatever you push — tell me if the overlapping files are not obvious and I will list them.

DaBlitzStein:
Refresh — merge dc821eef8.

- The changelog fragment `changelog.d/fixed/7786-agent-inference-settings-tristate.md` is present on the merged tip, and `CHANGELOG.md` itself stays identical to main (the earlier accidental deletion was already restored).
- `tsc --noEmit` zero errors on the merged tip.
- vitest 1721 passing, with only the known PromptsExperimentsModal flake, which passes isolated.

The feature diff is unchanged.

DaBlitzStein:
Closing this in favour of #7787, which contains it.

The two branches are not independent: `feat/modelparams-ladders` carries this branch's commits rebased — same subjects (`fix(models): agent inference settings win over the per-model override`, `fix(types): remove duplicate limits_known…`, `fix(dashboard): strict number parsing…`), different SHAs. Comparing the two diffs against `main` file by file, the 46 files here are a subset of the 59 there; the only path unique to this branch is its own changelog fragment, and #7787's fragment covers everything this one said and adds the `reasoning_effort` rule and the save-time warning behaviour. Both merges also produce a byte-identical `openapi.json` (sha `1cbacb65…`), which is the strong signal that there is no divergence hiding between them.

Splitting #7787 back down to just the step-ladder work would mean rewriting its history and leaving two PRs that have to land in order, which is more fragile than one coherent PR — and nothing is lost by consolidating, since every commit here already lives there.

Nothing to review separately: #7787 is the one to look at.

Issue #7782, opened 2026-08-21
bug(dashboard): the model parameter form dims the switch that would un-dim it, so it reads as broken


DaBlitzStein:
## What happens

Every parameter row in the model detail form (`SliderInput`, `crates/librefang-api/dashboard/src/components/ui/SliderInput.tsx`) carries its inactive styling on the row container:

```tsx
<div className={`space-y-1.5 ${!enabled ? "opacity-40" : ""}`}>
```

That container also holds the `role="switch"` button which is the only way to make the row active.
So an inactive row fades its own escape hatch to 40%, and the switch's off state is `bg-border-subtle` — the hairline token used for dividers — which is already the faintest colour in the palette before the opacity is applied on top of it.

## Why it is the default state, not an edge case

`ModelsPage` initialises every `*Enabled` flag to `false` and only flips one on when the model already carries an override for that field:

```tsx
tempEnabled: false, topPEnabled: false, maxTokensEnabled: false,
contextWindowEnabled: false, freqEnabled: false, presEnabled: false,
```

A model with no overrides — which is every model until someone sets one — therefore opens with **all seven** rows faded, and with all seven switches faded along with them.
The form does not read as "opt in per field". It reads as a screen that failed to load, or as a permission the account does not have.

## Accessibility

A `role="switch"` is an interactive control. Rendering it at 40% opacity in the divider token puts it well under the 3:1 contrast floor WCAG 2.2 sets for non-text UI components (1.4.11), in the state where it most needs to be found.

## Fix

Dim the values, never the control that restores them:

- move the `opacity-40` from the container onto the label, the number input, the range input and the tick row
- give the off state a token that reads as a control rather than as a divider

The dimming itself is worth keeping — it is what distinguishes "using the catalog default" from "overridden here" at a glance — so the fix is where it is applied, not whether.

## Regression cover

`SliderInput.test.tsx` pins both halves, because the obvious wrong fix is to drop the dimming altogether:

- the switch carries no `opacity-` class while the row is inactive
- the slider and the number input still do
- the off state does not use `bg-border-subtle`
- one click on the switch activates the row
- `aria-checked` tracks the state

## Related

- #7774 — the context window must be editable from every surface. Before that change this same form hardcoded `enabled={false}` with an empty `onChange` and no `onToggle` at all for the context window row, so that field had no switch to find in the first place. This issue is the layer underneath: even with a switch present, it was drawn where it could not be seen.

PR #7787, opened 2026-08-21
feat(dashboard): step ladders with a custom entry for the token-count fields

DaBlitzStein:
**Stacked on #7786** — it needs the `inherit` state to have a rung to show.

Targets `main` rather than #7786's branch because a cross-fork PR cannot use a fork branch as its base.
Until #7786 merges, the diff below also contains #7786's commit; review only `feat(dashboard): step ladders…`.
Merge #7786 first and this shrinks to its own commit.

## What this changes

The response-length and context-window fields in the agent editor and the agent-type editor become step ladders with a custom entry, replacing free sliders.

## Why a ladder

The useful values for a token count are an order-of-magnitude sequence, not a continuum.
Landing on exactly 131072 by dragging is a chore, the numbers in between mean nothing to any provider, and the slider gave no clue which values were sensible.

`inherit` becomes a rung rather than an empty box, so "this agent has no opinion, let the model's setting apply" is something an operator can point at instead of inferring from a blank field.

## Two ladders, not one

- Context: 8K · 32K · 128K · 256K · 512K · 1M · 2M
- Output: 1K · 4K · 8K · 16K · 32K · 64K · 128K

The output ladder deliberately stops at 128K.

Output tokens are not context tokens.
Gemini's 1M / 2M are how much it can *read*, and no model generates a million tokens of reply.
A ladder offering 1M output would assert that the value is valid and invite a setting the provider will refuse — worse than the slider it replaces, because a slider makes no claim.

`modelParamLadders.test.ts` pins both ladders and asserts the output one stays below the context one, so a future edit cannot quietly merge them.

## Caps

A limit the model actually declared trims the ladder, so no rung is offered that the endpoint cannot honour.
When the cap falls between two rungs it is appended, which is the only way to offer a model whose real maximum is, say, 20000.

A limit nobody measured trims nothing.
`limits_known === false` marks a discovery placeholder, and capping against one would hide rungs the endpoint may well support (#7780).

## Substantive changes

- New `src/lib/modelParamLadders.ts` — the two ladders, `formatTokens`, `ladderUpTo`, `isOnLadder`. Mirrors `CONTEXT_WINDOW_LADDER` / `MAX_OUTPUT_TOKENS_LADDER` in `librefang-types`, which the TUI editor uses; the test pins the values so a one-sided edit fails.
- New `src/components/ui/StepLadderInput.tsx`. `SliderInput.tsx` is untouched.
- `AgentManifestForm.tsx` — ladders for `max_tokens` and `context_window`, plus the four sampling knobs an agent could not previously set, plus `max_output_tokens`.
- An over-limit value is flagged in red and left exactly as entered: the field is not marked invalid and nothing is clamped.
- The agent editor and the agent-type editor share this form, so both get all of it.
- Locale keys in all five locales.

## Verification

- [ ] `pnpm vitest run`
- [ ] `pnpm tsc --noEmit`

Tests added:

- `modelParamLadders.test.ts` — ladder shape, the output-below-context invariant, formatting, capping including the between-rungs case, and no cap when none was sourced.
- `StepLadderInput.test.tsx` — inherit as a rung, every rung rendered, the custom field opening for an off-ladder value and staying open, re-selecting a preset, cap trimming, and the advisory not disabling anything.
- `AgentManifestForm.test.tsx` — all five preference knobs present, everything starting on inherit, the output ladder not offering 2M while the context ladder does, custom entry, over-limit flagged without changing the value, and silence when limits were never sourced.

Refs #7781, #7782


houko:

Conflicting files (15), against `main` at `4c3059e4`:

- `crates/librefang-api/dashboard/src/api.test.ts`
- `crates/librefang-api/dashboard/src/lib/agentModelPatch.test.ts`
- `crates/librefang-api/dashboard/src/lib/agentModelPatch.ts`
- `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx`
- `crates/librefang-cli/locales/en/main.ftl`
- `crates/librefang-cli/locales/ko/main.ftl`
- `crates/librefang-cli/src/tui/screens/agents.rs`
- `crates/librefang-runtime/src/model_catalog.rs`
- `crates/librefang-runtime/src/model_metadata.rs`
- `xtask/baselines/openapi.sha256`

houko (commented):

```
```

```
cargo xtask codegen --openapi
cargo xtask schema-check gen
```

houko (changes requested):
The step-ladder work looks well covered — `StepLadderInput.test.tsx`, `modelParamLadders.test.ts`, `AgentManifestForm.test.tsx` and `api.test.ts` are all extended — and the reasoning for two different ladders ("output tokens are not context tokens, Gemini's 1M / 2M are how much it can *read*, and no model generates a million tokens of reply") is the right instinct. The blocker is the changelog.

## Blocking

**1. This deletes 1,311 lines of `CHANGELOG.md`, including the entire released `## [2026.8.30]` section.**

```
@@ -9,1322 +9,34 @@
-## [2026.8.30] - 2026-08-30
-
-_554 PRs from 2 contributors since v2026.8.19._
```

That is the released history covering 554 PRs, and `.github/workflows/release.yml` slices the release body out of that section. Merging this ships a `CHANGELOG.md` with a whole release missing. It looks like a rebase resolution that took "ours" on a branch predating the release cut — restore the file to `main`'s content.

**2. `CHANGELOG.md` should not be edited at all.**

`CLAUDE.md`:

> ### Changelog: a new file under `changelog.d/`, NOT an edit to `CHANGELOG.md`
> Appending to the single `## [Unreleased]` section conflicts with every other open PR doing the same, in a file where the conflict carries no information — both sides are correct and the resolution is always "keep both".

#7786 does the same thing to the same file, so the two will conflict on a file where the conflict tells a reviewer nothing — precisely the failure the rule exists to prevent.

Move the bullets into `changelog.d/added/7787-….md` verbatim (drop the leading `- `, one sentence per line, continuation lines indented two spaces, ending `(#7787) (@DaBlitzStein)`). `cargo xtask collect-fragments` folds them into `[Unreleased]` at release time.

## Coordination

Three other open PRs edit the same model-parameter surface, and all four will conflict:

- **#8112** adds `top_p` / `frequency_penalty` / `presence_penalty` inputs to `AgentManifestForm` *and* rewrites `SliderInput`'s toggle geometry and opacity token.
- **#8106** changes `SliderInput`'s off-state token and enables the `context_window` slider on `ModelsPage`.
- **#7786** makes `max_tokens` / `temperature` `Option` and adds the other three as typed fields.

This PR replaces the free sliders with ladders, which subsumes most of what #8112 and #8106 are reaching for. Worth agreeing which of the four owns the model-parameter form before any of them merge — right now three different designs for the same control are in review simultaneously.

## Non-blocking

- "`inherit` is a rung rather than an empty box, so 'this agent has no opinion' is something you can point at" is the best sentence in the batch and should survive into the fragment intact.
- `Security` / `OpenAPI Drift` / `CI Gate` are red; Security is the pre-existing `cargo audit` pair (`lru` 0.18.0 RUSTSEC-2026-0253, `rand` 0.7.3 RUSTSEC-2026-0097) affecting most open PRs, and the drift should clear on a rebase with `openapi.json` and a recomputed `xtask/baselines/openapi.sha256`.

houko:

No commits of your own since my review on 2026-09-01T14:16:51Z; the only thing that moved the branch tip is a merge from `main`, which the `auto-update-branches` cron pushes by itself. The PR's "updated" timestamp has therefore advanced while the blocking items have not been touched. Worth flagging because that timestamp is what makes a PR look answered from outside — I nearly mis-triaged a dozen PRs on exactly that signal today.

Still open from that review:

- 1. This deletes 1,311 lines of `CHANGELOG.md`, including the entire released `## [2026.8.30]` section
- 2. `CHANGELOG.md` should not be edited at all

Separately, and this part is on me: the branch now conflicts with `main`, because I merged a batch of PRs while working through the queue. A rebase will be needed on top of whatever you push — tell me if the overlapping files are not obvious and I will list them.

DaBlitzStein:
Refresh — merge 2c9fe8650.

- The changelog fragment `changelog.d/fixed/7787-agent-inference-settings-override.md` is present on the merged tip, and `CHANGELOG.md` itself stays identical to main.
- CLI locale parity confirmed across the merged tip.
- Dashboard `tsc --noEmit` zero errors on the merged tip.

The feature diff is unchanged.

DaBlitzStein:
Pushed `1f9fc239a`: merged current `main` (this branch was six commits behind and conflicting).

Two conflicts, both resolved keeping every side:

- `crates/librefang-cli/locales/zh-CN/main.ftl` — this branch appends 25 `tui-agents-param-*` / `tui-event-model-params-*` / `agent-set-invalid-*` keys at the end of the file, and `main` (#8141) appends four `tui-memory-config-*` / `tui-goals-run-*` keys in the same place. No key collides. The branch's block goes first and #8141's after, matching the order git auto-merged into `en`, `ko` and `uk` without conflict. Key-set parity across the four locales verified separately: `missing=0 extra=0` against `en` for all of them.
- `xtask/baselines/openapi.sha256` — neither side of the conflict was valid, since `openapi.json` had already auto-merged into something different from both. `xtask/src/schema_check.rs` defines the baseline as a plain `sha256sum` of that file, so it was recomputed rather than picked: `1cbacb653bda2198c2f5c2862ab3e1b20f20c27e830f60c503243f6ca50749dc`.

**#7786 is now closed as absorbed by this PR.** Its commits already live here rebased, its 46 changed files are a subset of this branch's 59, and the only path unique to it was its own changelog fragment — whose content this branch's fragment already covers, plus the `reasoning_effort` rule and the save-time warning behaviour. Both merges also produce a byte-identical `openapi.json`, which is what confirmed there was no divergence hiding between them.

- `cargo check -p librefang-types -p librefang-hands -p librefang-kernel -p librefang-api -p librefang-cli -p librefang-runtime -p librefang-testing --all-targets` — exit 0.
- `cargo nextest run -p librefang-api -E 'binary(agent_model_params_test)'` — 7/7, the suite that pins the tri-state behaviour this PR is built on: `preferences_round_trip_through_all_three_states`, `out_of_range_preferences_are_rejected`, `operator_limit_outranks_the_catalog_limit`, `exceeding_a_known_limit_warns_and_does_not_clamp`, `exceeding_an_inferred_limit_stays_silent`, `context_window_is_checked_against_the_catalog_window`, `an_existing_explicit_value_survives_unrelated_patches`.
- `cargo nextest run -p librefang-types -p librefang-cli` — 1610/1610, including `i18n_checks test_no_untranslated_strings`, which is what would have caught a dropped key in the zh-CN resolution.
- `npx tsc --noEmit` clean; `npx vitest run` over the seven touched test files — 133/133.
- `i18n-parity` — failing set is ko/pl/uk with the same per-locale detail as `main` (`ko` missing `agents.children_count_one`, `pl`/`uk` 12 extra plural forms). Nothing added.

One merge-order note: this PR and #8013 (shared-folders editor) both touch `agentManifest.ts`, `AgentManifestForm.tsx` and all five locales, and neither is on `main` yet — so the second to land will conflict. The changes are additive in different regions, but a resolution that takes one side wholesale would silently drop the other's serializer hunk. Flagged on #8013 as well.

houko (approved):
Re-reviewed at the current head. Both blocking items are fixed, and the coordination problem resolved itself while this was open.

**1 — the released section is back.** `CHANGELOG.md` is out of the diff entirely; `2c9fe865 chore: merge origin/main to restore the released changelog` is the commit that did it. The `## [2026.8.30]` section and its 554 PRs are intact on this head.

**2 — one fragment, in `changelog.d/`.** `changelog.d/fixed/7787-agent-inference-settings-override.md`, and it explains the *why* rather than restating the title — the second sentence is the one that earns its place in the release body:

> Two instances of one agent type could not run the same model at different temperatures: `ModelConfig.max_tokens` and `.temperature` were plain numbers, so every agent carried a concrete 4096 / 0.7 whether or not anyone chose them, and letting the manifest win would have made per-model overrides unreachable for every agent in existence.

Recording the `reasoning_effort` exception, and *why* it keeps the opposite rule ("a fact about the endpoint rather than a preference"), is the part a future reader will need when they find one field resolving the other way round from its neighbours.

## The four-way overlap has collapsed to nothing

Worth updating what I wrote last time, since three of the four moved:

- **#7786 is closed**, and this PR carries its work — which is why the fragment describes the tri-state change rather than only the ladders.
- **#8112** dropped `SliderInput` from its diff entirely.
- **#8106** owns the toggle now, and only the toggle.

So there is no longer a question about which PR owns the model-parameter form; this one does, and nothing else is competing for `SliderInput`. Nothing to agree.

## One non-blocking item did not carry over

The sentence I said should survive into the fragment — *"`inherit` is a rung rather than an empty box, so 'this agent has no opinion' is something you can point at"* — is not in it. The fragment covers the resolution chain and the tri-state, but not the design decision that makes the ladder legible to an operator looking at it. Still non-blocking, and still the best sentence in the batch; one line at the end of the fragment would carry it.

Approving.

PR #7788, opened 2026-08-21
fix(providers): report output capacity, not a request-length preference

DaBlitzStein:
**Stacked on #7786** — it needs `limits_known`.

Targets `main` rather than #7786's branch because a cross-fork PR cannot use a fork branch as its base.
Until #7786 merges, the diff below also contains the two commits before it; review only `fix(providers): report output capacity…`.
Merge #7786 first and this shrinks.

## What this changes

The providers table's "max output tokens" column reports what a provider can emit, instead of what somebody asked it for.

## Why

The column answered with a per-model `ModelOverrides.max_tokens` when one was set, falling back to the catalog capacity otherwise.

That publishes a **preference** under a **capacity's** name.
`max_tokens` is how long a reply to *ask* for; `max_output_tokens` is how long a reply the endpoint can *produce*.
An operator reading "this provider gives at most N" was being shown what somebody had requested.

It was also not a correct preference display, which is what settles it.
An unset override does not mean "the catalog maximum will be requested" — it means the resolution chain falls through to `DEFAULT_MODEL_MAX_TOKENS` (4096).
So the old value was neither the capacity nor the effective request length.
It was **a number with no referent**, in a column an operator reads to make a decision.

## The two rules it dragged with it

The dashboard field that writes that override was built on the same false premise, and both of its rules are worse than the column:

- It was seeded with `override ?? catalogMax`, so an unset override displayed the capacity **as though somebody had chosen it**.
- Typing a value equal to the catalog capacity was treated as "same as the default, so clear it". It is not the default, so that rule **silently discarded a deliberate setting** and returned the model to 4096. An operator who wanted exactly that value could not set it.

That is the same family of defect as the column: it saves, it says yes, and it does not arrive.

## Substantive changes

- `provider_max_output_tokens` reports the representative model's capacity and nothing else. The `ModelOverrides.max_tokens` read is gone.
- The result is gated on `limits_known`, so an unsourced placeholder is not published as a provider's headline figure — the same category error from the other side (#7780). The dashboard renders "-".
- `limits_known` is exposed on `GET /api/models` and `GET /api/models/{id}` so surfaces can tell the difference.
- The editing field stays. `ModelOverrides.max_tokens` is a real preference and now feeds the agent resolution chain as the fallback when an agent inherits — it simply is not this column.
- The field is relabelled for what it does, and the column gets its own label (`providers.output_capacity`), because the two now mean different things.
- The draft rule moves to `src/lib/modelOverrideDraft.ts` so the save payload and the Save-button gate cannot drift apart, following the precedent `agentModelPatch.ts` set for exactly this. The model's capacity is not a parameter of that function, so no value of it can make a typed number vanish.
- `providers.max_tokens_hint_override` is deleted; it had no remaining caller.

## Verification

- [ ] `cargo clippy --workspace --all-targets`
- [ ] `cargo test -p librefang-api --test providers_routes_test`
- [ ] `pnpm vitest run`

Tests:

- The #6209 test that pinned the old contract is **rewritten, not dropped**: it now asserts the override does not move the column, and that the override is still stored and readable — because it is a real setting that feeds the agent chain.
- A second test seeds two providers whose entries differ in exactly one field, `limits_known`, so the assertion isolates that and nothing else.
- `modelOverrideDraft.test.ts` covers blank-clears, set, the typed-value-equals-capacity case that used to vanish, validation, and whitespace.

Refs #6209, #7780, #7781


DaBlitzStein:
Superseded — `limits_known` is on `origin/main`, and issue #7780 is closed with it.

Verified: `crates/librefang-api/src/routes/providers.rs` carries the flag through every model-shaped response — `:248` (the placeholder row), `:430` and `:670` (list and detail), `:1335`, and `:1471`.

The distinction this branch was opened for is therefore already expressed: the numeric capacity still flattens to `0` via `lim.context_window.unwrap_or(0)`, but `limits_known: false` sits beside it and tells the caller that the zero is a placeholder rather than a measurement. That is the same contract this branch proposed, reached by keeping the numeric field's shape stable and adding the flag next to it.

Closing rather than rebasing.

DaBlitzStein:
Cerré este PR el 2026-08-26 sin mergear, y su `modelOverrideDraft.ts` nunca llegó. Lo he medido hoy contra `origin/main` (`bcb932fba`): la regla que corregía sigue viva, así que esto no era código huérfano sino un bug en producción.

**Sigue vivo en `ProvidersPage.tsx:225`:**

```ts
const targetOverride = parsed != null && parsed !== catalogValue ? parsed : null;
```

con `catalogValue={catalogMaxOut}`. Un valor igual a la capacidad del catálogo se leía como «igual que el default, bórralo». No es el default: un `max_tokens` ausente resuelve por `inference_params.rs` a `DEFAULT_MODEL_MAX_TOKENS`, que son 4096 (`agent.rs:914`), y el catálogo no se consulta en ningún punto de esa cadena.

Dos formas de morder, las dos medidas:

- **Sin override guardado**, el campo se siembra con la cifra del catálogo, así que teclear ese mismo número dejaba el campo «no sucio», Guardar deshabilitado y el número imposible de guardar: un cuadro que muestra 16384 negándose en silencio a guardar 16384.
- **Con un override guardado**, elegir la cifra del catálogo escribía `overrides: {}`, borrando el ajuste del operador y dejando el modelo en 4096.

Rescatado en **#8368**, con la regla extraída a `modelOverrideDraft.ts` y aplicada también a `context_window`, que tenía el mismo camino de borrado silencioso. Añadí el seguimiento de si el operador tocó el campo, que tu versión no necesitaba porque en aquel momento el editor no sembraba desde un valor de presentación — sin él, abrir el drawer ya dejaba el campo sucio y un clic escribía la capacidad del catálogo para todo agente que nunca fijó un `max_tokens`.

Y corregidos los dos textos de ayuda de `max_tokens`, que decían que dejarlo en blanco usa el valor del catálogo. Es falso por la misma razón, y la interpolación imprimía la capacidad del catálogo como si fuera a donde revierte limpiar el campo.

PR #7796, opened 2026-08-22
fix(api): apply provider-only agent config updates

### LINKED ISSUE #7765: bug: changing only an agent's provider is accepted with 200 and discarded
## What happens

`PATCH /api/agents/{id}/config` applies the provider only when the same request also carries a non-empty model.

`crates/librefang-api/src/routes/agents/config.rs:888-901`:

```rust
if let Some(ref new_model) = req.model {
    if !new_model.is_empty() {
        let explicit_provider = req.provider.as_deref().filter(|p| !p.is_empty());
        if let Err(e) = state.kernel.set_agent_model(agent_id, new_model, explicit_provider) { … }
    }
}
```

So `PATCH {"provider": "litellm"}` on its own returns `200` and changes nothing. The field is read, filtered, and dropped, and the caller has no way to tell.

## Why it is worth flagging

The comment directly above that block documents the fix for #2380 — a case of fields silently discarded behind a `200` — and lists the fields that were rescued. The same shape survived two lines below it, for the field the endpoint is most likely to be called with on its own.

The dashboard happens not to trigger it: its save button is gated on both provider and model being non-empty (`crates/librefang-api/dashboard/src/pages/AgentsPage.tsx:944-946`). Any other client hits it — the TUI, a script, anything driving the API directly.

## Expected

A request carrying a provider applies it, keeping the agent's current model.

Worth routing through `set_agent_model` (`crates/librefang-kernel/src/kernel/agent_state.rs:84`) rather than around it, since that is where the provider-change semantics live: stripping the provider prefix, clearing stale per-agent `api_key_env` / `base_url`, and the canonical-session cleanup. Bypassing it was the original cause of #2380.

## Related gap

This was diagnosed from a live install where an operator reported a provider selection not persisting, and it could not be settled either way: the daemon logs method, path, status and latency, never the request body. There was no way to know what the browser had sent.

Debug-level logging of the body on `PATCH /api/agents/{id}/config` and `POST /api/agents` would make this class of report answerable from the log. It needs scrubbing first — those bodies carry `api_key_env`, `base_url` and system prompts — and `scrub_500` in the same module is the existing precedent for that.

houko:
## Summary

- Apply a non-empty provider from `PATCH /api/agents/{id}/config` even when the request omits the model, using the agent's stored model as the unchanged half of the update.
- Keep all model/provider changes on `set_agent_model`, preserving provider-prefix normalization, stale credential/base URL cleanup, persistence, and canonical-session invalidation.
- Map typed kernel errors consistently so provider-only, model-only, and combined requests all return 404 for an unknown agent, including a concurrent deletion after the provider-only pre-read.

## Root cause

The handler nested provider handling inside the non-empty model branch. A provider-only request therefore returned 200 without calling the kernel or changing any state, even though the request schema accepted the field.

## Verification

- The new provider-only regression test failed before the fix because the response was 200 while the stored provider remained `openai`.
- The unknown-agent regression test failed before typed error mapping because a model-only request returned 500 while a provider-only request returned 404.
- `cargo test -p librefang-api --lib routes::agents::monitoring_tests::patch_agent_config_ -- --nocapture` — 2 passed.
- `cargo test -p librefang-api --lib routes::agents::monitoring_tests` — 13 passed.
- `cargo clippy -p librefang-api --lib --tests -- -D warnings` — passed.
- `cargo fmt --all -- --check` — passed.
- `git diff --check` — passed.
- `python3 scripts/check-changelog-attribution.py --all-unreleased` — passed.

## Out of scope

- Sensitive request-body debug logging remains unchanged; adding it requires a separate redaction contract.
- The full API lib suite was not repeated on this branch because `main` still contains the known vault-key race fixed separately by #7793; the affected agents module and both new regression paths were run directly.

Closes #7765

### DIFF
--- a/crates/librefang-api/src/routes/agents/mod.rs
+++ b/crates/librefang-api/src/routes/agents/mod.rs
@@ -1927,6 +1927,83 @@ mod monitoring_tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
+    async fn patch_agent_config_provider_only_preserves_model_and_persists() {
+        let (state, _tmp) = monitoring_test_app_state();
+        let manifest = AgentManifest {
+            name: "patch-provider-only".to_string(),
+            model: librefang_types::agent::ModelConfig {
+                provider: "openai".to_string(),
+                model: "gpt-4.1".to_string(),
+                api_key_env: Some("OLD_PROVIDER_API_KEY".to_string()),
+                base_url: Some("https://old-provider.invalid/v1".to_string()),
+                ..Default::default()
+            },
+            ..AgentManifest::default()
+        };
+        let agent_id = state.kernel.spawn_agent_typed(manifest).unwrap();
+        let request = serde_json::from_value(serde_json::json!({"provider": "litellm"})).unwrap();
+
+        let (status, body) = json_response(
+            patch_agent_config(
+                State(state.clone()),
+                Path(agent_id.to_string()),
+                None,
+                Json(request),
+            )
+            .await,
+        )
+        .await;
+
+        assert_eq!(status, StatusCode::OK, "response: {body}");
+        let entry = state.kernel.agent_registry().get(agent_id).unwrap();
+        assert_eq!(entry.manifest.model.model, "gpt-4.1");
+        assert_eq!(entry.manifest.model.provider, "litellm");
+        assert_eq!(entry.manifest.model.api_key_env, None);
+        assert_eq!(entry.manifest.model.base_url, None);
+
+        let persisted = state
+            .kernel
+            .memory_substrate()
+            .load_agent(agent_id)
+            .unwrap()
+            .unwrap();
+        assert_eq!(persisted.manifest.model.model, "gpt-4.1");
+        assert_eq!(persisted.manifest.model.provider, "litellm");
+        assert_eq!(persisted.manifest.model.api_key_env, None);
+        assert_eq!(persisted.manifest.model.base_url, None);
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn patch_agent_config_model_and_provider_shapes_return_not_found_consistently() {
+        let (state, _tmp) = monitoring_test_app_state();
+        let missing_id = AgentId::new();
+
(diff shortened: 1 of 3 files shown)


crates/librefang-api/src/routes/agents/mod.rs
```diff
@@ -1926,6 +1926,83 @@ mod monitoring_tests {
         );
     }
 
+    #[tokio::test(flavor = "multi_thread")]
```
houko, line 1929: Per `docs/development/build-and-verify.md` ("Integration testing"), a route/wiring behavior change like this one is supposed to get a `#[tokio::test]` against `TestServer` in `crates/librefang-api/tests/agents_routes_integration.rs` (which already covers other `PATCH /config` shapes — see `test_patch_config_applies_api_key_env_and_base_url` there), exercising the real axum router over HTTP with `reqwest`. These two regression tests call `patch_agent_config` directly instead, matching the existing `monitoring_tests` style already used elsewhere in this file.

That in-process style does verify the fix and the status-code consistency, so I'm not confident this needs to block the PR — just flagging since the docs state reviewers should gate on a `TestServer`-based integration test for a route-shape change (the error-mapping change here does alter response status codes for existing model-only/combined requests, not just the new provider-only path). Worth a maintainer call on whether the existing coverage here is sufficient or whether a `tests/agents_routes_integration.rs` case should be added too.

---

houko (commented):
Reviewed against CLAUDE.md. Pushed two mechanical fixes on top of the existing commit: corrected the changelog fragment's trailing reference from the closed issue (#7765) to this PR's number (#7796), which is what the release automation matches to suppress the generated line, and split a two-sentence-per-line comment into one sentence per line per the repo's prose-wrapping convention. `cargo check -p librefang-api --lib`, `cargo clippy -p librefang-api --lib --tests -- -D warnings`, and `cargo test -p librefang-api --lib routes::agents::monitoring_tests` (13 passed) all still pass. One inline comment left on the test-location question — ambiguous enough to leave for a maintainer call rather than fix unilaterally.

---

PR #7801, opened 2026-08-22
fix(dashboard): authenticate protected image loads

### LINKED ISSUE #7738: Dashboard cannot display /api/uploads images when api_key is configured (401 on every <img>)
### Description

In any deployment that configures authentication (`api_key` in `config.toml`),
images served from `/api/uploads/{file_id}` — produced by `browser_screenshot`,
`image_generate`, or user attachments — **cannot be displayed in the dashboard**:
every request 401s.

Symptoms:
- Dashboard chat shows a broken image for screenshot/image tool outputs.
- Opening the URL directly in a browser (logged in or not) returns:
  ```
  {"error":"Missing Authorization: Bearer <api_key> header"}
  ```
- `curl -H "Authorization: Bearer <api_key>" ...` serves the file fine — so the
  file exists and the handler works; only the *client side* cannot authenticate.

Root cause — auth middleware (`crates/librefang-api/src/middleware.rs`):
1. `/api/uploads/{file_id}` is **not** in any public allowlist
   (`PUBLIC_ROUTES_GET_ONLY`, dashboard shell, `PUBLIC_ROUTES_DASHBOARD_READS`)
   → the request always reaches the auth check.
2. The deployment has `api_key` configured → the no-auth loopback fast-path is
   skipped → a credential is mandatory.
3. Accepted credentials are header/transport-based only: `Authorization: Bearer
   <api_key>`, `X-API-Key`, or WS subprotocol. **The dashboard session cookie is
   deliberately read only for `/dashboard/*` paths** (CSRF defense) — it is
   ignored for `/api/*`.
4. The dashboard renders chat images with a plain `">`
   (`dashboard/src/pages/ChatPage.tsx`, image-ref rendering) — `` cannot
   attach the `Authorization` header, and the cookie is rejected for API paths.
   → 401 → broken image.

Net effect: **in every authenticated deployment, generated images are invisible
in the dashboard**, and users cannot share them via link (401 without the key).

### Expected Behavior

Images from `/api/uploads` should be viewable by dashboard users. Two viable
fixes:
- (a) Server: gate `/api/uploads/{file_id}` reads under the same
  `require_auth_for_reads` public-read allowlist used by dashboard reads, so
  authenticated deployments that opt in can serve images without a header; or
- (b) Client: in `ChatPage`, fetch the image through the API helper (which
  attaches the key) and render it as an object URL, instead of a bare ``.

### Steps to Reproduce

1. Deploy with `api_key` set in `config.toml` (and dashboard auth enabled).
2. Ask an agent to run `browser_screenshot` (or generate an image).
3. In the dashboard chat, observe the broken image for the tool output.
4. Open `https://<host>/api/uploads/<file_id>` directly in a browser → 401
   `Missing Authorization: Bearer <api_key> header`.
5. Confirm the file is fine: `curl -H "Authorization: Bearer <api_key>"
   https://<host>/api/uploads/<file_id> -o out.png` → valid PNG.

### LibreFang Version

2026.7.31

### Operating System

Linux (x86_64)

### Logs / Screenshots

```
$ curl -o /tmp/shot.png http://aix.btaskee.internal/api/uploads/b41150a6-850d-49f7-870a-520ddd5ea60b
{"error":"Missing Authorization: Bearer <api_key> header"}

# Same request with the API key → serves the PNG (file exists, handler OK):
$ curl -H "Authorization: Bearer <api_key>" -o /tmp/shot.png \
    http://aix.btaskee.internal/api/uploads/b41150a6-... 
$ file /tmp/shot.png
PNG image data
```

houko:
Closes #7738

## Summary

- Fetch protected `/api/uploads/<id>` and `/api/media/artifacts/<id>` images with the dashboard's existing API credential, then render revocable object URLs.
- Apply the authenticated renderer to chat history attachments, pending upload previews, workflow image galleries, and media generation results.
- Keep remote and data image URLs on the direct rendering path so dashboard credentials are never sent to external origins.
- Abort in-flight requests and revoke object URLs when an image changes or unmounts.

## Verification

- `pnpm test` — 122 files, 1187 tests passed.
- `pnpm typecheck`
- `pnpm lint`
- `pnpm build` — 2049 modules transformed.
- Headless Chromium against the Vite development server — protected image requests carried `Bearer browser-secret`, image and link rendered the same blob URL, and the browser reported no console or page errors. The temporary harness was removed after verification.
- `python3 scripts/check-changelog-attribution.py --all-unreleased`
- `git diff --check`

## Out of scope

- Non-image attachment download links still require separate API-aware download handling. Anonymous access to uploads remains prohibited.

### DIFF
--- a/crates/librefang-api/dashboard/src/components/AuthenticatedImage.tsx
+++ b/crates/librefang-api/dashboard/src/components/AuthenticatedImage.tsx
@@ -1,0 +1,47 @@
+import { useEffect, useState, type AnchorHTMLAttributes, type ImgHTMLAttributes } from "react";
+import { fetchAuthenticatedImage, isAuthenticatedImagePath } from "../api";
+
+interface AuthenticatedImageProps extends Omit<ImgHTMLAttributes<HTMLImageElement>, "src"> {
+  src: string;
+  linkProps?: Omit<AnchorHTMLAttributes<HTMLAnchorElement>, "href" | "children">;
+}
+
+export function AuthenticatedImage({ src, linkProps, ...imageProps }: AuthenticatedImageProps) {
+  const [resolvedSrc, setResolvedSrc] = useState<string | undefined>(() => isAuthenticatedImagePath(src) ? undefined : src);
+
+  useEffect(() => {
+    if (!isAuthenticatedImagePath(src)) {
+      setResolvedSrc(src);
+      return;
+    }
+
+    const controller = new AbortController();
+    let objectUrl: string | undefined;
+    setResolvedSrc(undefined);
+
+    void fetchAuthenticatedImage(src, controller.signal)
+      .then((blob) => {
+        if (controller.signal.aborted) return;
+        objectUrl = URL.createObjectURL(blob);
+        setResolvedSrc(objectUrl);
+      })
+      .catch((error: unknown) => {
+        if (error instanceof DOMException && error.name === "AbortError") return;
+        setResolvedSrc(undefined);
+      });
+
+    return () => {
+      controller.abort();
+      if (objectUrl) URL.revokeObjectURL(objectUrl);
+    };
+  }, [src]);
+
+  const image = <img {...imageProps} src={resolvedSrc} />;
+  if (!linkProps) return image;
+
+  return (
+    <a {...linkProps} href={resolvedSrc}>
+      {image}
+    </a>
+  );
+}
--- a/crates/librefang-api/dashboard/src/pages/MediaPage.tsx
+++ b/crates/librefang-api/dashboard/src/pages/MediaPage.tsx
@@ -21,6 +21,8 @@ import { Badge } from "../components/ui/Badge";
 import { Button } from "../components/ui/Button";
 import { Input } from "../components/ui/Input";
 import { PageHeader } from "../components/ui/PageHeader";
+import { AuthenticatedImage } from "../components/AuthenticatedImage";
+import { isAuthenticatedImagePath } from "../api";
 import { useUIStore } from "../lib/store";
 import {
   Image as ImageIcon,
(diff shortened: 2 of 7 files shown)


crates/librefang-api/dashboard/src/components/AuthenticatedImage.tsx
```diff
@@ -0,0 +1,47 @@
...
+    if (!isAuthenticatedImagePath(src)) {
+      setResolvedSrc(src);
+      return;
+    }
+
+    const controller = new AbortController();
+    let objectUrl: string | undefined;
+    setResolvedSrc(undefined);
+
+    void fetchAuthenticatedImage(src, controller.signal)
```
houko, line 22: Per the dashboard's `AGENTS.md` ("Data layer — mandatory rules"): "Do NOT call `fetch()` or `api.*` directly inside a page or component file." `fetchAuthenticatedImage` is an `api.ts` export, and it's called here directly from a component's `useEffect`, bypassing the `src/lib/queries/` hooks layer entirely (no query key, no error-translation path shared with the rest of the app).

This slips past `no-inline-fetch.test.ts` only because that guard's regex (`/(?<[a-zA-Z0-9_.])fetch\s*\(/`) matches a bare `fetch(` call, not an `api.*` helper name like `fetchAuthenticatedImage(` — so the letter of the automated check is satisfied while the written rule in `AGENTS.md` is not.

The existing exception convention in this codebase (`AuditPage.tsx:122`, `ConnectWizardPage.tsx:96`) is a `// lint-disable-next-line dashboard/no-inline-fetch -- <reason>` comment directly above the raw `fetch()` call for cases that genuinely can't go through the query cache (blob downloads, one-shot probes). That precedent is for a documented, narrow exception with a stated reason — not silent.

Two ways to close this, and I'd leave the choice to you since it's a design call:
- Route this through a real hook in `src/lib/queries/` (e.g. wrapping the blob fetch in `useQuery`, with the object-URL creation/revocation handled via a `useEffect` on `data` rather than inside the fetch itself), consistent with every other domain in `queries/`.
- Or, if a cached blob genuinely doesn't fit the query-cache model (object URLs need imperative revoke tied to component lifecycle, so caching them is unsafe to share across consumers), keep the direct fetch but annotate it with the same `// lint-disable-next-line dashboard/no-inline-fetch -- <reason>` marker and a short comment explaining why, matching the established pattern — and note that today's `no-inline-fetch.test.ts` regex wouldn't actually see it either way, since it only scans for a literal `fetch(`, not `api.*` calls.

---

crates/librefang-api/dashboard/src/pages/MediaPage.tsx
```diff
@@ -343,26 +345,36 @@ function ImagePanel({
...
+                    className="w-full h-auto"
+                    linkProps={{
+                      target: "_blank",
+                      rel: "noopener noreferrer",
+                      className: "block rounded-xl overflow-hidden border border-border-subtle hover:border-brand/40 transition-colors",
+                    }}
+                  />
+                );
+              }
+              return (
```
houko, line 365: Branch condition changed from `img.url` (old code) to `imageSrc` (`isAuthenticatedImagePath(img.url) ? img.url : safeUrl(img.url)`). That's a real fix — the old code rendered ``'&lt;img src={img.url}&gt;'`` unsanitized even when `safeUrl()` rejected it — but it also changes what falls into this branch: if `img.url` is set but rejected by `safeUrl()` (e.g. a `javascript:` or otherwise disallowed scheme) and the provider response has no `data_base64`, this now falls through to ``'&lt;img src={'data:image/png;base64,${img.data_base64}'}&gt;'`` with `data_base64` undefined, i.e. `src="data:image/png;base64,undefined"` — a silently broken image instead of the "too large" warning or any indication the URL was rejected.

Fairly contrived (requires a provider response with an unsafe/non-authenticated `url` and no `data_base64`), so not blocking, but worth a guard (e.g. treat "url present but rejected" as its own case with a message) if that response shape is reachable in practice.

---

houko (commented):
Reviewed against `CLAUDE.md` / the dashboard's `AGENTS.md`. One rule-compliance finding on the API-hooks-layer contract (fetches route around `src/lib/queries/`), plus one correctness note on a fallback edge case in the same diff. No mechanical auto-fixes applied — both are judgment calls best left to the author. Everything else (changelog fragment format, commit messages/attribution, remote/data URL passthrough) checked out clean.

---

PR #7803, opened 2026-08-22
fix(media): make processing timeouts configurable

houko:
## Summary

- Add independent `[media] transcription_timeout_secs` and `ffmpeg_timeout_secs` settings, preserving the existing 60-second and 30-second behavior as defaults.
- Apply the transcription timeout to OpenAI-compatible, custom STT, Gemini, and ElevenLabs requests, and apply the ffmpeg timeout to OGA remuxing, video audio extraction, window extraction, and source probes.
- Reclassify media config reloads as restart-required because `MediaEngine` captures `MediaConfig` by value at boot.
- Expose both fields through `GET /api/config`, the generated config schema, dashboard labels in all five locales, and the English and Chinese configuration docs.

## Root cause

The transcription request and ffmpeg subprocess limits were hard-coded at their execution sites, so recordings could hit an invisible wall far below the documented media size limits. The reload planner also classified all media changes as immediately effective even though the live engine has no config rebuild path.

## Verification

- `cargo test -p librefang-runtime-media` — 108 passed
- `cargo test -p librefang-types --lib media::tests::` — 39 passed
- `cargo test -p librefang-kernel --lib config_reload::tests` — 48 passed
- `cargo test -p librefang-api --lib routes::config::manage::config_read_write_parity_tests` — 5 passed
- `cargo test -p librefang-api --test config_routes_integration get_config_exposes_writable_fields_that_used_to_be_write_only` — passed
- `cargo test -p librefang-api --test config_schema_golden kernel_config_schema_matches_golden_fixture` — passed
- `cargo clippy -p librefang-types -p librefang-runtime-media -p librefang-kernel -p librefang-api --all-targets -- -D warnings`
- `git diff --check`
- `python3 scripts/check-changelog-attribution.py --all-unreleased`
- Verified all four new dashboard translation keys are present and parse in en, ko, pl, uk, and zh JSON files.

## Out of scope

- Result windowing and `out_path` from item 2 of #6748 remain separate work owned by the issue author.
- The post-extraction audio-size bound from item 3 is unchanged.
- The existing Polish and Ukrainian locale files each contain eight plural-form keys beyond the English reference; this predates the PR and is not modified here.

Refs #6748


crates/librefang-api/dashboard/src/locales/en.json
```diff
@@ -2203,6 +2203,8 @@
     "fld_audio_prompt": "Audio Prompt",
     "fld_audio_provider": "Audio Provider",
     "fld_audio_transcription": "Audio Transcription",
+    "fld_transcription_timeout_secs": "Transcription Timeout (s)",
+    "fld_ffmpeg_timeout_secs": "FFmpeg Timeout (s)",
```
houko, line 2207: Nit: the `fld_*` keys in this section are alphabetically ordered (`fld_audio_transcription` → `fld_auto_approve` → `fld_base_url` → ...), and these two new keys break that ordering here. `fld_ffmpeg_timeout_secs` would sort between `fld_extraction_threshold` and `fld_fold_min_batch_size`; `fld_transcription_timeout_secs` would sort between `fld_totp_issuer` and `fld_usage_footer`. Same misplacement in `ko.json`, `pl.json`, `uk.json`, and `zh.json`. Not enforced by `i18n-parity.mjs` (it only checks key-set parity, not order), so this won't fail CI — just flagging since it's an established convention in this file.

---

houko (commented):
Applied one mechanical fix directly on this branch (changelog fragment referenced the tracking issue's number instead of this PR's number, which would have caused a duplicate auto-generated release-notes line). Left one non-blocking nit about locale key ordering.

---

PR #7812, opened 2026-08-23
feat(slack): forward inbound files and images to the agent

### LINKED ISSUE #7087: [Feature] Slack adapter: inbound file/attachment support (user → agent)
### Description

The Slack sidecar adapter is text-only on the **inbound** side: messages carrying
files/attachments (`event["files"]`, `event["files_remote"]`, image `attachments`) are
**silently dropped** — the user's image/video never reaches the agent. Outbound delivery
is covered by #6897 (`files.upload`), but inbound is not.

This blocks a very common content workflow: a user uploads an image/video into Slack
asking the agent to create or publish something, and the agent must **see the asset** to
act on it.

Verified in code:

- `slack.py` `on_send` only calls `chat_postMessage(channel=..., text=...)` — no
  `files.upload`, no image block, no `files_remote`.
- The event handler reads only text; file-bearing fields in the event payload are
  ignored, so user-uploaded files never produce a `Content.image` / `Content.file`
  message for the agent.
- The sidecar wire protocol already supports rich content (`Content.image(url, caption)`,
  `Content.file(url, filename)`, `Content.file_data(bytes, filename, mime_type)`,
  `Content.voice(...)`, `Content.video(...)`) — the gap is purely in the Slack adapter.

**Use cases (MKT team):** (1) member uploads a campaign image with *"Post this to
LinkedIn with caption X"* → agent sees the image (vision) and publishes; (2) video/audio
review — upload a video → agent reviews/transcribes → returns result; (3) after
completing, agent returns the final asset (resized image, post preview, document) —
outbound part → #6897.

**Proposal:**
1. **Inbound detect** — recognize file-bearing events (`files`, `files_remote`, image
   `attachments`) in the Slack event handler.
2. **Download & emit** — fetch the file via `files.info` / provided URL and emit
   `Content.image` / `Content.file` / `Content.video` / `Content.audio` so the agent can
   process the asset.
3. **Policy** — config-gated: size limit, extension allow-list, optional per-channel
   `download: true/false` (busy channels can disable to avoid spam).
4. **Tests** — unit test event parsing with files; integration test asserting the agent
   receives `Content.image` / `Content.file` correctly.

### Alternatives Considered

- **External-host + paste URL** — current workaround: files are dropped, members must
  upload to an external host and paste a URL. Very disruptive for a daily content
  workflow; loses the in-channel context and requires extra tooling.
- **Outbound-only support (#6897)** — solves delivering files from the agent, but does
  not let the agent *receive* user-uploaded assets, which is the primary need here.

### Additional Context

Related issues/PRs:

- #6897 — outbound `files.upload` (posted 2026-08-10, open) — companion of this request
- #5219 — sidecar protocol (content already models `File` / `FileData` / `Image`)
- #6730 — Slack adapter UX (reaction spam / oversized replies) — same `slack.py`
- Rust side already models `ChannelContent::File` / `FileData` — inbound only needs the
  adapter to emit the correct protocol; no kernel changes required
- Companion request: per-user identity + targeted notification (roster, DM-to-user,
  `owner_notice` routing) — delivering files to a specific user depends on that
  capability

houko:
Closes #7087

The Slack sidecar was text-only on the inbound side. A message carrying an upload arrives with subtype `file_share`, and the subtype filter in `parse_slack_event` dropped every subtype other than `message_changed`, so the user's image or clip was discarded before any content parsing ran.

## Changes

- `sdk/python/librefang/sidecar/adapters/slack.py`
  - `file_share` now shares the bare-message arm; every other subtype is still dropped.
  - New `parse_slack_files` / `_file_content` map a message's `files` array onto `Content.image` / `Content.video` / `Content.audio` / `Content.file`, keyed on `mimetype`, carrying `url_private_download` and using the message text as the caption. `duration_ms` becomes `duration_seconds` for video / audio.
  - One attachment per message (the wire carries one `ChannelContent`), extras counted in a warning; the attachment outranks the message text including a slash command. Both match the discord sidecar's precedent.
  - An upload with no comment is now a complete message: the empty-text drop only applies when there is no eligible attachment.
  - `SlackFilePolicy` holds the gate: `SLACK_FILE_DOWNLOADS` (bool, default `true`), `SLACK_FILE_MAX_BYTES` (default 10 MiB, the same cap as the outbound upload path), `SLACK_FILE_ALLOWED_EXTENSIONS` (empty = allow all, matching the `SLACK_ALLOWED_CHANNELS` convention already in this file), and the `SLACK_FILE_DOWNLOAD_CHANNELS` allow-list / `SLACK_FILE_DOWNLOAD_EXCLUDE_CHANNELS` deny-list pair. A non-integer `SLACK_FILE_MAX_BYTES` exits 2, following the `TWITCH_RATE_LIMIT_*` precedent; a value below 1 falls back to the default with a warning.
  - Files whose `mode` is `tombstone` or `hidden_by_limit` are refused — their `url_private` no longer serves bytes, and letting the daemon try only puts a `[File download failed]` line in the prompt.
  - All five knobs are declared as `SCHEMA` fields so the dashboard renders them.

- **Credential handling.** `url_private_download` 302s to a login page without the bot token, so the daemon needs it. The adapter declares `header_rules` for Slack's own file hosts (`files.slack.com`, `slack-files.com`); the daemon's `fetch_headers_for` (`crates/librefang-channels/src/sidecar.rs`) exact-matches the request host against those rules and attaches nothing for any other host. `_is_slack_file_url` applies the same pin adapter-side, refusing non-HTTPS URLs and URLs carrying userinfo (`https://files.slack.com@evil.example/x` resolves to `evil.example`), so a file registered through `files.remote.add` with a member-chosen `url_private` is never forwarded at all. The rules are only declared when forwarding is enabled, so `SLACK_FILE_DOWNLOADS=false` does not ship the token to the daemon.

- **Why the URL and not the bytes.** `download_media_blocks` in `crates/librefang-channels/src/bridge.rs` is what turns an inbound `Image` / `File` / `Voice` / `Audio` / `Video` URL into a vision image block, an audio transcription or a saved document path. Its match has no `FileData` arm (`_ => None` at bridge.rs:5873), and the inbound text arm renders `FileData` as `[User sent a local file: …]` (bridge.rs:4412), so an adapter that inlines inbound bytes delivers a placeholder and discards the payload. Forwarding the URL with a host-pinned auth rule is the mechanism the matrix sidecar already uses for MSC3916 media, and it needs no kernel change.

- **Deliberately not followed:** link-unfurl `attachments[].image_url`. That is preview metadata for a URL somebody pasted, not an upload, and following it would point the daemon at an arbitrary host. Noted in the module docstring.

- Docs: architecture note on `header_rules` and why inbound `FileData` is a dead end (`docs/architecture/sidecar-channels.md`); a "User Uploads" section, the `files:read` scope and the five env vars in the en and zh channel pages.

- Changelog fragment: `changelog.d/added/7087-slack-inbound-attachments.md`.

## Verification

`cd sdk/python && python3 -m pytest tests -q` — 2044 passed (was 2016 before this branch; 28 new cases). This is the exact invocation of CI's `Python SDK Tests` job (`.github/workflows/ci.yml`, `working-directory: sdk/python`, `python -m pytest tests -q`).

`cd sdk/python && python3 -m pytest tests/test_slack_adapter.py -q` — 185 passed.

New tests in `sdk/python/tests/test_slack_adapter.py`, all against the pure `parse_slack_event` / `parse_slack_files` entry points:

- `test_parse_event_file_share_emits_image_content` — a `file_share` event produces `{"Image": {url, caption, mime_type}}` with routing metadata unchanged.
- `test_parse_event_file_share_video_and_audio_and_document_variants` — mimetype dispatch and `duration_ms` → `duration_seconds`.
- `test_parse_event_file_share_oversize_is_rejected`, `test_parse_event_file_share_oversize_keeps_the_companion_text` — the size cap, and proof a rejected attachment does not swallow the text the user typed with it.
- `test_parse_event_file_share_disallowed_extension_is_rejected`, `..._allowed_extension_passes`, `..._extension_falls_back_to_filetype`.
- `test_parse_event_file_share_download_switch_off_drops_the_file`, `..._per_channel_exclude_list`, `..._per_channel_allow_list`.
- `test_inbound_attachment_parsing_performs_no_http` — replaces `_public_http_request` and `_http_request` with a function that raises, then parses with the switch on and off: the adapter never fetches an inbound file itself.
- `test_parse_event_file_share_rejects_non_slack_host` — an attacker host, a suffix lookalike (`files.slack.com.evil.example`), the userinfo trick and plain HTTP are all refused.
- `test_header_rules_pin_the_bot_token_to_slack_file_hosts`, `test_header_rules_absent_when_downloads_are_disabled` (asserts the token string appears nowhere in the `ready` event), `test_header_rules_surface_in_the_ready_event`, `test_is_slack_file_url_host_pinning` — the token is declared only for Slack file hosts.
- `test_parse_event_file_share_dropped_without_a_policy`, `test_parse_event_other_subtypes_are_still_dropped`, `test_parse_event_file_share_still_honours_self_skip_and_channel_filter` — no behaviour change for callers that pass no policy, and the existing filters still apply.
- `test_file_policy_defaults`, `test_file_policy_env_parsing`, `test_file_max_bytes_non_integer_exits_2`, `test_file_max_bytes_below_one_falls_back_to_default`.

No cargo command was run: this branch touches no Rust source.

## Left out

- **Multi-file uploads deliver one attachment.** The wire protocol carries one `ChannelContent` per message; `Content.media_group` exists but the inbound bridge has no `MediaGroup` arm in `download_media_blocks`, so it would land as a `[Media group: N items]` placeholder. Extras are logged rather than silently dropped.
- **A non-media file loses its companion text.** `ChannelContent::File` has no caption field, so `Content.file` cannot carry the message. Warned at the boundary, same as the discord sidecar. Giving `File` a caption is a change to `crates/librefang-channels/src/types.rs` and every adapter that constructs it — a different crate, so it is not folded in here.
- **Attachments with an unknown `size`** (external files sometimes omit it) are accepted; the daemon's own `channels_download_max_bytes` still bounds the fetch.

### DIFF
--- a/sdk/python/librefang/sidecar/adapters/slack.py
+++ b/sdk/python/librefang/sidecar/adapters/slack.py
@@ -17,9 +17,29 @@ Behaviour parity with the Rust adapter:
   must be ACK'd by echoing back ``{"envelope_id": "..."}``.
 * **Event handling**: only ``message`` and ``app_mention`` types
   produce ``message`` events. Subtype filter: bare messages pass,
-  ``message_changed`` extracts ``event.message`` (edit), every other
-  subtype is dropped (joins, leaves, file_share, etc.). Self-skip on
+  ``message_changed`` extracts ``event.message`` (edit), ``file_share``
+  passes as an ordinary message carrying ``files`` (#7087), every other
+  subtype is dropped (joins, leaves, topic changes, etc.). Self-skip on
   ``bot_id`` present OR ``user == bot_user_id``.
+* **Inbound attachments** (#7087): a message's ``files`` array becomes an
+  ``Image`` / ``Video`` / ``Audio`` / ``File`` content variant carrying
+  ``url_private_download``, and the adapter declares ``header_rules`` so the
+  daemon fetches that URL with the bot token — Slack's private file URLs
+  302 to a login page without it.
+  The URL is forwarded rather than the bytes because the daemon's media
+  pipeline is what produces a vision image block, an audio transcription or
+  a saved document path; inbound ``FileData`` is rendered as a text
+  placeholder and its payload discarded, so inlining bytes would deliver
+  nothing.
+  One attachment per message (the wire carries one ``ChannelContent``), the
+  attachment outranks the message text including a slash command, and the
+  message text rides along as the caption for every variant that has one.
+  Policy knobs: ``SLACK_FILE_DOWNLOADS``, ``SLACK_FILE_MAX_BYTES``,
+  ``SLACK_FILE_ALLOWED_EXTENSIONS``, ``SLACK_FILE_DOWNLOAD_CHANNELS`` and
+  ``SLACK_FILE_DOWNLOAD_EXCLUDE_CHANNELS``.
+  Link-unfurl ``attachments[].image_url`` is deliberately **not** followed:
+  it is preview metadata for a URL somebody pasted, not an upload, and
+  fetching it would point the daemon at an arbitrary host.
 * **Allowed channels**: empty list = allow all. When non-empty,
   channel must be in the list; DMs (``channel`` starts with ``D``)
   are exempt (the operator's per-user DM allowlist handles those).
@@ -56,6 +76,11 @@ Configure via ``[[sidecar_channels]]``::
     # SLACK_FORCE_FLAT_REPLIES = "false"
     # SLACK_REACTIONS = "true"
     # SLACK_PROGRESS_CARD = "true"
+    # SLACK_FILE_DOWNLOADS = "true"
+    # SLACK_FILE_MAX_BYTES = "10485760"
+    # SLACK_FILE_ALLOWED_EXTENSIONS = "png,jpg,pdf,mp4"
+    # SLACK_FILE_DOWNLOAD_CHANNELS = "C0123"
+    # SLACK_FILE_DOWNLOAD_EXCLUDE_CHANNELS = "C0789"
     # SLACK_ACCOUNT_ID = "workspace-prod"
 
 Secrets via ``~/.librefang/secrets.env``: ``SLACK_APP_TOKEN`` (the
@@ -77,6 +102,7 @@ import threading
 import time
 import urllib.parse
 import urllib.request
+from dataclasses import dataclass
 from typing import Any, Callable, Optional
 
 from librefang.sidecar import Content, Field, Schema, SidecarAdapter, protocol, run_stdio_main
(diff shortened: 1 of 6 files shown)


sdk/python/librefang/sidecar/adapters/slack.py
```diff
@@ -385,6 +408,177 @@ def parse_users_info(body: dict) -> tuple[Optional[str], Optional[str]]:
...
+    if not isinstance(url, str) or not url:
+        return False
+    try:
+        parsed = urllib.parse.urlsplit(url)
+        host = parsed.hostname
+    except (TypeError, ValueError):
+        return False
+    if parsed.scheme != "https" or not host:
+        return False
+    if parsed.username is not None or parsed.password is not None:
```
houko, line 475: Nit: the docstring on `_is_slack_file_url` calls out `https://files.slack.com@evil.example/x` by name as the reason for this userinfo check, but `test_is_slack_file_url_host_pinning` doesn't exercise that exact string — it only asserts the plain-host rejection cases. Worth adding one assertion for the userinfo form so the test suite actually covers the attack the comment documents (even though, as far as I can tell, the plain `host in SLACK_FILE_HOSTS` check would already reject it since `urlsplit(...).hostname` resolves to `evil.example` there — so this looks like a coverage gap rather than a live bug).

---

houko (commented):
Reviewed against CLAUDE.md. Solid, well-tested change (185/185 pytest, security design checked against the actual Rust `fetch_headers_for`/`download_media_blocks` it references). Pushed one mechanical fix: the new docstrings/comments were hard-wrapped at ~78 columns across sentence boundaries, which CLAUDE.md's prose-wrapping rule forbids — reflowed to one sentence per line, no logic changes. One minor test-coverage nit left inline.

---

PR #7818, opened 2026-08-23
feat(models): make the context window an editable model override

### LINKED ISSUE #7774: feat: model configuration, context window included, must be editable at any time from the WebUI, the TUI and the API

houko:
Closes #7774 (partially — see "Not in this PR").

## The problem

`ModelOverrides` carried 14 fields and none of them was `context_window`, so a model's window had exactly one persistence path: the catalog entry, which the next registry sync rewrites.
The dashboard documented this itself at `ProvidersPage.tsx:166` — "context_window has no persistence path in ModelOverrides and is overwritten on registry sync" — and worked around it by editing `max_tokens`, which is the per-request *output* cap, a different quantity with a similar name.

The consequence is the one the issue reports: a gateway proxying a self-hosted runtime reports `max_input_tokens: null`, the window stays unknown, `resolve_context_window` finds nothing, and the agent loop falls back to 8192 with a `WARN` telling the operator to set a field in every one of their agents' `agent.toml`.
The agent then refuses to continue over an overflow the provider is not enforcing.

## Substantive changes

- **`librefang-types`** — `ModelOverrides` gains `context_window: Option<u64>` and `max_output_tokens: Option<u64>`, both `skip_serializing_if = "Option::is_none"`, both in `is_empty()` (without which an overrides document carrying only a corrected window would be dropped by `set_overrides` the moment it was saved). New `EffectiveLimits` type next to `EffectiveCapabilities`.
- **Why those two, and not more** — they are the two capacity limits on `ModelCatalogEntry`, they are unknown from the same sources for the same reason (the issue's own litellm response reports `max_input_tokens: null` and `max_output_tokens: null` side by side), and neither had an operator path. Everything else on `ModelOverrides` is an inference parameter that already has one. `max_tokens` is deliberately left alone: it is the cap sent on the wire, not a fact about the model, and conflating the two is the confusion the issue is about.
- **`librefang-runtime`** — `ModelCatalog::effective_limits(entry)` and `effective_limits_for_manifest(provider, model)`. The second is the load-bearing one: overrides are keyed by `provider:model_id` and need no catalog entry, so it resolves for a model the catalog has never heard of — which is the reported case. It consults the manifest's own key first, then the reconciled entry's key (#6423 prefix reconciliation), so the value an operator typed against the id they can see is the one that wins.
- **`librefang-kernel`** — `resolve_context_window` now reads the override between the `agent.toml` value and the catalog. All four call sites (both `messaging.rs` paths, `agent_execution.rs`, and the compaction gate + context report in `agent_runtime.rs`) go through that helper, so they inherit the layer without changes; their precedence comments are updated.
- **`librefang-api`** — `GET /api/models`, `GET /api/models/{id}` and `GET /api/providers/{name}` report the **effective** `context_window` / `max_output_tokens`, with the raw catalog values preserved under a new `limits_catalog` object. This mirrors the existing `capabilities_catalog` treatment of `supports_*` exactly, and a revert-to-default control needs the unmerged value to name. `provider_max_output_tokens` (the #6209 headline) now consults `max_output_tokens` as a better fallback than the catalog, while keeping `max_tokens` ahead of it.
- **OpenAPI** — the three `/api/models/overrides/{id}` handlers were unannotated and therefore absent from the spec. They now carry `#[utoipa::path]` and are registered in `ApiDoc::paths(...)`.
- **Dashboard** — `ProviderMaxTokensSection` becomes `ProviderModelLimitsSection`: one model picker, two `ModelLimitEditor` instances (context window, max tokens). The stale comment is gone because the statement it made is no longer true. Reverting compares against `limits_catalog`, not the row's own now-effective `context_window` — reading the latter would make an active override look identical to the catalog default and silently delete itself on the next save. When no window is known for a model the drawer says so, so the assumption the runtime has been logging is visible next to the model it applies to.
- **Docs** — a "Per-Model Configuration Overrides" section in `configuration/providers/management`, en and zh, with the endpoint, the whole-document `PUT` semantics, and the precedence ladder.

## Precedence

Explicitly, for the context window:

1. `agent.toml` `[model] context_window` — the most specific layer, one agent stating its own window (#6568).
2. `model_overrides.json` `context_window` — the operator's correction for the model, inherited by every agent using it. **This PR.**
3. The catalog entry — whatever the registry declared, or a discovery pass assumed.
4. `8192` — the agent loop's conservative fallback, still logged.

A zero is "unknown" at every layer, never a limit, so a cleared field falls through instead of pinning the window to zero tokens and poisoning the budget math.

## Relationship to other in-flight work

- **#7780** (`fix/7780-discovered-model-capacity`) makes the discovery probe report real capacity instead of the hardcoded `131_072` in `merge_discovered_models`. That improves layer 3. This PR adds layer 2 above it and does not touch the probe — `an_operator_override_beats_the_discovered_context_window` is the test that pins the ordering between them, and it is written against a discovery-shaped entry so it stays meaningful once #7780 lands.
- **PR #7786** (DaBlitzStein's fork) adds a `limits_known` tri-state to `ModelCatalogEntry`. Complementary rather than overlapping: that flag describes what the *catalog* knows, this PR describes what the *operator* has corrected. Where they meet is `limits_catalog` in the JSON rows, which would be the natural place to also surface `limits_known` once #7786 merges. Nothing in either branch was edited from here.

## Verification

- `cargo check -p librefang-runtime --lib`, `cargo check -p librefang-kernel --lib`, `cargo check -p librefang-api --lib` — clean.
- `cargo test -p librefang-types --lib model_catalog` — 32 pass, including `overrides_file_without_capacity_limits_still_parses_as_absent`, `a_context_window_override_alone_is_not_an_empty_override_set`, `absent_capacity_limits_are_omitted_from_serialized_overrides`.
- `cargo test -p librefang-runtime --lib` — 8 new tests pass: `effective_limits_without_an_override_are_the_catalog_values`, `an_operator_override_beats_the_discovered_context_window`, `overriding_one_limit_leaves_the_other_on_the_catalog_value`, `a_zero_override_is_ignored_and_a_zero_catalog_value_reports_unknown`, `an_override_applies_to_a_model_the_catalog_does_not_know`, `an_unknown_model_with_no_override_resolves_to_nothing`, `both_the_manifest_key_and_the_reconciled_entry_key_are_honoured`, `a_limit_override_survives_a_catalog_reload`.
- `cargo test -p librefang-kernel --lib context_window_tests` — 15 pass (7 pre-existing, 6 new): `a_model_level_override_beats_the_catalog_value`, `a_model_level_override_resolves_a_model_the_catalog_does_not_know`, `the_agent_toml_value_still_wins_over_a_model_level_override`, `an_absent_limit_override_leaves_the_chain_unchanged`, `a_zero_model_level_override_falls_through_to_the_catalog`, `an_override_supplies_a_window_the_catalog_stores_as_zero`.
- `cargo test -p librefang-api --test providers_routes_test` — 62 pass, including the 3 new `#[tokio::test]`s against the real router: `context_window_override_flips_effective_value_on_every_model_surface`, `an_overrides_document_without_limits_leaves_reported_limits_unchanged`, `provider_headline_ranks_max_tokens_above_the_capacity_override`.
- Dashboard: `pnpm typecheck`, `pnpm lint`, `pnpm vitest run` — 122 files / 1191 tests pass, including 4 new `ProvidersPage` cases.
- `python3 scripts/check-changelog-attribution.py` — OK.

Not run locally: `dead_route_audit_test` and `openapi_path_coverage_test` (the drift guards for the new `#[utoipa::path]` annotations), and clippy. The shared build directory on this machine ran low on disk, so those are left to CI rather than forced through.
`openapi.json` and the generated SDKs are not regenerated here — the `OpenAPI Drift` job auto-commits that sync for same-repo PRs, and `cargo xtask` was outside the local build budget.

## Not in this PR

The issue asks for four surfaces; two are here (API, WebUI) and two are not:

- **A TUI models screen** (issue item 3). `crates/librefang-cli/src/tui/screens/` has no model screen at all, so this is a new screen rather than a field added to an existing one — a different crate and a self-contained piece of work.
- **Surfacing the assumed window in the *session* view** (issue item 5, backend half). The dashboard now flags a model with no known window, but the place the guess actually bites is `GET /api/agents/{id}/session/context`, whose `max_context_tokens` gives no hint of which layer produced it. Reporting that needs `resolve_context_window` to return its source and `context_report_for_session` to carry it — and that method is a `KernelHandle` trait method, so the signature change ripples through the trait, its impl and `kernel_api.rs`. Worth doing; it is a different change from the override layer and I did not want to bury it here. Say the word and I will open it as a follow-up rather than leaving it implied.

One unrelated pre-existing failure, reported rather than touched: `pnpm test:i18n-parity` fails on `main` for `pl.json` / `uk.json`, which carry eight `_few` / `_many` plural keys that `en.json` has no counterpart for. Those are legitimate CLDR plural categories for Slavic languages, so "remove the extra keys" is likely the wrong fix and the script's strict-parity rule is the thing to revisit. My keys are at parity across all five locales.


crates/librefang-api/dashboard/src/pages/ProvidersPage.tsx
```diff
@@ -163,74 +163,70 @@ function SetDefaultModelSection({ providerId, currentDefault, onSetDefault }: {
...
+  // that already equals the catalog value needs no override row, so treat it as
+  // "clear". A blank input also clears.
+  const targetOverride = parsed != null && parsed !== catalogValue ? parsed : null;
+  const dirty = targetOverride !== (overrideValue ?? null);
 
   const handleSave = async () => {
     if (!overrideKey || invalid) return;
     setSaving(true);
     try {
+      const next: ModelOverrides = { ...overrides };
```
houko, line 222: `ModelLimitEditor` is now instantiated twice (context window + max tokens) sharing the same `overrides` prop, and each instance's `handleSave` builds `next` from its own closure over that prop (`{ ...overrides }`) rather than the latest server state.

If a user edits both fields and clicks "Save" on each in quick succession — before the first mutation's cache invalidation re-renders `overrides` — the second save's `next` is spread from the pre-first-save snapshot, so it silently drops the first field's update (or restores its stale value) once it lands. This race didn't exist when there was only one editable field in `ProviderMaxTokensSection`.

Not flagging this as a hard blocker — it needs a deliberate double-interaction to trigger, and may be an acceptable tradeoff for now — but worth a decision: e.g. disable both Save buttons while either mutation is in flight (`updateOverrides.isPending` from either editor), or have the parent own a single in-flight `overrides` draft both editors read/write.

---

houko (commented):
Automated CLAUDE.md compliance pass. Checked deterministic prompt ordering, config-field completeness (`is_empty()`, serde attrs), route/OpenAPI registration, auth allowlist, integration-test coverage, dashboard hook/query-key conventions, changelog fragment format, prose wrapping, and the effective-limits precedence logic against their test suites. No mechanical fixes needed — one plausible correctness concern left inline for the author's judgment call.

---

PR #7819, opened 2026-08-23
feat(types): privacy pass over a manifest before registry promotion

### LINKED ISSUE #7771: feat: promoting an agent type to the registry needs a privacy pass over the manifest first
## Context

`origin/main` (`e61bbbca3`) exposes agent types read-only — `crates/librefang-api/src/routes/agent_templates.rs` registers only `GET /templates` and `GET /templates/{name}`.
Open PR #6931 adds local `POST`/`PUT`/`DELETE`.
Neither gives you a way to take an agent type you built on your own install and contribute it back to `librefang/librefang-registry`, which is the natural next step once operator-authored agent types exist and the obvious way for the registry to grow.

This issue is about that capability, and about one requirement it has to satisfy before it ships rather than after.

## The requirement

An agent type promoted from a running install is an `AgentManifest` that was written for one operator's machine, and publishing it verbatim would put that machine's details in a public repository.
On `origin/main` an `AgentManifest` carries, among portable fields:

- `workspace` — an absolute path into that operator's `~/.librefang/workspaces/`.
- `model.api_key_env` and `model.base_url`, and the same pair on every entry of `fallback_models` — which environment variable holds the operator's credentials, and any private or self-hosted provider endpoint.
- `workspaces[].mount` — absolute paths describing the host's filesystem layout.
- `exec_policy` — command and environment-variable allowlists, which are host-specific security policy.
- `context_injection` — free-form literal text, which is where internal documentation or customer names end up.
- `metadata` — an arbitrary operator key/value bag.

The registry's own validator requires only `name`, `description` and `module`, so none of the above is caught on the way in.
Once published, it is in git history.

## Suggested shape

Two separate operations, not one:

1. A sanitiser that returns a publishable copy with the instance-specific fields above removed or neutralised, leaving the portable half untouched — `name`, `description`, `model.system_prompt`, `model.provider` / `model.model`, `skills`, `capabilities.tools`, `channels`, `routing`, `tags`, `resources`, `priority`, `mcp_servers`, `thinking`, `compaction`.
2. A detector that reports what it found, with a bounded preview per finding, so the operator sees what is about to be dropped and confirms.

The detector should not be "whatever the sanitiser removes".
It also has to scan the fields the sanitiser *keeps*, because an operator can paste anything into a system prompt and no structural rule separates portable configuration from an internal hostname sitting inside free text.
Absolute paths, credential-shaped tokens and private hosts are worth flagging wherever they appear.

Keeping the two apart is the point: the operator decides, rather than having their file quietly rewritten on their behalf.

## Status

I have both functions written and unit-tested against `librefang-types` (roughly 600 lines including tests), but I have deliberately **not** opened a PR for them, because on `origin/main` there is nothing that could call them — it would be dead code landing ahead of the feature that needs it, and that is not a useful thing to hand a reviewer.

So: is this a direction you want?
If yes I will open it together with the promotion path, or immediately after #6931 if you would rather have the primitives reviewed separately.
If the answer is that agent types should not be publishable from an install at all, that is a fine answer too and I will drop it.

houko:
Closes #7771.

## What this is, and what it deliberately is not

The issue asks for the privacy pass that has to run *before* an operator-authored agent type can be contributed to `librefang/librefang-registry`.

I checked what the promotion path actually is on `origin/main` first, because the issue's premise matters: `crates/librefang-api/src/routes/agent_templates.rs` registers only `GET /templates`, `GET /templates/{name}` and `GET /templates/{name}/toml`; there is no promote/publish path for an agent type anywhere in `crates/` — not in the API, not in `librefang-cli` (`AgentCommands` has `new` / `spawn` / `list` / `chat` / `kill` / `delete` / `reset-uuid` / `merge-history` / `set`, no export or publish), and not in `librefang-extensions`. The only registry-contribution machinery that exists is `librefang-skills/src/registry_pr.rs`, and it is skill-shaped, not manifest-shaped.

So **the endpoint half awaits PR #6931**, which adds local `POST`/`PUT`/`DELETE` for agent types. This PR is the primitives plus their tests, wired at the one call site that exists today rather than landed as dead code.

## Substance

New module `crates/librefang-types/src/manifest_privacy.rs`, with the two operations kept separate exactly as the issue asks:

- `sanitize_for_publication(&AgentManifest) -> AgentManifest` — a publishable copy. Borrows the input and never writes a file.
- `scan_for_publication(&AgentManifest) -> Vec<Finding>` — what it found, with a bounded single-line preview per finding capped at 96 characters.

The detector is **not** a diff against the sanitiser. Each `Finding` carries `removed_by_sanitizer`: `true` means confirming publication is enough, `false` means the value sits inside a field promotion has to keep and the operator must edit it by hand. That second case is the one that matters — nothing structural distinguishes portable configuration from an internal hostname sitting in a system prompt.

### Classification

`CLASSIFICATION` in the module is the readable table; the doc comment carries the per-field reasoning. Summary:

- **Stripped — host path**: `workspace`, `workspaces` (both `path` and `mount` describe host layout, and the aliases are often the operator's own vocabulary).
- **Stripped — credential binding / private endpoint**: `model.api_key_env`, `model.base_url`, and the same pair on every `fallback_models` entry.
- **Stripped — host policy**: `exec_policy`, `capabilities.shell`, `capabilities.network`, `capabilities.ofp_connect`, `tool_exec_backend`, `context_engine` (plugin names, local Python hook paths, a sidecar endpoint).
- **Stripped — operator free text / metadata**: `context_injection`, `tools` (per-tool `params` is an untyped operator bag), `metadata`, `model.extra_params` and its `fallback_models` counterpart, `channel_overrides`, `author`.
- **Stripped — local wiring / state**: `triggers` (`target_agent` and `workflow_id` name records on this install), `is_hand`, `enabled`.
- **Portable**: `name`, `version`, `description`, `module`, `schedule`, `session_mode`, `model.provider` / `model.model` / `system_prompt` / sampling knobs, `skills`, `capabilities.tools` and the memory / spawn / message grants, `mcp_servers`, `channels`, `tags`, `routing`, `resources`, `priority`, `thinking`, `compaction`, `profile`, tool allow/blocklists, `allowed_plugins`, `response_format`, and the remaining behavioural knobs.

Two of those deserve a note because the issue's framing and the code disagree slightly, and the code is right: `mcp_servers` and `channels` are **names**, not bindings — MCP endpoints and headers, and channel tokens and chat ids, live in `KernelConfig`, never in a manifest. Stripping them would remove the reproducibility the type exists for while protecting nothing, so they travel. Conversely `author` is stripped even though the issue does not list it: an operator's `author` is routinely a local username or an e-mail address, and attribution for a contribution belongs to the pull request carrying it.

### Why the classification cannot silently rot

Three mechanisms, and the first is the one the issue asks for:

1. **Compile-time.** `sanitize_for_publication` destructures `AgentManifest` with no `..` rest pattern and rebuilds it with no `..Default::default()`. Adding a field to `AgentManifest` is therefore an `E0027` ("pattern does not mention field") on the destructure and an `E0063` ("missing field in initializer") on the construction, in this module, until someone names it on both sides and so decides what happens to it. The same holds for `ModelConfig`, `FallbackModel` and `ManifestCapabilities`, the three types whose interiors are reduced rather than kept or dropped whole. This is the mechanism the issue's "prefer an exhaustive match" asks for — a struct destructure without `..` gives the identical guarantee for a struct that a match without a wildcard arm gives for an enum.
2. **Runtime, for the readable table.** `classification_covers_every_serialized_manifest_field` asserts `CLASSIFICATION` and the serialized struct agree in **both** directions, so the table cannot fall behind the struct and a rename cannot leave a stale entry. The fixture populates every field carrying `skip_serializing_if` so all 58 keys are emitted.
3. **By construction, for the detector.** The retained-value scan walks the *serialized* publishable manifest rather than a list of field names, so a newly added portable field is scanned for pasted secrets from the moment it exists — no third place to update.

### Value-level scanners

The credential and PII shapes reuse `librefang_types::taint`'s existing denylist through `TaintSink::net_fetch()` (the sink that blocks `Secret` and `Pii`) rather than growing a second copy. Two adjustments were needed:

- The taint rules for well-known prefixes and opaque tokens anchor on the whole payload, so a key pasted mid-sentence is scanned token by token as well.
- `OpaqueToken` is skipped for values with no unbroken alphanumeric run of 20 or more characters. Without that gate it fires on every long model id — `accounts/fireworks/models/llama-v3p1-405b-instruct` and `claude-sonnet-4-20250514` both top out at eight — and a detector that cries wolf on `model.model` teaches the operator to click through the findings that matter. There is a test pinning that specifically.

Absolute-host-path and private-endpoint detection are new here. Both are documented as best-effort with their gaps stated: the POSIX path rule requires a known filesystem root, which means `/api/v1/messages` in a prompt is not reported and neither is a path under a non-standard root like `/nvme0/agents/`; the private-endpoint rule covers loopback, RFC 1918 and link-local literals (four octets required, so `10.0.0` is not read as an address), the `.local` / `.internal` / `.lan` / `.corp` family, and a single-label host behind an explicit scheme.

## Wiring

`GET /api/templates/{name}` gains an additive, read-only `promotion_preview` object: `findings`, `requires_review` (any finding the sanitiser cannot handle on its own), and `manifest_toml` — the scrubbed manifest, so an operator can attach it to a registry pull request by hand today. `manifest_toml` is `null` if the publishable copy cannot be rendered as TOML, with a `WARN`; that is not a reason to withhold the findings, which are the part that protects the operator.

Nothing existing changes: the raw `manifest_toml` at the top level is still the operator's file verbatim, and there is a test asserting that. The dashboard consumes only `/templates/{name}/toml`, so nothing in the SPA is affected and no locale strings are added.

## Verification

- `cargo test -p librefang-types --lib manifest_privacy` — **34 passed**. Includes one test per stripped category asserting the value does not survive; `no_planted_sentinel_survives_promotion`, which greps the serialized publishable manifest for all 19 planted sentinels at once; `portable_fields_survive_promotion` and `portable_collections_survive_promotion` for the benign half; `classification_covers_every_serialized_manifest_field`; `sanitizing_does_not_mutate_the_caller_s_manifest`; `publishable_manifest_round_trips_through_toml`; `detector_does_not_mistake_a_model_id_for_a_credential`; `detector_output_is_stable_across_runs`; `previews_are_bounded_and_single_line`; and shape-rule tables with both positive and negative cases.
- `cargo test -p librefang-api --test profiles_templates_routes_integration` — **13 passed**, 4 of them new `#[tokio::test]`s against the real router: the findings shape, the scrubbed-TOML round trip (host specifics absent, portable half present, raw file unmodified), a secret inside a retained system prompt setting `requires_review`, and silence on an already-portable template.
- `cargo check -p librefang-api --lib` — clean.

Not run: `cargo clippy`. Free space on the build host dropped from 45 GiB to 2.4 GiB during these runs (the shared target directory is contended by other concurrent sessions), so I stopped building rather than risk the machine and am relying on CI for the clippy gate.

One environment note for anyone reproducing locally: `librefang-api` cannot compile from a fresh worktree without `crates/librefang-api/static/react/` existing, because `include_dir!` panics on the missing directory. I created a gitignored placeholder; it is not part of this diff.

## Deferred

Only the half that cannot exist yet: there is no promotion endpoint or CLI command to gate, so nothing calls `sanitize_for_publication` on the way to a registry. Once #6931 lands its `POST`/`PUT`/`DELETE`, the promote path is `scan_for_publication` → show the operator the findings → confirm → `sanitize_for_publication` → the `registry_pr.rs`-shaped GitHub flow. I have not started that here because it would be guessing at #6931's shape.


crates/librefang-types/src/manifest_privacy.rs
```diff
@@ -0,0 +1,1818 @@
...
+    {
+        return true;
+    }
+    // A bare single-label host is only meaningful with an explicit scheme; without one, every ordinary word would match.
+    had_scheme && !lower.contains('.')
+}
+
+/// Split the host out of a URL-ish token, reporting whether a scheme was present.
+/// Returns `None` for tokens that cannot carry a host at all.
+fn split_host(token: &str) -> Option<(&str, bool)> {
```
houko, line 1001: `split_host`'s port-stripping can misparse a bare (unbracketed or non-schemed) IPv6 literal, letting a loopback/private address slip past `looks_like_private_endpoint`:

- For input `"::1"` (no scheme, no brackets — plausible as a raw manifest string value), `authority.rfind(':')` finds the *second* colon and strips `"1"` as a bogus port, leaving `host = ":"`. The later comparison against `"::1"` then fails to match, so the loopback literal isn't flagged.
- For `"[::1]:8080"` (bracketed host with an explicit port), the `authority.starts_with('[')` branch returns the whole `"[::1]:8080"` unstripped, so after `trim_start_matches('[').trim_end_matches(']')` you get `"::1]:8080"`, which also fails the `"::1"` equality check. It happens to still get caught today only via the "single-label host with scheme" fallback (since it still lacks a `.`), but a bare `"::1"` or `"[::1]:8080"` with no scheme goes undetected.

Since this module exists specifically to catch host-specific network literals before publication, a pasted IPv6 loopback/private address in a retained field (e.g. a system prompt or tag) would go unreported. Worth a look — scope/fix (special-casing bracketed vs. unbracketed IPv6 host+port splitting, plus test coverage) is a judgment call I'd leave to you rather than patch blind.

---

crates/librefang-api/src/routes/agent_templates.rs
```diff
@@ -237,8 +237,42 @@ async fn load_agent_templates(
...
+///
+/// `manifest_toml` is `null` when the publishable copy cannot be rendered as TOML.
+/// That is not a reason to withhold the findings, which are the part that protects the operator.
+fn promotion_preview(name: &str, manifest: &AgentManifest) -> serde_json::Value {
+    use librefang_types::manifest_privacy::{sanitize_for_publication, scan_for_publication};
+
+    let findings = scan_for_publication(manifest);
+    let requires_review = findings.iter().any(|finding| !finding.removed_by_sanitizer);
+
+    let publishable = sanitize_for_publication(manifest);
```
houko, line 275: Minor: `sanitize_for_publication` runs twice per request here — once inside `scan_for_publication` (via `collect_retained_findings`, line 255) to build the JSON walk, and again directly on line 258 for `manifest_toml`. Harmless given manifest size and this endpoint's traffic, but if you want to dedupe it, that's an API change to `manifest_privacy` (e.g. a combined `sanitize_and_scan` helper, or having `scan_for_publication` accept an already-sanitised manifest) rather than a local one-liner — flagging for your call rather than changing the module's public surface unilaterally.

---

houko (commented):
Automated review against CLAUDE.md. Pushed 3 mechanical fixes to this branch:
- fix(types): a clippy `explicit_counter_loop` failure in `truncate_preview` (was breaking `cargo clippy -- -D warnings`)
- docs(api) / docs(types): rewrapped hard-wrapped doc comments in the new `manifest_privacy` module and its API wiring to one sentence per line, per the repo's prose-wrapping convention

All 34 `manifest_privacy` unit tests and all 13 `profiles_templates_routes_integration` tests still pass after the fixes, and `cargo clippy -p librefang-types --lib` / `-p librefang-api --lib` are clean.

Left two inline comments on things I didn't fix unilaterally — a possible IPv6-literal detection gap in `looks_like_private_endpoint`/`split_host`, and a minor double-sanitization in `promotion_preview` — both are judgment calls on scope rather than safe one-line patches.

---

PR #7820, opened 2026-08-23
feat(memory): give agents a callable semantic-memory surface

### LINKED ISSUE #7808: Semantic memory has no agent-callable surface — agents get KV only, and `memory_search` silently resolves to exact-key recall
## Description

Semantic memory is a first-class subsystem in LibreFang — embeddings, cosine re-ranking, confidence decay, consolidation, duplicate detection — but **an agent cannot reach any of it**. The entire tool surface an agent gets for "memory" is three key-value operations against `kv_store`. Everything semantic is reachable only automatically (the recall the agent loop injects) or externally (the REST API).

The practical consequence: an agent can be *given* memories it never asked for, but can never *ask* for a memory. There is no "what do I know about X?" It can't check what it stored, can't clean up something wrong, can't add a durable fact on purpose.

### What exists today

| Surface | Semantic memory? |
|---|---|
| Builtin tools `memory_store` / `memory_recall` / `memory_list` | ❌ pure KV on `kv_store` |
| Automatic recall in the agent loop | ✅ but not agent-controllable |
| REST `/api/memory/**` | ✅ full CRUD + search + consolidate + stats |
| Agent-callable path to the above | ❌ **none** |

Confirmed by reading the tree:

*Line numbers below are against `main` @ `c5b0731`.*

- `crates/librefang-runtime/src/tool_runner/memory.rs:45,76,107` — `tool_memory_store` / `tool_memory_recall` / `tool_memory_list`, each forwarding to the `KernelHandle` method of the same name
- `crates/librefang-kernel/src/kernel/handles/memory_access.rs:77,150,185` — those forward to `substrate.structured_*`
- `crates/librefang-memory/src/structured.rs:47` → `SELECT value FROM kv_store WHERE agent_id = ?1 AND key = ?2`

`memory_recall` is an **exact-key lookup**. It never touches the `memories` table, never reads an embedding, never ranks by similarity.

- `crates/librefang-kernel-handle/src/memory_access.rs:13` — the `MemoryAccess` trait exposes exactly `memory_store:20`, `memory_recall:33`, `memory_list:45`, `memory_acl_for_sender:63`. **No search method exists on the trait at all**, so no tool can reach the semantic store even in principle.
- `ProactiveMemoryStore` is consumed by the REST layer (`librefang-api/src/routes/memory.rs`) and injected into the agent loop as an `Arc` (`agent_loop/prompt.rs`) — neither path has a tool-callable entry point.
- Even `DREAM_ALLOWED_TOOLS` (`crates/librefang-kernel/src/auto_dream/mod.rs`) is `["memory_store", "memory_recall", "memory_list"]` — the dream loop is KV-only too.

### Bug found while mapping this: `memory_search` is a trap

`crates/librefang-types/src/tool_compat.rs:23`

```rust
"memory_search" | "memory_recall" => Some("memory_recall"),
```

A model that calls `memory_search` — the single most natural name for the thing it wants — gets silently routed to **exact-key KV recall**. It does not error. It returns "not found" for a key that was never a key. The name promises semantic search and delivers a hashmap miss.

This is worth fixing regardless of the rest of this issue. Either make the alias resolve to a real search tool once one exists, or drop the alias so the call fails loudly instead of lying.

### Proposed tool surface

A naming note first, because this subsystem is already one bad name deep. The existing `memory_store` / `memory_recall` / `memory_list` are KV, and they own those names. Anything semantic needs a prefix that makes the substrate obvious at the call site, otherwise `memory_store` (KV, keyed) and a hypothetical `memory_remember` (semantic, keyless) read as synonyms and get picked at random — the same failure as `memory_search` below, just introduced deliberately. So: **`memory_semantic_*` throughout.**

Ranked by how much each unblocks:

1. **`memory_semantic_search(query, limit?, scope?, min_confidence?)`** — the core ask. Backed by `ProactiveMemoryStore::search_with_guard` (`proactive.rs:1781`), or `search_all_with_guard` (`:1830`) for cross-agent when RBAC permits.
2. **`memory_semantic_add(content, scope?, confidence?)`** — deliberate write to semantic memory. Today an agent can only KV-store or hope auto-extraction catches it. Backed by `add_with_guard` (`:1815`).
3. **`memory_semantic_forget(memory_id)`** — a wrong memory currently cannot be retracted by the agent that is being misled by it. Backed by `delete_with_guard` (`:1802`).
4. **`memory_semantic_stats()`** — self-introspection: how many memories, which scopes, what's decaying. Backed by `stats` (`:1286`) / `count` (`:1618`).
5. **`memory_semantic_consolidate()` / `memory_semantic_duplicates()`** — backed by `consolidate` (`:1574`) and `find_duplicates` (`:1661`). See the motivation note below on why these matter more than they look.

If the prefix reads as too long, the alternative is to give the semantic side a verb of its own (`recall_search` / `recall_add` / …) and leave `memory_*` to mean KV exclusively. Either works; what doesn't work is mixing both substrates under one prefix with no marker.

### Implementation seam

The underlying functions all already exist and are already exercised by the REST handlers — the HTTP layer in `routes/memory.rs` is a thin wrapper over `ProactiveMemoryStore`. **What's missing is the seam, not the logic.**

Per tool:

1. new method on `MemoryAccess` (`crates/librefang-kernel-handle/src/memory_access.rs:13`) forwarding to the store
2. impl in `crates/librefang-kernel/src/kernel/handles/memory_access.rs` (cf. `memory_store:77`)
3. name const in `mod tool_name` (`tool_runner/definitions.rs`, cf. `MEMORY_STORE:34`)
4. `ToolDefinition` entry in `builtin_tool_definitions()` (`definitions.rs:176`, cf. the `memory_store` entry at `:393`)
5. dispatch arm in `tool_runner/dispatch.rs` (cf. `"memory_store" =>` at `:231`)
6. decide on `ALWAYS_NATIVE_TOOLS` (`definitions.rs:128`) — `memory_semantic_search` probably belongs there; the rest can stay lazy behind `tool_search` / `tool_load`

The ACL story should mirror the KV tools: `enforce_memory_acl` + a namespace guard, cf. `tool_runner/memory.rs:45`.

## Alternatives Considered

**Current workaround: an MCP bridge.** I run a small stdio MCP server that wraps the REST API and re-exposes it as `memory_search` / `memory_add` / `memory_list` / `memory_consolidate`. It works, and it is exactly as fragile as it sounds — it produces the loop

```
daemon → stdio child → HTTP → same daemon on :4545
```

The failure that prompted this issue: the MCP child died and was not respawned. The agent lost all addressable long-term memory for a full day, while the data and the API behind it were healthy the entire time. Every `memory_search` came back as `Transport closed`, three consecutive failed iterations tripped the loop guard, and the agent hard-crashed mid-conversation. A subsystem that is present in-process should not be reachable only via a network round-trip to itself.

**Rejected: teaching agents to call the REST API via `web_fetch`.** Needs the API key in the agent's context, bypasses the tool ACL layer entirely, and gives up structured schemas.

### Motivation — semantic memory that can't be inspected can't be corrected

The bridge outage was the acute symptom, but the chronic one is worse. A stale memory ("service X is unavailable") keeps getting auto-recalled into context and keeps biasing the model, long after the underlying problem is fixed. I have watched an agent assert a tool was unavailable *in the same response in which it successfully called that tool* — the injected memory outvoted the tool result in front of it.

The agent cannot see that memory (no search), cannot correct it (no forget), and cannot merge the pile of near-duplicates reinforcing it (no consolidate). The only remedy available today is a full agent reset, which throws away everything else too. This is what makes items 3 and 5 above more than nice-to-haves.

### Secondary gap: no FTS index on `memories`

There is no `memories_fts` virtual table — the only FTS5 table in the schema is `sessions_fts` (`crates/librefang-memory/src/migration.rs:654`, rebuilt in v33). When no embedding is available, "search" degrades to a `content LIKE` scan (`semantic.rs:299`).

That also means the `fts_only` config knob has no FTS to fall back to for memories. Worth confirming whether that's intended; if the embedding provider is unreachable, memory search quietly becomes a substring scan.

### Recall ranking has no similarity floor

`recall_impl` (`semantic.rs:258`) over-fetches `(limit * 10).max(100)` candidates, re-ranks by cosine (`:552`), and truncates to top-k (`:563`, logged at `:565` as `Vector recall: N results from M candidates`) — **with no minimum-similarity threshold**. On a sparse memory set the top-5 will be filled with whatever exists, however unrelated. Vectors that fail to compare sink to `f32::NEG_INFINITY`, but merely-irrelevant ones are promoted on merit.

A `min_similarity` parameter on the proposed search tool (and ideally on auto-recall too, via `ProactiveMemoryConfig`) would let a caller ask for "nothing rather than noise". Related: `MEMORY_RECALL_LIMIT = 5` is hardcoded at `agent_loop/prompt.rs:309`.

### Environment

- Observed on LibreFang @ `0b83f9df6`; all source citations above re-read from `main` @ `c5b0731`
- macOS 15 (Apple Silicon), SQLite substrate, `text-embedding-3-small`

Happy to take a swing at the implementation if the shape above looks right — the seam is mechanical enough, and I'd rather get agreement on the tool naming (`memory_semantic_*` vs. a separate verb) and the `memory_search` alias question first.

houko:
Closes #7808.

## The gap

Semantic memory was a first-class subsystem with no agent-callable surface.
The whole tool surface an agent got for "memory" was three exact-key operations against `kv_store` (the `tool_name` block in `definitions.rs`); everything semantic was reachable only automatically (the recall `agent_loop::prompt` injects) or externally (REST `/api/memory/**`).
`MemoryAccess` had no search method at all, so no tool could reach the semantic store even in principle.

## How `memory_search` misled, precisely

`tool_compat.rs` mapped `"memory_search" | "memory_recall" => Some("memory_recall")`.
`normalize_tool_name` runs at the top of both `execute_tool_raw` and `execute_tool_with_sender_account`, so the rewrite happened before dispatch, before the allowlist check, and before anything the model could see.

The resulting chain was `memory_search` → `tool_memory_recall` → `KernelHandle::memory_recall` → `structured_get` → `SELECT value FROM kv_store WHERE agent_id = ?1 AND key = ?2`.
Three things made it a trap rather than a bug:

- **The argument shape does not match, and the mismatch is silent.** `tool_memory_recall` reads `input["key"]`. A model calling `memory_search({"query": "..."})` supplies no `key`, so `input["key"]` is `Value::Null`, `.as_str()` is `None`, and the call fails with `MissingParameter("key")` — an error naming a parameter the tool it *thought* it called does not have. A model that guesses `memory_search({"key": "deploy keys"})` instead hits the second failure below.
- **The failure is indistinguishable from an empty store.** On a miss `tool_memory_recall` returns `Ok(format!("No value found for key '{key}'."))` — not an error. A prose sentence about a "key" the model never chose, delivered as success.
- **The descriptions confirmed the wrong mental model.** `memory_recall`'s was `"Recall a value from the agent's memory by key."` — read as "this is how you look things up in memory". `memory_store`'s said "Store a value in the agent's memory... Use for persisting data across sessions". Neither said "key/value"; neither said "not a search". A model reading its own tool list had no way to tell the two stores apart.

The docs stated it outright: `docs/src/app/agent/tools/page.mdx` described `memory_recall` as **"Semantic search over the agent's memories"** and gave `memory_store` an argument list (`content`, `category`, `scope`) it has never accepted. Fixed here too, in both locales.

## Changes

**New tool surface.** The `memory_semantic_` prefix is load-bearing — `memory_store` (keyed) and a hypothetical `memory_remember` (keyless) read as synonyms and get picked at random.

- `memory_semantic_search(query, limit?, min_confidence?)` → `search_with_guard`. Default limit 5 (matches `MEMORY_RECALL_LIMIT`), clamped to `[1, 50]`. `min_confidence` is the "nothing rather than noise" floor from the issue; fragments with no stored confidence predate confidence tracking and are kept.
- `memory_semantic_add(content)` → `add_with_guard`, the same extraction pipeline `POST /api/memory` uses. Reports exactly what was persisted, and says **"Nothing was stored"** when the extractor declines — reporting a declined extraction as success would reproduce the failure this issue is about.
- `memory_semantic_forget(memory_id)` → `delete_with_guard`. Ownership is enforced downstream (`ProactiveMemory::delete` returns `Ok(false)` for another agent's memory), reported as "nothing was deleted".
- `memory_semantic_stats()` → `stats`.

Only `memory_semantic_search` joins `ALWAYS_NATIVE_TOOLS`, because the tool it competes with for the model's attention (`memory_recall`) already sits there. The other three stay behind `tool_search` / `tool_load`.

**Seam.** Four `MemoryAccess` methods with `Unavailable` defaults (the `WikiAccess` pattern), so every existing stub and embedder compiles unchanged; `#[async_trait]` added to the trait.

**Alias.** `memory_search` now resolves to `memory_semantic_search`; `memory_query` / `recall_search` join it, and `memory_add` / `memory_remember` / `memory_forget` / `memory_delete` map to their semantic counterparts. `memory_recall` still maps to itself. The four new names are added to `is_known_librefang_tool`.

**KV descriptions** now state what they do *not* do. `memory_recall`: "Look up one value by its EXACT key... This is not a search: it does no fuzzy, semantic, or substring matching... use `memory_semantic_search` instead."

**ACL** is resolved kernel-side and forwarded to the existing `*_with_guard` wrappers. This is the one place where copying the KV tools would have been wrong: `enforce_memory_acl` in the runtime does `check_read` / `check_write` only, so a runtime-side gate would have given namespace gating while **silently skipping `guard.redact_all`** — handing PII-tagged fragments to an LLM for a user whose policy forbids it — and would have missed `check_delete`'s `delete_allowed` flag.
`stats` has no `_with_guard` wrapper upstream, so it is gated inline on `check_read("proactive")` rather than being left as the one unguarded semantic read.
`AuthDenied` is mapped to `ToolError::PermissionDenied` so a refusal stays the soft `ToolExecutionStatus::Denied` and does not count toward the consecutive-hard-failure abort (the #5984 contract for the KV tools).
RBAC off (`memory_acl_for_sender` → `None`) yields a grant-everything ACL, matching `enforce_memory_acl`'s `Ok(())`; otherwise disabling RBAC would *tighten* memory access.

**Capability gating is applied twice, because once is not enough.** `capabilities.tools = ["memory_*"]` — the common declaration — glob-matches the new names, so the runtime allowlist cannot express "key/value yes, semantic no". `available_tools` therefore also consults the declared memory scopes: `memory_read` for `search` / `stats`, `memory_write` for `add` / `forget`.
The rules, each pinned by a test: undeclared (empty) scopes stay open, matching the reading `capabilities.tools` already has and keeping the default manifest able to reach the feature; a scope covers the store when it is `*`, `self.*`, or glob-matches `proactive`; an explicit `capabilities.tools` entry naming the tool overrides the gate, exactly as it does for the self-evolution tools.
`self.*` is accepted because every non-memory `ToolProfile` implies `memory_write = ["self.*"]`, which means "may write its own memory", not "barred from the semantic store" — matching it literally against `proactive` would have stripped the write tools from every profile-based agent.
All four are withheld when `[proactive_memory] enabled = false`, since the store is then never built (mirrors the existing browser / canvas gates).

**Determinism (#3298).** `MemoryStats::categories` is a `HashMap`. It is re-keyed through a `BTreeMap` before rendering — the stats result lands in the message history verbatim, and a reordered one invalidates the provider prompt cache for every later turn.

## Verification

- `cargo test -p librefang-runtime --test tool_runner_semantic_memory` — **19 passed**, new file. Argument forwarding and caller scoping; limit clamping (0→1, 7→7, 5000→50); rejection of missing/blank `query`, out-of-range `min_confidence`, blank/oversized `content`, malformed `memory_id`; the empty-result message naming which store was searched; `memory_search` landing on semantic search rather than KV recall; add reporting a no-op; forget reporting a miss; all four surfacing an ACL refusal as soft `Denied`; the `allowed_tools` allowlist blocking all four (through `execute_tool`, which is where that gate actually lives — not the raw dispatcher); `memory_*` admitting them; declarations present and `memory_semantic_search` in `ALWAYS_NATIVE_TOOLS`; `memory_recall`'s description disclaiming search.
- `cargo test -p librefang-kernel --lib -- semantic_memory scope_covers_own_memory explicitly_named_semantic` — **8 passed**. Every arm of the gate through the real path (`boot_with_config` → `spawn_agent` → `available_tools`): undeclared scopes open; wildcard open; `kv:*`-only withholds all four while the three KV tools survive; read and write halves gated independently; explicit declaration overrides while an undeclared sibling stays gated; subsystem-off strips all four; plus a coverage assertion that every `memory_semantic_*` name in `builtin_tool_definitions()` is classified, so a tool added later cannot slip through ungated.
- Determinism: `semantic_stats_rendering_is_byte_identical_across_insertion_orders` renders 24 categories in three insertion orders (forward, reversed, interleaved — reversal alone can pass for a container that happens to sort on read) and asserts byte equality plus actually-sorted keys. `builtin_tool_definitions_are_byte_stable` asserts repeated calls serialise identically and that no duplicate tool name makes the list order ambiguous.
- `cargo test -p librefang-types --lib -- tool_compat` — 3 passed (alias table).
- `cargo test -p librefang-import --lib -- test_tool_mapping test_is_known` — 2 passed (the mirrored assertion).
- `cargo check --lib` clean for `librefang-runtime`, `librefang-kernel`, `librefang-kernel-handle`.
- `cargo fmt` on the five touched crates.

**Not run: `cargo clippy`.** Clippy needs its own copy of the dependency graph, and free disk on this host was down to ~9 GiB; running it would have risked the machine for a lint gate CI already enforces. Flagged rather than glossed over.

**Not run: live daemon + real LLM** (human-only per CLAUDE.md). `memory_semantic_add` goes through the LLM extractor, so its distil-vs-decline behaviour against a real provider is worth one manual pass: with the daemon running, call `memory_semantic_add({"content": "the deploy key rotated in March"})`, then `memory_semantic_search({"query": "deploy key"})`, then `memory_semantic_forget` with the id the search returned, then `memory_semantic_stats({})`.

- **`DREAM_ALLOWED_TOOLS`** (`auto_dream/mod.rs`) is still `["memory_store", "memory_recall", "memory_list"]`, so the dream loop remains key/value-only even though Consolidate / Prune is exactly what these tools are for. Adding them — `memory_semantic_forget` especially — widens what an autonomous background loop may delete with no human in the loop. That is a policy call, not a wiring gap, so it is flagged rather than taken.
- **`memory_semantic_consolidate` / `_duplicates`** (item 5 in the issue). Same reason: `consolidate` merges and deletes rows across an agent's whole store, and whether an agent should trigger that on itself unattended is worth deciding before shipping the tool.
- **No `min_similarity`.** `MemoryItem` carries `confidence` but no similarity score — `recall_impl` re-ranks by cosine and discards the score before constructing items. A real similarity floor (and the `ProactiveMemoryConfig` counterpart the issue asks for) means plumbing the score through `librefang-memory`, a different crate from the seam this PR builds. `min_confidence` is the honest subset that needed no new plumbing.
- **No `memories_fts`.** Confirmed absent: the only FTS5 table is `sessions_fts` (`migration.rs`), so with no embedding provider "search" degrades to a `content LIKE` scan (`semantic.rs`) and `fts_only` has no memories index to fall back to. A schema migration is a separate change.

## Open questions for the maintainer

Both were the reporter's own open questions, and both are cheap to reverse now and expensive later.

1. **Naming.** This takes `memory_semantic_*` (the issue's first option) over a separate verb (`recall_*`). Renaming is cheap while only the alias table and four definitions reference the names; it gets expensive once agent manifests and skills name them.
2. **`memory_semantic_add` semantics.** Extraction-mediated (as shipped) means a deliberate "remember this" can legitimately store nothing — reported honestly, but still surprising for a tool an agent called on purpose. The alternative is a verbatim write via `add_with_level`, which needs a new `*_with_guard` wrapper in `librefang-memory`. Happy to switch, or to offer both behind a `verbatim` flag.


crates/librefang-kernel/src/kernel/handles/memory_access.rs
```diff
@@ -248,6 +295,109 @@ impl kernel_handle::MemoryAccess for LibreFangKernel {
...
+        let items = store
+            .search_with_guard(query, &agent.to_string(), limit, &guard)
+            .await?;
+        Ok(match min_confidence {
+            // A fragment with no stored confidence predates confidence
+            // tracking; keep it rather than silently deleting history from the
+            // caller's view when a floor is requested.
+            Some(floor) => items
+                .into_iter()
+                .filter(|i| i.confidence.is_none_or(|c| c >= floor))
```
houko, line 323: `limit` is applied by `search_with_guard` before this `min_confidence` filter runs, so the filter only ever narrows the top-`limit` ranked set — it never asks the store for a wider candidate pool to compensate. A caller passing `limit: 5, min_confidence: 0.8` can get back fewer than 5 items even when more than 5 fragments in the whole store would pass the confidence floor.

Not necessarily a bug — the tool doesn't promise "up to `limit` passing results" — but worth confirming this is the intended semantics (floor-that-narrows vs. floor-that-backfills), since a model asking for "5 confident memories" may silently get 1 or 2.

---

houko (commented):
Reviewed against CLAUDE.md's rules (deterministic prompt ordering, KernelHandle trait-injection pattern, config-field checklist, changelog format/attribution, commit conventions). This is an unusually careful, well-documented PR — pushed one small mechanical fix (a duplicate test assertion) directly to the branch; one inline nit below for a behavior worth confirming is intended.

---

PR #7822, opened 2026-08-23
fix(browser): fence preformatted blocks instead of folding them into prose

nevgenov:
`browser_read_page` loses every code block on a documentation page, the same way `web_fetch` does in #7821 — and for a different reason, in different code.

The walk in `EXTRACT_CONTENT_JS_TEMPLATE` has no branch for `pre`, so a listing falls through to the text-node branch, which trims every line. Measured against pages loaded in a live browser:

| page | `<pre>` in the DOM | fenced blocks before | after |
|---|---|---|---|
| Rust book | 15 | 0 | **15** |
| Python tutorial | 35 | 0 | **35** |
| Wikipedia (Rust) | 29 | 0 | **29** |
| Hacker News | 0 | 0 | 0 |

The text survives, but nothing separates it from the prose around it and its indentation is gone. A compiler diagnostic is the clearest loss, since the column markers that make it readable line up with nothing:

```
error[E0382]: borrow of moved value: `s1`
-->
src/main.rs:5:16
|
2 |     let s1 = String::from("hello");
  |         -- move occurs because `s1` has type `String`
```

## The change

A `pre` is taken as text rather than walked — descending is what trimmed it — and emitted as a fenced block, carrying any `language-` or `lang-` class from the element or from an inner `code`.

The fence is emitted for **any** `pre`, not only `pre > code`. Python's documentation marks listings up as highlighted spans with no `code` element anywhere, the RFC series uses a bare `<pre class="newpage">`, and so do diagnostics. Requiring `code` would leave all three exactly as they are today.

Link counts are identical before and after on every page measured, so the new branch takes nothing with it.

## Test

Added to the live-browser fixture harness from #7806, next to `live_extraction_preserves_structure_click_fallback_and_budget`, with a fixture carrying both shapes: a highlighted `pre > code` with a language class and a bare `pre` with no `code` element. It asserts the language reaches the fence, the indentation survives, both blocks are fenced exactly once, and the prose between them is untouched.

That harness needs a discoverable Chrome, which my machine does not have — both live tests fail here identically, including the pre-existing one on an unmodified `main` (`Chromium exited before printing DevTools URL`). I ran the same fixture through `chrome-headless-shell` over CDP by hand instead, and all five assertions hold:

```
Before.

```rust
fn main() {
    let x = 1;
}
```

Between.

```
>>> len(x)
3
```

After.
```

`2,295` tests pass otherwise; clippy `--all-targets` and fmt clean.

## Note

This is the same defect as #7821, fixed a second time because the two paths serialize independently — `web_fetch` parses an HTML string in Rust, the browser walks a live DOM in JS. They share nothing but `wrap_external_content`. Worth knowing, not worth blocking on: both fixes stand on their own.

Refs #6624

### DIFF
--- a/changelog.d/fixed/7822-browser-code-blocks.md
+++ b/changelog.d/fixed/7822-browser-code-blocks.md
@@ -1,0 +1,5 @@
+`browser_read_page` no longer folds code blocks into the prose around them.
+The walk had no branch for `pre`, so a listing fell through to the text-node branch, which trims every line: the text survived, but nothing separated it from the surrounding prose and its indentation was gone, which is what makes a compiler diagnostic unreadable.
+A `pre` is now taken as text rather than walked, and emitted as a fenced block carrying any `language-` or `lang-` class from the element or from an inner `code`.
+The fence is emitted for any `pre`, not only `pre > code`, because Python's documentation marks listings up as highlighted spans with no `code` element, the RFC series uses a bare `pre`, and so do diagnostics.
+Measured against pages loaded in a live browser, counting `pre` elements in the DOM against fenced blocks in the extraction: the Rust book 0 of 15 to 15, Python's tutorial 0 of 35 to 35, and the Rust Wikipedia article 0 of 29 to 29, with link counts identical before and after on every page (#7822) (@nevgenov)
--- a/crates/librefang-runtime/src/browser.rs
+++ b/crates/librefang-runtime/src/browser.rs
@@ -1488,6 +1488,12 @@ const EXTRACT_CONTENT_JS_TEMPLATE: &str = r#"(() => {
         return text.replace(/⟨(\d+)⟩/g, '($1)');
     }
 
+    // Read a language hint out of a `language-rust` / `lang-rust` class, so the reader knows what it is looking at.
+    function codeLanguage(cls) {
+        const m = /(?:language|lang)-([A-Za-z0-9+#-]+)/.exec(cls || '');
+        return m ? m[1] : '';
+    }
+
     const lines = [];
     // Parallel to `lines`: whether that line is a bullet a nested `li` already emitted.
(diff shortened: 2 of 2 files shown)


nevgenov:
An observation from having written this fix twice, offered because it seems worth saying out loud rather than sitting on. It asks for nothing and neither PR depends on it.

#7821 and this one fix the same defect — a listing reaching the model with its whitespace gone — in two places, with no shared code between them. `web_fetch` parses an HTML string in Rust; the browser walks a live DOM in JS. They share `wrap_external_content` and nothing else. So the two fixes are genuinely different code solving the same problem, and both had to reason separately about the same things: that a `pre` must not be reflowed, that the fence needs a language hint, that requiring `pre > code` would miss Python's documentation and the RFC series.

What made that necessary is that the browser path serializes *inside the page*. Looking at the four things the extraction does, they do not all need to be there:

- picking the root genuinely needs a live DOM, and it is also the part where our own rule beats the general-purpose ones — a results page is not an article, and a reader-oriented extractor gets it wrong;
- turning a DOM subtree into text needs HTML, not a live DOM;
- the cap and the marker table operate on that text, so they need neither.

Which suggests the seam is between the first and the second, rather than where it currently sits — with selection staying in the page and returning `root.outerHTML`, and serialisation plus budgeting being ordinary Rust that a test can call with a fixture and no browser at all. Today three of those four live in a JS string that only a live Chrome can execute, which is also why the fixture harness you added in #7806 has to launch one.

I am not proposing that here and I am not asking for a decision. Both PRs stand on their own and are worth merging on their own; if the answer is "the duplication is the cheaper of the two costs", that is a reasonable answer and I will stop bringing it up. Mostly I wanted the observation on the record while both diffs are fresh, since it is the kind of thing that is obvious when you have just written the same fix twice and invisible six months later.

PR #7827, opened 2026-08-24
feat(models): discover a custom gateway's models, and trust the capability flags it reports

### LINKED ISSUE #7775: bug: models served by an OpenAI-compatible gateway never appear in the model list
## What happens

`GET /api/models?provider=litellm` returns zero models on an install whose litellm gateway is configured, reachable, and answering completions.

The gateway lists them on the standard endpoint:

```
$ curl -H "Authorization: Bearer $KEY" http://litellm…/v1/models
sensor-model-generic
sensor-model-generic-high
pakllm
rodela-testing-model
embedding-high
```

Five models, one call, the same `/v1/models` shape every OpenAI-compatible server exposes. LibreFang lists none of them.

## Why

`list_models` (`crates/librefang-api/src/routes/providers.rs:282-305`) refreshes a live catalogue for exactly two providers:

```rust
let openrouter_in_scope = provider_filter … == "openrouter";
if openrouter_in_scope { crate::openrouter_catalog::refresh_if_stale(&state.kernel).await … }

let everyapi_in_scope = provider_filter … == "everyapi";
if everyapi_in_scope { crate::everyapi_catalog::refresh_if_stale(&state.kernel).await … }
```

Everything else is served from the checked-in catalogue. A provider whose models are not in that catalogue — which is every self-hosted gateway, by definition, since the model names are the operator's own — has nothing to show.

This is not specific to litellm. vLLM, LM Studio, llama.cpp behind a proxy, LiteLLM, any OpenAI-compatible endpoint: all expose `/v1/models`, none are queried.

## What it cascades into

The model being absent from the catalogue is also why nothing knows its context window, and that is where it stops being cosmetic:

```
WARN Model not in catalog — falling back to conservative context window.
     model=sensor-model-generic-high fallback=8192
WARN Context overflow unrecoverable — suggest /reset or /compact
```

The agent stops answering at ~17k tokens against a model that handles the request fine — `prompt_tokens=16968` succeeded at the provider in the same minute. The limit exists only in our own assumption.

And the operator cannot correct it: the model is not selectable, not configurable, and the one editable field within reach is the agent's `max_tokens`, which is the output cap and a different quantity.

## Expected

Any provider with an OpenAI-compatible base URL gets its `/v1/models` queried, the same way OpenRouter and EveryAPI already get their catalogues refreshed, so its models appear in the list and can be selected and configured.

Two details worth deciding explicitly rather than by accident:

- Gateways commonly answer `/v1/models` with ids and nothing else — litellm returns `max_tokens: null`, `max_input_tokens: null`, `max_output_tokens: null` on `/v1/model/info` for these. So discovery gives a usable list but no limits, and the operator has to be able to supply them (see #7774).
- Discovery should not be assumed to be free or reliable: it is a network call to the operator's own infrastructure, and it should degrade to the checked-in catalogue rather than failing the request, which is what the two existing refreshers already do (`tracing::warn!` then continue).

## Related

#7774 covers making a model's configuration — the context window included — editable at any time from every surface. This issue is the step before it: without discovery, there is no model to configure.

DaBlitzStein:
Closes #7775.

## What this does

Two commits, both about the custom-gateway catalog.
The second fixes a defect in the first, folded into the same PR rather than shipped and patched afterwards.

### 1. Discover a custom gateway's models from its models list

`GET /api/models` was the one surface that never triggered discovery for an operator-defined OpenAI-compatible gateway.
`GET /api/providers/{name}` probes the endpoint and merges what it finds, so a litellm or vLLM proxy registered under its own provider id only ever populated after someone opened the provider page — and the model picker, which is where those models are needed, stayed empty.

- `custom_gateway_catalog::refresh_if_stale` reads the gateway's `/v1/models` behind a TTL, so a listing does not become a network call per request.
- It is a no-op outside its scope: registry-known ids, EveryAPI, and any provider with no base URL or no usable auth.
  That lets `list_models` call it over every candidate id without a pre-check.
- Same TTL, same per-base-URL retry window, and the same warn-and-degrade-to-static-snapshot behaviour on a down or slow gateway that the existing OpenRouter / EveryAPI refreshers use.
- When the gateway also exposes litellm's optional `/v1/model/info`, declared context-window, output-limit and per-token cost figures are used; when it does not, or reports them as `null`, the model still appears with the catalog's existing "unknown" sentinel rather than being hidden.

Companion to #7776, which is why the switch that enables this could not survive a restart in the first place — the two are independent halves of the same operator-visible symptom.

### 2. Trust the gateway's own capability flags

`parse_model_info` harvested only the four numeric figures and threw the capability flags away, so `build_catalog_entries` could only carry `supports_vision` / `supports_tools` forward from a prior catalog entry.
A freshly discovered model has no prior entry, so every discovered gateway model was born with every capability `false`.

That is not cosmetic.
The agent loop reads `supports_vision` before the call: a model marked blind gets its attached images stripped and a bare file path substituted in their place.
The model then opens that path with the browser or the shell and "describes" the image from its filename — an answer that looks like vision and is not.

`LiveLimits` now carries `supports_vision` and `supports_function_calling`.
An explicit flag from the gateway wins, because the gateway is the authority on its own models; `null` means "it did not say" and keeps whatever was already known — the same unknown-versus-zero discipline the numeric limits already use.

## Verification

- Four wiremock integration tests in `crates/librefang-api/tests/providers_routes_test.rs`: models appear after a live refresh, a gateway that is down keeps the previous catalog instead of failing the request, a second listing inside the TTL makes no second network call, and a provider with no compatible base URL never probes at all.
- Three unit tests in `custom_gateway_catalog::tests` for the capability flags: `a_vision_flag_from_the_gateway_reaches_a_freshly_discovered_model`, `a_null_capability_keeps_what_was_already_known`, `an_explicit_denial_from_the_gateway_wins_over_the_old_value`.
- Behavioural check on a live deployment against a litellm instance: before the second commit the catalog reported every discovered model as non-vision and image attachments were replaced by a file path; after it, the catalog matched litellm's own flags on all three models the instance serves.
- No API surface change: this adds no route and no schema, so no OpenAPI or SDK regeneration.


houko:
@DaBlitzStein — same situation as #7826 for the first half of this, and I want to be straight about it: I merged a competing implementation after you opened this PR. `2ac0e3ef7` (#7816, merged 2026-08-24 11:26Z) closed #7775, the issue this PR closes.

**Commit 1 is superseded, and not by a better version so much as a different one.** #7816 reuses the existing `provider_health::local_provider_probe_targets()` + `probe_provider_cached()` and the shared 60 s `ProbeCache` that `GET /api/providers` already fills, where this branch adds `custom_gateway_catalog.rs` with its own TTL and per-base-URL retry window. There is no mechanical resolution: taking main's line leaves your 805-line module unreferenced (clippy dead-code failure, and `lib.rs` still declares it); taking yours drops main's broader coverage of built-in local ids; taking both means two discovery paths probing the operator's gateway on every `/api/models`. That is a judgment about your design, so I stopped rather than guess. Nothing has been pushed to this branch.

**Commit 2 is not superseded, and it is a live bug on `main` today.** I verified it:

- `crates/librefang-runtime/src/provider_health.rs:16` — `DiscoveredModelInfo` carries only Ollama-shaped fields (`families`, `capabilities`). No `supports_vision`, no `supports_function_calling`.
- No reference to litellm's `/v1/model/info` anywhere in that file.
- `model_catalog.rs:48` — `infer_capabilities(name, families)` derives `supports_vision` from `families` containing `"clip"` plus name heuristics. For an OpenAI-compatible gateway `families` is `None`, so a gateway that *declares* `supports_vision: true` has no way to say so.

The consequence you described stands: a freshly discovered litellm model is born `supports_vision = false`, the agent loop strips its images and substitutes a file path, and the model "describes" the image from its filename. That is a wrong answer presented as a real one.

Fixing it belongs on main's `provider_health` / `DiscoveredModelInfo` machinery rather than on this branch, so it is a rewrite of your commit, not a rebase of it.

**Your call on how that lands.** I would rather you got the credit for it, so: if you want to open it as its own PR against `provider_health`, it is yours. If you would rather not, say so — or say nothing for a couple of days — and I will write it citing this PR and you as the finder. What I am not going to do is quietly reimplement it a third time.

Recommending closure of this PR as superseded only after commit 2 has a home. Not closing it myself.

houko:

Conflicting files (1), against `main` at `4c3059e4`:

- `crates/librefang-api/src/routes/providers.rs`

DaBlitzStein:
Closing recommendation from a rebase attempt onto current `main` — this PR is superseded, and the half that is not has moved to #7957.

**Commit `a3385c6fa` ("discover a custom gateway's models from the models list") is already upstream.** `GET /api/models` on current `main` calls `refresh_discovered_models(&state, provider_filter.as_deref())`, whose doc comment cites the same **#7775** and describes the same defect this branch set out to fix. The rebase surfaced it as a direct conflict on `crates/librefang-api/src/routes/providers.rs`.

Upstream's version is the better one and this branch should not compete with it: it reuses `local_provider_probe_targets` — *"the single definition of 'participates in discovery'"* — probes through the shared `ProbeCache` with its 60s TTL, and runs the probes concurrently with `join_all`, so it stays on the same path as the periodic probe loop. This branch instead added a separate `crates/librefang-api/src/custom_gateway_catalog.rs` with its own staleness logic: a second discovery mechanism running beside the first.

**Commit `d9308a062` ("trust the gateway's own capability flags") only edits that same `custom_gateway_catalog.rs`.** With the module gone there is nothing left for it to apply to — but the bug it described is real and still present on `main`, so I have filed it separately with the trace through current code: **#7957**.

The short version of that trace: the OpenAI `/v1/models` parser builds `DiscoveredModelInfo::bare(name)`, so `capabilities` is empty; `resolve_discovered_capabilities` therefore falls back to `infer_capabilities(name, families)`; and a gateway model named by its operator (`team-default`, an internal id) gets `supports_vision` guessed from that string. A wrong `false` reaches `redact_images_for_text_only` in the agent loop, which swaps the image for its on-disk path. The `unwrap_or(true)` above it fails open on a catalog *miss* but not on a catalog *hit* carrying a wrong `false`.

No objection to closing this. Nothing is lost: the discovery half is in `main`, and the capability half is #7957, which I am happy to implement against upstream's `refresh_discovered_models` path rather than re-landing a parallel module.

DaBlitzStein:
Superseded by upstream: the gateway model discovery and capability-flag trust landed in #7816 (`fix(api): discover models from OpenAI-compatible gateways in /api/models`) and `d9308a06` (`fix(models): trust the gateway's own capability flags`).

The merge commit `16dcf2dd` explicitly lists `#7827` among the synced PRs.

PR #7833, opened 2026-08-24
feat(tui): backups tab in settings — create, restore with selection, delete

### LINKED ISSUE #7832: feat(tui): the terminal has no backup surface — create, list, restore and delete are dashboard/API only
The zip backup system is reachable from the HTTP API and from the dashboard.
The terminal has nothing.

`librefang tui` has a Settings screen with Providers, Models and Tools sub-tabs; none of them mention backups.
So on a headless box — the deployment shape the TUI exists for — creating an archive before an upgrade, listing what archives exist, restoring one, or deleting an old one all require hand-writing HTTP requests against `/api/backup`, `/api/backups` and `/api/restore`.

That is the same four-surface gap the project closes elsewhere: an operator working over SSH should not have to drop to `curl` for an operation the dashboard exposes with a button.

## Expected

A Backups sub-tab in the settings screen that:

- lists what `GET /api/backups` reports, newest first, with size and creation time;
- creates an archive;
- restores the selected archive, with the same options the dashboard restore has (keep-my-config clone mode, and which components the restore touches);
- deletes the selected archive.

## Note on key bindings

If the component toggles are bound to the number keys, they collide with the sub-tab switch keys, so the tab needs an explicit way back — `SettingsState::sub` persists across leaving and re-entering the Settings tab, so a tab that swallows the switch keys without an escape is a one-way door for the rest of the session.

DaBlitzStein:
Closes #7832.

**Stacked on #7831** (`feat/restore-onto-existing-system`), which adds the `keep_config` / `components` fields this tab sends.
Please merge that one first; the diff below is only the last commit.

## What this does

The settings screen gains a Backups sub-tab on key `4`, completing the fourth surface for the backup system:

- lists what `GET /api/backups` reports, newest first, with size and creation time;
- `c` creates an archive;
- `r` restores the selected archive with the same options the dashboard has — space toggles "keep my config" (clone mode), `1`-`8` toggle which components the restore touches;
- `d` deletes the selected archive.

## The one-way-door problem, and what it costs

Binding the component toggles to `1`-`8` collides with the sub-tab switch keys, so those are skipped while the Backups tab is active.
`SettingsState::sub` survives leaving and re-entering the Settings tab, so without a way out the tab would be a one-way door for the rest of the session: the operator could reach Backups and never get back to Providers, Models or Tools.

`Esc` returns to the Providers sub-tab and requests its refresh.
It is in the hint line and covered by a test.

## Verification

Six unit tests in `crates/librefang-cli/src/tui/screens/settings.rs::backups_tests`:

- `tab_4_opens_backups_and_requests_a_refresh`
- `esc_leaves_the_backups_tab`
- `space_toggles_clone_mode`
- `number_keys_toggle_components`
- `restore_sends_the_selection_with_the_filename`
- `create_is_one_keypress`

Strings added to all four TUI locales (`en`, `ko`, `uk`, `zh-CN`), labels in human language with the key names in the hint line.

## Note on the `AppEvent` diff

Adding variants to `AppEvent` makes rustfmt expand the enum's struct variants (`AgentSpawned`, `AgentKilled`, `MemoryKvSaved`, `TemplateTomlLoaded`, `SecurityChainVerified`, `ChatModelLabelLoaded`) from their current single-line form, and does the same to `SettingsAction::SaveProviderKey`.
Those hunks are the formatter's output, not hand edits; `rustfmt --check` on the files is clean with them and fails without them.

## Verification not performed here

The workspace build was not run locally — a cross-compile was occupying the shared target directory.
CI is the gate for `cargo check` / `clippy` / the test run.


crates/librefang-api/src/routes/backup.rs
```diff
@@ -574,6 +576,40 @@ fn restore_backup_blocking(
...
+            Some("workflows")
+        } else if name_str.starts_with("data/") {
+            Some("data")
+        } else {
+            None
+        };
+        if keep_config && component == Some("config") {
+            continue;
+        }
+        if let Some(selected) = &components {
```
houko, line 603: `components: Some(vec[])` (present-but-empty) behaves very differently from `components: None` (key omitted) here, and both new UI surfaces added in this PR send the empty form by default.

Walking through the match: with `selected = []`, every classified `component` (`Some("config")`, `Some("cron_jobs")`, …) fails `selected.iter().any(...)` and falls into `Some(_) => continue`, so it is skipped. Only unclassified/manifest-adjacent entries (`component == None`) survive. So sending `"components": []` restores almost nothing, not "restore everything" as the surrounding UI implies.

That is exactly what happens by default from both surfaces this PR adds, before the operator touches anything:
- TUI: `SettingsState::backup_components` defaults to `Vec::new()` and is sent verbatim in `spawn_restore_backup`'s JSON body (`crates/librefang-cli/src/tui/event.rs`, the `"components": components,` line) — pressing `r` without first pressing any of `1`-`8` sends `components: []`.
- Dashboard: `RuntimePage.tsx`'s `restoreComponents` state also defaults to `[]` and is always included in the `mutateAsync` payload, with the same effect via `restoreBackup`'s `options?.components`.

The one path that actually gets "restore everything" is the *no options at all* case exercised by `restore_without_options_writes_every_component_back` (the `"components"` key is omitted entirely there) — which none of the new UI code paths can produce, since both always send the field.

Suggest treating an empty selection the same as "no filter" here, e.g. `if let Some(selected) = &components.as_ref().filter(|s| !s.is_empty()) { ... }`, or have both call sites send `None`/omit the field when nothing is toggled, matching the "empty = restore all" comment in `RuntimePage.tsx`.

---

PR #5232, opened 2026-05-18
feat(sidecar,sdk): telegram full sidecar parity (formatter + full inbound/outbound), stdlib-only

houko:
Follow-up to #5228. Full sidecar parity for telegram so the in-process
`crates/librefang-channels/src/telegram.rs` can be removed. Three
commits, one tight cluster:

1. `daaa2712` — kill the `requests` dependency (the #5228 wart).
2. `134ef485` — promote to `SidecarAdapter` with capability
   negotiation (typing/reaction/interactive/thread/streaming).
3. `987fd31a` — **full parity**: faithful, function-by-function ports
   of the audited Rust (not re-derived).

## What `987fd31a` ports (with the Rust oracle it is pinned to)

- **Markdown → Telegram-HTML formatter** — byte-exact port of
  `formatter::markdown_to_telegram_html` (+ inline/fence/heading/list
  helpers). Tests assert the *same input→output pairs* as
  `crate::formatter`'s own `mod tests`, so the two implementations
  cannot drift silently. Outbound text now formats + sanitizes +
  chunks and sends `parse_mode=HTML` (was a naive
  `parse_mode=Markdown`), with the same plain-text retry on Telegram's
  "can't parse entities" 400.
- **`sanitize_telegram_html` security pass** — tag + `<a href>`
  scheme allowlist, attribute-injection escaping, unclosed-tag
  balancing (`rebuild_safe_tag` / `parse_attrs` / `is_safe_href`),
  including the exact Rust edge where a dropped unsafe-`<a>` leaves
  its now-unmatched `</a>` escaped.
- **`message_truncator` chunker** — `utf16_len` /
  `truncate_to_utf16_limit` / `split_to_utf16_chunks` with the
  HTML-entity boundary guard and the zero-progress guards.
- **Full inbound** — text/bot-command, photo, document, audio, voice,
  animation, video, video_note, location, sticker; `from` /
  `sender_chat` extraction; `callback_query` → `ButtonCallback`
  (+ fire-and-forget `answerCallbackQuery`); `edited_message`;
  reply-to context; `getFile` URL resolution with text fallback;
  `ALLOWED_USERS` by id **and** username.
- **Full outbound** `ChannelContent` dispatch for every variant
  (Image→sendPhoto, File→sendDocument/sendVoice, Voice/Video/Audio/
  Animation, Sticker, Location, MediaGroup, Poll, Interactive,
  EditInteractive, DeleteMessage), private-URL → multipart upload,
[...]

PR #7848, opened 2026-08-24
fix(sidecar): make SDK version skew visible instead of silent

houko:
Refs #7140 — root cause 1 only. #7159 covers candidates 2/3 and the `/think` stub; #7701 covers the session dimension. Neither touches `sidecar.rs`, `embedded_sdk.rs`, `routes/sidecar_describe.rs`, or `sdk/`, and both are still scoped that way as of tips `6e34904` and `bd5e722`.

## The divergence

The sidecar protocol version is written down in five places and was enforced in none:

- `docs/architecture/sidecar-protocol.md:70` — "`ready.params.protocol_version` is an optional integer. **Current value: `1`.**"
- `conformance/sidecar/corpus/events/ready_full.json` — `"protocol_version": 1`
- `sdk/python/tests/test_sidecar_conformance.py:73` — `protocol_version=1,` **passed in by the test itself**
- `sdk/rust/librefang-sidecar/tests/conformance.rs` — `Some(1)`, same
- `sdk/python/librefang/sidecar/runtime.py:110` — `protocol_version: Optional[int] = None`
- `sdk/rust/librefang-sidecar/src/runtime.rs:139` — `fn protocol_version(&self) -> Option<u32> { None }`

No first-party adapter overrode it, so **every `ready` frame from every deployment carried `"protocol_version": null`** while four artefacts said `1`. The daemon's own comment recorded the intent and the gap in one line — `sidecar.rs:258`, `/// Reserved for skew diagnostics (logged, never enforced)` — and the reader loop logged the value at INFO with nothing to compare it against, because no `SIDECAR_PROTOCOL_VERSION` existed anywhere in the repo.

Both conformance suites were green throughout, because each supplied the corpus's own value by hand rather than building the frame the way an adapter builds it.

The second half is the SDK package version. `librefang.__version__` exists and is on no wire: `Schema.to_dict()` emitted `name` / `display_name` / `description` / `fields` and nothing else, and `SidecarSchema` had no field to receive it. Meanwhile `pythonpath_with_embedded` skipped the binary-embedded SDK whenever `has_real_sdk_installed(command)` was true, and that probe was `python -c "import librefang.sidecar"` — importability only. A March pip install imports perfectly well, wins the precedence contest against the August tree compiled into the daemon, and left a `debug!` as the only trace. That is the reported 2026.3.2201-vs-2026.7.31 pairing, reproduced from the code path.

## Where the command path sits in this

The daemon dispatches a slash command off `ChannelContent::Command` (`bridge.rs:4078`). The Python Telegram adapter first learned to emit `Content.command` in #5232 (2026-05-18); its predecessor emitted `{user_id, user_name, text, channel_id, platform}` with no `content` key at all, and skipped every update without `msg.get("text")`. The reported deployment's SDK predates that commit by two months.

`content` also deserializes as a unit: a shape `ChannelContent` cannot model fails `SidecarMessageParams` outright, and `sidecar.rs:1420` warns and drops the **whole** `message` event — command included. Nothing pinned that boundary, because `Content::Command` was the only frozen-core content shape with no corpus fixture on either side: `events/` had `message_text.json` and `message_minimal.json` and nothing else.

## Changes

**One source of truth, checked at the boundary**
- `librefang_channels::sidecar::SIDECAR_PROTOCOL_VERSION` is the source of truth for the number.
- `classify_protocol_version` maps a declared version to `Match` / `Unspecified` / `Older` / `Newer`; the reader loop `warn!`s on all three non-matching cases with adapter name and both versions. Still never enforced — a skewed sidecar keeps running, it just stops being invisible. Split out of the loop so the decision is unit-testable without a subprocess, which is why it went unexamined.
- Both SDKs default `protocol_version` to their mirror of the constant, so an adapter declares it without a per-adapter line. A version that must be opted into is a version that stays unset.

**The SDK version reaches an operator**
- `Schema.to_dict()` emits `sdk_version` unconditionally (Python); `Schema::new` fills it from `CARGO_PKG_VERSION` (Rust SDK).
- `SidecarSchema.sdk_version: Option<String>`, surfaced on `GET /api/channels` discovery rows and in the dashboard configure drawer. Absent rather than null when the adapter reported none — an SDK too old to carry the field is exactly the deployment this diagnoses.
- `describe_sidecar` resolves the same interpreter and PYTHONPATH as the eventual spawn, so this is the SDK that will actually serve traffic.

**The stale-install shadow is audible**
- The spawn probe is now `import librefang, librefang.sidecar; print(librefang.__version__)`. The installed copy still wins — editable installs stay authoritative — but a version differing from `embedded_sdk_version()` logs a `WARN` naming both and the remedy.
- `embedded_sdk_version()` reads `sdk/python/pyproject.toml` through `include_str!`. The extracted tree cannot report its own version: `_package_version()` resolves it from a sibling `pyproject.toml` (absent there) or installed distribution metadata (absent for a PYTHONPATH-only tree), and answers `0+unknown`.

**Drift guards**
- `crates/librefang-channels/tests/sidecar_version_contract.rs` pins `SIDECAR_PROTOCOL_VERSION` against the Python SDK constant, the Rust SDK constant, the corpus fixture, and the architecture doc, and pins `embedded_sdk_version()` against the packaged Python version. Each assertion names the file it read.
- `conformance/sidecar/corpus/events/message_command.json` closes the command-frame hole, asserted producer-side in both SDK suites and consumer-side in the supervisor suite.
- `ready_full.json` is now built through `SidecarAdapter.ready_event()` in both SDK suites instead of a hand-passed literal, so the declarative defaults every adapter inherits are what the corpus asserts. That is the assertion that would have caught this in March.
- `test_every_first_party_adapter_announces_the_protocol_version` walks all 27 first-party adapters.

## Verification

```
cargo test -p librefang-channels --test sidecar_version_contract --test sidecar_protocol_conformance
    6 passed (version contract) / 6 passed (conformance)
cargo test -p librefang-channels --lib sidecar:: embedded_sdk      75 passed / 16 passed
cargo test -p librefang-api --test sidecar_sdk_version_test        1 passed
cargo test -p librefang-api --test openapi_spec_test --test openapi_path_coverage_test --test sidecar_describe_test
                                                                   7 passed
cargo test -p librefang-api --lib routes::channels                 16 passed
cargo clippy -p librefang-channels -p librefang-api --lib --tests -- -D warnings   clean
cargo fmt --all -- --check                                         clean
(cd sdk/rust/librefang-sidecar && cargo test -p librefang-sidecar) 36 passed, incl. 14 conformance
(cd sdk/python && python -m pytest tests -q)                       2048 passed
(dashboard) pnpm vitest run ChannelsPage locale-parity en-locale-coverage   46 passed
(dashboard) tsc --noEmit / eslint                                  clean
```

The drift guard was verified to fail, not just to pass: flipping the Python `PROTOCOL_VERSION` to 2 produces
`python_sdk_mirrors_the_protocol_version ... FAILED — sdk/python/librefang/sidecar/protocol.py disagrees with SIDECAR_PROTOCOL_VERSION, left: 2, right: 1`.

`openapi.json` was regenerated by `openapi_spec_test` (which writes the file) and the result is byte-identical to the committed edit.

No live daemon or real LLM was run.

## Deliberately not done

- **The bare `ready` frame now warns.** `events/ready_minimal.json` stays accepted and the corpus still pins that acceptance, but an adapter declaring nothing is `Unspecified`, not "assumed current" — so the in-repo `examples/sidecar-channel-{bash,go,node,python}` adapters will each log one WARN at spawn. Those files are owned by the open #7239 and are not touched here to avoid a conflict; declaring the version in them is a one-line addition per example whenever that PR lands.
- **Root cause 2 is untouched** — it is #7159's, and it is unmerged, so this does not close the issue.


crates/librefang-channels/src/embedded_sdk.rs
```diff
@@ -16,9 +16,8 @@
 //!
 //! ## Precedence
 //!
-//! The embedded copy is a **fallback**, not a hijack. On every spawn
-//! we run a one-shot `<command> -c "import librefang.sidecar"` (cached
-//! per command path) and:
+//! The embedded copy is a **fallback**, not a hijack.
+//! On every spawn we run a one-shot `<command> -c "import librefang, librefang.sidecar; print(librefang.__version__)"` (cached per command path) and:
 //!
```
houko, line 21: Judgment call, left for you rather than fixed.

This `## Precedence` paragraph predates the PR and is hand-wrapped at ~72 columns, but the diff touches it — only to swap the `-c` probe string. I reflowed the two lines whose text actually changed, since the probe string now spans the break, but left the surrounding sentences (the `- If the interpreter already imports…` / `- Otherwise (the new-user case…)` bullets and the `Skipping the inject when a real install exists…` paragraph) at their original wrap.

CLAUDE.md says the rule applies "to any paragraph you are already touching" but also "don't rewrap an untouched paragraph just to enforce this." A one-word substitution inside a bullet sits between those two clauses. Reflowing the whole section is a handful of lines and would make the module header self-consistent; leaving it keeps this PR's diff narrower. Your call — happy either way, and no action needed if you'd rather leave it.

---

houko (commented):
Automated CLAUDE.md conventions pass on this PR.

**Fixed and pushed** (`d026104`, comment text only — no code lines touched): the new doc comments across `sidecar.rs`, `embedded_sdk.rs`, `sidecar_version_contract.rs`, `sidecar_protocol_conformance.rs`, `sidecar_sdk_version_test.rs`, `sidecar_describe.rs`, both SDKs' `protocol.rs` / `protocol.py`, `runtime.py`, and the Rust SDK `conformance.rs` were hand-wrapped at ~72–80 columns, splitting single sentences across lines. CLAUDE.md's prose rule ("no column limit; break only at sentence boundaries") covers source doc-comments and multi-line prose comment blocks. Worth noting that `sdk/rust/librefang-sidecar/src/runtime.rs:139` and the rewritten `docs/architecture/sidecar-protocol.md` already followed the rule — the rest of the diff was the outlier, so this only makes the PR internally consistent.

Verification after the reflow:
- `cargo test -p librefang-channels --test sidecar_version_contract --test sidecar_protocol_conformance` → 12 passed
- `cargo clippy -p librefang-channels --lib --tests -- -D warnings` → clean
- `cargo fmt --all -- --check` → clean, in both the main workspace and `sdk/rust/librefang-sidecar`
- `cargo test -p librefang-sidecar` → 36 passed
- Python changes are docstring/comment-only and were AST-checked (pytest isn't installed in this environment)
- Rebased onto your `900b1f3` before pushing; `cargo check --workspace --lib` isn't runnable here (`librefang-desktop` needs GTK/dbus system libs), so `librefang-api` was not compile-checked — CI covers it, and the two files touched there are comment-only.

**Clean, no findings:** the changelog fragment (`changelog.d/fixed/7140-sidecar-version-drift.md` — new file, recognized section, correct `(#7848) (@houko)` attribution, prose explaining *why*); `SidecarSchema.sdk_version` serde handling and every construction site including the fallback schema; no determinism regression (the `HashMap` in `embedded_sdk.rs` is a process-local probe cache, never iterated into a prompt or serialized artifact); no new routes needing a `middleware.rs` `is_public` entry; no AI attribution anywhere in the diff.

One judgment call inline.

---

Issue #7605, opened 2026-08-16
`librefang message` cannot target a session (all callers share one canonical session) and auto-memory leaks context across sessions


otaviobrum:
## Description

We run a **public, customer-facing agent** (a pre-sales assistant on our website). Every website
visitor is a different person, but they all talk to the same agent. Two independent mechanisms make
one visitor's text end up inside another visitor's turn.

**Bug 1 — `librefang message` has no way to select a session.**
The CLI (`librefang message <agent> <text> --json`) always resolves to the agent's *canonical*
session (one row per agent in `canonical_sessions`). For an agent serving N unrelated users, all N
conversations are interleaved in a single history that is re-sent on every turn. Consequences, in
order of severity: (a) **privacy** — visitor A's message is in the model context of visitor B's turn;
(b) **persistent prompt injection** — an instruction injected by one visitor travels into the next
visitor's turn; (c) **cost/latency** — the shared history grows without bound and is billed on every
turn (we hit `Token limit would be exceeded: 305649 + 1024 reserved > 300000` on an agent whose
individual turns are ~15k tokens).

The REST API **already solves this**: `POST /api/agents/{id}/message` accepts `session_id`
("Optional explicit session ID (UUID string) to use for this message... Enables multi-tab /
multi-session UIs where the caller tracks which session each conversation belongs to"). The CLI does
not expose it. `--incognito` is not a substitute: it suppresses session writes entirely, so
continuity *inside* one conversation is lost too.

**Bug 2 — auto-memory ignores session boundaries and ignores the agent's declared capabilities.**
With `[proactive_memory] auto_memorize = true` / `auto_retrieve = true` in `~/.librefang/config.toml`
(the default), every turn is stored as a `[Past exchange] Them:.../You:...` record in a **per-agent**
memory store and is auto-retrieved into *later* turns of the **same agent**, regardless of session.
So even after fixing Bug 1 with `session_id`, visitor B still received visitor A's data.

Worse, the agent-level knobs that look like they should govern this do not:

* `capabilities.memory_read = []`, `capabilities.memory_write = []` and `capabilities.tools = []` in
  `agent.toml` — the turn still came back with a populated `memories_used` field.
* `[proactive_memory] enabled = false` **inside `agent.toml`** — this section does not exist in
  `~/.librefang/schema.toml` for agents, so the key is silently ignored. There appears to be **no
  per-agent way to turn auto-memorize/auto-retrieve off**; the only switch is global, which would
[...]

Issue #5227, opened 2026-05-18
Cross-chat context bleed: group msg mentioning agent contaminates DM session reply


f-liva:
## Symptom

When the same user writes a message in a group chat that mentions the agent, and then writes a separate message in their 1:1 DM with the agent, the agent's DM reply leaks content from the group message — paraphrasing words or topics from the group msg that were never written in the DM session.

## Repro (live)

1. Configure agent on WhatsApp with `[channels.whatsapp] group_policy = "mention_only"` and a `group_trigger_patterns` regex that matches the agent name.
2. From the SAME user phone number, in a group chat where the agent is a member, write a message that contains the agent name (e.g. "ora torchio Ambrogio con messaggi strani"). The gateway forwards it because the mention pattern matched.
3. ~2 minutes later, from the same user, in the 1:1 DM with the agent, write a different question (e.g. "che ora è in Italia?").
4. Agent replies in the DM, but the reply paraphrases the GROUP message ("Modalità test attiva — pronto per messaggi tecnici strani 🎩") rather than answering the DM question.

## Observed sequence

```
10:10:40Z  group_Livas      Federico (group): "...torchio Ambrogio con messaggi strani..."
                            [gateway] forward_dispatch (mention matched)
10:12:31Z  DM_Federico      Federico (DM):    "rapid-1: mi serve un favore"
10:12:36Z  DM_Federico      Federico (DM):    "rapid-2: dimmi solo che ora è in Italia"
10:12:42Z  DM_Federico      Federico (DM):    "rapid-3: e basta, in una riga"
10:13:01Z  DM_Federico      Ambrogio (DM):    "Modalità test attiva — pronto per
                                              messaggi tecnici strani 🎩"
                                              ^^^^^^^^^^^^^^^^^^^^^^^
                                              direct paraphrase of group msg
                                              never typed in DM
```

## Expected

DM session must not have visibility on group messages. Group session and DM session are separate (different channel/peer) and the agent loop should resolve them via distinct `SessionId::for_channel(agent, channel_key)` instances. The DM reply should answer the DM question only.

## Hypothesis

`SessionId::for_channel` may be deterministic per `(agent, channel:peer)`, but downstream model context (claude-code CLI persistent state, `auto_memorize`, prompt-cache reuse) is shared across sessions for the same agent. When two sessions interleave within a short window, the second session's first turn picks up the first session's tail.

Worth checking:
[...]

PR #7849, opened 2026-08-24
fix(memory): scope automatic memory to its session and declared capabilities

houko:
Refs #7605 — Bug 2. #7815 shipped Bug 1 (`librefang message --session-id`) and deliberately did not `Closes`; this does not `Closes` either, see **Not covered** below.

## What was actually broken, verified against `9581b5e37`

The maintainer's comment on the issue splits Bug 2 into three asks. Re-checked each before touching anything:

- **Per-agent `auto_memorize` / `auto_retrieve` override — already shipped in #4870.** `ProactiveMemoryOverrides::resolve_auto_retrieve` / `resolve_auto_memorize` (`crates/librefang-types/src/memory.rs`) resolve the `agent.toml` `[proactive_memory]` block against the global config, and `gated_proactive_memory_for_retrieve` / `_for_memorize` apply it. Covered by `test_proactive_memory_overrides_per_field_disable` and friends. Not redone; a test was added at the gate itself because there was none there.
- **Capability half — missing.** `git grep memory_read crates/librefang-runtime crates/librefang-memory` returned nothing, so an agent declaring `capabilities.memory_read = []` still had memories injected and still auto-memorized every turn.
- **Session scoping — missing.** The only isolation in `auto_retrieve` was the #5227 `chat_scope` filter, and `chat_scope` is `None` for every REST / CLI / dashboard caller, so the filter was a no-op exactly where a multi-user deployment lives.

### Session scoping (`librefang-memory`, `librefang-types`)

- `SESSION_SCOPE_METADATA_KEY` + `memory_session_scope_allows_recall` in `librefang-types`. An untagged memory passes (legacy rows, dashboard writes); a memory tagged for another session is dropped.
- **No `MemoryLevel::User` exemption**, unlike the chat predicate. The chats #5227 separates belong to one person; the sessions this separates routinely belong to different people, and "user-level" is exactly where an extractor files `PINE-77`. An exemption would leave the reported leak open on the highest-value rows.
- `ProactiveMemoryHooks::auto_memorize` / `auto_retrieve` take a `session_scope: Option<&str>`. `auto_memorize` stamps it; `auto_retrieve` filters on it, composed with (not replacing) the chat filter.
- The dedup candidate set in `add_with_decision` is filtered the same way. Without it, session B's turn could NOOP against a stranger's row — losing the fact — or UPDATE it in place, overwriting one visitor's memory with another's content.
- The substrate / context-engine recall in `setup_recalled_memories` gets the same filter. It is a second path into the prompt; gating only `auto_retrieve` would have left it open.
- The scope is the session the loop already reads and writes (`Session::id`), resolved by the ladder in `docs/architecture/session-mode-resolution.md`. No second notion of a session is derived.

### Capabilities (`librefang-types`, `librefang-runtime`, `librefang-kernel`)

- `ManifestCapabilities::memory_read` / `memory_write` become `Option<Vec<String>>`. Absent key → `None` → unrestricted, as every other capability list reads when missing; `memory_read = []` → `Some([])` → denies. Without the tri-state, `[]` and "no `[capabilities]` block" are the same value and the operator's explicit lockout cannot be expressed.
- `skip_serializing_if = "Option::is_none"` so an undeclared list survives a msgpack / JSON round-trip as undeclared rather than coming back declared-empty.
- New `ManifestCapabilities::allows_own_memory_read` / `allows_own_memory_write`, applied in `gated_proactive_memory_for_retrieve` / `_for_memorize`.
- `scope_covers_own_memory` moves from `LibreFangKernel` to `librefang_types::capability` (kernel keeps a delegating assoc fn) so the #7808 tool gate and the new auto gate cannot drift.
- `manifest_to_capabilities` now lets a declared list beat the profile's implied one even when empty.

### Config surface

- `[proactive_memory] session_scoped_recall` in `config.toml`, default **`true`**, plus `session_scoped_recall: Option<bool>` on `ProactiveMemoryOverrides` for `agent.toml` (per-agent overrides never live in `config.toml`, #5476). Struct field, `#[serde(default)]`, `Default` entry, `Serialize`/`Deserialize` — all four.

## Default chosen, and the one behaviour it changes

Default `true`, because this is the privacy leak the issue reports and a default that leaves it open fixes nothing for the reporter.

For an agent that configures nothing, almost nothing changes:

- a bare `librefang message` or a REST turn with no `session_id` runs in one canonical session, so the stamp is constant;
- a channel agent gets one session per chat, which #5227 was already separating;
- forks are forced `Persistent` and share the parent's session;
- rows written before this shipped carry no stamp and stay recallable everywhere, so no existing store goes dark.

The case that does change is **`session_mode = "new"`**: each invocation gets a fresh session, so a cron agent no longer auto-recalls what a previous fire memorized. That is the honest reading of "new", but an agent relying on the agent-wide pool for continuity wants `session_scoped_recall = false` in its `agent.toml`.

## Upgrade note

A manifest blob written into SQLite before this change stored an undeclared scope list as `[]`, which now decodes as declared-empty. Boot re-reads each agent's `agent.toml` and lets the disk copy win when it differs (`kernel/boot.rs`), so every agent with a manifest on disk — which `persist_manifest_to_disk` gives every registered agent — is corrected on the first restart. An agent whose `agent.toml` has since been deleted reads as declared-empty and loses automatic memory until the file is restored: loss of a convenience, logged at `debug` naming the capability, rather than a store quietly re-opening.

## Verification

Run with a scratch `CARGO_TARGET_DIR`.

- `cargo check --workspace --tests` — clean.
- `cargo clippy -p librefang-types -p librefang-memory -p librefang-runtime -p librefang-kernel --lib --tests -- -D warnings` — clean.
- `cargo test -p librefang-types --lib` — 956 passed.
- `cargo test -p librefang-memory --lib` — 374 passed.
- `cargo test -p librefang-runtime --lib agent_loop` — 283 passed.

New tests:

- `librefang-memory` `proactive::tests::auto_memorize_then_retrieve_does_not_cross_sessions_7605` — memorize under session A, retrieve under session B, assert empty; retrieve under A, assert non-empty so the filter does not over-prune.
- `librefang-memory` `proactive::tests::untagged_memories_and_unscoped_callers_keep_pre_7605_recall` — legacy untagged rows surface from every session, and a caller passing no scope gets the whole agent pool back.
- `librefang-memory` `proactive::tests::dedup_candidates_do_not_cross_sessions_7605` — an identical fact from a different session ADDs its own row instead of folding into the other session's.
- `librefang-runtime` `agent_loop::prompt::tests::recall_setup_drops_other_sessions_memories_7605` — the substrate recall path is filtered too, and untagged rows still surface.
- `librefang-runtime` `agent_loop::end_turn::tests::declared_empty_memory_scopes_close_both_automatic_memory_gates` — `memory_read = []` blocks retrieval, `memory_write = []` blocks capture.
- `librefang-runtime` `agent_loop::end_turn::tests::memory_read_and_write_scopes_gate_their_own_half_only`, `undeclared_memory_scopes_leave_both_gates_open`, `per_agent_proactive_memory_override_still_gates_each_half`.
- `librefang-runtime` `agent_loop::end_turn::tests::session_recall_scope_defaults_to_the_turns_own_session`, `session_recall_scope_can_be_turned_off_per_agent_and_globally`, `session_recall_scope_is_none_without_a_proactive_store`.
- `librefang-types` `memory::tests::session_scope_filter_hides_other_sessions_and_keeps_untagged_rows`, `session_scope_filter_has_no_user_level_exemption`, `session_scoped_recall_defaults_on_and_is_overridable_per_agent`, `session_scoped_recall_ignores_the_master_switch`, `session_scoped_recall_survives_a_config_roundtrip`.
- `librefang-types` `agent::tests::declared_empty_memory_scopes_deny_while_an_absent_key_stays_open`, `declared_memory_scopes_are_matched_against_the_agents_own_store`, `undeclared_memory_scopes_survive_a_serde_roundtrip_as_undeclared`.

No new HTTP route or dashboard change, so no `TestServer` test or vitest is owed.
Not run: live daemon against a real LLM — human-only.

## Not covered — why this is `Refs`, not `Closes`

Two items, both deliberately outside this diff and both needing a maintainer call:

1. **`GET` / `PATCH /api/memory/config` does not expose `session_scoped_recall`.** That route projects a curated subset of `[proactive_memory]` (`enabled`, `auto_memorize`, `auto_retrieve`, `extraction_model`, `max_retrieve`) and the new key is the same class of switch, so an operator flipping memory settings from the dashboard will not see it. Adding it is ~4 lines plus a `TestServer` test, but it is `librefang-api` — a different crate from the ones this PR is in — so I did not decide it unilaterally. Say the word and it goes in here.
2. **The dashboard's `agent.toml` writer drops an explicitly empty scope list.** `dashboard/src/lib/agentManifest.ts:526` emits `memory_read` only `if (form.capabilities.memory_read.length)`, so editing an agent through the form and saving turns `memory_read = []` back into an absent key — silently re-opening automatic memory. Pre-existing and uniform across every capability field, but this PR is what makes it load-bearing. Fixing it properly means teaching the form model the same tri-state (`string[] | null`) plus a colocated vitest, in the dashboard domain.


crates/librefang-kernel/src/kernel/tools_and_skills.rs
```diff
@@ -1549,21 +1549,12 @@ impl LibreFangKernel {
...
-    ///   * `*` — the unrestricted grant.
-    ///   * `self.*` — what every non-memory [`ToolProfile`] implies for
-    ///     `memory_write`. Those profiles mean "this agent may write its own
-    ///     memory", not "this agent may not use the semantic store", so
-    ///     matching it literally against `proactive` would strip the write
-    ///     tools from every profile-based agent.
-    ///   * anything glob-matching `proactive`, the namespace string the
-    ///     per-user ACL and the REST layer already use for this store — so an
-    ///     operator can name it explicitly.
+    /// Thin re-export of [`librefang_types::capability::scope_covers_own_memory`], which moved to the types crate in #7605 so the automatic memorize / retrieve gate in `librefang-runtime` answers this question identically to the tool gate here.
```
houko, line 1694: Worth a maintainer sign-off: `memory_read = []` / `memory_write = []` now mean two different things depending on path. `ManifestCapabilities::allows_own_memory_read`/`_write` (the new automatic-memory gate this PR adds, see `agent.rs` line ~1794) reads the raw manifest field and denies on a declared-empty list. This tool gate (`scope_covers_own_memory`, called from the kernel-resolved `Capability::MemoryRead`/`Write` list a few lines above) still treats a declared-empty list the same as an absent key — kept open for #7808 backward compat, per the new doc comment here.

Net effect: an operator who writes `memory_read = []` to lock an agent out of memory stops automatic recall/capture but does **not** stop the LLM from directly calling `memory_semantic_search`/`_add` if `capabilities.tools` still grants it. That's already true today for #7808 and isn't a regression from this PR, but this PR is what makes the field's *meaning* diverge between the two gates, and the changelog fragment ("`memory_read = []` … now actually stop automatic recall") doesn't call out that manual tool access is unaffected. Intentional, or worth unifying/documenting more loudly?

---

houko (commented):
Applied two mechanical fixes directly to this branch (pushed as `04929b0` and `8bfc211`):
- Repaired a doc-comment splice in `crates/librefang-memory/src/proactive.rs` where the new #7605 test doc comments had been inserted mid-sentence into the pre-existing comment for `test_auto_memorize_stamps_chat_scope_5227`, leaving a garbled sentence and an orphaned trailing line ~180 lines later.
- Rewrapped the new #7605 doc/line comments across 10 files to one-sentence-per-line, per CLAUDE.md's prose-wrapping rule (they were hard-wrapped at a fixed column, splitting some sentences and merging others).

One design point flagged inline (`crates/librefang-kernel/src/kernel/tools_and_skills.rs`): the new automatic-memory gate and the pre-existing #7808 tool gate now treat a declared-empty `memory_read`/`memory_write` list differently.

One more, not inline since the file isn't touched by this diff: the dashboard's `agent.toml` writer (`crates/librefang-api/dashboard/src/lib/agentManifest.ts:526-527`) only emits `memory_read`/`memory_write` when the array is non-empty (`if (form.capabilities.memory_read.length)`), so there's no way to produce the declared-empty `memory_read = []` this PR makes load-bearing for denying automatic memory through the dashboard UI — clearing the list in the form just makes the key disappear (→ absent → unrestricted), the opposite of what an operator would expect. Pre-existing and out of scope per the PR description's own "Not covered" section, just flagging since it's now load-bearing.

---

PR #7855, opened 2026-08-24
feat(audit): repair a broken chain without discarding history

### LINKED ISSUE #7702: bug: audit hash-chain integrity fails on boot — chain break at seq 204 (proteo instance)

houko:
Closes #7702.

## Credit

The analysis behind this, and the shape of the recovery — diagnose the first break, classify it, sever from the break forward rather than wiping the table, record a `ChainReanchored` marker chained onto the last survivor, drive it from a `librefang security audit-reanchor` command that refuses to run against a live daemon — is @DaBlitzStein's, from #7747.
That PR also ruled out two candidate root causes with evidence rather than leaving them as open possibilities, which is what narrowed the search to the stale-snapshot append.
This PR exists because the review on #7747 stalled: every commit there since the CHANGES_REQUESTED review has been a bare `Merge branch 'main'`, and the underlying issue is a production log that cannot verify.
Rather than keep waiting, I implemented the requested changes directly.
I wrote the code here rather than copying that diff, so the three review items are addressed structurally rather than patched, but the design is theirs and #7747 should be closed as superseded — that is a recommendation for the author, not something I am acting on.

## Premise check

`diagnose_chain`, `reanchor_after_break` and `audit-reanchor` do not exist anywhere on `main`.
This is an implementation of #7702's recovery request, not an adjustment to shipped code.

## Complementary to #7847

#7847 (merged) made `record_with_context` derive `seq` and `prev_hash` from the durable tail inside the same `BEGIN IMMEDIATE` transaction as the INSERT, so a stale-snapshot append can no longer fork the chain.
That protects logs going forward and does nothing for a log that is already broken — and the affected host's log is already broken.
Prevention landed; this is repair.
Between them, #7702's two requests — recovery tooling, and an answer to what caused it — are both satisfied, so this closes the issue.

## The problem this solves

Before this, the only recovery LibreFang shipped was `librefang security audit-reset`, whose single mode is `DELETE FROM audit_entries`.
It restores verification by destroying the evidence.
An operator with a forked chain was choosing between a log that never verifies again and no log at all, which is not a choice a compliance deployment can make.

A Merkle chain admits exactly one predecessor per row, so a repair necessarily severs one side of the break.
The only real question is whether the severed rows are destroyed or preserved, and this preserves them.

## Changes

**`crates/librefang-runtime-audit/src/recovery/`** — new module.

- `diagnose_chain(&Connection)` walks `audit_entries` in `seq` order and returns the first `ChainBreak { seq, kind, expected, found }`. `ChainBreakKind` separates a `PrevHashMismatch` (the link to the previous row is wrong — the fork shape) from a `HashMismatch` (the row's own stored hash does not match its content — an edit to the row). The walk seeds from the first row's own `prev_hash`, exactly as `with_db` seeds `chain_anchor` on boot, so a database whose prefix retention has pruned is not reported as corrupt. It accepts the pre-delimiter (v1) hash layout on the same terms `verify_integrity` does, so a log old enough to predate the delimiter fix is not reported as broken either.
- `plan_reanchor(&Connection)` returns what the repair would touch — rows before, severed, surviving — without mutating, so the CLI dry run prints numbers from the same query the repair uses.
- `reanchor_after_break(&mut Connection, archive_dir, anchor_path, &ChainBreak)` performs the repair. The rows at and after the break are written to `<data_dir>/audit-archive/audit-severed-<seq>-<timestamp>.jsonl` and fsynced **before** anything is deleted; then the delete, then a `ChainReanchored` marker chained onto the last survivor whose `detail` is a JSON document naming the break, the severed count, the archive file name and the archive's SHA-256. Committing that digest into the chain is what keeps the preserved rows tamper-evident after the repair.
- New `AuditAction::ChainReanchored` variant (append-only, so existing persisted hashes are unaffected).

**`crates/librefang-cli`** — `librefang security audit-reanchor`, dry-run by default, `--confirm` to perform.
The daemon guard applies to the dry run too, because a diagnosis taken against a live writer describes a state that may already have changed by the time the operator acts on it.

**Docs** — `docs/operations/audit-chain-recovery.md` (how to tell an anchor mismatch from a chain break, what produced it, how to run the repair, how to read and verify an archive, when `audit-reset` is still the right tool) and a SECURITY.md bullet under Audit.

## The three standing review items on #7747

1. **Reuse the canonical anchor writer.** Addressed by placing the repair in `librefang-runtime-audit` rather than the CLI: `AuditLog::write_anchor` is called directly, so the `"<seq> <hash>\n"` line format and the temp-file-plus-rename are shared by construction and cannot drift. There is nothing to keep in sync. `reanchor_writes_an_anchor_in_the_format_the_reader_accepts` asserts the exact bytes and round-trips them back through `read_anchor`.
2. **End-to-end test that reopens `AuditLog::with_db_anchored` after recovery.** `reanchor_leaves_a_log_that_reopens_and_verifies` builds the real fork shape (two chains merged into one table), asserts the broken table fails through `with_db_anchored`, repairs it, reopens through the same path, and asserts `verify_integrity() == Ok(())` plus that the surviving prefix kept its original hashes. Three more tests reopen after recovery in other configurations.
3. **`COUNT(*)` inside the recovery transaction.** The count is taken inside the transaction, after both the delete and the marker INSERT, and that value — not `marker_seq + 1` — is what reaches the anchor. `reanchor_anchors_the_row_count_not_the_marker_seq` prunes a prefix first so the two diverge (marker at seq 5, four rows persisted) and asserts the anchor carries 4.

The third blocking item in my review — that `reanchor_after_break` trusted a stale `ChainBreak` and that `find_daemon_in_home` does not exclude every other SQLite writer — is closed by re-diagnosing inside the repair's own `BEGIN IMMEDIATE` transaction and aborting unless the break matches field for field.
`BEGIN IMMEDIATE` takes the same RESERVED lock the append path takes, so the guard is a real writer exclusion rather than a daemon-file probe.
`reanchor_refuses_a_break_that_moved_and_mutates_nothing` covers it, and asserts no archive is written on the refusal path either.

## Verification

- `cargo test -p librefang-runtime-audit` — 55 passed, of which 12 are new: `diagnose_chain_returns_none_for_an_intact_chain`, `diagnose_chain_reports_the_fork_as_a_prev_hash_mismatch`, `diagnose_chain_distinguishes_an_edited_row_from_a_fork`, `diagnose_chain_accepts_a_pruned_prefix`, `reanchor_leaves_a_log_that_reopens_and_verifies`, `reanchor_writes_an_anchor_in_the_format_the_reader_accepts`, `reanchor_anchors_the_row_count_not_the_marker_seq`, `reanchor_archives_every_severed_row_and_commits_the_digest`, `reanchor_refuses_a_break_that_moved_and_mutates_nothing`, `reanchor_refuses_when_the_chain_already_verifies`, `plan_reanchor_counts_what_the_repair_would_touch`, `a_first_row_hash_mismatch_severs_everything_into_the_archive`.
- `cargo test -p librefang-cli --test i18n_checks` — 3 passed; the new strings are present in all four CLI locales (en, ko, uk, zh-CN).
- `cargo clippy -p librefang-runtime-audit --lib --tests -- -D warnings` and `cargo clippy -p librefang-cli --bins -- -D warnings` — clean.
- `cargo check -p librefang-cli --bins` — clean.
- `scripts/check-changelog-attribution.py` — clean.

No API route, no dashboard surface: the command is offline by design and refuses to run while the daemon holds the database, so there is nothing for `TestServer` to exercise.

## Not done here

- No live-daemon run. The command is human-only to execute against a real `~/.librefang/librefang.db`; the automated coverage goes through `with_db_anchored` against a real pool, which is the same boot path.
- The archive is a file next to the database rather than a table, so it can be lost independently of the database. Committing its SHA-256 into the chain means a lost or altered archive is detectable, not recoverable. Moving it into a `audit_severed_entries` table would need a `librefang-memory` migration — a different crate — so it is a deliberate design choice rather than a deferral, and the file form has the advantage that the forensic record can be moved off the host.


crates/librefang-runtime-audit/src/recovery/mod.rs
```diff
@@ -0,0 +1,461 @@
...
+        .map_err(|e| format!("cannot create the archive {}: {e}", tmp.display()))?;
+    file.write_all(&body)
+        .and_then(|()| file.sync_all())
+        .map_err(|e| format!("cannot write the archive {}: {e}", tmp.display()))?;
+    drop(file);
+    std::fs::rename(&tmp, &path)
+        .map_err(|e| format!("cannot publish the archive {}: {e}", path.display()))?;
+    // Durability of the rename itself; without it a crash can leave the directory entry unrecorded even though the file's contents were synced.
+    if let Ok(dir) = std::fs::File::open(archive_dir) {
+        let _ = dir.sync_all();
```
houko, line 428: The directory fsync's failure is silently swallowed here (`if let Ok(dir) = ... { let _ = dir.sync_all(); }`), with no error return and no log line. The doc-comment above `write_archive` frames this fsync as part of the crash-durability guarantee ("both the file and its directory are fsynced before the rename returns, so a crash cannot leave a half-written archive"), but if `File::open` or `sync_all()` fails here, that guarantee silently doesn't hold and the operator gets no signal.

Also, as written, the directory fsync happens *after* `rename` (line 424), not before it — which is actually the standard correct technique for making a rename durable, but it doesn't match the doc-comment's "fsynced before the rename returns" wording. Worth tightening the comment either way.

Is swallowing this specific failure intentional (the content fsync + rename already gives the guarantee that matters, and this is best-effort), or should it propagate an error / log a warning so a failed directory fsync isn't silently lost?

---

houko (commented):
Reviewed against CLAUDE.md conventions. Pushed a small mechanical fix for hard-wrapped doc-comments (project convention is one sentence per line, no column-limit wrapping); left one inline question about a silently-swallowed fsync error. Everything else checked (deterministic JSON ordering on the hash-relevant path, changelog fragment format/attribution, config-field completeness) came back clean.

---

PR #7858, opened 2026-08-24
feat(workflows): step agent types with find-or-spawn

### LINKED ISSUE #7712: Workflow steps: reference an agent type with find-or-spawn

houko:
Closes #7712.

## Credit

The analysis and the resolution design here are **@DaBlitzStein's**, from #7715 — the find-or-spawn shape, the canonical name-derived UUID via `agent_identities` (#4614), the ByType-specific step error, and the observation that dry runs must not spawn are all theirs.
This PR exists only because the review on #7715 stalled: every commit since the CHANGES_REQUESTED review has been a bare `Merge branch 'main'`, and four PRs (#7717, #7725, #7726, #7727) are stacked on top of it, so landing the root unblocks the most work of anything in the queue.
I recommend #7715 be closed as superseded by this one — that is a recommendation for @DaBlitzStein and the maintainers to act on, not something I have done or will do.
Nothing in #7715 has been closed, labelled, commented on, or pushed to.

## Premise check

`StepAgent::ByType` and the find-or-spawn resolution are **absent from `main`** — `StepAgent` there has only `ById` and `ByName`, and there is no template-loading path for step agents at all.
So this is an implementation of the issue, written fresh against current `main`, not an adjustment of existing code.
Where it reuses #7715's approach it is called out above; the diff was written rather than copied, and the three review items are built into the design instead of patched on.

## The three review items

**1. Exactly-one-of validation on the routing keys.**
`id` / `name` / `type` are now counted before a `StepAgent` is constructed, in both the kernel deserializer (TOML/JSON workflow definitions) and the HTTP `POST` / `PUT /api/workflows` handlers.
Presence is decided on the key, not on "the key holds a string": `{"agent_id": 7, "agent_name": "b"}` used to fail the `as_str()` cast, look unset, and resolve to `ByName` — binding the step to the agent the author had stopped naming.
An explicit JSON `null` still counts as absent so a client that serializes unset fields round-trips.
Integration tests cover every ambiguous pair and the all-three case on **both** create and update (8 cases), plus a non-string key, an empty key, and — for update — an assertion that the rejected `PUT` left the stored step list untouched rather than partially replaced.

**2. The `workflow_create` `builtin_tool_definitions` hunk is not here.**
No change to `crates/librefang-runtime/src/tool_runner/definitions.rs`.
It belongs with #6943 and must land atomically with a dispatcher, or agents discover an advertised tool that answers `ToolError::NotFound`.

**3. Structured template-loading error.**
New `librefang_kernel::agent_template` returns `TemplateLoadError` with variants `Missing` / `Unreadable` / `Malformed` / `NameMismatch` / `InvalidName`, each naming the path it was looking at, plus a stable `kind()` for structured logs and an `is_missing()` for the one case where "try a different name" is the right advice.
The search stops at the first directory that *has* the manifest file, so a malformed `workspaces/agents/<type>/agent.toml` is reported as malformed instead of falling through to the registry copy — silently serving a different manifest than the one on the operator's disk is how an edit appears to have no effect.
Template identity is pinned: the raw TOML is inspected for a `name` key before deserializing (because `AgentManifest` is `#[serde(default)]` and would turn an omitted name into the literal `"unnamed"`), an omitted name is pinned to the directory name, and a *declared* name that disagrees is refused.
That refusal is what makes the duplicate-race reuse safe — since `manifest.name == requested` always holds, the entry re-read after an `AgentAlreadyExists` is the same agent this spawn was building.
Reuse now matches **only** `AgentAlreadyExists`; every other spawn failure returns an error rather than re-probing the registry and binding the step to whatever holds that name.

## Other substantive changes

- `StepAgent::ByType { template }` serializes as `{"type": "..."}` and as `{"agent_type": "..."}` in the HTTP step shape, so a workflow round-trips through `GET /api/workflows/{id}` as a type rather than decaying to a name.
- The six byte-identical `match agent_ref { ById .. ByName .. }` closures — kernel `run_workflow`, kernel operator-timeout auto-resolve, `KernelHandle` workflow runner, channel bridge, and the API background / resume / operator drivers — collapse into one `KernelApi::resolve_step_agent`. Six copies of one policy is six places to forget a variant, and `ByType` is exactly the kind of variant a missed copy fails *silently* on: the step resolves to `None` and the run reports a missing agent rather than an unhandled reference.
- `KernelApi::preview_step_agent` is the non-mutating twin, used by `dry_run_workflow`. `POST /api/workflows/{id}/dry-run` is documented as side-effect free; a preview that mints an agent per invocation turns "show me what this would do" into "do part of it". The preview reports the canonical id the real run *would* land on, read from `agent_identities` without registering.
- The type agent is spawned top-level (`parent = None`) deliberately: it outlives the run that first needed it and is shared by every later run of every workflow naming that type, so parenting it to whichever run happened to be first would give it that agent's lifetime and capability ceiling.
- Template search order is `<home>/workspaces/agents/<type>/agent.toml` then `<home>/registry/agents/<type>/agent.toml` — the two agent-template directories that actually exist on `main`; a local template shadows the registry checkout, matching the precedence `librefang-kernel-router` already uses for hands. Type names carrying a path separator, `..`, or a NUL are rejected before touching the filesystem.
- **Dashboard (required, not optional):** `CanvasPage`'s `buildSteps` sent `agent_id` *and* `agent_name` on every step, because a canvas node carries both so the card can render the name. Under the new rule that payload is a 400, so saving from the canvas would have broken. Routing-key selection moved into `stepAgentPayload()` in `lib/canvas.ts` (pure, unit-tested) and emits exactly one key.
- Dashboard, same file: `LoadedWorkflowStep.agent` was typed `{ id?, name? }` while the API has always emitted `{ agent_id }` / `{ agent_name }`, so the no-saved-layout load path silently restored no agent binding at all. Fixed to the real shape, including `agent_type`. `agentType` is carried through node data, the import validator, the node card, and the template-load path so a type binding survives a canvas round trip; picking a concrete agent in the node editor clears it rather than leaving both for `buildSteps` to choose between. No new UI and no new strings, so no locale changes.
- Docs updated in `docs/src/app/agent/workflows/page.mdx` and the zh mirror.

## Deliberately not changed, with the trade-off

`WorkflowTemplateStep::agent` in `librefang-types` is a single `Option<String>` that cannot record *how* a step addresses its agent.
Saving a workflow as a template therefore records the type name and `instantiate` reads it back as `ByName` — the same flattening `ById` has always had there.
The name still resolves once the type has been spawned; what is lost is find-or-spawn for a workflow instantiated from a saved template into a daemon that has never run that type.
Fixing it means widening the persisted template format in a different crate, plus the dashboard template shape and the bundled `workflows/templates/*.json`, so I have flagged it rather than bundled it. Say the word and I will open it as its own PR.

## Verification

- `cargo test -p librefang-kernel --lib step_agent` — 20/20. Kernel: spawn, reuse, missing, corrupt-not-missing, name mismatch, rejected-manifest-is-a-spawn-failure (asserting no registry entry is left to fall back onto), preview-does-not-spawn, preview-missing. Deserializer: by-type, type round trip through TOML, `{"type": ...}` serialization, every ambiguous pair + all three, non-string key, explicit null.
- `cargo test -p librefang-kernel --lib agent_template` — 10/10, covering precedence, fall-through, missing, two malformed shapes, malformed-does-not-fall-through-to-registry, name mismatch, name pinning, and path traversal.
- `cargo test -p librefang-api --test workflows_routes_integration` — 85/85, including the 8 ambiguity cases across create and update, the `agent_type` create and update round trips, and the two dry-run cases (`workflow_dry_run_previews_agent_type_without_spawning` asserts the registry count is unchanged).
- `cargo test -p librefang-kernel --test workflow_integration_test` — 8/8; `-p librefang-api` `workflow_lifecycle_test` 17/17, `workflow_operator_action_test` 12/12, `workflow_operator_nodes_test` 20/20, `workflow_pause_resume_test` 14/14, `dead_route_audit_test` 2/2, `openapi_path_coverage_test` 1/1.
- `cargo clippy -p librefang-kernel -p librefang-api --lib --tests -- -D warnings` — clean. `cargo fmt --all -- --check` — clean.
- Dashboard: `pnpm run build` (not just tsc), `pnpm run test` 1208/1208 across 124 files, `pnpm run lint` — all clean.
- Not run locally: workspace-wide build and the live-daemon paths, which need a real LLM and are human-only.


crates/librefang-kernel/src/agent_template.rs
```diff
@@ -0,0 +1,420 @@
...
+        Some("must not be empty")
+    } else if requested.contains('/') || requested.contains('\\') {
+        Some("must not contain a path separator")
+    } else if requested == "." || requested == ".." {
+        Some("must not be a relative path component")
+    } else if requested.contains('\0') {
+        Some("must not contain a NUL byte")
+    } else {
+        None
+    };
```
houko, line 160: `validate_type_name` rejects `/`, `\`, `.`/`..`, and NUL, but not a bare `:`.

On Windows, `PathBuf::join`/`push` special-cases a path with a drive prefix but no root (e.g. `c:temp`): it *replaces* `self` rather than joining onto it (documented behavior of `Path::push`). A type name like `"c:temp"` contains none of the rejected characters, so it passes this check, but `dir.join("c:temp")` in the caller would then discard `dir` entirely and resolve relative to that drive's current working directory instead of under `workspaces/agents/`/`registry/agents/`.

That's the exact class of thing this function's doc comment says it prevents ("a name carrying a path separator or `..` would otherwise let a workflow definition read an `agent.toml` from anywhere on the host") — the claim holds for `/`/`\`/`..` but not for the Windows drive-prefix escape. Worth either rejecting `:` here too, or confirming the existing `/api/templates/{name}` precedent this mirrors already handles it (in which case this comment can be resolved).

---

crates/librefang-kernel/src/kernel/step_agent.rs
```diff
@@ -0,0 +1,222 @@
...
+            }
+            // A concurrent resolver for the same type won between our
+            // `find_by_name` miss and this register. Reusing the winner is
+            // correct — and is the ONLY spawn failure where reuse is correct.
+            // Any other error (rejected module path, reserved name,
+            // tool_exec override) means this manifest was never registered,
+            // so an entry that happens to hold the name belongs to something
+            // else and binding the step to it is the silent mis-binding this
+            // resolution path exists to avoid.
+            Err(KernelError::LibreFang(LibreFangError::AgentAlreadyExists(_))) => self
```
houko, line 211: This `AgentAlreadyExists` race-reuse arm is the one path this PR calls out as load-bearing ("the ONLY spawn failure where reuse is correct"), but the added tests (e.g. `step_agent_by_type_reuses_the_registered_agent`) all seed the collision *before* calling resolve — none drives two concurrent `find_or_spawn_agent_type` calls (e.g. via `tokio::join!`) through the actual `find_by_name` miss → register → lose-the-race sequence this arm exists for.

Given how carefully the surrounding comment reasons about why reuse is safe *only* here, it'd be worth a genuine-concurrency test to prove the race path itself resolves correctly (same canonical id on both sides), not just that a pre-existing entry is reused when hit non-concurrently.

---

houko (commented):
Automated CLAUDE.md-rules pass. No mechanical fixes needed — checked deterministic-ordering, config-field, serde-skip, auth-allowlist, route-registration, dashboard query/mutation-hook, prose-wrapping, changelog-fragment, and AI-attribution rules, all clean. Two ambiguous items left as line comments for a maintainer call: a possible Windows drive-prefix path-validation gap, and a race-arm that's asserted but not exercised under real concurrency.

---

PR #7859, opened 2026-08-24
feat(api): agent-type CRUD on /api/templates, with a non-destructive PUT

### LINKED ISSUE #7731: Agent types editor: catalog-backed skill/tool pickers, workspace-agent guards, TUI parity
## Motivation

The agent-types editor (dashboard) is a text-based form: tools are a comma-separated string and skills cannot be edited at all (flagged in #6931 review). Workspace agents listed as types expose Edit/Delete that would write a shadowing template. The TUI templates screen only lists builtins — operator-created types are invisible and cannot be spawned from.

## Proposal

- Dashboard editor: catalog-backed MultiSelectCmdk pickers for skills (GET /api/skills) and tools (GET /api/tools), mirroring the agent editor; provider/model stay free-text with hints (spawn defaults).
- Workspace-agent source entries show "managed via Agents" instead of Edit/Delete.
- TUI templates screen lists operator-created types from templates/ and spawns them from their real TOML (tools/skills/model preserved).
- API already round-trips the flat shape (skills included) — no route change.

## Acceptance

tsc + locale parity green; TUI compiles; doc update in docs/architecture/agent-created-workflows.md.

### LINKED ISSUE #7740: bug: editing an agent type from the dashboard silently destroys every field the flat JSON does not carry
## What happens

`PUT /api/templates/{name}` rebuilds the whole TOML from a flat 7-key JSON body and overwrites the file with `std::fs::write`.
Anything an operator put in the agent type's `.toml` by hand — `[compaction]`, `max_history_messages`, `[[triggers]]`, `[resources]`, `[autonomous]`, `mcp_servers`, `tool_allowlist`, … — is gone after the first save from the WebUI.
The endpoint returns `200`.

Evidence:

- `crates/librefang-api/src/routes/agent_templates.rs:517` rebuilds from the request JSON, `:524` writes the file.
- `crates/librefang-types/src/agent.rs:991` `agent_type_json_to_toml` reads exactly 7 keys (`name`, `description`, `system_prompt`, `provider`, `model`, `tools[]`, `skills[]`, plus `channels[]` and `routing{}`); everything else comes from `..Default::default()` at `:1055`, `:1059`, `:1061`.

That drops 49 manifest fields plus 9 `ModelConfig` fields on every save.

## Second, related defect: `channels` and `routing` do not survive the round-trip

`manifest_to_agent_type` (`crates/librefang-api/src/routes/agent_templates.rs:431-441`) emits only `name`, `description`, `system_prompt`, `provider`, `model`, `tools`, `skills` — it never emits `channels` or `routing`.
The dashboard does expect them: `crates/librefang-api/dashboard/src/api.ts:3724-3733` declares them on `interface AgentType` and `crates/librefang-api/dashboard/src/pages/AgentTypesPage.tsx:79-84` reads `type.channels ?? []` / `type.routing?.…`.

So the GET returns `undefined` for both, the form renders them empty, and saving posts `channels: []` / `routing: undefined` — which the converter then omits.
**Net effect: a channel allowlist or a model-tier routing block configured in the editor is erased the next time anyone opens and saves that agent type.**

The GET already exposes the full manifest under the `manifest` key (`agent_templates.rs:325-338`); the page just does not use it.

## Expected

Saving an agent type should preserve every field it does not explicitly edit.

## Suggested fix

- Make `manifest_to_agent_type` emit `channels` and `routing` so the editor round-trips what it already renders.
- Make `PUT /api/templates/{name}` non-destructive: read the existing TOML, merge field by field, or accept a full `manifest_toml` the way `PATCH /api/agents/{id}` already does (`crates/librefang-api/src/routes/agents/lifecycle.rs:1277-1296`).
- Regression test: write an agent type TOML carrying `[compaction]` and `max_history_messages`, `PUT` a flat body, assert both survive; and assert `channels`/`routing` round-trip through GET → PUT → GET.

houko:
Implements agent-type create / read / update / delete on `/api/templates`, plus the dashboard editor that drives it.

Closes #7740
Closes #7731
Refs #7722

## Credit, and why this PR exists

The analysis and the design are @DaBlitzStein's — #6931 for the CRUD work, #7740 for the destructive-`PUT` report, #7731 for the editor bullets. I have implemented their conclusions; the thinking is theirs.

#6931 is stacked roughly a hundred commits deep on unmerged parents, so its ~10,000-line diff is almost entirely borrowed code around a single commit of its own. That shape is what stalled it — reviewers cannot separate the contribution from the base. This PR is the same substance written directly on top of current `main`, where the diff is small enough to actually read.

I would recommend closing #6931 as superseded once this lands. That is a recommendation only — I have not commented on, labelled, closed or pushed to it.

## Premise check first

Grepped `origin/main` before writing anything:

- `crates/librefang-api/src/routes/agent_templates.rs` registers read handlers only — `list_profiles`, `get_profile`, `list_agent_templates`, `get_agent_template`, `get_agent_template_toml`.
- `agent_type_json_to_toml`, `manifest_to_agent_type`, `agent_types_dir` and `dashboard/src/pages/AgentTypesPage.tsx` exist nowhere in the tree.
- The catalog on `main` is *single*-source: it reads `workspaces/agents/` only, so the operator-authored store the write verbs need does not exist either.

So nothing here is a port of an existing implementation. The endpoints, the storage layout, the shared merge type and the dashboard page are all written from scratch against the defect analysis on the two issues.

## The central change: `PUT` is a patch, not a constructor

`AgentManifest` has 58 fields. The editor shape carries 7. Rebuilding the document from those 7 and writing it over the operator's file — the approach on #6931 — resets the other 51 to their defaults and answers `200`: `[[triggers]]`, `[compaction]`, `max_history_messages`, `mcp_servers`, `tool_allowlist`, `session_mode`, `[workspaces]`, `channels`, `[exec_policy]` and `fallback_models` all disappear the first time anyone opens the editor and saves.

`update_agent_type` therefore parses the stored manifest, applies the request body over it field by field, and writes the merged result. A key the client did not send leaves its field exactly as it was found.

## Substantive changes

**`crates/librefang-types/src/agent_type.rs`** (new) — the shared merge rules, in `librefang-types` rather than in the route, so the agent-facing `agent_type_create` tool of #7722 inherits them by using the same type instead of re-deriving them.

- `AgentTypeSpec` — the flat shape as a patch. Every field is `Option`, so absent and empty stay distinguishable: `Some("")` clears, `None` keeps. `unwrap_or("You are a helpful AI agent.")` cannot express that difference, which is how a deliberately blank system prompt became canned text on disk.
- `#[serde(deny_unknown_fields)]` — under patch semantics a typo'd key would otherwise deserialize to "field absent", read as "keep the old value", drop the edit, and still answer `200`.
- `AgentTypeSpec::apply_to` — the merge. `agent_type_spec_of` — its exact inverse, for the `GET` side; the two sit adjacent deliberately, because a field one can write that the other does not emit is precisely what loses data across a `GET` → edit → `PUT` cycle.
- `AgentTypeSpec::into_new_manifest` — the create path, and the only manifest-construction path in the change. **It writes an exhaustive struct literal with no `..Default::default()` rest pattern.** That is the structural guard, not a style preference: `AgentManifest` gains fields regularly, and a rest pattern keeps compiling while silently deciding, on behalf of whoever added field 59, what a newly-created agent type does with it. Spelling all 58 out converts that into a compile error at the one place where the decision has to be made, and the doc comment tells the next person to add the field with a reason rather than reach for the rest pattern. `ModelConfig` and `ManifestCapabilities` are exhaustive for the same reason.

**`crates/librefang-api/src/routes/agent_templates.rs`** — the endpoints.

- `POST /api/templates` (201), `PUT /api/templates/{name}` (200), `DELETE /api/templates/{name}` (200), merged into the existing domain router. Storage is `~/.librefang/agent-types/{name}.toml`, one flat file per type so a create or an edit is exactly one atomic rename with nothing to leave half-built.
- **Atomic writes.** `persist_agent_type` renders and lands the document through the existing `crate::atomic_write` (temp file, `sync_all`, `rename`). `std::fs::write` truncates in place, so a failure partway through would leave a truncated `agent.toml` where a valid one used to be — the worst failure mode for a file the daemon parses at spawn. A manifest that cannot be re-rendered as TOML is refused with a 500 and nothing is written.
- **Create is race-free.** `Path::exists()` followed by a write is check-then-act: two concurrent creates of one name both see "absent" and the second silently replaces the first, defeating the 409. `create_agent_type_file` claims the path with `File::create_new` — atomic create-if-absent at the OS level — and removes the claim if the fill fails. This was flagged on #6931 and deferred there; it is in the code I was already writing, so it is fixed here.
- **Dual-source guard.** `GET` reads two sources — the `agent-types/` store and live agent workspaces — while the write verbs only reach the first. Every row now carries `source` and `editable`, and `PUT`/`DELETE` aimed at a workspace-sourced name answer `409 template_not_editable` rather than a bare `404` that would tell an operator nothing about why a row they can see refuses to save. `POST` refuses a name that already belongs to a live agent (`template_name_taken`) so the shadowing collision cannot be created deliberately; one arising afterwards resolves to the writable copy and is logged at `WARN`.
- Identity comes from the URL: `PUT` discards a `name` in the body, which would otherwise move the document out from under the path that addressed it.
- `GET /api/templates/{name}` gains `source`, `editable` and `spec` (the flat projection) **alongside** the pre-existing nested `manifest` and `manifest_toml`, which existing clients read. Nothing was flattened away, so `templates_get_known_template_returns_manifest` still passes — #6931 broke that test by reshaping the response.
- Nine new error strings across all nine `errors.ftl` locales. The four locales that were already missing `api-error-template-invalid-manifest` / `-read-failed` got those too, since I was in the files.

**Dashboard** — `src/pages/AgentTypesPage.tsx` plus `lib/queries/agentTypes.ts`, `lib/mutations/agentTypes.ts`, an `agentTypeKeys` factory in `keys.ts` with coverage in `keys.test.ts`, a route, a nav entry, an `agent-types` entry in `SPA_ROUTES` (`webchat.rs`), and 35 strings in all five locales.

- Skills and tools are **catalog-backed `MultiSelectCmdk` pickers** over `GET /api/skills` and `GET /api/tools`, unioned with what the type already references so an identifier the registry does not know still renders as a chip rather than vanishing from the form and, with it, from the saved document (#7731 bullet 1).
- `skills` round-trips. #6931's form had no `skills` member at all, so every save through it emptied the skill list.
- Workspace-sourced rows render **"Managed via Agents"** linking to `/agents` instead of Edit/Delete controls the API refuses (#7731 bullet 2).
- Provider and model are free text with an explicit `default` placeholder and a hint saying what `default` inherits — no canned substitution client-side either.
- A standing note in the editor says the save rewrites only the fields shown and preserves the rest, so the patch semantics are visible to the operator rather than implied.

**TUI parity.** `crates/librefang-cli/src/tui/event.rs` read only `workspaces/agents/` on the in-process backend. Left alone it would have silently omitted every type this PR lets you create, while the daemon backend showed them — a defect this change would have introduced. It now reads both sources with the same precedence as the API, and resolves a spawn through `local_agent_type_path`. This is the one place I crossed a crate boundary; it is around sixty mechanical lines, and leaving it would have shipped a surface disagreeing with itself.

**Docs.** `docs/architecture/agent-types.md` (new) covers storage, dual sourcing, the patch semantics and the exhaustive-literal mechanism. `docs/src/app/integrations/api/system/page.mdx` and its `zh` counterpart document the new verbs and the corrected response shapes. #7731's acceptance named `docs/architecture/agent-created-workflows.md`, which does not exist on `main`; the content landed at the path above.

## Auth

The new verbs inherit the existing blanket `role >= Admin` write gate and are deliberately **not** added to `is_owner_only_write`. That list Owner-gates surfaces expanding what can execute *beyond what an Admin already has*. An Admin can already `POST /api/agents` with an arbitrary manifest — any provider, any system prompt, any tool grant — so gating agent types at Owner while leaving direct spawn at Admin buys nothing and only makes the two inconsistent. Stated here rather than decided quietly, since the question was raised on #6931.

## Verification

- `cargo test -p librefang-api --test agent_types_routes_integration` — 10 `#[tokio::test]`s, all against the real production router from `server::build_router`, so a handler that exists but was never merged into `server.rs` fails here.
  - `update_preserves_every_field_the_dashboard_form_never_sends` — **the test #6931 lacked.** Seeds a manifest carrying `[[triggers]]`, `tool_allowlist`, `mcp_servers`, `max_history_messages`, `session_mode` and `[compaction]`, sends exactly the 7-key body the dashboard sends, then asserts all six survived on disk. It fails against a rebuild-from-body handler and passes against the merge.
  - `update_writes_blank_fields_through_instead_of_substituting_canned_text` and `create_writes_a_blank_system_prompt_through_unchanged` — one per canned-substitution case.
  - `skills_round_trip_through_the_editor_shape` — create, GET, replay the GET verbatim, PUT omitting `skills`, PUT with an explicit `[]`; asserting keep, keep, clear.
  - `a_workspace_agent_row_is_readable_but_refuses_the_write_verbs` — the dual-source guard, including that the live agent's manifest is untouched.
  - `create_refuses_a_name_that_belongs_to_a_live_agent`, `crud_lifecycle_through_the_production_router` (which also asserts a refused duplicate did not overwrite the original), `write_verbs_reject_names_that_would_escape_the_agent_types_directory`, `a_typo_in_a_save_body_is_rejected_rather_than_silently_ignored`, `update_pins_identity_to_the_url_rather_than_the_body`.
- `cargo test -p librefang-types agent_type` — 8 unit tests on the merge rules themselves.
- `cargo test -p librefang-api --test dead_route_audit_test --test openapi_path_coverage_test --test profiles_templates_routes_integration --test webchat_spa_fallback_test` — the drift guards, plus the 16 pre-existing template tests, all green.
- `cargo clippy -p librefang-api --lib --tests -- -D warnings`, `cargo clippy -p librefang-cli --bins -- -D warnings`, `cargo fmt --check` — clean.
- `pnpm run build` (the real Vite build, not just `tsc`) and `pnpm run test` — 125 files, 1210 tests, green.

Live daemon plus a real LLM was not exercised; nothing here touches an LLM call path.

## Not done here

`openapi.json` and the generated SDKs are not regenerated in this branch — `cargo xtask` is outside what I run locally. The three handlers are registered in `openapi.rs`, which is what `dead_route_audit_test` and `openapi_path_coverage_test` check, so the `openapi-drift` job's auto-codegen commit should close the gap on push. If it does not, that regeneration is the one follow-up.


crates/librefang-api/src/routes/agent_templates.rs
```diff
@@ -382,27 +565,271 @@ pub async fn get_agent_template_toml(
...
+            // parse is exactly the data loss this handler exists to prevent.
+            tracing::warn!("Invalid agent type manifest for '{name}': {e}");
+            return ApiErrorResponse::internal(invalid_manifest).into_json_tuple();
+        }
+    };
+
+    spec.apply_to(&mut manifest);
+    manifest.name = name.clone();
+
+    match persist_agent_type(&name, &manifest) {
```
houko, line 765: `update_agent_type`'s read-modify-write (read at line 730, `apply_to` at 762, `persist_agent_type` at 765) has no lock, ETag, or version check. Two concurrent `PUT`s to the same name both read the same stored manifest, apply different patches, and the second write silently wins — both requests get `200`, and one edit is lost with no signal to either caller.

This is the same class of problem `create_agent_type_file`'s `File::create_new` deliberately hardens against on the create path, but nothing analogous exists here. Given the PR's stated goal is "no silent data loss," is this an accepted trade-off for a low-traffic operator-admin surface, or worth a per-file mutex / optimistic version check? All the new integration tests are sequential, so this path isn't exercised either way.

---

crates/librefang-api/src/routes/agent_templates.rs
```diff
@@ -382,27 +565,271 @@ pub async fn get_agent_template_toml(
...
+    NameTaken,
+    Failed(String),
+}
+
+/// Write a new agent type, refusing to overwrite one that already exists.
+///
+/// `Path::exists()` followed by a write is check-then-act: two concurrent creates of the same name both observe "absent" and the second silently replaces the first, which is exactly the 409 this endpoint promises not to do.
+/// Claiming the path with `File::create_new` — an atomic create-if-absent at the OS level — lets exactly one of them through.
+/// The claim is then filled by the same atomic rename every other write here uses, and removed again if that fails, so a failed create leaves no empty file behind for the catalog to trip over.
+fn create_agent_type_file(name: &str, manifest: &AgentManifest) -> CreateOutcome {
```
houko, line 616: `create_agent_type_file` claims the name with `File::create_new` (a real, empty, 0-byte file on disk) and only then fills it via `atomic_write`. If the process is killed between the `create_new` success and the `atomic_write` call (or between that call and the cleanup `remove_file` on its error path), a zero-byte, unparsable `{name}.toml` is left behind permanently: future `POST` sees the name as taken (409), and `GET`/`PUT` fail on it with `invalid_manifest`/500, with no reconciliation path found in this diff.

Low-probability window, but is this an accepted trade-off of the create-if-absent approach, or worth a startup-time sweep for zero-byte/unparsable files under `agent-types/`?

---

crates/librefang-api/dashboard/src/lib/queries/keys.ts
houko, line 1: `agentTypeKeys.*` is a cache fully separate from the pre-existing `agentKeys.templates()`, even though both back `GET /api/templates` (the latter feeds the spawn-form template picker on `AgentsPage.tsx`). The create/update/delete mutations in `lib/mutations/agentTypes.ts` invalidate only `agentTypeKeys.*`, never `agentKeys.templates()`, so the spawn-form dropdown can show a stale template list right after an edit made from the new Agent Types page.

In practice `agentKeys.templates()`'s query has no explicit `staleTime` and the spawn form is a different route, so a remount likely refetches — the real-world staleness window may be negligible. Flagging so a human confirms this dual-cache-of-one-resource split is intentional rather than an oversight of the invalidation contract; if intentional, a one-line comment on `agentTypeKeys` noting the split would save the next reader the same question.

---

houko (commented):
Automated daily review pass. Pushed a small mechanical fix (reflowed a few hard-wrapped doc comments to one sentence per line per CLAUDE.md's prose-wrapping rule); everything else checked out clean against project conventions (deterministic ordering, atomic writes, exhaustive struct literals, auth wiring, path-traversal guards, locale parity, changelog fragment format, dashboard query/mutation conventions). Left three inline comments on real but judgment-call concerns — none of them are asserted as bugs, just worth the author's confirmation.

---

PR #7862, opened 2026-08-24
feat(workflows): reversible step agent binding and per-step session choice

houko:
## Credit

The analysis and the design of this are @DaBlitzStein's, in #7728 and the issue thread of #7724.
The reversible-binding shape here is the fix for the one review item on that PR, and the agent-type / fresh vocabulary comes from their #7715 and #7717.

#7728 is stacked about a hundred commits deep on those two unmerged parents, so its diff is roughly 10,500 added and 3,700 removed lines across 122 files — almost all of it borrowed from #7715, #7717, #7722 and #6931 — while its own contribution is the single commit `377f45f`.
Required checks and review on it are exercising other people's code, and it has stalled there.
That diff also deletes about 2,943 lines of `CHANGELOG.md` as a rebase artifact.
This PR carries the same intent on top of current `main` where it is small enough to read; **I'd recommend #7728 be closed as superseded**, but that is a recommendation for its author, not a decision — I have not touched it.

## The premise did not hold

The issue says #7715/#7717 "wired step agent types (find-or-spawn from a template, with per-run fresh instances) through the kernel, all resolver sites, and the API parser".
On `main` they did not, because neither PR has merged:

- `StepAgent` in `crates/librefang-kernel/src/workflow.rs` has exactly two variants, `ById` and `ByName`. There is no `ByType`, no `template`, no `fresh`.
- `create_workflow` / `update_workflow` reject any step without `agent_id` or `agent_name` — a payload carrying `type` gets a 400.
- There is no `useAgentTypes` hook and no `/api/agent-types` route.

So the agent-type source cannot be exposed yet: a selector emitting `type` would build a control whose every use fails at the API boundary.
I have deliberately **not** re-implemented the kernel half here — #7715 is already open, based on `main`, not a draft, and reviewable at +489/-15; duplicating it would step on live work and conflict with it.
The type source lands when #7715 and #7717 do, and slots into the third arm of the source selector this PR introduces.
That is why this is `Refs #7724`, not `Closes`.

### Canvas step editor: an explicit, reversible agent source

`components/StepAgentBinding.tsx` (new) owns the binding. The source — `instance` or `name` — is state, and only the control belonging to the active source is rendered.

That shape is the review item. The alternative is one `<select>` whose controlled value is forced to `""` while another binding field is set; that select can never emit the value that would clear the other field, so once a step leaves the concrete-agent binding it cannot come back. Rendering exactly the control that belongs to the active source removes the door rather than widening it: no control is ever fed a value contradicting its own state.

The switch is lossless in both directions. Leaving `instance` carries the selected agent's name into the name field; returning re-selects the live agent answering to that name, and falls back to an empty, still-selectable dropdown when none does.

`agent_name` is not a lesser binding — it is the durable one. An agent's id changes when it is killed and respawned; its name does not. The canvas previously offered only `agent_id`, so a workflow authored in it silently stopped resolving after a respawn.

### Per-step session selection

`WorkflowStep::session_mode` is fully wired on `main` — kernel field, `parse_step_session_mode` at both POST and PATCH, and `workflow_to_json` — and had no UI on any surface.
The panel now exposes all three states: agent default, reuse the agent's session, fresh session each run.

Note the vocabulary difference from #7717 and don't conflate them: this is a fresh **session** (a clean conversation for the step), which `main` supports today. #7717's `fresh` is a fresh **agent instance** spawned per run. Both are wanted; only the first is reachable.

### Three round-trip bugs found while reading the same code

- **`GET /api/workflows/{id}` shape mismatch.** `workflow_to_json` emits `steps[].agent` as `{"agent_id": …}` / `{"agent_name": …}`; `CanvasPage` read `agent.id` / `agent.name`. Both are always `undefined`, so a workflow with no saved layout hydrated as nodes with no agent, rendered as "click to assign", and `buildSteps` dropped every one of them — loading and re-saving such a workflow emptied it. Nothing pinned that contract; a test does now.
- **Name-bound steps were stripped on save.** `NodeConfigPanel` rebuilt `agentName` from `agents.find(a => a.id === agentId)`, so opening a step bound only by name and pressing Save unbound it. The node's "missing agent" badge and the run animation read `data.agentId` directly for the same reason and had the same blind spot.
- **The TUI workflow creator could not create anything.** `spawn_create_workflow` posted `"steps": steps_json` — the raw string, not an array — and `create_workflow` reads `req["steps"].as_array()`, so every submission was rejected with `Missing 'steps' array`. The rejection was then reported as a success, because the match arm was `Ok(resp)` with no status check, and the created id was read from `body["id"]` when the route answers `workflow_id`. All four are fixed.

### TUI parity: scoped honestly

The issue asks for the same fields "wherever steps are authored" in the TUI.
The TUI's workflow screen authors steps as **one free-text JSON blob** (`create_steps`, wizard stage 2), not per-field controls — so there is no selector to add, and it already accepts any field the API parses once the blob actually reaches the API as an array.
What was wrong there was the blob itself: the placeholder taught `[{"action": "..."}]`, a key the API has never accepted.
It now shows the real shape, including `session_mode`. Per-field TUI step authoring would be a new screen, not a field, and is not in this PR.

## Files

- `crates/librefang-api/dashboard/src/components/StepAgentBinding.tsx` — new: the control plus its pure helpers (`bindingFromNodeData`, `bindingToNodeData`, `switchAgentSource`, `stepAgentFields`, `isStepBound`, `normalizeSessionMode`).
- `crates/librefang-api/dashboard/src/pages/CanvasPage.tsx` — uses it; fixes the load shape, the save path, the bound-step predicate in `buildSteps` / the node badge / the run animation.
- `crates/librefang-api/dashboard/src/lib/canvas.ts` — `agentSource` / `sessionMode` on `CanvasNodeData`; the panel's two Tailwind class constants moved here so the panel and the extracted control share one definition.
- `crates/librefang-api/dashboard/src/api.ts` — `WorkflowStep.session_mode`.
- Five dashboard locales — 11 new `canvas.*` keys each.
- `crates/librefang-cli/src/tui/event.rs` — `parse_workflow_steps_json` + `StepsJsonError`, status handling, `workflow_id`.
- Four CLI locales — 4 new keys each, plus the corrected steps placeholder.
- `crates/librefang-api/tests/workflows_routes_integration.rs` — the binding-vocabulary drift guard.

## Verification

- `pnpm test` (vitest): **1226 passed, 126 files**, including the new `StepAgentBinding.test.tsx` — 18 cases covering the source switch in both directions, the lossless round trip, a hand-typed name surviving a bounce through `instance`, the session selector set and cleared, the payload projection, and the datalist.
- `pnpm run typecheck` clean; `pnpm run lint` clean on the touched files.
- `pnpm run build` — real Vite build, green (not just `tsc`).
- `node scripts/i18n-parity.mjs` — the five dashboard locales are at parity on the new keys; the script's only complaints are the eight pre-existing Slavic `_few`/`_many` plural forms it does not model and the vitest gate does.
- `cargo check -p librefang-cli --bins` clean.
- `cargo test -p librefang-cli --bins workflow_steps_json` — 3 passed.
- CLI locale coverage verified against the rules in `tests/i18n_checks.rs`: all four locales at 2032 keys, zero drift, every new key referenced from Rust and present everywhere.
- `cargo fmt` clean.

**Not verified locally:** the new test in `workflows_routes_integration.rs` has not been compiled or run. The shared build volume this worktree uses fell to 11 GiB free while I was working and my build budget stops at 12 GiB, so I stopped rather than risk filling the disk. Same reason for `cargo clippy -p librefang-cli`. Both are on the CI path; the test uses only the `boot` / `json_request` / `get` helpers its immediate neighbour in the same file uses.

## Deferred, deliberately

- The agent-type (template, find-or-spawn) source, and a fresh-agent-instance-per-run toggle. Both need `StepAgent::ByType { template, fresh }`, which is #7715 + #7717. The selector added here takes a third arm without restructuring.

Refs #7724


houko:
Automated review pass — verification and one merge-conflict flag, no code issues found.

**Verification (filling the gaps noted as untested locally):**
- `cargo check -p librefang-cli --bins` — clean
- `cargo clippy -p librefang-cli --all-targets -- -D warnings` — zero warnings
- `cargo test -p librefang-cli --bins workflow_steps_json` — 3/3 pass
- `cargo test -p librefang-api --test workflows_routes_integration workflow_get_names_the_step_agent_binding_the_way_create_reads_it` — pass

Changelog fragments, locale parity (5 dashboard + 4 CLI locales, same 4 keys each), and commit message all match this repo's conventions.

**Merge conflict with `main`:** this branch is based on `c571623`, and `main` has since picked up #7858 (`StepAgent::ByType` / find-or-spawn agent types) — the very feature this PR's description says it deliberately deferred ("slots into the third arm of the source selector this PR introduces"). That feature made overlapping edits to the same files this PR touches: a test merge locally shows real conflicts in `crates/librefang-api/dashboard/src/lib/canvas.ts` (1 hunk) and `crates/librefang-api/dashboard/src/pages/CanvasPage.tsx` (6 hunks) — not just adjacent-line noise, but both sides changing the step/agent-binding shape.

Since #7858 already landed, reconciling this needs a design call about how the `instance` / `name` source selector in `StepAgentBinding.tsx` should now compose with the `type` source — that's a decision for @houko, not something to resolve mechanically. Flagging so it doesn't get merged with unresolved conflicts once the rebase is attempted.

---

PR #7867, opened 2026-08-24
fix(memory): give the prompt memory section an owned budget

nevgenov:
Lands the half of #7756 §1.3 that stands on its own, as suggested in that thread: give the memory section an owned budget and stop cutting bullets mid-sentence. No plugin, no seam, independent of the RFC outcome.

## Problem

Three producers feed the memory section and none of them owns its total.

- Substrate recall is capped at `MEMORY_RECALL_LIMIT = 5` (`agent_loop/prompt.rs:309`), fetched wider and truncated back.
- Proactive items are appended after that cap.
- The formatter then took the first ten and capped each at 500 characters.

So the section's real ceiling was the product of two independent caps — nothing stated it, and `cap_str` sits downstream of all three producers knowing nothing about how many contributed.

The per-item cut was also unconditional, landing mid-clause whenever content exceeded it. Measured on one production corpus: 57% of raw memory rows exceed 500 characters, median length 599, while no extracted fact does — median 105. So the truncation this fixes is almost entirely a property of one class of memory.

## Change

**An explicit section budget** at the value the ceiling already implicitly had, so the limit becomes stated and enforceable without changing how much context an agent receives.

**Per-item share recomputed as the section is built.** An item shorter than its share hands the surplus to later items, which is the common case given the length asymmetry above. A per-item ceiling keeps a single oversized row — an attachment stored verbatim, for instance — from filling the section when few memories are present, and the key label is capped so an unbounded one cannot spend its item's whole share.

**Framing charged up front.** The `- `, the newline and the `[key] ` wrapper come out of the allowance rather than being discovered afterwards, and the omission notice is reserved for before any bullet is written — the discipline `librefang_types::memory::format_memories_with_budget` already uses. A final clamp holds the contract in release, where `debug_assert` is compiled out.

**Sentence-boundary truncation**, falling back to a word boundary and then a hard cut. Both boundary searches reject a break before the halfway mark of the window, since an early one discards more than it keeps: an unbroken tail such as a URL otherwise collapses a bullet to whatever preceded its first space. Every shortened bullet carries the marker, so a memory cut at a full stop is not mistaken for one that ended there.

**Dropped items are reported** with their count.

## Tests

Every property is pinned by a test verified to fail under a mutation that reintroduces the defect: the budget total, the per-item share, the per-item and key ceilings, each truncation branch, the marker surviving the clamp, and the notice reserve.

Two things learned the hard way and applied throughout:

- Sweeps are derived from the constants they depend on. A literal sweep only covers the window its constants happened to produce when it was written, and stops covering as soon as one moves.
- Bounds that guard a constant are literals, for the mirror-image reason: derived from the constant they guard, they move with it and stop detecting the change.

The three constant changes documented as tripwires each fail a test rather than silently altering behaviour.

## Verification

- `cargo test -p librefang-runtime --lib prompt_builder` — 105 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2346 passed, 2 failed; the same two as on the untouched base (`browser::tests::live_extraction_…` needs a Chromium binary that is absent here, and `agent_context::tests::poisoned_cache_lock_recovers_cached_context` flakes under the parallel run on the base too). Verified by stashing this change and re-running there.
- `cargo check --workspace --lib`, `cargo clippy -p librefang-runtime --lib --tests -- -D warnings`, `cargo fmt --all -- --check` — clean.

The branch went through nine adversarial self-review passes before this was opened for review, each in a separate session against the previous round's diff. They found, among other things, that the budget was not actually enforced (5074 against a stated 5000), that a long key could render the section empty, and several tests that did not fail under mutation of the behaviour they were named for. All of it is folded into the single commit here; I have not kept the intermediate history, since it is my process rather than part of the change.

## Not included

The other half of §1.3 — who owns the budget across the three producers rather than only within the formatter — needs the producers to agree on a contract and belongs with the seam discussion. This bounds the section where it is assembled, which holds regardless of how that resolves.

Refs #7756

### DIFF
--- a/changelog.d/fixed/7867-memory-section-budget.md
+++ b/changelog.d/fixed/7867-memory-section-budget.md
@@ -1,0 +1,4 @@
+The prompt's memory section now enforces an explicit character budget instead of relying on the product of two independent caps that nothing stated or checked.
+A shortened memory is cut at a sentence boundary where one fits and at a word boundary where that keeps most of the text, so a bullet ends mid-token only when neither is available.
+Memories dropped for space are reported with their count.
+(#7867) (@nevgenov)
--- a/crates/librefang-runtime/src/prompt_builder.rs
+++ b/crates/librefang-runtime/src/prompt_builder.rs
@@ -615,6 +615,172 @@ pub fn build_memory_section(memories: &[(String, String)]) -> String {
     out
 }
 
+/// Maximum number of memory bullets rendered into the section.
+pub const MEMORY_SECTION_MAX_ITEMS: usize = 10;
+
+/// Character budget for the memory bullets, excluding the framing header.
+///
+/// The section has three independent producers — substrate recall capped at `MEMORY_RECALL_LIMIT`, proactive items appended after it, and the `MEMORY_SECTION_MAX_ITEMS` cut here — and before this budget existed none of them owned the total.
+/// Each item was independently capped, so the section's real ceiling was the product of the two caps and nothing stated it.
+///
+/// The value is the former implicit ceiling (10 items × 500 chars), but it is a ceiling on the section rather than on content alone, so the reachable maximum is lower.
+/// The `- ` and newline framing each bullet is charged against its allowance, and a notice reserve is held back whenever memories were dropped.
+/// Measured worst case is 5000 characters of body — the budget exactly — against the 5060 the unbudgeted version produced, about 1 % less content in exchange for a limit that is stated and enforced.
+/// At roughly four characters per token the budget is about 1250 tokens.
+pub const MEMORY_SECTION_BUDGET_CHARS: usize = 5000;
+
+/// Below this an item is not worth rendering — a fragment this short carries no usable context and only spends budget.
+///
+/// The equal share never decreases — a bullet can never cost more than the share it was granted — and its floor is about 490.
+/// The guard compares the share minus this item's framing, whose worst case is `MAX_BULLET_FRAMING_CHARS` — 73 with the key capped at 64, since `cap_str` adds three characters of its own — so it cannot fire as the constants stand.
+/// It is kept as a tripwire: raising the item count, lowering the budget or widening the key cap should drop a memory rather than emit an unusable fragment, and skipping one item leaves the rest of the section intact.
+pub const MEMORY_SECTION_MIN_ITEM_CHARS: usize = 120;
+
+/// Characters held back for the omission notice.
+///
+/// The notice is `- [+N further memories not shown]\n`: 33 characters of fixed text plus the count.
+/// The count is bounded only by the caller's input, so the widest form is a twenty-digit `usize`, giving 53.
+/// The reserve keeps headroom above that so a future rewording does not silently overrun.
+/// Held back before any bullet is emitted, because discovering afterwards that the notice does not fit leaves no way to honour the budget without dropping a bullet that was already rendered.
+const OMISSION_NOTICE_RESERVE: usize = 96;
+
+/// Worst-case framing charged to one bullet: `- `, the newline, the `[key] ` wrapper, the capped key, and the `...` that `cap_str` appends when it truncates.
+///
+/// The ellipsis is counted here for the same reason `SKILL_BOILERPLATE_OVERHEAD` counts it — `cap_str` returns up to `max_chars + 3`, so a cap of 64 renders up to 67.
+const MAX_BULLET_FRAMING_CHARS: usize = MEMORY_SECTION_MAX_KEY_CHARS + 9;
+
+/// Whether an item can be skipped for its own framing cost.
+///
+/// Derived from the constants rather than asserted, so that widening the key cap, raising the item count or lowering the budget flips it automatically.
+/// When it is true a notice can appear with fewer memories than the item cut allows, and the reserve has to be held for that case too — otherwise the notice is written past the budget.
+const SKIP_POSSIBLE: bool = skip_possible(
+    MEMORY_SECTION_BUDGET_CHARS,
+    MEMORY_SECTION_MAX_ITEMS,
+    MEMORY_SECTION_MIN_ITEM_CHARS,
+    MAX_BULLET_FRAMING_CHARS,
+);
+
+/// Whether the smallest equal share can fall under the floor once an item's
+/// worst-case framing is taken out of it.
+///
+/// A free function rather than an inline expression so the rule can be exercised at inputs the shipped constants do not reach.
+/// Comparing `SKIP_POSSIBLE` against a re-derivation beside it proves nothing: a definition that drops a term still evaluates to the same value at the shipped constants, and the disagreement only appears once one of them moves.
+const fn skip_possible(budget: usize, items: usize, floor: usize, framing: usize) -> bool {
+    budget / items < floor + framing
+}
+
+/// Ceiling for the rendered key label.
+///
+/// Keys reach this function through a `pub` entry point and a `pub` field, and nothing upstream bounds their length.
+/// An unbounded key is charged against its own item's allowance, so without this cap a long enough label pushes the allowance below `MEMORY_SECTION_MIN_ITEM_CHARS` and the memory cannot render at all.
+/// A label longer than this carries no additional meaning to the model.
+/// `cap_str` appends `...` when it truncates, so the rendered label reaches 67; `MAX_BULLET_FRAMING_CHARS` accounts for that.
+pub const MEMORY_SECTION_MAX_KEY_CHARS: usize = 64;
+
+/// Ceiling for a single bullet, independent of how much budget is free.
+///
+/// The section budget alone is not sufficient protection: with few memories the per-item share grows to the whole section, and a single pathological row — an attachment stored verbatim, for instance — would then fill it.
+/// Keeping the historical per-item cap means this change makes the total explicit without altering how much any one memory may contribute.
+pub const MEMORY_SECTION_MAX_ITEM_CHARS: usize = 500;
+
+/// Marker appended whenever content was dropped, so the model can tell a shortened memory from a complete one.
+const TRUNCATION_MARKER: &str = "...";
+
+/// Truncate at the last sentence boundary that fits, falling back to a word boundary and then to a hard cut.
+///
+/// Cutting mid-sentence is what the previous fixed per-item cap did, and on a corpus where raw conversational excerpts dominate it left the majority of bullets ending mid-clause.
+/// A truncated sentence reads as a claim the speaker did not finish, which is worse than a shorter but complete one.
+///
+/// Every truncating branch appends [`TRUNCATION_MARKER`], including the sentence-boundary one.
+/// A cut that ends on a full stop is indistinguishable from a memory that simply ended there, and silent shortening is worse than visible shortening: the model has no way to know context is missing.
+///
+/// The returned string never exceeds `max_chars` including the marker, so callers can budget against the limit they passed rather than against the limit plus an unstated overhead.
+fn truncate_at_sentence_boundary(s: &str, max_chars: usize) -> String {
+    if s.chars().count() <= max_chars {
+        return s.to_string();
+    }
+    // The marker is part of the budget, so the content window shrinks to make
+    // room for it.
+    let marker_chars = TRUNCATION_MARKER.chars().count();
+    let content_chars = max_chars.saturating_sub(marker_chars + 1);
+    if content_chars == 0 {
+        // Unreachable from the section, whose allowance floor is
+        // `MEMORY_SECTION_MIN_ITEM_CHARS`, but the helper is `max_chars`-generic
+        // and must not hand back more than it was given.
+        // Counted in chars, not bytes: `safe_truncate_str` takes a byte count,
+        // and passing a char count to it only happens to work while the marker
+        // is ASCII.
+        return TRUNCATION_MARKER.chars().take(max_chars).collect();
+    }
+    let cut = s
+        .char_indices()
+        .nth(content_chars)
+        .map(|(i, _)| i)
+        .unwrap_or(s.len());
+    let window = safe_truncate_str(s, cut);
+
+    // Terminal punctuation followed by whitespace, or at the very end of the
+    // window. Bare `.` is not enough — abbreviations and version numbers.
+    let sentence_end = window
+        .char_indices()
+        .filter(|&(i, c)| {
+            if !matches!(c, '.' | '!' | '?' | '…') {
+                return false;
+            }
+            let after = &window[i + c.len_utf8()..];
+            after.is_empty() || after.starts_with(char::is_whitespace)
+        })
+        .map(|(i, c)| i + c.len_utf8())
+        .next_back();
+
+    // A boundary before the halfway mark discards more than it keeps, so it is
+    // rejected in favour of the next strategy down.
+    // Both boundary searches need this guard, not just the sentence one: the
+    // word fallback runs precisely when the sentence guard rejected a break,
+    // and `rfind` is just as capable of landing at position 1 when the window
+    // ends in an unbroken token such as a URL, a path or a hash.
+    let halfway_ok = |end: usize| end * 2 >= window.len();
+
+    if let Some(end) = sentence_end.filter(|&e| halfway_ok(e)) {
+        return clamp_to(
+            format!("{} {TRUNCATION_MARKER}", window[..end].trim_end()),
+            max_chars,
+        );
+    }
+
+    let out = match window.rfind(char::is_whitespace) {
+        Some(sp) if sp > 0 && halfway_ok(sp) => {
+            format!("{}{TRUNCATION_MARKER}", window[..sp].trim_end())
+        }
+        _ => format!("{}{TRUNCATION_MARKER}", window.trim_end()),
+    };
+
+    clamp_to(out, max_chars)
+}
+
+/// Last line of defence for the `max_chars` contract.
+///
+/// The reserve is sized for the separator the sentence branch inserts, and an arithmetic slip there would be invisible in release, where `debug_assert` is compiled out.
+/// `format_memories_with_budget` clamps for the same reason.
+/// Applied on every return path, including the sentence branch — that branch is the one which spends the reserved character, so guarding only the paths that keep a spare would leave the case this exists for uncovered.
+fn clamp_to(s: String, max_chars: usize) -> String {
+    if s.chars().count() <= max_chars {
+        return s;
+    }
+    // Trim content, never the marker. A plain `take(max_chars)` eats the
+    // marker's tail instead, leaving a bullet that ends in `..` — the signal
+    // that context was cut, silently degraded into something that reads like
+    // punctuation. That failure is invisible to a length assertion, which is
+    // the only thing a clamp can otherwise be tested by.
+    let marker_chars = TRUNCATION_MARKER.chars().count();
+    if max_chars <= marker_chars {
+        return TRUNCATION_MARKER.chars().take(max_chars).collect();
+    }
+    let keep = max_chars - marker_chars;
+    let body: String = s.chars().take(keep).collect();
+    format!("{}{TRUNCATION_MARKER}", body.trim_end())
+}
+
 /// Format recalled memories as a natural personal-context block.
 ///
 /// Used by both the system prompt (appended to the Memory section) and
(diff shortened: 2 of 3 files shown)


nevgenov:
Ran an adversarial review pass over this branch before asking for yours. It found that the central claim of the PR body was false, so I am correcting that here rather than letting the diff quietly fix it.

## What I claimed and what was true

**"Give the section a token budget."** The budget was not enforced. Measured 5074 characters against a stated 5000 — three leaks, all charged after the fact:

- the per-item allowance covered content but not the `- ` and newline wrapped around it;
- `truncate_at_sentence_boundary` could return `max_chars + 3` once it appended the marker, so a caller budgeting against `max_chars` was short by the marker on every item;
- the omission notice was appended outside the budget entirely.

I cited `format_memories_with_budget` as the convention being followed. I had copied its notice wording and not its discipline: it reserves for the footer before emitting bullets (`max_chars.saturating_sub(64)`), clamps at the end, and pins the result with a `debug_assert`. This now does all three.

**"The truncation marker stays `...`."** Not on the sentence branch, which is the preferred one — it returned the text with no marker at all. So in the common case truncation became *silent*: a memory cut at a full stop was indistinguishable from one that ended there, and the model lost the signal that context was missing. That is worse than the mid-clause cut this PR set out to fix. Every truncating branch now appends the marker.

**"Without changing how much context an agent receives."** False for a specific and not-rare class. The word-boundary fallback used `rfind` with no guard on where the whitespace landed, while the sentence branch had exactly such a guard. A window ending in one unbroken token — a URL, a path, a base64 blob — put the last whitespace near the start and threw the rest away: 45 characters where the old code emitted 503. Both boundary searches now share the guard.

## Test defects

Three of the eight tests did not fail under mutation of the behaviour they were named for.

- `memory_section_total_stays_within_budget` used a fixture whose only whitespace sat at position 1, so every bullet collapsed through the path above and the section measured 138 characters against a 5000 ceiling. A budget test that never approaches the budget. Fixture now has no whitespace, and it counts the whole body rather than sampling lines that start with `- ` — that filter also missed continuations of multi-line memories, which are the dominant shape in a real corpus.
- The share test took the *longest* bullet, which is the last one, and the last is ceiling-bound regardless. Deleting the share recomputation outright left it green. It now compares the first bullet against the last: early items are share-bound, late ones ceiling-bound, and removing the share makes them equal.
- `truncation_is_utf8_safe` asserted `cut.is_char_boundary(cut.len())`, which is true of every `String`. Replaced with a check that the kept text is a prefix of the input.

The per-item ceiling had no test at all, and my first attempt at one derived its bound from the constant it was guarding — so raising the ceiling kept it green. It now asserts against a literal, with a comment saying why.

Each fix is pinned by a test I verified fails under a mutation that reintroduces the defect. Eight mutations, eight caught. Two of them initially appeared to survive and did not: both had made a constant dead and failed to compile under `-D warnings`, which my harness had been reading as a pass.

## Also

Added the `changelog.d/fixed/` fragment I had omitted, and unwrapped the new doc-comments to one sentence per line per the prose rule in `CLAUDE.md`.

## Verification

- `cargo test -p librefang-runtime --lib prompt_builder` — 94 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2335 passed, 2 failed; same two as on the untouched base, verified by stashing and re-running there.
- `cargo check --workspace --lib`, `cargo clippy -p librefang-runtime --lib --tests -- -D warnings`, `cargo fmt --all -- --check` — all clean.

The scope note from the original body still stands: this bounds the section where it is assembled, not the contract between its three producers.

nevgenov:
Second review pass, over the fix commit rather than the original. It found four defects that fix introduced or left, and two more claims of mine that were false. Correcting them here for the same reason as last time.

## Claims of mine that were wrong

**"The notice names whichever limit applied."** It cannot. A bullet never costs more than the share it was granted, so the share is non-decreasing and its floor is `MEMORY_SECTION_BUDGET_CHARS / MEMORY_SECTION_MAX_ITEMS` ≈ 488 — four times `MEMORY_SECTION_MIN_ITEM_CHARS`. The budget-stop branch is unreachable, so the notice always took one arm. Confirmed empirically across 16 configurations of count and length: the budget wording never appeared. The changelog line I wrote about naming the limit was a no-op, one always-constant string for another.

The branch is gone. The notice states the count and no cause, and the `MEMORY_SECTION_MIN_ITEM_CHARS` guard stays as a documented tripwire for a future constant change rather than as a live path — its doc no longer describes a stop that cannot happen.

**"Reserves, clamps, and pins with a `debug_assert` — this now does all three."** It did two. There was no clamp, and `debug_assert` is compiled out of release, so the invariant had no protection there at all. A clamp is in now. I want to be explicit that it is unreachable defence rather than tested behaviour: removing it leaves every test green, because the arithmetic that would trip it is correct. What pins that arithmetic is the new punctuation test below.

**`MEMORY_SECTION_BUDGET_CHARS` doc: "without changing how much context an agent receives."** Measured against the base formatter, it does change it: 4903 characters of body against 5060, about 3 % less. Both causes came from my fix commit — the bullet framing is now charged to the item allowance where it used to be free, and the notice reserve was held back unconditionally. The reserve is now taken only when something can actually be dropped, which returns 96 characters to the common case; the remaining difference is real and the doc says so.

## Tests that did not catch what they were named for

Two mutations survived the previous round.

- Measuring the truncation window in **bytes** instead of characters keeps roughly a third of the allowance for three-byte scripts — 251 characters instead of 499 for Cyrillic at a 500 limit — with every test green. `truncation_is_utf8_safe_and_faithful` asserted only that the kept text was a valid prefix of the input, which a naive byte cut satisfies. There is now a lower-bound assertion across two-, three- and four-byte fillers.
- The reserve is `marker_chars + 1`; the extra character pays for the separator the sentence branch inserts. Dropping it pushes the result past the limit the helper documents, and none of the four samples in `truncation_never_exceeds_its_limit` placed terminal punctuation at the window edge, which is the only position where that character is spent. A test now does exactly that, across four limits.

Also fixed: the `content_chars == 0` early return passed a character count to `safe_truncate_str`, whose parameter is a byte count. Unreachable from the section, and harmless only while the marker stays ASCII.

## Mutation coverage

Nine mutations, eight killed, each by the test named for that behaviour. The ninth is the release clamp, unreachable by construction as described above.

A note on method, since it bit me twice: `cargo test` prints `error: test failed` on a failing test, not only on a compile error, and my harness was treating any `error` line as "the mutation did not build". Combined with the earlier case where a mutation made a constant dead under `-D warnings`, I had two different ways of reading a caught mutation as a survivor and one of reading a survivor as unbuildable. Mutations are now discriminated on `could not compile` specifically, and the matrix above was re-run under that.

## Verification

- `cargo test -p librefang-runtime --lib prompt_builder` — 97 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2338 passed, 2 failed; the same two as on the untouched base.
- `cargo check --workspace --lib`, `cargo clippy -p librefang-runtime --lib --tests -- -D warnings`, `cargo fmt --all -- --check`, changelog attribution check — all clean.

nevgenov:
Third review pass, over the second fix commit. It found a defect that fix introduced, one it left, and four more claims of mine that were false. Same correction as before.

## The defect I introduced

**A long key emptied the section.** The previous commit started charging the `[key] ` framing against each item's allowance — that was the fix for the budget overrun. Nothing upstream bounds key length, and `format_memory_items_as_personal_context` is `pub` over a `pub` field. At 375 characters the allowance drops below `MEMORY_SECTION_MIN_ITEM_CHARS`, the guard fires, and its `break` discards every remaining memory too. Ten memories rendered **zero** bullets and 35 characters of body.

The label is now capped, which bounds framing at 70 against a share floor of 490, and the guard skips the item rather than ending the section.

## What it left

**The release clamp guarded the wrong branches.** Its own comment says it exists for the separator the sentence branch inserts — and it sat after that branch's early return. It covered only the word and hard paths, which by construction have a spare character, and not the one that spends it. Now applied on every return path.

## Claims of mine that were false

All four are numbers describing an earlier state of this same branch, written into docs and not revisited when the code moved under them in the same commit.

- `MEMORY_SECTION_BUDGET_CHARS` claimed a 4903-character worst case; measured 4998. It also claimed about 3 % less content than the base formatter where the real figure is about 1 % — 4998 against 5060. The 4903 was the value *before* I made the notice reserve conditional, three lines away in the same diff.
- `OMISSION_NOTICE_RESERVE` justified itself with "a twenty-digit `usize` gives 87 characters". That measures a notice wording this branch had already deleted; the current text gives 53.
- `MEMORY_SECTION_MIN_ITEM_CHARS` argued its own unreachability from the equal share never decreasing. True of the share, but the guard compares the share *minus framing*, which is a different quantity — and that is exactly the gap the key walked through.
- The changelog fragment described the URL collapse and the missing truncation marker as fixed defects. Both were introduced and repaired inside this branch, never appeared in a release, and `changelog.d/README.md` says fragments ship in the release body verbatim. A reader would have been told about regressions they never had.

## The test that did not test what it was named for

`truncation_respects_its_limit_with_punctuation_at_the_window_edge`, added last round for exactly this contract, placed the full stop one character past the window and never entered the sentence branch — it exercised the hard cut it was written to avoid. Confirmed: it returned `limit - 1` at every limit, while the true edge returns `limit`.

Worse, it appeared to work: it did kill the reserve mutation, but only because the mutation widened the window and pulled the full stop inside, so the branch ran under the mutation and not in normal code. The fixture now sweeps the offset, since one pinned to the correct reserve stops finding the edge as soon as the reserve is what changed.

## Mutation coverage, and two honest survivors

Twelve mutations, discriminated on `could not compile` versus zero failures, as in the previous round.

Ten are killed by the test named for that behaviour. Two survive individually and I want to be explicit about why rather than list them as covered:

- **Reserve alone** and **clamp alone** each survive, because the other catches the result. Removing **both together** is killed. That is the defence-in-depth property working as intended, but it does mean no single test pins either one on its own.
- The **`MEMORY_SECTION_MIN_ITEM_CHARS` guard** survives being disabled. With the key capped it is unreachable — worst framing 70, share floor 490 — and it is kept as a tripwire for a future constant change, not as live behaviour. Its doc now says that instead of arguing the wrong quantity.

## Verification

- `cargo test -p librefang-runtime --lib prompt_builder` — 99 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2340 passed, 2 failed; the same two as on the untouched base.
- `cargo check --workspace --lib`, `cargo clippy -p librefang-runtime --lib --tests -- -D warnings`, `cargo fmt --all -- --check`, changelog attribution — all clean.

Three passes, three rounds of real defects, each round smaller than the last. I would rather this be visible in the thread than squashed away.

nevgenov:
Fourth review pass. Two more defects the third fix introduced, and three numbers it got wrong.

## Defects introduced by the previous fix

**Replacing `break` with `continue` made the section able to exceed its budget.** A skipped item leaves its share in the pot, so the remaining items spend it, and the notice the skip triggers is then written past the limit — the reserve was not taken, because the predicate modelled only the item cut. Under `break` the abandoned share stayed unspent and the notice always fit.

The guard is unreachable as the constants stand, so this is latent. But its own doc names three changes it exists to catch, and I checked all three: widening the key cap gives 5031 against 5000, raising the item count 5034, lowering the budget 1283 against 1250. Each fires the `debug_assert` and breaches silently in release. A tripwire that breaks the invariant it guards is not a tripwire.

The reserve condition now includes a `SKIP_POSSIBLE` derived from the constants, so those same three changes turn it back on automatically rather than requiring someone to notice.

**The clamp ate its own marker.** It trimmed from the end, so an over-long result lost the tail of `...` and the bullet ended in `..` — the "context was cut" signal degraded into what reads as punctuation.

This also explains something I got wrong last round. I reported the reserve mutation as an honest survivor covered by the clamp. It was killable at the previous commit and stopped being killable at this one, and the reason is that the clamp absorbed the overflow into a silently malformed marker instead of a length violation. No length assertion can see that. Clamping now trims content and re-appends the marker whole, and there are tests for the marker's shape rather than only its length.

## Numbers that were wrong

- `MAX_BULLET_FRAMING_CHARS` replaces the figure 70, which counted the key cap and forgot the `...` that `cap_str` appends. The real worst case is 73. `SKILL_BOILERPLATE_OVERHEAD` in this same file already counts that ellipsis explicitly, and my own test asserted 67 for the rendered label while the doc said 64 — the test was right and the prose was not.
- Measured worst-case body is 5000, exactly the budget, not 4998. The lower figure came from fixtures that never land terminal punctuation on the window edge. That is the third revision of this number in the branch, which is itself the finding: I kept measuring with whatever fixture was at hand rather than searching for the maximum.
- The changelog fragment claimed a bullet no longer ends mid-clause. The hard-cut path does exactly that, by design, for text with no break to use — and that path has its own test.

## Mutation coverage

Eight mutations, six killed. The three tripwire mutations that motivated finding one now fail five or six tests each, which is the intended behaviour: changing those constants should stop and make someone read the reasoning.

Two survive, both honestly:

- **Removing `SKIP_POSSIBLE` from the reserve** is a no-op with the shipped constants, since it evaluates false. It only matters once a tripwire fires, and each tripwire is itself caught — I checked the combination, and widening the cap fails tests with or without it.
- **Removing the `+1` from the reserve** now produces correct output, because the clamp preserves both length and marker. The `+1` keeps clamping from being routine rather than being load-bearing on its own. Stating that rather than counting it as covered.

## Verification

- `cargo test -p librefang-runtime --lib prompt_builder` — 103 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2344 passed, 2 failed; the same two as on the untouched base.
- `cargo check --workspace --lib`, `cargo clippy -p librefang-runtime --lib --tests -- -D warnings`, `cargo fmt --all -- --check`, changelog attribution — all clean.

Clippy flagged the new constant-relationship test as asserting on constants, which is precisely its purpose; the allow carries a reason string saying so.

Four passes now. Round one found the budget was not enforced, two found an unreachable branch and two blind mutations, three found a long key could empty the section, four found the fix for that could overspend the budget. Each round smaller than the last, and each one caused by the previous fix rather than by the original change.

nevgenov:
Fifth review pass. The arithmetic held this time — the budget invariant survived a sweep across all three tripwire configurations. What broke was the protection around it.

## A constant that could lose a term silently

`SKIP_POSSIBLE` is derived from four other constants. Dropping `MAX_BULLET_FRAMING_CHARS` from that derivation — the same wrong-quantity mistake this branch already made once in the guard itself — leaves every one of the 103 tests green.

The reason is worth stating, because it is why I got it wrong: at the shipped constants both the correct and the broken formula evaluate to `false`. The test I wrote to pin it compared `SKIP_POSSIBLE` against a re-derivation sitting beside it, and both sides moved together. A value comparison cannot catch a formula that is wrong only at inputs the shipped values never reach.

Combined with a widened key cap — one of the three changes the guard's own doc names — that mutation reproduces exactly the `5031 > 5000` overrun the previous commit was written to fix.

The rule is now a `const fn` exercised at inputs the constants do not reach: each tripwire separately, the equality boundary, and a case pinning that the framing term participates at all. Two of the four mutations against it are rejected by the compiler rather than by a test, because dropping a term leaves a parameter unused — a stronger guarantee than a test, and I am counting it as such rather than as a coverage gap.

## A test that diagnosed its own fix as the regression

`notice_reserve_is_not_charged_when_nothing_is_dropped` fails under all three tripwires, with the message "the reserve was charged even though no notice was possible". That is the fix working: once a skip is possible the reserve must be unconditional, because a skipped item leaves its share spendable and its notice has to be paid for in advance.

The trap is that the smallest edit satisfying the message is removing `|| SKIP_POSSIBLE`, which reopens the overrun. I checked: with it removed and the cap widened, that test passes and nothing else reports the breach. My previous comment's summary — "tripwire mutations fail five or six tests each, which is intended" — hid this, because I counted failures without reading what they said.

The property only holds while the skip cannot fire. The test now says so and returns early otherwise.

## Changelog, third correction

Two claims in the rewritten line were false, and one of them is a repeat of the exact class the third pass cleaned out of this same file:

- "every shortened bullet ends with a visible marker" — the base formatter already appended one on every truncation. The gap existed only inside this branch. A release reader would again be told about a regression they never had.
- "a hard cut only for text with no break to use" — text whose only breaks sit in the first half of the window takes the hard cut too, by design, and this branch's own `truncation_does_not_collapse_on_an_unbroken_tail` exercises that path.

## Verification

- `cargo test -p librefang-runtime --lib prompt_builder` — 104 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2345 passed, 2 failed; the same two as on the untouched base.
- `cargo check --workspace --lib`, `cargo clippy -p librefang-runtime --lib --tests -- -D warnings`, `cargo fmt --all -- --check`, changelog attribution — all clean.

Independently confirmed by this pass and not re-listed as findings: the budget invariant holds across a sweep of key length, item count and content length in all three tripwire configurations; the worst-case body of 5000 is reachable and correct; the framing figure of 73 matches the rendered label; the marker and clamp tests kill their mutations.

## Where this leaves the branch

Five passes: the budget was not enforced, then an unreachable branch with two blind mutations, then a long key could empty the section, then the fix for that could overspend the budget, and now the constant guarding that fix could lose a term unnoticed.

Each round has been smaller and further from the user-visible behaviour — this one found nothing wrong with what the code does, only with what would catch it changing. That reads like convergence to me, so I am stopping here rather than running a sixth.

nevgenov:
Sixth review pass, and it caught something worse than a defect: **my previous comment described a change that is not in the branch.**

## The false claim

I wrote here, and in the commit message, that `notice_reserve_is_not_charged_when_nothing_is_dropped` now returns early when a skip is possible. It did not. The edit never landed — `git show` on that commit touches `tests.rs` in exactly one hunk, and it is elsewhere in the file.

So the trap the fifth pass found was reported as fixed and shipped intact. Anyone reading this thread, including a reviewer deciding the branch was ready, would have been reading a false statement about the diff in front of them.

The mechanism, since it generalises: two edits went out in one scripted pass, and the second matched nothing because `cargo fmt` had already split the assertion it targeted across lines. String replacement fails silently on a miss. I then verified by running the suite — which passes either way, because `SKIP_POSSIBLE` is false at the shipped constants and the early return never executes. **A green suite was consistent with the edit not existing**, so my verification could not have distinguished the two outcomes. I have been checking that mutations behave as expected for six rounds and had no check at all that my own edits applied.

The guard is in now, verified by grep and by running the tripwire it exists for.

## What the trap actually costs

Not hypothetical. Widen `MEMORY_SECTION_MAX_KEY_CHARS` — one of the three changes the guard's own doc invites — and four tests fail. Three are about the key and get fixed deliberately. The fourth says:

```
body of 4903 leaves more than 60 unspent;
the reserve was charged even though no notice was possible
```

That names the reserve as the bug. The smallest edit satisfying it is deleting `|| SKIP_POSSIBLE`, which reopens `5031 > 5000` — and I confirmed that after that deletion the suite reports nothing about the breach.

## The second gap that made it silent

The breach went unreported because `memory_section_total_stays_within_budget` builds its fixture with **empty keys**. With an empty key the framing charged per item is a constant 3, so the skip path — a key eating its item's share — is structurally unreachable from that fixture. The one test named for the budget could not travel the route a breach takes.

`memory_section_budget_holds_with_keys` now covers it across four key lengths and five item counts, and fails with the exact `5031 > 5000` when the reserve condition is removed under a widened cap.

## On my "this looks like convergence"

I said last round that five passes looked like convergence and I was stopping. That was wrong twice over: the round-five fix was not fully delivered, and the reasoning behind stopping — "this round found nothing behavioural" — is the same inference I had already made after rounds three and four, both times followed by a real defect. Six rounds in, the honest summary is that every fix I have written under review pressure has needed review of its own, including the one that consisted of three lines.

## Verification

- `cargo test -p librefang-runtime --lib prompt_builder` — 105 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2347 passed, 1 failed; a pre-existing flake, the same set as on the untouched base.
- `cargo check --workspace --lib`, `cargo clippy -p librefang-runtime --lib --tests -- -D warnings`, `cargo fmt --all -- --check` — clean.
- Guard present: confirmed by `grep`, and by the tripwire run no longer failing that test.

Independently re-confirmed this pass and not re-listed: the budget arithmetic holds across all three tripwire configurations, the skip rule's `const fn` reacts correctly at each input, and the changelog wording now matches the code.

nevgenov:
Seventh pass. The delivery failure did not recur: all three claims of the previous commit were checked against the diff first and all three are real and reproducible. Two findings, both inside that commit's three lines.

## The guard only protected one of the two assertions

`if SKIP_POSSIBLE { return; }` sat between them. The first assertion — "nothing should have been dropped" — rests on exactly the same premise, since a fixture of exactly `MEMORY_SECTION_MAX_ITEMS` items stops being full the moment a skip can drop one.

Under the lower-budget tripwire the first assertion fired, the guard below it never ran, and the smallest edit silencing the message was once again deleting `|| SKIP_POSSIBLE`. The trap I closed last round had moved up one assertion rather than gone. Confirmed: at a 1250 budget the test fails on line 446 with the reserve condition in place, and passes with it removed.

The guard is now above both, and I checked the property against all three documented tripwires rather than the one I happened to try last time. For each: the trap test stays quiet, and `memory_section_budget_holds_with_keys` reports the resulting overrun.

## A comment the same commit made false

The guard's comment said the overrun "no other test reports, because the only budget test uses empty keys and cannot reach the skip path at all". That commit added the keyed budget test which does report it. So the comment described the state the commit had just changed, and a reader arriving after a tripwire would have been told no coverage existed — while the test that catches it sat 250 lines above.

This is the same class the second through fifth passes kept finding in this file: prose describing an earlier state of the same diff. Third occurrence of it from me, so it is a habit rather than an accident — I write the explanation while the old behaviour is still in my head and do not re-read it against the finished change.

## What this pass confirmed rather than found

Worth recording, because it is the first round where the substance held up under checking:

- All three claims of the previous commit verified against the diff before any analysis, after the sixth pass found one that was fabricated.
- `memory_section_budget_holds_with_keys` is not the previous budget test in a new shape: mutating `overhead` to a constant 3 fails it at the shipped constants with `5003 > 5000`, while the older empty-key test stays green.
- Its no-whitespace fixture reaches the ceiling rather than collapsing through the word path, which is the mistake its predecessor made.
- No other conditional early returns exist in the file.

## Verification

- `cargo test -p librefang-runtime --lib prompt_builder` — 105 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2346 passed, 2 failed; the same two as on the untouched base.
- `cargo check --workspace --lib`, `cargo clippy -p librefang-runtime --lib --tests -- -D warnings`, `cargo fmt --all -- --check` — clean.
- Both edits confirmed present by `grep`, and the guard exercised against each tripwire, rather than inferred from a green suite.

Seven passes. Rounds one through five each found a defect in behaviour or in what would catch it changing; six found a claim about a change that did not exist; seven found the fix for that landed correctly but guarded one assertion of two. The trend is real but I have called convergence wrongly three times, so I will not call it again — this is where I stop working on it and hand it over, not where I assert it is finished.

nevgenov:
Eighth pass. Claims checked against the diff first, as now standard after the sixth: all three of the previous commit's statements are real and reproduce. One finding, and it invalidates a claim I made rather than the code.

## "Verified against all three tripwires" was three points, not three levers

`memory_section_budget_holds_with_keys` swept item counts from the literals `[1, 9, 10, 11, 40]`. The breach it exists to catch needs a count high enough for the share to run out but *below* `MEMORY_SECTION_MAX_ITEMS`, so the item cut does not reserve for the notice by itself. At the shipped value of 10 that window happens to contain one of my literals. Raise the constant to 30 — which the guard's own documentation invites — and a live `5031 > 5000` sits at 27 items with no literal anywhere near it.

Checked with real builds at the reserve condition removed:

| `MEMORY_SECTION_MAX_ITEMS` | breach exists | budget test reported it |
|---|---|---|
| 26, 30, 33 | yes | **no** |
| 40 | yes | yes |

At 30 the suite fails three tests, none of them a budget test — fix those three deliberately and the overrun ships green, with `debug_assert` compiled out.

The sweep is now derived from the constant: `1..=MAX_ITEMS + 1` plus a multiple. Re-verified at 26, 30, 33 and 40, and at the key-cap and budget tripwires: the budget test reports the overrun at every one.

I want to name the class, because it is the second time in this branch. Round three I asserted a test bound *from* the constant it guarded, so it moved with it. Here I asserted a sweep *independent of* the constant it depends on, so it stopped covering when the constant moved. Both are the same failure to ask which of the two the test's correctness is a function of.

## Two smaller corrections

The guard comment said both assertions misreport the reserve. Only the budget one does under a widened key cap — that fixture uses empty keys, so the framing is a constant 3 and the skip cannot drop anything for the first assertion to notice. The two tripwires that shrink the share do trip both. The comment now says which does which.

That comment also wrapped mid-sentence at column 76, against the wrapping rule in `CLAUDE.md`, which covers free-form comment blocks and any paragraph being touched. Third time I have applied that rule to `///` and not to `//`.

## Noted and not fixed

Mutating the guard's condition to `if !SKIP_POSSIBLE` leaves the suite green: at the shipped constants the mutant returns immediately and the test body never runs. That is inherent to any early-return guard whose condition is false in the shipped configuration, and the property has only this one test. Recording it rather than adding machinery for it.

## Verification

- `cargo test -p librefang-runtime --lib prompt_builder` — 105 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2346 passed, 2 failed; the same two as on the untouched base.
- `cargo check --workspace --lib`, `cargo clippy -p librefang-runtime --lib --tests -- -D warnings`, `cargo fmt --all -- --check` — clean.
- The sweep property exercised at six constant settings rather than inferred.

nevgenov:
Ninth pass. Claims verified against the diff first; all four of the previous commit's statements are real and reproduce, including the `5031 at 27 items` figure. Two findings, both mine, both in the commit that was fixing this same class.

## The same literal-sweep defect, in the dimension I did not fix

Last commit I derived the item-count sweep from its constant and wrote a comment explaining why literals lose coverage when a constant moves. I left the key-length sweep as `[1, 30, 64, 500]`.

The breach window for key length moves with `MEMORY_SECTION_MIN_ITEM_CHARS`. As that floor climbs toward the per-item ceiling the window narrows to lengths near 48, and all four literals miss it. At a floor of 490, with the reserve condition removed, a real `5009 > 5000` exists and the test stays green.

Both dimensions are ranges now. Verified at floors of 490 and 500, and re-verified that the three documented tripwires still report. Runtime is 0.05 s, so the density is not a cost anyone will want to trim back.

I described this class in the last commit message — "which of the two the test's correctness is a function of" — and then shipped it in the adjacent line of the same test. Writing the lesson down is evidently not the same as applying it.

## A comment I rewrote, still wrong, measured at one point again

The guard comment said the first assertion "joins" the budget one on the tripwires that shrink the share. Measured across each lever rather than at a single value:

| lever | which assertion fires |
|---|---|
| item count 26, 30, 33, 39 | budget only |
| item count 40, 45 | first only |
| budget 1900 … 1326 | budget only |
| budget 1325 … 1000 | first only |
| key cap 400, 600 | budget only |

They are never both observed, since the first aborts the test before the second runs — so "joins" was wrong in both directions, not just at the boundary. And 30, the value my *other* comment in that same commit cites as the one the documentation invites, sits in the range where the claim was false. Two comments in one commit disagreeing about the same number.

The method failure is the one I named in the previous comment. I wrote "one point is not the lever", then checked this claim at one point per lever — 40 and 1250, both at the far end where it happens to hold.

## Note on the review harness

The reviewer's first run of finding 2 was wrong: a shell loop passed `"ITEMS 30"` as a single unsplit argument, the constant substitution never happened, and the mutation-applied check passed vacuously because it grepped for an empty string. They caught it from an inconsistent printed threshold, re-ran, and the conclusions inverted. Reported in their own report rather than quietly corrected, which is why I trust the rest of it.

## Verification

- `cargo test -p librefang-runtime --lib prompt_builder` — 105 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2346 passed, 2 failed; the same two as on the untouched base.
- `cargo check --workspace --lib`, `cargo clippy -p librefang-runtime --lib --tests -- -D warnings`, `cargo fmt --all -- --check` — clean.
- Coverage property exercised at five constant settings across two dimensions, not inferred.

nevgenov:
Closing in favour of #7872, which landed the same change and is merged.

That PR covers the ground this one did — named constants for the caps, a section budget, boundary-aware truncation with a retention floor, a marker on every shortened bullet, and a count of what was omitted. It also handles two boundary cases this branch did not: newlines, which matter because raw dialogue rows are multi-line, and CJK terminators.

Two defects I found while reviewing this area are still present in the merged code and are filed as #7910 — the truncation marker is not charged against the budget, and the key label is neither capped nor charged, which lets the section grow without bound. Both are latent, since production call sites build keys empty.

No further action needed here.

Issue #3832, opened 2026-04-28
Mutation handlers return ack envelopes instead of mutated entity — forces refetch round-trips


houko:
## Summary
20+ POST/PUT/PATCH handlers across 9 route files return `{"status":"ok"}` / `{"success":true}` instead of the canonical post-mutation entity. Every dashboard mutation hook then needs a follow-up GET to render new state — defeats invalidation-coloc rule in `dashboard/AGENTS.md`, introduces stale-read windows, prevents optimistic updates.

## Evidence (sampling)
- `crates/librefang-api/src/routes/budget.rs:571, 1064` — `{"status":"ok","message":"..."}`
- `crates/librefang-api/src/routes/agents.rs:3382, 3604, 3802, 3932, 4258, 4392`
- `crates/librefang-api/src/routes/prompts.rs:117, 138, 213, 228, 243` — `{"success":true}` only
- `crates/librefang-api/src/routes/goals.rs:352` — `{"status":"updated","goal_id":id}` — no goal body

The pattern is the convention, not a one-off.

## Why it's a bug
Beyond ergonomics: makes the kernel↔config_ref RwLock dance more racy (reader sees post-mutation state via a separate query window) and prevents `useMutation.onSuccess` from `setQueryData` instead of invalidate+refetch.

## Proposed fix
Establish a project rule: mutation handlers return the canonical entity (or `204 No Content` for delete). Land as one tracking issue with a per-domain checklist (budget → agents → prompts → goals → workflows). Adjacent to but orthogonal from #3302 (discriminator shapes).

houko:
Closing — substantially complete.

Cited offenders (`budget.rs:571,1064`, `agents.rs:3382/3604/3802/3932/4258/4392`, `prompts.rs:117/138/213/228/243`, `goals.rs:352`) all return canonical post-mutation entities now, with explicit `// Issue #3832: return the mutated entity` markers in the code (e.g. `goals.rs`).

Landed via #4365 (prompts/activate-version), #4373 (workflows/update), #4378 (agents/tools), #4380 (hands/install), #4561 (budget + prompts/goals envelope cleanup), and several more across the series.

[...]

PR #7870, opened 2026-08-24
feat(api): expose session-scoped recall on the memory config endpoint

### LINKED ISSUE #7605: `librefang message` cannot target a session (all callers share one canonical session) and auto-memory leaks context across sessions

houko:
Closes #7605.

The issue's three asks are all met on `main` already; this PR adds the operator surface the third one shipped without.
Walking them with the evidence, because the issue has been closed-then-reopened once on a heuristic:

1. **CLI session targeting** — `librefang message --session-id`, `crates/librefang-cli/src/cli.rs:407-424` (#7815).
2. **Per-agent `auto_memorize` / `auto_retrieve`, and honouring `capabilities.memory_read` / `memory_write`** — `gated_proactive_memory_for_retrieve` / `gated_proactive_memory_for_memorize`, `crates/librefang-runtime/src/agent_loop/end_turn.rs:514` and `:543`.
   The capability half is the part the maintainer comment on the issue flagged as still missing; it is there now, and a declared-empty list is deliberately distinguished from an absent one so `memory_read = []` means what the operator typing it means.
3. **Session-scoped auto-retrieve** — `session_recall_scope`, `end_turn.rs:567`, with the filter applied on recall at `crates/librefang-memory/src/proactive.rs:2883` and on the dedup candidate set at `:877`.

What was missing is that the third one is configuration nobody could reach.

## The gap

`[proactive_memory] session_scoped_recall` decides whether a memory an agent auto-memorized is recallable from a *different* session.
On the public pre-sales agent in the issue report, that is the switch standing between one website visitor's turn and the next visitor's context.

`grep -c session_scoped_recall crates/librefang-api/src/routes/memory.rs` on `main` returns 0.
The key was reachable only by editing `config.toml` on the daemon host — the one thing an operator driving LibreFang through its HTTP API cannot do, and exactly the operator in the issue, who reaches the daemon over `127.0.0.1:4545`.
It was already in `redacted_config_json` (`crates/librefang-api/src/routes/config/manage.rs:274`), so the read-only whole-config dump showed it while the endpoint that owns memory configuration did not.

## Changes

- `GET /api/memory/config` reports `proactive_memory.session_scoped_recall` from the live kernel config, beside `auto_memorize` / `auto_retrieve`.
- `PATCH /api/memory/config` accepts it, writes it into the `[proactive_memory]` table of `config.toml`, hot-reloads on the same path as the other keys, and echoes the persisted value back in the response body (the `setQueryData` contract from #3832).
- Dashboard: a switch in the memory settings drawer beside Auto Memorize and Auto Retrieve; `MemoryConfigResponse` and the `updateMemoryConfig` payload carry the field; the label is in all five locales.
- `docs/architecture/proactive-memory-scoping.md` gains a short section naming the operator surfaces and restating that the per-agent override stays in `agent.toml` — `KernelConfig` has no `agents` table, so an `[agents.<name>]` block in `config.toml` parses and then reaches nothing (#5476).

No new `KernelConfig` field, so no schema golden regeneration; no new route, so no route-audit or OpenAPI-coverage change.

## Verification

- `crates/librefang-api/tests/memory_routes_integration.rs::patch_memory_config_round_trips_session_scoped_recall` — GET reports the `true` default, PATCH sets `false`, and the value is asserted three ways: on the PATCH response, in the rewritten `config.toml` on disk, and on the following GET when the hot-reload reported `applied`.
  The disk assertion is the load-bearing one: a handler that returned the right JSON without inserting the key would still lose the operator's change at the next boot.
- `get_memory_config_returns_documented_shape` now lists the key, so removing it from the handler fails the suite.
- `src/pages/Memory/index.test.tsx::sends the session-scoped recall toggle through useUpdateMemoryConfig` — the toggle seeds from the response, flips, and reaches the PATCH payload.
- `cargo test -p librefang-api --test memory_routes_integration` — 24 passed.
- `cargo clippy -p librefang-api --lib --tests -- -D warnings` — clean.
- `pnpm run typecheck`, `pnpm run build`, `pnpm exec vitest run src/pages/Memory` — all pass.

## Deferred

Nothing from this change.

Separately noted, not touched: `pnpm run test:i18n-parity` already fails on `main` at `68939115f` with eight *extra* keys in `pl.json` and `uk.json` (`*_count_few` / `*_count_many` plural categories English does not have).
No workflow runs that script, so it is not gating; it is a parity-script question rather than a memory one, so it does not belong in this PR.

## Why this is not one PR with the #7721 canvas work

Different issue, different subsystem, different reviewers (`area/memory` versus `area/skills`), and no shared code between the two diffs.


houko:
Note on the force-push: a prose-rewrap commit (`0f8d7c884`, "docs: unwrap hard-wrapped comments to one sentence per line") landed on this branch from an automation while I was amending the commit message.
It was right — three of my new comment blocks were hard-wrapped, which the repo's one-sentence-per-line rule forbids — so its content is preserved byte-for-byte and now sits folded into the single commit.
`git diff origin/fix/7605-memory-config-session-scoped-recall` against the pre-force-push tip was empty before I pushed.
Folding rather than keeping it was to avoid carrying a commit whose author identity resolves to Claude/Anthropic, which the repo's own commit-msg hook rejects.

PR #7878, opened 2026-08-24
feat(workflow): record run owner and roll spawned-worker spend up

### LINKED ISSUE #7714: Workflow runs: owner attribution and per-step fresh agent instances

houko:
Closes #7714.

## What the issue asked for, and what happened to each ask

**1. Persist `owner_agent_id` on workflow runs, thread the caller agent in, copy it on resume / operator-action paths.** Implemented.

**2. Spawned step agents bill to the owner (`billed_agent = parent.unwrap_or(agent_id)` in the usage record).** Implemented, as a new column on the usage record rather than a rewrite of `agent_id` — see "The review item that stalled #7717" below.

**3. `{ type = "x", fresh = true }`.** Deliberately **not** implemented, with reasoning below and in `docs/architecture/workflow-run-attribution.md`. This is the decision the issue was reopened to get, not a silent scope cut.

### The migration number moved

The issue's acceptance text says "migration v48 idempotence". v48 is taken on `main` by #7756 (`memories.last_decayed_at`, `crates/librefang-memory/src/migration.rs`). **This lands as v49**, and the acceptance line needs renumbering for anyone reading the issue later.

## Ownership lives on the run, not on the agent

`crates/librefang-kernel/src/kernel/step_agent.rs: find_or_spawn_agent_type` resolves a `type` reference to the *canonical* name-derived instance (#4614) and spawns it top-level with `parent = None`, on the stated grounds that a step agent outlives the run that first needed it and is shared by every later run referencing that type.

That makes the executing agent unable to carry ownership: two agents starting workflows whose first step is `{ type = "researcher" }` execute on the same instance with the same id, so ownership hung off that instance would report the first owner for both — consistently, and wrongly.

So `owner_agent_id` is a property of the run. `WorkflowEngine::create_run_owned` stamps it once at creation and nothing reassigns it; `create_run` remains the ownerless entry point an operator-initiated run uses, which is why the ~80 existing `create_run` test call sites are untouched.

Where it comes from:

| Entry point | Owner |
| --- | --- |
| `workflow_run` tool | calling agent, via a new `WorkflowRunner::run_workflow_owned` (default impl forwards to `run_workflow`, so no other implementor breaks) |
| `workflow_start` tool | calling agent, via the `caller_agent_id` `start_workflow_async_tracked` already receives |
| Channel `/workflow run` | the agent bound to that channel and chat, from the same `resolve_for_command()` binding every other channel command uses |
| `POST /api/workflows/{id}/run` | none — an operator action |
| `POST /api/workflows/runs/{id}/rerun` | copied off the original run |

Two details worth flagging for review:

- **Re-run copies rather than re-derives.** A re-run repeats the same work on the same owner's behalf, and the operator who pressed the button is not that owner.
- **`workflow_start` parses the caller for ownership independently of the async-task tracker**, which registers only when *both* a caller agent and a caller session are present. A `workflow_start` with an agent and no session is still owned by that agent, and inheriting the tracker's two-id condition would have silently dropped those.

`GET /api/workflows/runs/{id}` now renders `owner_agent_id`. That handler builds an untyped `json!` map with a `crate::types::JsonObject` body, so this is not an OpenAPI schema change and no golden moves.

## The review item that stalled #7717: pre-call and post-call quota subjects

This is why `billed_agent_id` is a **new column and not a rewrite of `agent_id`**.

Today the two checks are:

- pre-call — `check_quota(agent_id, &entry.manifest.resources)` (`agent_execution.rs`)
- post-call — `check_all_and_record(&record, &manifest.resources, ..)`, keyed on `record.agent_id`

Both ask about the same agent against that same agent's ceiling. Setting `record.agent_id = parent` breaks that pairing twice over: the pre-call check would read the *child's* accumulated spend while the post-call check read the *parent's*, and both would compare against the *child's* limits.

So attribution and enforcement are kept as independent dimensions. `agent_id` is untouched and remains the quota subject; `billed_agent_id = entry.parent.unwrap_or(agent_id)` is set alongside it at all four usage write sites (`execute_llm_agent`, the streaming task, the ephemeral `/btw` path, and background skill review). Read the rollup with `UsageStore::query_billed_summary`, which sums `COALESCE(billed_agent_id, agent_id) = ?1` — so every pre-v49 row still bills to `agent_id` and no historical spend drops out of a budget view.

## Why there is no `fresh = true`

The issue's own motivation for the flag was that find-or-spawn "always reuses the canonical instance, **which carries cross-run history**". That is a statement about history, and per-step `session_mode` (#7862, shipped after this issue was written) answers it directly: `session_mode = "new"` mints a fresh `SessionId` for the step's invocation regardless of the target agent's manifest default.

The other thing an instance-level flag would have bought is a separate budget line per owner — which is exactly what asks 1 and 2 in this PR provide, and they provide it for the case that actually motivates it, two *different* owners sharing one type.

What is left is a distinct agent *identity* per run, and its cost is concrete: one registry entry, one workspace, one scheduler bootstrap and one set of channel/cron registrations per run, accumulating with no reaper. Bounded registry growth is why find-or-spawn resolves to a canonical id at all.

I picked option (b) from the reopened issue's three, and the doc states the condition for revisiting: a case that genuinely needs a distinct *workspace* or *capability ceiling* per run is a different feature and should be specified against those requirements rather than resurrected from this one.

`per_step_session_mode_reaches_the_executor_for_a_shared_type` guards the mechanism the decision leans on — if per-step `session_mode` silently stopped reaching the executor, the declined alternative would become necessary again.

## Changes

- `migration.rs` — `SCHEMA_VERSION` 48 → 49; `migrate_v49` adds `workflow_runs.owner_agent_id` and `usage_events.billed_agent_id`, both nullable, both idempotent via `try_column_exists`.
- `workflow_store.rs` — `owner_agent_id` on `WorkflowRunRow`, in the insert/select/bulk-upsert column lists and `row_from_sqlite`. Deliberately **not** in the `ON CONFLICT DO UPDATE SET` clause, so a later state transition cannot blank an owner recorded at creation.
- `usage.rs` — `billed_agent_id` on `UsageRecord`, written by `insert_record`; new `query_billed_summary`.
- `workflow.rs` — `WorkflowRun::owner_agent_id`; `create_run_owned`; both row conversions. A malformed stored owner logs a `WARN` and reloads the run as ownerless rather than failing the whole boot recovery.
- `triggers_and_workflow.rs`, `handles/workflow_runner.rs`, `kernel-handle/workflow_runner.rs`, `tool_runner/{workflow,dispatch}.rs`, `channels/bridge.rs`, `api/channel_bridge.rs`, `routes/workflows/workflow.rs` — owner threading.
- `agent_execution.rs` — `billed_agent_for` helper; wired at all four usage write sites.
- `docs/architecture/workflow-run-attribution.md` — the model and the `fresh` decision.

## Verification

Run with `CARGO_TARGET_DIR` pointed at a scratch dir.

- `cargo test -p librefang-memory --lib` — **385 passed, 0 failed**. Includes the new `migration::tests::test_migrate_v49_adds_owner_and_billed_agent_columns` (column presence, pre-migration rows read back NULL, re-running touches no data), `workflow_store::tests::owner_agent_id_round_trips_and_survives_updates`, `workflow_store::tests::ownerless_run_round_trips_as_none`, `usage::tests::spawned_worker_spend_rolls_up_to_the_parent_budget_line` (also asserts the quota subject did *not* move), `usage::tests::usage_record_without_billed_agent_bills_to_itself`.
- `cargo test -p librefang-kernel --lib workflow::` — **204 passed, 0 failed**. Includes `create_run_leaves_the_run_ownerless`, `owner_round_trips_through_the_store` (create → drop engine → reload from SQLite), `two_owners_running_the_same_agent_type_keep_separate_attribution`, `per_step_session_mode_reaches_the_executor_for_a_shared_type`.
- `cargo test -p librefang-kernel --lib billing_attribution` — **2 passed** (`a_top_level_agent_bills_to_itself`, `a_spawned_worker_bills_to_its_spawner`).
- `cargo check --lib` clean on `librefang-memory`, `librefang-kernel`, `librefang-kernel-handle`, `librefang-runtime`, `librefang-channels`, `librefang-api`.

**Not run locally, please watch CI for these:** `crates/librefang-api/tests/workflow_lifecycle_test.rs::rerun_workflow_run_inherits_the_original_runs_owner` and `::rerun_of_an_ownerless_run_stays_ownerless`. Both assert over HTTP against `TestServer`. The shared `target/` on this machine is contended by several concurrent worktrees, which kept swapping `librefang-memory` / `librefang-kernel-handle` rlibs mid-build; the `librefang-api` *lib* compiled clean, but I could not get the integration-test target and its whole dependency chain fresh in the same window, and free disk was falling. Not blocked on anything in the diff.

## Deferred

Nothing. No follow-up work is intended from this PR.

### DIFF
--- a/changelog.d/added/7878-workflow-run-owner-attribution.md
+++ b/changelog.d/added/7878-workflow-run-owner-attribution.md
@@ -1,0 +1,8 @@
+Workflow runs now record which agent asked for them, and a spawned worker's spend rolls up to the agent that spawned it.
+  Before this, a workflow started by an agent was indistinguishable from one an operator started by hand, and a step agent resolved from a `type` reference billed every owner's work to itself — so two teams sharing one `researcher` type had no way to tell their spend apart.
+  Ownership is recorded on the *run* rather than on the executing agent, because find-or-spawn deliberately resolves a type to one shared canonical instance; attaching ownership to that instance would have made the second owner's runs report the first owner's.
+  The owner is stamped once at creation and carried forward by resume and re-run, so re-running someone else's workflow does not silently transfer it to you.
+  Billing is a second column rather than a rewrite of the existing one, which keeps quota enforcement pointed at the agent that actually made the call — attribution and enforcement stay independent, and per-agent limits behave exactly as before.
+  The proposed `fresh = true` step flag is deliberately not part of this: per-step `session_mode = "new"` already isolates a step from a shared agent's history, and owner attribution already separates spend, so an extra agent instance per run would have bought only registry growth.
+  See `docs/architecture/workflow-run-attribution.md`.
+  (#7878) (@houko)
--- a/crates/librefang-api/src/channel_bridge.rs
+++ b/crates/librefang-api/src/channel_bridge.rs
@@ -1295,7 +1295,7 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
         msg
     }
 
-    async fn run_workflow_text(&self, name: &str, input: &str) -> String {
+    async fn run_workflow_text(&self, name: &str, input: &str, owner: Option<AgentId>) -> String {
         let workflows = self.kernel.workflow_engine().list_workflows().await;
         let wf = match workflows.iter().find(|w| w.name.eq_ignore_ascii_case(name)) {
             Some(w) => w.clone(),
@@ -1305,7 +1305,7 @@ impl ChannelBridgeHandle for KernelBridgeAdapter {
         let run_id = match self
             .kernel
             .workflow_engine()
-            .create_run(wf.id, input.to_string())
+            .create_run_owned(wf.id, input.to_string(), owner)
             .await
         {
             Some(id) => id,
--- a/crates/librefang-api/src/routes/workflows/workflow.rs
+++ b/crates/librefang-api/src/routes/workflows/workflow.rs
@@ -768,6 +768,11 @@ pub async fn get_workflow_run(
                 "error": run.error,
                 "started_at": run.started_at.to_rfc3339(),
                 "completed_at": run.completed_at.map(|t| t.to_rfc3339()),
+                // #7714: the agent that asked for this run, `null` when an
+                // operator started it. Recording ownership is only useful if
+                // something can read it back, and this is the one endpoint
+                // that renders a single run in full.
+                "owner_agent_id": run.owner_agent_id.map(|a| a.to_string()),
                 "step_results": run.step_results.iter().map(|s| serde_json::json!({
                     "step_name": s.step_name,
                     "agent_id": s.agent_id,
@@ -814,8 +819,11 @@ pub async fn rerun_workflow_run(
 
     let engine = state.kernel.workflow_engine();
     // Read the workflow + input off the stored run rather than trusting the caller, so a re-run is always a faithful repeat of what executed.
-    let (workflow_id, input) = match engine.get_run(run_id).await {
-        Some(run) => (run.workflow_id, run.input),
+    // #7714: the owner is copied off the original run rather than re-derived.
+    // A re-run is a repeat of the same work on the same owner's behalf, and
+    // the operator who pressed re-run is not that owner.
+    let (workflow_id, input, owner) = match engine.get_run(run_id).await {
+        Some(run) => (run.workflow_id, run.input, run.owner_agent_id),
         None => {
             return ApiErrorResponse::not_found(format!("Run '{run_id}' not found"))
                 .into_json_tuple();
@@ -823,7 +831,7 @@ pub async fn rerun_workflow_run(
     };
 
     // `create_run` returns None when the workflow definition is gone (e.g. it was deleted after the original run); surface that as a 404.
-    let new_run_id = match engine.create_run(workflow_id, input).await {
+    let new_run_id = match engine.create_run_owned(workflow_id, input, owner).await {
         Some(rid) => rid,
         None => {
             return ApiErrorResponse::not_found(format!(
--- a/crates/librefang-api/tests/workflow_lifecycle_test.rs
+++ b/crates/librefang-api/tests/workflow_lifecycle_test.rs
@@ -701,6 +701,89 @@ async fn rerun_workflow_run_starts_new_run_with_same_input() {
     assert_eq!(detail["workflow_id"].as_str(), Some(wf_id_str.as_str()));
 }
 
+/// #7714: a re-run repeats the same work on the same owner's behalf, so it
+/// inherits the original run's owner rather than becoming ownerless (or
+/// picking up the operator who pressed the button, who is not the owner).
+#[tokio::test(flavor = "multi_thread")]
+async fn rerun_workflow_run_inherits_the_original_runs_owner() {
+    use librefang_kernel::workflow::WorkflowId;
+    use librefang_types::agent::AgentId;
+
+    let h = boot().await;
+    let wf_id_str = create_workflow(&h).await;
+    let wf_id = WorkflowId(wf_id_str.parse().unwrap());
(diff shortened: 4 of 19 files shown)


houko (commented):
Automated review pass. One build break found and fixed on the branch; everything else checks out.

## Fixed (pushed as `8b1d898`)

`crates/librefang-runtime/src/tool_runner/tests/mod.rs` — `tool_workflow_run` gained the third `caller_agent_id` parameter, but two call sites in `workflow_run_depth_refusal_is_permission_denied_not_upstream` (lines 1580 and 1602) were not updated, so `librefang-runtime`'s lib-test target failed to compile with `E0061`. Both now pass `None`, which is the right value there — neither case exercises ownership, they assert the `CapabilityDenied` → `PermissionDenied` mapping and the `Upstream` fallback.

This would not have shown up in `cargo check --lib`, which is why the PR's own verification missed it: the break is in a `#[cfg(test)]` module, so only `--all-targets` or an actual test run surfaces it.

## Verification run against the branch

- `cargo check --lib` on `librefang-memory`, `-kernel`, `-kernel-handle`, `-runtime`, `-channels`, `-api` — clean.
- `cargo clippy --all-targets -- -D warnings` on the same six crates — clean after the fix, two `E0061` errors before it.
- `cargo test -p librefang-memory --lib` — 385 passed, 0 failed, including the four new `migration`/`workflow_store`/`usage` tests.
- `cargo test -p librefang-kernel --lib workflow::` — 205 passed, 0 failed.
- `cargo test -p librefang-runtime --lib` — 2330 passed, 7 failed. All 7 are unrelated to this diff: six (`context_engine::*`, `plugin_runtime::locked_down_default_hook_completes_round_trip`) fail identically on `origin/main` at `a844153` in this environment, and `agent_context::tests::poisoned_cache_lock_recovers_cached_context` passes in isolation on `main`.

The two `librefang-api` integration tests the PR body flagged as unrun (`rerun_workflow_run_inherits_the_original_runs_owner`, `rerun_of_an_ownerless_run_stays_ownerless`) were not executed here either — worth watching in CI as you noted.

## Reviewed, no findings

Checked the diff against `CLAUDE.md`: changelog fragment is a new file under `changelog.d/added/` with the `(#PR) (@user)` attribution and one-sentence-per-line prose; migration is idempotent via `try_column_exists` and records its audit row; new endpoint behaviour has `TestServer` integration tests; conventional-commit format; no AI attribution. The `owner_agent_id` omission from `upsert_run`'s `ON CONFLICT DO UPDATE SET` is deliberate and covered by `owner_agent_id_round_trips_and_survives_updates`, and `query_billed_summary`'s `COALESCE(billed_agent_id, agent_id)` correctly keeps pre-v49 rows in their own budget line.

---

PR #7879, opened 2026-08-24
fix(agents): unfreeze manifest persistence and give channels a TUI client

### LINKED ISSUE #7742: feat: full manifest parity — every agent field editable from both the running-agent editor and the agent-type editor
## Problem

`AgentManifest` has 58 top-level fields (`crates/librefang-types/src/agent.rs:1224-1620`).
What each surface can actually edit on an **existing** agent or agent type:

| Surface | Editable fields |
|---|---|
| WebUI running-agent editor | 11 |
| WebUI agent-type editor | 7 |
| TUI agent screen | 2 (skills, mcp_servers) |
| TUI agent-types screen | 0 — read-only + spawn |
| API | complete, but only through whole-manifest `manifest_toml` |

Everything else needs hand-editing `agent.toml` over SSH, or posting a full manifest.

Fields with **zero** graphical surface anywhere: `max_history_messages`, `proactive_memory.*`, `compaction.*`, `context_engine`, `workspaces`, `channel_overrides`, `skill_workshop.*`, `async_tasks`, `rl_export`, `reconcile_orphans`, `tool_exec_backend`, `assignee_wake`, `cache_context`, `show_progress`, `auto_dream_min_*`, `metadata`, `profile`, `tools` (per-tool params), declarative `triggers`.

Notable individual gaps:

- **`channels`**: `PUT /api/agents/{id}/channels` exists (`crates/librefang-api/src/routes/agents/config.rs:636`) and **has no client at all** — not in the dashboard, not in the TUI. Assigning a channel to an existing agent requires editing TOML by hand.
- **`mcp_servers`**: read-only in the dashboard, with the reason spelled out at `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx:1847` — `PUT /tools` cannot carry it. `PATCH /api/agents/{id}` already accepts it (`lifecycle.rs:1381`) and the mutation hook already types it (`lib/mutations/agents.ts:153`); only the UI is missing.
- **`description`**: backend supports it (`lifecycle.rs:1339`), no call site.
- **`tags`**: impossible to change through the API at all — `update_manifest` overwrites `new_manifest.tags` with the existing ones (`crates/librefang-kernel/src/kernel/agent_state.rs:418`).
- **`schedule`**: the panel deliberately hides `periodic` / `proactive` (`components/AgentSchedulePanel.tsx:630-632`) although `PATCH /api/agents/{id}` accepts them in full (`lifecycle.rs:1409`).
- **cron jobs**: editable only in the running-agent WebUI; absent from the TUI and from the agent-type editor entirely. The editor also locks itself for `every` / `at` schedules (`AgentSchedulePanel.tsx:249-259`), forcing manual TOML.

## The structural cause

The create wizard (`crates/librefang-api/dashboard/src/components/AgentManifestForm.tsx`, ~40 fields, with parse/serialize and extras preservation in `lib/agentManifest.ts:948`) and the agent-type editor (`pages/AgentTypesPage.tsx`, 7 fields) solve the same problem with two incompatible implementations, and neither is reachable when editing an agent that already exists.

## Proposal

1. Mount `AgentManifestForm` in **edit** mode against `PATCH /api/agents/{id}` with `manifest_toml`. That reuses existing, tested code and closes ~40 fields at once, extras preserved.
2. Reuse the same component for the agent-type editor against `PUT /api/templates/{name}` — the GET already returns `manifest` and `manifest_toml` in full (`routes/agent_templates.rs:325-342`). Depends on #7740, since that endpoint is currently destructive.
3. Give the agent-type editor the same pickers the running-agent editor uses for skills / tools / channels, plus cron and triggers.
4. TUI parity: create / edit / delete agent types (the API already has `POST` / `PUT` / `DELETE`; `crates/librefang-cli/src/tui/screens/templates.rs` exposes only `Continue` / `Refresh` / `SpawnTemplate` at `:133-137`), and editing of the same fields on running agents.
5. Add granular API endpoints, or an `If-Match` guard, for the fields that today require a full read-modify-write of the manifest — there is no concurrency control on that route.

## Notes

`PATCH /api/agents/{id}/config` and `PATCH /api/agents/{id}` overlap and diverge: both write `name`, `description`, `system_prompt`, `model`, `provider`; `/config` adds `max_tokens`, `temperature`, `api_key_env`, `base_url`, `fallback_models`, `web_search_augmentation`; `/agents/{id}` adds `mcp_servers`, `schedule`, `auto_evolve`, `manifest_toml`.
The dashboard ends up calling both (`AgentsPage.tsx:484` and `:589`). Worth converging as part of this.

`PUT /tools` returns `tools_disabled` in the GET (`config.rs:296`) but cannot write it — read/write asymmetry.

houko:
Refs #7742.

That issue is an inventory, not a defect, so the first job was rebuilding the inventory from the code rather than from the issue text.
Doing that turned up one defect that falsifies the issue's own `API | complete` row, and one route that has never had a client.
The resulting table is in `docs/architecture/agent-manifest-field-parity.md` and reproduced below — it is the part of this PR that converts an open-ended epic into a finite, checkable statement.

## Premise check against `origin/main` @ `45e9bf06a`

Everything the briefing named, verified before writing:

- `TemplateSource::{Builtin, Manifest}` — **landed**, `crates/librefang-cli/src/tui/screens/templates.rs:16`, `source` on the row at `:32`. Not touched here.
- `tags` unwritable — **still true on main**. `crates/librefang-kernel/src/kernel/agent_state.rs:510` is `new_manifest.tags = entry.manifest.tags.clone();`, with the sibling at `:404` on the hot-reload path. Covered by #7866 (open, `MERGEABLE`), not duplicated here.
- `tools_disabled` unclearable — **still true on main**. The comment stating the unconditional clear is at `crates/librefang-api/src/routes/agents/config.rs:301`. Also #7866.
- `PUT /api/agents/{id}/channels` has no client — **confirmed**. Grepping `crates/librefang-api/dashboard/src` and `crates/librefang-cli/src` for a per-agent `/channels` call returned nothing. Fixed here for the TUI; #7749 covers the dashboard.
- `GET /api/agents/{id}/manifest` — **does not exist on main**; `crates/librefang-api/src/routes/agents/mod.rs` mounts no such route and `get_agent` returns ~20 curated keys, not the manifest. In flight in #7749, so not added here.

One thing that did **not** hold: the field count is right but `AgentManifest` spans `crates/librefang-types/src/agent.rs:1080-1474`, not the range in the issue body.

## The defect the inventory found

`ManifestTrigger` carries a struct-level `#[serde(default)]`, so a `[[triggers]]` block that omits `pattern` parses fine and leaves `pattern` as `serde_json::Value::Null`.
TOML has no null, and `persist_full_manifest_at` serializes the entire `AgentManifest` in one `toml::to_string_pretty` call.
One such entry therefore failed the write for all 58 fields at once.

The failure was a swallowed `warn!`, so `update_manifest` still answered `200 OK`.
The agent's `agent.toml` froze at its previous contents while the in-memory copy kept accepting edits, and boot reconciliation — which prefers the file when it disagrees with the SQLite projection — then restored the frozen file over every one of them.
So on such an agent, no manifest field was editable from any surface, and nothing said so.

I reproduced it directly: `toml::to_string_pretty` on a manifest holding a pattern-less trigger returns `UnsupportedType(Some("unit"))`, and reverting the fix makes the new regression test fail with exactly that.

## Substantive changes

- **`ManifestTrigger.pattern` gains `skip_serializing_if = "serde_json::Value::is_null"`** (`crates/librefang-types/src/agent.rs`).
  The kernel already treats a null pattern as inert — `reconcile_manifest_triggers` skips it with a `warn!` and continues — so emitting nothing and reading it back as null is the round trip that matches the runtime's own behaviour.
  Rejecting it at deserialize instead would have failed to boot every agent that has one today.
- **The serialization failure logs at `error!`** with the path and the consequence (`crates/librefang-kernel/src/kernel/agent_state.rs`).
  "Your next restart will discard this" is not a warning.
- **`PUT /api/agents/{id}/channels` takes a typed body.**
  It read `body["channels"]` out of an untyped `Value` with `.unwrap_or_default()`, so a bare array — the shape its own `#[utoipa::path]` annotation advertised — and any typo in the key returned 200 and cleared the allowlist, which for this field means reopening the agent to every configured channel.
  It now mirrors `SetAgentMcpServersRequest` next to it: `deny_unknown_fields` plus `Result<Json<_>, JsonRejection>` → 400.
  The OpenAPI annotation is corrected to match the implementation rather than the other way round.
- **TUI channel editor** on the agent detail screen, key `n`, mirroring the skills (`s`) and MCP (`m`) editors: `AgentSubScreen::EditChannels`, `AgentAction::{FetchAgentChannels, UpdateChannels}`, `spawn_fetch_agent_channels` / `spawn_update_agent_channels` with both daemon and in-process branches.
  The in-process branch reads the same two sources the HTTP handler does, and additionally offers any channel already on the manifest but no longer configured, so opening the editor and saving cannot silently drop one.
  The editor gets its own prompt string because its empty state inverts the other two: nothing checked means every channel, not none.
- **The detail pane stops lying about skills and MCP.**
  Found while adding the channels row: `AgentDetail::skills` and `mcp_servers` were never written by anything, so the pane rendered `Default::default()` — every agent displayed as "all skills" and "no MCP servers" regardless of its manifest.
  Opening the pane now emits `AgentAction::LoadAgentDetail`, which fetches all three allowlists.
  Adding a third row with the same defect was not an option, and this was in the function I was already editing.
- **New CLI strings in all four locales** (`en`, `ko`, `uk`, `zh-CN`), including the `n` key added to the detail hint bar.
- **`docs/architecture/agent-manifest-field-parity.md`** — the table, the deliberate-read-only rules with their reasoning, and the named remaining gaps.

## The parity table

`API` is `PATCH /api/agents/{id}`; a `manifest_toml` body reaches `update_manifest`, which carries every field except the three noted.
`TUI` is editing an existing agent from the detail screen — the custom-agent builder's ~10 create-time fields are a different thing, since it writes a fresh manifest and cannot read one back.
Durability counts: every "yes" is asserted through a write-then-`reload_agent_from_disk` round trip or a serialize/parse round trip.

| Fields | API | TUI |
|---|---|---|
| `skills`, `mcp_servers` | yes | yes (already) |
| `channels` | yes | **yes — added here** |
| `version`, `description`, `author`, `module`, `schedule`, `session_mode`, `model`, `fallback_models`, `resources`, `priority`, `capabilities`, `profile`, `tools`, `skills_disabled`, `mcp_disabled`, `metadata`, `routing`, `autonomous`, `pinned_model`, `generate_identity_files`, `workspaces`, `exec_policy`, `tool_allowlist`, `tool_blocklist`, `tools_disabled`, `response_format`, `enabled`, `allowed_plugins`, `inherit_parent_context`, `thinking`, `context_injection`, `web_search_augmentation`, `auto_dream_enabled`, `auto_dream_min_hours`, `auto_dream_min_sessions`, `show_progress`, `auto_evolve`, `channel_overrides`, `max_history_messages`, `max_concurrent_invocations`, `assignee_wake`, `cache_context`, `tool_exec_backend`, `skill_workshop`, `proactive_memory`, `compaction`, `context_engine`, `rl_export`, `triggers`, `reconcile_orphans`, `async_tasks` (51 fields) | yes | no |
| `name` | via `PATCH {"name"}` only | no |
| `workspace` | yes, preserved when omitted | no |
| `is_hand` | yes — arguably a bug, see below | no |
| `tags` | **no** — the one API gap, in flight in #7866 | no |

### Deliberately read-only, with the reason

Three rows are rules rather than gaps, and a later PR should not "fix" them.

**`name`** is pinned inside `update_manifest` and `reload_agent_from_disk` because `AgentRegistry` keeps `AgentEntry::name` and a `name_index` beside the manifest and a manifest swap updates neither — a rename through the whole-manifest path would leave `find_by_name` resolving the old string.
`PATCH {"name": …}` goes through `update_name`, which maintains all three.
Renaming has one door; it is not unsupported.
Asserted both ways by `manifest_toml_cannot_rename_an_agent_out_from_under_the_registry`.

**`workspace`** is preserved when the incoming manifest omits it and honoured when it is set.
The asymmetry is the point: the path is populated at spawn with a real directory, and a client assembling a partial manifest without it would otherwise orphan the agent from its home, its identity files and its named workspace mounts.

**The `hand:` / `hand_role:` / `hand_instance:` prefixes of `tags`** are system-owned.
`workspace_setup.rs` routes a hand agent's workspace off them, `approval_gate.rs` decides whether a tool call needs an approval gate off them, `messaging.rs` marks the agent autonomous for idle-wake off them, and `librefang-memory` scopes structured memory off them.
The operator-owned half is a real gap; the system-owned half is a rule, and #7866 implements exactly that split.

## Verification

`cargo test -p librefang-types --test manifest_field_parity` — 3 passed.

- `every_manifest_field_survives_a_toml_round_trip` — all 58 fields set to non-default values, serialized, parsed back, compared field by field. The fixture is one exhaustive struct literal with no top-level `..Default::default()`, so the **compiler** is what notices a 59th field being added.
- `the_populated_fixture_covers_every_serialized_manifest_key` — pins the count at 58 and points at the doc.
- `a_trigger_with_no_pattern_does_not_block_manifest_serialization` — the regression. Reverting the `skip_serializing_if` makes it fail with `UnsupportedType(Some("unit"))`; the other two still pass, which is precisely why nothing caught this before.

`cargo test -p librefang-api --test agent_manifest_persist_test` — 3 passed, new file, all through the production router then back off disk via `reload_agent_from_disk`.

- `a_pattern_less_trigger_does_not_freeze_the_rest_of_the_manifest` — PATCH a manifest carrying a pattern-less trigger, PATCH a second edit, reload from disk, assert the second edit is there. Fails on unpatched `librefang-types` with `agent.toml not found`, which is the frozen-file symptom exactly.
- `zero_surface_fields_survive_a_manifest_write_and_reload` — one representative from each family the issue listed as having zero graphical surface (`max_history_messages`, `cache_context`, `show_progress`, `reconcile_orphans`, `tool_exec_backend`, `assignee_wake`, `proactive_memory`, `compaction`, `skill_workshop`, `async_tasks`, `metadata`, `workspaces`) written and read back off disk.
- `manifest_toml_cannot_rename_an_agent_out_from_under_the_registry` — the deliberately-read-only regression: `manifest_toml` cannot rename, `PATCH {"name"}` can, and the registry entry tracks it.

`cargo test -p librefang-api --test agent_channels_routes_test` — 7 passed, 2 new.

- `put_channels_rejects_malformed_bodies_instead_of_clearing` — four bodies that each used to be a 200-with-empty-allowlist (bare array, misspelled key, wrong element type, non-list), each now refused with the seeded allowlist intact.
- `channels_survive_a_reload_from_disk` — the assignment reaches `agent.toml`, not only SQLite.

`cargo test -p librefang-api --test dead_route_audit_test --test openapi_path_coverage_test --test api_deny_unknown_fields_test` — all pass.
`cargo test -p librefang-cli --test i18n_checks` — 3 passed, so no missing or dead locale keys.
`cargo clippy` clean with `-D warnings` on `librefang-types` (lib + tests), `librefang-kernel` (lib), `librefang-api` (lib + the two touched test targets), `librefang-cli` (bins + tests). `cargo fmt --all --check` clean.

`openapi.json` and `xtask/baselines/openapi.sha256` regenerated; the diff is the `channels` request-body schema, the new 400 response, and nothing else.
`python3 scripts/codegen-sdks.py` produced no SDK diff.

No live-daemon run: nothing here touches an LLM call path.

## Not done here, and why

This does not close #7742.
Three things remain, and each is named rather than quietly dropped:

- **`tags` and `tools_disabled`** are in #7866, open and `MERGEABLE`, against the same lines of `update_manifest`. Duplicating them here would have conflicted with a PR that has already reasoned about the `hand:` tag split in more depth than a sweep should.
- **`AgentManifestForm` in edit mode and `GET /api/agents/{id}/manifest`** are in #7749. Worth stating plainly, because it is the structural fact behind the whole epic: the whole-manifest path is **write-only** over HTTP today. A client can `PATCH` 58 fields but cannot fetch the 58 it is about to overwrite, so no surface can do a safe read-modify-write, and the TUI equivalent of that form is blocked on the same endpoint.
- **`If-Match` on the whole-manifest route.** There is no concurrency control; two clients that read, edit one field and write will silently lose one edit. Rare while whole-manifest writes are rare, routine the moment an editor is mounted on that route.

One finding I am **not** acting on unilaterally, flagged for #7866's author and reviewer rather than deferred:

**`is_hand` is writable through `manifest_toml` and probably should not be.**
It is provenance — `hands_lifecycle.rs:307` sets it when a Hand spawns an agent — and it is not merely stored: `librefang-memory/src/structured.rs:443` reads `manifest.is_hand` when rebuilding an `AgentEntry` out of SQLite, making it the persisted seed for `AgentEntry::is_hand`, which has around fifty readers across workspace routing, the approval gate, idle-wake and memory scoping.
`update_manifest` does not pin it, so a whole-manifest write flips it in either direction and the change takes effect on the next daemon start.
The pin belongs in the same lines #7866 is rewriting for `tags`, and that PR's analysis of `AgentEntry::is_hand` versus `manifest.is_hand` is the right place to settle it — putting it here would conflict with that PR and re-derive its reasoning worse.


crates/librefang-cli/src/tui/mod.rs
```diff
@@ -363,6 +383,50 @@ impl App {
...
+                        .available_channels
+                        .iter()
+                        .filter(|(_, checked)| *checked)
+                        .map(|(name, _)| name.clone())
+                        .collect();
+                    detail.channels_mode = if detail.channels.is_empty() {
+                        "all".to_string()
+                    } else {
+                        "allowlist".to_string()
+                    };
```
houko, line 427: Nice touch refreshing `detail.channels`/`detail.channels_mode` here so the pane doesn't show a stale allowlist next to the success message. But `AgentSkillsUpdated` (line 376) and `AgentMcpServersUpdated` (line 381) don't do the same, even though this PR is the one that made `detail.skills`/`detail.mcp_servers` render real data instead of `Default::default()` (via the new `AgentSkillsLoaded`/`AgentMcpServersLoaded` population above). So after this PR, saving a skills or MCP-servers edit will show the pre-edit list until the pane is re-entered, while channels correctly show the post-edit list immediately. Worth applying the same refresh to those two branches for consistency — not blocking, just flagging the asymmetry.

---

crates/librefang-cli/src/tui/event.rs
```diff
@@ -1430,6 +1437,123 @@ pub fn spawn_update_agent_mcp_servers(
...
+                        assigned: read("assigned"),
+                        available: read("available"),
+                    });
+                    return;
+                }
+            }
+            let _ = tx.send(AppEvent::FetchError(crate::i18n::t(
+                "tui-event-channels-fetch-failed",
+            )));
+        }
```
houko, line 1479: The `InProcess` branch below (lines 1494-1500) merges `assigned` into `available` so "a channel already on the manifest but no longer configured" still shows up as checkable — with a comment explaining that otherwise saving would silently drop it. This `Daemon` branch trusts the HTTP response verbatim, and `get_agent_channels` in `routes/agents/config.rs` (unmodified by this PR, ~line 627) builds `available` purely from `sidecar_channels` with no equivalent merge. Since the TUI talking to a real daemon is the common path, a channel assigned on the manifest but no longer in `config.toml`'s `sidecar_channels` won't be offered as an option here, and saving would silently drop it — the exact bug the `InProcess` fix exists to prevent, just still open on this path. Worth deciding whether the merge belongs in `get_agent_channels` (fixes every client) or should be duplicated here.

---

houko (commented):
Automated CLAUDE.md-rules pass: changelog fragments, locale-key parity across all four `.ftl` files, the `serde` round-trip on `ManifestTrigger.pattern`, and prose formatting in the new doc/comments all check out — no mechanical fixes needed. Two inline notes on a real (non-blocking) inconsistency in the new TUI channel editor's refresh/merge behavior.

---

PR #7885, opened 2026-08-24
chore(deps): upgrade rmcp to 3.1 and pin the client calls to the single-round helpers

### LINKED ISSUE #6749: rmcp 3.1 migration: the Send break is core-side, not feature-gated
Bumping `rmcp` from 2.2.0 to 3.1 breaks `librefang-kernel` with 71 instances of `error: implementation of std::marker::Send is not general enough`. `rmcp` itself, `librefang-runtime-mcp` and `librefang-runtime` all compile clean at 3.1 — only the kernel fails.

Dependabot's attempt was #6710, closed in favour of a hand-authored migration. This issue carries what was learned so the next attempt does not restart from zero.

## The feature-bisect lever does not exist

The cheapest hypothesis was that rmcp 3.1's newly feature-gated `bytes` / `base64 0.23.0` pulled in the trigger, so dropping `transport-streamable-http-client-reqwest` and then `reqwest` would name it. That experiment is **not runnable, and would not have been informative anyway**.

Dropping `transport-streamable-http-client-reqwest` fails `librefang-runtime-mcp` itself with three `E0432` unresolved imports, at `crates/librefang-runtime-mcp/src/lib.rs:1753`, `:1754` and `:1918`. `connect_streamable_http` (`lib.rs:1739`) and `extract_auth_required` (`lib.rs:1918`) carry no `#[cfg(feature = …)]` guard, so the kernel is never typechecked in that configuration — a "0 errors" reading there would mean "the kernel was not compiled", not "the errors vanished". Adding cfg gates to force it would change runtime-mcp's own async-fn graph at the same time as the feature set, confounding the single variable the experiment exists to isolate.

Four independent facts, each verified against source, say no feature can be the trigger:

1. **`librefang-kernel` has zero rmcp coupling.** No `rmcp` entry in `crates/librefang-kernel/Cargo.toml`, and the only match for `rmcp` under `crates/librefang-kernel/src/**` is a prose comment at `kernel/mcp_setup.rs:189`. Nothing feature-gated can enter the kernel's type graph by name.
2. **Not one of the 71 error notes names an rmcp type.** Aggregated: `&kernel::LibreFangKernel` ×30, `&librefang_channels::types::SenderContext` ×24, `&str` ×8, `&std::path::Path` ×3, then one each of `&workflow::Workflow`, `&tokio::sync::RwLock<HashMap<WorkflowId, Workflow>>`, `&Arc<LibreFangKernel>`, `&DeferredToolExecution`, `&bool`, and a closure at `triggers_and_workflow.rs:989`. All librefang-side. This is region-generalization propagation, not an rmcp type losing `Send`.
3. **The transport is already type-erased before the kernel sees it.** `crates/librefang-runtime-mcp/src/lib.rs:802` defines `type DynRmcpClient = rmcp::service::RunningService<RoleClient, Box<dyn rmcp::service::DynService<RoleClient>>>`, stored as `McpInner::Rmcp(DynRmcpClient)`. `StreamableHttpClientTransport` cannot appear in a kernel future regardless of feature selection.
4. **The `bytes` / `base64` gating is structurally identical between 2.2.0 and 3.1.1.** Both gate them behind `client-side-sse`, `reqwest-tls-no-provider` and `transport-streamable-http-client*`. Those features were already enabled at 2.2.0, which is green. 3.1 did not newly expose them to this workspace.

So there is no feature to drop and no upstream feature-gating bug to report.

## What did change in rmcp's core

Diffing `rmcp/src/service.rs` between the vendored 2.2.0 and 3.1.1 sources — none of it feature-gated:

**`pub trait ServiceRole` went from a pure associated-type trait with zero methods to five methods.** The interesting one is `invalidate_response_cache(_peer: &Peer<Self>, _notification: &Self::PeerNot) -> impl Future<Output = ()> + MaybeSendFuture` at `service.rs:149` — an RPIT default method taking reference parameters, which is a textbook region-generalization hazard, and the only new `MaybeSendFuture` instance in the crate (total occurrences went 123 → 124). The other four are `configure_direct_peer`, `peer_cancelled_params`, `enforce_request_association` and `enforce_peer_request_association`.

**Supertrait bounds tightened into associated-type equalities.** `type Req: … + GetMeta` became `GetMeta<Metadata = RequestMetaObject>`, same for `PeerReq`, and `PeerNot`'s `GetMeta` became `GetMeta<Metadata = NotificationMetaObject>`.

Both live in core, so the conclusion is the same either way. Isolating which one is responsible needs a `[patch.crates-io]` vendored rmcp and was out of scope.

Worth recording why the earlier docs.rs check missed this: checking the auto-trait impls on `DynService` / `RunningService` / `Peer` was a valid negative, but those listings show the `where` clauses on the structs, not new default methods on the `ServiceRole` bound they are generic over.

## The actual next lever, which needs a decision before anyone writes it

Type-erase `McpConnection` out of the `KernelApi` `#[async_trait]` `Send` proof — box the futures at the runtime-mcp → kernel boundary so the kernel's `async_trait` `Send` obligations stop being computed through runtime-mcp's inference-derived opaque future types.

The blast radius is real and entirely librefang-side. The 71 sites: 18 in `crates/librefang-kernel/src/kernel_api.rs`, 14 in `kernel/triggers_and_workflow.rs`, 11 in `kernel/handles/agent_control.rs`, 10 in `kernel/handles/workflow_runner.rs`, with the remainder scattered across `goal_lifecycle`, `task_registry`, `approval_gate`, `cron_tick`, `accessors`, `inbox`, `mod` and `messaging` / `background_lifecycle`. The signature that carries rmcp into the proof is `kernel_api.rs:146-149` `fn mcp_connections_ref(&self) -> &tokio::sync::Mutex<Vec<librefang_runtime::mcp::McpConnection>>`, backed by `kernel/subsystems/mcp.rs:44`.

That is a refactor across `kernel_api.rs`, `kernel/subsystems/mcp.rs`, `kernel/subsystem_forwards.rs` and every `mcp_connections_ref` caller. It should not ride along inside a version bump.

## Two behavioural decisions the migration owes regardless of how `Send` resolves

These are the real reason this cannot be a dependabot merge, and both are already decided (see the closing comment on #6710) — recorded here so they are not re-litigated:

- **`call_tool` vs `call_tool_once`.** In 2.2 `call_tool` returned after one round; in 3.1 it internally drives multi-round tool results to completion. LibreFang calls it at `crates/librefang-runtime-mcp/src/lib.rs:2982` from inside per-agent semaphores and a `tokio::time::timeout` budget, so a call that can now loop internally changes latency, cancellation and lock-hold behaviour. The migration uses `call_tool_once` to preserve today's semantics.
- **Client response caching is new and default-enabled.** It will be disabled: a cached result served for a side-effecting tool is a correctness bug, and MCP gives us nothing to distinguish the safe tools.

Each gets a `#[tokio::test]` pinning it, so a future rmcp bump that flips either default fails a test rather than silently changing behaviour.

## Upstream report worth filing

At `modelcontextprotocol/rust-sdk`, on the correct subject: `ServiceRole` gaining default methods — notably `invalidate_response_cache` returning `impl Future + MaybeSendFuture` with `&Peer<Self>` / `&Self::PeerNot` reference parameters — plus the `GetMeta<Metadata = …>` associated-type-equality tightening, as a semver-minor source-breaking change for downstreams holding an rmcp client inside an `#[async_trait]` `Send` boundary. Frame it as "the 3.0 migration guide omits a Send / region-generalization break", flag both candidates, and state plainly that they were not individually isolated.

## Not urgent

No advisory forces the upgrade — `Security` and both `cargo-deny` jobs are green on the 3.1 branch. rmcp stays pinned at 2.x until this is scheduled.

## Reproduction, already built

A worktree is left at `/Volumes/Lexar/worktrees/rmcp-bisect` on branch `chore/rmcp-3.1-bisect` (off `main` @ c403439a), uncommitted, restored to the exact failing configuration: `Cargo.toml:143` bumped to `rmcp = "3.1"` with the original full feature list, `Cargo.lock` resolved to rmcp 3.1.1, and the three `rmcp::model::Meta::new` → `rmcp::model::RequestMetaObject::new` renames applied at `crates/librefang-runtime-mcp/src/lib.rs:2961`, `:2974` and `:6336`. Those three renames are the complete API-surface change 3.1 requires; the error census is 100% homogeneous with no `E0xxx` codes at all.

Verified API shapes for whoever continues: `RequestMetaObject::new()` exists, `RequestMetaObject: DerefMut` so `.insert(…)` still works, and `CallToolRequestParams.meta` is still `Option<RequestMetaObject>` (`rmcp-3.1.1/src/model.rs:4057`), so `params.meta = Some(meta)` typechecks unchanged.

houko:
Closes #6749.

## What changed

- `Cargo.toml`: `rmcp` 2.0 (resolved 2.2.0) → 3.1 (resolved 3.1.4), all four features unchanged.
  `cargo update -p rmcp` moves exactly one package — no new transitive dependencies.
- `rmcp::model::Meta` → `rmcp::model::RequestMetaObject` at three sites in `crates/librefang-runtime-mcp/src/lib.rs` (two in `call_tool_with_caller`, one in the test module).
- `client.call_tool(params)` → `client.call_tool_once(params)` and `client.read_resource(params)` → `client.read_resource_once(params)`, each with a `match` on the new response enum.

## Why the `_once` variants

rmcp 3.1 adds an inherent `impl<S> RunningService<RoleClient, S>` block (`src/service/client.rs:1897`) that shadows two methods we previously reached through `Deref<Target = Peer<RoleClient>>`.
The inherent `call_tool` and `read_resource` are MRTR drivers — they loop up to `DEFAULT_MRTR_MAX_ROUNDS` (10) SEP-2322 `input_required` rounds, and they capture `S` in the returned future.

Both consequences are unwanted here:

- The loop would have run inside the existing `tokio::time::timeout`, so a budget written for one request would have silently covered ten.
- The captured `S` is what an inherent `async fn` leaks to its caller, and it made every future holding an `McpConnection` non-`Send`.

`call_tool_once` / `read_resource_once` send exactly one request and are `Send`-clean.
They are also the semantically faithful choice: `RootsClientHandler` builds its `ClientCapabilities` with only `roots` set, declaring neither the tasks extension nor elicitation, so a compliant server has no license to send `input_required` or a task response.
The non-`Complete` match arms return an error rather than silently discarding a response we never asked for.

`ProtocolVersion::LATEST` is `2025-11-25` in both 2.2.0 and 3.1.4, so nothing changes on the wire.

## Measurement

The `Send` breakage is not local to the MCP crate — `librefang-runtime-mcp` compiles clean against 3.1.4 with only the `Meta` rename, because `Send` is demanded at the spawn sites in the kernel, not where the future is written.

| State | `cargo check -p librefang-kernel --lib` |
| --- | --- |
| Unmodified `origin/main` (baseline control) | 0 errors |
| rmcp 3.1.4, `Meta` rename only | 36 errors |
| rmcp 3.1.4, `read_resource_once` only | 36 errors |
| rmcp 3.1.4, `call_tool_once` only | 36 errors |
| rmcp 3.1.4, both `_once` calls | 0 errors |

All 36 are `implementation of std::marker::Send is not general enough`, spread across `cron_tick.rs`, `inbox.rs`, `goal_lifecycle.rs`, `accessors.rs`, `messaging.rs`, `background_lifecycle.rs`, `task_registry.rs`, `triggers_and_workflow.rs`, `kernel_api.rs`, and `handles/{agent_control,approval_gate,workflow_runner}.rs`.
Each of the two call sites reproduces the full set on its own, so neither fix is redundant.

The baseline row matters: the diagnostics name `&bool`, `&str` and `&std::path::Path` as failing `Send`, which is not credible on its face, so the run against unmodified `origin/main` is what establishes that the 36 are attributable to this bump rather than to a pre-existing break or a poisoned shared target.

## Verification

- `cargo check -p librefang-runtime-mcp --lib` — 0 errors.
- `cargo check -p librefang-runtime-mcp --tests` — 0 errors.
- `cargo check -p librefang-runtime --lib` — 0 errors.
- `cargo check -p librefang-kernel --lib` — 0 errors.
- `cargo fmt --check -p librefang-runtime-mcp` — clean.
- No route or config surface is touched, so no new `TestServer` integration test applies.
- Workspace-wide build, clippy and the full test lanes are left to CI.

## Not done here

Nothing deferred.
The MRTR and tasks extensions that 3.1 exposes are new capability surface, not part of restoring parity with 2.2 behaviour; adopting either would mean declaring the matching `ClientCapabilities` and designing how an `input_required` round reaches the agent loop, which is a feature decision rather than a dependency upgrade.


Cargo.lock
```diff
@@ -7986,14 +7986,16 @@ dependencies = [
 
 [[package]]
 name = "rmcp"
-version = "2.2.0"
+version = "3.1.4"
```
houko, line 7989: The PR description states "`cargo update -p rmcp` moves exactly one package — no new transitive dependencies," but this `Cargo.lock` diff touches 18 package blocks, not one: alongside `rmcp` itself, `anstyle-query`, `anstyle-wincon`, `colored`, `dirs-sys`, `errno`, `generator`, `muda`, `nu-ansi-term`, `os_pipe`, `rustix`, `rustls-platform-verifier`, `socket2`, `tempfile`, `termina`, `tray-icon`, `uds_windows`, and `winapi-util` all get bumped (mostly `windows-sys`/`windows-link` unification onto 0.61.x).

None of those are reachable from `rmcp`'s dependency tree — `tray-icon`, `muda` and `colored` are `librefang-desktop`/`librefang-cli` dependencies with no relation to MCP. That suggests an unscoped `cargo update` ran rather than `-p rmcp` alone.

Functionally this is probably harmless (patch/minor `windows-sys` unification), but it means the PR's own verification table — which only exercises `librefang-runtime-mcp`, `librefang-runtime` and `librefang-kernel` — doesn't cover the crates whose transitive deps actually moved (`librefang-desktop`, `librefang-cli`). Worth either re-running a scoped `cargo update -p rmcp --precise 3.1.4` to get a minimal lockfile diff, or extending the verification/PR description to acknowledge the wider bump.

(Separately verified: `cargo check -p librefang-runtime-mcp --lib --tests` and `cargo clippy -p librefang-runtime-mcp --lib --tests -- -D warnings` are both clean against this branch.)

---

houko (commented):
Reviewed against CLAUDE.md conventions. Code changes (rmcp 3.1 migration, `_once` client calls, `Meta`→`RequestMetaObject` rename) look correct — verified `cargo check -p librefang-runtime-mcp --lib --tests` and `cargo clippy -p librefang-runtime-mcp --lib --tests -- -D warnings` both pass clean on this branch, and `cargo fmt` is clean. Commit message and changelog fragment both follow repo conventions. One inline note on `Cargo.lock` scope below — informational, not blocking.

---

Issue #6699, opened 2026-08-01
feat: ephemeral agent spawning (subagent pattern)


DaBlitzStein:
## Summary

Agents should be able to spawn ephemeral workers — like Claude Code's `Agent` tool — that run a task with caller-configured tools, return a result, and vanish.
No workspace, no DB persistence, no cleanup debt.

## Motivation

Today `spawn_agent` always creates a permanent agent with full ceremony (7 dirs, 9 identity files, SQLite, capability/scheduler/trigger registration).
There is no lightweight "run this task, get the result, disappear" path.
The closest is `/btw` (`send_message_ephemeral`), but it's hardcoded to zero tools and requires a pre-existing agent.

## Proposed design

Three spawn modes via extended `tool_agent_spawn`:

| Mode | Workspace | SQLite | Tools | Cleanup |
|------|-----------|--------|-------|---------|
| **Ephemeral** (`ephemeral: true`) | None | None | Caller-configured | Auto |
| **Ephemeral from type** | None | None | Defined by agent type | Auto |
| **Permanent** (today) | Full | Full | From manifest | Manual |

### Core types
- `AgentLifecycle` enum (`Permanent` / `Ephemeral`)
- `EphemeralSpawnRequest` struct (system_prompt, agent_type, model, tools, skills, message, max_iterations)
- `EphemeralSpawnResult` struct (response, cost_usd, iterations)

### Agent types store
- `~/.librefang/templates/` directory with `.toml` manifests
- CRUD API: `GET/POST/PUT/DELETE /api/agent-types`
- Dashboard page with card grid, create/edit form, quick run

### Kernel
- `spawn_ephemeral_worker` on `LibreFangKernel` (modeled on `send_message_ephemeral` but with tools)
- `spawn_ephemeral` + `spawn_ephemeral_detailed` on `AgentControl` trait
- Template resolution: `~/.librefang/templates/` first, falls back to `workspaces/agents/`

### Runtime
- `tool_agent_spawn` extended with `ephemeral: true` flag
- When ephemeral: builds `EphemeralSpawnRequest`, returns result text directly (not agent ID)

### API
- `POST /api/agents/spawn-ephemeral` — HTTP entry point
- Agent types CRUD with JSON accept/return

### Dashboard
- Agent Types page (sidebar, between Agents and Chat)
- Cards with name/description/model, Quick Run, Edit, Delete
- Spawn ephemeral from Chat page (future)

## Implementation status

[...]

PR #7887, opened 2026-08-24
fix(kernel): bill ephemeral worker usage to its parent

houko:
`main` does not compile. `librefang-kernel` fails with:

```
error[E0063]: missing field `billed_agent_id` in initializer of `UsageRecord`
  --> crates/librefang-kernel/src/kernel/ephemeral_spawn.rs:484:28
```

Every open pull request inherits this on its next CI run, so it needs to land before anything else does.

## How it happened

Two pull requests that were each green against their own base:

- #7878 added `billed_agent_id: Option<AgentId>` to `UsageRecord` (`crates/librefang-memory/src/usage.rs:70`) and filled it at every constructor that existed at the time.
- #7885 landed shortly after, and #7875 before it had introduced `crates/librefang-kernel/src/kernel/ephemeral_spawn.rs` — a new `UsageRecord` constructor that #7878's base did not contain.

Neither change is wrong. This is the case the repository's merge guidance calls out: the `main` ruleset has no strict required-status-checks policy, so each pull request merges on CI run against its own base, and two independently green changes can break `main` together.

## The fix

`billed_agent_id: Some(parent_id)`.

An ephemeral worker has no registry entry of its own, so there is no `AgentEntry` for `billed_agent_for` to read a parent from.
The parent IS the billed agent by construction — the worker runs on its behalf and against its budget — which is the same answer `billed_agent_for` returns for a parented agent, since it is `entry.parent.unwrap_or(entry.id)`.
That also matches what #6699 asked for: budget attributed to the parent.

The record already set `agent_id: parent_id`, so this makes the two fields agree rather than introducing a new attribution rule.

## Verification

- `cargo check -p librefang-kernel --lib` — clean (was E0063 before the change).
- `cargo test -p librefang-kernel --lib ephemeral` — 8 passed, including `an_ephemeral_worker_cannot_exceed_its_parents_tools` and `test_send_message_ephemeral_does_not_modify_session`.

Found by the `Kernel Ignored / Restart Override` job on #7886, which was the first lane to compile the kernel after both changes were present.


crates/librefang-kernel/src/kernel/ephemeral_spawn.rs
```diff
@@ -483,6 +483,9 @@ impl LibreFangKernel {
         );
         let usage_record = librefang_memory::usage::UsageRecord {
             agent_id: parent_id,
+            // An ephemeral worker has no registry entry of its own, so there is no `AgentEntry` for `billed_agent_for` to read a parent from.
+            // The parent IS the billed agent by construction — the worker runs on its behalf and its budget — which is the same answer that helper returns for a parented agent (#7714, #6699).
+            billed_agent_id: Some(parent_id),
```
houko, line 488: No test pins this assignment. CLAUDE.md's integration-testing mandate is scoped to route/wiring changes, so it doesn't strictly require one here, but this is exactly the kind of "compiles either way, silently wrong either way" field that #7714's own migration notes call out (`usage.rs:1523`: "Every pre-#7714 call site leaves `billed_agent_id` unset"). Worth a small unit test asserting `check_all_and_record`/`billed_agent_for`-style rollup for the ephemeral-worker path specifically, so a future refactor that reintroduces a missing/self-referential `billed_agent_id` here fails a test instead of only a compile.

Note the assignment is a no-op for the `COALESCE(billed_agent_id, agent_id)` read path today, since `agent_id` is already `parent_id` on this record (line 485) — the field is set for consistency with the `billed_agent_for` contract, not because it changes current query results. Leaving this as a comment since it's a coverage gap, not a bug.

---

houko (commented):
Applied two mechanical CLAUDE.md fixes directly on the branch: renamed the changelog fragment to lead with the PR number (repo convention, every other fragment follows it), and rewrapped the new kernel comment to sentence boundaries per the prose-wrapping rule. `cargo check -p librefang-kernel --lib` and `cargo clippy -p librefang-kernel --all-targets -- -D warnings` are clean; `check-changelog-attribution.py` passes. Left one non-blocking comment on a test-coverage gap.

---

Issue #4868, opened 2026-05-10
Channel `/new` wipes ALL agent sessions, not just the channel-derived one


neo-wanderer:
## Summary

The channel-side `/new` command (Telegram, Discord, Slack, etc.) currently calls the kernel's `reset_session(agent_id)`, which deletes **every** session for that agent — the default session plus every per-channel session and any cron/trigger-spawned ones — and creates one fresh session in its place.

For users who run an agent across multiple surfaces (e.g., dashboard + Telegram + Discord on the same hand), typing `/new` in any one channel silently destroys the in-flight history on all the others.

## Repro

1. Same agent reachable from the dashboard and from Telegram.
2. Have an active session on each (different `SessionId`s — one is `SessionId::for_channel(agent,"telegram:<chat>")`, the other is the default).
3. Send `/new` in Telegram.
4. Observe: dashboard session row is gone (`GET /api/sessions/<dashboard-sid>` → 404).

## Code path

- `crates/librefang-channels/src/bridge.rs:4730` — `"new"` command handler calls `handle.reset_session(aid)` with no scope.
- `crates/librefang-kernel/src/kernel/session_ops.rs:168` — `reset_session(agent_id)`:
  - line 176: `get_agent_session_ids(agent_id)` — iterates all sessions
  - line 199-202: `delete_agent_sessions(agent_id)` — transactional delete of every session row for that agent
  - The comment at line 173-174 documents the agent-wide intent: *"Auto-save session summaries for ALL sessions (default + per-channel) before clearing, so no channel's conversation history is silently lost."*

Summaries are saved to memory before the delete, but the full transcripts are unrecoverable.

## Expected

`/new` invoked from a channel should reset **only** that channel's `SessionId::for_channel(agent, "<channel>:<chat>")`. Other channels and the dashboard session should be untouched.

The agent-wide reset is still useful as a separate explicit operation (e.g., dashboard "Reset agent" button, `POST /api/agents/{id}/reset`), but the channel `/new` user-mental-model is "start a fresh chat *here*", not "wipe everything across every surface."

## Proposed change

1. Add an optional scope to the kernel API:
   ```rust
   pub fn reset_session(&self, agent_id: AgentId, scope: ResetScope) -> KernelResult<()>
   // where
   pub enum ResetScope { Agent, Session(SessionId) }
   ```
[...]

PR #7896, opened 2026-08-25
fix(channels): scope broadcast turns to the chat's session so /new can reach them

houko:
Takes over the session half of #7140 from @DaBlitzStein's #7701, which has sat mergeable-but-stale on a fork I cannot push to.
The diagnosis is theirs: a channel `/new` can clear the per-chat session while the conversation the user is looking at lives in the agent's canonical `entry.session_id`, so the command acks and nothing changes.
I recommend #7701 be closed as superseded once this lands — a recommendation only; I have not touched it.

## Why the implementation differs from #7701

#7701 resets the canonical session unconditionally alongside the derived one.
Verifying that on current `main` turned up two things that changed the shape of the fix.

**The `/new` no-op #7701 was filed against has already been fixed, by a different mechanism.**
#7705 landed `session_scope` (`crates/librefang-channels/src/bridge.rs:2851`), the single derivation shared by `build_sender_context` and the reset command arms.
The Telegram case in #7701's description — sender id in `SENDER_USER_ID_KEY` metadata, `platform_id` empty, so `/new` cleared `for_channel(agent, "telegram")` while the chat lived in `for_channel(agent, "telegram:<user>")` — is covered by `session_scope_agrees_with_build_sender_context_when_platform_id_is_empty`.

**Resetting the canonical session unconditionally would destroy the dashboard conversation.**
The WebUI chat sets `use_canonical_session: true` (`crates/librefang-api/src/ws.rs:1221`), so the agent's canonical session *is* the dashboard conversation.
`reset_one_session` deletes the row it is given (`crates/librefang-kernel/src/kernel/session_ops.rs:454`), so a Telegram `/new` would wipe the dashboard history of every agent shared between the two surfaces, under an ack that reads "Other surfaces untouched." — the collateral damage `reset_channel_session` was introduced to avoid in #4868.
`/compact` is worse: it passes `force = true`, which bypasses the "nothing to do" short-circuit at `crates/librefang-kernel/src/kernel/agent_runtime.rs:502`, so an empty canonical session would buy a real LLM call on every `/compact`.

So instead of clearing the canonical session, this closes the path by which a channel turn reaches it at all.

## The remaining path

`resolve_dispatch_session_id` (`crates/librefang-kernel/src/kernel/mod.rs:686`) derives `SessionId::for_sender_scope` only when a `SenderContext` is present, and otherwise falls back to `entry.session_id`.
Broadcast fan-out was the one channel dispatch that sent without one — `handle.send_message(aid, &t)` at `bridge.rs:4677` (parallel) and `:4699` (sequential); every other channel send path passes `sender_ctx`.
A sender with `[channels] broadcast` routing therefore accumulated their whole conversation in the canonical session, and `/new` cleared a per-chat session nothing had ever written to.

## Changes

- Broadcast dispatch builds the same `SenderContext` the rest of `dispatch_message` uses and sends through `send_message_with_sender`, in both the parallel and the sequential strategy.
  Each target now keeps its own session for the chat, and gets the peer-scoped memory lookup and `/think` override every other channel turn gets.
- `/new`, `/reboot` and `/compact` apply to every agent in the broadcast fan-out, not just the one the router chain resolves.
  Without this the fix would be half a fix: the turns land in per-chat sessions, and the reset still reaches only one of the agents answering.
  Targets are the union of the router-resolved agent and the broadcast targets, so a chat without broadcast routing resets exactly what it reset before.
- Identical acks are deduplicated, so single-agent wording is unchanged.

## Verification

- `broadcast_dispatch_carries_the_per_chat_session_scope` (`crates/librefang-channels/tests/bridge_integration_test.rs`) — drives a broadcast message through the real `BridgeManager` and asserts both targets were dispatched with scope `("telegram", Some("vip_user"))`, failing if either arrives with no sender context.
- `new_resets_every_broadcast_target`, `reboot_and_compact_reach_every_broadcast_target`, `new_without_broadcast_resets_only_the_resolved_agent` (`crates/librefang-channels/src/bridge.rs`) — fan-out coverage plus the no-regression case, which also pins the ack wording.
- **Not compiled locally.** This machine's disk was at 8 GiB free with a 44 GB shared `target/`, under the floor I work to, so `cargo check`/`clippy`/`nextest` did not run; `rustfmt --check` is clean on both files. CI is the gate — please do not merge red.

## Scope

`Refs #7140`, not `Closes`: root cause 2 (`resolve_for_command` consults only the router chain, `bridge.rs:6946`) and root cause 3 (`set_conversation_binding` still has no production caller, `crates/librefang-memory/src/channel_binding_store.rs:123`) are untouched and remain open, as your 2026-08-23 audit on the issue records.
Root cause 1 (#7848) and the `/think` half of root cause 4 (#7854) are confirmed present on `main` at `eac1a6bdf`.

### DIFF
--- a/changelog.d/fixed/7896-broadcast-session-scope.md
+++ b/changelog.d/fixed/7896-broadcast-session-scope.md
@@ -1,0 +1,5 @@
+`/new`, `/reboot` and `/compact` now clear the session a broadcast chat is actually talking in.
+Broadcast fan-out was the one channel dispatch that reached the kernel without a `SenderContext`, and the kernel derives the per-chat `SessionId::for_sender_scope` only when a sender context is present — so those turns accumulated in the agent's canonical session, the one the dashboard chat writes to, while the reset commands addressed a per-chat session nothing had ever written to.
+The command acked success and the bot went on answering out of the history the user had just asked it to forget.
+Broadcast dispatch now carries the same sender context every other channel turn carries, so each target keeps its own session for the chat, and the three reset commands cover every agent in the fan-out rather than the single agent the router chain happens to resolve.
+The canonical-versus-derived session framing is @DaBlitzStein's, from #7701. (#7896) (@houko)
--- a/crates/librefang-channels/src/bridge.rs
+++ b/crates/librefang-channels/src/bridge.rs
@@ -4661,6 +4661,10 @@ async fn dispatch_message(
                 }
             }
 
+            // Broadcast dispatch carries the same `SenderContext` every other channel turn carries (#7140, session dimension).
+            // Without it the kernel's session resolver falls past its channel branch — `resolve_dispatch_session_id` only derives `SessionId::for_sender_scope` when a sender context is present — and lands the turn on the agent's canonical `entry.session_id`, the session the dashboard chat writes to.
+            // The conversation the broadcast user sees then lives somewhere `/new` cannot reach: the reset commands address the per-chat session, so they clear an empty row and report success while the visible history survives.
+            let broadcast_sender_ctx = build_sender_context(message, overrides.as_ref());
             let strategy = router.broadcast_strategy();
             let mut responses = Vec::new();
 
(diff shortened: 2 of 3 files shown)


houko:
Compile verification, which the PR body correctly recorded as missing.

The branch was written while this machine had 6-10 GiB free against a 44 GB shared target directory, so nothing was compiled at the time. Disk has since been reclaimed and the checks now ran:

- `cargo check -p librefang-channels --tests` — clean.
- `cargo test -p librefang-channels` — 628 passed, 0 failed, across all five test binaries.
- `cargo clippy -p librefang-channels --all-targets -- -D warnings` — clean.

No code changed to achieve this; it is the same commit the PR was opened with.

Issue #3538, opened 2026-04-27
migrate_v13 / v18 skip INSERT INTO migrations audit row — table diverges from user_version


houko:
## Summary
Every migration ends with `INSERT OR IGNORE INTO migrations (version, applied_at, description) VALUES (N, ...)` — except v13 and v18, which skip it entirely. v17 also has an unguarded `let _ = ALTER TABLE ... ADD COLUMN second_factor_used` that swallows real errors as well as "duplicate column".

## Location
- `crates/librefang-memory/src/migration.rs:461-529` (v13)
- `crates/librefang-memory/src/migration.rs:633-641` (v18)
- `crates/librefang-memory/src/migration.rs:626` (v17 unguarded ALTER)

## Failure
- Operator queries `SELECT * FROM migrations ORDER BY version` for diagnostics; sees gaps at v13 and v18 → assumes those migrations didn't apply.
- Schema is actually current (`user_version=23` is set unconditionally at end of `run_migrations`), but the audit log lies.
- Tooling that uses the `migrations` table as ground truth (e.g. report.rs) makes wrong decisions; an automated rollback that "v18 wasn't applied" leaves a stranded `totp_lockout` table.
- v17's `let _ = ALTER ... ADD COLUMN second_factor_used` succeeds silently for "already exists" AND for any other ALTER failure (lock, disk, schema). Downstream INSERTs that include the column then fail at runtime, not upgrade time.

## Fix
- Add the missing `INSERT OR IGNORE INTO migrations` rows for v13 and v18.
- Replace v17's `let _ = ALTER ...` with `if !column_exists(conn, "approval_audit", "second_factor_used") { conn.execute(...)?; }` — same pattern as v14/v15/v16.

PR #7904, opened 2026-08-25
feat(ephemeral): persist a run record under the spawning agent

### LINKED ISSUE #7752: feat: ephemeral sub-agent runs should persist under the parent, not vanish and not orphan
## The false dichotomy

Ephemeral workers (`agent_spawn` with `ephemeral: true`) were spawned with `incognito: false`, so every disposable run wrote its session to SQLite and pushed the exchange into episodic memory — under a throwaway `AgentId` with no registry entry.

The rows are unreachable: nothing can look them up, nothing can clean them, and the path is designed to be called cheaply and often.
Worst of both worlds — the cost of persistence with none of the value.
The changelog for that work claimed "no DB persistence", which was not true.

The immediate fix was `incognito: true`: better to write nothing than to write garbage.
But "vanish" is not the right end state either, and this issue is about the third option.

## What is actually wanted

The question is not *whether* to persist a sub-agent run. It is **under whose identity**.

Claude Code is the reference here: a subagent's transcript is saved, but it hangs off the parent session rather than floating free. That makes it auditable after the fact, and resumable — you can pick a subagent back up with its context intact, because the record is attached to something that still exists.

Applied here:

- An ephemeral run persists **as a child of the session that spawned it**.
- The parent can enumerate its children: what was delegated, what came back, what it cost.
- Retention follows the parent — when the parent session is compacted or deleted, its ephemeral children go with it. No orphans by construction.
- Episodic memory writes are attributed to the parent, or suppressed, but never filed under an id that does not exist.

## What already exists

`LoopOptions` already carries `parent_session_id: Option<SessionId>` (`crates/librefang-runtime/src/agent_loop/types.rs:134`), alongside `is_fork` (`:59`) and `incognito` (`:69`).

So the plumbing for parent linkage is present; the ephemeral spawn path in `crates/librefang-kernel/src/kernel/messaging.rs` simply does not use it — it passes neither a parent session nor a resolvable identity, which is what forced the `incognito` workaround.

## Proposal

1. Ephemeral spawns pass the spawning session as `parent_session_id`.
2. Persist the run under that parent instead of under a throwaway agent id, so `incognito` stops being the only way to avoid orphans.
3. Expose the children: an API and dashboard/TUI view answering "this agent fired 12 ephemeral workers this turn — here is what each did and what each cost".
4. Cascade retention from the parent, so cleanup needs no separate sweeper.
5. Keep `incognito: true` available as an explicit opt-in for callers who genuinely want no trace, rather than as the default hiding a design gap.

## Why it matters

Today an operator watching an agent misbehave through ephemeral workers has nothing to inspect: the runs left no reachable trace, and now leave none at all.
Cost attribution has the same shape — spend is billed to the parent (fixed separately), but the *work* that produced the spend is invisible.

## Relationship to #7744

#7744 proposes that whatever an agent creates belongs to the principal it acted for.
This is the sharpest instance of that principle: a sub-agent run is work performed on behalf of someone, and it should be filed under them.
Resolving #7744's ownership chain gives this issue its answer for free; this issue is the concrete case that motivates it.

houko:
Closes #7752

## What this is

An ephemeral worker (`agent_spawn` with `ephemeral: true`, merged in #7875) runs one turn under its parent's identity and then vanishes: no registry entry, no persisted session, and a mission workspace deleted on the way out.
#7878 gave the parent the *spend* through `usage_events.billed_agent_id`, but the work behind the spend had no record at all — so an operator watching an agent misbehave through disposable workers had nothing to inspect, and "what did this agent delegate, and what did each one cost" had no answer.

This adds the run record the issue asks for: an `ephemeral_runs` table written by the kernel after each run, and `GET /api/agents/{id}/ephemeral-runs` to read it back under the parent.

## A run record is not a session row, and that distinction is the design

The issue's proposal 2 reads as "stop using `incognito` to avoid orphans", and the one-line version of that — clear the worker's `incognito` flag so the agent loop persists its session under the parent's `AgentId` — is wrong.

`incognito` gates far more than the session write.
It also gates the episodic-memory write (`end_turn.rs:257`), the context-engine `after_turn` advance (same block), the proactive-memory `auto_memorize` pass (`end_turn.rs:330`), and ~20 further sites across `mod.rs` / `run_streaming.rs` / `tool_call.rs`.
Every one of those keys on `session.agent_id`, which for an ephemeral worker **is the parent** (`ephemeral_spawn.rs`, session construction).

So clearing the flag would fold a delegated sub-run's exchange into the parent's own episodic recall — teaching the parent it said things it never said and never saw — advance the parent's context engine on text it did not produce, and file the worker's throwaway session among the parent's real conversations in every session-list view.
That is the outcome @houko's own triage comment ruled out ("Episodic memory: suppressed, not re-attributed").

The record here is written by the kernel, after the run, out of band from the agent loop's persistence boundary. `incognito: true` stays exactly where it is.

## A run record is also not a surviving workspace

#7860's `MissionWorkspace` deletes the worker's scratch directory on every exit path — success, error, and panic unwind — by design.
That is untouched. Keeping a record of *what a worker did* is a different question from keeping *the scratch files it did it with*, and the answer to the second stays "no".
Both properties are asserted together in the same tests rather than separately, so a future change cannot buy one by quietly trading away the other.

## Substantive changes

- **`crates/librefang-memory/src/ephemeral_run_store.rs`** (new): `EphemeralRunStore` — `record_run`, `list_for_parent`, `rollup_for_parent`, `delete_for_agent`, plus `truncate_for_record`.
- **Migration v51**: `ephemeral_runs`, indexed on `(parent_agent_id, finished_at DESC)`, `status` constrained to `completed` / `failed`. Purely additive; no existing table touched and no existing row changes meaning.
- **`substrate.rs`**: the cascade. `remove_agent_inner` now deletes an agent's run records in the same transaction as its sessions and structured rows — "retention follows the parent, no orphans by construction".
- **Retention cap**: each insert trims its own parent back to the most recent `MAX_RUNS_PER_PARENT` (200) rows in the same transaction. The ephemeral path is designed to be called cheaply and often; an unbounded per-parent history would recreate a milder version of the row-bloat problem this feature replaces. A compiled constant rather than a config knob, so the table stays bounded without an operator having to know the knob exists — say the word if you want it configurable and I will add the full `KernelConfig` checklist.
- **Text cap**: the delegated task and the worker's answer are stored truncated to 4096 chars, clipped on a `char` boundary and marked. Model output is unbounded and this table takes one row per spawn.
- **`ephemeral_spawn.rs`**: records on completion *and* on a failure inside `run_agent_loop` (status `failed`, with the error). A failure that never reached the loop — depth refusal, quota, unknown tool, driver resolution — records nothing, because no run happened. Best-effort, like the usage record beside it: the tokens are already spent, so a bookkeeping failure must not discard an answer the caller paid for.
- **`GET /api/agents/{id}/ephemeral-runs`**: `runs` plus a `rollup` (count, failures, cost, tokens). Owner-scoped like the sibling `/stats` and `/events`, and for a stronger reason — a run record carries conversation content, not just counters. `limit` defaults to 30, caps at 200.

Attribution agrees with #7878 rather than restating it: the row's `parent_agent_id` is the same id the usage row bills to, which is what `billed_agent_for(entry) = entry.parent.unwrap_or(entry.id)` yields for a parented agent. A test asserts the two views of one run cannot disagree about the owner.

## Drive-by fix found while editing `migration.rs`

`migrate_v50` recorded its audit row as **version 49**, where `INSERT OR IGNORE` silently dropped it against v49's existing row.
The #3538 backfill then re-inserted version 50 with the placeholder description `audit-row backfill (#3538)` and logged `Migration audit drift detected and self-healed` — on **every fresh install**.

`test_every_migration_records_audit_row` passed anyway, because it runs *after* that backfill and only checks the rows exist. It now also asserts the backfill inserted nothing, which is the assertion that actually catches this. Verified by reintroducing the typo: the strengthened test fails with the intended message, and passes with the fix.

## Verification

- `cargo check -p librefang-memory -p librefang-kernel -p librefang-api --lib` — clean
- `cargo clippy -p librefang-memory -p librefang-kernel -p librefang-api --lib --tests -- -D warnings` — clean
- `cargo test -p librefang-memory --lib -- ephemeral migration::` — 53 passed
- `cargo test -p librefang-kernel --lib -- ephemeral mission a_run_record` — 28 passed
- `cargo test -p librefang-api --test agent_ephemeral_runs_test` — 5 passed
- `cargo test -p librefang-api --test dead_route_audit_test --test openapi_path_coverage_test` — 3 passed
- Merged `origin/main` (through #7881, which also touches `ephemeral_spawn.rs`) and re-ran all of the above green.

New tests, by the claim each makes:

| Test | Claim |
| --- | --- |
| `a_completed_ephemeral_run_is_visible_under_its_parent` (api) | the run is reachable over HTTP under its parent, **and** the mission workspace is still gone |
| `an_agent_with_no_ephemeral_runs_gets_an_empty_list` (api) | a parent that spawned nothing is unaffected and sees no one else's runs |
| `the_limit_parameter_caps_rows_without_capping_the_rollup` (api) | paging does not distort the cost rollup |
| `unknown_and_invalid_agent_ids_are_refused` / `the_route_requires_authentication` (api) | 400 / 404 / 401 match the sibling observability routes |
| `a_finished_ephemeral_run_is_recorded_under_its_parent` (kernel) | the three properties at the kernel layer: reachable, workspace gone, bystander untouched |
| `a_run_record_and_its_usage_row_agree_on_the_owner` (kernel) | the ledger and the run list cannot disagree about who owns a run |
| `removing_a_parent_removes_its_ephemeral_run_records` (kernel) | the cascade |
| `a_failed_run_leaves_no_mission_workspace_behind` (kernel, extended) | a refusal *before* the loop fabricates no record — the boundary of "a record exists iff a run started" |
| store + migration unit tests | trim cap, trim scoping, rollup arithmetic, status rejection, `char`-boundary truncation, table shape, audit-row regression |

## Not done here

`openapi.json` / `sdk/` / `xtask/baselines/` are not regenerated in this PR — the `OpenAPI Drift` job auto-commits them back to same-repo PR branches, and this is a same-repo branch. The `dead_route_audit` and `openapi_path_coverage` guards both pass, so the annotation and the route registration agree.

No dashboard UI. The issue's proposal 3 asks for a dashboard/TUI view; this PR lands the record and the API it needs, and the SPA work is a separate crate and a separate review. Happy to follow it up if you want it in the same release.


crates/librefang-kernel/src/kernel/ephemeral_spawn.rs
```diff
@@ -546,4 +607,46 @@ impl LibreFangKernel {
...
+            .list_for_parent(&parent_id.0.to_string(), limit)
+            .map_err(KernelError::LibreFang)
+    }
+
+    /// Aggregate spend and run count across an agent's retained ephemeral runs.
+    pub fn ephemeral_run_rollup_for_agent(
+        &self,
+        parent_id: AgentId,
+    ) -> KernelResult<librefang_memory::EphemeralRunRollup> {
+        librefang_memory::EphemeralRunStore::new(self.memory.substrate.pool())
```
houko, line 648: `ephemeral_runs_for_agent` / `ephemeral_run_rollup_for_agent` (here and below) have no production caller — `observability.rs`'s `list_agent_ephemeral_runs` handler constructs its own `EphemeralRunStore` from `state.kernel.memory_substrate().pool()` directly rather than going through these kernel methods. That mirrors the existing `get_agent_stats` pattern, so it's not a new anti-pattern, but it does leave these two as test-only public API surface. Worth a maintainer call: should the route call through these for symmetry, or should they be dropped as unused indirection?

---

crates/librefang-api/tests/agent_ephemeral_runs_test.rs
```diff
@@ -0,0 +1,276 @@
...
+    .await;
+    assert_eq!(status, StatusCode::NOT_FOUND);
+}
+
+/// The route sits behind the same auth layer as the rest of `/api/agents`.
+///
+/// A run record carries the delegated task and the worker's answer verbatim, so an unauthenticated read would leak conversation content, not just counters.
+#[tokio::test(flavor = "multi_thread")]
+async fn the_route_requires_authentication() {
+    let h = boot().await;
```
houko, line 263: This tests 401-when-unauthenticated, but there's no test for cross-owner denial (agent B requesting agent A's runs while authenticated as a non-admin B). The route's own doc comment argues this endpoint needs owner-scoping *more* than `/stats`/`/events` because it carries verbatim conversation content — worth adding the one test neither of those siblings has either: a second author gets 404, not the run records. Not a regression (siblings lack it too), but given the PR explicitly foregrounds the stronger privacy stakes here, closing the gap seems worth it.

---

crates/librefang-api/src/routes/agents/observability.rs
```diff
@@ -523,3 +523,170 @@ pub async fn agent_logs(
...
+    pub output_tokens: u64,
+}
+
+#[derive(Debug, Clone, serde::Serialize, utoipa::ToSchema)]
+pub struct EphemeralRunsResponse {
+    pub runs: Vec<EphemeralRunView>,
+    pub rollup: EphemeralRunRollupView,
+}
+
+/// GET /api/agents/{id}/ephemeral-runs — what this agent delegated to ephemeral workers, and what each one cost (refs #7752).
```
houko, line 598: `EphemeralRunRollupView` sums only the retained rows (capped at `MAX_RUNS_PER_PARENT = 200`), which is documented in the store/route/migration comments but not signaled in the wire response itself (no `truncated`/`window` field). An operator diffing this rollup against the all-time `usage_events` total for the same agent could read the mismatch as a bug rather than the intended trailing-window design. Not blocking since it's a disclosed, deliberate tradeoff — just flagging in case a future dashboard consumer needs to surface the distinction to end users.

---

houko (commented):
Automated review against CLAUDE.md.

Pushed a mechanical fix (commit `353384b`, rebased to `6d4c86b`): reflowed the new doc comments and inline explanatory comments across `ephemeral_spawn.rs`, `migration.rs`, `substrate.rs`, `ephemeral_run_store.rs`, `observability.rs`, `kernel/tests.rs`, and `agent_ephemeral_runs_test.rs` to one sentence per line, per CLAUDE.md's "Prose wrapping" rule (new prose must break only at sentence boundaries, not at a column width). Verified `cargo check` and `cargo clippy -D warnings` clean on `librefang-memory`, `librefang-kernel`, `librefang-api` afterward.

Otherwise this PR is solid: the route requires auth like its `/stats`/`/events` siblings, no HashMap/HashSet reaches a prompt, `MAX_RUNS_PER_PARENT`/`MAX_TEXT_LEN` are deliberately compiled constants (not half-done config wiring), there's a real `#[tokio::test]`-against-router integration suite, no new circular dependency, the char-boundary truncation and v51 migration DDL are correct, the cascade delete is transactional, and the `migrate_v50` audit-row fix is correct and doesn't disturb already-migrated databases.

Left three inline comments on things worth a maintainer's judgment call rather than a mechanical fix (unused kernel wrapper methods, a missing cross-owner-denial test, and the rollup's undisclosed retention-window field) — none of them blocking.

---

houko:
Auto-merge disarmed deliberately, to force a merge order — nothing here is wrong.

**Three open PRs each claim a migration number and two of them claim the same one.** This PR and #7916 both take **v51**; #7919 takes **v52**.

Ordering matters more than it looks. `run_step!(N, f)` fires on `current_version < N` where `current_version` is read once at the top of `run_migrations`, so a database that reaches N will never run a migration numbered below N added later. And the ladder has to stay contiguous: `migration::tests::test_every_migration_records_audit_row` asserts a fresh ladder needs no audit-row backfill, so skipping a number fails the build. I confirmed that by renumbering this branch to v53 with 51 and 52 absent — `left: 2, right: 0`, exactly that test.

So the three have to land in ascending order, and the last one's number is only knowable once the earlier two are on `main`. Plan:

1. **#7916 merges first, keeping v51.** Auto-merge is armed there.
2. **#7919 second, at v52.** Disarmed until #7916 lands.
3. **This PR last, renumbered to v53** at that point — not now, because renumbering ahead of the base would leave a 50 → 53 gap and turn this PR red on the contiguity test for no reason.

Nothing is required of this PR until then. When #7919 is on `main`: merge `origin/main`, renumber `migrate_v51` → `migrate_v53` (`migration.rs` and the doc reference in `ephemeral_run_store.rs`), re-run `cargo test -p librefang-memory --lib`, and re-arm.

I introduced this by arming three migration-carrying PRs for auto-merge at once, where merge order was going to be decided by whichever CI run finished first.

houko:
Verification update while CI is unavailable — this branch was checked locally against current `main` (`3624319cc`).

- Merges clean into `main`.
- `cargo test -p librefang-memory --lib` — **404 passed, 0 failed**.
- It still declares `SCHEMA_VERSION = 51`.

That last point is the thing to watch. **v51 is only correct if this merges before #7916**, which also claims v51. #7916 has auto-merge armed and this one does not, so in practice #7916 lands first and this branch then has to become **v53** (with #7919 taking v52) — that is the ordering I documented earlier and it has not changed.

If the order ever flips and this lands first, no renumber is needed. Either way it must not merge with a number another merged migration already used: `run_step!(N, f)` fires on `current_version < N` read once at the top of `run_migrations`, so a database that reaches N never runs a lower-numbered migration added afterwards, and `migration::tests::test_every_migration_records_audit_row` additionally requires the ladder to stay contiguous.

Auto-merge stays off here until #7916 and #7919 are on `main`.

houko:

PR #7915, opened 2026-08-25
feat(memory): give episodic memory an exit — TTL plus a per-row size cap

### LINKED ISSUE #7911: Episodic memory has no exit: written every turn, never distilled, never expires, attachments inlined
Splitting §1.2 out of #7756 as you suggested, so the retention question can move without waiting on the seam decision. This is the part where the remaining work is a choice rather than a fix, so it names the choices instead of leaving them open.

## Symptom

The episodic layer has no exit. It is written unconditionally, never distilled, never expires, and inlines attachments whole.

Measured on one production instance, full corpus of 975 rows accumulated over about 100 days:

| class | rows | share |
|---|---|---|
| raw dialogue (`[Past exchange]`, plus the legacy `User asked:` shape) | 769 | 78.9% |
| extracted facts, 12 categories | 151 | 15.5% |
| imported | 55 | 5.6% |

Write rate on a normal day is 20 raw rows to 3 facts. The largest single row is **201 765 characters** — a PDF attachment stored verbatim through `[Past exchange]\nThem: [Attached PDF: …]` — and it had been retrieved 63 times.

## Root cause

**Written unconditionally.** `crates/librefang-runtime/src/agent_loop/end_turn.rs:269` writes one row per turn, gated only on `!is_fork && !incognito`. No `auto_memorize` gate, no `extraction_threshold`, no category — `HashMap::new()` for metadata, scope `episodic`.

**Never an input to extraction.** `auto_memorize` reads the live turn, not the stored rows — `end_turn.rs:333`, `&ctx.session.messages[end_turn.new_messages_start..]`. The two writers are parallel consumers of the same turn, so the arrow from episodic to semantic that the architecture is usually described as having does not exist in code.

**Never expires.** Nothing has a TTL on scope `episodic`. `evict_if_over_cap` (`crates/librefang-memory/src/proactive.rs:1133`) now prefers raw dialogue, which is the right order, but it only runs at the cap.

**Auto-Dream is not a transform.** It is an agent turn prompted to use `memory_recall` / `memory_store`. A control run on this instance produced +2 rows and zero deletions, against a write rate of 20 rows a day.

## What the field does here

Not to argue from authority, but because the shape is settled elsewhere and the divergence is specific.

The classical distinction is Tulving's: episodic memory holds events with time and place, semantic memory holds decontextualised knowledge. The part usually dropped in implementations is that **semantic memory is derived from episodic, and the episodic trace then fades**. You remember that Paris is the capital of France without remembering the lesson. Consolidation without forgetting is not the model; it is half of it.

Current systems land the same way:

- mem0 separates lifetimes by tier — conversation, session, user — and describes promoting relevant episodic content into semantic memory, with TTL on the long tail to bound storage.
- Generative Agents (Park et al., 2023) pair the observation stream with a reflection step that synthesises higher-order statements, and retrieve on recency × importance × relevance rather than similarity alone.
- Consolidation-routing systems (MIRIX, EverMemOS) route content into typed stores and cluster related entries upward.

What LibreFang already has that these have: per-turn extraction with an ADD/UPDATE/DELETE/NOOP decision, which is mem0's design and works well here — 151 correctly categorised facts, zero extraction failures in the logs. What it does not have is the second half: nothing consolidates, and nothing retires.

## The decisions, since that is what is blocking

Each of these is a policy call rather than a defect, which is presumably why none has been made. Stating the trade rather than the preference.

**1. Does episodic get a TTL, and is it time-based or capacity-based?**
Time-based bounds growth predictably and is one migration plus one query. Capacity-based (keep the last N per agent) never surprises an operator with a sudden gap but leaves the tail unbounded in bytes. The 201 765-character row argues that bytes, not rows, are the quantity that hurts.

**2. Does distillation happen, and where?**
Extending Auto-Dream is cheaper — it already walks memory and has a model attached — but it is agentic, so it is best-effort and its throughput is unpredictable, currently two orders of magnitude below the write rate. A deterministic pass is predictable and costs an LLM call per batch. Doing neither is defensible if TTL alone is judged enough; then the position is "episodic is a short-lived buffer and facts come only from the live turn", which is coherent and should be written down.

**3. Is the write gated at all?**
Right now every turn is stored. Gating on the same signal the extractor uses would cut volume at the source, but it also means a turn that produced no fact leaves no trace, which removes the "what did we talk about" capability the layer nominally provides. Alternatively gate only on size, which would have caught the PDF.

**4. Attachments.**
Inlining a 200 KB document into a memory row is the single largest contributor here. Storing a reference plus a short extract changes what the row is for. This one seems least controversial to me but it does change the shape of the record.

**5. Does `confidence` stop being mutable?**
Related but separable, and it interacts with the seam RFC: if decay is derived from elapsed time at read rather than written back, the field becomes extractor certainty only, and lifecycle stops being state a provider has to own. @samvallad33 made this point on #7756 and it is the same one Part 2 of that RFC was making.

## What I am not claiming

The retrieval side of this is measured and does **not** support "raw dialogue crowds facts out of the prompt". Controlling for age with a single-month cohort — 98 raw rows against 93 facts — median retrieval count was 29.5 for raw against 110 for facts. The ranker prefers facts by roughly 3×. Raw dialogue dominates the retrieved set because there is 5× more of it, so this is a corpus-composition problem and not a ranking one. Fixing retention is the lever; reweighting retrieval is not.

## Offer

Happy to implement whichever way these are called, and to test against a live instance with about a thousand real rows — that instance is where every number above comes from. Corpus statistics are shareable even though the corpus is not, if the distributions would help anyone else reason about the trade.

Refs #7756

houko:
Closes #7911

## Premise, verified against the tree

Every claim in the issue holds on `origin/main`.

- The per-turn write is unconditional. `crates/librefang-runtime/src/agent_loop/end_turn.rs:257` gates only on `!ctx.opts.is_fork && !ctx.opts.incognito`, then composes `[Past exchange]\nThem: …\nYou: …` and calls `remember_interaction_best_effort` (`crates/librefang-runtime/src/agent_loop/prompt.rs:66`), which writes scope `"episodic"` with `HashMap::new()` metadata straight to `MemorySubstrate`.
- Episodic is never an input to extraction. `end_turn.rs:330` passes `&ctx.session.messages[end_turn.new_messages_start..]` to `auto_memorize` — the live turn, not the stored rows. The two writers are parallel consumers of the same turn.
- Nothing expired it. `crates/librefang-memory/src/decay.rs::run_decay` named exactly two scope literals, `"session_memory"` and `"agent_memory"`, and `MemoryDecayConfig` carried exactly two TTLs. `episodic` is the table default (`crates/librefang-memory/src/migration.rs:397`) and appeared in neither.
- `ConsolidationEngine::consolidate` (`crates/librefang-memory/src/consolidation.rs:95`) is not a distiller: it multiplies `confidence` by a decay factor and merges near-verbatim duplicates by Jaccard overlap. It never deletes by age and never summarises.
- The per-agent cap does not cover the per-turn path at all. `evict_if_over_cap` (`crates/librefang-memory/src/proactive.rs:1133`) lives on `ProactiveMemoryStore`, and the per-turn writer bypasses that store entirely — the cap only runs on the turns where the extractor also stored something.
- No size bound existed anywhere. `sanitize_for_memory` strips channel envelopes and trims whitespace; it does not truncate.

## Changes

**Episodic TTL.** `MemoryDecayConfig` gains `episodic_ttl_days`, default 90.
`run_decay` sweeps it on the same soft-delete path as the other two scopes, and the existing daily retention sweep (`prune_soft_deleted_memories`) hard-deletes those rows and reclaims their embedding BLOBs.
The TTL is measured from `accessed_at`, which `recall_with_embedding` refreshes, so a row that keeps being retrieved keeps living — the sweep retires the tail, not the working set.
90 days is deliberately longer than the AGENT default of 30 because an episodic row is often the only record that a conversation happened.
`0` disables expiry for this scope only.

**Per-row size cap.** `MemoryConfig` gains `max_episodic_chars`, default 8000, `0` disables.
`budget_interaction_halves` (`agent_loop/message.rs`) splits the budget across the two halves of the exchange: a side that fits inside its half keeps everything and releases the remainder to the other side, so a 200 KB user message can never truncate the agent's reply away.
Cuts land on `char` boundaries and append `… [truncated]`.
It is applied in `end_turn.rs` *before* the embedding call, so it bounds embedding cost as well as stored bytes — the 201765-character row in the issue was also a 201765-character embedding request.

**Plumbing.** The cap is read off `MemorySubstrate` (`set_max_episodic_chars` / `max_episodic_chars`, set once in `kernel/boot.rs` next to `set_consolidation_duplicate_threshold`) rather than threaded through `LoopOptions`, because the runtime's per-turn writer already holds the substrate and `[memory]` is restart-required in `build_reload_plan`, so a value read once at boot cannot go stale.

**`decay_scope` helper.** The two copy-pasted per-scope blocks in `run_decay` became one function rather than three copies. Behaviour for SESSION and AGENT is unchanged — same SQL, same `datetime()` wrapping, same malformed-`accessed_at` warn.

**Docs.** `docs/architecture/episodic-memory-lifecycle.md` states what the layer is for, what removes rows from it, and — as the issue asked — writes down explicitly that distillation and write-time relevance gating are *not* implemented, so the honest description of the layer is "a bounded, self-expiring buffer of recent exchanges".

## Verification

Nothing was run against a live daemon.

`cargo check --workspace --lib` — clean.
`cargo clippy --workspace --all-targets -- -D warnings` — exit 0, zero warnings.

`cargo test -p librefang-memory --lib decay` — 24 passed, 0 failed. New tests:

- `decay::tests::episodic_memories_expire_after_their_ttl`
- `decay::tests::episodic_ttl_is_measured_from_last_access_not_creation` — a row created 400 days ago but accessed yesterday survives, which is the property that makes a time-based policy safe here
- `decay::tests::episodic_ttl_zero_disables_only_the_episodic_scope` — the knob is per-scope, not a master switch
- `decay::tests::expired_episodic_rows_are_hard_deleted_by_the_retention_sweep` — decay soft-deletes, the retention sweep reclaims the row and its BLOB, so the exit is complete rather than half-built

`cargo test -p librefang-memory --lib` — 399 passed, 0 failed, 0 ignored (whole crate, after the additions above).

`cargo test -p librefang-memory --lib substrate::tests::substrate_episodic` — 2 passed, 0 failed. These sit at the injection site rather than only at the implementation site, which is the failure mode CLAUDE.md warns about — a default that compiles fine and silently disables the feature:

- `substrate::tests::substrate_episodic_cap_default_matches_the_configured_default` — the substrate's compiled fallback and `MemoryConfig::default().max_episodic_chars` are in different crates and nothing but this assertion stops them drifting
- `substrate::tests::substrate_episodic_cap_round_trips_through_the_boot_setter` — including that `0` survives as `0` rather than being coerced back to the default

`cargo test -p librefang-types` — 5 / 49 / 3 / 4 / 2 passed across its suites, 0 failed; this is what covers `MemoryDecayConfig` in `config_default_roundtrip`.

`cargo test -p librefang-runtime --lib budget_interaction` — 5 passed, 0 failed:

- `agent_loop::tests::budget_interaction_halves_zero_cap_is_unbounded`
- `agent_loop::tests::budget_interaction_halves_under_budget_is_untouched`
- `agent_loop::tests::budget_interaction_halves_gives_a_short_reply_its_full_length` — the 201765-character attachment case from the issue; the reply survives whole
- `agent_loop::tests::budget_interaction_halves_splits_evenly_when_both_sides_are_large`
- `agent_loop::tests::budget_interaction_halves_cuts_on_char_boundaries`

`cargo test -p librefang-api --test config_schema_golden` — 1 passed, 0 failed. The fixture was regenerated and the diff is exactly the two new fields plus their defaults and doc text, reviewed in the commit.

No migration. Both fields are config-only, and `[memory]` is already classified restart-required as a whole struct in `build_reload_plan`, so `docs/operations/config-reload.md` needs no new row.

## Deferred, with reasons

- **Distillation** (issue decision 2). A deterministic episodic→semantic pass costs an LLM call per batch and needs an `AuxTask` slot, a cheap-tier fallback and a financial-DoS guard of the kind `SkillWorkshopReview` already has. That is a feature in the kernel's aux-task domain, not a retention fix, and it should not ride in on a PR whose job is bounding growth. The doc added here states the current position explicitly so it is a written-down choice rather than an omission.
- **Write-time relevance gating** (issue decision 3). Gating on the extractor's signal removes the "what did we talk about" capability the layer nominally provides, which is a product call for the maintainer, not an implementation detail. The size gate — the half of decision 3 the issue calls least controversial — is implemented here.
- **`confidence` becoming immutable** (issue decision 5). Explicitly cross-cutting with #7756's seam decision; changing it here would pre-empt that RFC.


nevgenov:
Ran the TTL default against the instance, since you asked for the distribution. 2084 live episodic rows, 1.91 M characters, one installation, several agents, roughly five months of history.

**The protection works as you described.** Every one of the 466 rows older than 90 days by `created_at` has genuinely been recalled — median `access_count` 26. The 101 rows that were never recalled all have `accessed_at == created_at` and a maximum age of 88 days, so they are young rather than overlooked. On this corpus `accessed_at` is tracking real use, and measuring from it does retire the tail rather than the working set.

**It retires in bursts, not on a gradient.** TTL from `accessed_at`, same corpus, walking the threshold:

| threshold | rows retired | characters |
|---|---|---|
| 88 days | 367 | 212 000 |
| 89 days | 0 | 0 |
| 90 days | 0 | 0 |

The cliff is one timestamp: 365 rows share `accessed_at` to the same second. So at the 90-day default this corpus retires nothing today and 367 rows the day after tomorrow, in a single sweep — each of those a soft-delete plus, later, a BLOB reclaim. Not a correctness problem, but if the sweep is unbatched it is worth knowing that the realistic shape of the work is one large burst rather than a trickle.

I could not identify what wrote those 365 rows in one second. It is not dashboard polling — the bulk listing paths go through `recall_readonly` (#5839), and these rows carry high access counts rather than being untouched. A maintenance operation at some point is the likeliest explanation and I cannot rule it in or out from the data alone. Flagging it because if such bulk touches happen routinely, they postpone every affected row by a full TTL, and that would matter for the policy in a way this single snapshot cannot show.

**Steady state, extrapolating past the cliff:** about 550 rows and 326 000 characters, roughly 17 % of the episodic bytes.

**For comparison, the size cap does the same magnitude of work from 8 rows.** Eight rows exceed 8000 characters and carry 274 935 characters above it — 14.4 % of the entire episodic corpus, including the 201 765-character row from the issue. TTL needs 550 rows and three months to reclaim what the cap reclaims from eight rows immediately, and the cap also bounds the embedding request, which the TTL cannot do retroactively.

On the 90-day default itself I do not think this corpus argues for changing it. The number that would argue is the shape of the `accessed_at` distribution on an installation that is older than five months, and mine is not.

Happy to rerun after upgrading, which would also produce the census for #7916.

Refs #7911

PR #7916, opened 2026-08-25
fix(memory): stamp stored vectors with the model that produced them

### LINKED ISSUE #7912: The configured embedding model is an unmeasured per-deployment bet, and changing it silently invalidates stored vectors
Filing the embedder-observability piece from #7756 §1.4 as you asked, with the measurements linked back there.

The short version: the configured embedding model moved retrieval quality more than any retrieval-method difference I could resolve, and there is no way for an operator to know that, no way to measure it on their own corpus, and no guard against changing it.

## What was measured

LLM-judged nDCG@10 over three real corpora. For each query the union of every arm's top-10 was shuffled so the judge could not see provenance, each pair graded 0–3; confidence intervals bootstrapped over paired per-query differences. Five runs.

The cosine arm is the one that matters, since that is what the runtime uses on the embedding path:

| corpus | `bge-m3` 4-bit (configured) | `multilingual-e5-large` (unquantized) | delta |
|---|---|---|---|
| agent-A memory, 978 rows, conversational | 0.6025 | 0.6984 | **+0.096** |
| agent-B memory, 749 rows, sales domain | 0.6187 | 0.6503 | +0.032 |
| documentation corpus, 718 chunks | 0.5873 | 0.5943 | +0.007 |

For scale: the method has a measurable noise floor of about 0.04. The embedder-independent lexical arms moved by −0.040 and −0.016 between two runs that changed only the vector arm, because the judged pool is the union of all arms. Every method-level difference I chased across five runs sat inside or barely outside that floor. The embedder swap cleared it outright on one corpus.

Worth noting the gain is strongly corpus-dependent: near-nothing on technical documentation, where exact terms dominate and lexical matching already does well, and large on conversational memory, which is reworded and where semantics carry the match.

## Three gaps

**1. Nothing measures this per deployment.**

Producing the table above required an out-of-tree harness reading a copy of the database and reimplementing the ranking path. An operator has no supported way to ask whether their configured embedder is costing them, and the daemon reports the embedding subsystem healthy either way — because it is healthy, in the sense of returning vectors of the right shape.

**2. Changing `embedding_model` silently invalidates every stored vector.**

`memories.embedding` is a bare `BLOB` (`crates/librefang-memory/src/migration.rs:494`) with no record of which model produced it, and I could find no re-embedding path. Both models here are 1024-dimensional, so the dimension guard never fires: cosine is computed between incompatible spaces and returns meaningless similarities, with no error anywhere.

That is the worst failure shape available — an operator edits one config line and retrieval quietly goes random for every pre-existing row.

**3. No prefix support, which strands about half the available gain.**

A large part of the current embedding landscape — the e5 family among them — is trained on an asymmetric `query:` / `passage:` pair. Core sends raw text; `passage:` appears nowhere in the tree.

Measured on agent-A, same queries, same judge, only the prefixes toggled:

| configuration | nDCG@10 |
|---|---|
| configured model | 0.6025 |
| `e5-large`, no prefixes — what a config swap gets today | 0.6499 |
| `e5-large`, with prefixes — unreachable | 0.6984 |

So the reachable half, +0.047, sits at the noise floor and I would not act on it. The other +0.049 is a within-model comparison on an identical corpus and query set, and it is behind a feature core does not have.

## Suggested shape

Ordered by cost, and the first is the one I would take regardless of the others.

**Stamp the model on the vector**, or on the store, and warn at boot when the configured model differs from what the stored vectors were built with. Turns a silent corruption into a startup message. Does not require re-embedding to exist — it only requires the operator to be told.

**Optional prefixes** in the embedder config, defaulting to empty so nothing changes for current models.

**A re-embedding path**, if a model change is meant to be supported at all rather than merely detected.

Measuring this on an operator's own corpus is a bigger question and belongs with the harness discussion on #7756, not here.

## Caveats

One judge model throughout, so systematic judge bias is uncontrolled. All three corpora come from one installation and share a language and a pipeline; the documentation corpus differs in text type but not in either. And this compares two models, not embedders in general — it may be about 4-bit quantization specifically rather than about `bge-m3`, which I did not separate.

Happy to run the same harness against any model pair that would inform the choice, and to test a patch on a live instance with about a thousand real rows.

Refs #7756

houko:
Closes #7912

## Premise, verified against the tree

Gap 2 of the issue — "changing `embedding_model` silently invalidates every stored vector" — holds exactly as written on `origin/main`.

- `memories.embedding` is a bare BLOB with no provenance. `crates/librefang-memory/src/migration.rs:494` is the only place it is declared, and no later migration adds a model column: the full set of `ALTER TABLE memories ADD COLUMN` sites is `embedding` (v3), `image_url` / `image_embedding` / `modality` (v15), `peer_id` (v16), `last_decayed_at` (v48) and `deleted_at` (v31). `grep -rn embedding_model crates/librefang-memory` returned nothing.
- The only guard was the length check. `cosine_similarity` (`crates/librefang-types/src/memory.rs:971`) returns `None` when the vectors differ in length, and `librefang-types` already has `test_cosine_similarity_length_mismatch` pinning that. It catches a dimension change and nothing else — two 1024-dimensional models slip straight past, which is exactly what the issue measured.
- No re-embedding path exists. `SemanticStore::update_embedding` can replace one row's vector, but nothing sweeps the store after a model change.
- Nothing warned. The daemon logs `Embedding driver configured from memory config` / `auto-detected` at boot and never compares the resolved model against what is already on disk.

One thing the issue does not say that this PR had to handle: **the configured string is not the model identity**. `kernel/boot.rs:1457-1550` substitutes `default_embedding_model_for_provider()` when `embedding_model` is left at either built-in placeholder, and `detect_embedding_provider()` picks the provider from the environment when none is configured. Only the pair resolved at boot identifies the vector space, so that is what gets stamped.

Gaps 1 (per-deployment retrieval measurement) and 3 (asymmetric `query:` / `passage:` prefixes) are not addressed here — see Deferred.

## Changes

**Migration v51.** `memories.embedding_model TEXT DEFAULT NULL`, plus `idx_memories_embedding_model` for the census and for any future re-embedding sweep. `NULL` means "written before this migration" and is treated as comparable everywhere — see below.

**Stamp on write.** `SemanticStore::remember_with_embedding_and_peer` records the active model on any row it writes with a vector. A text-only row is left unstamped: it has no embedding space to belong to, and stamping it would inflate the census with rows a model change cannot affect.

**Re-stamp on re-embed.** `update_embedding` now writes `embedding_model` alongside the new vector. Without this the repair path would be self-defeating: a row rebuilt in the current space would keep its old stamp and stay permanently excluded from scoring.

**Recall refuses to score a foreign vector.** `recall_impl` reads the stamp with the vector and leaves `similarity = None` for any candidate stamped with a *different* model. That is the same treatment a dimension-mismatched vector already received — it sorts to the bottom on the existing `NEG_INFINITY` sentinel, and a `min_similarity` floor drops it, because an unmeasured row cannot clear a measured minimum. One `WARN` per recall, not per row.

**Dedup withholds a foreign vector.** `get_embeddings_batch` no longer hands out vectors stamped with a different model. Its only caller is `ProactiveMemoryStore::find_duplicates`, which compares two *stored* vectors and merges the rows that score above the threshold — a merge is destructive, so those pairs drop to the existing Jaccard word-overlap fallback rather than being merged on a meaningless score.

**Boot census and warning.** `MemorySubstrate::embedding_model_census()` groups live vector-carrying rows by producing model, returning a `BTreeMap` so the rendered summary is byte-identical across boots for the same store. `kernel/boot.rs` sets the resolved identity, runs the census and emits one `WARN` naming every model in the store that is not the active one, with its row count, and what to do about it. A census failure downgrades to a warn and boot continues — provenance is observability and must not be able to stop a daemon starting.

**`NULL` is deliberately comparable.** Rows predating v51 keep being scored. The overwhelmingly common case is an operator who has never changed their model, for whom every historical row is valid; dropping their whole corpus out of retrieval on upgrade would be a far worse default than the risk it removes. The census reports them under `(unstamped, pre-v51)` so the size of that population is visible rather than assumed.

**Docs.** `docs/architecture/embedding-provenance.md` covers the failure shape, why the stamp is the resolved pair rather than the configured string, the write/recall/dedup lifecycle, and what is still uncovered.

## Verification

Nothing was run against a live daemon.

`cargo check --workspace --lib` — clean.
`cargo clippy --workspace --all-targets -- -D warnings` — exit 0, zero warnings.
`cargo test -p librefang-memory --lib` — 404 passed, 0 failed, 0 ignored.

New tests:

- `migration::tests::test_migrate_v51_adds_embedding_model_column_nullable_and_indexed` — column present, index present exactly once, a legacy-shaped INSERT reads back `NULL`, and re-running the step twice touches neither the index nor the data
- `semantic::tests::a_stored_vector_records_the_model_that_produced_it`
- `semantic::tests::a_text_only_row_is_not_stamped`
- `semantic::tests::no_configured_model_means_no_stamp`
- `semantic::tests::recall_does_not_score_a_vector_from_a_different_model` — two *identical* vectors written under two different models, so without the guard the stale row would score a perfect 1.0 and be indistinguishable from the fresh one
- `semantic::tests::a_similarity_floor_drops_vectors_from_a_different_model`
- `semantic::tests::a_pre_stamp_vector_is_still_scored`
- `semantic::tests::re_embedding_a_row_restamps_it_with_the_active_model`
- `semantic::tests::get_embeddings_batch_withholds_vectors_from_a_different_model` — the destructive dedup path; an unstamped pre-v51 vector stays available, a foreign-stamped one is withheld
- `semantic::tests::the_census_counts_live_vectors_by_producing_model`
- `semantic::tests::the_census_ignores_soft_deleted_rows`

**Migration number:** `SCHEMA_VERSION` goes 50 → 51. 48 (#7756), 49 (#7714) and 50 (#7808) are already on `main`. This is the only one of my two open memory PRs that needs a migration — #7915 (issue #7911) is config-only — so there is no collision between them.

No config field, so no `build_reload_plan` classification, no `docs/operations/config-reload.md` row and no golden-schema regeneration.

## Deferred, with reasons

- **A per-deployment retrieval-quality harness** (gap 1). The issue itself places this with the harness discussion on #7756 rather than here, and it is a new measurement subsystem — corpus sampling, an LLM judge, bootstrapped intervals — not a change to the memory store.
- **Asymmetric `query:` / `passage:` prefixes** (gap 3). This is a change to the embedding drivers in `librefang-runtime` / `librefang-llm-drivers`, a different domain from the store, and #7912 measures the reachable half of the gain at +0.047 against a stated noise floor of ~0.04 — the author says they would not act on it. Worth its own PR with its own measurement, not a rider on this one.
- **A re-embedding sweep.** The issue orders detection first and says so explicitly: "Does not require re-embedding to exist — it only requires the operator to be told." The pieces a sweep needs are in place after this PR — the stamp, the census, and an `update_embedding` that re-stamps — but the sweep itself needs a background job, a rate limit against the embedding provider and a progress surface, which is kernel scheduling work rather than store work.
- **The external HTTP vector store.** With `vector_backend = "http"` similarity is computed by the backend and the SQLite rows are only hydration, so provenance there belongs to the backend's own index and this stamp cannot enforce it. Noted in the doc.


crates/librefang-kernel/src/kernel/boot.rs
```diff
@@ -1562,6 +1574,41 @@ impl LibreFangKernel {
...
+                        .collect();
+                    if !stale.is_empty() {
+                        warn!(
+                            active_model = %identity,
+                            stale = %stale.join(", "),
+                            "Stored embeddings were produced by a different model than the one now configured. \
+                             Cosine similarity across two embedding spaces is meaningless, so those vectors are \
+                             skipped during vector recall. Restore the previous embedding_model, or re-embed \
+                             the affected rows."
+                        );
```
houko, line 1601: The `stale` computation includes `UNSTAMPED_EMBEDDING_MODEL` rows, but the warning text says those vectors "are skipped during vector recall" — which is false for that bucket. `migrate_v51`'s doc comment and the `a_pre_stamp_vector_is_still_scored` test both establish that a `NULL` stamp is deliberately treated as *comparable*, not skipped.

Concretely: on the very first boot after this migration ships, every existing deployment with historical memories will have 100% of its rows land in `(unstamped, pre-v51)`, which differs from `identity` by construction. This warning will fire on essentially every upgrade, every boot, telling the operator their vectors are being skipped when in fact none of them are — the opposite of what a provenance warning should do.

Filtering the stale list to exclude `UNSTAMPED_EMBEDDING_MODEL` (or wording the message to distinguish "unstamped, still scored" from "different model, skipped") would make the warning match what `recall_impl`/`get_embeddings_batch` actually do.

---

houko (commented):
Reviewed against `CLAUDE.md`'s conventions and read through the write/recall/dedup/census lifecycle. `cargo check -p librefang-memory --lib`, `cargo clippy -p librefang-memory --all-targets -- -D warnings`, and `cargo clippy -p librefang-memory --lib` (via a separate check) all came back clean in this environment (kernel/desktop crates couldn't be checked here — missing system `dbus-1` dev headers, unrelated to this diff).

One inline comment on the boot-time warning's false-positive on unstamped rows (see below).

One additional finding that isn't inline because it's in a file this PR doesn't touch: `crates/librefang-memory/src/consolidation.rs`'s `ConsolidationEngine::consolidate` (the periodic text-only dedup sweep, `merge_embeddings_weighted` around line 239, `UPDATE ... embedding = ?4` around line 262) confidence-weight-averages two rows' embeddings once `text_similarity` judges them duplicates, with no check that they came from the same embedding model and no re-stamp of `embedding_model` on the merged row. That's the same failure mode this PR closes everywhere else (recall, dedup batch fetch) — a cosine/average across two embedding spaces being treated as meaningful — except here it produces and keeps a fabricated blended vector under a stamp that no longer describes it. Worth a follow-up given it's a different file/domain from this PR's scope (store write/recall/dedup guards vs. the separate text-only consolidation engine), but flagging now since it's the same bug class.

---

houko:
Merge-order note: this PR takes migration **v51** and #7919 takes **v52**.

Because the audit backfill inserts a row for every version in `1..=user_version`, a database that reaches 52 has already claimed 51 and would never run a later-added `migrate_v51`. This PR therefore has to land first. Auto-merge is armed here and disarmed on #7919 until it does — nothing is required of this PR beyond going green.

houko:
Correction to my earlier note — the chain is three deep, not two.

#7904 also claims **v51**, the same number as this PR. Full order: **this PR (v51) → #7919 (v52) → #7904 (v53, renumbered at that point)**. Auto-merge is armed here and disarmed on both of the others; nothing is required of this PR beyond going green.

houko:

PR #7918, opened 2026-08-25
fix(ci): remove the duplicate timeout key that stops release-tag parsing

houko:
## The failure

Every branch pushed since `#7284` merged shows a red `release-tag` run: `conclusion: failure`, zero jobs, no log. That is GitHub Actions refusing to start the workflow, not a job failing inside it — and it reads like a broken test rather than a malformed file, which is why it survived a while.

`.github/workflows/release-tag.yml` declared `timeout-minutes` twice on the `tag` job:

```yaml
  tag:
    name: Create release tag
    if: github.ref == 'refs/heads/main'
    runs-on: ubuntu-latest
    timeout-minutes: 15      # from #7836, the blanket timeout pass
    environment: release-tag
    timeout-minutes: 10      # from #7284, a deliberate value
```

I introduced it. `main` had just gained `timeout-minutes: 15` from the blanket pass in #7836 while the #7284 branch was adding `timeout-minutes: 10` to the same job, and resolving that by keeping both lines is wrong in a way "keep both sides" usually is not.

## Why every check passed

`yaml.safe_load` silently keeps the *last* of two identical keys. So the file is valid YAML, `check-workflow-timeouts.py` saw a `timeout-minutes` and was satisfied, and `check-changelog-attribution.py` had nothing to say. Nothing in the repository could see a difference between this file and a correct one — only GitHub could.

## Changes

- Removed the `timeout-minutes: 15` line. `10` is kept because `scripts/tests/test_release_tag_workflow_safety.py:188` asserts exactly that value; `15` was the blanket default, `10` was chosen for this job.
- `scripts/check-workflow-timeouts.py` now parses with a `SafeLoader` subclass that raises on a duplicate mapping key instead of collapsing it. That guard already reads every workflow file, so this costs nothing and closes the hole rather than fixing one instance of it.

## Verification

- Full scan of all 48 workflow files on `origin/main` with a duplicate-key-strict loader: `release-tag.yml` was the only one. Nothing else from the same merge sweep is affected.
- `python3 scripts/check-workflow-timeouts.py` — `OK: all 136 workflow jobs declare timeout-minutes`.
- `python3 scripts/tests/test_release_tag_workflow_safety.py` — passes.
- `python3 -m pytest scripts/tests -q` — 55 passed, 27 subtests passed.
- **Negative control**: re-added the duplicate line and re-ran the guard:

```
check-workflow-timeouts: release-tag.yml has a duplicate key 'timeout-minutes' on line 40.
GitHub Actions refuses to start a workflow with a duplicate key, and reports it as a
failed run with no jobs. Keep one of the two lines.
exit=1
```

## Note

`release-tag` is not a required status check, so this was noise rather than a merge blocker — but it is noise on every single branch, and it masks a real red `release-tag` run if one ever happens.


scripts/check-workflow-timeouts.py
```diff
@@ -29,6 +33,29 @@
...
 
+class DuplicateKey(Exception):
+    """A mapping in a workflow file declares the same key twice."""
+
+
+class _StrictLoader(yaml.SafeLoader):
+    """``SafeLoader`` that refuses a duplicate mapping key instead of keeping the last one."""
+
+
+def _no_duplicate_keys(loader: _StrictLoader, node, deep: bool = False) -> dict:
```
houko, line 44: Non-blocking: `_no_duplicate_keys` fully constructs the mapping (and its values) before returning, whereas PyYAML's stock `SafeConstructor.construct_yaml_map` is a generator that yields the empty `dict` first and populates it after. That two-step form is what lets a mapping contain a self-referencing alias (an anchor on the mapping used again inside one of its own values). This constructor would raise on that shape instead of building it. GitHub Actions workflow YAML essentially never does that (anchors here are shared step/job blocks, not self-referential), and the PR's full 48-file scan came back clean, so this is very unlikely to bite in practice — flagging only because it's a real behavior narrowing versus the loader being replaced, not because I expect it to matter here.

---

houko (commented):
Reviewed against CLAUDE.md. Pushed two mechanical fixes directly to the branch (prose-wrap in the new docstring, changelog fragment filename convention); left one non-blocking implementation note inline.

---

PR #7919, opened 2026-08-25
feat(channels): bulk channel member enumeration, kept out of the channel_dm authorization set

### LINKED ISSUE #7086: [Feature] Per-user identity and targeted notification for multi-user channels (Slack focus)
### Description

On channels that many real users share (Slack groups; likewise Telegram/WhatsApp groups), agents cannot (a) know **who** is in the channel, and (b) proactively **notify a specific
user** (e.g. "your task is done") without posting to the whole group. We run LibreFang
agents in multi-user Slack groups and DMs; identity and targeted delivery are blocking us.

Concretely, three gaps (verified in code):

1. **No roster / `group_members` for Slack** — the Slack adapter never emits
   `group_members` metadata (only WhatsApp does), so the `group_members` tool has no data
   on Slack; the agent cannot answer "who is in this channel?".
2. **No targeted notification to one user** — `channel_send(channel, target, content)`
   refuses cross-chat dispatch (#6117): in a group the only valid target is the group
   itself. `notify_owner(reason, summary)` was meant as the escape hatch, but
   `owner_notice` in the reply envelope is **not routed by any sidecar adapter** (BC-01:
   Telegram/Discord/Slack ignore it; only the in-process whatsapp-gateway consumes it),
   so on Slack it is a no-op for delivery. There is no tool to send a message to a
   specific Slack user by id.
3. **Sender display name is the raw user id** — the adapter sets `user_name = user_id`
   verbatim and never calls `users.info`, so even with the `[display_name]:` prefix in
   DMs (#4666/#4776), on Slack the agent only sees the opaque id `U09…` unless operators
   hand-configure `[users]` entries.

**Use cases:** (1) request attribution — attach the requester's identity (id + name) when
creating a task in our external PM tool (Odin) so it notifies the right person; (2)
completion notification — DM the requester ("✅ your task is done — here's the result")
instead of spamming the group; (3) name-aware conversation — greet/reference users by
name, know department/email where configured.

**Proposal:**
1. **Roster** — Slack adapter fetches + caches `conversations.members` + `users.info` and
   emits `group_members` metadata (like WhatsApp), enabling the existing roster tool.
2. **Targeted delivery** — an explicit "deliver to user id X" path (allow `channel_send`
   to a per-user DM conversation owned by the bot, or route `owner_notice` to the session
   peer on adapters that can DM); builds on delivery-target presets #6506 and
   per-conversation routing #6127 / RFC #5671.
3. **Identity** — optional, config-gated `users.info` resolution with cache + fallback so
   `display_name` is human-readable; `[users]` config remains the authoritative override.

### Alternatives Considered

- **Reply only in the originating channel/thread** — works for DMs, but in groups the
  reply is visible to everyone and cannot target one member; insufficient for private
  task-completion notices.
- **Hand-configured `[users.<id>]` entries** — works for a handful of operators, but does
  not scale to a channel roster and still gives no runtime member list or live profile
  data; display names must be maintained manually and stay stale.
- **`notify_owner`** — intended escape hatch, but not routed by Slack (BC-01), so not an
  option on our primary channel.
- **Custom sidecar/webhook channel** — technically possible (a "different-channel"
  dispatch stays allowed), but it bypasses LibreFang's built-in identity model, needs our
  own user-id mapping, and is a workaround rather than a platform feature.

### Additional Context

Related issues/PRs:

- #6493 / #6497 — cross-user compacted-summary leak (fixed; per-user gating exists)
- #2217 — WhatsApp: all conversations shared one session (fixed; per-conversation direction)
- #5671 — RFC: channel inbound routing / per-conversation binding
- #6127 — per-conversation agent routing for multi-agent groups
- #6117 — cross-chat dispatch guard (why `channel_send` cannot DM a member)
- #6395 — DM/group policy envs are only read by the WhatsApp adapter
- #4776 — sender identity prefix in DMs (Slack still shows raw ids)
- #6897 + companion inbound-files request — file delivery to a specific user depends on
  the targeted-delivery capability above

houko:
Closes #7086.

This is gap 1's bulk half — the last piece left open after #7865, #7874 and #7901.
`channel_members` could answer "who has spoken here?" but not "who is in here?", because the per-sender roster upsert was the only thing writing rows and no adapter emitted a platform member list.

## The security argument for the split

The obvious implementation — fill `group_roster` from Slack's `conversations.members` — is the one thing that must not be done.
`channel_dm` uses exactly those rows as its authorization set (`tool_runner/channel.rs`, authorized against the current conversation's roster), so bulk-populating them would silently widen the set from "people this agent has interacted with" to "everyone the platform lists", letting an agent privately message a channel member who has never addressed it.
That is a privilege escalation, not a feature, and it would reopen #6117 through a side door.

**Shape chosen: a `source` column on the roster row (`observed` / `enumerated`), not a second table.**

- The natural key `(channel_type, chat_id, user_id)` is identical in both sets. A member who was enumerated and then speaks is one row changing classification — an `UPDATE` — rather than two rows in two tables that every read would have to `UNION` and de-duplicate with a priority rule. That implicit reasoning at each read site is exactly what the split exists to remove.
- With two tables the narrowing would be "which table you happened to query", which is the accidental kind of guarantee. With a column it is an explicit `AND source = 'observed'` living in one named method.
- `RosterStore::observed_members` is a separate method from `RosterStore::members`, not a boolean flag, so an authorization check cannot reach the wider set by defaulting an argument. The same split is carried up through `KernelHandle::roster_observed_members`, whose default impl returns nothing — the fail-closed direction.
- Promotion is one-way. Speaking upgrades an enumerated row to observed; `upsert_enumerated` pins `source = group_roster.source` on conflict so a channel-wide sweep can never demote someone back. Without that pin the control would switch itself off on the member-list TTL.
- An unrecognised `source` value parses as `Enumerated`, not `Observed` — an unreadable classification loses DM reachability rather than gaining it.
- `upsert_enumerated` also leaves `last_seen` alone on conflict. For an observed row that column means *last heard from*, and a retention sweep keyed on it must not read a channel-wide enumeration as everyone having just spoken.

## Changes

**Storage**

- `crates/librefang-memory/src/migration.rs` — **schema v52** adds `group_roster.source TEXT NOT NULL DEFAULT 'observed'`. Every pre-migration row defaults to `observed`, which is correct because the sender upsert was the only writer; reading them any other way would retroactively revoke `channel_dm` for every existing deployment.
- `crates/librefang-memory/src/roster_store.rs` — `RosterSource`, `RosterMember`, `upsert_enumerated`, `observed_members`; `members` now returns both sets with their classification.

**Kernel / bridge**

- `KernelHandle::roster_observed_members` (new, fail-closed default) alongside `roster_members`, implemented over the real store in `librefang-kernel`.
- `ChannelBridgeHandle::roster_upsert_enumerated`, implemented in `librefang-api/src/channel_bridge.rs`.
- `bridge.rs: upsert_enumerated_members_into_roster` — reads the `group_members` metadata that `extract_group_members` already parsed and, until now, dropped. Wired into both inbound paths next to the existing sender upsert. Capped at `MAX_ENUMERATED_MEMBERS_PER_MESSAGE = 1_000`: the sidecar wire is not a trusted source of list lengths, and one message must not become an unbounded run of SQLite transactions.
- The `extract_group_members` doc comment, corrected in #7901 to say the bulk list is dropped, is updated again now that it feeds the enumerated half.

**Tools**

- `tool_channel_dm` authorizes against `roster_observed_members`, never `roster_members`. Its refusal now names the classification so the model can pick a reachable member instead of retrying blind.
- `tool_channel_members` reports every member with its `source`, plus `observed_count` / `enumerated_count`, and a note saying which members `channel_dm` will refuse. Tool definitions updated to match — a model that assumed every listed member was reachable would otherwise discover the split one refusal at a time.

**Slack adapter** (`sdk/python/librefang/sidecar/adapters/slack.py`)

- `SLACK_ENUMERATE_MEMBERS` (**default off**), `SLACK_MEMBER_LIST_TTL` (1 h), `SLACK_MEMBER_LIST_MAX` (500).
- `parse_conversations_members`, `_MemberListCache`, `_lookup_channel_members` (paginates, caps, caches every exit path including failures with the short `NEGATIVE_TTL_SECS` cooldown), `_apply_enumerated_members` wired into both inbound envelope paths.
- Needs `channels:read` / `groups:read`.
- A sweep **never** issues `users.info`; names come from the identity cache `SLACK_RESOLVE_DISPLAY_NAMES` fills. Resolving 500 people who have never spoken would spend the workspace's entire per-method budget.
- Member ids are sorted and de-duplicated before being stamped, so the metadata is byte-identical across sweeps that page in a different order (#3298 — this list reaches an LLM prompt through `channel_members`).

## Privacy

Off by default, behind its own switch, in the same style as `SLACK_RESOLVE_DISPLAY_NAMES` — and for a strictly larger reason.
That knob changes how well the daemon names the handful of people who have addressed the agent; this one changes *how many people* it stores at all, from those who spoke to everyone the workspace lists in the channel, most of whom never interacted with it.
The two are independent: enumeration on with resolution off stores opaque ids, which is the least-data configuration that still answers the question.
Documented in `docs/src/app/integrations/channels/core/page.mdx` (new "Channel Member Enumeration" section) and `docs/src/app/agent/tools/page.mdx`.

## Verification

- `cargo test -p librefang-memory` — 399 unit + 2 integration pass, including 6 new `roster_store` tests: `enumerated_members_are_listed_but_not_observed`, `speaking_promotes_an_enumerated_member_to_observed`, `enumeration_never_demotes_an_observed_member`, `enumeration_does_not_touch_last_seen_of_an_existing_row`, `pre_migration_rows_default_to_observed`, `an_unknown_source_is_not_dm_authorized`.
- `cargo test -p librefang-kernel --test kernel_handle_contract_broader` — 8 pass, including the two new end-to-end tests below.
- `cargo test -p librefang-runtime --test tool_runner_channel_dm --test tool_runner_channel_members` — 10 + 5 pass.
- `cargo test -p librefang-channels --lib roster_write_path` — 12 pass, 6 new over the enumeration write path.
- `python3 -m pytest sdk/python/tests -q` — 2181 pass, 16 new over `conversations.members` parsing, the member-list cache, pagination, the cap, the default-off switch and the no-`users.info` guarantee.

### The security assertion, with its negative control

`enumerated_members_are_reportable_but_never_dm_authorized` (`crates/librefang-kernel/tests/kernel_handle_contract_broader.rs`) runs against a real `MemorySubstrate`, so it asserts over the store's SQL rather than over a mock agreeing with itself.

Replacing `AND source = ?3` with a no-op predicate:

```
---- enumerated_members_are_reportable_but_never_dm_authorized stdout ----
assertion `left == right` failed: channel_dm's authorization set must exclude a member who has only ever been enumerated
  left: ["U1", "U7"]
 right: ["U1"]

---- speaking_promotes_an_enumerated_member_and_a_later_sweep_does_not_revoke_it stdout ----
assertion failed: kh.roster_observed_members("slack", "C0DESIGN").expect(...).is_empty()

test result: FAILED. 6 passed; 2 failed
```

Filter restored: `test result: ok. 8 passed; 0 failed`.

The tool layer has its own control. Pointing `tool_channel_dm` at `roster_members` instead of `roster_observed_members`:

```
---- channel_dm_refuses_an_enumerated_member_who_has_never_spoken stdout ----
an enumerated-but-never-observed member must be refused: Delivered privately to U7 on slack; the rest of the conversation cannot see it. Message sent to U7 on slack

---- channel_dm_reaches_only_the_named_member stdout ----
  left: [RosterQuery { channel: "slack", chat_id: "C0DESIGN", set: "all" }]
 right: [RosterQuery { channel: "slack", chat_id: "C0DESIGN", set: "observed" }]

test result: FAILED. 7 passed; 3 failed
```

That first line is the escalation itself: a private message delivered to someone who has never addressed the agent.
Restored: `test result: ok. 10 passed; 0 failed`.
The mock records *which* read each call used, so the assertion fails on the wide read even where the refusal would still have happened for another reason.

## Migration numbering — needs coordination with #7916

This takes **v52** and skips v51, which #7916 has claimed.
The `run_step!` guard is `current_version < N`, so a gap costs nothing going up. But the audit-trail backfill at the end of `run_migrations` inserts a row for every version in `1..=user_version`, so a database that reaches 52 has already claimed 51 and would never run a `migrate_v51` added afterwards.
**Whichever of the two merges second must renumber to the next free version rather than filling the gap.** The reasoning is recorded in a comment at the `run_step!` site so it is not rediscovered later.

## Deferred

- **Roster retention is still unanswered** — the open question from #7874, restated in #7901, and this PR makes it larger rather than resolving it. Roster rows have no automatic expiry, and enumeration stores identities for people who never interacted with the agent at all. That belongs in its own issue with a retention policy (TTL on `last_seen`, a prune sweep, or an operator command), not bolted onto this one; the docs say plainly that enabling the switch is a personal-data decision. Not fixed here because a retention sweep is a scheduler/config concern rather than a channels one.
- Only Slack enumerates. Other adapters need no change to benefit — the `group_members` metadata contract is channel-agnostic and the bridge write path is shared, so any adapter can opt in by emitting the key.

## Drive-by

`migrate_v50` recorded its audit row as `VALUES (49, ...)`, so the ladder never wrote a description for version 50 (the #3538 backfill papered over it with a generic row). One-character fix in a file this PR already touches.


houko:
Auto-merge disarmed deliberately, to force a merge order — not because anything here is wrong.

This PR takes migration **v52** and #7916 takes **v51**. The audit backfill inserts a row for every version in `1..=user_version`, so a database that reaches 52 has already claimed 51 and would never run a later-added `migrate_v51`. If this merges first, #7916's migration is silently dead on every database that took this one.

So: **#7916 must land first.** Auto-merge is armed there and disarmed here. Once #7916 is on `main`, re-arm this one — v51 then v52 is the correct sequence and no renumbering is needed.

If #7916 is instead abandoned or renumbered, this PR should be renumbered to the next free version at that point rather than left holding a gap.

houko:
Correction to my earlier note — the chain is three deep, not two.

#7904 also claims **v51**, the same number as #7916. Full order: **#7916 (v51) → this PR (v52) → #7904 (v53, renumbered at that point)**.

I also confirmed the ladder must stay contiguous, not merely ascending: `migration::tests::test_every_migration_records_audit_row` asserts a fresh ladder needs no audit-row backfill, so a 50 → 53 gap fails the build (`left: 2, right: 0`). That is why #7904 keeps v51 for now and renumbers only once its base is known.

houko:
Removed the drive-by `migrate_v50` audit-version fix from this branch in `4149a0cbd`.

It duplicated #7925 exactly — the same single line — so whichever of the two merged second would have hit a conflict or landed as a silent no-op. #7925 owns the fix: it has an issue (#7924) and two tests (`no_migration_relies_on_the_audit_backfill`, `v49_and_v50_each_describe_their_own_ddl`) verified to fail against the old literal, where this branch carried it with no coverage and no mention in the body.

Nothing else here changes; `cargo test -p librefang-memory --lib migration::` still passes 42/42 after the revert.

houko:

PR #7921, opened 2026-08-25
feat(provisioning): declarative resource provisioning and include-aware config checksums

### LINKED ISSUE #6695: RFC: add managed configuration and declarative provisioning mode

houko:
Refs #6695. Lands the second half of @whatnick's RFC — declarative resource provisioning — plus the one design question left open that blocked the first half from being trustworthy.

## What was already on `main`, re-verified today

The RFC's status comment predates three merges. Checked against `origin/main` at `3624319cc`:

- **§4, the dashboard provenance surface** — landed in #7868 (`a0acb2c9`), which also settled the lock-or-classify decision on `change_password`, the sidecar-channel pair, the extension pair and the MCP routes under `mcp_runtime_store = "file"`.
- **§1 and §6, the Kubernetes artefacts** — landed in #7902 (`d520adbe`): `deploy/kubernetes/overlays/managed-config/` with a ConfigMap, a read-only mount outside `/data`, both env vars, a `checksum/config` rollout annotation, a rollback procedure, and a `kind` e2e job that asserts the 423.
- **The relocation defect** (writes landing in a file nothing reads) — fixed in #7886 via `config_path_boot`.

That leaves declarative provisioning, which did not exist in any form, and the `include` checksum question.

## Declarative resource provisioning

Managed mode locks `config.toml` as a whole. This is the same contract one level down, for resources that do not live in that file.

`LIBREFANG_PROVISIONING_PATH` points at a deployment-owned tree:

```text
/etc/librefang/provisioning/agents/researcher.toml
```

- **`crates/librefang-kernel/src/provisioning.rs`** — the scan, the state file, and `plan()`, a pure decision table over plain data so every case is testable without a booted kernel.
- **`crates/librefang-kernel/src/kernel/provisioning_ops.rs`** — `apply_provisioning()`, called once from `boot_with_config_at` after the registry restore and *before* the default-assistant fallback, so a deployment that declares its own agents does not also get an `assistant` it never asked for.
- The resource identifier is the manifest's `name`, not the file name. `AgentManifest::name` deserialises a missing key to `"unnamed"`, so the reconcile requires the key to be present rather than accepting the deserialiser's default — otherwise `nmae = "researcher"` provisions `unnamed` and the next such typo collides with it invisibly.
- Decision table: create / apply / **adopt** an existing agent under a declared name (`kubectl apply` semantics — id, session and history survive) / no-op on a byte-identical file / recreate one deleted out of band.
- `LIBREFANG_PROVISIONING_PRUNE` decides what happens to an orphan. The default **releases** it — provenance dropped, agent keeps running and becomes editable again — which is what makes removal reversible: putting the file back re-adopts the same agent instead of colliding with a tombstone. Only the exact word `delete` removes anything, so a typo can never be the reason an agent is deleted.
- **Failures never fail the boot.** A file that is unreadable, not UTF-8, not valid TOML, nameless or duplicate is recorded and skipped; the rest of the tree still applies. An unrecognised subdirectory (`channels/`, `workflows/`) is reported rather than ignored, so a deployment is told the kind is unsupported instead of believing it provisioned something.
- The tree is an **input, never a write target**: a provisioned agent's `source_toml_path` is deliberately not set to its declaring file, so `persist_manifest_to_disk` never retries a doomed write against a read-only mount — the failure mode the managed-mode migration write-back was fixed to avoid.

### Both variables are environment, not `KernelConfig`

Same reason `LIBREFANG_CONFIG_MODE` is: a setting that decides what may be written must not itself be writable through the surface it governs. A `[provisioning]` key would be settable from the dashboard in mutable mode, and a write that turns provisioning off is a write that unlocks every provisioned resource. No new config field, so no `build_reload_plan` classification and no `config_schema_golden` churn.

### The lock

`guard_provisioned_write` (`routes/mod.rs`) is the resource-level counterpart of `guard_config_write` — same `423 Locked`, same envelope, `code: "resource_provisioned"` because the remedy differs. `guard_provisioned_agent` (`routes/agents/mod.rs`) applies it to nine handlers: `PATCH /agents/{id}`, `DELETE /agents/{id}`, the `model` / `tools` / `skills` / `mcp_servers` / `channels` PUTs, `PATCH /agents/{id}/config`, `PATCH /agents/{id}/identity`, and per item inside `DELETE /agents/bulk` so one deployment-owned agent does not fail an operator's other deletes.

The guard is on the *definition*, not the agent. Suspend, resume, stop, message, sessions, files and every read stay open — the RFC's "operational actions and mutable runtime state remain usable" criterion, asserted in a test rather than assumed. Everything the tree does not declare stays fully mutable.

### `GET /api/provisioning/status`

Per-resource `source`, applied `checksum`, current `source_checksum`, `drifted`, `present`, plus `failures[]` — the only place a refused file survives after the boot log scrolls. No write route: a provisioning tree is deployment-owned, so it is rollout-only exactly as managed configuration is.

## Include-aware configuration checksums (RFC design question 6)

`GET /api/config/status` hashed the primary file's bytes alone, so with `include = [...]` an edit to an included file changed the effective configuration and left the checksum identical — a false negative for the operator using it to confirm a rollout landed, and the stated reason `scripts/check-k8s-manifests.py` banned `include` in a managed ConfigMap outright.

- With no include the digest is **unchanged, byte for byte**, so #7902's shipped annotation and its `kind` assertion keep matching.
- With includes it is the digest of `sha256sum` output over the closure in include order, reproducible as `(cd /etc/librefang && sha256sum config.toml extra.toml) | sha256sum`. A new `includes` field lists what contributed; `modified_at` now reports the newest mtime across the same set.
- The walk is tolerant where the loader is strict — this feeds a status endpoint, so an absolute, traversing, missing or circular include shortens the list rather than failing the request.
- The checker now **allows** an include whose target is another key of the same ConfigMap and computes the composite annotation, failing on the cases the daemon could not honour: a target the manifest does not render, a `/`-containing key, an absolute or `..` path, and an include in a file mounted with a `subPath`.

## Kubernetes and docs

The managed-config overlay now renders `agents/researcher.toml` into a `librefang-agents` ConfigMap, mounts it read-only at `/etc/librefang/provisioning/agents` (a ConfigMap exposes keys flat, hence the subdirectory), sets `LIBREFANG_PROVISIONING_PATH`, and carries a `checksum/provisioning` annotation. The tree is reconciled at boot only, so an edit that does not roll the pod changes nothing — `check_provisioning` in the manifest checker enforces that, and additionally fails the build on a declaration that is not valid TOML, declares no `name`, is not a `.toml` key, assigns a credential a literal value, or is mounted writable or at the wrong path.

`docs/operations/declarative-provisioning.md` is the operator contract; `deploy/kubernetes/README.md` gains a "Provisioned agents" section; `docs/operations/managed-config.md` gains the include-checksum semantics.

## Verification

Every command run locally, in this worktree.

- `cargo test -p librefang-kernel --lib` — **1685 passed, 0 failed**. Includes 15 new `provisioning::tests::*` (the plan decision table, sorted scanning, per-file failures, duplicate names, unsupported subdirectories, state round-trip and malformed-state degradation, the prune-policy resolution rule) and 5 new `config::tests::*` (`provenance_checksum_of_a_single_file_is_the_raw_byte_digest`, `provenance_checksum_covers_included_files`, `composite_checksum_matches_the_documented_sha256sum_pipeline`, `source_walk_skips_unsafe_and_missing_includes_instead_of_failing`, `source_walk_terminates_on_a_circular_include`).
- `cargo test -p librefang-api --lib` — 1101 passed, 0 failed.
- `cargo test -p librefang-api --test provisioning_test` — **13 passed, 0 failed** (new file). Named: `provisioning_is_disabled_and_inert_without_the_environment_variable`, `a_declared_agent_is_created_at_boot_and_reported_with_its_provenance`, `manifest_writes_to_a_provisioned_agent_are_refused_with_the_documented_shape` (all nine routes), `operational_routes_stay_open_on_a_provisioned_agent`, `a_runtime_created_agent_stays_writable_beside_a_provisioned_one`, `editing_a_declaration_after_boot_reports_drift_without_changing_the_applied_checksum`, `a_malformed_declaration_is_reported_without_failing_the_boot`, `a_second_boot_over_an_unchanged_tree_changes_nothing`, `a_changed_declaration_is_applied_to_the_existing_agent_on_the_next_boot`, `removing_a_declaration_releases_the_agent_under_the_default_prune_policy`, `removing_a_declaration_deletes_the_agent_under_the_delete_prune_policy`, `an_existing_agent_is_adopted_when_the_deployment_declares_its_name`, `a_provisioned_agent_deleted_out_of_band_is_recreated_on_the_next_boot`.
- `cargo test -p librefang-api --test config_managed_mode_test` — 14 passed (unchanged behaviour, including the checksum-unchanged-across-a-refused-write case).
- `cargo test -p librefang-api --test config_path_unification_test` — 2 passed. `--test agents_clone_bulk_integration` — 27 passed.
- Drift guards: `dead_route_audit_test` 2 passed, `openapi_path_coverage_test` 1 passed, `openapi_spec_test` 1 passed, `config_schema_golden` 1 passed (schema unchanged — no new `KernelConfig` field). `openapi.json` regenerated and `xtask/baselines/openapi.sha256` rewritten as plain `sha256sum openapi.json`.
- `kubectl kustomize deploy/kubernetes/{base,overlays/managed-config}` piped through `python3 scripts/check-k8s-manifests.py` — both clean (3 and 5 manifests). Each new guard was also exercised against a deliberately broken copy of the overlay and produced the intended message: stale `checksum/provisioning`, a declaration with no `name`, invalid TOML, a literal `api_key`, an include naming an unrendered key, an absolute include, and a `/`-containing include.
- Not run: `cargo nextest` (not installed on this host) and any live daemon — CI covers the former, the latter is human-only.

- **`channels/` and `workflows/` provisioning.** Both persist through surfaces this reconcile does not model — channels into `config.toml` itself, already covered by the whole-config lock, and workflows into a SQLite-backed registry with its own run state — so neither is a matter of pointing the same scan at another subdirectory. An unrecognised subdirectory is reported as a failure rather than silently skipped, so a deployment that tries one is told.
- **In-place reload of the tree.** Boot-time only, matching the rollout-only contract managed configuration settled on.
- **Field-level ownership inside `config.toml`.** The RFC's remaining open design question, now answered rather than deferred: ownership is expressed per *resource*, which is the granularity a deployment has actually asked for, and layering a second axis over the existing `WRITABLE_EXACT_PATHS` allowlist would produce a matrix whose interesting cases are all corners. Recorded in `managed-config.md`.

I have deliberately **not** used `Closes #6695` — see the issue comment for exactly what is and is not covered.


crates/librefang-kernel/src/kernel/provisioning_ops.rs
```diff
@@ -0,0 +1,266 @@
...
+                                    info!(
+                                        agent = %name,
+                                        "Pruned provisioned agent — its declaration left the tree"
+                                    );
+                                }
+                            }
+                            None => tracing::debug!(
+                                agent = %name,
+                                "Provisioned agent already gone; nothing to prune"
+                            ),
```
houko: `Action::Prune` counts `report.pruned += 1` and drops the key from `next` unconditionally, even when `self.kill_agent(entry.id)` returns `Err` (e.g. the agent is hand-owned and the kill guard refuses it, per the same guard `bulk_delete_agents`/`kill_agent` already special-case elsewhere in this PR).

Two effects of that:
- `GET /api/provisioning/status`'s `report.pruned` claims a resource was pruned when it wasn't — an operator watching that count after setting `LIBREFANG_PROVISIONING_PRUNE=delete` would see a wrong success count on a failed delete, with the actual error visible only in the boot log `WARN`.
- Because the key isn't carried into `next.resources` either, the state file loses this resource's provenance entirely. On the *next* boot, `plan()` sees `desired` without the key and `previous` without the key too — no `Prune`/`Release` action is generated for it at all, so a delete that failed once is never retried and the agent quietly becomes an ordinary untracked runtime agent (not reported in `failures[]`, not counted as `failed`).

Compare with the `Apply`/`Create` failure branch just above, which does push a `ProvisioningFailure` and bump `report.failed` instead of `report.applied`/`created`. Seems like `Action::Prune` should follow the same pattern: only increment `pruned` and drop provenance on success, and on `Err` push a `ProvisioningFailure` (bump `report.failed`) while keeping the previous provenance in `next` so the next reconcile retries the prune instead of silently giving up.

No test currently exercises a `kill_agent` failure during prune, so this path isn't covered either way — flagging as a question rather than pushing a fix, since the right retry semantics might be a deliberate call I'm missing.

---

houko (commented):
Reviewed the diff against CLAUDE.md's project conventions (deterministic ordering, config-field completeness, changelog fragment format, route registration/auth, environment-variable-not-config-field pattern for lock-deciding settings). Everything else lines up — one edge-case question left inline on the prune-failure accounting.

---

PR #7922, opened 2026-08-25
feat(memory): budget the prompt memory section per class

### LINKED ISSUE #7920: Prompt memory section is 92 % raw dialogue by characters: extracted facts are crowded out by row size, not by rank
Follows the episodic/semantic split from #7911. That issue asked what removes episodic rows; this one is about what they do while they are alive — they take 92 % of the prompt memory section by characters, and in roughly a third of turns the section contains no extracted fact at all.

Measured on one installation, agent-A: 999 live rows — 794 episodic, 198 `user_memory`, 7 `agent_memory`.

## Root cause

Recall on the prompt path does not distinguish memory classes. Both call sites pass a filter that leaves `scope` at its default:

```rust
// crates/librefang-runtime/src/agent_loop/prompt.rs:369 and :439
Some(MemoryFilter {
    agent_id: Some(ctx.session.agent_id),
    peer_id: ctx.sender_user_id.map(str::to_owned),
    ..Default::default()          // scope: None — episodic and extracted facts rank together
}),
```

The section is then filled greedily from that single ranked list (`prompt_builder.rs:760`), bounded by `MEMORY_BULLET_LIMIT = 10` (`:623`), `MEMORY_BULLET_MAX_CHARS = 500` (`:626`) and `MEMORY_SECTION_MAX_CHARS = 5000` (`:637`).

**Ranking is not the problem.** Episodic rows take 77.4 % of the slots against a 79.5 % share of the corpus — very slightly *under* their base rate, so they are not being over-ranked.

The problem is size. An episodic row averages 1167 characters because it inlines a whole exchange; an extracted fact averages 133. So at parity of slots the two classes are nowhere near parity of budget:

| class | share of corpus | share of slots | **share of characters** |
|---|---|---|---|
| episodic | 79.5 % | 77.4 % | **91.7 %** |
| extracted facts | 19.8 % | 22.5 % | 8.2 % |

Over 80 real user messages, replaying the production ranking against the stored vectors: **in 29 % of turns the top-10 is entirely episodic** — no extracted fact reaches the prompt. Median episodic share of a turn's section: 90 %.

## What a fixed budget buys

Same 5000-character budget, same 500-character bullet cap, same 80 queries. Judge grades every candidate 0–3; the pool is the union of all arms, shuffled, so provenance is hidden. Arms differ only in how the budget is divided between the two classes.

| policy | items placed | grade 0 | 1 | 2 | 3 | usefulness (2+3 only) |
|---|---|---|---|---|---|---|
| current — one mixed list | 10.0 | 5.8 | 2.1 | 0.5 | 1.6 | 5.80 |
| 30 % of budget to facts | 24.4 | 15.6 | 5.2 | 0.8 | 2.8 | 9.94 |
| 50 % | 28.6 | 18.8 | 6.1 | 0.9 | 2.9 | 10.43 |
| 70 % | 33.0 | 22.1 | 7.0 | 1.0 | 2.9 | **10.82** |
| facts only | 38.9 | 26.6 | 8.6 | 1.1 | 2.5 | 9.75 |

Splitting roughly doubles the number of directly-answering records that reach the prompt, 1.6 → 2.9, at no extra retrieval or token cost. Turns containing at least one grade-3 record go from 68 % to 85 %.

Two results here argue against the obvious over-correction:

**Dropping episodic entirely is worse, not better.** Facts-only maximises the naive sum of grades but is the weakest of the alternatives once grade-1 records are discarded, and it has the lowest grade-3 coverage (80 %). An individual episodic row grades *higher* than an individual fact — mean grade falls from 0.79 to 0.47 as the fact share rises. The gain is from fitting more small records, not from the classes differing in quality in the direction one might assume.

**The exact split is not resolvable from this data.** Running the identical seed twice moved the baseline arm by about 12 %, and the 30/50/70 spread (9.94 / 10.43 / 10.82) sits inside that. The finding is "divide the budget", not "divide it 70/30".

One number about the status quo worth stating plainly: of the 10 records in the current section, **5.8 are graded "not related" to the message that retrieved them**.

## Proposed fix

Allocate `MEMORY_SECTION_MAX_CHARS` per class rather than first-come from one ranked list — episodic rows can no longer consume the budget a fact would have used, because they are not drawing from the same pool. The split fraction should be configurable rather than compiled in, given that the measurement cannot justify a specific value.

This needs `scope` to survive to the point where the section is built; today the filter discards the distinction before `prompt_builder` ever sees it.

Deliberately not proposed: dropping episodic from retrieval, and distillation. The first is refuted above. The second is the open question from #7911 and this result does not bear on it — the gain here needs no LLM pass.

## Caveats

One installation, one judge model, and a single corpus whose class ratio (4:1 episodic) is itself a consequence of the unconditional per-turn writer — an installation with a different ratio would see a different effect size.

The larger limitation is that the judge graded records independently. If episodic rows carry narrative continuity that is worth more assembled than as separate bullets, per-record grading cannot see it, and that is a real argument against pushing the fact share high. It is also why "divide the budget" is the claim here rather than a specific division.

Happy to share the harness or rerun it against a different split, budget or corpus.

Refs #7911, #7872, #7910

houko:
Closes #7920

## ⚠️ Depends on #7914 — must merge after it

This branch is cut from `fix/7910-memory-section-budget`, not from `main`, so the diff below currently includes #7914's commits.
#7914 rewrites the exact constants and the exact charging logic this PR builds on: it reserves the truncation marker out of the budget, adds `MEMORY_KEY_DISPLAY_CAP`, charges the whole rendered line, and grows `MEMORY_SECTION_MAX_CHARS` from 5000 to 5730 via `MEMORY_BULLET_FRAMING_OVERHEAD`.
Basing on `main` would have produced a conflict in every hunk and silently reverted that accounting.
"70 % of the budget" here means 70 % of the same number #7914 charges against, framing included.

**Auto-merge is deliberately left off.** Merging ahead of #7914 would land its commits out of order through this branch.

## What the issue established

Measurement, analysis and the A/B are @nevgenov's, on one installation's 999 live rows and 80 real user messages.

Recall on the prompt path does not distinguish memory classes — both call sites leave `MemoryFilter.scope` at its default, so raw dialogue and extracted facts rank in one list, and the section is filled first-come from it.

Ranking was not the problem: raw dialogue took 77.4 % of the slots against a 79.5 % share of the corpus, slightly *under* its base rate.
Size was. A raw-dialogue row inlines a whole exchange and averages 1167 characters; an extracted fact averages 133.
So at parity of slots the character split is 91.7 % / 8.2 %, and in 29 % of turns the section is entirely raw dialogue — no extracted fact reaches the prompt at all.

The A/B, same budget and same per-bullet cap, judge-graded with provenance hidden:

| policy | items placed | usefulness (grades 2+3) |
|---|---|---|
| current, one mixed list | 10.0 | 5.80 |
| 30 % of budget to facts | 24.4 | 9.94 |
| 50 % | 28.6 | 10.43 |
| **70 %** | 33.0 | **10.82** |
| facts only | 38.9 | 9.75 |

Facts-only is *worse* than 70/30, so this is a split and not an exclusion.

## Changes

- **`librefang-types::memory`** — `memory_is_raw_dialogue(scope, source, metadata)` plus `MemoryFragment::is_raw_dialogue()`, with `EPISODIC_SCOPE` and `MEMORY_CATEGORY_METADATA_KEY` constants.
  The predicate is the same three-part test `SemanticStore::eviction_candidates` applies in SQL (scope `episodic`, source `Conversation`, no `category`), lifted into the shared crate so the prompt builder and the eviction cap cannot drift apart on what a class is. Null / empty-string / non-string `category` values are handled exactly the way `COALESCE(json_extract(...), '') = ''` handles them.
- **`prompt_builder`** — `format_memory_items_by_class` and `build_memory_section_by_class`.
  Facts fill first against `MEMORY_FACT_BUDGET_PERCENT` (70 %) of `MEMORY_SECTION_MAX_CHARS`; dialogue then fills against the *whole* section budget, so it inherits whatever facts did not spend; then any facts held back by their own ceiling get a second pass over whatever dialogue left in turn.
  Neither share can be wasted in either direction.
- **The total is unchanged.** Both classes charge one running counter against `MEMORY_SECTION_MAX_CHARS`, so the section costs exactly as many characters as before and #7914's hard cap stays hard. What changes is how many records those characters buy.
- **`MEMORY_CLASS_BULLET_LIMIT = 40`** replaces `MEMORY_BULLET_LIMIT = 10` on the class-aware path.
  Keeping the count at 10 would have handed most of the measured benefit back: at the observed fact size a 70 % share holds about 29 facts, and the winning arm placed 33 records. The character budget stays the binding constraint (`MEMORY_BULLET_MIN_CHARS` plus framing means the section can never hold more than ~70 bullets whatever this constant says); 40 is a legibility backstop, not a size bound.
- **`agent_loop::prompt::build_prompt_setup`** partitions the recalled fragments by class before building the section. `partition` preserves rank order within each class and the two classes are rendered contiguously, facts first — a fixed arrangement of a fixed input, so the section stays byte-identical across processes and the provider prompt cache still hits (#3298). A determinism assertion is part of the first test.
- **Config knob** — `KernelConfig.memory_fact_budget_percent: Option<u8>`, `None` → 70.
  The measurement establishes "divide the budget", not "divide it 70/30": rerunning the identical seed moved the baseline arm by ~12 % and the 30/50/70 spread sits inside that. A compiled-in constant would have overstated what the data supports.
  Plumbed through `LoopOptions.memory_fact_budget_percent` (the same shape `max_history_messages` uses) and populated at all five kernel `LoopOptions` sites. Classified NOOP in `build_reload_plan` (read live from `config.load()` per turn, so the ArcSwap swap is the whole reapply action), with the row added to `docs/operations/config-reload.md` and the schema golden regenerated.
- The class-blind `format_memory_items_as_personal_context` / `build_memory_section` stay, unchanged in behaviour, for callers that have no class information. Note that the kernel's `build_system_prompt` sites all pass `recalled_memories: vec![]` — recall happens in the agent loop — so the class-aware path is the live one.
- `remember_interaction_best_effort` now writes `EPISODIC_SCOPE` rather than a bare `"episodic"` literal, since it is the writer the classifier is defined against.

## The other half: recall never handed the section both classes

Raised in review, verified before acting on it, and it holds — with one more defect than suspected.

**The live branch is the context-engine one.** `boot.rs:1725-1744` binds `context_engine` as `Some(engine)` unconditionally (`build_context_engine` always returns a `Box`), and `tools_and_skills.rs:1874` hands it to every agent without a plugin allowlist. So `agent_loop/prompt.rs:339` is the branch production takes, not the substrate branches at `:431` / `:485`.

**`DefaultContextEngine::ingest` (`context_engine.rs:570-612`) recalls `max_recall_results` rows — 5 by default — class-blind.** The supplemental substrate fetch that widens that window was gated on `chat_scope_active || session_scope.is_some()`, so on a dashboard, REST or cron turn the engine's top five *was* the section's entire candidate pool.

**`prompt.rs:531` then capped class-blind** (`memories.truncate(MEMORY_RECALL_LIMIT)`, limit 5) after a fetch that was itself 5 when unscoped (`:334-338`).

On a store that is 4:1 raw dialogue the top five rows are raw dialogue — the same 29 % of turns the issue measured as reaching the prompt with no fact at all — so the budget split had nothing to place and spilled its fact share back to dialogue. Correct spill behaviour, and a no-op for the case the split exists to fix.

**And a third one, in this PR's own first commit:** `proactive_item_to_fragment` rebuilt the scope from `item.level`, but `MemoryItem::from_fragment` maps the scope through `MemoryLevel`, whose catch-all arm folds `episodic` into `Session` (`memory.rs:36-45`). A raw-dialogue row arriving through proactive memory came back labelled `session_memory`, and the class split — which reads the scope — would have filed it as an extracted fact and given it a fact-sized share of the budget. `auto_retrieve` recalls with no scope filter (`proactive.rs:2984-3020`), so it returns raw dialogue routinely; this was not hypothetical.

### What changed

- **Per-class candidate quotas** replace the class-blind `truncate`. `MEMORY_RECALL_LIMIT_DIALOGUE = 5` keeps the historical window for a class whose 30 % share renders about three rows; `MEMORY_RECALL_LIMIT_FACTS = 25` is sized to what the fact share can actually render (~29 bullets at the measured 133-character mean), so the character budget rather than the candidate window decides how many facts appear. No spill between quotas — the section's own fill already spills, and a wider dialogue window could not be rendered. **Neither class can lose:** a class-blind top-5 could never have contained more dialogue rows than the dialogue quota allows.
- **The supplemental substrate fetch is ungated** (still skipped in `stable_prefix_mode`), so the engine path gets a candidate pool wider than the engine's own class-blind five.
- **`MemoryItem` gains an optional `scope`**, set by `from_fragment` and used by `proactive_item_to_fragment`, which also restores the real `MemorySource` instead of hard-coding `Conversation`. Same shape `similarity` uses (#7808) for a value the conversion used to discard.
- `DEFAULT_SEMANTIC_SEARCH_LIMIT`'s doc comment claimed to match `MEMORY_RECALL_LIMIT`; that constant is now a per-class pair, so the comment says which half it tracks and why the tool limit was deliberately not widened with it.

### Retrieval cost — stated, not assumed

The issue's "no extra retrieval or token cost" was about an offline replay against stored vectors. It does not apply to this half, so here is the actual shape:

| | before | after |
|---|---|---|
| candidate window | 5 | 30 (5 dialogue + 25 facts) |
| `recall_fetch_limit` | 5, or 50 when scope-filtered | 120 |
| supplemental fetch on the engine path | scope-filtered turns only | every non-`stable_prefix_mode` turn |

What that costs in the store: **nothing on either ranked path.** `recall_impl` (`semantic.rs:323-337`) bounds the candidate scan by `MAX_FTS_CANDIDATES` (500) on the FTS path and `max(limit * 10, MAX_BRUTEFORCE_CANDIDATES)` = 5000 on the embedding path — 120 × 10 = 1200 is below 5000, so the scan is byte-for-byte what it was. Only the plain `LIKE` fallback, which applies `LIMIT limit` directly, goes from 5 to 120 rows. The added supplemental recall is one more local SQLite/vector query on turns that previously skipped it; every channel turn already ran it.

**#5227 is preserved with room to spare.** Its guarantee is that the fetch outruns the post-filter so a window of rows stamped for another chat cannot filter down to zero. The inflation is unchanged in form (`* 4`, floor 50) and the number it produces went up, not down: 50 → 120 on scoped turns. `engine_recall_widens_fetch_to_avoid_chat_scope_starvation_5227` and `recall_setup_drops_other_sessions_memories_7605` both still pass.

## Verification

- `cargo test -p librefang-runtime --lib` — **2414 passed, 0 failed**, run six times.
- `cargo test -p librefang-runtime --lib prompt_builder` — 101 passed (4 new).
- `cargo test -p librefang-runtime --lib agent_loop::prompt::tests` — 7 passed (4 new, 3 pre-existing #5227 / #7605 guards).
- `cargo test -p librefang-types --lib` — 979 passed (1 new).
- `cargo test -p librefang-kernel --lib config_reload` — 50 passed, incl. `every_config_field_is_reload_classified` and `doc_reload_table_matches_classified_reload_fields`.
- `cargo test -p librefang-api --test config_schema_golden` — golden regenerated and green.

New tests:

Section budget (`prompt_builder::tests`)
- `memory_section_class_split_admits_facts_a_mixed_list_crowds_out` — the class-blind formatter admits **zero** facts from a corpus of the issue's shape; the class-aware one admits all eight, keeps dialogue, does not interleave the classes, and is reproducible.
- `memory_section_class_split_stays_inside_the_total_budget` — 60 facts + 20 dialogue at `None/0/30/50/70/100/250`; never exceeds `MEMORY_SECTION_MAX_CHARS`.
- `memory_section_class_split_spills_an_unused_share_to_the_other_class`
- `memory_section_class_split_empty_on_both_classes_empty`, `fact_budget_percent_falls_back_and_clamps`

Candidate selection (`agent_loop::prompt::tests`)
- `recall_setup_carries_both_memory_classes_7920` — 40 raw-dialogue rows and 10 facts, with a deterministic two-axis embedding driver putting every dialogue row above every fact. Asserts the **candidate set**, not the rendered section, because that is where the loss happens.
- `recall_setup_carries_both_classes_through_the_context_engine_7920` — the same, on the branch production takes, with `max_recall_results = 5`.
- `recall_candidate_selection_caps_each_class_independently` — quotas are ceilings on their own class, rank order preserved, an absent class neither costs nor gives.
- `proactive_conversion_preserves_the_memory_class` — pins the lossy `MemoryLevel` fold and asserts the round trip still reads as raw dialogue.

Types (`librefang_types::memory::tests`)
- `raw_dialogue_is_the_per_turn_writer_signature`

### Negative controls

**Budget split** — `format_memory_items_by_class` reverted to one mixed list, one budget, `MEMORY_BULLET_LIMIT`:

```
memory_section_class_split_spills_an_unused_share_to_the_other_class ... FAILED
  dialogue's share was wasted: facts spent 1520, own share is 4011
memory_section_class_split_admits_facts_a_mixed_list_crowds_out ... FAILED
  class split did not admit the extracted facts
    left: 0    right: 8
```
The rendered section under the control was ten `zzdialogue` bullets followed by `(8 further remembered details are not shown here.)`. Restored: `4 passed; 0 failed`.

**Candidate selection, control A** — `select_recall_candidates(...)` reverted to `memories.truncate(MEMORY_RECALL_LIMIT)`, everything else kept:

```
recall_setup_carries_both_memory_classes_7920 ... FAILED
  every extracted fact in the store must reach the section: got 0 facts, 30 dialogue rows
    left: 0    right: 10
recall_setup_carries_both_classes_through_the_context_engine_7920 ... FAILED
  no extracted fact survived the engine path's candidate selection
```

**Candidate selection, control B** — only the supplemental fetch re-gated on `chat_scope_active || session_scope.is_some()`, selection left in place:

```
recall_setup_carries_both_classes_through_the_context_engine_7920 ... FAILED
  no extracted fact survived the engine path's candidate selection; candidates =
  [("episodic", ...), ("episodic", ...), ("episodic", ...), ("episodic", ...), ("episodic", ...)]
```
Five candidates, every one of them raw dialogue — the production shape exactly. The non-engine test still passes under B, so the two halves are independently necessary. Restored: `7 passed; 0 failed`.

## Also in here

`agent_context`'s cache tests now take a serialization guard. `poisoned_cache_lock_recovers_cached_context` asserts on process-global poison state that *any* concurrent cache access clears by design; it is safe under `cargo nextest`, which CI runs and which gives every test its own process, but simply adding tests elsewhere in the crate reshuffled the `cargo test` schedule enough to break it. Confirmed as a parallelism artifact — green at `--test-threads=1`, green on the base branch — and fixed rather than left for the next person to rediscover.

One unrelated flake was observed on this machine and is **not** from this change: `chatgpt_oauth::tests::test_callback_server_releases_listener_after_timeout` failed once in six full-suite runs and passes in isolation. It binds a local listener and asserts on a timeout; different module, no shared state with anything here.


nevgenov:
Thank you for the attribution, but the table in #7920 needs a correction before this merges, and it lands on the exact thing you flagged at the end. **The recall-limit change is not a separate decision — it is the load-bearing half, and the budget split alone is a no-op.**

## The split alone cannot bind

I re-derived what the section builder actually receives. Substrate recall truncates to `MEMORY_RECALL_LIMIT = 5`; `auto_retrieve` then appends up to `max_retrieve` (10 by default), deduplicated by content. Both draw from the same corpus with the same filter and the same cosine ranking, so the second subsumes the first and the builder receives the top 10.

Across the same 80 turns, with each row capped at `MEMORY_BULLET_MAX_CHARS`:

| | |
|---|---|
| median characters in the top 10 | 2614 |
| maximum | 4927 |
| turns where the total exceeds the 5000-character budget | **0 of 80** |

The budget never binds. The section is never forced to choose between a fact and a dialogue row, so a rule governing how it chooses changes nothing — with the current recall limit `format_memory_items_by_class` renders the same records as the class-blind path, in a different order.

## Where the measured gain actually came from

My B and C arms filled facts from the **full ranked candidate list**, not from the ten fragments the builder is handed. That is where 24–33 placed records came from. The builder sees 2.26 facts per turn on average, and **zero in 29 % of turns**. A budget split cannot place a fact that was never recalled.

So my table conflated two changes and presented them as one. The honest decomposition:

1. **Admit more facts past the recall truncation.** This is where the entire measured effect lives — roughly 26 facts instead of 2.26.
2. **Then** split the budget, because once step 1 lands the total is about 7500 characters against a 5000 budget and the section *does* have to choose. Without step 2, dialogue reclaims the space by arriving first.

Neither half does anything alone. Step 1 without step 2 overflows and truncates class-blind; step 2 without step 1 is what this PR currently is.

The "no extra retrieval or token cost" claim survives, but for a narrower reason than I stated: cosine is already computed over the whole candidate window (`MAX_BRUTEFORCE_CANDIDATES`), so taking more of the ranked result costs no additional scoring, and the token cost stays bounded by the unchanged section budget. What it costs is nothing — but it has to actually happen, and today the truncation prevents it.

## One correction to the PR description

> proactive-memory items are appended after that cap and do reach the section as facts

They reach it as whatever class they are. `ProactiveMemoryStore::auto_retrieve` builds `MemoryFilter::agent(agent_id)` with no scope constraint — the comment above the call reads "Search across all memory levels" — so it returns the same ~80/20 mixture as the substrate path. There is no fact-specific channel into the section today; that is why the 29 % figure is what it is.

## Suggestion

Yes, please do it in this PR. Splitting them would land a change that measures as zero and a follow-up that carries all of the effect, which makes both halves hard to evaluate and would leave the merged intermediate state looking like the fix.

Class-aware truncation at constant total breadth is the smaller version and I would take it: reserve part of the recall window for facts rather than raising the window. That keeps the retrieval shape unchanged and is enough to make the budget split load-bearing.

I can rerun the harness against whatever reserve you pick, on the same corpus and queries, before you merge — say the word and I will post the numbers here rather than have you take mine on faith a second time.

PR #7923, opened 2026-08-25
feat(scripts): add a memory retrieval evaluation harness

houko:
Refs #7756. Does not close it — this answers the harness question that has been sitting unanswered on that thread and leaves the rest of the issue where it is.

## The question this answers

@nevgenov described a ~300-line Python retrieval-comparison harness on #7756, said it "does not fit in-tree as-is", and offered three shapes: a script under `scripts/`, a Rust in-tree version with frozen judgments that CI could run, or a linked repository. They explicitly declined to pick.

This is the decision, implemented rather than described.

**In-tree, under `scripts/`, Python, stdlib-only, operator-run, and never in CI beyond a hermetic self-test.**

Why not the other two:

- **Frozen judgments in Rust, CI-able.** The judged pool is the union of every arm's top-k, so adding an arm changes what the judge sees for every arm. Frozen labels therefore only cover the arms captured at freeze time — which is precisely the case the harness exists to serve, comparing a *new* arm against the baseline. It would also hide the noise floor, since the floor is a consequence of pool churn, and report sub-floor differences as results. Determinism bought at the price of answering only questions already asked.
- **Out of tree.** That is the failure mode already observed. The compiled-in embedder went unmeasured for the life of the corpus not because anyone was careless but because measuring it required a program that lived nowhere. A linked repository is exactly as far away as no repository.

## Where the hermetic/non-hermetic line sits

The harness splits cleanly, and the split is the whole design.

**Never in CI:** judging and embedding. Both need live endpoints, so the tool is not hermetic and not deterministic — re-running the same commit gives slightly different nDCG. That is correct for a decision harness and wrong for a test. No workflow trigger, no schedule, no `workflow_dispatch`, and no new secret: this repository has an explicit stance that PR runners never see a provider key (`ci.yml` says so where `xtask integration-test` is invoked with `--skip-llm`), and nothing here asks it to change.

**In CI, on every PR:** the arithmetic. nDCG@10, the paired bootstrap, pool construction and the judge-response parse rule are pure functions of numbers, and a subtle error in any of them produces output that looks exactly like correct output. `--self-test` runs 35 synthetic-input tests in ~0.1s with no network and no secret. New job `Memory Eval Harness Self-Test`, modelled on `image-vuln-gate` — unconditional, stdlib-only, registered in `ci-gate`'s `needs`.

## Substantive changes

- **`scripts/memory-retrieval-eval.py`** — the harness. Stdlib only; no `requirements.txt` is introduced, matching the convention `check-skills-supply-chain.py` states in its header and `read_email.py` follows (the nearest precedent: a live-endpoint, env-credentialed operator tool that CI never invokes).
  - Corpus is opened `mode=ro&immutable=1` against a **copy**. It never contacts the daemon, never binds a port and never writes to what it reads.
  - Arms: `cosine` over the stored `embedding` BLOBs (`decode_embedding` mirrors `embedding_from_bytes` in `semantic.rs`, trailing-partial-float behaviour included), `fts5` via `memories_fts MATCH … ORDER BY rank` — the daemon's own index from migration v50 (#7808), not a reimplementation of BM25 that can drift — and weighted `rrf` over the two.
  - `--reembed` rebuilds every corpus vector with a candidate embedding model and refuses to mix re-embedded and stored rows, so the control run that overturned the headline finding on #7756 is reproducible here. `--query-prefix` / `--passage-prefix` make the asymmetric e5 contract a measurable variable, which it is not anywhere else (#7912, gap 3).
  - Judgments and embeddings are cached on disk keyed by model and content, so adding an arm costs the pairs it newly contributes rather than a full re-spend.
- **The paired bootstrap is mandatory.** There is no unpaired mode and `bootstrap_paired_ci` rejects unaligned input. Between-query variance dwarfs between-arm variance at this corpus size, so an unpaired comparison cannot detect the effect at all — it reports "no difference" for a difference that is really there. The pool is a set union so a document retrieved by three arms is graded once and every arm reads the same grade, which is what makes the differences paired in the first place.
- **The noise floor is reported, not assumed.** Every comparison is classified `noise` / `significant but below the noise floor` / `significant`, and the middle verdict is the one people skip. `--noise-floor RUN_A.json RUN_B.json --stable-arms fts5` recomputes it from two runs using arms that could not have changed, which is the diagnostic that caught the retraction on #7756.
- **A query set under ~30 is flagged in the report** as too small for the bootstrap to mean anything. The runs on #7756 counted 59–80.
- **Judge responses outside 0–3 or unparseable are discarded whole, never clamped**, and one bad pair drops the whole query so every arm stays scored over the same pool. Clamping a 26 to a 3 substitutes plausible data for bad data and the result is indistinguishable from a real judgment.
- **`docs/development/memory-retrieval-eval.md`** — the decision, the operator procedure, how to read the output, and what the method does not settle.
- **`.github/workflows/ci.yml`** — the self-test job, plus its registration in `ci-gate`.

## Verification

- `python3 scripts/memory-retrieval-eval.py --self-test` — 35 passed, 0 failed, 0.098s, no network.
- `python3 scripts/check-workflow-timeouts.py` — OK, all 137 workflow jobs declare `timeout-minutes`.
- End-to-end exercised against a synthetic SQLite corpus (5 rows, `memories` + `memories_fts`, real little-endian f32 BLOBs) with `http_json` replaced by a fake: all three arms rank, the pool is judged, the report renders, `--out` writes per-query scores, the second run serves entirely from the judgment cache, and `--noise-floor` reads two run files back. Not committed — it is a smoke check, and the contracts it covers are pinned by `--self-test`.
- Nothing was run against a live daemon, a live judge or a live embedding endpoint.

No Rust changed, so no `cargo` verification applies. No config field, no migration.

## Deferred

- **A second judge model.** Every run on #7756 shares one judge and none of them controlled for systematic judge bias. The cache is keyed by judge model so a second judge starts cold rather than blending two judges into one number, but running one is an operator decision and needs a second endpoint.
- **Corpus-composition measurement** (retrieval rate by age cohort, class ratios, length distributions). That is the diagnostic that mattered for episodic retention and it belongs with #7911, not here.
- **Any change to what the daemon ships.** This measures; it decides nothing. Replacing one compiled-in guess with another compiled-in guess on the strength of three corpora that share a language, an embedder and an extraction pipeline is the same unvalidated bet, and the reporter said so first.

## Credit

The method, the noise-floor diagnostic, the discard-don't-clamp rule and every number quoted in the docs are @nevgenov's, from #7756 and #7912. The corpora are real production data, and the public retraction of their own headline finding — with the numbers — is the most useful thing in that thread.

### DIFF
--- a/scripts/memory-retrieval-eval.py
+++ b/scripts/memory-retrieval-eval.py
@@ -1,0 +1,1238 @@
+#!/usr/bin/env python3
+"""Compare memory retrieval strategies on a real corpus, with an LLM judge and paired confidence intervals.
+
+Why this exists
+---------------
+
+LibreFang compiles in one retrieval strategy and one embedding model, and until now there was no supported way
+for an operator to ask whether either choice was costing them anything on their own corpus.
+That question was answered once, by @nevgenov on #7756, using an out-of-tree Python harness run against a
+read-only copy of a production database — five runs over three real corpora, ~5800 LLM-judged pairs.
+The headline finding of the first run ("the retriever the runtime hardcodes is the worst of five arms") did not
+survive the reporter's own control run: it was an artifact of a 4-bit-quantized embedder, and swapping the
+embedding model moved the vector arm further (+0.096 / +0.032 nDCG@10) than any difference between retrieval
+methods that any of the five runs could resolve.
+
+Both the finding and its retraction are the argument for this file.
+The measurement was decisive twice, and both times it required work that lived nowhere in the tree.
+
+What it does
+------------
+
+For each query, every arm produces a ranked top-k.
+The union of those lists is pooled and shuffled, so the judge grades documents without seeing which arm
+retrieved them or in what position.
+An LLM grades each (query, document) pair 0-3.
+nDCG@10 is computed per arm per query, and every arm is compared against the leader with a confidence interval
+bootstrapped over *paired* per-query differences.
+
+The pairing is not a detail.
+Between-query variance dwarfs between-arm variance on corpora this size: an unpaired comparison cannot detect
+the effect at all, and will report "no difference" for a difference that is really there.
+This harness therefore has no unpaired mode, and `bootstrap_paired_ci` refuses inputs that are not aligned
+per query.
+
+What it deliberately does not do
+--------------------------------
+
+It does not run in CI, on any trigger, ever.
+It needs a live judge model and a live embedding endpoint, so it is neither hermetic nor deterministic — the
+same commit re-run gives slightly different nDCG.
+That is fine for a decision harness and wrong for a test.
+
+The one part that *is* hermetic is the arithmetic — nDCG, the paired bootstrap, pool construction, and the
+judge-response parse rule — and getting that subtly wrong is the failure mode nobody would notice.
+So it is unit-tested with synthetic inputs under `--self-test`, which is what CI runs, in about a second, with
+no network and no secret.
+
+Reading the numbers
+-------------------
+
+The method has a measurable noise floor of roughly 0.04 nDCG, and it is measurable rather than assumed.
+Because the judged pool is the union of all arms, changing one arm changes what the judge sees for every arm:
+in the reporter's runs the embedder-independent lexical arms moved by -0.040 and -0.016 between two runs whose
+only changed variable was the vector arm.
+Any difference below that floor is not a result.
+`--noise-floor RUN_A.json RUN_B.json` recomputes it for your own setup from two runs, using the arms you
+declare embedder-independent.
+
+Three things that each cost the reporter a re-run, and where this harness stands on each:
+
+* e5-family embedders are trained on an asymmetric `query:` / `passage:` pair, and omitting the prefixes
+  narrowed the relevant-vs-irrelevant margin from 0.12 to 0.07 — enough to produce a false confirmation of the
+  hypothesis under test.
+  Guarded: `--query-prefix` and `--passage-prefix` make this a measurable variable here, even though the
+  daemon cannot yet send prefixes (#7912, gap 3).
+* Comparing a candidate embedder against the stored vectors is meaningless, because a query in one embedding
+  space scored against vectors from another is not a comparison of anything.
+  Guarded: `--reembed` rebuilds the corpus vectors with the configured endpoint, and refuses to mix the two.
+* Comparing per-record retrieval counts without controlling for age inverts the conclusion — a same-month
+  cohort gave 29.5 against 110 where the uncontrolled numbers said the opposite.
+  Not guarded, because it is a corpus-composition question rather than a retrieval one; if you extend this
+  tool that way, control for age.
+
+Safety
+------
+
+The corpus is opened read-only and immutable, and it must be a *copy*.
+This tool never contacts the daemon, never binds a port, and never writes to the database it reads.
+Corpus text is sent to whatever judge endpoint you configure — that is a third party unless you point it at
+something you run, so choose it deliberately.
+
+Usage
+-----
+
+    python3 scripts/memory-retrieval-eval.py --self-test
+
+    cp ~/.librefang/memory.db /tmp/corpus.db
+    export LIBREFANG_EVAL_JUDGE_URL=https://.../v1/chat/completions
+    export LIBREFANG_EVAL_JUDGE_MODEL=...
+    export LIBREFANG_EVAL_JUDGE_KEY=...
+    export LIBREFANG_EVAL_EMBED_URL=https://.../v1/embeddings
+    export LIBREFANG_EVAL_EMBED_MODEL=...
+    python3 scripts/memory-retrieval-eval.py run \\
+        --corpus /tmp/corpus.db --queries queries.txt --arms cosine,fts5,rrf \\
+        --out run-bge-m3.json
+
+    python3 scripts/memory-retrieval-eval.py --noise-floor run-a.json run-b.json --stable-arms fts5
+
+To compare a different embedding model, re-embed the corpus rather than scoring new queries against old
+vectors:
+
+    python3 scripts/memory-retrieval-eval.py run \\
+        --corpus /tmp/corpus.db --queries queries.txt --arms cosine,fts5 --reembed \\
+        --embed-model intfloat/multilingual-e5-large --query-prefix 'query: ' --passage-prefix 'passage: ' \\
+        --out run-e5.json
+
+Stdlib only, on purpose: no virtualenv, no requirements file, `python3 scripts/memory-retrieval-eval.py` and
+it runs, exactly like every other tool in `scripts/`.
+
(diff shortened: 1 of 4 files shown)


scripts/memory-retrieval-eval.py
```diff
@@ -0,0 +1,1177 @@
...
+    url = args.embed_url or os.environ.get("LIBREFANG_EVAL_EMBED_URL")
+    model = args.embed_model or os.environ.get("LIBREFANG_EVAL_EMBED_MODEL")
+    key = os.environ.get("LIBREFANG_EVAL_EMBED_KEY")
+    if not url or not model:
+        raise SystemExit(
+            "embedding needs an endpoint: set LIBREFANG_EVAL_EMBED_URL and LIBREFANG_EVAL_EMBED_MODEL "
+            "(or --embed-url / --embed-model)"
+        )
+    response = http_json(url, {"model": model, "input": list(texts)}, key)
+    return [item["embedding"] for item in response["data"]]
```
houko, line 608: Both `embed_texts` and `make_embedder` take `response["data"]` in list order and zip it positionally against the submitted batch, without reading each item's `index` field back.

That's consistent with the OpenAI embeddings contract in the common case, but it sits oddly next to `reembed_corpus`'s own stated principle a few lines up ("this replaces *every* vector or fails; it never mixes stored and re-embedded rows") — a proxy or "OpenAI-compatible" endpoint that reorders `data` under batching would silently mispair vector[i] with document[i] while the length check still passes, which is exactly the failure mode the harness otherwise goes out of its way to guard against (see the discard-don't-clamp judge parsing, and the vector-count check just below this line).

Worth sorting by `item["index"]` before zipping if you want the same guarantee here, or a one-line comment noting it's an accepted assumption if not.

---

scripts/memory-retrieval-eval.py
```diff
@@ -0,0 +1,1177 @@
...
+# ── CLI ──────────────────────────────────────────────────────────────────────────────────────────────────
+
+
+def build_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        description="Compare memory retrieval strategies on a real corpus with an LLM judge.",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="Never runs in CI beyond --self-test. Needs a live judge and a live embedding endpoint.",
+    )
+    parser.add_argument("command", nargs="?", default="run", choices=["run"], help="what to do (default: run)")
```
houko, line 1117: Minor: `command` only ever has one valid value (`choices=["run"]`), and `main()` doesn't actually dispatch on `args.command` — `--noise-floor` and `--self-test` are handled as flags checked before this positional is consulted at all. Was a multi-command CLI (`run` / `noise-floor` as subcommands) the original plan, with this left over from an earlier shape? If not, this positional could go and `run` stay the only path, which would also remove the now-unreachable `choices` check.

---

houko (commented):
Automated pass against CLAUDE.md conventions. Pushed one mechanical fix (rewrapped the new script's docstrings/comments to one-sentence-per-line, matching the convention already followed in the docs/changelog additions in this PR — verified `--self-test` still passes, 35/35). Two smaller points left as inline comments for a human call: possible embedding-response reordering risk, and a seemingly vestigial CLI positional argument. No other findings in the arithmetic (nDCG, paired bootstrap, verdict classification) — it checks out against the self-test suite.

---

nevgenov:
Ran `--self-test` (35 tests, 0.087 s) and read it against the way I actually used my own version. Three things, one of which I think blocks a first run producing a meaningful number.

## 1. Nothing excludes the document the query came from

`--queries` takes "real user messages, not generated ones", and on a LibreFang corpus a real user message is stored verbatim inside a `[Past exchange]` row in the same store. So the query is a near-copy of a document in the corpus, and every arm retrieves it.

Measured on the corpus from this thread, 37 queries drawn the way the help text recommends:

| | |
|---|---|
| query's own source row ranked **first** | 49 % |
| in the top 10 | 76 % |
| cosine of that row to its own query | median **0.820**, max 0.931 |

The judge then grades that document highly, because it does contain the question. Every arm collects the same free hit near rank 1, which inflates absolute nDCG for all of them and *compresses the differences between them* — which is the quantity the harness exists to measure. It pushes results toward "no difference" and would make the noise floor look larger than it is.

This was the first thing that went wrong in my own runs and the reason my harness carries `q["source"]` and drops that id per query. I have no view on which shape you prefer — an optional second column in the queries file (`id<TAB>text`), or a documented temporal-split recipe where queries come from messages newer than every document in the corpus. The second is what I used for the most recent measurement and it needs no format change, but it does need to be written down, because the natural way to build a queries file is `SELECT` the user turns out of the same database.

## 2. `verdict()` calls a straddling interval significant

```python
if max(abs(low), abs(high)) < noise_floor:
    return "significant but below the noise floor"
return "significant"
```

`max` means the middle category only fires when the *whole* interval is inside the floor. An interval like `[+0.01, +0.20]` against a 0.04 floor reports plain `significant`, even though the true effect is plausibly 0.01 — inside the floor, and exactly the case the middle category was introduced to catch.

The bound that has to clear the floor is the one nearest zero, so for an interval that already excludes zero that is `min(abs(low), abs(high))`.

The three self-test cases are `(-0.01, 0.05)`, `(0.005, 0.02)` and `(0.05, 0.12)` — both bounds on the same side of the floor in each. The straddling case is the only one where the two readings disagree and it is the one not pinned.

For calibration: my most recent comparison came out at `+0.39 [+0.05, +0.75]` against this floor. Both rules call it significant, but only just, and under `max` I would not have been told how close it was.

## 3. Per-pair judging costs about 30× a pooled call, which is worth stating in the docs

`make_judge` returns `(query, document) -> grade`, so a run is one request per judged pair rather than one per query. With three arms at `--top-k 10` the pool runs 20–30 documents, so 80 queries is roughly 2000 requests against 80.

I am not arguing for the pooled form — per-pair removes position effects inside the list and means a malformed response costs one pair instead of a whole query, and I think that is the better trade. But it changes what an operator should expect from a first run enough that it belongs next to the runtime estimate, and it makes `--judge-cache` load-bearing rather than a convenience. It is also worth saying explicitly that per-pair grading is uncalibrated across documents: the judge never sees two candidates side by side, so a grade of 2 in one call and a 2 in another are only as comparable as the rubric makes them.

## What I checked and found correct

`fts_match_expression` strips to alphanumerics and quotes every token, so operators, stray quotes and `-` in a real message are data; `str.isalnum()` is Unicode-aware, so the Cyrillic corpus survives it. `build_pool` is a set union over `sorted(rankings)` with a seeded shuffle, which is what makes the differences paired and reproducible. `dcg` uses the `(2^g − 1)` formulation. `bootstrap_paired_ci` resamples queries rather than observations and refuses unaligned input. `parse_judge_grade` discarding out-of-range responses whole is the right call and matches what a clamped `26 → 3` did to one of my runs.

Points 1 and 2 are small diffs. I will rerun my corpus through it once the source-exclusion question is settled and post the numbers next to my own harness's, which is the comparison that would actually validate this rather than my reading it.

Issue #7746, opened 2026-08-20
feat: IAM integration — map external identity provider groups, roles and scopes onto LibreFang authorization


DaBlitzStein:
## Where we are

OIDC login exists (#182, implemented; the flow lives in `crates/librefang-api/src/oauth.rs`).
A user can authenticate against an external provider.

What happens after that is the gap: authorization is entirely local and entirely flat.
`crates/librefang-api/src/routes/users.rs:184` defines the full permission vocabulary:

```rust
const VALID_ROLES: &[&str] = &["owner", "admin", "user", "viewer"];
```

Four hardcoded strings, assigned by hand, per user.

So an organisation running Keycloak, Okta, Entra, Authentik or Google Workspace already has its group structure, its role assignments and its scopes — and LibreFang throws all of it away at the door, then asks an operator to re-enter a subset of it by hand, one user at a time, and to remember to undo it when someone leaves.

That last part is the real risk: **deprovisioning does not propagate.** Revoking access in the IdP does not revoke it here.

## What is needed

**1. Consume the claims we are already handed.**
Read `groups`, `roles` and `scope` from the ID token / userinfo response, with the claim names configurable — every provider spells them differently (`groups`, `realm_access.roles`, `resource_access.<client>.roles`, `wids`, custom claims).

**2. Map external identities onto local principals.**
Operator-defined mapping rules, e.g. IdP group `platform-oncall` → LibreFang group `oncall`, IdP role `librefang-admin` → local `admin`.
Depends on #7745 for the group side.

**3. Re-evaluate on every login, not just the first.**
Group membership and roles refresh from the token each time, so removing someone from a group in the IdP takes effect here without manual cleanup.

**4. Replace the four fixed roles with real permissions.**
Scopes or capabilities that can be composed, so an organisation can express "may run workflows but not create agent types" without patching `VALID_ROLES`.
Keep the four names as presets so nothing breaks.

**5. Enforce it where it matters.**
Route-level authorization derived from the resolved principal, not just the presence of a valid token.

## Interaction with the ownership model

#7744 proposes `Principal = User | Group | Role` for everything an agent creates.
This issue is what makes the `Group` and `Role` arms trustworthy: without an IdP as the source of truth, group and role membership is hand-maintained and drifts.

## Scope note

Deliberately separate from #7745. Groups are useful standalone, even in a deployment with no external IdP; this issue is about not discarding what the IdP already knows.

[...]

PR #7928, opened 2026-08-25
feat: everything an agent creates belongs to the principal it acted for

### LINKED ISSUE #7744: feat: ownership model — everything an agent creates belongs to the principal it acted for (user, group, or role)
## The gap

An agent can create durable things on its own: workflows (`workflow_create`), skills (skill workshop, evolution), cron jobs (`cron_create`), triggers, other agents (`agent_spawn`), memories.
None of them records who it was created *for*.

`workflow_create` is the case that surfaced it.
The tool now logs the calling agent, but that is the executor, not the owner — and a log line is not ownership: logs rotate, the workflow outlives them.

The principle this is missing: **an agent acts on behalf of someone. Whatever it creates belongs to that someone, not to the agent.**

## Why an owner field is not enough

Ownership in a real deployment is not always one human.
A workflow built during a support shift belongs to the support team; a compliance cron belongs to whoever holds that duty this quarter; an ops runbook belongs to a role, and the humans filling it change.

So the owner should be a **principal**, not a user id:

```
Principal = User(id) | Group(id) | Role(name)
```

The same shape IAM systems converge on, for the same reason: identity outlives the individual.

## What already exists

- `SenderContext.user_id` (`crates/librefang-channels/src/types.rs:344`) — platform-specific sender.
- `ChannelMessage.sender.librefang_user` (`crates/librefang-channels/src/types.rs:110`) — the canonical LibreFang identity, mapping a platform user onto a real system user.
- Full user CRUD at `/api/users` (`crates/librefang-api/src/routes/users.rs`), with a flat `role: String` field (`:81`, `:123`).
- `BindingContext.roles` (`crates/librefang-channels/src/router.rs:27`) — roles already travel through channel binding resolution.

What does not exist: groups (no `Group` type, no `group_id`, anywhere), and any notion of a principal that unifies the three.

## Where the chain breaks

The identity is captured at the edge and lost before it reaches creation.
`caller_agent_id` reaches `create_workflow`; the user behind that turn does not.
The same symptom shows up in metering, where the ephemeral spawn path writes `user_id: None` outright.

The chain that needs to carry a principal end to end:

```
channel message / API request / TUI  →  session  →  agent turn  →  tool call  →  created artifact
```

## Proposal

1. Introduce a `Principal` type (`User` / `Group` / `Role`) in `librefang-types`.
2. Add groups: a `Group` entity and user-group membership, alongside the existing users surface.
3. Thread the resolved principal through the session and the agent turn, so every tool call knows who it is acting for. Where an agent runs with no human behind it (cron, autonomous loop), the principal is the agent's own configured owner — explicit, never absent.
4. Stamp `owner: Principal` on every artifact an agent can create: workflows, skills, cron jobs, triggers, spawned agents, memories.
5. Use it: per-principal quotas, "show me what I own" filters in the dashboard and TUI, cascade decisions when a user or group is removed, and a real authorisation check on mutations of an owned artifact.

## Why this matters beyond bookkeeping

Without an owner there is no one to hold accountable and no one to throttle.
A prompt-injected agent can create a workflow whose steps target a more privileged agent, and there is currently no principal to attribute that to, no quota to stop it, and no way to revoke what it made.

## Scope note

This is deliberately filed as the whole model rather than a `created_by` column on `Workflow`.
Adding that one field would touch ~40 struct literals across three crates plus the dashboard type and the OpenAPI schema — the same blast radius as doing it properly, for a fraction of the value, and it would have to be redone once groups and roles arrive.

houko:
Closes #7744.

The design in that thread is @DaBlitzStein's, worked out with @houko across five comments, and this implements it as settled there — the two-arm principal, the tagged struct, the optional field with a config default, and the refusal to put any of it in `manifest.metadata`. All four decisions are restated below with the reasoning that produced them, because the issue was closed and reopened once and the argument should not have to be reconstructed from a comment thread again.

## Depends on #7913 — must merge after it

Based on `origin/feat/7745-user-groups`, not on `main`, and targeted at `main`.

A principal is `User | Group`. The `Group` arm has nothing behind it on `main` today, so basing this on `main` would mean either inventing a second groups concept or shipping an arm that names nothing. `Principal::group_named` derives its id from a `[[groups]] name`, and `KernelConfig::principal_name` resolves it back through the `groups` list #7913 adds.

**Do not merge this before #7913**, or its commits land out of order. Auto-merge is deliberately off.

## What "acted for" resolves to today

Establishing this first, because the answer decides the whole shape.

After #7884, the authenticated caller already reaches the agent turn, typed, as `owner: Option<UserId>`:

- read at `crates/librefang-api/src/routes/agents/messaging.rs:39` from `AuthenticatedApiUser`
- threaded through `KernelApi` (`kernel_api.rs:418, 672, 701`) into `send_message_full_inner` (`kernel/messaging.rs:1011`) and `execute_llm_agent` (`kernel/agent_execution.rs:373`)
- consumed for exactly two decisions: provider-credential selection (`resolve_driver_for_owner`, `agent_execution.rs:1060`) and spend attribution / per-user budget gating (`billed_user_id`, `agent_execution.rs:1466`)

And it **stopped there**. `run_agent_loop` had no owner parameter (`crates/librefang-runtime/src/agent_loop/mod.rs:458`), and `ToolExecContext` had no `UserId` field — the runtime's only notion of "who is driving this turn" was `manifest.metadata["sender_user_id"]`, a *platform* identifier that steers per-sender tool authorization and `peer:{user_id}:KEY` memory scoping.

Every non-HTTP entry point arrives with `owner = None`, structurally: they all funnel through `send_message_full` (`messaging.rs:920`), whose signature has no owner and which hardcodes `None` at `:947` — cron (`cron_tick.rs:348`), triggers (`triggers_and_workflow.rs:688`), workflow steps (`:1332`), `agent_send` (`messaging.rs:205`), channel bridges, forks.

So the increment is: carry that value one hop further, and give the ownerless turns something to resolve to.

## The four decisions

**1. Two arms, not three.** `Principal = User(UserId) | Group(name)`. The `Role` arm is dropped: `UserRole` (`crates/librefang-kernel/src/auth.rs:25`) is an ordinal privilege level `Viewer=0 … Owner=3`, not an identity. "Owned by admin" means "owned by everyone at or above admin", which is a permission predicate, and it has no defined answer to the question ownership must answer on deletion — you cannot delete `admin`, so nothing can cascade. The rotating-duty example in the issue body is a group with rotating membership, which #7913 provides and #7746 will populate from IdP claims.

**2. A tagged struct, not a Rust enum.** `struct Principal { kind: PrincipalKind, id: Uuid }`, serialized as `{"kind":"user","id":"…"}`. Every artifact carrying an owner is `Serialize`/`Deserialize` and reaches `openapi.json` and four generated SDKs. An enum's serde representation is a choice that has to be re-made the day an `Agent` or `Service` arm is added; a struct with an explicit `kind` string has one shape forever, and a new kind is an additive change every generated client already tolerates. The cost — `match` over `kind` is not exhaustiveness-checked — is accepted because there are few such sites and the wire shape is the expensive thing to get wrong. `wire_shape_is_a_flat_tagged_struct` pins the literal.

**3. Optional, with a config default.** `agent.toml: owner` and `config.toml: default_owner` are both `Option<String>`. A required field would invalidate every existing `agent.toml`, and an ownership model that cannot be adopted incrementally is not adopted at all. A malformed spec logs a `WARN` and falls through to the next step rather than failing the turn.

**4. Not in `manifest.metadata`.** The principal rides on `LoopOptions` (`agent_loop/types.rs`) and `ToolExecContext` (`tool_runner/dispatch.rs`) as a typed field. That untyped bag is exactly how `sender_user_id` became a stringly value with no declared type and no single authority over who writes it.

## Which resources get an owner, and what was left out

**In: workflows and cron jobs (including `schedule_create`, the natural-language front end for the same artifact).**

Both are agent-creatable with a tool, both are the case the issue names, and both are persisted as serde structs in JSON files — `<home>/workflows/<id>.workflow.json` and `<home>/data/cron_jobs.json` — so `#[serde(default, skip_serializing_if)]` covers the upgrade with **no SQLite migration at all**. That matters right now: v51/v52/v53 are claimed by #7916, #7919 and #7904, and this PR takes no rung.

Workflows are the sharpest case. `Workflow` carried *no* provenance of any kind, and the kernel's own comment said so — `crates/librefang-kernel/src/kernel/handles/workflow_runner.rs:661`, "Provenance trace, not an authorization gate: workflows have no ownership model". That comment is now replaced by a recorded owner.

**Out, with reasons:**

- **Memories** (`memories`, `kv_store`). Already scoped by `agent_id` + `peer_id` and gated by a namespace ACL. Adding a third scoping dimension to a hot SQLite path needs a migration on the contended ladder *and* a decision about how ownership composes with the existing ACL — which is enforcement, not recording, and belongs with the PR that enforces.
- **Spawned agents** (`agents` table). Wants a migration, and there is a separate latent bug there worth naming: `AgentEntry.parent` is never written to any column and every load path hardcodes `parent: None` (`crates/librefang-memory/src/structured.rs:452, 664, 1048`), so agent lineage lives in RAM and dies at restart. Fixing that is a schema change plus three load sites, in a different crate, and it deserves its own issue rather than being smuggled in here.
- **Skills.** `librefang-skills` already records a per-version author in `.evolution.json`. It also has a real one-line bug — `evolution.rs:1051` hardcodes `author: "agent-evolved"` in the written `skill.toml`, discarding the `author` parameter it was handed — but that is a different crate and a different domain, so per CLAUDE.md I am raising it here rather than deciding unilaterally: **@houko, want that one-liner in this PR, or as its own issue?**
- **Triggers.** Have no agent-facing create tool, so they fall outside "everything an agent creates". They already carry `agent_id`.
- **Approval-deferred tool calls.** See below — this is the one decision the issue thread left genuinely open, and I did not close it.

## What an owner is stored as

A `Principal` is `{ kind, id }` with the id derived UUIDv5 in a per-kind namespace (`LIBREFANG_GROUP_NAMESPACE` alongside the existing `LIBREFANG_USER_NAMESPACE`), so `user:alice` and `group:alice` can never collide.

Two string forms, deliberately not interchangeable:

- `Display` / `FromStr` are the canonical storage form, `kind:uuid`. One `TEXT` column holds a whole principal and `WHERE owner = ?` filters on it, for whenever an artifact does move into SQLite.
- `Principal::from_spec` is the operator form, `user:alice` / `group:oncall` / bare `alice`, for `config.toml` and `agent.toml`. The value after the colon is **always** a name and never a uuid, because every other identity reference in `config.toml` names a principal by name and accepting both would make `user:<uuid-shaped-thing>` silently ambiguous. Both directions are pinned by tests, including the documented corner where a uuid-shaped *name* is hashed like any other name.

UUIDv5 is what makes the owner queryable without a reverse index: the derivation is a pure function, so a "what does the on-call group own" filter is built by deriving the principal from the name and comparing ids. The reverse — id back to a display name — is `KernelConfig::principal_name`, a linear scan of the two operator-authored lists rather than a stored mapping that could disagree with `config.toml` after an edit. `GET /api/workflows/{id}` returns both `owner` (machine-readable) and `owner_name` (`null` when unowned *or* when the principal is no longer declared, which the doc comment says explicitly).

## Precedence, and why the manifest does not win

`resolve_acting_principal(authenticated, manifest_owner, config_default)`:

1. **The authenticated caller.** Same `owner: Option<UserId>` that already picks the caller's provider credential and bills their budget, so ownership, credential choice and spend attribution read one identity by construction rather than three that can disagree.
2. **`agent.toml: owner`.** For the turns with no human — this is where group ownership is reached in practice: `owner = "group:compliance"` on the agent that runs the quarterly cron.
3. **`config.toml: default_owner`.**
4. **`None`.**

The manifest deliberately does *not* override the authenticated caller. An override would mean a support agent configured `owner = "group:support"` silently relabels Alice's workflow as the team's — the opposite of "the principal it acted for", and it would make the recorded owner useless as an audit answer. `the_authenticated_caller_wins_over_every_configured_fallback` pins it.

Forks are cleared fork-aware, mirroring the existing `effective_owner` rule for billing: what a fork creates is recorded against the agent's configured owner or nothing, never against the person whose turn spawned it.

## What happens to existing rows

**`None` means unowned, visible to all.** Stated, tested, and not a default nobody chose.

Every workflow definition and cron job written before this field deserializes to `None` (`a_workflow_created_before_ownership_deserializes_as_unowned`, `a_job_written_before_ownership_deserializes_as_unowned`). Because ownership restricts nothing yet, an unowned artifact behaves exactly as it did before — which is what makes the field safe to add with no migration and no backfill. `skip_serializing_if` keeps the unowned on-disk shape byte-identical, so the upgrade does not rewrite existing files on the next persist (`an_unowned_workflow_omits_the_key_entirely_when_serialized`, and the cron twin).

The first artifact created with no resolvable principal logs one `WARN` per daemon naming both config keys (`principal::warn_once_unowned`); every subsequent one is `DEBUG` with the artifact id. Once per process, not once per artifact — a deployment that has declared no principals creates unowned artifacts continuously, and a per-artifact warning would be pure noise in exactly the deployment least equipped to act on it.

## Records only. Nothing is enforced.

No branch reads the owner as a permission. Per-sender tool authorization, approval routing and `peer:{user_id}:KEY` memory scoping all still use `sender_id`, unchanged. The dispatch table does not consult `acting_principal` for anything but stamping.

Two things that look like enforcement and are not:

- **`PUT /api/workflows/{id}` carries the stored owner over unchanged**, exactly as it already carries `created_at`, and the field is deliberately not readable from the request body. That is not an authorization check — anyone who could edit before can still edit. It stops an *edit* from silently being a *transfer*, which would make the recorded owner worth less than the log line it replaced. Transfer, if wanted, belongs on its own route with its own authorization.
- **The create routes read the caller from the auth extension, not the body.** Same class of hole #7884 closed on `POST /api/agents/{id}/message`: a body field naming its own owner is a field the caller chooses.

## Changes

**`librefang-types`**
- New `principal` module: `Principal`, `PrincipalKind`, `LIBREFANG_GROUP_NAMESPACE`, `resolve_acting_principal`, `warn_once_unowned`, `Display`/`FromStr`/`from_spec` and their error types. The module doc carries the full argument for every decision above.
- `AgentManifest.owner: Option<String>` (`#[serde(default, skip_serializing_if)]`), plus the `Default` entry, the exhaustive-destructure guards in `manifest_privacy.rs` (dropped on publication — a principal names a `[[users]]` entry in *this* deployment) and `config/mod.rs`, and `agent_type.rs` (a type is a template, not an instance).
- `KernelConfig.default_owner: Option<String>` with all four required pieces, plus `default_owner_principal()` and `principal_name()`.
- `CronJob.owner: Option<Principal>`.

**`librefang-kernel` / `librefang-kernel-handle`**
- `Workflow.owner: Option<Principal>`; `WorkflowFile` parses an operator-declared one; `WorkflowTemplate` instantiation leaves it `None`.
- `KernelHandle::create_workflow` and `CronControl::cron_create` take `owner: Option<Principal>` as a typed argument, never inside the JSON payload. `WorkflowSummary` reports it.
- `execute_llm_agent` resolves the principal once, next to the `owner` it already holds; the streaming path resolves it in `send_message_streaming_with_sender_and_opts`, the first point where the manifest is in hand — the same place `compaction_config` is resolved, and fork-aware.
- `build_reload_plan` classifies `default_owner` as read-live/no-op, with the drift-guarded row in `docs/operations/config-reload.md`.

**`librefang-runtime`**
- `LoopOptions.acting_principal` → `ToolExecContext.acting_principal` → `tool_workflow_create` / `tool_cron_create` / `tool_schedule_create`.
- `execute_tool_with_sender_account` takes it; the positional `execute_tool` shim passes `None`.

**`librefang-api`**
- `AuthenticatedApiUser::owner_principal()` — returns `None` for the synthetic root sentinel (master key, loopback, `allow_no_auth`), because that sentinel means "authentication is off", names no `[[users]]` entry, and recording it would write a principal that resolves to nothing and looks real. That is what lets an unauthenticated deployment fall through to `default_owner`.
- `POST /api/workflows`, `POST /api/cron/jobs`, `POST /api/schedules` stamp the caller; `PUT /api/workflows/{id}` preserves; `workflow_to_json` reports `owner` + `owner_name`.
- MCP HTTP bridge (`routes/network.rs`) and the channel `/schedule add` bridge pass `None` — a header-supplied peer jid and a platform sender are routing keys, not credentials.

**Docs**: `docs/architecture/agent-manifest-field-parity.md` (58 → 59 fields, table renumbered), `docs/operations/config-reload.md`.

## Verification

- `cargo test -p librefang-types` — 1006 lib tests pass, including 15 new ones: 9 in `principal::tests` (namespace separation, agreement with `UserId::from_name`, pinned uuid literals, canonical round-trip, canonical parse rejecting a name where a uuid belongs, spec form always hashing, blank/unknown-kind rejection, the frozen wire shape, `as_user_id` narrowing), 6 in `principal::resolution_tests` (the caller winning over both fallbacks, the manifest arm, the config arm, unowned, a malformed manifest spec falling through, a bare name meaning a user), plus 4 in `config::tests` (`default_owner` defaulting and being omitted, a group spec, a malformed spec resolving to `None` rather than failing the load, `principal_name` resolving both arms and refusing to cross them) and 3 in `scheduler::tests` (pre-#7744 job parses unowned, unowned omits the key, a stamped owner round-trips).
- `cargo test -p librefang-kernel --lib` — 1670 pass. Includes 5 new in `handles::workflow_runner::tests`: the owner is stamped from the argument, an `owner` in the spec is ignored, a pre-ownership definition parses unowned, unowned omits the key, a stamped owner round-trips through the persisted JSON.
- `cargo test -p librefang-kernel --test kernel_handle_contract_cron_spawn` — 6 pass, 2 new: an owner survives the scheduler's own store via `cron_list`, and an unowned job leaves the key absent rather than null.
- `cargo test -p librefang-runtime` — 2400 lib tests plus the tool-runner integration suites. 5 new: `cron_create_forwards_the_acting_principal_to_the_kernel`, `cron_create_forwards_no_principal_when_the_turn_resolved_none`, `a_model_supplied_owner_in_the_tool_input_cannot_choose_the_principal`, `workflow_create_forwards_the_acting_principal_to_the_kernel`, `a_model_supplied_owner_in_the_spec_cannot_choose_the_principal`. Each asserts both that the principal arrived as a typed argument *and* that it did not arrive inside the model's own payload.
- `cargo test -p librefang-api --test artifact_ownership_test` — 7 new integration tests against the production `routes::workflows::router()` behind the real `middleware::auth`, each write followed by an independent `GET` (refs #3721): `creating_a_workflow_records_the_authenticated_caller_as_its_owner`, `an_owner_named_in_the_request_body_cannot_choose_the_owner`, `editing_a_workflow_is_not_a_transfer`, `an_unauthenticated_create_falls_back_to_the_configured_default_owner`, `with_nothing_configured_a_workflow_is_recorded_unowned`, `creating_a_cron_job_records_the_authenticated_caller_as_its_owner`, `creating_a_schedule_records_the_authenticated_caller_as_its_owner`.
- `cargo test -p librefang-api --test workflows_routes_integration --test owner_scoped_routes_integration --test config_schema_golden` — 102 pass.
- Drift guards: `dead_route_audit_test`, `openapi_path_coverage_test`, `openapi_spec_test` all pass and `openapi.json` is **unchanged** — the workflow/cron responses are declared as `JsonObject`, so the added fields are not a schema change and `xtask/baselines/openapi.sha256` needs no edit. `kernel_config_schema.golden.json` was regenerated by its own `--ignored regenerate_golden` path for the `default_owner` key (+7 lines).

One unrelated flake seen once and green on re-run: `agent_context::tests::poisoned_cache_lock_recovers_cached_context`, last touched by #7029, in a file this PR does not modify.

Live daemon and real-LLM paths were not exercised — human-only.

## Deferred, and what #7746 needs

**Deferred approvals stay unowned, and that is the one decision I did not make.** `Kernel::build_deferred_tool_exec_context` passes `acting_principal: None`, with the reasoning recorded at the call site rather than only here. `DeferredToolExecution` is persisted, so a principal written into it is resolved at defer time and consumed at resume time — possibly after a restart, a config reload, or the deletion of the user it names — making it an assertion about the past presented as a live one. The three candidate answers are (a) stamp it and accept the staleness, (b) re-resolve at resume and fail closed, (c) refuse to defer a tool call that needs an owner. Choosing between them requires the enforcement semantics this increment deliberately lacks: with nothing reading the owner, all three are indistinguishable. Concretely, a `workflow_create` or `cron_create` that goes through the approval gate is recorded unowned.

**Also not in this PR:** the remaining artifact classes listed above; any enforcement (per-principal quotas, "show me what I own" filters, cascade-on-delete, an authorization check on mutations of an owned artifact); dashboard and TUI surfacing of the owner.

**For #7746 (IAM mapping):** everything it needs is in place. `Principal::group_named` derives a group id from a `[[groups]] name`, so an IdP group claim mapped onto a local group name produces the same principal on every node with no lookup table. `KernelConfig::roles_for_user` and `groups_for_user` (from #7913) are the resolvers, and the `UserConfig.role` privilege ladder stays deliberately separate from group-conferred roles — #7746 is where an operator-defined mapping connects the two, and the test #7913 added to pin that separation should be updated consciously, not deleted. The `Role` arm this PR dropped is expressed there as IdP-fed groups.


houko:
Answering your question about `evolution.rs:1051`: **its own PR — #7929.**

You were right to ask rather than fold it in; `librefang-skills` is a different crate and a different domain from the ownership thread, which is exactly the CLAUDE.md bar for checking first.

Confirmed the impact before fixing: `skill_workshop/storage.rs`'s `CandidateKind::Create` arm passes `Some(&candidate.agent_id)`, so every skill approved through the workshop shipped a manifest claiming `agent-evolved` while the real producer sat in `.evolution.json` — recorded where nobody looks, wrong where everybody does. Negative control has it: restoring the literal fails `create_skill_records_the_caller_as_the_manifest_author` with `left: "agent-evolved", right: "scout"`.

The other item you flagged — `AgentEntry.parent` never persisted (`structured.rs:452/664/1048`) — I have not touched. Leaving it for you to raise as its own issue with the evidence you already have, rather than restating it second-hand.

houko:
Correcting my previous comment: I said I would leave the `AgentEntry.parent` finding for you to file. That was the wrong call — filing it is my job here, and "leave it for later" is exactly the deferral pattern CLAUDE.md warns about.

Filed as **#7930**, with the evidence verified rather than restated:

- No `parent` column exists. The v1 `CREATE TABLE agents` (`migration.rs:328-335`) has six columns, and the later `ALTER TABLE agents ADD COLUMN` migrations add only `session_id`, `identity`, `source_toml_path` (`:1859`, `:1865`, `:1871`).
- Hence `parent: None` hardcoded at all three hydration sites you cited (`structured.rs:452/664/1048`).
- The part that makes it user-visible rather than merely dead: `routes/agents/mod.rs:344` serialises `"parent_agent_id": e.parent…`, so after any restart the API asserts every agent has no parent. A populated field claiming a false thing, not an absent one a client can ignore.
- `kernel/tests.rs:1012` passes because it never round-trips through the store.

I also noted `children: Vec<AgentId>` has the same shape and the same problem, and that the fix has a schema decision to make (column vs manifest blob; whether `children` should be stored at all when it is derivable from `parent`).

The other thing you raised, `evolution.rs:1051`, is fixed in #7929.

PR #7933, opened 2026-08-25
docs(claude): record the honest CI-liveness test and the sentinel check

houko:
Documentation only. Two operational lessons from a session where both mistakes were made repeatedly.

## 1. Judging whether CI is executing

A workflow file GitHub cannot parse produces a run that completes as `failure` within seconds **without ever taking a runner**. A queue full of those is indistinguishable from throughput unless you exclude them by name.

`gh run list --status in_progress` counts *runs*, not jobs, and those startup failures pass through the in-progress state on their way to completed. A poll on "in-progress is non-zero" therefore reports a live pool against a dead one. That happened repeatedly here: every "runners alive" signal vanished on the follow-up query, while a manually dispatched probe sat queued for two and a half hours.

The only condition that means anything is now spelled out in the doc: a run whose workflow is not one of the known startup-failing ones reached success or failure, with a timestamp later than a baseline recorded before watching started.

Three monitors in this session got it wrong three different ways, and the doc names all three so the next one is not written the same way:

- polling the in-progress count
- comparing against a `jq` expression that yields `null` on an empty array, so it never equals the baseline and fires forever
- counting matches with a `select(...)` whose negated `test(...)` did not bind to the intended field, which silently counted everything

The rule the doc lands on: make the check print the timestamped run itself and read it, never a derived boolean.

## 2. A `cargo check` that finishes too fast

In a shared `CARGO_TARGET_DIR` — which this repo mandates — a check can report finishing in under a second with no `Checking` lines, because the artifacts are already present from another worktree. Touching sources frequently does not force a rebuild.

Appending `compile_error!("SENTINEL");` to the crate's `lib.rs` and confirming cargo reports it is one command and settles it. The doc says to do that before trusting such a green, and before reporting one.

Both entries sit beside the existing "Batch merging is runner-pool bound" and truncation bullets, which cover adjacent failure modes from the same incident.

## Changes

- `CLAUDE.md`: two new bullets in the GitHub Collaboration section.
- `changelog.d/documentation/claude-md-ci-liveness.md`.

No code, no workflow, no configuration.


CLAUDE.md
```diff
@@ -198,6 +198,16 @@ The rules you must not break without asking:
...
+  `gh run list --status in_progress` counts **runs**, not jobs, and those startup failures flicker through `in_progress` on their way to `completed` — so a poll on "in_progress != 0" reports a live pool against a dead one.
+  The only condition that means anything: **a run whose workflow is not one of the known startup-failing ones reached `success` or `failure`, with a timestamp later than a baseline you recorded before you started watching.**
+  Three separate monitors in one session got this wrong three different ways — polling `in_progress`, comparing a timestamp against a `jq` expression that yields `null` on an empty array (`null != baseline`, so it fires forever), and counting matches with a `select(...)` whose `test(...)|not` did not bind to the field intended.
+  Write the check to produce the *timestamped run itself*, print it, and read it — never a derived boolean.
+- **A suspiciously fast `cargo check` is not evidence of anything until you prove the compiler saw your tree.**
+  In a shared `CARGO_TARGET_DIR` a check can print `Finished in 0.35s` with no `Checking` lines because the artifacts are already there.
+  `touch`ing sources often does not force a rebuild.
```
houko, line 209: This section is introduced as "Full policy, with the incidents that produced each rule: `docs/development/github-collaboration.md`" — but neither new bullet (CI-liveness test, cargo-check sentinel) has matching detail added to that doc. Worth adding the incident write-up there too, or dropping the "full policy lives elsewhere" framing for these two if it's meant to be self-contained here. Leaving this as a comment rather than a fix since it's a judgment call on where the detail belongs, not mechanical.

---

houko (commented):
Fixed two mechanical CLAUDE.md-rule violations directly on the branch:
- The new cargo-check bullet had two sentences on one line, violating the "one sentence per line" prose rule for newly added prose.
- The changelog fragment filename didn't lead with the PR number per `changelog.d/README.md`'s naming convention.

One remaining item left as a comment below since it's a judgment call, not mechanical.

---

PR #7947, opened 2026-08-27
feat(dashboard): edit a cron or proactive schedule from the agent panel

DaBlitzStein:
## What the panel did

`AgentSchedulePanel` could only edit a *continuous* schedule. For the other two modes it rendered a dead end:

```tsx
// Periodic / proactive: editing those requires manifest changes
// we don't yet surface here. Hide the toggle rather than offering
// a button that silently overwrites the user's schedule with continuous.
<span …>{t("agents.detail.schedule_manifest_only")}</span>   // "manifest-controlled"
```

So a cron agent or a proactive agent showed its schedule and then told the operator to go edit `agent.toml` by hand.

## What it does now

That placeholder is replaced by a real Edit button for each of the two modes:

- **Periodic** — a cron expression field, refusing a blank expression.
- **Proactive** — a comma-separated conditions field, refusing an empty list.

Both write through the existing `PATCH /api/agents/{id}` path the continuous editor already uses. No new endpoint.

### Why the empty guard matters, and why it is not optional

The conditions field is **not seeded from the live schedule** — the panel has no read of the current condition list to seed it with. So an empty field means *"the operator typed nothing"*, never *"the operator wants no conditions"*.

Without the guard, opening Edit on a proactive agent and pressing Save would send `conditions: []`, wipe a schedule the operator never looked at, and report it as a success with a green toast. The cron field has the same shape and the same guard, for the same reason.

Clearing conditions on purpose is still possible by switching the schedule mode away from proactive — an explicit choice rather than the side effect of opening a drawer.

A hint under the field states the non-seeding out loud rather than leaving the operator to discover it: *"This replaces the full condition list — it isn't pre-filled with the current one."*

## Why one PR and not two

The two editors replace **one** placeholder that covered both non-continuous modes. Landing them separately leaves the panel in a state where one mode is editable while the other still renders a "manifest-controlled" label whose i18n key is already half-removed. They go together.

## Verification

`pnpm vitest run src/lib/__tests__/ src/components/AgentSchedulePanel.test.tsx` → **34 passed**, including the three dashboard i18n gates (missing keys, uncopied fallbacks, dead keys).

i18n: the three `proactive_conditions*` keys are present in all five locales; `schedule_manifest_only` is removed from all five, since this PR deletes its only use.

`pnpm typecheck` → clean.

## Correction to the earlier description

The previous body framed this as a one-line guard against a pre-existing bug in `saveConditions`. That was wrong: `saveConditions` does not exist on `origin/main` — this PR introduces it, along with the editor it belongs to. The guard is part of the new editor's contract, not a fix to shipped code. The diff has not changed shape; the description had understated it.


houko (commented):
Merging — the empty-conditions guard is the right fix, and the whitespace-and-commas-only test (`" , ,  "` filtering down to an empty list) covers the case a bare `.trim()` check would have let through.

One thing worth following up on, which your own comments already identify: the conditions editor submits a *replacement* list while never showing the current one.

```tsx
// Same limitation as the cron draft above: the schedule summary string
// doesn't carry the current condition list, so this can only submit a
// *replacement* list, not show what's live today.
```

The guard stops the zero-condition wipe, but not the partial one: an agent with four conditions, an operator who opens Edit and types one, and Save replaces all four with that one — a successful save, no warning, three conditions gone.
That is the same failure this PR fixes, one step short of empty.

Two ways out, either fine: read the live conditions (the manifest is already fetched by `AgentManifestForm`, so the data exists on the client) and seed the textarea from them, or label the field as replacing rather than editing so the operator knows what Save costs.
Seeding is better — it makes the field an editor rather than a trapdoor — but the label alone would close the surprise.

Not blocking, since today's behaviour is strictly better than what it replaces.

PR #7965, opened 2026-08-29
fix(api): report restored backup entries with archive separators

houko:
`main` is red on `Test / Windows` (shard 1/2) after #7174. This fixes one of the two defects and makes the other diagnosable.

## Fixed: restored entries carried the host separator

`restore_backup_blocking` pushed `entry_name.to_string_lossy()` onto the restored list. `entry_name` is a `Path`, so it renders with `\` on Windows and the reported entries stop matching the `/`-separated archive keys every caller compares against — `archive_entry_key` is the normalisation the classification filters in the same loop already go through.

```
routes::backup::tests::restore_accepts_a_small_highly_compressible_entry
  left: ["data\\a2a_tasks.db-shm"]
 right: ["data/a2a_tasks.db-shm"]
```

New `restored_entries_are_reported_with_archive_separators` drives a real archive through `restore_backup_blocking`, asserts the literal `data/cron_jobs.json`, and separately rejects any host separator in the result — so a platform that renders `\` cannot quietly agree with itself the way a `Path::join`-built expectation would.

## Not fixed yet, but no longer invisible

The same Windows run fails two round trips in `pairing_backup_extra_test`:

```
restore_round_trips_an_agent_workspace_file_to_its_original_path
restore_without_options_writes_every_component_back
  restore failed: {"error":"Backup restore incomplete","restored_files":20,"error_count":2}
```

Two entries fail their `create` or `extract` step. I could not determine which two: those reasons only reach `tracing::error!(errors = ?errors, ...)`, no subscriber is installed in the integration tests, and the 500 body deliberately carries counts only. The CI log therefore contains the count and nothing else.

Ruled out by reading the code rather than guessing:
- The compression-ratio and per-entry/total size guards return `Err` and abort the whole restore, so they cannot produce a partial result with `error_count: 2`.
- `prepare_restore_target` failures propagate with `?` — also an abort, not a per-entry error.
- `restore_root` uses `Path::strip_prefix`, which is component-based and correct on Windows.

So the failures are `open_restore_target` or `std::io::copy`. Rather than change security-sensitive extraction code on a hypothesis I cannot test — the Windows lane runs only on push to `main`, never on a PR — the harness now installs a `tracing_subscriber` at `ERROR` with a test writer. The next `main` run names both entries and their OS error.

## Verification

- `cargo test -p librefang-api --lib routes::backup` — 14 passed
- `cargo test -p librefang-api --test pairing_backup_extra_test` — 29 passed
- `cargo fmt --check` — clean

Windows is not reproducible locally; this change is separator-only plus test instrumentation, so Unix green is meaningful for the part it fixes.

## Deferred

The two `create`/`extract` failures, deliberately — with the evidence path landing in this PR rather than a guess.

### DIFF
--- a/changelog.d/fixed/7965-backup-restore-entry-separators.md
+++ b/changelog.d/fixed/7965-backup-restore-entry-separators.md
@@ -1,0 +1,2 @@
+Backup restore now reports restored entries with the archive's `/` separators on Windows rather than the host's, so a client matching the reported list against the component names it asked for no longer sees every entry as unrecognised.
+(#7965) (@houko)
--- a/crates/librefang-api/src/routes/backup.rs
+++ b/crates/librefang-api/src/routes/backup.rs
@@ -980,7 +980,11 @@ fn restore_backup_blocking(
                 "archive exceeded the total decompression limit".to_string(),
             ));
         }
-        restored.push(entry_name.to_string_lossy().to_string());
+        // `entry_name` is a `Path`, so rendering it hands back `\` separators on Windows and the
+        // restored list stops matching the `/`-separated archive keys every caller compares
+        // against. `archive_entry_key` is the normalisation the classification filters above
+        // already go through.
+        restored.push(archive_entry_key(&entry_name));
     }
 
     Ok(RestoreOutcome {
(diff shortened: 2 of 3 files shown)


houko:
Narrowing from the newest `main` Windows run (33242909288), which reproduces the same failures on `a02ee8293`.

Shard 2 carries a fourth failure I had not seen in the earlier run:

```
restore_of_the_data_component_covers_every_data_entry
  restore failed: {"error":"Backup restore incomplete","restored_files":9,"error_count":2}
```

This one restores **only the `data` component** and still hits exactly two per-entry errors, where the full round trips restore 20 and hit the same two. **Both failing entries are therefore inside `data/`** — the count does not change when every other component is filtered out.

That is consistent with the two entries being files the live test kernel holds open on Windows (the SQLite database and its sidecar), which is a real product limitation on that platform rather than a test artefact: a restore cannot truncate a file another handle owns.

What I still cannot settle by reading, and why the instrumentation in this PR is the right next step rather than a guess:

- `std::fs::write` (pre-#7174) and `open_restore_target` (post-#7174) both open with `write | create | truncate`, and the `O_NOFOLLOW` addition is Unix-only — so the open flags are identical on Windows and cannot by themselves explain a regression.
- The one thing #7174 did change on the Windows open path is the root: targets are now joined onto `std::fs::canonicalize(&home_dir)`, which returns a `\\?\` verbatim path there, instead of the plain `home_dir`.

Whether that path spelling is what flipped these two entries from success to failure is exactly what the captured `tracing::error!(errors = ?errors, ...)` line will say, by naming both entries and their OS error. Changing the extraction root on that hypothesis, in code that exists to bound a zip bomb and reject symlink escapes, without being able to run Windows, is not a trade I want to make blind.

Summary for `main`: shard 1 fails 3 tests (one fixed here), shard 2 fails 1. All four are #7174 backup-restore regressions; nothing else on Windows regressed across the 22 merges.

PR #7971, opened 2026-08-29
fix(channels): delete stale sidecar and restore /goal command bridge

DaBlitzStein:
## Summary

- `DELETE /api/channels/sidecar/{name}` decided whether a channel existed by trying to strip its block from the root `config.toml`, while the dashboard's delete button is driven by `configured` from `GET /api/channels`, which the daemon serves from the in-memory config.
  These disagree in two real states, both returning `404 no configured sidecar channel named X` for a channel that was actually running: a sidecar declared via an `include = [...]` file (kernel merges it into the live config, so it spawns and supervises normally, but the root `config.toml` has no block to remove), and a `config.toml` edited or rewritten out from under a running daemon (in-memory config and supervisor still carry the channel with nothing left on disk to strip).
  `remove_sidecar_block_anywhere` now falls back to the file that actually declares the entry, and the second case is now treated as a reconcile — falling through to a config reload plus forced bridge restart to drop the row and stop the orphaned child process.
  A name declared in no config file and absent from both the live config and the adapter map still returns 404.
- A conflict resolution had taken the `main` side of a diff and dropped `create_and_start_goal` from the channel bridge trait, its adapter implementation, and the `/goal` command match arm — while leaving the `goal` `CommandDef` (`telegram_menu: true`) standing, so Telegram still advertised `/goal` in its command menu but answered "Unknown command: /goal".
  Restores `ChannelBridgeHandle::create_and_start_goal`, the `"goal"` match arm (parsing `--loop-engineering`, resolving the target agent through the same `resolve_for_command` path the `/workflow` arm uses), and the `KernelBridgeAdapter` implementation, which persists the goal document first and then schedules the run via `KernelApi::start_goal_run` — so a failed schedule (kernel self-handle unset) is reported distinctly from a failed create, and the operator is told the goal exists and just needs a daemon restart rather than losing the request.

## Verification

- `cargo check -p librefang-channels -p librefang-api --lib` — clean, no warnings.
- During cherry-pick onto current `main`, two drift issues surfaced against the branch the commits were originally authored on and were fixed in place:
  - `resolve_for_command()` is a synchronous closure (used without `.await` at every other call site in `bridge.rs`); the `/goal` arm's stray `.await` has been removed.
  - `KernelApi::start_goal_run` takes 3 arguments (`goal_id`, `agent_id`, `max_iterations`) on current `main` — the call site's extra `verify_agent_id` / `verify_max_retries` / `evaluator_model` arguments (none of which exist anywhere else in the codebase) have been removed to match, mirroring the call in `routes/goals.rs`. `loop_engineering` is still persisted on the goal document.
- New tests carried over from the original commits: `channel_remove_test.rs` (sidecar deletion for both the include-file and reconcile cases) and `test_handle_command_goal_is_dispatched` (asserts `/goal` reaches its arm instead of the unknown-command fallthrough).

## Out-of-scope follow-ups

None.


houko (commented):
Two unrelated changes in one PR, and the `/goal` half is now superseded by something already on `main`.

**`create_and_start_goal` duplicates the kernel helper, in the API layer.**

#7994 merged a few minutes ago with the same trait method on the same impl, implemented as a delegation:

```rust
async fn create_and_start_goal(&self, agent_id: AgentId, description: &str, loop_engineering: bool)
    -> Result<String, String>
{
    librefang_kernel::goal_runner::create_and_start_goal(self.kernel.as_ref(), agent_id, description, loop_engineering)
        .map(|launch| launch.message(description))
}
```

This branch instead hand-builds the goal document inline — `uuid::Uuid::new_v4()`, an ad-hoc `serde_json::json!` with nine fields, a `structured_modify` against `GOALS_STORAGE_KEY`, then `start_goal_run`.
That is the kernel's own persistence format reimplemented one layer up, and the two will drift the first time a field is added to `Goal`: the kernel path gets it, this one does not.
It is also a straight conflict — same file, same trait, same method name.

The concrete risk is not hypothetical.
#8010 (also merged) exists because goals in the store had UUID-typed `Option` fields written as `""` rather than `null`, which made `load_goal` return `None` and every start/pause/resume answer 500.
A second writer constructing the document by hand is exactly how that class of mismatch gets reintroduced.
Rebasing onto `main` and dropping this method should leave the sidecar-delete change standing on its own.

**The sidecar-delete reconciliation is the good half, and it should be its own PR.**

The reasoning is right: `GET /api/channels` answers from the in-memory config and the supervisor owns the child, so "no block left on disk" and "no such channel" are genuinely different states, and 404 for a channel that is actively delivering messages is the wrong answer.
Resolving `include = [...]` files before concluding the block is absent is the other half of the same correctness, and the differentiated error string on the reconcile path is a nice touch.

One question on that half: the `live` check reads `state.kernel.config_ref().sidecar_channels` and `channel_adapters_ref()`, then falls through to `reload_channels_from_disk`.
If the block genuinely is absent from every file on disk *and* absent from the merged config but an adapter is still registered — an orphaned child from a previous config — does the reload actually stop it, or does it only reconcile channels the new config still knows about?
That is the case the `warn!` describes, and it is worth a test alongside the include-file one you already added.

houko (changes requested):
Both sidecar-delete bugs are well diagnosed, the reconcile reasoning ("nothing left on disk is NOT the same as no such channel") is right, and the two integration tests pin the real scenarios including the rodela incident. Three things before merge.

## Blocking

**1. `remove_sidecar_block_anywhere` stops at the first hit, so a doubly-declared sidecar comes back.**

```rust
if super::sidecar_toml::remove_sidecar_block(config_path, name)? {
    return Ok(true);
}
for included in included_files_with_sidecars_blocking(config_path)? {
    if super::sidecar_toml::remove_sidecar_block(&included, name)? {
        return Ok(true);
    }
}
```

The doc comment's own premise is that the root can *shadow* an included entry — which means both files can declare `name` at once. In that case this removes the root block, returns, and the included block survives. The subsequent `reload_channels_from_disk` re-merges it, so the delete reports `status: "removed"` and the channel is still there — a fresh version of the bug this PR is fixing.

Remove from every file that declares it and return whether any removal happened:

```rust
let mut removed = super::sidecar_toml::remove_sidecar_block(config_path, name)?;
for included in included_files_with_sidecars_blocking(config_path)? {
    removed |= super::sidecar_toml::remove_sidecar_block(&included, name)?;
}
Ok(removed)
```

Worth a third test: the same name in the root *and* an included file, asserting both blocks are gone and the repeat delete 404s.

**2. The `/goal` half of the title is not in the diff.**

`crates/librefang-channels/src/bridge.rs` gains three comment blocks on `test_handle_command_goal_is_dispatched` and no code. The test existed and passed before this PR; nothing is restored. The comments are good — the "a conflict resolution once dropped the handler while leaving the declaration in place" note is worth having — but the title ("…and restore /goal command bridge") and the PR body should say "document why this test exists" rather than implying a fix. As written the release notes would claim a repair that did not happen.

**3. No changelog fragment.**

Two operator-visible fixes (delete works for include-declared sidecars; delete reconciles an orphaned live child instead of 404-ing) and `changelog.d/` has no entry. Per `CLAUDE.md` these go in as prose explaining *why*, and both have a concrete incident behind them, so they will read well: `changelog.d/fixed/7971-....md`.

## Non-blocking

- The `live` check ORs "still in the in-memory config" with "still in `channel_adapters_ref()`". Those are two genuinely different reconciliation states, and the `warn!` collapses them into one message. Logging which one tripped would make the next rodela-style incident diagnosable from the log alone.
- The two-branch error string in the bridge-restart failure path is good, but the `live` case now returns 500 having deliberately done nothing to disk. That is correct, though the message ("config.toml had no block to remove and the bridge restart failed — the sidecar is still running") is the one an operator will paste into an issue — worth adding what to do next, e.g. that a daemon restart will drop it.

houko:

No commits of your own since my review on 2026-09-01T13:48:30Z; the only thing that moved the branch tip is a merge from `main`, which the `auto-update-branches` cron pushes by itself. The PR's "updated" timestamp has therefore advanced while the blocking items have not been touched. Worth flagging because that timestamp is what makes a PR look answered from outside — I nearly mis-triaged a dozen PRs on exactly that signal today.

Still open from that review:

- 1. `remove_sidecar_block_anywhere` stops at the first hit, so a doubly-declared sidecar comes back
- 2. The `/goal` half of the title is not in the diff
- 3. No changelog fragment

DaBlitzStein:
Pushed `e53172662`: merged current `main`, which is now an ancestor of this branch.

No hand-written merge resolution — the diff against git's own automerge tree is empty.

Verification on the pushed tip, all seven tests requested, started and passed:

```
Starting 7 tests across 116 binaries (3381 tests skipped)
PASS [129.295s] ws::tests::goal_command_strips_the_loop_engineering_flag
PASS [129.299s] ws::tests::goal_command_creates_and_reports_the_goal
PASS [137.498s] channel_remove_test delete_reconciles_a_live_sidecar_that_is_no_longer_on_disk
PASS [162.683s] ws::tests::goal_command_without_a_description_returns_usage
PASS [140.465s] channel_remove_test delete_unknown_sidecar_404s
PASS [150.102s] channel_remove_test delete_removes_sidecar_declared_in_an_included_file
PASS [156.340s] channel_remove_test delete_removes_configured_sidecar_then_404s_on_repeat
Summary [286.017s] 7 tests run: 7 passed (7 slow), 3381 skipped
```

Cardinality checked rather than assumed: four `channel_remove` plus three `goal_command` is seven asked for, seven started, seven passed.

Worth recording, because it nearly produced a false report: an earlier run of this branch showed failures including `goal_command_creates_and_reports_the_goal`, and **those were not real**. The worktree had been reused and carried uncommitted changes from earlier work, so an hour of verification ran against a contaminated tree. Re-run on a clean tree from the remote tip, the same tests pass and the merge SHA is unchanged. Any failure reported for this branch before that point should be disregarded.

The 129-162 s durations on tests that are normally fast are machine load, not this branch: several parallel builds were running.

DaBlitzStein:
Pushed `53b83195c`, a fast-forward over the previous tip that folds in work which had been sitting uncommitted in a working tree: `remove_sidecar_block_anywhere` now also sweeps the files pulled in by `include`, with its test and changelog fragment.

**The recovered test was failing, and the fixture was wrong rather than the implementation.** The fixture wrote `TELEGRAM_BLOCK` and *then* `include = ["channels.toml"]`. Since the block ends inside the `[sidecar_channels.env]` table, that `include` lands under `sidecar_channels.env` in TOML rather than at the root, so `document.get("include")` returns `None`, the included file is never scanned, and the block survives.

Confirmed by parsing before and after the fix: root `include` present with `['channels.toml']`, and `env` back to just `{'ALLOWED_USERS': '1,2'}`. The sibling test that already passed puts `include` first, which is what pointed at the fixture rather than the code.

Verification on the pushed tip — eight requested, eight started, eight passed:

```
PASS [111.017s] ws::tests::goal_command_strips_the_loop_engineering_flag
PASS [111.023s] ws::tests::goal_command_creates_and_reports_the_goal
PASS [127.925s] ws::tests::goal_command_without_a_description_returns_usage
PASS [132.929s] channel_remove_test delete_reconciles_a_live_sidecar_that_is_no_longer_on_disk
PASS [ 25.089s] channel_remove_test delete_removes_configured_sidecar_then_404s_on_repeat
PASS [ 25.083s] channel_remove_test delete_removes_sidecar_declared_in_an_included_file
PASS [  3.184s] channel_remove_test delete_unknown_sidecar_404s
PASS [  8.243s] channel_remove_test delete_strips_a_sidecar_declared_in_root_and_include_at_once
Summary [137.908s] 8 tests run: 8 passed (4 slow), 3381 skipped
```

The last one is the recovered test, green.

Stated rather than glossed: the pre-commit hook ran `rustfmt`, which reformatted one `std::fs::write` call across three lines **after** that run. The fixture ordering was re-checked afterwards and is intact (`include` first, line 236), and the change is formatting only — but the tree that was tested and the tree that was committed differ by that line break.

DaBlitzStein:
Addressed at `065b4593a` and `003658cc9`.

**1 — `remove_sidecar_block_anywhere` removes from every file that declares the name** and returns whether any removal happened, so a sidecar declared in both the root and an included file no longer resurvives the `reload_channels_from_disk` re-merge with the delete reporting `status: "removed"`.

`delete_strips_a_sidecar_declared_in_root_and_include_at_once` is the third test you asked for. Restoring the early return fails it and leaves the other four green.

Writing it surfaced a second problem in the same function that your review did not name: the include scan ran **after** the root block had already been rewritten, so an unreadable include file failed the whole call with the root already stripped — a half-applied delete behind a 500 the operator could not retry out of. Collecting the include list first makes the only read-side failure abort before anything on disk changes.

**2 — the title and body no longer claim a repair that did not happen.** The `/goal` half is three comment blocks on a test that already existed and passed; it is now described as documenting why the test exists. The "a conflict resolution once dropped the handler while leaving the declaration in place" note stays, since that is the thing worth having.

**3 — `changelog.d/fixed/7971-sidecar-delete-include-and-reconcile.md`** covers both operator-visible fixes, written as the incidents behind them rather than a restatement of the title. Trailing newline included.

Non-blocking: the `warn!` now records **which** reconciliation state tripped rather than collapsing "still in the in-memory config" and "still in `channel_adapters_ref()`" into one message, and the 500's error string says what to do next.

Note for re-running: this suite boots a kernel per test, so under load it fails at the harness line with `BootFailed("Memory init failed: … database is locked")` rather than at an assertion. All five pass serially on an idle machine.

Ready for another look.

crates/librefang-api/src/routes/channels.rs
```diff
@@ -1806,12 +1837,41 @@ pub async fn delete_sidecar_channel(
...
+    // sidecar is delivering messages. Reconcile instead: fall through to the
+    // reload, which drops the row and stops the orphaned child.
+    let in_live_config = state
+        .kernel
+        .config_ref()
+        .sidecar_channels
+        .iter()
+        .any(|sc| sc.name == name);
+    let has_adapter = state.kernel.channel_adapters_ref().contains_key(&name);
+    let live = !removed && (in_live_config || has_adapter);
```
houko, line 1947: `live` treats "the include scan found no block" as "nothing is declared on disk", but `included_files_with_sidecars_blocking` only reads the **first level** of `include` and only substring-matches `[[sidecar_channels]]`. Every declaration it cannot see now produces a false success instead of the old 404.

Nested include: `config.toml` has `include = ["a.toml"]`, `a.toml` has `include = ["channels.toml"]`, and `channels.toml` declares `[[sidecar_channels]] name = "email"`. The kernel resolves includes recursively (`crates/librefang-kernel/src/config.rs:561`, "Recursively resolve includes in the included file"), so the channel is fully live. On `DELETE /api/channels/sidecar/email` the scan reads only `a.toml`, finds no header, so `removed == false`; `in_live_config == true`, so `live == true`; the handler skips the 404, `reload_config()` re-merges the still-present block, the forced `reload_channels_from_disk` respawns the child, and the response is `{"status":"removed"}`.

Same outcome for the inline form in a first-level include — `sidecar_channels = [{ name = "email", command = "..." }]` — which the kernel merges as an ordinary array of tables but the `body.contains("[[sidecar_channels]]")` heuristic never matches.

The operator is told the channel is gone while its sidecar keeps delivering messages, which is strictly worse than the 404 this replaces. The failure-branch text at the bottom of the handler ("the sidecar is still running; a daemon restart will drop it") is wrong in the same cases — a restart re-merges the block.

Suggestion: only take the reconcile branch when the scan is known to be exhaustive — resolve includes recursively and parse rather than substring-match, so `removed == false` really does mean "declared nowhere reachable".

crates/librefang-api/src/routes/channels.rs
```diff
@@ -966,6 +966,37 @@ fn sidecar_discovery_rows() -> Vec<serde_json::Value> {
...
+/// either behind means the reload re-merges the survivor and the delete
+/// reports `removed` while the channel comes back. Every file that declares
+/// the name is stripped; `false` still means "declared nowhere", which is the
+/// only case that deserves a 404.
+fn remove_sidecar_block_anywhere(
+    config_path: &std::path::Path,
+    name: &str,
+) -> Result<bool, String> {
+    // Scan the includes before the first write: an unreadable include file then fails the whole delete, instead of leaving the root block already stripped behind a 500 the operator cannot retry out of.
+    let included_files = included_files_with_sidecars_blocking(config_path)?;
```
houko: This makes a *missing* include file fatal to every sidecar delete, including sidecars declared entirely in the root `config.toml`.

`included_files_with_sidecars_blocking` returns `Err("failed to read included config ...")` for any unreadable entry, and here that error is propagated before `remove_sidecar_block` touches the root file. But a missing include is a state the kernel deliberately tolerates: `resolve_config_includes` fails, `load_config` logs `Config include resolution failed, using root config only` (`crates/librefang-kernel/src/config.rs:120-126`) and the daemon boots normally.

Concretely: `config.toml` = `include = ["overrides.toml"]` (file absent) plus `[[sidecar_channels]] name = "telegram"`. The daemon runs the telegram sidecar. `DELETE /api/channels/sidecar/telegram` now returns 500 and strips nothing — a delete that worked before this PR is permanently broken until someone hand-edits `config.toml`. The same 500 also replaces the intended 404 on a repeat delete once the declaring include file has been removed.

Fail-closed is right for the `configure` caller, which refuses the whole save whenever any include declares a sidecar. The delete only needs the list to know *where else to strip*, so suggest tolerating `ErrorKind::NotFound` per include entry on this path — mirroring the kernel's own tolerance — while still surfacing permission and parse errors.

crates/librefang-api/src/routes/channels.rs
```diff
@@ -966,6 +966,37 @@ fn sidecar_discovery_rows() -> Vec<serde_json::Value> {
...
+    let mut removed = super::sidecar_toml::remove_sidecar_block(config_path, name)?;
+    for included in included_files {
+        removed |= super::sidecar_toml::remove_sidecar_block(&included, name)?;
```
houko: This turns `included_files_with_sidecars_blocking` from a read-only check into a **write** target, and it resolves include paths the kernel itself refuses to load.

`resolve_config_includes` rejects absolute paths, `..` traversal, and anything escaping the config directory (`crates/librefang-kernel/src/config.rs:516-543`: "rejects absolute path" / "rejects path traversal" / "escapes config directory"). The scan applies none of that: `raw_path.is_absolute()` is honoured as a path to read, and `parent.join(raw)` walks upward without complaint.

So with `include = ["/etc/librefang/channels.toml"]` (or `["../shared/channels.toml"]`) the kernel runs on the root config only — nothing in that file is live — yet the delete resolves it, matches the `[[sidecar_channels]]` substring, rewrites the file through `atomic_write`, and answers 200 `removed`. A TOML file outside the daemon's config directory has been modified, and the response describes a runtime change that never happened.

Suggest applying the same absolute / traversal / escape validation the kernel uses before any include file is written, so the delete can only mutate files the kernel would actually merge.

crates/librefang-api/src/routes/channels.rs
```diff
@@ -966,6 +966,37 @@ fn sidecar_discovery_rows() -> Vec<serde_json::Value> {
...
+/// reports `removed` while the channel comes back. Every file that declares
+/// the name is stripped; `false` still means "declared nowhere", which is the
+/// only case that deserves a 404.
+fn remove_sidecar_block_anywhere(
+    config_path: &std::path::Path,
+    name: &str,
+) -> Result<bool, String> {
+    // Scan the includes before the first write: an unreadable include file then fails the whole delete, instead of leaving the root block already stripped behind a 500 the operator cannot retry out of.
+    let included_files = included_files_with_sidecars_blocking(config_path)?;
+    let mut removed = super::sidecar_toml::remove_sidecar_block(config_path, name)?;
```
houko: The multi-file walk is atomic per file but not across files: the root is stripped first, and a failure in any include write propagates through `?` after the root has already been rewritten.

Take the case the new `delete_strips_a_sidecar_declared_in_root_and_include_at_once` test covers — `telegram` in both `config.toml` and `channels.toml` — but with `channels.toml` in a directory the daemon cannot write (read-only mount, or the file is root-owned). The root block is stripped, `remove_sidecar_block(&included, ...)` fails inside `atomic_write` at `open`/`rename`, and the request returns 500 with the root already modified.

The retry is then worse than the first attempt. If the include has become unreadable rather than unwritable, the pre-write scan errors and the request 500s forever; if it is readable but still unwritable, the root has nothing left to strip so `removed == false`, `in_live_config == true` still holds, `live` is taken, and the endpoint answers 200 `removed` while the surviving include declaration is re-merged by the reload.

The comment on line 991 claims the pre-write scan avoids "leaving the root block already stripped behind a 500 the operator cannot retry out of", but it only covers *unreadable* includes — an *unwritable* one lands exactly in the state it describes. Snapshotting each file's contents before the first write and restoring them all on any failure (the pattern `write_sidecar_configuration` already uses via `read_file_snapshot` / `restore_sidecar_file`) would make the whole walk all-or-nothing.

crates/librefang-api/src/routes/channels.rs
```diff
@@ -1806,12 +1837,41 @@ pub async fn delete_sidecar_channel(
...
+    // 404 there tells the operator the channel does not exist while its
+    // sidecar is delivering messages. Reconcile instead: fall through to the
+    // reload, which drops the row and stops the orphaned child.
+    let in_live_config = state
+        .kernel
+        .config_ref()
+        .sidecar_channels
+        .iter()
+        .any(|sc| sc.name == name);
+    let has_adapter = state.kernel.channel_adapters_ref().contains_key(&name);
```
houko, line 1946: The `has_adapter` half of `live` never converges: nothing on this path removes the stale adapter entry, so the same DELETE keeps returning 200 `removed`.

`mesh.channel_adapters` is only ever cleared by the `ReloadChannels` dispatch (`crates/librefang-kernel/src/kernel/config_reload_ops.rs:308`), and that dispatch does not run when the plan diff is empty — which is exactly the situation this branch was added for ("the plan diff can legitimately come back empty when the in-memory config already matched disk and only the supervisor was out of step"). `start_channel_bridge_with_config` only *inserts* adapters; it never removes a key that no longer has a config entry.

So for `in_live_config == false`, `has_adapter == true`, nothing on disk: `live` is taken, `reload_config()` yields no `sidecar_channels` diff, the map is not cleared, `reload_channels_from_disk` repopulates from a config that has no such channel, and the stale `email` key is still in the map when the request returns. A second DELETE takes the same branch and answers 200 `removed` again — and every one of those calls tears down and restarts the entire bridge, killing and respawning every *other* sidecar child as collateral.

Both new tests exercise only the `in_live_config` variant, where the plan diff does clear the map and the follow-up delete correctly 404s. Suggest removing the target's plain and `name:name` keys from `channel_adapters_ref()` on the reconcile path (and covering the adapter-only case in a test), so the state actually reconciles.

crates/librefang-api/src/routes/channels.rs
```diff
@@ -966,6 +966,37 @@ fn sidecar_discovery_rows() -> Vec<serde_json::Value> {
...
+/// the name is stripped; `false` still means "declared nowhere", which is the
+/// only case that deserves a 404.
+fn remove_sidecar_block_anywhere(
+    config_path: &std::path::Path,
+    name: &str,
+) -> Result<bool, String> {
+    // Scan the includes before the first write: an unreadable include file then fails the whole delete, instead of leaving the root block already stripped behind a 500 the operator cannot retry out of.
+    let included_files = included_files_with_sidecars_blocking(config_path)?;
+    let mut removed = super::sidecar_toml::remove_sidecar_block(config_path, name)?;
+    for included in included_files {
```
houko: The walk visits each file once, but `remove_sidecar_block` strips only the **first** matching entry inside a file: `(0..aot.len()).find(...)` followed by a single `aot.remove(idx)` (`crates/librefang-api/src/routes/sidecar_toml.rs:268-277`).

So if one file declares `[[sidecar_channels]] name = "email"` twice — a hand-edit, or a copy-paste of a block — the delete removes one, returns `true`, the handler answers 200 `removed`, and the reload re-merges the surviving duplicate and respawns the sidecar. That is the same "reported `removed` while the channel comes back" failure the cross-file walk was added to prevent, just within a single file rather than across two.

The doc comment above asserts the stronger invariant ("Every file that declares the name is stripped"). Draining every match — a `while let Some(idx) = ...` loop, or a `retain` on the array of tables — would make the comment true and close the last variant of this bug.

crates/librefang-api/src/routes/channels.rs
DaBlitzStein: Confirmed and fixed.
The scan now resolves includes recursively (nested includes and diamond graphs, matching `resolve_config_includes`) and real-parses each file for a `sidecar_channels` key instead of substring-matching `[[sidecar_channels]]`, so the inline `sidecar_channels = [{ ... }]` form is caught too.
Applies the same absolute-path and `..`-traversal rules the kernel enforces, so it never resolves a file the kernel itself would refuse to merge.
Added unit tests for a two-level-deep include and for the inline array-of-tables form, plus an HTTP-level integration test (`delete_strips_a_sidecar_declared_two_include_levels_deep`) that reproduces the exact false-404 you described and passes now.
Verified red first: reverted the recursion (capped at depth 1) and the real parse (back to substring match), all three new tests failed with the described symptom, then restored the fix and they pass.

crates/librefang-api/src/routes/channels.rs
DaBlitzStein: Confirmed and fixed.
The scan now takes a `tolerate_missing` flag: the delete path passes `true` so a vanished include (`NotFound`) is skipped rather than failing the whole scan, mirroring the kernel's own tolerance for a broken include chain.
`configure`'s conflict check keeps passing `false` and stays fail-closed, exactly as you said it should, since an unreadable include there might still declare a shadowing block it cannot verify.
Permission and parse errors still propagate on both paths.
Added `unreadable_included_config_is_skipped_when_tolerated` alongside the existing fail-closed test, and reran the full `channel_configure_test` suite (9 tests) to confirm the strict path is untouched.
Verified red: with the tolerance removed, the new test fails with the exact `NotFound` propagation you flagged, then passes once restored.

crates/librefang-api/src/routes/channels.rs
DaBlitzStein: Confirmed and fixed.
The recursive walk now skips (never follows) any include entry that is an absolute path or contains a `..` component, the same two rules `resolve_config_includes` enforces, before ever reading or writing through it.
Added `absolute_and_traversal_includes_are_skipped`, which seeds `include = ["/etc/passwd", "../escape.toml"]` and asserts neither is resolved.
Verified red: with the validation removed, the scan actually reads and TOML-parses `/etc/passwd` and the test fails on the resulting parse error, which is itself a demonstration of why this needed to be a hard skip rather than best-effort; restoring the check fixes it.

crates/librefang-api/src/routes/channels.rs
DaBlitzStein: Confirmed and fixed.
Every candidate file (root plus every included file that declares the block) is now snapshotted before the first write, and if any file's `remove_sidecar_block` call fails, every snapshot is restored via `restore_sidecar_file` before the error is returned — so a failure partway through the walk no longer leaves an earlier file's block already stripped.
Added `delete_restores_the_root_when_a_later_include_write_fails`: root and an include both declare the same name, the include's `sidecar_channels` is a scalar (not an array-of-tables) so its write deterministically errors after the root has already committed its strip, and the test asserts the root is byte-identical to its pre-request content afterward.
Verified red: with the restore loop replaced by a no-op, the test fails with the root left stripped, exactly the state you described; restoring the loop fixes it.

crates/librefang-api/src/routes/channels.rs
DaBlitzStein: Confirmed and fixed.
When `in_live_config` is false (the has_adapter-only orphan), the reconcile branch now removes the target's plain and `name:name` qualified keys from `channel_adapters_ref()` directly, rather than waiting on the `ReloadChannels` dispatch that never fires here because the plan diff is empty.
The `in_live_config == true` case is left to the existing reload-diff mechanism, which already clears the map correctly (covered by the pre-existing `delete_reconciles_a_live_sidecar_that_is_no_longer_on_disk` test).
Added `delete_clears_a_stale_adapter_entry_with_no_config_anywhere`, which inserts a stub adapter under both keys with no config anywhere and asserts both are gone after one delete, then that a repeat delete 404s.
Verified red: with the removal disabled, the test fails on the very first assertion (`the plain adapter key must be cleared directly`); restoring it fixes it.

crates/librefang-api/src/routes/channels.rs
DaBlitzStein: Confirmed and fixed.
`remove_sidecar_block` now drains every matching entry (a `while` loop over the array-of-tables) instead of removing only the first match, so the doc comment's "every file that declares the name is stripped" is actually true within a single file too.
Added `remove_drains_every_duplicate_entry_in_one_file`, seeding two `email` entries and one unrelated `telegram` entry in the same file, asserting both `email` entries are gone and `telegram` survives.
Verified red: with the old single-`find`-plus-`remove` logic restored temporarily, the test fails with one `"email"` entry still present; restoring the drain loop fixes it.

DaBlitzStein:
Every one of the 6 review threads on this PR now has an inline reply, and the branch is at `251766828`.

houko (approved):
LGTM. Both findings are fixed at the root rather than patched at the symptom.

**Nested includes / substring matching.** `included_files_with_sidecars_blocking` is now a real recursive walk: `MAX_INCLUDE_DEPTH = 10` pinned against the kernel constant with a comment saying so, a canonicalized `seen` set so a diamond or cycle terminates, and a genuine `toml_edit` parse testing `document.get("sidecar_channels")` instead of substring-matching the array-of-tables header. That closes both halves — the two-level chain and the inline `sidecar_channels = [{ … }]` form the kernel merges identically.

The tests are the good kind. `nested_include_is_resolved_recursively` states in its doc-comment how to reproduce the original bug ("reverting the recursion — calling `walk` only at depth 0's direct includes — fails this test with `hits.is_empty()`"), which is the difference between a test that guards the fix and one that merely passes alongside it. `absolute_and_traversal_includes_are_skipped` is a good catch that was not in my comment: the scan must skip exactly what `resolve_config_includes` rejects, or it would read — and later write to — a file the kernel would never load.

**The non-converging `has_adapter` orphan.** `adapters.remove(&name)` plus the `{name}:{name}` key, gated on `!in_live_config`, with a comment explaining why only that case needs the direct removal (when `in_live_config` is also true the reload's own diff clears the map, and there is a named test covering it). The `warn!` now carries `in_live_config` and `has_orphan_adapter` separately, so which of the two reconciliation states tripped is answerable from the log alone.

The `bridge.rs` hunk is comment and test additions only, and its `test_handle_command_goal_is_dispatched` comment records a real past incident — a conflict resolution dropping the `/goal` handler while leaving `telegram_menu: true` in the declaration, so the bot menu offered a command that answered "Unknown command".

All checks green; the `BLOCKED` state is the stale `CHANGES_REQUESTED` from my earlier round, which this clears.

One process note before merging: #8228 landed a large rewrite of `crates/librefang-channels/src/bridge.rs` after this PR's CI ran. Git reports the branch as mergeable and this PR only adds to the tests module, but the `main` ruleset has no strict status-check policy, so I am updating the branch to re-run CI against the new base rather than merging on a green that predates it.

DaBlitzStein:
Pushed `8cfb3b9d8` onto this branch. It is rescued work — it was sitting uncommitted in a worktree and was in no branch, so it would have been lost when that worktree was cleaned up. Flagging it explicitly because it arrives on a branch that has already been reviewed.

Two independent fixes, both in `routes/channels.rs`.

**The orphaned adapter key.**
The DELETE handler removed `{name}` and a guessed `{name}:{name}`, but `channel_bridge.rs:3057` registers the qualified key as `format!("{name}:{aid}")`, where `aid` is the *adapter's* `account_id` — not the channel's name.
The guess only ever matched the coincidence where those two happen to be equal, which is what a single-account deployment produces.
On a multi-account deployment `telegram:acct-42` survived every delete and every reload, which is the orphan this branch exists to converge.
It now removes every key the channel owns: the plain key, and anything under the `{name}:` prefix.
An adapter with no `account_id` registers only the plain key (`channel_bridge.rs` guards the qualified insert behind `if let Some(ref aid)`), so that case is covered by the same predicate rather than needing its own.

**The include walk was missing the kernel's third containment rule.**
The doc-comment on `included_files_with_sidecars_blocking` said it applied "the same two rules the kernel enforces". The kernel applies three — `config.rs:540` is `canonical.starts_with(&canonical_dir)`.
The two that were present are string checks on absolute paths and `..` components, and neither can see a symlink.
A relative symlink carrying no `..` passes both, then resolves wherever it points; if the target declares `sidecar_channels` it joins the list that `remove_sidecar_block_anywhere` **rewrites**, so the endpoint could edit a file outside the config directory — one the kernel itself refuses to merge.
The new check skips rather than propagating an error, unlike the kernel: this walk answers "where could a reachable block be stripped", and a file the kernel will not merge cannot be holding a live one.

On reuse rather than mirroring the rule: I looked at sharing the kernel's predicate and did not, for two reasons worth stating.
Extracting it would collapse the kernel's two distinct error messages (one for the include, one for the config dir) into one, which is a behaviour change in a PR that is not about that.
A cross-crate behavioural equivalence test would need `resolve_config_includes` made `pub`, which widens the kernel's API for a test.
The local helper names the exact kernel rule it mirrors and why the error semantics deliberately differ, which seemed the honest trade. Happy to revisit if you disagree.

**Verification.**
Both halves were confirmed failing first, against this branch without the fix, and for the right reason rather than a broken harness:

- `delete_clears_a_stale_adapter_entry_keyed_by_a_real_account_id` — "``email:acct-42`` must be cleared — the qualified key is keyed by the adapter's account id, not by the channel name". The other 8 tests in that file passed in the same run, so the new one discriminates.
- `symlinked_include_escaping_the_config_dir_is_skipped` — "a symlink resolving outside the config directory must not become a file the delete handler rewrites: `["/tmp/.tmpRp8tiT/elsewhere.toml"]`", naming the escaped path. Its control case, `symlinked_include_inside_the_config_dir_is_still_followed`, passed in that same run, so the check is not blanket-blocking symlinks.

Both green after. `cargo check --workspace --all-targets --exclude librefang-desktop` and `cargo clippy --workspace --all-targets --exclude librefang-desktop -- -D warnings` are both clean.
`cargo test -p librefang-api --lib` is 1163 passed with 4 failures, all of them test-kernel boots under load from concurrent builds on this host, all passing in isolation, and none in a file this change touches.
All 16 channel / sidecar / config integration targets pass, including `channel_remove_test` at 9/9.

DaBlitzStein:
Correcting the verification claim in my previous comment, because the way I obtained it was unsound even though the result stands.

I had written `cargo check … | tail -5; echo "EXIT=$?"`. `$?` there is `tail`'s exit status, not cargo's, so the "clean, exit 0" I reported for `check` and `clippy` was never actually read from cargo.

Both have now been re-run with the exit code read directly and on a freshly cleaned target directory: **`cargo check --workspace --all-targets --exclude librefang-desktop` and `cargo clippy … -- -D warnings` are exit 0, zero errors, 28 crates recompiled.** I also re-ran the tests after that clean rebuild — the two symlink cases pass, and `channel_remove_test` is 9/9. So the previous comment's conclusions are correct; only the evidence behind two of its lines was.

The clean rebuild mattered for a second reason worth recording. Two worktrees sharing one `CARGO_TARGET_DIR` had left this crate linking a `librefang-types` rmeta built from the other branch, which produced `unresolved import` errors for modules that plainly exist in the tree. Nothing to do with this change, but it is why the first pass was not trustworthy end to end.

Also worth noting about the adapter-key test, since it is the kind of red that can fool you: on the first attempt **all nine** tests in `channel_remove_test` failed on the same line of a shared helper. That was SQLite contention at test-kernel boot from concurrent builds on the host, not the defect. Run serially and retried until the kernel booted, exactly one test fails — the new one — with the message quoted in my previous comment, and the other eight pass. That is the run the red is taken from.

PR #7785, opened 2026-08-21
feat(goals): gate autonomous goal runs on a verifier and an evaluator

DaBlitzStein:
First of the five PRs from the #6505 split.

#6505 closed with the note that the goal runner with a verification loop is something the project wants, and that it should land in pieces against `main` rather than be lost.
This is the runner itself; the `/goal` CLI command and the TUI screen are the next one.

The goal-runner skeleton is already on `main`, so this is narrower than the closed PR: it adds the verification machinery to the loop that exists, and touches nothing outside the goals domain.

## What it does

A goal run ended the moment the agent wrote `GOAL_DONE`.
That made the worker the sole judge of its own work — the one check a long-horizon loop most needs, and the one it did not have.

Setting `loop_engineering` on a goal adds two judges that are not the worker:

- **A verifier agent** (`Goal::verify_agent_id`) reads each iteration's output and returns `VERDICT: PASS|FAIL|NEEDS_REWORK`.
  A rejection goes back to the generator carrying the verifier's stated reason, up to `verify_max_retries` rework rounds, and the reworked reply replaces the rejected one.
  Until the verifier passes the work, `GOAL_DONE` does not end the run.
- **An evaluator model** (`Goal::evaluator_model`) makes one cheap yes/no read of the goal against the latest output, and can conclude the goal is met even when the agent never claimed it.

The agent can also record a reusable lesson with `GOAL_LEARNED: <one sentence>`.
Captured lessons are replayed into later iterations' prompts, persisted under `goal_learnings_<id>`, and written into a `goal-learned-*` skill through `evolution::create_skill` — the same prompt-injection scan every other skill-creation path uses, which matters here because the body is model-authored text.

All three are inert unless the goal opts in.
With `loop_engineering` off the prompt is byte-identical to the one it was and the run still makes exactly one LLM call per iteration.
`plain_goal_prompt_is_unchanged_by_the_new_sections` and `a_plain_run_makes_no_extra_llm_calls` assert both halves of that, because an evaluator that fired on every goal would double every existing operator's bill silently.

A run now also gives up after five consecutive non-rate-limit tick failures (`MAX_ERROR_STREAK`) instead of spending its whole iteration budget rediscovering a deleted agent or a revoked key.
That is separate from the existing rate-limit breaker so a transient rate-limit does not also count toward it.

## Sub-agents are delegated, not provisioned

The prompt directs the agent to its own `agent_spawn` / `agent_send` tools, which run under the capabilities its operator granted it.
[...]

PR #7973, opened 2026-08-29
feat(goals): add pause/resume and a configurable loop cadence

DaBlitzStein:
## Summary

- Add `GoalRunPhase::Paused` (between `Running` and `Finished`) and a `tick_interval_secs: Option<u64>` field on `Goal` (default 2s, clamped 1-86400 seconds, matching cron's own `MAX_EVERY_SECS` ceiling).
- `GoalRunner::pause`/`GoalRunner::start` (kernel): pause is cooperative -- the loop finishes its current turn, checkpoints iteration count and progress into the shared KV (the `goal_runs` table's `phase` column has a `CHECK` constraint that does not admit `paused`), and exits. `start()` auto-detects and resumes from that checkpoint when one exists.
- `stop()` (cancel) now also discards any pause checkpoint, so a cancelled goal's next start is a genuine fresh run.
- `GoalRunner::state()` falls back to a persisted pause checkpoint when the in-memory registry has no live entry, so a paused run stays observable via `GET /api/goals/{id}/run` after its loop task self-cleans.
- The run loop now reads `tick_interval_secs` from the freshly-loaded goal each iteration instead of a hard-wired 2-second constant.
- Kernel: `LibreFangKernel::goal_run_pause` / `goal_run_resume` (the latter is the same path as `goal_run_start` -- resume is start-with-a-checkpoint). `KernelApi` trait gets matching `pause_goal_run` / `resume_goal_run` methods.
- Fix: goal-run ticks now carry the goal id as `SenderContext.chat_id`, so `send_message_full`'s session derivation scopes on `(agent, "autonomous", goal_id)` instead of just `(agent, "autonomous")`. Previously every loop-mode goal of one agent collapsed onto a single session, so two concurrent runs interleaved their prompts into one conversation history.
- API: `POST /api/goals/{id}/pause` and `POST /api/goals/{id}/resume` (409 when there is no paused run to resume -- without that guard, resume on a goal with no checkpoint would silently restart it from iteration 0). `start_goal_run` and the new `resume_goal_run` share a `start_or_resume` body.
- API: `POST /api/goals` and `PUT /api/goals/{id}` validate and persist `tick_interval_secs`; a blank string or `null` clears the override back to the default, matching the existing `parent_id` / `agent_id` clear-signal convention (#6562).

## Out of scope

Two fixes originally scoped for this change -- preserving `GOAL_LEARNED:` marker case and letting an evaluator complete a goal even without an explicit `GOAL_DONE` marker -- do not apply to this codebase's current goal runner: there is no learnings-capture mechanism, no evaluator consultation, and no `GOAL_LEARNED` marker parsing here at all (`parse_tick` only recognizes `GOAL_PROGRESS:` / `GOAL_DONE` / `GOAL_COMPLETE` / `GOAL_BLOCKED`). Confirmed via `grep -r 'GOAL_LEARNED\|evaluate_goal\|verify_agent_id' crates/` returning no hits anywhere in the workspace. Nothing to fix.

## Test plan

- [x] `cargo check -p librefang-types -p librefang-kernel -p librefang-api --lib` -- clean
- [x] `cargo clippy -p librefang-types -p librefang-kernel -p librefang-api --all-targets -- -D warnings` -- clean
- [x] `cargo test -p librefang-types --lib goal` -- 12 passed
- [x] `cargo test -p librefang-kernel --lib goal_runner` -- 21 passed (includes new pause/resume/cadence unit tests)
- [x] `cargo test -p librefang-kernel --lib goal_session_scope` -- 3 passed
- [x] `cargo test -p librefang-api --test goals_routes_integration` -- 55 passed (includes new pause/resume/cadence integration tests)
- [x] `cargo test -p librefang-api --test dead_route_audit_test --test openapi_path_coverage_test` -- clean (new routes are not `#[utoipa::path]`-annotated, matching the existing `/start` / `/stop` convention, so neither test needs updating)
- [x] `cargo test -p librefang-api --test openapi_spec_test` -- regenerated `openapi.json`, no diff (expected, per above)
- [x] `python3 scripts/codegen-sdks.py` -- regenerated SDKs, no diff
- [x] `cargo fmt --check` -- clean

Refs #5744


houko (commented):
Good change overall — separating "pause keeps the checkpoint" from "stop discards it" is the right distinction, `/resume` being `/start` plus a precondition is the right factoring, and rejecting an out-of-range `tick_interval_secs` at the API while the runner still clamps defensively is the correct division (the operator hears about their number; a bad stored value can never wedge a loop).

One thing to fix: `/resume` answers `409` for a goal that does not exist.

In `start_or_resume`, the `require_paused` guard runs before the goal is looked up:

```rust
if require_paused {
    let paused = state.kernel.goal_run_state(goal_id)
        .is_some_and(|run| run.phase == GoalRunPhase::Paused);
    if !paused {
        return (StatusCode::CONFLICT, Json(json!({
            "error": "This goal has no paused run to resume. Use POST /api/goals/{id}/start to begin a new run."
        })));
    }
}

// ... only after this does the handler read the store and 404 on a missing goal
```

`goal_run_state` returns `None` for a goal that was never created, which is indistinguishable here from a goal that exists with no paused run.
So `POST /api/goals/<typo'd-uuid>/resume` returns `409` with a message telling the operator to `POST /api/goals/{id}/start` — advice that will fail too, because there is no such goal.
`404` is the honest answer, and the well-formed-uuid case is exactly when a typo produces this.

Moving the guard below the store read and the goal lookup fixes it: by then a missing goal has already produced its `404`, and `409` means only what it says.
Worth an integration test alongside `goal_run_pause_signals_a_live_run` — resume against a random uuid, assert `404`.

Smaller note, no change needed: `validate_tick_interval` runs twice on the update path — once before the transaction for the range check, then the value is re-read from `req` inside it.
That is the right shape (validate outside, apply inside), but the second read repeats the `is_clear_signal` branch, so if the clear semantics ever change they have to change in both places.
A comment at the in-transaction site pointing at the validator would be enough.

houko (dismissed):
Careful work: rejecting an out-of-range `tick_interval_secs` at the API while the runner clamps defensively is the right division, the constants carry their reasoning (`MIN_… = 1, not 0: a zero-second cadence removes the only gap between consecutive provider calls`), and the integration tests cover the null-clears-the-override case that usually gets missed. Two gaps.

## Blocking

**1. Neither new route has `#[utoipa::path]`, and both drift guards are blind to that by construction.**

`grep -c "utoipa::path"` over this diff returns 0, so `POST /api/goals/{id}/pause` and `POST /api/goals/{id}/resume` never reach `openapi.json` or any generated SDK. CI is green because of how the two guards are scoped — `openapi_path_coverage_test.rs`'s own module doc says it:

> The dead-route audit asserts that every path **in the spec** is wired into the axum router. This test asserts that every handler **annotated with `#[utoipa::path(...)]` in the crate source** is actually referenced from `ApiDoc::paths(...)`.

Spec→router and annotation→spec. A handler with no annotation at all is in neither set, so nothing fails. Add `#[utoipa::path]` to both handlers, list them in `ApiDoc::paths(...)` in `openapi.rs`, regenerate `openapi.json`, and recompute `xtask/baselines/openapi.sha256`.

This matters immediately: #8029 adds `pauseGoalRun` / `resumeGoalRun` to the dashboard client against these exact paths, and the four SDKs are generated from the spec.

**2. No changelog fragment.**

Pause/resume on autonomous goal runs and a configurable loop cadence are both operator-visible, and the cadence one has a cost story worth telling in the release notes (the `MIN_GOAL_TICK_INTERVAL_SECS` rationale is already written — reuse it): `changelog.d/added/7973-....md`.

## Coordination

This PR is what makes `GoalRunPhase::Paused` real. Two open PRs are already written against it:

- **#8067** widens the dashboard's `phase` union to include `"paused"` and adds a `Pause`-icon badge. Without this PR that value cannot occur; with it, it can.
- **#8029** adds the `pauseGoalRun` / `resumeGoalRun` client functions and mutation hooks for the two routes added here.

Both should land after this one, and #8067's badge work would benefit from #1 being fixed first so its TS types can come from the spec rather than being hand-written.

## Non-blocking

- `pause_goal_run` answers `200 {"ok": true, "paused": false}` when there was nothing to pause. That is defensible (the request succeeded; the run was already not running), but a client cannot distinguish it from a successful pause without reading the body. Worth one line in the handler doc saying so, since #8029's mutation only checks the HTTP status.
- The pause checkpoint lives in shared-memory KV and `clear_pause_checkpoint`'s doc calls it "load-bearing: a checkpoint that outlives its pause would silently seed the next run". Worth an integration test for the daemon-restart case — pause, drop and rebuild the runner, and assert `state()` still reports `Paused` from the checkpoint rather than the run having vanished. The fallback path in `state()` is written for exactly that and is currently only exercised in-process.

houko (commented):
Downgrading my earlier review from "request changes" to a comment.

On reflection the two things I raised are additive rather than defects: the code is sound, CI is green, and the design — rejecting an out-of-range `tick_interval_secs` at the API while the runner clamps defensively — is right. Blocking a green PR on a missing annotation was a heavier call than it deserved, so this is your call to make rather than mine to gate.

Both asks stand, and I would still take them before merge rather than after:

**`#[utoipa::path]` on the two new handlers.** `grep -c "utoipa::path"` over the diff is 0, so `POST /api/goals/{id}/pause` and `/resume` never reach `openapi.json` or the four generated SDKs. CI is green because of how the guards are scoped — `openapi_path_coverage_test.rs`'s module doc says it checks *annotated handler → `ApiDoc::paths(...)`*, and `dead_route_audit_test.rs` checks *spec → router*. A handler with no annotation is in neither set, so nothing fails. It matters immediately because #8029 adds `pauseGoalRun` / `resumeGoalRun` against these exact paths.

**A changelog fragment.** `changelog.d/added/7973-....md`. The `MIN_GOAL_TICK_INTERVAL_SECS` rationale already in the source ("a zero-second cadence removes the only gap between consecutive provider calls, turning a run into a tight loop billing tokens as fast as the provider answers") is the sentence the release notes want.

Worth restating the sequencing, since three PRs depend on this one: it is what makes `GoalRunPhase::Paused` real, so #8067 (dashboard phase badges, widens the union to include `"paused"`), #8029 (pause/resume controls) and #7784 (TUI `tui-goals-phase-paus`) all need it first.

houko (changes requested):
The pause/resume design is right and the reasoning in the doc comments is unusually good — the "why not the `goal_runs` table" note on `goal_pause_key`, the ordering constraint on reading the checkpoint before `stop_locked` clears it, and the `goal_tick_sender_context` note on why per-goal scoping does not cost prompt-cache reuse are all load-bearing and all correct. The session-scope fix is a real bug fix and its three tests assert the right properties.

Two defects and one convention item before this can go in.

## 1. `POST /api/goals/{id}/resume` replaces the run's iteration cap with the default, and can terminate the resumed run immediately

`resume_goal_run` dispatches with no body:

`crates/librefang-api/src/routes/goals.rs:206-210`

```rust
pub async fn resume_goal_run(
    State(state): State<Arc<AppState>>,
    Path(id): Path<String>,
) -> (StatusCode, Json<serde_json::Value>) {
    start_or_resume(state, id, None, true).await
}
```

`start_or_resume` reads the cap only from that body, so `max_iterations` is `None` on every resume:

```rust
let max_iterations = match body.as_ref().and_then(|body| body.0.get("max_iterations")) {
    None => None,
    ...
```

`goal_run_start` then substitutes the compiled default (`goal_lifecycle.rs:33`):

```rust
let max = max_iterations.unwrap_or(DEFAULT_GOAL_MAX_ITERATIONS).max(1);
```

and `GoalRunner::start` seeds `iteration` from the checkpoint while taking `max_iterations` from the argument (`goal_runner.rs:635-636`):

```rust
iteration: resume.as_ref().map(|r| r.iteration).unwrap_or(0),
max_iterations,
```

`ResumePoint.max_iterations` is parsed out of the checkpoint at `goal_runner.rs:331` and used by `state()` at `:498`, but never by `start()`. That asymmetry is the bug: `GET /run` reports the paused run's real cap, and `/resume` then runs it with a different one.

Failure scenario:

1. `POST /api/goals/{id}/start` with `{"max_iterations": 100}`.
2. The run reaches iteration 30. `POST /api/goals/{id}/pause` checkpoints `{iteration: 30, max_iterations: 100}`.
3. `POST /api/goals/{id}/resume`. The cap becomes 25.
4. `run_loop`'s top-of-loop guard (`goal_runner.rs:917`) sees `30 >= 25` and breaks with `MaxIterationsReached` before the first turn.
5. `start` → `stop_locked` already cleared the checkpoint, and the non-`Paused` exit calls `clear_pause_checkpoint` again, so the run is settled and the 30 iterations of progress are unrecoverable. Resume destroyed the thing it was asked to continue.

It also misfires in the other direction: start with `{"max_iterations": 5}`, pause at iteration 3, resume, and the run now gets a cap of 25 the operator never set.

The checkpoint value is already loaded — `start` should prefer `resume.max_iterations` over the argument when resuming (or `resume_goal_run` should read it and pass it through). Whichever layer you pick, please also decide explicitly what an explicit `max_iterations` on a resume body should mean, and say so in the doc comment.

Neither existing test catches this because both resume paths hard-code the cap: `pause_checkpoints_and_start_resumes_from_it` (`goal_runner.rs:1998`) calls `runner.start(goal_id, agent_id, 100, ...)` with the same 100 it started with, and the four HTTP tests cover pause-signalling, the idle-pause `false`, the no-checkpoint 409 and malformed ids — none drives a successful resume through the route. A test that starts with a non-default cap, pauses, resumes via `POST /resume`, and asserts the resumed `max_iterations` is what the operator set is the discriminator.

## 2. A paused run's `started_at` / `updated_at` are minted fresh on every poll; the persisted `paused_at` is never read

`persist_pause_checkpoint` stores it (`goal_runner.rs:310`):

```rust
"paused_at": Utc::now().to_rfc3339(),
```

`ResumePoint` (`:293-298`) has no `paused_at` field and `load_pause_checkpoint` (`:319`) does not parse it, so the value is write-only. `state()`'s checkpoint fallback then stamps both timestamps with the clock (`:492-503`):

```rust
let now = Utc::now();
Some(GoalRunState {
    ...
    started_at: now,
    updated_at: now,
})
```

So two consecutive `GET /api/goals/{id}/run` calls on a goal that is not moving return two different `started_at` values, and any client computing "paused for how long" gets approximately zero, always. For a feature whose point is that a suspended long-horizon run stays observable, the timestamps are the part an operator reads.

The fix is small since the data is already persisted: carry `paused_at` on `ResumePoint`, parse it, and use it for `updated_at`. Please also checkpoint the run's real `started_at` and restore it, in both `state()` and `start()` — `start()` currently stamps `started_at: now` on resume too (`:639`), so a resumed long-horizon run loses the timestamp of when it actually began. If you decide a resume legitimately begins a new segment and should re-stamp, that is defensible, but it needs to be the documented choice rather than a side effect.

## 3. No changelog fragment

The diff touches seven files and none is under `changelog.d/`. This adds two HTTP routes, a new `GoalRunPhase` variant, and a new persisted `Goal` field — user-visible surface that belongs in the release notes. Per `changelog.d/README.md` this wants a fragment in `changelog.d/added/` named for the PR, ending `(#7973) (@DaBlitzStein)`. Worth two entries if you want the cadence control described separately from pause/resume, since they are independently useful.

## 4. Nit: `goal_tick_sender_context`'s `display_name` parameter

`goal_lifecycle.rs:136-146` takes `display_name: &str`, and the sole production caller (`:53`) passes `SYSTEM_CHANNEL_AUTONOMOUS` — the same constant the function already hardcodes into `channel`. The three tests pass it too. It is a degree of freedom nothing varies and nothing could sensibly vary, so it reads as configurable when it is not; drop the parameter and use the constant directly in the body.

## Verified separately

The "Out of scope" note checks out: `GOAL_LEARNED`, `evaluate_goal` and `verify_agent_id` have no hits in the workspace, and `parse_tick` recognises only the four markers you list. Nothing to fix there.

DaBlitzStein:
Merge-order warning for this PR and #7785, found by merging every open branch into one integration tree — the two are compatible, but one specific conflict has a resolution that compiles fine and silently drops a feature.

Both branches touch `run_loop` and `start_goal_run`. That part is fine: git merges the signature additively, and `kernel_api.rs` comes out clean with `start_goal_run` extended and `resume_goal_run` intact. The features genuinely coexist.

The problem is the call site in `routes/goals.rs`. #7785 replaces the call with its own version carrying the loop-engineering parameters, and in doing so removes the `require_paused` branch this PR adds. Whoever resolves that conflict by taking one side wholesale **loses either resume or loop-engineering, and the build stays green** — there is no compile error and no failing test to catch it, because each side is internally consistent.

The resolution that keeps both:

```rust
let started = if require_paused {
    state.kernel.resume_goal_run(goal_id, agent_id, max_iterations)
} else {
    state.kernel.start_goal_run(goal_id, agent_id, max_iterations,
        loop_engineering, verify_agent_id, verify_max_retries, evaluator_model)
};
```

Two more things the second one to land will hit in the same area: the test modules in `goal_runner.rs` end up interleaved inside each other's bodies and have to be rebuilt from both branches rather than patched, and this PR's cadence test calls `run_loop` with the old signature, so it needs updating to the widened one.

Flagging it on both PRs. Nothing to change here — the note is for merge time.

DaBlitzStein:
Pushed `be55d2c5b`: merged current `main`, which this branch was 67 commits behind.

**No hand-written merge resolution to review** — `git merge-tree --write-tree <previous tip> origin/main` against the committed tree reports zero lines of difference, so git's automerge and the committed result are identical.

Verification on the pushed tip: `cargo check` of the five touched crates — exit 0, 27 crates compiled. Scoped goal tests — 110 of 112 passing, and both of the remaining two green on a re-run at lower concurrency. `goal_create_rejects_an_out_of_range_tick_interval` failed at 40 s under load 18.8 and passes in 2.4 s at load 4; both failures carried `Memory init failed: … database is locked` at test-kernel boot, which is the connection pool timing out when the process cannot get scheduled — not a shared file, since `mock_kernel` puts its SQLite under its own tempdir.

Worth restating the merge-order note from earlier, since this refresh does not change it: this PR and #7785 both touch the `run_loop` / `start_goal_run` call site in `routes/goals.rs`, and resolving that conflict by taking one side wholesale silently drops either resume or loop-engineering while still compiling. The merge here came out clean only because #7785 is not in this tree — the collision appears when the two are combined.

DaBlitzStein:
Completing the verification for `5ddf22d55`, which was pushed with only the compile check.

- `cargo check -p librefang-api -p librefang-kernel -p librefang-types --all-targets` — exit 0.
- `cargo nextest run -p librefang-api -E 'binary(goals_routes_integration)'` — **50 tests run, 50 passed, 0 skipped**, in 29.98 s.

No hand-written merge resolution: `git diff` between git's own automerge tree and this branch's HEAD is empty.

The 29.98 s is worth recording next to a number from earlier today, because it is the cleanest illustration of a false red this suite produces. The **same binary** took 792 s in a sibling worktree and reported 8 failures, all `database is locked`, all clustered at 31-34 s, with **zero assertion failures**. Re-run at one thread they passed in about 1.9 s each. Here, on an unloaded machine, the whole binary finishes in half a minute with everything green.

That is the honest way to read `Memory init failed: … database is locked` from `mock_kernel`: it is the connection pool's deadline expiring because the process cannot get scheduled, not a lock anyone holds — the test kernel builds its database in a fresh `tempfile::tempdir()` that no other test can even see.

Also worth stating: the conflict anticipated in `routes/goals.rs` between this branch and the loop-engineering one **does not exist against `main`**. It only appears when the two are merged with each other, where `resume_goal_run` has to forward the parameters `start_goal_run` gained rather than being filled with defaults — filling them compiles and silently drops the verifier on resume.

DaBlitzStein:
Addressed at `3e5202001`, all four points.

**1 — the resume keeps the paused run's cap.** `GoalRunner::start` resolves `max_iterations` itself now, in the order: explicit argument, then the checkpoint's own cap, then the compiled default. It is resolved there because that is the only layer that knows whether this is a resume — by the time `goal_run_start` sees a `None` it cannot tell it apart from an operator asking for the default.

On your explicit question about what an explicit cap on a resume body should mean: **it is honoured**, and the doc comment says so. A resume naming `max_iterations` is a deliberate re-budgeting of the remaining run, and refusing it would leave no way to extend a run about to hit its cap.

The discriminator you asked for is `goal_run_resume_keeps_the_paused_runs_iteration_cap`, driven through `POST /resume`. Dropping the checkpoint fallback fails it with your scenario's numbers:

```
assertion `left == right` failed: an absent cap must restore the paused run's,
not the compiled default
  left: 25
 right: 100
```

`an_explicit_cap_outranks_the_checkpoints_on_resume` and `goal_run_resume_rejects_an_invalid_iteration_cap` cover the other two directions.

**2 — `paused_at` is read now, and `started_at` survives.** `ResumePoint` carries both, `state()` uses `paused_at` for `updated_at` and the checkpoint's `started_at` for `started_at`, so two consecutive `GET /run` calls on a motionless goal return the same timestamps and "paused for how long" is answerable. `start()` restores the original `started_at` rather than re-stamping: a resume continues the run it checkpointed rather than opening a new segment, and that is the documented choice rather than a side effect.

**3 — two changelog fragments**, pause/resume and the cadence control separately, as you suggested.

**4 — `display_name` is gone** from `goal_tick_sender_context`; the constant is used directly.

27 kernel tests green, including the four new ones and the five HTTP cases.

Ready for another look.

DaBlitzStein:
Cross-reference found while building the integration branch (`main` + all 41 open PRs). **This PR and #7785 cannot both merge as they stand**, and neither is wrong on its own — which is why both are green.

Both change the same two functions in `crates/librefang-kernel/src/goal_runner.rs`, in ways that do not compose:

| | `GoalRunner::start` | `run_loop` |
|---|---|---|
| `main` (`95c258cc4`) | `pub fn start<F, Fut>(` | `async fn run_loop<F, Fut>(` |
| #7785 (loop engineering) | `pub fn start<F, Fut, L, E, Efut>(` + `on_learnings_captured`, `evaluate_goal`, `loop_engineering`, `verify_agent_id`, `verify_max_retries`, `evaluator_model` | `<F, Fut, L, E, Efut>` with the same six |
| this PR (pause/resume) | keeps `<F, Fut>`, resolves `max_iterations` internally from the pause checkpoint, and takes `started_at` from it | adds the `pause: Arc<AtomicBool>` flag |

Whichever lands second does not compile: the first's call sites pass an argument list the second's signature does not have, and every `run_loop` call site in the test module has the same problem from the other direction.

**The reconciliation is small and I have already validated it**, on an integration branch that carries both:

- Keep #7785's generics and `#[allow(clippy::too_many_arguments)]`, and #7785's parameter list.
- Keep this PR's body wholesale — the checkpoint read before `stop_locked`, the `max_iterations` precedence (explicit → checkpoint → compiled default), and `started_at: resume.as_ref().and_then(|r| r.started_at).unwrap_or(now)`.
- `run_loop` takes both `stop` and `pause`, and #7785's six loop-engineering parameters.
- `GoalRunState`'s reconstruction from a checkpoint in `state()` needs `verify_agent_id` / `verify_max_retries` / `evaluator_model`, which the checkpoint does not store. Reading them back from the goal document — the same place `start`'s caller reads them from — keeps a paused run reporting the verifier its resume will actually use, instead of a blank that reads as "no gate on this run".
- `KernelApi::resume_goal_run` and `goal_run_resume` take the four loop-engineering arguments too, so `/resume` does not silently drop a verifier the operator configured.

With that in place, both feature sets pass together: 65 kernel goal tests green, including `verifier_rejection_sends_the_work_back_to_the_generator` (#7785) alongside `pause_checkpoints_and_start_resumes_from_it`, `an_explicit_cap_outranks_the_checkpoints_on_resume` and `run_loop_waits_the_goals_configured_tick_interval` (this PR).

**What I am not doing unilaterally:** rebasing one of these onto the other. That decides the merge order for two of your PRs and turns whichever loses into a stacked branch that cannot merge alone. Tell me which one should go first and I will push the rebase with the reconciliation above the same day — it is one function pair plus the test call sites, and I have the resolved version ready.

DaBlitzStein:
From a surface-parity audit of all 41 open PRs against `CLAUDE.md`'s rule that the TUI and the WebUI are as mandatory as the API, never a follow-up.

**`tick_interval_secs` reaches no surface at all.** This branch adds it to the goal document (`librefang-types/src/goal.rs`) with `MIN` / `MAX` / `DEFAULT` constants and validates it at the API boundary (`routes/goals.rs`), and the runner honours it — `run_loop_waits_the_goals_configured_tick_interval` pins that. But grepping all 41 PR heads for `tick_interval_secs` under `crates/librefang-api/dashboard` and `crates/librefang-cli` returns **zero files**: not the dashboard, not the TUI, not the CLI, in any open branch.

It is also not a `KernelConfig` field, so it never appears in `kernel_config_schema.golden.json` and the schema-driven editors (#8184's TUI config screen, the dashboard's ConfigPage) cannot pick it up for free the way they do for `tool_exec.default_timeout_secs` or `skills.promotion.*`. The only way to set it is to hand-edit the goal's JSON.

A validated knob with bounds nobody can reach is the shape this rule exists to prevent. The cheapest honest fix is a field on the goal create/edit form in the dashboard plus the matching TUI input, but I would rather ask than guess: do you want that in this PR, or split out with this one noting the gap?

Not blocking anything I can see, and the backend half is sound.

DaBlitzStein:
Addendum to the collision note above: **it is a three-way, not a two-way.** #8029 carries its own copy of this backend, and it is a strict subset of this one — `start` and `run_loop` unchanged from `main`, and `resume_goal_run` at `routes/goals.rs:206-211` taking no body and passing `None` where this PR's `:209-214` forwards `body: Option<Json<Value>>`.

The routes themselves are safe: both register byte-identical `.route("/goals/{id}/pause")` and `.route("/goals/{id}/resume")` lines that collapse into one registration. What does not survive a careless resolution is the re-budgeting your approval singled out — resolve the handler bodies toward #8029 and a resume silently falls back to the compiled default cap, with `an_explicit_cap_outranks_the_checkpoints_on_resume` on the other side of the file that was replaced.

So the order decision for #7785 and this PR should cover #8029 too. Noted there as well.

DaBlitzStein:
Stacked on #7785 and pushed at `19eef8698`, which resolves the `GoalRunner` collision from this side. The order picked is **#7785 → this PR → #8029**.

**The reconciled signature.** `GoalRunner::start` and `run_loop` are now `<F, Fut, L, E, Efut>` — #7785's generics and its six loop-engineering parameters — carrying this PR's body wholesale: the checkpoint read before `stop_locked`, the `max_iterations` precedence (explicit → checkpoint → compiled default), and `started_at: resume.as_ref().and_then(|r| r.started_at).unwrap_or(now)`. `run_loop` takes both `stop` and `pause`.

Three things needed a decision rather than a merge, and each is small:

- **`GoalRunner::state()` reconstructing a paused run from its checkpoint** now needs `verify_agent_id` / `verify_max_retries` / `evaluator_model`, which the checkpoint does not store. They come back from the goal document — the same place `start`'s caller reads them from — so a paused run reports the verifier its resume will actually use rather than a blank that reads as "no gate on this run". Extending the stored checkpoint would have been a schema change this PR has no reason to make.
- **`goal_run_resume` and `KernelApi::resume_goal_run` take the four loop-engineering arguments too.** A resume that dropped them would silently lose the gate the operator configured on the goal, which is the same class of silent loss as the cap this PR exists to preserve.
- **`KernelApi::start_goal_run` keeps #7785's wide signature**, since it is the layer #7785 widened.

**Verification, on the stacked tree:**

- `cargo test -p librefang-kernel --lib goal` → **65 passed, 0 failed**, with both feature sets in one binary: `verifier_rejection_sends_the_work_back_to_the_generator`, `verifier_pass_lets_goal_done_finish_the_run` and `the_evaluator_can_finish_a_goal_the_agent_never_claimed` from #7785, alongside `pause_checkpoints_and_start_resumes_from_it`, `an_explicit_cap_outranks_the_checkpoints_on_resume`, `resume_restores_the_checkpoints_cap_and_start_time` and `run_loop_waits_the_goals_configured_tick_interval` from here.
- `cargo test -p librefang-api --test goals_routes_integration` → **65 passed**.
- `cargo clippy -p librefang-kernel -p librefang-api --all-targets -- -D warnings` → clean.

**What is deliberately not here:** I lifted the reconciliation from an integration branch that also carries #7781, and `kernel_api.rs` came across with #7781's `set_agent_model_routing` attached. That is stripped — `grep -c "set_agent_model_routing\|model_router\|AgentRouterOverride"` on `kernel_api.rs` is 0. Nothing from any other open PR is in this diff.

**#8029 is the remaining leg.** It carries its own strict subset of this backend, and once this lands it should keep only its dashboard layer — which is what its title says it is. Noted there.

DaBlitzStein:
**The stacking commit I pushed earlier did not build. Fixed at `bb3127819`; the branch is also refreshed onto current `origin/main`.**

`19eef8698` widened `goal_run_start` on the kernel and `resume_goal_run` on the `KernelApi` trait, and left `start_goal_run` on the trait at its three-argument shape. That leaves two call sites disagreeing in opposite directions:

- `kernel_api.rs` — the trait impl calls the kernel's seven-argument `goal_run_start` with three
- `goal_runner.rs:188` — calls the trait's three-argument `start_goal_run` with seven

→ two `E0061`s, `could not compile librefang-kernel`.

**How it got past me, since that is the part worth recording.** The edit existed and was verified — `cargo check -p librefang-kernel -p librefang-api --all-targets` clean, 65 goal tests green, `clippy -D warnings` clean. All of it ran against a working tree that contained the fix, and the fix never got staged into the commit. **What I tested and what I pushed were not the same tree**, and every signal I had was about the tree I tested.

The corrected commit is verified the same way, but this time against what is actually committed: `cargo check -p librefang-kernel -p librefang-api --all-targets` clean, `cargo test -p librefang-kernel --lib goal` 65 passed.

I then audited the other nine branches I pushed today for the same shape — working tree versus committed HEAD versus pushed remote, per branch. **All nine identical on all three.** This was the only one.

Nothing about the reconciliation itself changed: `start`/`run_loop` keep `<F, Fut, L, E, Efut>` with #7785's six loop-engineering parameters and this PR's resume body, and the three decisions I described above still hold. The only difference is that the trait now says what the kernel already said.

Separately, and thanks to a review pass: the comment on `GoalRunner::state()` claimed all three loop-engineering fields come back from the goal document. Two do; `verify_max_retries` is not a `Goal` field at all — it arrives through `/start`'s optional body and neither the document nor `ResumePoint` persists it, so the reconstruction uses the compiled default. That is still the right value to report, because a resume without an explicit number runs under the same default, but the comment sent the reader hunting for a field that does not exist. Corrected in the integration branch and it will come across on the next push here if you want it in this PR — say the word and I will cherry-pick it.

crates/librefang-kernel/src/goal_runner.rs
```diff
@@ -690,22 +1154,46 @@ impl GoalRunner {
...
+    let (verify_agent_id, verify_max_retries, has_evaluator) = {
+        let s = state.lock().await;
+        (
+            s.verify_agent_id,
+            s.verify_max_retries.max(1),
+            s.evaluator_model.is_some(),
+        )
+    };
+
     let mut iteration: u32 = 0;
```
houko: `run_loop`'s local `iteration` still starts at 0 on a resume, so nothing the checkpoint restored survives the first tick.

`start()` seeds `initial.iteration` from the checkpoint (line 891) and `resume_restores_the_checkpoints_cap_and_start_time` asserts it — but the counter that actually drives `if iteration >= max_iterations` (line 1229) and `s.iteration = iteration + 1` (line 1362) is this local, and it is unconditionally 0.

With the scenario the doc comments describe — a run started at `max_iterations = 100`, paused at iteration 30 — a resume:

- runs 100 more turns rather than the remaining 70 (130 total), because the cap check compares against a counter that restarted;
- overwrites `s.iteration` back to `1` on the first completed tick, so `GET /api/goals/{id}/run` reports the run going 30 to 1 and the checkpointed count is lost;
- prompts the agent with `Iteration 1 of 100`.

Both pause/resume tests pass over this because the resumed `send_message` is `std::future::pending()`, so no tick ever lands and `s.iteration` is never rewritten; `pause_checkpoints_and_start_resumes_from_it` reads the state immediately after `start()` for the same reason.

The fix is to thread the resume iteration into `run_loop` (or seed it from `state.lock().await.iteration` alongside the verifier config read at line 1182), plus a regression test whose resumed `send_message` actually resolves.

crates/librefang-kernel/src/goal_runner.rs
```diff
@@ -736,23 +1230,130 @@ async fn run_loop<F, Fut>(
...
+                // decorative: the agent could close its own goal by asserting
+                // completion in an output the verifier had just rejected.
+                let done = verified && (parsed.done || evaluator_done);
+                let new_status = if done {
+                    Some(GoalStatus::Completed)
                 } else {
-                    parsed.progress
+                    Some(GoalStatus::InProgress)
                 };
+                let new_progress = if done { Some(100) } else { parsed.progress };
```
houko: The verifier gate is bypassable with `GOAL_PROGRESS: 100`.

When the verifier rejects an iteration `done` is correctly `false`, but `new_progress` still falls through to `parsed.progress`, and `patch_goal` writes it to the goal document. The next top-of-loop check is `if matches!(goal.status, Completed | Cancelled) || goal.progress >= 100 { break GoalRunPhase::Finished }`.

So an agent that ends its reply with `GOAL_PROGRESS: 100` instead of `GOAL_DONE` closes the run on the following iteration whatever the verifier said. The comment just above says the gate exists so the agent cannot "close its own goal by asserting completion in an output the verifier had just rejected" — `GOAL_PROGRESS: 100` is that same assertion through the other marker.

Suggest clamping self-reported progress on an unverified iteration (`parsed.progress.map(|p| p.min(99))` when `rejection.is_some()`), or gating the `goal.progress >= 100` exit on the iteration having been verified.

crates/librefang-kernel/src/goal_runner.rs
```diff
@@ -803,13 +1416,25 @@ async fn run_loop<F, Fut>(
...
 
         iteration += 1;
 
+        let tick_secs = goal
+            .tick_interval_secs
+            .unwrap_or(DEFAULT_GOAL_TICK_INTERVAL_SECS)
+            .clamp(MIN_GOAL_TICK_INTERVAL_SECS, MAX_GOAL_TICK_INTERVAL_SECS);
         tokio::select! {
-            _ = tokio::time::sleep(TICK_INTERVAL) => {}
+            _ = tokio::time::sleep(Duration::from_secs(tick_secs)) => {}
```
houko: A pause request is not observed during the inter-tick sleep, and this PR raises that sleep's ceiling to 24 hours.

`pause()` deliberately does not abort the task ("the loop needs to run to completion of its current turn"), and the only pause check is at the top of the loop (line 1206). This `select!` watches `shutdown_rx` but not `pause` / `stop`, so with `tick_interval_secs = 3600` a pause lands up to an hour later, and at the new `MAX_GOAL_TICK_INTERVAL_SECS = 86400` up to a day later — while the route doc tells the operator to "poll `GET /api/goals/{id}/run` for the phase to reach `paused`".

Adding a `tokio::sync::Notify` (or a bounded poll) to this `select!` fixes it without weakening the checkpoint guarantee: the agent turn is already finished by the time control reaches here, so waking early only skips dead time.

crates/librefang-kernel/src/goal_runner.rs
```diff
@@ -819,11 +1444,34 @@ async fn run_loop<F, Fut>(
...
 
-    {
+    let snapshot = {
         let mut s = state.lock().await;
         s.phase = final_phase;
         s.updated_at = Utc::now();
+        s.clone()
+    };
+
+    if final_phase == GoalRunPhase::Paused {
```
houko, line 1572: Pausing a `loop_engineering` run destroys every lesson it captured.

This branch returns before the `if loop_engineering && !learnings.is_empty()` block at line 1487, so nothing is written to `goal_learnings_<goal_id>` and `on_learnings_captured` never fires. `persist_pause_checkpoint` has no field for learnings either, and the resumed `run_loop` starts from `let mut learnings: Vec<String> = Vec::new()` (line 1197).

A run that captured ten `GOAL_LEARNED:` lessons over 30 iterations and is then paused loses all ten: no durable store entry, no pending draft, and they stop being replayed into the prompt after the resume. The changelog fragment says a pause checkpoints "iteration count, progress and start time" — the learnings belong in that list, or at minimum should be persisted here before the early return.

crates/librefang-kernel/src/kernel/goal_lifecycle.rs
```diff
@@ -44,36 +53,167 @@ impl LibreFangKernel {
...
+        // human has read them. They go into the skill workshop's `pending/`
+        // queue (#3328), the same place every other machine-proposed skill
+        // waits, rather than straight into the installed skills directory.
+        //
+        // The workshop's cap / TTL settings are read once here, at run start:
+        // the hook fires on a background task after the run ends, and reaching
+        // back into the registry from there would mean carrying a kernel handle
+        // for two `u32`s.
+        let skills_dir = self.home_dir().join("skills");
+        let workshop = self
```
houko: `workshop.enabled` is read but never checked, so goal runs write pending skill drafts for agents that never opted into the workshop.

The workshop's own gate is `cfg.enabled && cfg.auto_capture && cfg.max_pending > 0` (`skill_workshop/mod.rs:123` and `:148`), and `storage::save_candidate` only enforces `max_pending == 0`. `queue_learnings_as_pending_skill` forwards `workshop.max_pending` / `workshop.max_pending_age_days` and drops `enabled` and `auto_capture`.

`SkillWorkshopConfig::default()` is `enabled: false` — "Default-OFF — opt-in per the original #3328 acceptance criteria" — and it is also what the `unwrap_or_default()` here produces for an agent not currently in the registry. So any `loop_engineering` goal whose agent emits `GOAL_LEARNED:` starts filling `~/.librefang/skills/pending/<agent>/` on a daemon where the workshop was never turned on, and `goal_lessons_wait_in_pending_instead_of_installing_themselves` asserts that behaviour with `SkillWorkshopConfig::default()`.

Suggest an early return in `queue_learnings_as_pending_skill` when `!workshop.enabled || !workshop.auto_capture`. The durable `goal_learnings_<id>` entry still keeps the lessons, so nothing is lost by respecting the opt-in.

crates/librefang-api/src/routes/goals.rs
```diff
@@ -246,9 +308,59 @@ pub async fn start_goal_run(
...
+                .into_json_tuple();
+            }
+        }
+    };
+    let evaluator_model = goal["evaluator_model"]
+        .as_str()
+        .map(str::trim)
+        .filter(|s| !s.is_empty())
+        .map(str::to_string);
+    let verify_max_retries = body
```
houko: `verify_max_retries` is taken from the request body with no validation and a lossy cast, unlike `max_iterations` 35 lines above.

- `{"verify_max_retries": 100000}` is accepted, and each round is a verifier turn plus a generator turn (`goal_runner.rs:1265-1315`), so a single iteration can issue 200k provider calls. `max_iterations` returns 400 rather than allowing a budget like that.
- `{"verify_max_retries": 4294967296}` truncates through `as u32` to `0`, which `.max(1)` in the runner turns into 1 — silently the opposite of what was asked.
- `{"verify_max_retries": "3"}` is silently dropped to `None` instead of 400.

Suggest mirroring the `max_iterations` arm: 400 on a non-integer or out-of-range value, with a defensible ceiling.

crates/librefang-api/src/routes/goals.rs
```diff
@@ -398,6 +572,32 @@ pub async fn create_goal(
...
 
+    let loop_engineering = req["loop_engineering"].as_bool().unwrap_or(false);
+    // Same boundary rule as `agent_id`: reject a bad verifier id here rather
+    // than storing junk that `start_goal_run` has to reject later.
+    let verify_agent_id_str: Option<String> = req
+        .get("verify_agent_id")
+        .and_then(|v| v.as_str())
+        .map(|s| s.trim().to_string());
+    if let Some(ref vid) = verify_agent_id_str {
+        if vid.parse::<uuid::Uuid>().is_err() {
```
houko: `create_goal` 400s on a blank `verify_agent_id`, while every other id field on this handler treats blank as "unset".

`parent_id` and `agent_id` go through `optional_uuid_field`, where `is_clear_signal` maps `""` / `null` to "not set" (#6562), and `update_goal_by_id` uses `is_clear_signal` for `verify_agent_id` too. Only create rejects it: `req.get("verify_agent_id").and_then(as_str).map(trim)` yields `Some("")`, which fails `parse::<Uuid>()` and answers `Invalid verify_agent_id`.

The dashboard is shielded because `handleCreateGoal` strips the blank key, but any caller posting a complete body — the generated SDKs, a script, `curl` — gets a 400 for leaving the verifier unset. Routing it through `optional_uuid_field` fixes the inconsistency and canonicalizes the stored value at the same time: the current path stores the raw trimmed string, so a braced or upper-case UUID round-trips un-normalized while `agent_id` does not.

Same shape one field down: an `evaluator_model` of `""` is stored as `""` on the goal document rather than omitted.

crates/librefang-types/src/goal.rs
```diff
@@ -190,6 +229,9 @@ pub fn parse_goal_args(args: &str) -> Option<(String, bool)> {
 pub enum GoalRunPhase {
     /// The runner loop is active and driving the assigned agent.
     Running,
+    /// An operator paused the run; iteration count and progress are
+    /// checkpointed so a later start resumes rather than restarts.
+    Paused,
```
houko, line 234: Adding this variant breaks `librefang goal ... --watch`, which is not in the diff.

`crates/librefang-cli/src/commands/goal.rs:193`'s `terminal_phase_message` maps only `finished` / `max_iterations_reached` / `rate_limited` / `stopped`, and `classify_poll` treats anything else with `running == false` as `PollVerdict::Unobservable`. A paused run is exactly that shape (`get_goal_run` sets `running = phase == Running`), so pausing a watched goal makes the CLI print `cmd-goal-watch-poll-error` counting up to `MAX_UNOBSERVABLE_POLLS` and then exit non-zero as though the daemon had become unreachable — rather than reporting that the operator paused it.

`crates/librefang-cli/src/tui/screens/goals.rs` already maps `"paused"`, so the TUI is fine; it is only the `--watch` classifier that needs the new phase, and `every_terminal_phase_has_a_summary` is the test that would have caught it since it enumerates the variants by hand.

houko:
**`crates/librefang-api/dashboard/src/api.ts:4582`**

This union is missing `"paused"`, and there is no `goals.run_phase_paused` key in any of the five locale files.

The comment above `goalRunPhaseBadge` says `phase` is "typed from the wire contract rather than `string` so a caller cannot pass a typo'd literal" — but `GoalRunPhase::Paused` is now a value the server sends that this type says cannot exist. `GoalsPage.tsx:150` (`t("goals.run_phase_" + run.phase, { defaultValue: run.phase.replace(/_/g, " ") })`) therefore renders a bare untranslated `paused` with the `default` badge styling, in every language, while `run_phase_max_iterations_reached` and friends are all present in `en.json`.

Related, and worth stating in the PR body if deliberate: no `pauseGoalRun` / `resumeGoalRun` wrapper or mutation hook was added, so the two new endpoints have no dashboard client.

DaBlitzStein:
Fixed at `c4d581e3f` — you were right on both halves.

The union at `api.ts:4582` gains `"paused"`, and `goalRunPhaseBadge` gains a case for it (the warning palette — the run is alive but not progressing, same family as `stopped`), so the badge no longer falls through to the default styling. `goals.run_phase_paused` lands in all five locales, inserted next to its siblings.

On the related note: the missing `pauseGoalRun` / `resumeGoalRun` wrappers and mutation hooks are deliberate, not an oversight — they arrive in #8029, which is stacked on this branch and adds the dashboard controls that would call them. Stating it here as you asked.

Verification on the pushed tip: `tsc --noEmit` clean, and the dashboard suite runs 205 files / 1808 tests with 0 failures (under `LANG=en_US.UTF-8`; the analytics test is locale-sensitive on this machine, unrelated to this change).

crates/librefang-kernel/src/goal_runner.rs
```diff
@@ -366,27 +653,100 @@ impl GoalRunner {
...
+            // configuration, so these come back from the goal document — the same
+            // place `start()`'s caller reads them from. Reconstructing them from
+            // the clock-free source keeps a paused run's reported verifier the one
+            // its resume will actually use, instead of a blank that reads as "no
+            // gate on this run".
+            verify_agent_id: goal
+                .as_ref()
+                .filter(|g| g.loop_engineering)
+                .and_then(|g| g.verify_agent_id),
+            verify_max_retries: match goal.as_ref() {
```
houko, line 760: A pause loses the run's `verify_max_retries`, the same way it used to lose `max_iterations`.

`persist_pause_checkpoint` (line 476) writes agent_id / iteration / max_iterations / last_progress / started_at / paused_at and nothing else, so this fallback has no stored budget to restore and substitutes the compiled default.

An operator who starts a loop-engineering run with `POST /api/goals/{id}/start {"verify_max_retries": 8}` gets 8 recorded on the run state (line 903). After a pause, once the loop task self-cleans its registry slot, `GET /api/goals/{id}/run` reports `verify_max_retries: 3` — a budget the run never had. `POST /api/goals/{id}/resume` with no body then sends `verify_max_retries: None` (`routes/goals.rs:337`) and the runner falls back to 3 as well, so the 8 is gone with no way to recover it after a daemon restart.

This is the loss 3e5202001 just fixed for the iteration cap, one field over. Persisting it in the checkpoint next to `max_iterations` and giving it the same precedence — explicit argument, then the checkpoint's, then the default — closes it.

houko:
Re-reviewed at `c4d581e3f`. Only one commit has landed since the last review pass — `c4d581e3f` ("the run-phase union and the badge must know about `paused`"), which is a good catch on the dashboard side and is not one of the eight findings. **All eight earlier findings are still open at this head.** I have not re-posted them; the original threads still apply line-for-line, so this is just the verification result plus one new item.

| Finding | Status at `c4d581e3f` |
| --- | --- |
| `run_loop`'s `iteration` restarts at 0 on resume (`goal_runner.rs:1191`) | open |
| Verifier gate bypassable with `GOAL_PROGRESS: 100` (`goal_runner.rs:1356`) | open |
| Pause not observed during the inter-tick sleep (`goal_runner.rs:1437`) | open |
| Pausing a `loop_engineering` run discards its learnings (`goal_runner.rs:1454`) | open |
| `workshop.enabled` / `auto_capture` never checked (`goal_lifecycle.rs:329`) | open |
| `verify_max_retries` unvalidated, lossy `as u32` (`routes/goals.rs:337`) | open |
| `create_goal` 400s on a blank `verify_agent_id` (`routes/goals.rs:578`) | open |
| `librefang goal --watch` cannot classify `paused` (`commands/goal.rs:193`) | open |

Two notes on things that moved:

`3e5202001` ("resume a paused run under its own cap and timestamps") is the right fix for the cap and the timestamps, and it makes the iteration finding sharper rather than resolving it. `start`'s new doc comment now reasons "with the iteration count restored from the same checkpoint, a smaller default ends the resumed run at the top of its first loop" — that is only true if `run_loop` sees the restored count, and `run_loop` still opens with `let mut iteration: u32 = 0;`. The cap check at line 1229 compares 0 against the cap, so the precedence rule the commit added is currently guarding against a failure mode the code cannot reach, while the one it was meant to prevent — a resume running its full cap again from zero — is live. Both resume tests pass over it because their `send_message` is `std::future::pending()`, so no tick ever lands and `s.iteration` is never rewritten.

`c4d581e3f` fixed the `paused` phase in the dashboard's `GoalRunState` union and badge, but the CLI's `--watch` classifier is a separate enumeration of the same phases: `terminal_phase_message` in `crates/librefang-cli/src/commands/goal.rs:193` still lists only four phases, and `classify_poll` treats anything else with `running == false` as `Unobservable`. Pausing a watched goal therefore prints `cmd-goal-watch-poll-error` five times and exits 1 as though the daemon were unreachable. The file is outside this PR's diff, which is why the finding is anchored on the `Paused` variant in `librefang-types`.

One new item, posted inline: the pause checkpoint carries no `verify_max_retries`, so `state()` reports the compiled default for a paused run and a bodyless `/resume` re-budgets the verifier — the same loss `3e5202001` just closed for `max_iterations`.

crates/librefang-kernel/src/goal_runner.rs
DaBlitzStein: Fixed. `run_loop` reads its starting `iteration` from the same state snapshot as `verify_agent_id`/`verify_max_retries` (goal_runner.rs:~1241) instead of a hardcoded 0, so a resumed run continues counting from the checkpoint. Flagging for transparency: the first version of the regression test only asserted the final iteration count, which turned out to be identical with or without the fix (5, reached either via 2 ticks from 3 or 5 ticks from 0) — it passed on the unfixed code too. Replaced with a turn-count assertion (exactly 2 ticks, not 5) plus an assertion on the first resumed prompt's text ("Iteration: 4 of 5"), which does discriminate.

crates/librefang-kernel/src/goal_runner.rs
DaBlitzStein: Fixed, and specifically in the form PR #7785 already shipped for the same underlying bug (same second writer: `goal_update`, the tool the system prompt tells the agent to call, patches `goal.progress`/`goal.status` directly and bypasses `parse_tick`). Adopted #7785's reader-side gate verbatim: the top-of-loop completion check now only accepts a bare `progress >= 100` / `status == Completed` as done when `verify_agent_id.is_none()`; `Cancelled` stays unconditioned since it's a legitimate external stop, not something the gate itself produces. The progress clamp from the first pass is kept as defense in depth (scoped to `!verified`, matching #7785's shape exactly instead of an unconditional clamp, which would have wrongly pinned a legitimate no-verifier 100 at 99). Added a counterpart test proving the plain no-verifier path still finishes on bare `GOAL_PROGRESS: 100` without `GOAL_DONE`, so the two PRs' fixes and tests should converge cleanly on merge rather than one reverting the other.

crates/librefang-kernel/src/goal_runner.rs
DaBlitzStein: Fixed. The inter-tick sleep wakes every second (`PAUSE_POLL_INTERVAL`) instead of sleeping the whole configured interval in one shot, re-checking `pause`/`stop` between wakes. Implemented with a fixed deadline + `sleep_until` rather than chained `sleep(1s)` calls, after a self-review caught that chaining would drift by scheduling latency accumulated once per slice (~1 minute over the 24h maximum tick interval) — a fixed deadline can't drift. Also added a test for a case this change touches: pause and shutdown signalled together mid-sleep now both resolve through the pre-existing pause-before-shutdown check at the top of the loop (documented there), so they resolve to `Paused` — the old flat `select!` used to hardcode `Stopped` on a shutdown signal fired during that sleep specifically, bypassing that policy in that one spot.

crates/librefang-kernel/src/goal_runner.rs
DaBlitzStein: Fixed. `ResumePoint` now carries the run's `learnings`, persisted with the checkpoint and read back into the resumed run's own accumulator (`initial_learnings`), instead of resetting to empty on every pause. Verified with `learnings_survive_a_pause_and_resume_cycle`; tightened after a self-review pointed out the original assertion (an exact 2-element list) raced the scheduler — under load a second pre-pause tick can land before `pause()` takes effect, which the fix doesn't affect but the exact-order assertion did not tolerate. It now asserts the pre-pause and post-resume lessons are both present without pinning the exact count.

crates/librefang-kernel/src/kernel/goal_lifecycle.rs
DaBlitzStein: Fixed. `should_queue_learnings` (`enabled && auto_capture`) gates `queue_learnings_as_pending_skill` itself now, not just its caller — moved inside after a self-review noted the caller-side guard left the function's own tests exercising the opt-in at all (its two pre-existing tests both called it directly with the disabled default and asserted a draft got queued, which is exactly what the fix stops). Updated those two tests to `enabled: true` and added a new one asserting a disabled workshop queues nothing.

crates/librefang-api/src/routes/goals.rs
DaBlitzStein: Fixed. `verify_max_retries` now validates identically to `max_iterations`: 400 on 0, non-integer, or out-of-`u32`-range, instead of the old bare `as u32` cast that silently wrapped. Verified with `goal_run_start_rejects_invalid_verify_max_retries` (0, a string, a float, `u32::MAX + 1`). One thing I'm deliberately NOT doing and flagging rather than letting slide as addressed: the original finding named `100000` as a concrete concern (200k provider calls in one iteration), and that value is still accepted — there's no ceiling below `u32::MAX` here, same as `max_iterations` has none. If you want a defensible ceiling, I'd add it to both fields together rather than just this one; happy to do that as a follow-up if that's the call.

crates/librefang-api/src/routes/goals.rs
DaBlitzStein: Fixed. `verify_agent_id` now goes through `optional_uuid_field` (blank/null = not set, matching `parent_id`/`agent_id`'s #6562 handling) instead of 400ing on a blank string. `evaluator_model` now filters a blank string to `None` instead of storing `Some("")`, matching the field's own documented no-evaluator-configured meaning for `None`. Verified with two new integration tests; the existing non-blank-non-UUID rejection test is untouched and still passes.

crates/librefang-types/src/goal.rs
DaBlitzStein: Fixed. `terminal_phase_message` now has a `"paused"` arm, so `--watch` reports a paused run instead of treating it as unclassified state and burning its retry budget before giving up with a generic message. Both enumeration tests extended; locale key added to all four CLI locales, checked against the i18n coverage tests.

crates/librefang-kernel/src/goal_runner.rs
DaBlitzStein: Fixed, same pattern as `max_iterations` from the earlier round on this field's neighbor. `ResumePoint` now carries `verify_max_retries`, persisted with the checkpoint and read back with the same precedence order (explicit argument → checkpoint → compiled default). Verified with a new test: a run started with `verify_max_retries: 8`, paused, then resumed with an empty body, reports 8 instead of falling back to the default.

DaBlitzStein:
Heads-up on a commit I pushed to this branch, so it is not an unexplained one in your diff: **`1a99d69b9`, `fix(kernel): cancel discards a pause checkpoint it cannot read`.**

It fixes a finding you raised on **#8029** (`crates/librefang-kernel/src/goal_runner.rs`, the `stop_locked` checkpoint delete), not one raised here. It landed on this branch rather than there because `goal_runner.rs` is kept byte for byte identical between this PR and #8029, and a divergent third copy of that file broke the integration branch earlier. #8029 has since merged this branch and is identical again — verified with `git diff --quiet fork/feat/goal-pause-resume -- crates/librefang-kernel/src/goal_runner.rs`.

What it changes, at `goal_runner.rs:818` — your suggestion, applied as written:

`load_pause_checkpoint` answers `None` for two different things: "no checkpoint", and "there is a row I could not read". A substrate error is swallowed by its `.ok().flatten()?`, and a row whose `agent_id` is missing or not a UUID exits through the same `?`. `stop_locked` gated the delete on that `None`, so in both cases it skipped the delete and left the row behind — and the next `/start` auto-resumes from a checkpoint, so a transient storage failure at cancel time silently resurrected the run the operator had just cancelled. That is what the comment directly above the delete says cancel exists to prevent.

The delete is now unconditional and the read only decides the return value, which still reports whether a resume point was readable. It costs nothing: `clear_pause_checkpoint` was already a no-op on a missing key.

Two things about the test at `goal_runner.rs:3884` worth stating rather than leaving for you to find:

- The substrate read error is not reachable from a healthy `MemorySubstrate::open_in_memory`, so the test drives the other input that produces the same `None` over a row that exists — the unparseable `agent_id`. Same branch, same defect, but it is not literally the trigger described, and the doc comment says so.
- It asserts the **raw key** via `structured_get`, not `load_pause_checkpoint`. The neighbouring `stop_discards_a_pause_checkpoint_so_the_next_start_is_fresh` can assert the latter because its checkpoint is readable; on this input it is `None` both before and after the delete, so it cannot tell the fix from the bug. There are also two preconditions before `stop()` — unreadable **and** present — so the test cannot pass vacuously.

Verification: reverting only the two production lines gives

```
---- goal_runner::tests::stop_discards_a_checkpoint_it_cannot_read stdout ----
panicked at crates/librefang-kernel/src/goal_runner.rs:3921:9:
cancel must delete the checkpoint row it could not read, or the next start resumes a run the operator cancelled

test result: FAILED. 50 passed; 1 failed
```

Both preconditions passed before that panic. Restored: `test result: ok. 51 passed; 0 failed`. `cargo clippy -p librefang-kernel --all-targets -- -D warnings` clean. Changelog fragment added under this PR's number, `changelog.d/fixed/7973-goal-cancel-discards-unreadable-checkpoint.md`.

Fast-forward push, no force-push, so your loaded review and its inline anchors are untouched. I have not changed anything else on this PR — the rest of it is its own author's.

DaBlitzStein:
Every one of the 9 review threads on this PR now has an inline reply, and the branch is at `1a99d69b9`.

houko:
Structural note before the code review, so it does not arrive as a surprise later: this PR is one of six that carry the goal-loop feature, and none of them is stacked on another. #7785 (verifier/evaluator gate), #7973 (pause/resume + loop cadence), #8224, #8229, #8230 and this one each contain the other's kernel, API and dashboard changes as their own diverged revision — `git merge-base --is-ancestor` is false for every pair except #8029, which is genuinely stacked on #7973.

The 18 shared files are the whole feature (`goal_runner.rs`, `goal_lifecycle.rs`, `kernel_api.rs`, `types/goal.rs`, `routes/goals.rs`, `goals_routes_integration.rs`, the dashboard goals surface), and the revisions differ — `goal_runner.rs` is +2189 on #7785 and +1131 on #8229 — so whichever merges second gets a conflict whose correct resolution is not "keep both" and not "pick a side".

Full write-up and the suggested order are on #7785. I am holding the code review of this PR until its diff is only the surface it is named for, because a rebase will move all of it.

DaBlitzStein:
Refreshed against `origin/main` in `94838a130` — the branch was 23 commits behind.

One conflict, and it needed a real decision rather than a textual union: `GoalsPage.tsx`.

#8067 rewrote `goalRunPhaseBadge` from the `{bg, text, dot}` Tailwind-class shape to `{variant, icon}` so `Badge` owns the padding, border and dot. This branch had added a `"paused"` arm in the old shape. Taking either side alone loses something — upstream's side has no `paused` phase at all, and `GoalRunState["phase"]` includes `"paused"`, so the exhaustive switch would not compile without it.

Resolved by carrying `paused` into the new vocabulary as `{ variant: "warning", icon: Pause }`, matching the warning colour it already had, with `Pause` added to the existing `lucide-react` import.

DaBlitzStein:
Refreshed against `origin/main` and ready for another look.

**All nine findings from the review are fixed, each answered in its own thread** with the reasoning and the regression test that covers it: the restarted iteration counter on resume, the `GOAL_PROGRESS: 100` verifier bypass, the pause latency across the inter-tick sleep, the learnings destroyed by a pause, the skill-workshop opt-in being ignored, `verify_max_retries` validation, the blank `verify_agent_id` on create, `--watch` not recognising the new phase, and `verify_max_retries` lost across a pause.

**One defect the refresh itself introduced, found and fixed in `0a83858b5`.**
Merging `origin/main` brought in #8067's rewrite of the phase badge from `{bg,text,dot}` to `{variant,icon}`, and the `paused` arm was carried over into the new vocabulary as `{ variant: "warning", icon: Pause }`.
That is correct for the feature, but #8067's fallback test used `paused` as its example of a phase the badge does *not* recognise — a premise this PR retires by definition.
The test then read the warning variant and its icon as a failure of the neutral branch, and `Dashboard Build` went red on it.
The unknown-phase case now names a phase no arm matches, so it tests what its title claims, and `paused` gets its own assertions on the same axes so the arm is pinned rather than free to fall back through `default` in a future merge.

Verification on `0a83858b5`:

- `LANG=en_US.UTF-8 pnpm vitest run` — 207 files, 1860 tests, all passing.
- `pnpm typecheck` — clean.
- `CI Gate` was already green on the Rust side for the refreshed head; the only red was the dashboard test above.

One note for whoever reads the earlier CI run: `serviceWorker.test.ts > returns an error Response when both cache and network miss` also failed in that run with a 5000 ms timeout, and it is not a defect.
It passes on its own and passes in the full suite; it timed out under parallel load on a saturated machine.

houko (approved):
Re-reviewed at the current head. Both blocking findings are fixed, and the first more thoroughly than I asked.

**1 — resume no longer destroys the run it was asked to continue.** Two layers: omitting `max_iterations` restores the cap the paused run was already under instead of substituting the default (`3e520200`), and an explicit cap at or below the checkpoint's completed iteration count is now refused with a 400 rather than accepted into an immediate `MaxIterationsReached` exit that clears the checkpoint (`378b8bfa`). The second case is one I described but did not ask you to guard — the request could never have advanced the run, so refusing it is right.

**2 — the paused run's timestamps** come from the checkpoint rather than being minted per poll, and `seed_pause_checkpoint` puts a test in exactly the state `/resume` reads.

**3 — changelog fragments** are present, including one for the resume-cap refusal.

DaBlitzStein:
Rebasado sobre `57ad92341` (origin/main, release v2026.9.14). Nueva punta: `ed8fa353d`.

**Resoluciones no triviales:**

- **`GoalsPage.tsx` — reestructuración en pestañas (#8251, `cd140da8e`) contra los controles de loop engineering.**
  Main traía la página partida en dos paneles, el formulario de creación dentro del panel `goals`, `ref={createTitleRef}` y el botón de crear en la cabecera; la rama traía los selectores de agente y verificador, el checkbox de loop engineering, el input de modelo evaluador y el badge de fila, escritos contra la estructura plana anterior.
  Se conserva la estructura de main y se reinjertan los controles de la rama en su nueva anidación.

- **`GoalsPage.tsx` — `case "paused"` traducido al vocabulario de #8067.**
  HEAD traía `goalRunPhaseBadge` devolviendo `{variant: BadgeVariant, icon}`; la rama traía `case "paused"` escrito contra el `{bg,text,dot}` anterior.
  La propia rama ya había resuelto exactamente esta colisión en su merge `94838a130`, así que se reproduce esa resolución literal — `case "paused": return { variant: "warning", icon: Pause };`, con su comentario de dos líneas y `Pause` añadido al import de `lucide-react` — en vez de inventar una nueva.

- **`routes/goals.rs` — la llamada de arranque movida detrás de la lectura de configuración.**
  El commit `9cdaea90e` mueve `start_goal_run` para que se ejecute después de resolver `loop_engineering` / verificador / evaluador desde el documento del objetivo, porque ahora los necesita como argumentos.
  Ese commit fue escrito cuando solo existía `start_goal_run`; aquí ya existe la bifurcación `if require_paused { resume_goal_run(...) } else { start_goal_run(...) }` de pausa/reanudación, así que se conserva la bifurcación y se le pasan los siete argumentos a **las dos** ramas.

- **`kernel_api.rs`, `goal_lifecycle.rs`, `goal_runner.rs`, `goal.rs` — lo que el rebase se comió, y el porqué.**
  Esta rama llevaba su integración en **commits de merge** (`19eef8698 merge: stack on #7785`, `94838a130 Merge origin/main`).
  `git rebase` linealiza y **descarta los merges**, así que las resoluciones que vivían solo ahí desaparecen sin que ningún conflicto lo señale.
  Lo perdido y restaurado en `ed8fa353d`:
  - `goal_run_resume` había perdido sus cuatro argumentos de loop engineering tanto en el método inherente del kernel como en el trait `KernelApi`, mientras `routes/goals.rs` seguía llamándolo con los siete. Eso no compila.
  - El checkpoint de pausa había dejado de reconstruir `verify_agent_id`, `verify_max_retries` y `evaluator_model` desde el documento del objetivo, con lo que una ejecución pausada reportaba verificador en blanco.
  El criterio de la restauración es comprobable: todo fichero que main no ha tocado desde la base de fusión vuelve al contenido que ya tenía la punta de la rama, que para esos ficheros es por construcción el resultado correcto del rebase.
  `kernel_api.rs` sí lo tocó main, así que ahí se parte de la versión de main y se recupera solo la declaración y la implementación ensanchadas de `resume_goal_run`.

**Verificado** (dashboard, con `LANG=en_US.UTF-8`):

- Conjunto de ficheros idéntico antes y después del rebase: 38 ficheros.
- Los cinco `locales/*.json`: JSON válido, 0 duplicadas (con `object_pairs_hook`), 0 claves muertas resucitadas, las 9 claves de la rama presentes en los cinco idiomas.
- Coherencia de aridad entre las tres capas: `goal_lifecycle.rs`, `kernel_api.rs` y `routes/goals.rs` coinciden con la punta original.
- `npx tsc --noEmit`: 0 errores. `npx eslint .`: 0. `npx pnpm build`: OK.
- `npx vitest run`: **Test Files 1 failed | 209 passed (210)**, **Tests 1 failed | 1929 passed (1930)**.
  El único rojo es `PromptsExperimentsModal` / "stops selection when every variant has a traffic bucket" (timeout de 5 s), ajeno a este PR.

**NO verificado**: la mitad Rust (`librefang-kernel`, `librefang-cli`, `librefang-api`, `librefang-types`) no se ha compilado ni testeado; no se ejecutó `cargo` de ningún tipo.
Dado que la restauración descrita arriba es precisamente de código Rust, conviene que la comprobación de la integración empiece por ahí.

**Revisiones sin atender**: quedan 9 hilos de revisión sin resolver, 2 de ellos todavía vigentes.
Este rebase no los aborda.

DaBlitzStein:
### Ahora se apila sobre #7785 (`2adb4cba4`)

Esta rama llevaba su propia copia del trabajo de loop engineering, congelada en el punto al que #7785 había llegado cuando se cortó la rama: `43b4ce22b`, «docs(goals): say what an unresolvable evaluator_model does».
Todo lo que #7785 añadió después faltaba aquí, así que integrar las dos en paralelo resolvía la misma región dos veces y con resultados distintos.

**Sobre qué se apila y por qué**: #7785 va debajo porque es la copia más nueva del trabajo compartido — 977 líneas suyas faltaban en esta rama — y porque no necesita nada de nadie.
Esta rama es la segunda de la pila porque solo necesita a #7785, y porque es la que #8029, #8224 y #8230 necesitan a su vez (les faltaban 65, 80 y 85 líneas suyas respectivamente).

La fusión no usó la base que git elegía por defecto.
Los árboles de `43b4ce22b` (#7785) y `57c368907` (la copia de esta rama) difieren **solo** en lo que esta rama añade por debajo, así que apliqué exactamente `43b4ce22b..0734e56f0` — los siete commits nuevos de #7785 — sobre este árbol.
Eso redujo el conflicto de nueve ficheros a cinco, y esos cinco solo donde los arreglos tardíos de las dos ramas tocan las mismas líneas.

**Qué se conservó de cada lado.** Del lado de #7785, por ser la versión más nueva del trabajo compartido:

- `stop` pasa a `Arc<StopFlag>` — quién levantó la parada, no solo que está levantada — para que un `PUT /api/goals/{id}` con estado terminal no lo revierta la iteración en vuelo, y para que un `POST /stop` a secas siga guardando su contabilidad.
- `evaluator_goal_statement` impide llamar al juez de compleción con un objetivo vacío.
- `GateStreak` da a cada pata del verificador su propio cortacircuitos.
- `optional_uuid_field` para `verify_agent_id`, y un agente ya no puede ser su propio verificador (en creación y en actualización, comprobado sobre el par efectivo tras la escritura).

De este lado, injertado encima en vez de revertido a la forma de #7785:

- `max_iterations` sigue siendo `Option` hasta `GoalRunner::start`, para que un resume no pise el tope bajo el que estaba la ejecución pausada. El `unwrap_or(DEFAULT_GOAL_MAX_ITERATIONS)` que #7785 tiene en `goal_run_start` sigue eliminado.
- `resolve_workshop_config` mantiene la puerta de opt-in del taller de skills.
- `goal_tick_sender_context` mantiene el `chat_id` por objetivo que separa las sesiones.
- `pause: Arc<AtomicBool>` y `PAUSE_POLL_INTERVAL` conviven con el nuevo `StopFlag`.

`goal_run_resume` termina con sus siete parámetros — `goal_id`, `agent_id`, `max_iterations`, `loop_engineering`, `verify_agent_id`, `verify_max_retries`, `evaluator_model` — que es como lo llama `crates/librefang-api/src/routes/goals.rs`.

**Dos colisiones que git fusionó sin dejar marcador**, quitadas a mano:

- `goal_run_start_rejects_invalid_verify_max_retries` quedaba **definido dos veces** en `goals_routes_integration.rs`. Me quedé con el de #7785, que además cubre un valor negativo, y conservé el comentario de doc de esta rama.
- `progress_100_without_goal_done_still_finishes_with_no_verifier_configured` quedaba **definido dos veces** en `goal_runner.rs`. Me quedé con el de #7785, que afirma sobre el número de turnos y el progreso persistido en vez de solo sobre la fase.

**Invariante.** Para cada fichero que toca #7785, líneas suyas ausentes del resultado: **76**, en 6 ficheros, todas justificadas una a una en el mensaje del commit de fusión.
Son formas antiguas que esta rama sustituye deliberadamente — el `TICK_INTERVAL` fijo frente a la cadencia configurable, la unión de fases sin `paused`, la firma de 3 argumentos de `GoalRunner::new_with_store`, el `SenderContext` en línea que aquí es `goal_tick_sender_context`, el `unwrap_or(DEFAULT_GOAL_MAX_ITERATIONS)` — o líneas idénticas con otra indentación.
Contra el tip anterior de esta misma rama: **119** líneas, todas sustituciones por la versión más nueva de #7785, más las 6 del test duplicado que se quitó.

También he quitado una segunda línea `Optional body:` del doc comment de `start_goal_run`, que contradecía a la de arriba omitiendo `verify_max_retries`; ya estaba así antes de la fusión.

**Verificado**: `npx tsc --noEmit` limpio, `npx eslint .` limpio, `npx vitest run` 209/210 ficheros y 1.930/1.931 tests en verde, `pnpm build` limpio, `rustfmt --check` limpio en todos los `.rs` tocados.
El único test rojo es `PromptsExperimentsModal`, el caso del bucket de tráfico: es ajeno a este cambio y falla igual en `main`.

**No he compilado ni ejecutado la parte Rust**: el `target` compartido está en uso por otro trabajo. CI es la puerta.

El push fue **avance rápido** — el tip anterior `ed8fa353d` sigue siendo ancestro de `2adb4cba4` — así que no se reescribió historia y ningún comentario de revisión quedó huérfano.

**Revisiones**: los 9 hilos en línea tienen respuesta de @DaBlitzStein pero **ninguno está marcado como resuelto**; la última revisión de @houko es `APPROVED` (2026-09-12). No he tocado ni cerrado ningún hilo.

DaBlitzStein:
### Esta rama no compilaba: arreglado en `8fbb34da5`

Al ampliar `run_loop` de doce a catorce parámetros (`pause` en la posición once, `initial_learnings` en la catorce) y cambiar `stop` de `Arc<AtomicBool>` a `Arc<StopFlag>`, se quedaron sin actualizar:

- **16 llamadas** del módulo de test con la firma antigua de doce argumentos.
- **5 llamadas** que siguen pasando `Arc::new(AtomicBool::new(false))` donde ahora va el flag.
- **1 sitio en el bucle de pausa**: `stop.load(Ordering::SeqCst)` sobre un `StopFlag`, que no tiene `load`. Ahora es `stop.is_raised()`, igual que la comprobación de la cabecera del bucle.

Por qué no saltó antes: un `cargo check -p librefang-kernel` **sin** `--all-targets` compila solo la lib y no mira el módulo de test, y el error de la lib queda por delante tapando el resto.

`cargo check -p librefang-kernel --all-targets`: 0 errores.

PR #7835, opened 2026-08-24
feat(tui): shared folders editor for agents

DaBlitzStein:
Closes #7834.

**Stacked on #7749.**
The editor reads and writes the agent's manifest through `GET /api/agents/{id}/manifest`, which that PR adds; this branch merges `feat/agent-editor-parity` so it builds standalone.
Merge #7749 first and this PR reduces to its single last commit — `86d4945 feat(tui): shared folders editor for agents`.

## What this does

`w` in the agent detail opens a shared-folders editor:

- `a` adds a folder (name, path, read-write or read-only)
- `d` deletes the selected row
- Enter / Tab cycle the three fields of a row
- `s` saves, `Esc` goes back

The agent detail hint line advertises `w` alongside the existing `s` / `m` bindings.

## Why the write goes through the whole manifest

`PATCH /api/agents/{id}` with `manifest_toml` replaces the entire manifest.
So saving fetches the current TOML, replaces only the `[workspaces]` table, and PATCHes that back.
Building a manifest from the editor's own fields would silently drop every field the editor does not render — which is most of them.

Rows with an empty name or path are abandoned edits rather than declarations: they are filtered out before the write, so a half-typed row cannot produce a `[workspaces]` entry that the kernel then fails to resolve.

## Verification

Four unit tests in `crates/librefang-cli/src/tui/screens/agents.rs::workspaces_tests`:

- `adding_a_folder_focuses_the_name_field`
- `typing_then_tab_fills_name_and_path`
- `deleting_removes_the_selected_row`
- `save_emits_only_complete_rows`

Strings added to all four TUI locales (`en`, `ko`, `uk`, `zh-CN`).

## Verification not performed here

The workspace build was not run locally — a cross-compile was occupying the shared target directory.
CI is the gate for `cargo check` / `clippy` / the test run.

PR #7976, opened 2026-08-29
feat(dashboard,tui): show agent token footprint and recent calls

DaBlitzStein:
## Summary
- Dashboard: new "Token footprint" section in the agent detail drawer showing `injected_footprint_tokens` (from #7972) and the five most recent LLM calls from the existing events endpoint.
- TUI: `$` keybinding in the agent detail fetches the footprint and recent calls, rendering both in the ratatui detail pane.
- No new API endpoint or query hook — uses the existing `useAgentDetail` and `useAgentEvents` data.
- Locale keys added to all five dashboard languages (en, ko, pl, uk, zh) and four TUI languages (en, ko, uk, zh-CN).

Depends on #7972 (adds `injected_footprint_tokens` to the agent detail response).

## Verification
- `npx tsc --noEmit` — clean
- `vitest run locale-parity` — 8/8 passed
- `cargo clippy -p librefang-cli --all-targets -- -D warnings` — clean


houko (commented):
The feature is only reachable by pressing a key nothing tells you about.

`$` in the agent detail view is what fetches the footprint, and the only string that mentions it is the section header:

```
tui-agents-detail-tokens = Token footprint  ($ to refresh)
```

That header renders inside `if let Some(usage) = &state.token_usage`, so it exists only *after* the fetch has already happened.
Before the first press there is nothing on screen, in the hint bar, or in any locale file that names `$` — the hint is behind the door it unlocks.
Either put it in the detail view's hint bar alongside the other keys, or render the section with a "press $ to load" placeholder so the affordance exists before the data does.

Two smaller things in `crates/librefang-cli/src/tui/screens/agents.rs`:

**`AgentTokenUsage.total_tokens` does not hold a total.**
It is assigned from `body["injected_footprint_tokens"]` — the static system-prompt-plus-tools cost, which is per-request overhead, not a cumulative total for the agent.
The rendered label says "injected" so the screen is honest, but the field name is not, and the next reader to reach for `total_tokens` will get a number that means something else.
`injected_footprint_tokens` matches both the API field and the value.

**`recent: Vec<(String, u64, u64, f64)>`** costs a trip back to the producer to find out that the fields are model, input, output, cost — and at the render site it is destructured positionally, so transposing input and output would compile and be wrong.
A four-field struct with names would carry that at both ends. (#7995 adds a `UsageEventRow` with exactly these fields; if that lands, it is the shape to mirror.)

The locale coverage is right, by the way — all five dashboard locales including `pl.json`, which is the file #8013 is currently red on missing.

Unrelated: the `AgentAction` enum hunk reformats `ChatWithAgent`, `UpdateSkills`, `UpdateMcpServers` and `UpdateChannels` from one line each to four, which is rustfmt churn on variants this PR does not otherwise touch.
Harmless, but it triples the size of that hunk for a reviewer looking for the one added variant.

houko (changes requested):
Solid: it reuses the existing `GET /api/agents/{id}/events` (which already serves `usage_events` rows with exactly `model` / `input_tokens` / `output_tokens` / `cost_usd`), `injected_footprint_tokens` is already emitted by `GET /api/agents/{id}`, every locale is genuinely translated, and the i18n allowlist entry is a real technical format string in the right section. Two correctness issues in the TUI half.

## Blocking

**1. `AgentTokenUsageLoaded` carries no agent id, so a late response paints the wrong agent.**

```rust
AppEvent::AgentTokenUsageLoaded(usage) => {
    self.agents.token_usage = Some(usage);
}
```

`spawn_fetch_agent_token_usage` runs on its own thread per selection and makes two sequential HTTP calls. Select agent A, then B before A's fetch returns, and A's numbers land in `self.agents.token_usage` and render under B's name — with nothing to tell the operator the token count and cost belong to a different agent. Even without a race, `token_usage` is not cleared on selection change, so B shows A's figures until B's fetch completes.

Both are fixed by the same change: carry the agent id in the event, drop it when it does not match the current selection, and set `token_usage = None` when the selection changes.

**2. Both fetches swallow failures into a real-looking zero.**

```rust
if let Ok(body) = client.get(…).send().and_then(|r| r.json::<serde_json::Value>()) {
    usage.total_tokens = body["injected_footprint_tokens"].as_u64().unwrap_or(0);
}
```

A 404, a 500, a connection refused and a body missing the field all produce `total_tokens: 0` and an empty `recent` list, which renders as "this agent has no token footprint and has made no calls". Distinguish unavailable from zero — an error arm on the event, or at minimum a `status_msg`.

Related: the whole function is wrapped in `if let BackendRef::Daemon { .. } = backend`, with no `else`. In-process TUI users get no event at all, so the panel silently never populates. Same shape as the history fetch in #8073; both would benefit from an explicit "requires daemon mode" state.

## Non-blocking

- The dashboard fetches the default 30 events and then `.slice(0, 5)`, while the TUI asks for `?limit=5`. Same panel, same five rows — worth passing `limit: 5` on the dashboard query too, or at least agreeing on one number.
- Worth noting for #7995: that PR adds `UsageStore::recent_events`, which duplicates `list_agent_events_recent` — the method this PR's endpoint already uses. Since this is the consumer #7995's new method was presumably written for, and it does not need it, that method can be dropped there.
- No changelog fragment (`changelog.d/added/7976-....md`). A new per-agent cost/token panel in two surfaces is release-note material.

houko:

No commits of your own since my review on 2026-09-01T13:56:11Z; the only thing that moved the branch tip is a merge from `main`, which the `auto-update-branches` cron pushes by itself. The PR's "updated" timestamp has therefore advanced while the blocking items have not been touched. Worth flagging because that timestamp is what makes a PR look answered from outside — I nearly mis-triaged a dozen PRs on exactly that signal today.

Still open from that review:

- 1. `AgentTokenUsageLoaded` carries no agent id, so a late response paints the wrong agent
- 2. Both fetches swallow failures into a real-looking zero

DaBlitzStein:
Pushed `bd8803109`.

Merging `main` reported "Already up to date" — the previous head was itself a merge of `59ca61045`, so this branch was not stale. That confirms the red gate was only the cancelled run, with no test failure behind it.

Three things fixed here because they were in this PR's own code:

1. **The token-footprint fetch swallowed every error.** Both requests in `spawn_fetch_agent_token_usage` were wrapped in `if let Ok(...)`, so a 404, a 500 and a refused connection all rendered as a footprint of zero tokens with no recent calls — indistinguishable from an idle agent. On the in-process backend the `$` key did nothing at all. Both now go through `daemon_response` (the helper `main` introduced in #8141) and emit `AppEvent::FetchError` with the reason, with new keys in all four locales.
2. The diff had removed the `source_template` doc comment from #8018 in `dashboard/src/api.ts` — restored, and the new field documented alongside it.
3. Added the missing changelog fragment, `changelog.d/added/7976-agent-token-footprint.md`.

The one missing test was added too, covering `$` with and without the detail pane open. It was confirmed failing first — with the expected id changed to `agent-8` it fails with `left: "agent-7"` — so it discriminates rather than passing by construction.

Verification on the pushed tip: `cargo nextest run -p librefang-cli` — **495/495**, with `dollar_key_requests_the_footprint_for_the_open_agent_only`, `test_no_untranslated_strings`, `test_no_dead_locale_keys` and `test_locales_cover_used_i18n_keys` green by name. `cargo check` / `cargo clippy -- -D warnings` / `cargo fmt --check` all clean. `npx tsc --noEmit` clean; `npx vitest run src/pages/AgentsPage.test.tsx src/lib/queries/agents.test.tsx` — 2 files, 14 tests. `i18n-parity` failing set unchanged from `main`. Changelog attribution check passes.

Note on the first item: this is the fourth place in the TUI with the same shape — a `spawn_*` fetch that discards the error and emits its success event anyway, so a failure renders as empty data. #8144 fixed it for the memory and goals panes, #8059 for the Auxiliary tab, #8073 for version history, and this one for the token footprint. Filing that separately as a pattern rather than continuing to fix instances.

houko (changes requested):
Re-reviewed at the current head. Blocking #2 is fixed properly. Blocking #1 is untouched, so I am leaving this one open for it.

## #2 — done, and the daemon-only case with it

`daemon_response(...)` wraps both fetches, and each failure arm now sends `AppEvent::FetchError(message)` and returns instead of falling through to `AgentTokenUsageLoaded` with a zeroed struct. A 404, a 500 and a connection refused are no longer indistinguishable from "this agent has made no calls".

The `let BackendRef::Daemon { .. } = backend else` with a `tui-agents-token-usage-daemon-only` message is the explicit state I asked for on the in-process path — previously that branch produced no event at all and the panel just never populated.

## #1 — the stale-response race is unchanged

The event still carries no agent id:

```rust
AgentTokenUsageLoaded(crate::tui::screens::agents::AgentTokenUsage),
```

and the handler still applies it unconditionally:

```rust
AppEvent::AgentTokenUsageLoaded(usage) => {
    self.agents.token_usage = Some(usage);
}
```

`spawn_fetch_agent_token_usage` takes `agent_id: String`, uses it to build both URLs, and then drops it on the floor at the send. So the two halves of what I reported are both still live:

- **Late response paints the wrong agent.** Select A, select B before A's two sequential HTTP calls return, and A's token count and cost render under B's name with nothing indicating whose numbers they are.
- **No reset on selection change.** `token_usage` is initialised to `None` once at construction (`screens/agents.rs`) and never cleared when the selection moves, so B shows A's figures until B's own fetch completes — which happens on every selection change, race or no race.

Both close with the same edit: put the id in the event, ignore a payload whose id is not the current selection, and set `token_usage = None` when the selection changes. The action variant already carries the id (`AgentAction::FetchAgentTokenUsage(String)`), so it is a matter of threading it through rather than plumbing anything new.

Worth pinning with a test, since this is invisible in manual use unless you happen to click fast on a slow daemon: dispatch `AgentTokenUsageLoaded` for agent A while the selection is B, and assert `token_usage` stays `None`.

## Still non-blocking

The dashboard fetches the default page and does `.slice(0, 5)` while the TUI asks for `?limit=5`. Same panel, same five rows — worth agreeing on one number, but not a reason to hold this.

DaBlitzStein:
Addressed at `0d382d26e`.

**#1 is closed both ways.** `AgentTokenUsageLoaded` is a struct variant carrying `agent_id` now, and `apply_token_usage` only folds the payload in when it matches the current selection; `token_usage` is set to `None` when the selection changes, so B no longer shows A's figures while its own fetch is in flight. The id was already on `AgentAction::FetchAgentTokenUsage`, so it was threading rather than new plumbing.

Both tests, and the red separates the two halves cleanly. Removing the id guard fails the race test and leaves the reset test green, because the reset lives elsewhere:

```
a_footprint_for_another_agent_is_ignored ... FAILED
opening_another_agent_drops_the_previous_footprint ... ok
```

The first is the one you asked for — `AgentTokenUsageLoaded` dispatched for A while the selection is B, asserting `token_usage` stays `None`. Both green with the fix.

On the non-blocking point: the dashboard's `.slice(0, 5)` and the TUI's `?limit=5` still disagree in mechanism. Left as-is here since you did not want it holding the PR, but flagging that I read it and did not silently skip it.

Ready for another look — and the new P2 on `extraction_model` in #7985 is being handled there.

crates/librefang-api/dashboard/src/pages/AgentsPage.tsx
```diff
@@ -3130,6 +3130,37 @@ export function AgentsPage() {
                 )}
               </div>
 
+              {detailAgent.injected_footprint_tokens != null && detailAgent.injected_footprint_tokens > 0 && (
```
houko: The `> 0` half of this guard treats a genuine zero as "no data", and it also gates the **Recent calls** list, which does not depend on the footprint at all.

`available_tools` returns an empty `Vec` for an agent with `tools_disabled = true` (`crates/librefang-kernel/src/kernel/tools_and_skills.rs:177`), and the footprint is `estimate_token_count(&[], Some(&system_prompt), Some(&[]))` (`crates/librefang-api/src/routes/agents/lifecycle.rs:1078`). So for a tools-disabled agent whose manifest carries no `system_prompt` (identity files supply the persona instead), the daemon truthfully answers `injected_footprint_tokens: 0` — and the drawer then renders nothing at all: not the "0" that would confirm the agent is cheap, and not the five recent LLM calls, which are sourced from `/api/agents/{id}/events` and are perfectly available.

Suggest gating the section on presence only, and gating the recent-call rows on their own data:

```suggestion
              {detailAgent.injected_footprint_tokens != null && (
```

crates/librefang-cli/src/tui/event.rs
```diff
@@ -2240,6 +2250,70 @@ pub fn spawn_update_agent_channels(
...
+        match daemon_response(
+            client
+                .get(format!("{base_url}/api/agents/{agent_id}"))
+                .send(),
+            || crate::i18n::t("tui-agents-token-usage-failed"),
+        )
+        .and_then(|r| r.json::<serde_json::Value>().map_err(|e| e.to_string()))
+        {
+            Ok(body) => {
+                usage.total_tokens = body["injected_footprint_tokens"].as_u64().unwrap_or(0);
```
houko: `unwrap_or(0)` turns a missing field into a confident "injected 0".

bd8803109 already established the rule for this panel — a fetch that did not produce a number must say so rather than paint a zero — but that only covers the transport/status failure. A 200 whose body has no `injected_footprint_tokens` (a TUI talking to a daemon older than #7972, or a future rename of the key) lands here and reports an agent with zero static footprint, which is indistinguishable from the real zero case and is the one number the operator pressed `$` to get.

Prefer `as_u64()` returning `None` to go down the same `FetchError` path as a non-200:

```suggestion
                usage.total_tokens = match body["injected_footprint_tokens"].as_u64() {
                    Some(n) => n,
                    None => {
                        let _ = tx.send(AppEvent::FetchError(crate::i18n::t(
                            "tui-agents-token-usage-failed",
                        )));
                        return;
                    }
                };
```

crates/librefang-cli/src/tui/screens/agents.rs
```diff
@@ -1426,6 +1469,30 @@ fn draw_detail(f: &mut Frame, area: Rect, state: &AgentSelectState) {
                 ]));
             }
 
+            if let Some(usage) = &state.token_usage {
```
houko: This block is appended to a `Paragraph` that is rendered without `.scroll(...)`, so anything past the bottom of `chunks[0]` is silently clipped.

Budget on an 80x24 terminal: 1 row tab bar (`tui/mod.rs:3005`) + 2 rows screen border + 1 row hint bar leaves the detail pane 20 rows. A typical daemon agent already draws 14–16 lines (blank, ID, Name, State, Provider, Model, Created, Active, optional Tags/Caps/Parent/Children, blank, Skills, MCP, Channels), and this block adds 2 + up to 5 more. So pressing `$` on an agent that has tags and capabilities either shows only part of the recent-call list or, on a shorter terminal, produces no visible change whatsoever — no error, no scroll, nothing to distinguish it from an unbound key.

Either render this block first (right under the header, where it cannot be pushed off), or give the paragraph a scroll offset.

crates/librefang-cli/locales/en/main.ftl
```diff
@@ -2086,6 +2086,10 @@ tui-agents-title-templates = Templates
 tui-agents-title-custom-name = Custom — Name
 tui-agents-title-custom-desc = Custom — Description
 tui-agents-title-custom-prompt = Custom — System Prompt
+tui-agents-detail-tokens = Token footprint  ($ to refresh)
```
houko, line 2098: `tui-agents-hints-detail` is not updated, in any of the four locales, so `$` never appears in the detail screen's hint bar — it still reads `[s] Edit skills  [m] Edit MCP  [n] Edit channels  [p] Model params  [c] Chat  [k] Kill  [Esc] Back` (`crates/librefang-cli/locales/en/main.ftl:2127` and the ko / uk / zh-CN equivalents).

The only in-app mention of the key is `($ to refresh)` inside `tui-agents-detail-tokens`, which is rendered by the panel that only exists after the operator has already pressed `$`. As shipped, the feature is reachable only by reading the source or the changelog. Adding `[$] Tokens` to the four hint strings costs four lines.

crates/librefang-cli/src/tui/event.rs
```diff
@@ -2240,6 +2250,70 @@ pub fn spawn_update_agent_channels(
     });
 }
 
+pub fn spawn_fetch_agent_token_usage(
+    backend: BackendRef,
+    agent_id: String,
+    tx: mpsc::Sender<AppEvent>,
+) {
+    std::thread::spawn(move || {
+        let BackendRef::Daemon { base_url, api_key } = backend else {
```
houko: Every other per-agent fetch on this screen serves both backends — `spawn_fetch_agent_skills` (`event.rs:1802`), `spawn_fetch_agent_mcp_servers` and `spawn_fetch_agent_channels` all have a `BackendRef::InProcess(kernel)` arm — so in in-process TUI mode `$` is the one key on the detail screen that can only ever produce an error line.

Both halves are computable locally: the footprint is `compactor::estimate_token_count` over `kernel.available_tools(aid)` plus `manifest.model.system_prompt` (exactly what `routes/agents/lifecycle.rs:1078` does), and the recent calls are `memory_substrate().usage().list_agent_events_recent(aid, 5)`, the same call the route makes. That is an `InProcess` arm of roughly a dozen lines rather than a capability gap.

crates/librefang-api/dashboard/src/pages/AgentsPage.tsx
DaBlitzStein: Fixed — gated the panel on `injected_footprint_tokens != null` only, extracted as a named type-predicate `hasTokenFootprintData` (matching the existing `cloneResultNotice` pattern in this file since the drawer has no render harness). The recent-calls list already had its own independent `.length > 0` gate and is now reached regardless of the footprint value. Added `hasTokenFootprintData` unit tests in `AgentsPage.test.tsx` (zero, non-zero, null/undefined) — verified red against the old `> 0` guard, green with the fix.

crates/librefang-cli/src/tui/event.rs
DaBlitzStein: Fixed — `.as_u64()` now matches on `Some`/`None`, and `None` routes down the same `FetchError` path as a non-200 response instead of `unwrap_or(0)`. Added a raw-socket one-shot mock server in `event.rs`'s test module (no existing HTTP mock helper covered response-body shape, only transport failure) and a regression test asserting a 200 missing `injected_footprint_tokens` produces `FetchError`, not a fake zero — verified it panics/fails against the pre-fix `unwrap_or(0)`, passes with the fix.

crates/librefang-cli/src/tui/screens/agents.rs
DaBlitzStein: Fixed — moved the token-footprint block from the end of the `lines` vec to right after the fixed header (Model line), before the variable-length optional sections (Created/Active/Tags/Caps/Parent/Children/Skills/MCP/Channels), so it can no longer be pushed past the bottom of the pane. Went with relocation over a scroll offset since the screen has no scroll state today and this is the smaller change.

crates/librefang-cli/locales/en/main.ftl
DaBlitzStein: Fixed — added `[$] Tokens` (and localized equivalents) to `tui-agents-hints-detail` in all four locales (en/ko/uk/zh-CN).

crates/librefang-cli/src/tui/event.rs
DaBlitzStein: Fixed — added a `BackendRef::InProcess` arm to `spawn_fetch_agent_token_usage` mirroring `routes/agents/lifecycle.rs:1078`: `librefang_kernel::compactor::estimate_token_count` over `kernel.available_tools(aid)` + `manifest.model.system_prompt` for the footprint, and `memory_substrate().usage().list_agent_events_recent(aid, 5)` for the recent calls (the `KernelApi` trait method needed a fully-qualified call to avoid an ambiguity with the `AgentSubsystemApi`/`McpSubsystemApi`/`SkillsSubsystemApi` traits already imported in this file). The now-unreachable `tui-agents-token-usage-daemon-only` locale key was removed from all four locales.

DaBlitzStein:
Every one of the 5 review threads on this PR now has an inline reply, and the branch is at `9c27f301c`.

houko:
Structural note before the code review: this PR is part of a group of six that restructure the same agent-manifest editing surface without being stacked on each other — #7835, #7749, #7976, #8028, #8231 and #8041, plus #8013 which has now merged. `git merge-base --is-ancestor` is false in both directions for every pair, so they are diverged copies rather than a stack, and `tui/screens/agents.rs` / `AgentsPage.tsx` / `agentManifest.ts` are being rewritten by several of them at once against bases that predate the others.

Full write-up, with the shared-file counts per pair, is on #7835. I am holding the code review here until a merge order is picked and the rest are rebased — reviewing now means reviewing the same restructuring several times and having most of those reviews invalidated by the first merge.

Any findings already open on this PR still stand on their own.

DaBlitzStein:
@houko — the block from `2026-09-03` is asking for an edit that is already at the head, so this is a request for re-review rather than a new push. Nothing changed here; I checked because the PR is still `CHANGES_REQUESTED`.

**Blocking #1 — all three parts are in, at `cf4f3d843`:**

- **The event carries the id.** `crates/librefang-cli/src/tui/event.rs:356` is `AgentTokenUsageLoaded { agent_id, usage }`, and both send sites (`:2295`, `:2353`) pass the `agent_id` that `spawn_fetch_agent_token_usage` already had.
- **A payload for another agent is ignored.** `screens/agents.rs:277` — `apply_token_usage(&mut self, agent_id: &str, usage)` folds the payload in only `if self.detail.as_ref().is_some_and(|d| d.id == agent_id)`. `tui/mod.rs:515` destructures both fields and routes through it rather than assigning unconditionally.
- **The selection change clears it.** `screens/agents.rs:527` sets `token_usage = None` on the path that opens a detail, and `:269` does the same on reset — so B no longer shows A's figures while B's own fetch is in flight, race or no race.

**And the test you asked for is there**: `a_footprint_for_another_agent_is_ignored` (`screens/agents.rs:1922`) puts `agent-2` in `detail`, dispatches a payload for `agent-1`, asserts `token_usage` stays `None`, then dispatches `agent-2`'s own and asserts it lands.

**The non-blocking one does not reproduce, though.** The dashboard is not fetching a different number for the same panel — `useAgentEvents(id, 30)` backs **two** surfaces: the Logs tab at `AgentsPage.tsx:2317`, which renders the full list as a terminal-style turn feed, and the footprint panel at `:3163`, which slices 5. Dropping the request to `?limit=5` to match the TUI would truncate the Logs tab from 30 rows to 5. Both surfaces already render the same five rows in the footprint panel; the difference is one query serving two panels on the web side and only one on the TUI side. Happy to change it if you would rather the footprint panel had its own query, but as it stands there is no number to agree on.

Your five inline findings from `2026-09-07` were each answered in their own thread on `2026-09-09`; none is outstanding.

houko (approved):
Re-reviewed at the current head. Blocking #1 is fixed.

`0d382d26` gives the event the agent id it was missing — `AgentTokenUsageLoaded { agent_id, usage }` — and the handler routes it through `apply_token_usage(&agent_id, usage)`, which applies the payload only when `detail.id == agent_id`. A response that lands after the operator has moved to another agent is dropped instead of being painted onto whoever is on screen.

With #2 already done in the previous round, that closes both.

PR #7980, opened 2026-08-29
fix(cli): deduplicate ClawHub marketplace results by slug

### LINKED ISSUE #7978: fix(cli): TUI marketplace shows duplicate skills from ClawHub index
The ClawHub index sometimes returns entries for the same slug with different capitalisation (e.g. "Prd" and "prd").
`parse_clawhub_results` in `crates/librefang-cli/src/tui/event.rs` forwards every entry verbatim, so the TUI marketplace renders the skill twice — and pressing install on one appears to install both.

Fix: deduplicate by lowercased slug before returning.

DaBlitzStein:
## Summary
- The ClawHub index can return entries for the same slug with different capitalisation (e.g. "Prd" and "prd"), causing the TUI marketplace to render a skill twice.
- Deduplicates by lowercased slug in `parse_clawhub_results` before returning.

## Verification
- `cargo check -p librefang-cli --lib` clean

Closes #7978

### DIFF
--- a/crates/librefang-cli/src/tui/event.rs
+++ b/crates/librefang-cli/src/tui/event.rs
@@ -1957,7 +1957,6 @@ fn parse_clawhub_results(body: &serde_json::Value) -> Vec<ClawHubResult> {
         .and_then(|v| v.as_array())
         .or_else(|| body.as_array());
 
-    let mut seen = std::collections::HashSet::new();
     items
         .map(|arr| {
             let mut seen: std::collections::HashSet<String> = std::collections::HashSet::new();
--- a/crates/librefang-cli/src/tui/event.rs
+++ b/crates/librefang-cli/src/tui/event.rs
@@ -1957,6 +1957,7 @@ fn parse_clawhub_results(body: &serde_json::Value) -> Vec<ClawHubResult> {
         .and_then(|v| v.as_array())
         .or_else(|| body.as_array());
 
+    let mut seen = std::collections::HashSet::new();
     items
         .map(|arr| {
             let mut seen: std::collections::HashSet<String> = std::collections::HashSet::new();


houko (commented):
This is the same fix as #7958, on the same function and the same line, and #7958 is the stronger version of it — so I have merged that one and left this open for you to close or repurpose.

Both add a lowercased-slug `HashSet` filter to `parse_clawhub_results`.
The differences:

- #7958 declares `seen` **inside** the `.map()` closure; this PR declares it outside, in the enclosing function.
  Same behaviour today because the closure runs at most once, but the outer binding is live even on the `None` path and would be shared if that `map` ever became a loop.
- #7958 carries three tests — same slug in different casings collapses to one row, two distinct slugs sharing a display name stay two rows, and the bare-array shape (no `items` envelope) goes through the same path.
  This PR has none, and the bare-array case is exactly the one an untested dedupe tends to miss.
- #7958 explains in a comment why first-wins is correct (the index returns relevance order, so the first spelling of a slug is the one the server ranked highest) and ships a changelog fragment.

Nothing here is wrong; it is just the smaller half of a change that already landed.

houko (changes requested):
This PR has no changes — `+0/-0`, 0 files. `gh pr diff 7980` returns nothing.

The two commits on `fix/clawhub-dedupe` cancel each other out:

```
fix(cli): deduplicate ClawHub marketplace results by slug
fix(tui): remove dead unused HashSet in clawhub dedupe
```

so the branch tip is identical to `main` and merging it would fix nothing. The described behaviour — `parse_clawhub_results` collapsing entries whose slugs differ only in case — is not present on this branch.

Please either push the actual fix or close this and reopen from a branch that carries it. If it turns out the dedupe already landed on `main` through another PR, close this one and close #7978 from whichever PR did the work, so the issue is attributed to the change that fixed it.

When the fix is pushed, it also needs:

- A unit test on `parse_clawhub_results` with two entries differing only in slug capitalisation, asserting one result survives and that it is the expected one — case-insensitive dedupe has to pick a winner, and "whichever came first in the index response" is a decision worth pinning rather than inheriting.
- A changelog fragment (`changelog.d/fixed/7980-....md`).

DaBlitzStein:
Closing — this branch has no changes (+0/-0): the two commits cancel each other out, so the tip is identical to main. The case-insensitive slug dedupe it describes already landed on main via #7958 (parse_clawhub_results now filters on lowercased slug, first-wins). Issue #7978 is closed against #7958.

PR #7981, opened 2026-08-29
fix(cli): a failed skill install in the TUI says what the daemon reported

### LINKED ISSUE #7979: fix(cli): a failed skill install in the TUI says what the daemon reported
When a ClawHub skill install fails (HTTP 4xx/5xx), `spawn_install_skill` catches the error arm with a generic `"Failed to install {slug}"` message.
The daemon response body usually contains a structured `error` field with the real reason (e.g. "YAML parse error", "skill not found on registry").

Fix: read the response body on non-success status and surface the daemon's error message, falling back to `"HTTP {status}"` when the body has no `error` field.

DaBlitzStein:
## Summary
- When a ClawHub skill install fails (HTTP 4xx/5xx), `spawn_install_skill` now reads the daemon response body and surfaces its `error` field instead of a generic "Failed to install" message.
- Falls back to `HTTP {status}` when the body has no `error` field, and to the connection error string on transport failure.
- Updated i18n key `tui-event-skill-install-failed` in all 4 CLI locales (en, ko, uk, zh-CN) to accept a `$detail` parameter.

## Verification
- `cargo check -p librefang-cli --lib` clean

Closes #7979


DaBlitzStein:
Duplicate of #7959, which wraps the HTTP fallback through i18n correctly.

PR #7983, opened 2026-08-29
fix(skills): map marketplace connectivity errors to a dedicated variant

DaBlitzStein:
## Summary
- ClawHub: 3 new tests verify that HTML responses from search, get_skill, and install map to `MarketplaceUnavailable` instead of the generic `Network` error
- Skillhub: extract `parse_skillhub_search_body` and `parse_skillhub_index_body` so the markup-gate + dual-format parsing logic is unit-testable without a live network call
- Skillhub: missing-skill lookup now returns `NotFound` instead of `Network`
- 7 new unit tests for the extracted Skillhub parsers, including env-var override coverage

## Verification
- `cargo check -p librefang-skills --lib` passes
- `cargo test -p librefang-skills` passes (274 tests, 0 failures)


houko (commented):
Merging — the graceful-degradation tests are the right ones (HTML body, valid JSON, and genuinely malformed JSON kept as a separate `Network` error rather than folded into `MarketplaceUnavailable`), and extracting the parse out of the network call is what makes them possible without a live socket.

Two notes for later, neither blocking.

`parse_skillhub_index_body` is a one-line forward to `crate::parse_marketplace_json("Skillhub index", index_url, body)` with no logic of its own.
The doc comment carries real information, but the indirection means a reader chasing the index parse now goes through a hop that does nothing.
Calling the shared helper directly with the comment moved to its call sites would read the same and be one layer shallower.

The `Skill '{slug}' not found in Skillhub index` error changed from `SkillError::Network` to `SkillError::NotFound`, which moves that response from `502` to `404`.
That is the right status — a slug the index does not list is not a connectivity failure — but it is a user-visible API change that neither the title nor the PR body mentions, so it will surprise anyone reading the diff for the connectivity work alone.

This PR also has no `changelog.d/` fragment; its bullet (`fixed/7387-skillhub-marketplace-dead-api.md`) is in #7988 instead.
That fragment describes the `503` mapping and the `LIBREFANG_SKILLHUB_URL` override, but not the `404`, so whichever PR ends up carrying it is worth a line about the status change too.

PR #7984, opened 2026-08-29
fix(llm): mute reasoning_effort after the gateway rejects it

DaBlitzStein:
## Summary
- Add unsupported-parameter detection patterns covering litellm, OpenAI, vLLM/FastAPI error shapes
- Per-model mute map suppresses reasoning_effort for 10 min after rejection
- Prevents noisy error logs and wasted round-trips to gateways that don't support the param

## Verification
- `cargo check -p librefang-llm-driver --lib` passes
- `cargo check -p librefang-llm-drivers --lib` passes


houko (commented):
The strip-and-retry-then-mute design is right, and the TTL argument for why the mute must expire — gateway parameter support is operator configuration, a permanent set would need a daemon restart to revoke — is the correct call.
Two things to fix before this lands.

**The new const was inserted into the middle of another item's doc comment.**

`crates/librefang-llm-drivers/src/drivers/openai.rs`, at the `REASONING_EFFORT_MUTE_TTL` hunk:

```rust
/// There is no cross-vendor convention for requesting extended thinking over the OpenAI-compatible wire; `reasoning_effort` … is the most widely supported opt-in …
/// Budgets below 1024 tokens are treated as thinking-off — the same minimum-budget gate the anthropic driver applies.
/// How long a model stays muted after rejecting `reasoning_effort`.
///
/// Short enough that an operator who reconfigures the gateway sees the field
/// again without a daemon restart, …
const REASONING_EFFORT_MUTE_TTL: std::time::Duration = std::time::Duration::from_secs(600);

fn reasoning_effort_for_budget(budget_tokens: u32) -> Option<&'static str> {
```

Those first two lines documented `reasoning_effort_for_budget`.
They now attach to the const, so the TTL constant is documented as being about cross-vendor thinking conventions and token budgets, and the function it was written for has no doc comment at all.
Moving the const above the whole block, with the original paragraphs left on the function, fixes it.

**`reasoning_effort_rejected` does a full-map sweep on the read path.**

```rust
fn reasoning_effort_rejected(&self, model: &str) -> bool {
    let now = std::time::Instant::now();
    self.reasoning_effort_unsupported
        .retain(|_, at| now.duration_since(*at) < self.reasoning_effort_mute_ttl);
    self.reasoning_effort_unsupported.contains_key(model)
}
```

`DashMap::retain` takes a write lock on every shard, and this runs on every `build_request` — so a check that only ever needs to read one key serialises all of them, on the hot path, for a map that holds at most a handful of entries.
Reading the one entry expresses the same intent without the sweep:

```rust
self.reasoning_effort_unsupported
    .get(model)
    .is_some_and(|at| now.duration_since(*at) < self.reasoning_effort_mute_ttl)
```

An expired entry then simply reads as "not muted" and is overwritten on the next rejection, which is the behaviour the TTL wants anyway.
If you want the map bounded as well, prune on the write path in `record_reasoning_effort_rejected` — that one is rare by construction.

**Non-blocking:** this PR has no `changelog.d/` fragment; the bullet for it (`fixed/litellm-reasoning-effort-unsupported-param.md`) is sitting in #7988 instead.
It would be better here, where the PR number is known — see the separate comment on #7988.

houko (changes requested):
Turning the permanent mute into a TTL is the right call, the reasoning in the field doc is well argued, and the mock-gateway tests that assert *which* attempts carried the field are exactly the right shape. Three things.

## Blocking

**1. The new const swallows `reasoning_effort_for_budget`'s doc comment.**

The insertion lands between an existing doc block and the function it documents:

```rust
/// There is no cross-vendor convention for requesting extended thinking over the OpenAI-compatible wire; `reasoning_effort` … is the most widely supported opt-in …
/// Budgets below 1024 tokens are treated as thinking-off — the same minimum-budget gate the anthropic driver applies.
/// How long a model stays muted after rejecting `reasoning_effort`.
///
/// Short enough that an operator who reconfigures the gateway sees the field
/// again without a daemon restart, …
const REASONING_EFFORT_MUTE_TTL: std::time::Duration = std::time::Duration::from_secs(600);

fn reasoning_effort_for_budget(budget_tokens: u32) -> Option<&'static str> {
```

So the const's rustdoc now opens with two paragraphs about the OpenAI-compatible thinking convention and the 1024-token floor, and `reasoning_effort_for_budget` — the function those paragraphs describe — is left undocumented. Move the const above the existing doc block (or below the function).

**2. `reasoning_effort_rejected` sweeps the entire map on every request.**

```rust
fn reasoning_effort_rejected(&self, model: &str) -> bool {
    let now = std::time::Instant::now();
    self.reasoning_effort_unsupported
        .retain(|_, at| now.duration_since(*at) < self.reasoning_effort_mute_ttl);
    self.reasoning_effort_unsupported.contains_key(model)
}
```

`DashMap::retain` walks and write-locks **every shard**, and this runs from `build_request` on every completion and every stream — so a predicate that reads one key serializes all concurrent requests through the whole map, for all models, on each call. It also mutates behind a `&self` that reads as a query.

The single-key form is O(1) and touches one shard:

```rust
fn reasoning_effort_rejected(&self, model: &str) -> bool {
    match self.reasoning_effort_unsupported.get(model) {
        Some(at) if at.elapsed() < self.reasoning_effort_mute_ttl => true,
        Some(_) => { drop(_); self.reasoning_effort_unsupported.remove(model); false }
        None => false,
    }
}
```

(with care to drop the `Ref` before `remove`, or use `remove_if`). Stale entries for models never asked about again just sit there, which is fine — the map is bounded by the number of models this endpoint serves.

**3. No test covers the TTL, which is the whole point of the change.**

The tests prove strip-and-retry and that the field is omitted on the next call. Nothing asserts that the mute *expires* — and `reasoning_effort_mute_ttl` was made a per-instance field specifically so a test could shorten it. Construct a driver with a millisecond TTL, trigger a rejection, sleep past it, and assert the next request carries `reasoning_effort` again (the mock server's `seen` vector already records exactly that). Without it, the field is injectable-but-uninjected and the self-healing behaviour the doc promises is unverified.

## Non-blocking

- Re-recording restarts the TTL, and the doc says that is intended. Worth noting the consequence: a gateway that keeps rejecting is re-probed every 600 s per model regardless, so the "genuinely rejecting gateway is not retried on every turn" claim holds at a floor of one wasted 400 per model per 10 minutes. That is a reasonable trade, just worth being explicit since the const is the only knob.
- The added `"does not support parameter"` (singular) and `"unrecognized request argument"` patterns have no test case; the three new tests cover the litellm, OpenAI-code and Python-`TypeError` shapes only. Two more lines in `test_is_unsupported_parameter_error_*` would cover the patterns you are adding rather than the ones that already existed.
- `Security` / `CI Gate` are red on pre-existing `cargo audit` advisories in transitive dependencies (`lru` 0.18.0 RUSTSEC-2026-0253, `rand` 0.7.3 RUSTSEC-2026-0097), not on this diff.
- No changelog fragment. "A gateway that rejects `reasoning_effort` is now remembered for ten minutes instead of forever, so reconfiguring it no longer needs a daemon restart" is a real operator-visible change: `changelog.d/fixed/7984-....md`.

DaBlitzStein:
All review points are addressed in `4bbe622ff`, and the branch now merges cleanly into `main` (merge commit `773db32c7` brings it up to `07cb6fc76`; conflicts were in `openai.rs`'s two strip-and-retry comment blocks and `is_unsupported_reasoning_effort_error`'s rustdoc, resolved to the upstream wording).

**Blocking 1 — const swallowed `reasoning_effort_for_budget`'s doc comment.** `REASONING_EFFORT_MUTE_TTL` now sits above the doc block in `crates/librefang-llm-drivers/src/drivers/openai.rs`, so the two paragraphs about the OpenAI-compatible thinking convention are back on the function and the const carries only its TTL doc.

**Blocking 2 — the full-map sweep.** `reasoning_effort_rejected` no longer calls `DashMap::retain`: it reads the single entry for `model` (`get(model).is_some_and(|at| at.elapsed() < self.reasoning_effort_mute_ttl)`), touches one shard, and mutates nothing behind `&self`. An expired entry reads as "not muted" and is overwritten by the next recorded rejection, so the map stays bounded by the number of models the endpoint serves.

**Blocking 3 — the TTL is now tested end to end.** `mute_expires_and_field_is_resent_after_the_ttl_passes` builds a driver with a 50 ms TTL, learns a live 400 from the mock gateway, sleeps past the TTL, and asserts the `seen` sequence is `[true, false, true, false]` — the second `complete()` puts `reasoning_effort` back on the wire and re-learns the 400. The pre-existing `build_request`-level expiry tests still pass.

**Non-blocking, also done.** The const doc now states the consequence of re-record restarting the clock (one relearned 400 per still-rejecting model per TTL window). The two patterns this PR added to `UNSUPPORTED_PARAM_PATTERNS` (singular "does not support parameter", OpenAI's "unrecognized request argument") have cases in `test_is_unsupported_parameter_error_singular_and_unrecognized_shapes`. The changelog fragment landed here as `changelog.d/fixed/7984-reasoning-effort-mute-expires.md` instead of riding along in #7988.

Verification, all scoped per the repo rules: `cargo check -p librefang-llm-driver -p librefang-llm-drivers` clean; `cargo clippy -p librefang-llm-driver -p librefang-llm-drivers --all-targets -- -D warnings` zero warnings; `cargo test -p librefang-llm-drivers --lib openai::tests::` 114 passed, `cargo test -p librefang-llm-driver --lib` 73 passed, including the three mute-expiry tests (`reasoning_effort_mute_expires_after_the_ttl`, `a_fresh_rejection_still_mutes_the_field`, `mute_expires_and_field_is_resent_after_the_ttl_passes`). One unrelated test (`gemini_cli::complete_returns_json_message_and_usage`) fails on this host with "Text file busy" while spawning the gemini CLI — environment, not this diff.

houko (approved):
All three fixed, verified on the head.

**1 — rustdoc.** `REASONING_EFFORT_MUTE_TTL` moved to `:747` and `reasoning_effort_for_budget` has its own doc block back at `:749-752`. The const also gained a sentence I did not ask for and that is the useful one: because a re-recorded rejection restarts the clock, the floor "not retried on every turn" holds at is one relearned 400 per still-rejecting model per TTL window. That is the actual cost of the design, stated.

**2 — the map sweep is gone**, and the result is cleaner than what I proposed:

```rust
fn reasoning_effort_rejected(&self, model: &str) -> bool {
    self.reasoning_effort_unsupported
        .get(model)
        .is_some_and(|at| at.elapsed() < self.reasoning_effort_mute_ttl)
}
```

No `retain`, one shard, and no mutation at all — so it reads as the query it is, rather than a query that quietly writes behind `&self`. Skipping the opportunistic `remove` is the right trade: the map is bounded by the models this endpoint serves, and a stale entry is overwritten on the next rejection.

**3 — the TTL is covered twice**, unit and end-to-end. `mute_expires_and_field_is_resent_after_the_ttl_passes` is the one that matters, and it asserts the right thing: `seen == [true, false, true, false]`, so the third HTTP attempt carrying `reasoning_effort` again *is* the assertion, not a side effect of one. Shortening `reasoning_effort_mute_ttl` to 50 ms on the instance is exactly what the per-instance field was for, so the injectable-but-uninjected gap is closed.

`CI Gate` green, and `mergeable` is `BLOCKED` on review rather than a conflict — so this one escaped the rebase churn my merge sweep caused elsewhere.

Approving.

PR #7985, opened 2026-08-29
feat(tui): interactive memory config editing in the Memory screen

DaBlitzStein:
## Summary
- Add `ConfigField` enum for navigable config rows (auto-memorize, auto-retrieve, extraction model)
- Support j/k navigation, space/enter toggling, inline model name editing with visible cursor
- Track dirty state and save with `s` key via `PATCH /api/memory/config`
- Show unsaved work indicator; Esc discards unsaved changes
- Wire `SaveConfig` action through event loop to `spawn_save_memory_config`
- Add i18n keys for unsaved/saving states across all 4 locales (en, ko, uk, zh-CN)

## Verification
- `cargo check -p librefang-cli` passes
- `cargo clippy -p librefang-cli -- -D warnings` clean
- `cargo test -p librefang-cli -- tui::screens::memory::tests` — 12 tests pass (5 existing + 7 new)

## Test plan
- [ ] `cargo check --workspace --lib` passes
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [ ] `cargo test -p librefang-cli -- memory` runs the new TUI tests
- [ ] Toggle auto_memorize marks panel dirty
- [ ] Model editing swallows navigation keys
- [ ] Esc during model edit reverts to original value


houko (commented):
A successful save reports the word "on".

`crates/librefang-cli/src/tui/mod.rs`:

```rust
AppEvent::MemoryConfigSaved(ok) => {
    self.memory.status_msg = if ok {
        crate::i18n::t("tui-memory-config-on")
    } else {
        crate::i18n::t("tui-memory-config-save-failed")
    };
}
```

`tui-memory-config-on` is the toggle label — `on` / `켬` / `увімк.` / `开` — paired with `tui-memory-config-off` for rendering a boolean row.
Using it as the save confirmation puts a bare "on" in the status line where the operator is looking for confirmation that the write landed, and in the CJK locales it is a single character with no verb in it at all.

This PR adds `tui-memory-config-saving`, `tui-memory-config-save-failed` and `tui-memory-config-unsaved` to all four locales but no `tui-memory-config-saved`, which is the key this arm wants.
Adding it to the four `main.ftl` files and using it here is the whole fix.

Second, smaller: `config_dirty` is cleared when the save is *dispatched*, not when it succeeds.

```rust
self.config_dirty = false;
self.status_msg = crate::i18n::t("tui-memory-config-saving");
```

If the request fails, the status line says "Save failed" but the unsaved-changes marker is already gone, so the panel simultaneously reports a failure and claims there is nothing outstanding.
Restoring `config_dirty = true` in the `MemoryConfigSaved(false)` arm keeps the two consistent.

The `ConfigField` enum and the "typing swallows navigation keys" guard are both the right call — that part is good.

houko (changes requested):
The wiring is correct — `PATCH /api/memory/config` exists and reads exactly `proactive_memory.auto_memorize` / `auto_retrieve` / `extraction_model` (`routes/memory.rs:1843-1852`) — and the `ConfigField` enum with its "adding a row cannot silently shift what a keypress edits" rationale is the right shape. Three bugs in the save path.

## Blocking

**1. The success message says "on".**

```rust
AppEvent::MemoryConfigSaved(ok) => {
    self.memory.status_msg = if ok {
        crate::i18n::t("tui-memory-config-on")
    } else {
        crate::i18n::t("tui-memory-config-save-failed")
    };
}
```

`tui-memory-config-on` is the boolean row label — "on" / "켬" / "увімк." / "开". After a successful save the operator sees a bare "on". The PR adds `-saving` and `-save-failed` but no `-saved`; add it to all four `.ftl` files.

**2. `config_dirty` is cleared on keypress, before the save is known to have worked.**

```rust
KeyCode::Char('s') => {
    if let Some(cfg) = &self.config {
        let action = MemoryUIAction::SaveConfig { … };
        self.config_dirty = false;
        self.status_msg = crate::i18n::t("tui-memory-config-saving");
```

If the PATCH fails, the status line says "Save failed" but the unsaved-changes marker is already gone — and its own string is the thing telling the operator their edits are still pending ("Unsaved changes — press s to save, Esc to discard"). So a failed save silently downgrades to "everything is saved" with a transient error next to it. Clear the flag in the `MemoryConfigSaved(true)` arm instead.

Note this also means updating the new test that currently asserts the premature behaviour:

```rust
assert!(!state.config_dirty, "a save clears the unsaved marker");
```

It should be "a *successful* save clears the unsaved marker", driven through the event rather than the keypress — plus a companion asserting the flag survives a failure.

**3. In-process mode leaves the panel on "Saving…" forever.**

`spawn_save_memory_config` is entirely inside `if let BackendRef::Daemon { .. } = backend`, with no `else`. With no daemon, nothing is spawned, no `MemoryConfigSaved` ever arrives, and the status line keeps saying "Saving..." with no way to clear it. Send a failure event (or a dedicated "requires daemon mode" message) from the other arm.

## Non-blocking

- `.map(|r| r.status().is_success()).unwrap_or(false)` collapses a connection error, a 400 and a 500 into one bare `false`. The panel has a status line that could carry the status code and body — worth passing the reason through, since "Save failed" with no detail is the report an operator will file an issue about.
- Esc clearing `config_dirty` and dropping edits is deliberate and the hint string says so, so no objection — but a second Esc-to-confirm when dirty would cost little on a form where the only other exit is `s`.
- No changelog fragment (`changelog.d/added/7985-....md`). Editable memory config in the TUI is user-facing.

houko:

No commits of your own since my review on 2026-09-01T13:59:26Z; the only thing that moved the branch tip is a merge from `main`, which the `auto-update-branches` cron pushes by itself. The PR's "updated" timestamp has therefore advanced while the blocking items have not been touched. Worth flagging because that timestamp is what makes a PR look answered from outside — I nearly mis-triaged a dozen PRs on exactly that signal today.

Still open from that review:

- 1. The success message says "on"
- 2. `config_dirty` is cleared on keypress, before the save is known to have worked
- 3. In-process mode leaves the panel on "Saving…" forever

DaBlitzStein:
Completing the verification for `b847e99c1`, which was pushed with only the compile check. Run twice independently, with matching results.

```
cargo nextest run -p librefang-cli -E 'kind(lib) | kind(bin)'
Summary [4.940s] 505 tests run: 505 passed, 0 skipped     EXIT=0
```

The seven tests this PR adds, all in `tui::screens::memory::tests`, looked up one by one by exact name rather than inferred from the total:

`abandoning_a_typed_model_restores_what_was_there`, `accepting_a_typed_model_stops_it_being_inherited`, `moving_between_rows_edits_the_row_it_shows`, `saving_sends_exactly_what_the_panel_shows`, `saving_without_data_does_nothing`, `toggling_a_row_marks_the_panel_unsaved`, `typing_a_model_name_does_not_trigger_shortcuts`.

## The part that matters is `main`'s side, not this branch's

This PR touches `tui/event.rs` and `tui/mod.rs`, which `main` had just rewritten to add the model-parameters screen. A merge that quietly dropped one side would still leave this branch's own tests green, so they prove very little on their own.

**All eleven `tui::screens::model_params::tests` — `main`'s screen, in the same files — pass on the merged tree**, including `only_edited_fields_are_sent_and_a_cleared_field_sends_null`, `stepping_walks_the_ladder_and_returns_to_inherit_below_it` and `an_unknown_cap_leaves_the_whole_ladder_available`. **197 tests under `tui::` in total, all green.** Both sides exercise the shared TUI state and neither has clobbered the other.

Runtime side, since the PR also touches `proactive_memory.rs`:

```
cargo nextest run -p librefang-runtime -E 'test(proactive_memory)'
Summary [0.309s] 67 tests run: 67 passed, 2618 skipped
```

Durations of 10-34 ms at load 1.15. Several suites in this repo were producing 31-35 s failures with `Memory init failed: … database is locked` earlier under heavy parallel load; nothing resembling that here, and at this load a red would have been credible rather than dismissible.

`cargo clean -p librefang-types -p librefang-runtime` was run first: manually pruning a shared target leaves cargo's fingerprints inconsistent, and it then reuses another worktree's `rmeta` — which produced compile errors naming fields from unrelated branches three times today.

Also worth recording: `kind(lib)` alone returns zero tests here. `librefang-cli` is a binary crate with no library target, so its unit tests are `kind(bin)`; `kind(lib) | kind(bin)` is what CI runs and what the repo's own guidance specifies.

crates/librefang-cli/src/tui/screens/memory.rs
```diff
@@ -129,16 +175,100 @@ impl MemoryState {
...
+                        self.config_editing_model = true;
+                    }
+                }
+            },
+            KeyCode::Char('s') => {
+                if let Some(cfg) = &self.config {
+                    let action = MemoryUIAction::SaveConfig {
+                        auto_memorize: cfg.auto_memorize,
+                        auto_retrieve: cfg.auto_retrieve,
+                        extraction_model: cfg.effective_extraction_model.clone(),
```
houko: [P2] Preserve the configured extraction model when saving boolean edits

Saving always sends `cfg.effective_extraction_model` as the writable `extraction_model`, even when the operator only toggled auto-memorize or auto-retrieve.
Those are different values: `memory_config_get` returns the boot-resolved model name without its provider prefix, while the raw setting can be `provider/model`, absent to inherit, or a newer value awaiting restart.
`spawn_fetch_memory_config` keeps only the effective name, and `spawn_save_memory_config` unconditionally writes it back.

For example, `extraction_model = "groq/llama-3.3-70b-versatile"` is returned as effective model `llama-3.3-70b-versatile`; toggling a boolean and saving replaces the provider-qualified setting with that bare name.
An inherited model also becomes explicitly pinned, and a pending model change can be overwritten with the old boot value.
Keep the raw configured value separately from the display value and omit `extraction_model` from the PATCH unless the operator edited it.
Add a save-path test showing that a boolean-only edit preserves a provider-qualified or inherited setting.

houko (changes requested):
One new finding from the current-head review. Existing discussion and reviews were checked to avoid duplicating earlier feedback.

DaBlitzStein:
The P2 is addressed at `b4b20c3fc`.

`MemoryConfig` carries `configured_extraction_model` — the raw setting, verbatim — alongside the effective one, and the save omits `extraction_model` entirely unless the operator actually edited it. A `config_model_edited` flag gates that, and `apply_config` clears it on every refetch so a previous edit cannot be resent after the panel reloads.

Worth noting for the record: **the API was already returning both values.** `memory_config_get` sends `extraction_model` (what the file says) and `effective_extraction_model` (what is running), with a comment distinguishing them, plus `extraction_model_source`. The TUI was discarding the raw one and echoing the effective one back. So this is a client-side fix with no API change and no OpenAPI regeneration.

Four tests, and reverting to the unconditional write fails three of them — one per damage you named:

```
a_boolean_only_save_does_not_pin_an_inherited_model
  left: Some("inherited-fast")            right: None
a_boolean_only_save_leaves_a_provider_qualified_model_alone
  an untouched model must be left out of the save, not rewritten
  left: Some("llama-3.3-70b-versatile")   right: None
a_refetch_stops_a_previous_edit_from_being_saved_again ... FAILED
```

The second shows the provider prefix being lost — `groq/llama-3.3-70b-versatile` collapsing to the bare name — and the first shows an inherited setting becoming explicitly pinned. `an_edited_model_is_saved_as_typed` covers the case where the operator did edit it, so the omission is not just "never send it".

18 tests green in that module, i18n guards too.

Ready for another look.

crates/librefang-cli/src/tui/event.rs
```diff
@@ -2507,13 +2514,84 @@ pub fn spawn_fetch_memory_config(backend: BackendRef, tx: mpsc::Sender<AppEvent>
...
+        }
+        let body = serde_json::json!({ "proactive_memory": proactive_memory });
+        let result = match client
+            .patch(format!("{base_url}/api/memory/config"))
+            .json(&body)
+            .send()
+        {
+            Ok(resp) => {
+                let status = resp.status();
+                if status.is_success() {
```
houko, line 2602: **`status.is_success()` is not the success signal this endpoint documents.**

`memory_config_patch` ends with `(StatusCode::OK, Json(body))` on every path that got as far as writing the file, and its own contract comment says clients **MUST** inspect `body.status`, not the HTTP status. `"partial"` means the TOML was written but `reload_config()` failed, with the validator output in `reload_error` and `restart_required: true`.

Scenario: an operator has an unrelated invalid section hand-edited into `config.toml`. `s` returns 200 with `status: "partial"`, so `apply_save_result(Ok(()))` clears the unsaved marker and prints "Saved" — while the running kernel is still on the boot snapshot and nothing on the panel says a restart is needed. The same reads as a clean save whenever `restart_required` comes back `true`.

Parse the body and treat anything other than `status == "applied"` (or a true `restart_required`) as a distinct outcome rather than silent success.

crates/librefang-cli/src/tui/screens/memory.rs
```diff
@@ -129,16 +217,115 @@ impl MemoryState {
...
+                KeyCode::Esc => {
+                    self.config_editing_model = false;
+                    if let Some(cfg) = &self.config {
+                        self.config_model_buf = seed_model_buf(cfg);
+                    }
+                }
+                KeyCode::Enter => {
+                    self.config_editing_model = false;
+                    if let Some(cfg) = &mut self.config {
+                        cfg.effective_extraction_model = self.config_model_buf.clone();
```
houko: **Overwriting `effective_extraction_model` with the typed setting makes the panel report a model that is not extracting anything.**

`MemoryConfigView::effective_extraction_model` is documented as "the model extraction actually runs on", and the kernel is explicit that this is a boot-time snapshot: `HotAction::UpdateProactiveMemory` pushes the new `[proactive_memory]` table onto the existing store but **does not rebuild the extraction driver** (`crates/librefang-kernel/src/kernel/subsystems/memory.rs`, `MemoryExtractionResolution` doc comment; `config_reload_ops.rs:284`). `GET /api/memory/config` splits `extraction_model` from `effective_extraction_model` for exactly this reason.

Scenario: extraction is running on an inherited slow model. The operator types `groq/llama-3.1-8b`, presses Enter, then `s`. The panel drops the "inherited" tag and reports `groq/llama-3.1-8b` as the effective model, but every turn keeps extracting on the old one until the daemon restarts. There is no refetch after a successful save, so the panel never corrects itself — and this panel is the only surface an operator has for that question, which is the whole premise of the PR.

Keep the resolved value on `effective_extraction_model` (render the draft/pending setting as its own span, or blank the effective one), and refetch after a save so the panel shows what the daemon reports rather than what was typed.

crates/librefang-cli/src/tui/screens/memory.rs
```diff
@@ -129,16 +217,115 @@ impl MemoryState {
...
+    /// Fold the outcome of a config save back into the panel.
+    ///
+    /// The unsaved-changes marker is cleared here and nowhere else: it is the
+    /// only thing telling the operator their edits are still pending, so
+    /// dropping it on the keypress turns a failed PATCH into "everything is
+    /// saved" with a transient error beside it.
+    pub fn apply_save_result(&mut self, result: Result<(), crate::tui::event::FetchFailure>) {
+        self.status_msg = match result {
+            Ok(()) => {
+                self.config_dirty = false;
```
houko: **Clearing `config_dirty` / `config_model_edited` unconditionally drops edits made while the PATCH was in flight.**

`s` spawns the request and returns; the panel stays interactive. Scenario: press `s`, then — before the response lands (up to the 5 s client timeout) — press `space` on `auto_retrieve`, or accept a new model in the editor. Both set the flags again. The success event then clears them and prints "Saved". The second edit was never sent, a later `s` will not send the model (`config_model_edited` is back to `false`), and no marker tells the operator anything is still pending: the panel shows the new value and claims it is saved.

Snapshot the values that went out with the request and only clear the flags when the panel still matches that snapshot.

crates/librefang-cli/src/tui/event.rs
```diff
@@ -2507,13 +2514,84 @@ pub fn spawn_fetch_memory_config(backend: BackendRef, tx: mpsc::Sender<AppEvent>
...
+        };
+        let client = make_daemon_client(api_key.as_deref());
+        let mut proactive_memory = serde_json::json!({
+            "auto_memorize": auto_memorize,
+            "auto_retrieve": auto_retrieve,
+        });
+        if let Some(model) = extraction_model.as_deref().map(str::trim) {
+            // An emptied field is not a request to configure the empty string,
+            // and the endpoint has no "unset" for this key — leave it alone.
+            if !model.is_empty() {
```
houko, line 2590: **An emptied model field silently no-ops while the panel reports success.**

`Enter` on an empty buffer is accepted upstream (`memory.rs`: `configured_extraction_model = Some("")`, `effective_extraction_model = ""`, `config_model_edited = true`). This branch then drops the key, the PATCH carries booleans only, the endpoint returns 200, and `apply_save_result(Ok(()))` prints "Saved" and clears the marker.

Scenario: the operator backspaces the model to empty intending to go back to inheriting the default, presses Enter and `s`. The panel afterwards shows a blank extraction model with no "inherited" tag and says "Saved", while `config.toml` still holds the old `extraction_model`. Nothing on screen distinguishes that from a successful clear.

Reject the empty buffer where it is accepted (keep the editor open, or restore the previous value) rather than accepting a draft this function cannot express.

crates/librefang-cli/src/tui/screens/memory.rs
```diff
@@ -129,16 +217,115 @@ impl MemoryState {
...
+                    }
+                }
+                ConfigField::ExtractionModel => {
+                    if let Some(cfg) = &self.config {
+                        self.config_model_buf = seed_model_buf(cfg);
+                        self.config_editing_model = true;
+                    }
+                }
+            },
+            KeyCode::Char('s') => {
```
houko, line 341: **`s` PATCHes even when nothing was edited, and a PATCH rewrites the whole config file.**

There is no `config_dirty` guard here. `memory_config_patch` does a read-modify-write through `toml::to_string_pretty(&table)`, so it serialises the entire document from the parsed value — every comment in the operator's `~/.librefang/config.toml` is dropped, and a full `reload_config()` runs.

Scenario: `c` opens the panel, a stray `s` fires a save of the values just fetched. Nothing changed semantically, but the config file comes back reformatted with all comments stripped. Guard on `self.config_dirty` and make the no-change case a no-op.

crates/librefang-cli/src/tui/screens/memory.rs
```diff
@@ -109,13 +180,30 @@ impl MemoryState {
...
+    /// Put a freshly fetched configuration on the panel.
+    ///
+    /// Clearing the edited flag is the point of having a method: the fetch
+    /// answers with what the file now says, so any draft the operator accepted
+    /// before it is stale, and a save carrying it would write back a value the
+    /// panel is no longer showing.
+    pub fn apply_config(&mut self, config: MemoryConfigView) {
+        self.config = Some(config);
+        self.loading = false;
+        self.config_model_edited = false;
```
houko, line 257: **`apply_config` clears `config_model_edited` but leaves `config_dirty` set, so a refresh leaves a permanent false "unsaved" banner.**

Scenario: toggle `auto_memorize` (dirty), press `r`. The refetch replaces `self.config` with the daemon's values — discarding the toggle — but `config_dirty` stays `true`, so the yellow "Unsaved changes — press s to save, Esc to discard" line stays up over a panel that matches the server exactly. Pressing `s` then PATCHes values nobody edited (see the separate note on the missing dirty guard).

Clear `config_dirty` here too: a fresh fetch is by definition the saved state.

crates/librefang-cli/src/tui/screens/memory.rs
```diff
@@ -370,16 +583,33 @@ fn draw_config(f: &mut Frame, area: Rect, state: &MemoryState) {
...
     }
 
+    if state.config_dirty {
+        lines.push(Line::from(""));
+        lines.push(Line::from(Span::styled(
+            crate::i18n::t("tui-memory-config-unsaved"),
+            Style::default().fg(theme::YELLOW),
+        )));
+    }
+    if !state.status_msg.is_empty() {
```
houko: **`status_msg` is shared with the KV browser footer, so config-save messages leak across sub-screens.**

`draw_kv_browser` renders the same `state.status_msg`. Scenario: press `s` in the config panel, `Esc`, then `Enter` into an agent — the KV list footer reads "Saving..." / "Saved". In the other direction, a stale `tui-mod-saved-key` message from the KV screen renders inside the config panel where it reads as the result of a config save.

A dedicated field for the config panel keeps each screen's status honest.

crates/librefang-cli/src/tui/event.rs
```diff
@@ -2507,13 +2514,84 @@ pub fn spawn_fetch_memory_config(backend: BackendRef, tx: mpsc::Sender<AppEvent>
...
+        let BackendRef::Daemon { base_url, api_key } = backend else {
+            // Without this the panel sits on "Saving..." forever: nothing is
+            // spawned, no event arrives, and the operator has no way to tell
+            // the save is never going to happen.
+            let _ = tx.send(AppEvent::MemoryConfigSaved(Err(
+                FetchFailure::RequiresDaemon,
+            )));
+            return;
+        };
+        let client = make_daemon_client(api_key.as_deref());
```
houko: **The default 5 s client timeout is a read timeout on a write that does far more than 5 s of work.**

`make_daemon_client` is the 5 s variant. This handler reads `config.toml`, serialises and atomically rewrites it, then runs a full `reload_config()` — which can rebuild embedding/extraction drivers and re-scan skills.

Scenario: the reload takes longer than 5 s. The TUI reports "Save failed: operation timed out" and keeps the unsaved marker, although the file was written and the reload completed server-side. The operator retries and PATCHes a second time. `make_daemon_client_with_timeout` already exists — give this write a budget sized for the reload.

crates/librefang-runtime/src/proactive_memory.rs
```diff
@@ -31,6 +31,21 @@ impl librefang_memory::proactive::EmbeddingFn for EmbeddingBridge {
     }
 }
 
+/// Wall-clock ceiling on the single extraction call that turns a conversation
+/// slice into candidate memories.
+///
+/// Named rather than inlined because the caller in `agent_loop::end_turn` has
+/// to budget for it: a ceiling here that the end-of-turn budget does not cover
+/// would cut off extractions that were about to succeed.
+pub const EXTRACTION_TIMEOUT_SECS: u64 = 30;
```
houko, line 41: **This rationale does not describe the code.** `git grep EXTRACTION_TIMEOUT_SECS` / `DECISION_TIMEOUT_SECS` matches only this file, and `crates/librefang-runtime/src/agent_loop/end_turn.rs` has no wall-clock budget around the `auto_memorize` call at all — there is nothing budgeting for this ceiling, so "named rather than inlined because the caller in `agent_loop::end_turn` has to budget for it" claims a coupling that does not exist.

Either wire the constant into an actual end-of-turn budget or state the real reason (two related ceilings kept side by side). Worth noting this hunk is also unrelated to the TUI change the rest of the PR makes.

crates/librefang-cli/src/tui/screens/memory.rs
```diff
@@ -129,16 +217,115 @@ impl MemoryState {
...
+                            .flatten(),
+                    };
+                    // `config_dirty` stays set until the daemon confirms the
+                    // write — see `apply_save_result`.
+                    self.status_msg = crate::i18n::t("tui-memory-config-saving");
+                    return action;
+                }
+            }
+            KeyCode::Char('r') => return MemoryUIAction::LoadConfig,
             KeyCode::Esc | KeyCode::Char('q') => {
```
houko, line 378: **The hint line was not updated for the new bindings.** `tui-memory-config-hint` is still `r refresh · Esc back` in all four locales, so `j`/`k`, `space` and `s` are undiscoverable until something is already dirty (the unsaved banner is the only place `s` is mentioned, and it only appears after an edit).

Also worth reconsidering: `Esc` here clears `config_dirty` and returns without confirmation, and the mutated `self.config` is only corrected because re-entry via `c` happens to refetch.

crates/librefang-cli/src/tui/screens/memory.rs
DaBlitzStein: Confirmed, and fixed in `a07b3540b`.

`MemoryConfigView` now carries the raw setting and the resolved name as two fields (`configured_extraction_model` / `effective_extraction_model`, `crates/librefang-cli/src/tui/screens/memory.rs:40-47`), the fetch keeps both (`event.rs:2521`), and the save omits `extraction_model` entirely unless the operator opened the editor and accepted a value (`memory.rs:348-353`). The editor also seeds from the configured setting rather than the split half, so editing `groq/llama-3.3-70b-versatile` starts from the whole string (`seed_model_buf`, `memory.rs:56`).

Two save-path tests, as asked. Both fail against the pre-fix behaviour — reverting the save to send `cfg.effective_extraction_model` unconditionally:

```
---- a_boolean_only_save_leaves_a_provider_qualified_model_alone stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1262:17:
assertion `left == right` failed: an untouched model must be left out of the save, not rewritten
  left: Some("llama-3.3-70b-versatile")
 right: None

---- a_boolean_only_save_does_not_pin_an_inherited_model stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1286:18:
  left: Some("inherited-fast")
 right: None
```

The inherited case is the second half of your example and it does reproduce: an unset `extraction_model` came back as the boot-resolved `[default_model]` name and the save pinned it.

crates/librefang-cli/src/tui/event.rs
DaBlitzStein: Confirmed — the endpoint's own contract comment says exactly this, and the client was reading only the HTTP status. Fixed in `22fd993aa`, and this push (`2a02fc044`) makes it testable.

I verified the contract against the handler rather than the description: `memory_config_patch` returns `(StatusCode::OK, Json(body))` on every path that reached disk, and `"partial"` carries `reload_error` with `restart_required: true` (`crates/librefang-api/src/routes/memory.rs:1890-1900`).

The body handling now lives in `interpret_memory_config_patch_body` (`crates/librefang-cli/src/tui/event.rs:2541`) instead of inline in the request thread, so the discrimination can be tested without standing up an HTTP server. Anything that is not `status == "applied"` becomes a `FetchFailure::Error` carrying the validator output, which the panel renders as a failure and which leaves the unsaved marker up.

Reverting that function to `Ok(())`:

```
---- tui::event::tests::a_partial_memory_config_patch_is_not_a_success stdout ----
panicked at crates/librefang-cli/src/tui/event.rs:5224:22:
a partial save must not read as success, got Ok(())
```

One thing I did not do: `restart_required: true` on an otherwise `"applied"` save is still treated as success. The extraction model is in that state after *every* save — a config reload swaps the `[proactive_memory]` table onto the running store without rebuilding the extraction driver — so failing the save on it would mark every model change as failed. The panel reports it on the model row instead (see the thread on `effective_extraction_model`). Say the word if you would rather have it as a distinct third outcome on the status line.

crates/librefang-cli/src/tui/screens/memory.rs
DaBlitzStein: Confirmed on both counts, and both are now fixed — the first half in `a07b3540b`, the refetch and the display consequence in `2a02fc044`.

Accepting a model no longer touches `effective_extraction_model`; it writes only `configured_extraction_model` (`crates/librefang-cli/src/tui/screens/memory.rs:288-300`), so the resolved value keeps meaning "what extraction actually runs on".

Your second point turned out to matter more than the first, because fixing only the first left the panel lying in a new way. The pending span was gated on `config_model_edited`, which a successful save clears — so the moment the save landed the row went back to showing only the old model, with no sign of the value just written. The operator was told "Saved" and shown what looks like a revert.

So the indicator now follows whether the setting is actually what is running, not whether this session typed it (`memory.rs:698-709`), and names the two states apart: `pending save` before the write, `saved — restart to use it` after. That second one is also now visible on a plain fetch, which it never was — a hand-edited `config.toml` that the daemon has not restarted into reads as diverged the moment the panel opens.

The comparison is deliberately not a second provider split. `MemoryExtractionTarget`'s docs say boot accepts both `provider/model` and `provider:model` and strips the prefix with `strip_provider_prefix`; re-deriving that here is precisely the "two derivations disagree" failure your `GET` comment warns about. `extraction_model_awaits_restart` (`memory.rs:78`) does a suffix test instead, and returns `false` when `effective_extraction_model` is empty — extraction inactive, a sidecar, or a failed driver build — since none of those is a pending restart.

And the refetch: a clean save now returns `MemoryUIAction::LoadConfig` (`memory.rs:435`, dispatched at `mod.rs:628-635`).

Red cycle, `extraction_model_awaits_restart` stubbed to `false`:

```
---- a_saved_model_the_daemon_has_not_picked_up_is_reported_as_awaiting_a_restart stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1498:9:
the file names one model and the extractor is running another

---- a_clean_save_refetches_what_the_daemon_now_reports stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1547:9:
a save moves the setting without moving the extractor; only a fetch can say what each one is now
```

And with the comparison reduced to a naive `configured != effective`, which is the version that looks right and is not:

```
---- a_provider_qualified_setting_that_is_already_running_needs_no_restart stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1518:13:
groq/llama-3.3-70b-versatile is the spec boot resolved to the running model, not a pending change

---- a_setting_with_no_running_model_is_not_a_pending_restart stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1537:9:
assertion failed: !extraction_model_awaits_restart(&cfg)
```

Coverage boundary, stated plainly: those tests pin the predicate, not the rendering. The render is a three-line `if/else if` over it (`memory.rs:698-709`) and is not under test — the draw path needs a `Frame`, and there is no render-test harness on this screen to extend.

crates/librefang-cli/src/tui/screens/memory.rs
DaBlitzStein: Confirmed, fixed in `a07b3540b`.

`s` snapshots what the request carries into `pending_save` (`crates/librefang-cli/src/tui/screens/memory.rs:181`, set at `:366-370`), and `apply_save_result` clears the markers only when the panel still matches that snapshot (`:415-423`). An edit that lands in the window keeps `config_dirty` and `config_model_edited` set, so a later `s` still sends it.

The refetch added in this push (`2a02fc044`) had to respect the same rule, or it would have become a second way to lose that edit — `apply_config` takes the daemon's answer as truth. It is skipped on exactly the path where the snapshot no longer matches.

Reverting the snapshot comparison to an unconditional clear:

```
---- an_edit_made_while_a_save_is_in_flight_is_not_marked_saved stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1450:9:
the second toggle arrived after the snapshot the in-flight save carried

---- a_save_that_raced_an_edit_does_not_refetch_over_it stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1563:9:
refetching here would overwrite the edit that arrived after the snapshot
```

crates/librefang-cli/src/tui/event.rs
DaBlitzStein: Confirmed, fixed in `a07b3540b` — and fixed where you said, at the point the draft is accepted rather than downstream.

`Enter` on an empty or whitespace-only buffer keeps the editor open and returns without accepting anything (`crates/librefang-cli/src/tui/screens/memory.rs:288-295`). The endpoint has no unset for this key — `memory_config_patch` only inserts `extraction_model` when the request carries a string (`crates/librefang-api/src/routes/memory.rs:1851`), and unlike the `[memory]` keys it does not route through `apply_optional_string_patch`, so a `null` would not remove it either. An empty draft is not a setting this panel can express, so it is refused where it is typed.

The save path keeps a belt-and-braces skip for an empty model (`event.rs:2587-2592`), but it is no longer reachable from the editor.

Reverting the rejection:

```
---- emptying_the_model_buffer_and_pressing_enter_keeps_the_editor_open stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1193:9:
an empty draft is not a setting a save can express; the editor must stay open
```

crates/librefang-cli/src/tui/screens/memory.rs
DaBlitzStein: Confirmed, fixed in `a07b3540b`.

`s` returns early when nothing is dirty (`crates/librefang-cli/src/tui/screens/memory.rs:345-347`), so the stray-keypress case is a no-op and never reaches the endpoint.

The reason is worth keeping on the record because it is not obvious from the client side: `memory_config_patch` parses the file into a `toml::Value` and re-serialises the whole document with `toml::to_string_pretty` (`crates/librefang-api/src/routes/memory.rs:1863`), so every comment in the operator's `config.toml` is dropped by a round trip that changed nothing.

Reverting the guard:

```
---- saving_with_nothing_edited_does_nothing stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1474:9:
the endpoint does a full read-modify-write of config.toml; a no-op save still strips every comment in it
```

crates/librefang-cli/src/tui/screens/memory.rs
DaBlitzStein: Confirmed, fixed in `a07b3540b`: `apply_config` clears `config_dirty` as well as `config_model_edited` (`crates/librefang-cli/src/tui/screens/memory.rs:255-262`). A fresh fetch is the saved state by definition, banner included.

Reverting that line:

```
---- a_refetch_with_nothing_else_edited_leaves_nothing_to_save stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1363:9:
a fresh fetch replaces the draft; the yellow banner must not survive it

---- a_refetch_stops_a_previous_edit_from_being_saved_again stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1334:9:
a fresh fetch is by definition the saved state, discarded draft included
```

crates/librefang-cli/src/tui/screens/memory.rs
DaBlitzStein: Confirmed, fixed in `a07b3540b`: the config panel writes to its own `config_status_msg` (`crates/librefang-cli/src/tui/screens/memory.rs:164`) and renders only that (`:753-758`). `status_msg` stays the KV browser's. The fetch-failure path in `mod.rs:636-649` was moved onto the config field too — that fetch only ever happens for this screen, so its error belongs there.

Both directions of your scenario are closed by the split: nothing in the config panel writes `status_msg`, and nothing in the KV browser writes `config_status_msg`.

No test on this one, and I would rather say so than imply otherwise: asserting it needs a render assertion on two different sub-screens, and the screen has no render-test harness to extend. The guarantee here is structural — the two fields have disjoint writers — rather than pinned by a test.

crates/librefang-cli/src/tui/event.rs
DaBlitzStein: Confirmed, fixed in `22fd993aa`: the PATCH uses `make_daemon_client_with_timeout(.., Duration::from_secs(30))` (`crates/librefang-cli/src/tui/event.rs:2582`) instead of the 5 s `make_daemon_client`.

30 s to match the other write paths in this file that trigger daemon-side work rather than a read. The retry-a-completed-write scenario you describe is the one that matters — a timeout here is indistinguishable to the operator from a rejected save, and the second PATCH is a full read-modify-write of `config.toml` on top of a reload that already succeeded.

No test: this is a client construction inside a thread-spawning function, and pinning it would take an HTTP fixture that deliberately stalls. Verified by reading the call site.

crates/librefang-runtime/src/proactive_memory.rs
DaBlitzStein: Confirmed — the rationale claimed a coupling that does not exist. Fixed in `c8684d4f0`.

I checked rather than taking the description: `git grep` for either constant matches only `crates/librefang-runtime/src/proactive_memory.rs`, they are used solely as `timeout_secs` on the two extraction calls (`:582`, `:707`), and `agent_loop::end_turn` wraps the `auto_memorize` call in no wall-clock budget at all. The comment now states the real reason — the two ceilings sit together so they can be compared, and this one is enforced inside the extraction call — rather than inventing a caller that budgets for it.

On the hunk being unrelated: fair, and I want to be accurate about what it is. `origin/main` has these as bare `Some(30)` / `Some(15)` at the same call sites, so naming them is this PR's own change, not a pre-existing comment I wandered into. It is still not a TUI change. I have left it in because dropping it puts the magic numbers back, but if you would rather this PR were TUI-only I will pull the hunk out and send it separately — your call.

crates/librefang-cli/src/tui/screens/memory.rs
DaBlitzStein: Both confirmed. The hint was fixed in `a07b3540b`, the `Esc` behaviour in this push (`2a02fc044`).

`tui-memory-config-hint` now names every binding, in all four locales with real translations rather than English copied across: `j/k select · space/Enter toggle or edit · s save · r refresh · Esc back` (`crates/librefang-cli/locales/{en,ko,uk,zh-CN}/main.ftl`). The new `tui-memory-config-restart-required` key added in this push is translated in all four as well.

On `Esc`: you were right that the correctness depended on an accident, and I took that as the defect rather than the missing confirmation. `Esc` cleared the marker but left the mutated values on `self.config`, where they read as configuration the daemon had reported — corrected only because re-entry through `c` happens to refetch. That accident fails exactly when it matters: while the refetch is in flight, or when it fails and the panel keeps rendering local edits as the daemon's own state.

So the panel now keeps the last fetched configuration (`config_saved`, `crates/librefang-cli/src/tui/screens/memory.rs:188`) and `Esc` restores it (`:383-390`), which makes the hint's "Esc to discard" true instead of half-true.

```
---- esc_puts_back_the_configuration_the_daemon_reported stdout ----
panicked at crates/librefang-cli/src/tui/screens/memory.rs:1597:9:
the discarded toggle must be back to what the daemon reported
```

I did not add a confirmation prompt. With the values genuinely restored, `Esc` now does what the hint says it does, and a confirmation step on a documented discard felt like the wrong trade against a keypress operators will hit constantly. Happy to add one if you disagree — it is a small change now that the discard path is real.

DaBlitzStein:
Every one of the 11 review threads on this PR now has an inline reply, and the branch is at `2a02fc044`.

houko (approved):
LGTM. All six findings are fixed, and several of them are fixed in a way that makes the class of mistake harder to repeat rather than just patching the instance.

- **`status.is_success()` was the wrong success signal.** Now `interpret_memory_config_patch_body` discriminates on `body.status`, returns the `reload_error` text alongside `tui-memory-config-save-partial`, and — the part that matters most — is split out of the request thread specifically so the discrimination is unit-testable without standing up an HTTP server. The doc-comment restates the endpoint's contract at the call site, so the next person does not have to find it in `memory_config_patch`.
- **The emptied model field silently no-opping.** Guarded with `map(str::trim)` + `!model.is_empty()`, and the doc on `spawn_save_memory_config` now explains why `None` (not-edited) and an empty edit are different cases and what sending the displayed name instead would break — stripping a `provider/` prefix, pinning a model that was inheriting, or overwriting a change still waiting on a restart.
- **`s` PATCHing when nothing was edited.** `if !self.config_dirty` at `memory.rs:345` short-circuits, so a full read-modify-write that drops every comment in the operator's `config.toml` no longer happens on an idle keypress.
- **`apply_config` leaving a permanent false "unsaved" banner.** `config_dirty = false` with the reasoning inline ("a fresh fetch is by definition the saved state").
- **The `proactive_memory` rationale that described nothing.** Rewritten honestly: it now says the caller does *not* wrap `auto_memorize` in a timeout and that the ceiling is enforced inside the extraction call, instead of claiming a caller-side budget that `end_turn.rs` does not have. `DECISION_TIMEOUT_SECS` gets a real reason for being shorter rather than an assertion.
- **The stale hint line.** `tui-memory-config-hint` now reads `j/k select · space/Enter toggle or edit · s save · r refresh · Esc back`, so the bindings are discoverable before something is dirty.

All checks green.

Issue #6553, opened 2026-07-22
`agent kill --purge_identity` (aka `agent kill <UUID>`, canonical-UUID purge) cascades DELETE into `audit_entries`, breaking the WORM Merkle chain that `security verify` guarantees


otaviobrum:
# Description

`security verify` walks `audit_entries` from the lowest existing `seq` upward, recomputing each row's hash from its content and checking `row[N].prev_hash == row[N-1].hash`. That is the entire tamper-evidence guarantee of the audit trail — and by design the trail is documented/expected to be append-only (WORM): nothing should ever remove a row once written.

We found that it isn't. `librefang agent kill <AGENT_ID>` (the CLI's `--help` text: *"the agent's canonical UUID binding is purged, orphaning any prior sessions / memories under the old UUID"*) performs a cascading DELETE that **also removes the `audit_entries` rows belonging to the purged agent** — including the killed agent's own `AgentKill` audit row's predecessor(s) and any other rows tied to that agent's identity. This is not documented as touching `audit_entries` at all; the help text only mentions sessions/memories.

The result: every time a test/probe agent is purged, `audit_entries` develops a gap. Downstream, `prev_hash` on the row *after* the gap points to the hash of a row that no longer exists, so `security verify` reports a **chain break** — indistinguishable, without out-of-band investigation, from an attacker having tampered with the trail. For a product whose main audit-trail selling point is tamper-evidence, "the tool itself silently breaks its own tamper-evidence as a side effect of a routine, non-destructive-sounding admin command" is a serious integrity defect.

**How we found it and confirmed it's the purge, not corruption:**

We run `security verify` periodically as part of our own audit playbook. It started failing with `chain break at seq 247`. Investigating (isolated copy of the DB, production untouched — see Steps below), we found **three separate gaps**, each immediately preceding an `AgentKill ... purge_identity=true` row for a short-lived test/probe agent:

| gap (missing `seq`) | date (UTC) | next surviving row |
|---|---|---|
| 246 | 2026-07-10T01:13Z | seq 247, `AgentKill name=soc-l1-analyst-wprobe, purge_identity=true` |
| 655–658, 664 | 2026-07-11T19:18Z | seq 665, `AgentKill name=revert-probe, purge_identity=true` |
| 1678–1686 | 2026-07-18T13:44Z | seq 1687, `AgentKill name=gate-probe-tmp, purge_identity=true` |

[...]

PR #7986, opened 2026-08-29
feat(cli): add purge command to remove every trace of an agent

DaBlitzStein:
## Summary

`librefang agent purge --agent <name>` removes every trace of an agent in one idempotent operation: the roster entry (cascading to sessions, memories and KV rows), the workspace directory, the agent-type template, the identity record, and the cron jobs, triggers and channel bindings that would otherwise be inherited by the next agent registered under the same name.
The CLI opens the database directly, so it works with no daemon running, and a path-shaped agent name is rejected before any filesystem operation happens.
`agent_purge::purge_agent()` lives in `librefang-kernel` so the API and the TUI can reuse the same plan-then-execute pair rather than growing a second implementation.

## Substantive changes

- The configuration that names the purge target is resolved through `try_load_config`, the strict loader, and a failure aborts with a nonzero status before any substrate is opened.
  `load_config` answers a bad or missing file with `Ok(defaults)`, so `--config /srv/instance-b/config.toml` against an unreadable file used to move the deletion onto the default installation and purge its unrelated agent of the same name.
  The one tolerated absence is a config file missing from the *default* location when none was requested by path.
- The database is resolved from the configuration (`memory.sqlite_path`, else `data_dir`), matching what `cmd_audit_reset` already does, instead of a hardcoded `home/data/librefang.db` that missed the live database whenever `data_dir` had been moved.
- The plan the operator confirms now opens with the selected database path, so a purge aimed at the wrong installation is visible before the prompt rather than after.
- The destructive path refuses to run while a daemon is up, unless `--force` is passed; `--dry-run` is unaffected.
- Cron jobs, triggers and channel bindings belonging to the agent are removed.
  This is the residue that matters most in practice: cron and trigger entries key on `AgentId`, and an id derived from the agent name is deterministic, so the next agent registered under that name silently inherits the previous one's schedule.
  Channel instance defaults and conversation bindings key on the name itself and inherit the same way.
  The removal reuses the existing `CronScheduler::remove_agent_jobs` / `TriggerEngine::remove_agent_triggers`, which `kill_agent` already uses on the live path — the purge simply never called them because it does not boot a kernel.
- A workspace override is no longer able to resolve onto a directory that is not the agent's own: `workspace = "."` resolved to the workspaces root itself, and the deletion was attempted against it.
  A workspace shared with another live agent is likewise reported rather than deleted.
- A hand agent's workspace override, which `backfill_workspace_dir` writes as an absolute path under the hands subtree, is now validated against the whole workspaces root instead of the `agents/` subtree, so it resolves instead of being reported missing.
- The orphan scan runs even when the name is live again, so rows left by an earlier incarnation under a different id are found rather than skipped.
- The live-agent guard reads raw ids via `list_agents` instead of `load_all_agents`, which silently drops any row whose manifest blob fails to deserialize.
  A live agent with a corrupt manifest was therefore indistinguishable from an orphan, and its rows were eligible for cascade.
- An id with rows in agent-scoped tables that cannot be attributed to any name is **reported and never touched** (`other_orphans_present`).
  No agent-scoped table stores a name, so an id this command cannot derive from the requested name may belong to an entirely different agent; deleting it would be guessing.
- A failure to persist `agent_identities.toml` is surfaced instead of being reported as success — `AgentIdentityRegistry::purge` only logs the error and returns as if the write had happened.
- `--dry-run` documents honestly that opening the database still runs migrations and the FTS reconcile.
  A read-only open was considered and rejected: there is no precedent for one in this codebase and it would hard-fail on a database that has not been migrated yet, which is a worse failure mode than the one being documented.

## Verification

- `cargo test -p librefang-kernel -- agent_purge` — 22 tests, including `a_workspace_override_pointing_at_the_root_itself_is_never_deleted`, `a_workspace_shared_with_another_live_agent_is_never_deleted`, `cron_and_trigger_residue_is_removed_so_a_recreated_agent_does_not_inherit_it`, `channel_binding_residue_is_removed_by_name`, `stale_orphans_from_an_earlier_incarnation_are_found_even_when_the_name_is_live_again`, `a_live_agent_with_a_corrupt_manifest_is_never_treated_as_an_orphan`, `unattributable_orphan_rows_are_reported_but_never_touched`, `a_hand_agents_absolute_workspace_override_is_resolved_not_reported_missing` and `identity_persist_failure_is_reported_not_silently_swallowed`.
- `cargo test -p librefang-cli --bins -- purge` — 11 tests, including `an_unloadable_explicit_config_purges_nothing_from_the_default_installation`, `a_missing_explicit_config_purges_nothing_from_the_default_installation`, `a_missing_default_config_still_purges_the_default_installation`, `db_path_follows_configured_data_dir_not_a_hardcoded_home_join` and the four cases around the daemon-running guard.
- Every one of these was confirmed red before it was trusted: the production block was reverted, the test was run, the literal failure was recorded, and the block was restored.
  The one that mattered most produced `remove workspace /tmp/.../workspaces/agents/.` in the failure output, which is the unfixed code actually attempting to delete the workspaces root.
- `cargo clippy -p librefang-kernel -p librefang-cli --all-targets -- -D warnings` — clean, run again after `rustfmt`.
- `i18n_checks` green, with real translations for the six new keys in `uk` and `zh-CN` rather than the English text copied across.

## Left out

Nothing deferred.
The cron, trigger and channel-binding removal was in scope by the time it was found, since the deletion primitives already existed and only the wiring was missing.

## Fixed during review (round 2, `870a38e6d`)

An internal review of `bcbe5bc18` found eight issues.
None had a maintainer thread, so they are recorded here rather than answered inline.
All eight are addressed; each new or changed test was confirmed red against the reverted production block before being trusted.

- **The shared-workspace guard compared resolved directories for equality, so a *nested* workspace of another live agent was deleted anyway.**
  `alpha` with `workspace = "shared"` and a live, registered `beta` with `workspace = "shared/team"` need no exotic configuration, and `remove_dir_all(<workspaces>/agents/shared)` took beta's subtree with it while the report called it a clean removal.
  The guard now tests containment in both directions.
  Its input also widened from `manifest.workspace` to every directory an agent occupies: the `[workspaces]` table is the second way an agent gets a directory, `ensure_named_workspaces` resolves a `path` entry against the same root, and nothing forbids one under `agents/`.
  Declarations resolving outside the workspaces root are ignored, since the directory being purged is always inside it — otherwise a single `mount = "/"` would veto every workspace deletion in the installation.
  Pinned by `a_nested_workspace_of_another_live_agent_is_never_deleted` and `another_agents_named_workspace_under_this_one_is_never_deleted`.
- **A purely informational scan could veto the entire command.**
  `other_orphan_ids_exist` deletes nothing and no other step's safety depends on its answer, but its error went into `plan.failures`, which `purge_agent` refuses to execute on — so a diagnostic failure blocked the cleanup with no flag to get past it.
  It now degrades to "could not determine" with a `WARN`.
  The query also asks SQLite for existence instead of materialising every distinct id of every agent-scoped table into a `Vec<String>` and filtering in Rust, which additionally removes the one place a non-TEXT value in an agent-scoping column could raise `InvalidColumnType`.
  `a_failing_orphan_scan_never_vetoes_the_purge` fails against the old code with `read orphan ids from memories: Invalid column type Blob at index: 0, name: agent_id`.
- **Neither the daemon guard nor the database resolution had a test at its injection site.**
  Both helpers had unit tests that stayed green with their call sites deleted from `cmd_purge`, because the three existing `cmd_purge` tests ran against a default installation where `home_dir/data` and `data_dir` are the same directory.
  `cmd_purge_opens_the_database_data_dir_points_at_not_a_home_join` points `data_dir` at an unrelated tree and exits 1 on `no database at <home>/data/librefang.db` against the old join; `cmd_purge_refuses_while_a_daemon_answers_in_this_home_unless_forced` binds a loopback socket answering `/api/health`, and with the guard's call site removed the command deletes the agent-type template and returns 0.
  Both drive the command through `--config` rather than `LIBREFANG_HOME`, so neither adds to the set of tests mutating process-wide state.
- **A caveat-only plan announced a deletion that never happened.**
  With nothing attributable to the name but `other_orphans_present` set, the operator saw "About to permanently purge 'alpha':" followed by a note, then no prompt, no action and exit 0.
  The caveats now hang under the "nothing to purge" line.
  The report's stdout lines were extracted into `outcome_lines` so what the operator is told is assertable; `a_caveat_only_plan_does_not_announce_a_deletion` compares the whole vector.
- **The printed database path no longer named everything the command deletes.**
  Cron jobs, triggers, agent-type templates and workspaces all hang off `home_dir` while the database is resolved through `sqlite_path` / `data_dir`, which this PR deliberately decoupled — on a relocated install the operator confirmed one path and got deletions in two trees.
  The home directory is printed alongside it (`purge-home-line`, four locales).
- **`ko` lost six keys to this branch.**
  It was at exact parity at `0037a0710` (2309 keys, same as `en`) and this branch left it six behind, which `test_locales_cover_used_i18n_keys` does not catch because it checks `en`, `uk` and `zh-CN` only.
  The six are restored with real Korean, along with the reworded dry-run header this branch changed in the other three; all four locales are back at 2316.
- **`identity_persist_failure_is_reported_not_silently_swallowed` passed for free under root.**
  It strips the home directory to `0o500` and relies on the write failing, which root ignores — and the repo's own no-toolchain path (`Dockerfile.rust-dev`) runs as root.
  `persist_tmp_path` mixes a timestamp into the temporary filename and the destination has to stay readable for the plan, so the sabotage cannot be made permission-independent; instead the test now probes whether the sabotage bites and skips with a printed reason when it does not.
  Confirmed by inverting the probe: the test then reports `skipped: this process writes to a 0o500 directory`, proving the write really fails here and the assertions ran for real.
- **`find_daemon_in_home` was evaluated eagerly on the two paths that discard it.**
  `--dry-run` and `--force` each paid up to 3 s of connect and read timeout for a value the guard drops on its first line; the probe is now a closure, and the two bypass tests fail if it is ever called.

Two items from the same review were deliberately not changed, with the reasoning here rather than in a follow-up:

- **`agent_identities.toml` is still written twice** — once inside `AgentIdentityRegistry::purge`, which swallows the error, and once after, to get a result worth reporting.
  The clean fix is to thread a `Result` out of `purge()`, but `kernel/agent_runtime.rs:829` and `librefang-api/src/routes/agents/identity.rs:218` both depend on the current signature, so it is a cross-crate API change carrying a behaviour fix of its own (that API route returns 200 on a failed persist).
  The current code is correct and errs toward under-reporting; the cost is one extra atomic rename on a manual command.
  Say the word and it becomes its own PR.
- **`cron.persist()` makes the load-time quarantine durable for other agents' jobs.**
  `CronScheduler::load` disables any persisted job that fails re-validation, and persisting writes that `enabled: false` for agents this purge never touched.
  Reporting it would need a new `PurgeReport` field and four locale keys for a state the daemon writes at its next boot anyway, and merging the removals into a fresh read of the file would resurrect a job that crash-loops the scheduler.
  The trade-off is now documented at the call site instead.

### Verification for this round

- `cargo test -p librefang-kernel --lib` — 1767 passed, 0 failed (the 25 `agent_purge` tests included).
- `cargo test -p librefang-cli --bins` — 513 passed, 0 failed.
- `cargo test -p librefang-cli --test i18n_checks` — 3 passed.
- `cargo clippy -p librefang-kernel -p librefang-cli --all-targets -- -D warnings` — clean, re-run after `rustfmt`.
- `cargo check --workspace --lib --exclude librefang-desktop` — clean (this host has no GTK headers).
- Red proofs, each with the literal failure recorded: the containment guard reverted to equality (both nested tests fail), `other_orphan_ids_exist` reverted to the materialising form (`Invalid column type Blob`), `purge_db_path` replaced by the old home join (exit 1), the daemon guard's call site deleted (the template is deleted and the command returns 0), the probe made eager (both bypass tests panic on "the daemon probe ran on a path that discards its answer"), and the caveat header restored (the first line reads "About to permanently purge 'alpha':").


houko (commented):
The purge looks for the agent-type in a directory that is not the agent-type store, and the test agrees with the code rather than with the tree — so it passes while the behaviour is wrong.

`crates/librefang-kernel/src/agent_purge.rs`:

```rust
let type_file = home
    .join("workspaces")
    .join("agent-types")
    .join(format!("{agent_name}.toml"));
```

The canonical store is `<home>/agent-types/`, not `<home>/workspaces/agent-types/`.
`librefang_types::registry_paths::installed_agent_types_dir` is `home_dir.join("agent-types")` — there is a test on it, `installed_agent_types_dir_is_home_join_agent_types`, and #8000 (merged) routes the registry fan-out, the CRUD API, the spawn resolvers and the CLI all through that same helper.

So on a real installation `type_file.is_file()` is false, nothing is removed, and `report.agent_type_removed` stays `false` — the command reports success having left the agent-type behind, which is exactly the lingering-data problem this command exists to solve.

The unit tests do not catch it because `home_with` builds the fixture at the same wrong path:

```rust
let types = dir.path().join("workspaces").join("agent-types");
```

Both sides agree with each other and neither agrees with the tree.
Switching the production path to `librefang_types::registry_paths::installed_agent_types_dir(home)` and the fixture with it fixes both — and using the helper rather than a literal is what keeps this from drifting again the next time the directory moves.

Two other things worth settling before this lands.

**No confirmation on a destructive command.**
`librefang purge --agent foo` goes straight to `remove_dir_all` on the workspace and a cascading `remove_agent` in the database, with no prompt, no `--yes`, and no `--dry-run`.
A typo in `--agent` is unrecoverable, and the command is aimed at operators cleaning up by hand, which is precisely when a name gets mistyped.
A confirmation prompt with a `--yes` escape (or at minimum a `--dry-run` that prints the `PurgeReport` without acting) would match the blast radius.

**The module doc overstates its reach.**
`agent_purge.rs` opens with "shared by the CLI, the API, the dashboard and the TUI", and in this PR only `cmd_purge` calls it.
Either the other three callers belong in this PR, or the doc should say the CLI is the only caller today and the free-function shape is what keeps the others cheap to add.

houko (changes requested):
Good module boundary (a free function so the CLI can run without a daemon), an honest `PurgeReport`, a path-traversal guard with a test, and all four locales translated. But it does not clean the case the module doc says it exists for, it looks for agent types in the wrong directory, and it is a `remove_dir_all` with no confirmation.

## Blocking

**1. The agent-type path is wrong, and the test fixture hides it.**

```rust
let type_file = home.join("workspaces").join("agent-types").join(format!("{agent_name}.toml"));
```

The store is at `home/agent-types/`, not `home/workspaces/agent-types/` — `librefang_types::agent_type_store::agent_types_dir_in` is:

```rust
pub fn agent_types_dir_in(home_dir: &std::path::Path) -> PathBuf {
    home_dir.join("agent-types")
}
```

and its doc says why that helper exists: *"Both spellings route through this one function so a reader and a writer can never disagree about where the store is (#6699)."* The API agrees — `agent_templates.rs:145` documents `~/.librefang/agent-types/{name}.toml`.

So `agent_type_removed` is always `false` against a real installation and the template survives the purge. `it_removes_the_workspace_and_the_agent_type` passes only because `home_with()` writes the fixture to the same wrong path the code reads. Use `agent_types_dir_in(home)` (and `agent_type_path`-style joining) so the test would have caught this.

While you are there: `validate_agent_type_name` lives in that same module. Reusing it instead of the hand-rolled `contains('/') || contains('\\') || contains("..")` check keeps one definition of a safe name.

**2. The headline case — a roster entry that is already gone — is not cleaned.**

The module doc:

> Deleting an agent removes it from the roster … and (for an agent whose roster entry is already gone) its sessions and memories in the database with nothing left pointing at them. **This module is the one place that cleans all of it.**

and the CLI help: *"For agents the operator already deleted but whose data lingers."*

But the database work is entirely inside:

```rust
if let Some(entry) = entries.iter().find(|e| e.name == agent_name) {
    substrate.remove_agent(entry.id)…
}
```

No roster entry means no `entry.id`, so `remove_agent` is never called and the orphaned sessions, memories and KV rows stay exactly where they were. The purge reports `roster_entry_removed: false` and — with #1 also false — `is_empty()`, so the CLI prints *"Nothing to purge: 'x' left no trace in this installation"* while the rows are still there.

That is the whole reason to have the command. It needs a name-keyed path to the agent id: search `sessions` / `memories` for rows whose agent no longer exists, or record the id somewhere the purge can recover it. Whatever the mechanism, the test to write first is: seed sessions and memories for an agent id, remove only the roster row, purge by name, assert the rows are gone.

**3. A `remove_dir_all` on a user workspace with no confirmation, no `--yes`, no `--dry-run`.**

`librefang purge --agent foo` deletes the agent's entire workspace directory immediately — user files, not only LibreFang state — and it is irreversible. Every comparable destructive surface in this project gates on a confirmation (the dashboard's `ConfirmDialog`, "Clear all"), and the repo's own safety hooks block unguarded `rm -rf` against exactly this kind of target.

Please add `--yes` (required when stdin is not a TTY, interactive prompt when it is) and a `--dry-run` that prints the `PurgeReport` it *would* produce. `PurgeReport` is already shaped for that — computing it without acting is a small refactor and makes the command safe to explore.

## Non-blocking

- Both doc comments claim shared use — *"shared by the CLI, the API, the dashboard and the TUI"* and *"the shared implementation the API, the dashboard and the TUI use too"* — but only the CLI calls it in this diff. Reword to say it is written to be shared, or wire one more caller.
- The workspace and agent-type deletions run after the roster removal with `?` on each, so a failure to delete the workspace leaves the roster entry already gone and the report unreturned. Given the function's idempotency promise ("purging an agent that is already partly gone cleans up whatever remains"), a partial failure would be better reported than propagated — return the `PurgeReport` alongside the error, or collect failures and report all of them.
- `MemorySubstrate::open(&db, 0.01)` — the bare `0.01` reads as a magic number at the call site. A named const or a comment naming the parameter would help.
- No changelog fragment. A new destructive CLI command is exactly the kind of thing the release notes must carry: `changelog.d/added/7986-....md`.
- `Security` / `CI Gate` are red on pre-existing `cargo audit` advisories in transitive dependencies (`lru` 0.18.0, `rand` 0.7.3), not on this diff.

DaBlitzStein:
Reworked the purge command to address the review feedback; commit `fa67720de`.

- **Read-only preview before delete**: new `plan_purge()` returns a `PurgePlan` (`preview: PurgeReport` + `failures`), surfaced via `--dry-run` and the confirmation prompt, so the operator sees exactly what will be removed. A plan with roster read failures is never executed.
- **Truthful reporting**: `PurgeReport` now distinguishes `roster_entry_removed`, `orphaned_data_removed`, `identity_record_removed`, `workspace_removed`, and `agent_type_removed`. `purge_agent` returns `PurgeOutcome { report, failures }` so partial success is visible and rerunnable (the module is idempotent).
- **Agent-type handling via the canonical store**: paths resolve through `agent_type_path_in` and names validate via `validate_agent_type_name`, and the orphan path recovers the agent id through `AgentIdentityRegistry` plus the deterministic name-derived UUID, skipping any id a live roster entry still holds.

Verification: `cargo clippy -p librefang-kernel -p librefang-cli --all-targets -- -D warnings` clean; `cargo test -p librefang-kernel --lib agent_purge` — 8 passed (purge_by_name_cleans_rows_that_outlived_the_roster_entry, orphan_rows_are_found_through_the_identity_registry_even_for_random_ids, a_live_agents_id_is_never_treated_as_an_orphan, dry_run_plan_matches_what_the_real_purge_removes, it_removes_the_workspace_and_the_agent_type, it_leaves_every_other_agent_alone, purging_something_already_gone_is_not_an_error, a_path_shaped_name_never_reaches_the_filesystem). Branch is now mergeable.

houko (dismissed):
All three blocking points resolved, and the design grew a fail-safe I did not ask for and would not have thought of.

**1 — the agent-type path goes through the shared helper.** `agent_type_path_in` and `validate_agent_type_name` are imported from `librefang_types::agent_type_store` (`agent_purge.rs:17`), the hand-rolled traversal check is replaced by `validate_agent_type_name` (`:97`), and the test fixture now builds its path with `agent_types_dir_in` (`:294`). That is the part that matters: reader and writer now derive the location from one function, so a future divergence fails the test instead of hiding in it. (`home.join("workspaces").join("agents")` at `:166` is the workspace directory, correctly a different tree — I checked that was not collateral.)

**2 — the headline case works.** `plan_purge` collects `orphan_agent_ids` by scanning for rows whose id no roster entry holds, `purge_agent` calls `remove_agent` for each, and `is_empty()` now counts `orphaned_data_removed` — so the "Nothing to purge: left no trace" message no longer fires over rows that are still there. `purge_by_name_cleans_rows_that_outlived_the_roster_entry` is the test, with `delete_roster_row_only` as the fixture, which is exactly the shape I asked for.

**3 — both flags, and the non-TTY gate is real.** `--dry-run` runs `plan_purge` and prints without writing. The destructive path warns, then `prompt_yes_no(…, false)` defaults to no, so an EOF on non-TTY stdin answers "no" and `--yes` is effectively required there.

## The fail-safe is the best thing in this round

`PurgePlan.failures`, and the fact that `purge_agent` returns early when it is non-empty rather than merely documenting the rule:

> A plan with failures must not be executed: a roster read failure means live agents cannot be told apart from orphans, and executing could cascade a live agent's data.

That is a real hazard created by the orphan-scanning fix itself — the moment "no roster entry holds this id" becomes a deletion trigger, a failed roster read makes *every* id look orphaned. Refusing to act on an incomplete plan is the correct response, and I verified it is enforced in code at the top of `purge_agent`, not just asserted in the doc. Splitting `plan_purge` out for `--dry-run` is what made that check possible, so the two changes reinforce each other.

## One inaccuracy to settle before merge

`plan_purge`'s doc comment claims more than the code does:

> Confirmation prompts and `--dry-run` show this, so the operator confirms what will actually happen rather than a guess.

`--dry-run` does. The confirmation prompt does not — `cmd_purge`'s destructive path never calls `plan_purge`; it prints `purge-confirm-warning`, which is a static enumeration of every category that *could* be removed ("the workspace directory and any agent-type of the same name"), and then prompts. So the operator confirms a generic worst case, not the computed plan: they are not told which workspace path is about to go, or whether one exists for this agent at all.

Either fix the sentence or make it true. I would make it true — the plan is already computed and printable, so the destructive path becomes `plan_purge` → print preview → prompt → `purge_agent`. The double plan is cheap, and for an irreversible `remove_dir_all` over user files, showing the actual paths before asking is the whole point of having a preview. A doc comment that asserts a safety property the code does not deliver is the worse half of this, because the next person will trust it.

`CI Gate` green, `mergeable` is `BLOCKED` on review rather than a conflict.

Approving — the remaining item is a two-line change and I would rather not hold the command over it, but please do resolve it one way or the other rather than leaving the doc as it stands.

houko:
Approved above, but the branch now conflicts and GitHub cannot auto-update it — `gh pr update-branch` fails with "Cannot update PR branch due to conflicts". That is my merge ordering: #7784 (`feat(cli): add the goal command and a Goals tab to the TUI`) landed as `8ad0dc7aa` a few minutes ago, and it adds a subcommand through the same insertion points this PR does.

Seven files conflict, and all seven are the "both sides added an entry to the same list" shape:

- `crates/librefang-cli/src/cli.rs` — #7784 added a `Goal` variant, this adds `Purge`.
- `crates/librefang-cli/src/commands/mod.rs` — both register a new command module.
- `crates/librefang-cli/src/main.rs` — both add a dispatch arm.
- `crates/librefang-cli/locales/{en,ko,uk,zh-CN}/main.ftl` — both append a block of new keys.

Keep both sides everywhere; there is no semantic interaction between `goal` and `purge`. The only thing to watch is the locale files: #7784 landed `cmd-goal-*` plus `tui-goals-*` keys and #7986 lands `purge-*`, so a resolution that takes one side wholesale in any of the four files will silently drop the other command's strings — and `crates/librefang-cli/tests/i18n_checks.rs` will catch it as missing keys rather than as a merge mistake, which is a confusing way to find out.

I have not touched the branch myself since force-pushing someone else's branch is not mine to do. #7998 was in the same cluster and updated cleanly, so this is the only one that needs a hand.

Nothing about the review changes — the three blocking points stay resolved, and the one item I left open (the confirmation prompt not showing the computed plan, while `plan_purge`'s doc says it does) is still worth settling in the same push.

DaBlitzStein:
Refreshed onto current `main` — the branch had gone `CONFLICTING` against the `goal` command landing. Head is now `fb17e682b`; both commands coexist in the `Commands` enum (`Goal` at `cli.rs:304`, `Purge` at `cli.rs:322`).

Verification on the merged state:

- `cargo check -p librefang-kernel -p librefang-cli --all-targets` clean, `clippy -D warnings` clean on both crates.
- `cargo nextest run -p librefang-kernel --no-fail-fast` — **1908 passed**, 0 failed.
- `cargo nextest run -p librefang-cli -E 'kind(lib) | kind(bin)' --no-fail-fast` — **481 passed**. Baseline before the merge was 459; the +22 are the upstream goal-command tests, so coverage grew rather than regressed.
- Purge i18n: the 12-key purge block (`purge-confirm-warning` … `purge-removed-workspace`) verified present with identical key sets in all four CLI locales (`en`, `ko`, `uk`, `zh-CN`).

houko (changes requested):
Re-reading this at head `fb17e682` before merge, and I have to walk back my approval: the orphan scan sees four of the roughly twenty tables the cascade clears, so the headline case is still under-delivered — just narrower than in round 2. Two other path/name issues below, plus the item from my approval that is still open.

## Blocking

**1. `has_agent_rows` scans 4 tables; `remove_agent` clears ~20, so the orphan gate misses most of the residue it exists to find.**

```rust
for table in ["sessions", "canonical_sessions", "memories", "kv_store"] {
```

and the doc above it:

> Scans the same tables the substrate's `remove_agent` cascade clears, so anything the scan can see, the cascade removes.

The implication in that sentence is true; the sentence itself is not, and the gap is the direction that matters. `execute_structured_agent_deletes` (`crates/librefang-memory/src/structured.rs:838-874`) also clears `usage_events`, `task_queue`, `entities`, `relations`, `events`, `pending_approvals`, `goal_runs`, `approval_audit`, `prompt_experiments`, `prompt_versions`, `experiment_variants`, `experiment_metrics`; `execute_session_agent_deletes` adds `sessions_fts`; `execute_ephemeral_run_agent_deletes` adds `ephemeral_runs`.

`has_agent_rows` gates the cascade. An orphan whose surviving rows are in any of those unscanned tables returns `Ok(false)`, `orphan_agent_ids` stays empty, `remove_agent` is never called, and the CLI prints *"Nothing to purge: 'x' left no trace in this installation"* over rows that are still there. That is the same failure mode as round 2's blocking point 2.

This is not hypothetical, and the file itself records why. The cascade grew over time:

```rust
// pending_approvals (v26 — #3611) was missing from this
// cascade; the audit found that on `remove_agent` the table
// would retain rows for the deleted agent … (audit: agent-cascade-delete-missing-tables)
```

and `goal_runs` came in at v42. Any installation that deleted an agent before those versions has residue in exactly the tables the scan cannot see — which is precisely the population `librefang purge` is for. Time-based retention diverging between `sessions` (`session.rs:1391`) and `usage_events` (`usage.rs:1705`) produces the same state on a current build.

Note the predicates are not uniform, so a scan list cannot just be table names: `events` keys on `source_agent`, `ephemeral_runs` on `parent_agent_id`, everything else on `agent_id`, and `agents` on `id`. That argues for one shared `&[(table, column)]` const in `librefang-memory` that both the cascade and the scan iterate, rather than a second hand-maintained list here — the audit above is what a duplicated list costs.

**2. The workspace path is a literal, and the two things that move it are ignored.**

```rust
let workspace = home.join("workspaces").join("agents").join(agent_name);
```

`workspaces_dir` is configurable (`crates/librefang-types/src/config/types.rs:3835`), and the resolved location comes from `effective_agent_workspaces_dir()` (`types.rs:7140`) = `workspaces_dir.unwrap_or(home/workspaces)/agents`. An agent can also carry a per-agent override, `AgentManifest.workspace` (`agent.rs:1233-1236`), which `resolve_workspace_dir` (`workspace_setup.rs:572-618`) honours as a path relative to the root or as an absolute path inside it.

With either in play `workspace.is_dir()` is false, the report says nothing was there, and the directory survives the purge — reported as success. This is the same shape as round 2's blocking point 1, where the agent-type file was read from a hand-written path instead of the shared helper: the helper exists, and using it is what stops the next relocation from silently breaking this. (Reading the override needs the manifest, which is more work than swapping in a helper — if that is out of scope for this PR, then the report needs to say the workspace location was not resolvable rather than claim nothing was there.)

**3. `validate_agent_type_name` is the wrong validator for an agent name, and it refuses agents that legitimately exist.**

```rust
if let Err(reason) = validate_agent_type_name(agent_name) {
    failures.push(format!("invalid agent name {agent_name:?}: {reason}"));
```

That helper is `[A-Za-z0-9_-]{1,64}` (`agent_type_store.rs:70-81`) because it names a file in the agent-type store. Agent names are not bound by it — the only rule at the registry boundary is `validate_agent_name` (`agent.rs:204-213`), which rejects the reserved `_operator:` prefix and nothing else. So an agent named `my.bot`, `data agent` or `研究员` is ordinary, and `librefang purge --agent 研究员` refuses to do anything at all.

This one is my fault: I suggested reusing `validate_agent_type_name` in round 1, and it is right for deriving the agent-type filename — it is wrong as the gate on the whole operation. The traversal defence you need is that the name cannot contain a separator or `..`; keep that check (or apply `validate_agent_type_name` only to the agent-type branch) and let the rest of the purge run for names it rejects.

Worth pairing with point 2: the default workspace directory is `safe_path_component(agent_name, …)` (`workspace_setup.rs:538-548`), which strips exactly the characters `validate_agent_type_name` rejects. So once names outside that set can be purged, `join(agent_name)` is no longer the right join either — another reason both should come from the same helper as the spawn side.

## Still open from my approval

`plan_purge`'s doc still reads "Confirmation prompts and `--dry-run` show this", and at `fb17e682` the destructive path in `cmd_purge` still prints the static `purge-confirm-warning` and prompts without ever calling `plan_purge`. No code commit has landed since the approval (`fa67720d` predates it; `fb17e682` is a merge from main), so this is unchanged rather than disputed. Same request as before: fix the sentence or make it true, and I would still make it true.

## Not a problem, since I checked

- `PurgeReport`'s `is_empty()` counting `orphaned_data_removed`, the refusal to execute a plan with failures, and the "only unbind the identity record when every cascade succeeded" rule all still hold at this head.
- The post-approval concern about names and `safe_path_component` diverging does **not** apply today, because `validate_agent_type_name` and `safe_path_component` filter on the identical character set — it only becomes live once point 3 is fixed, which is why I raised them together.

DaBlitzStein:
Completing the verification for `82efb4337`, which was pushed with only the compile check.

```
Starting 2 tests across 24 binaries (1909 tests skipped)
PASS [0.051s] agent_purge::tests::purge_by_name_cleans_rows_that_outlived_the_roster_entry
PASS [0.053s] agent_purge::tests::dry_run_plan_matches_what_the_real_purge_removes
Summary [0.055s] 2 tests run: 2 passed, 1909 skipped
```

Two asked for, two started, two passed, on a machine at load 0.86 — so there is no ambient explanation available for either outcome, in whichever direction it had gone.

`dry_run_plan_matches_what_the_real_purge_removes` is the one that matters here: it pins that the dry run does not lie about what the real purge will delete. A discrepancy there is the kind an operator only discovers after running the destructive version.

The 51 and 53 millisecond durations are worth noting next to the rest of this refresh round: several suites in this repo were producing failures at 31-35 s with `Memory init failed: … database is locked`, which is the connection pool's deadline expiring under machine load rather than any lock. Nothing resembling that signature here.

`cargo clean -p librefang-types -p librefang-kernel` was run first, deliberately. Manually pruning a shared target directory leaves cargo's fingerprints inconsistent, and it will then reuse another worktree's `rmeta` — which produced errors naming fields belonging to an entirely different branch three separate times today, each looking exactly like a real compile failure.

crates/librefang-cli/src/commands/purge.rs
```diff
@@ -0,0 +1,211 @@
...
+/// fires — the substrate just requires one.
+const PURGE_DECAY_RATE: f32 = 0.01;
+
+pub(crate) fn cmd_purge(config: Option<&Path>, agent: &str, yes: bool, dry_run: bool) -> i32 {
+    // The config decides where the workspaces root is (`workspaces_dir`), so
+    // purge has to read it rather than assume `{home}/workspaces/agents`.
+    // `load_config` already reports a parse failure on stderr (#5186); falling
+    // back to the defaults keeps the command usable on a broken config, which
+    // is a state a cleanup command should expect to meet.
+    let config = librefang_kernel::config::load_config(config).unwrap_or_else(|e| {
```
houko: [P1] Abort purge when the selected configuration cannot be loaded

The new configuration-aware path catches a load error and substitutes `KernelConfig::default()` before selecting the database and workspace root.
For `librefang --config /srv/instance-b/config.toml purge --agent worker --yes`, a TOML error can therefore switch the deletion target to the default installation and purge its unrelated `worker` instead.
The warning does not make that target substitution safe, and the preview prints removal categories rather than the selected database path.

Use the strict config loader and return a nonzero exit status before opening a substrate on any configuration failure.
The ordinary loader also falls back on unreadable/missing files and some other failures, so merely replacing `unwrap_or_else` is insufficient.
A regression test should supply an invalid explicit config while a separate default installation contains the same agent name, and assert that no deletion is attempted.

houko (changes requested):
One new finding from the current-head review. Existing discussion and reviews were checked to avoid duplicating earlier feedback.

DaBlitzStein:
Addressed at `428d61a3b`, all four points — and thank you for walking back the approval, the four-table scan was the real gap.

**1 — one shared list, read by both halves.** `AGENT_SCOPED_TABLES` is a new `&[(AgentTableGroup, &str, &str)]` const in `librefang-memory`, carrying the column per table because the predicates are not uniform (`events` on `source_agent`, `ephemeral_runs` on `parent_agent_id`, `agents` on `id`). The `remove_agent` cascade and `has_agent_rows` both iterate it, so a table added to one is added to the other by construction. `audit_entries` stays out, with the #6553 reasoning on the constant itself.

Restricting the scan back to the old four tables fails exactly one test and leaves the other twelve green:

```
orphan_rows_outside_the_old_four_table_scan_are_found_and_purged ... FAILED
test result: FAILED. 12 passed; 1 failed
```

**2 — the workspace comes from the shared helper**, so a configured `workspaces_dir` is honoured, and a per-agent `AgentManifest.workspace` override is read too. Where it genuinely cannot be resolved the report says the location was unresolvable rather than claiming nothing was there — `an_orphans_missing_workspace_is_reported_as_unresolved_not_absent` pins that distinction.

**3 — `validate_agent_type_name` no longer gates the whole operation.** The traversal defence is the separator/`..` check; the agent-type charset applies only to the agent-type branch. `names_outside_the_agent_type_charset_are_purged_not_refused` covers a name your review named.

**4 — `plan_purge` is called on the destructive path**, so the sentence about confirmation prompts showing the plan is true rather than fixed by weakening it.

13 tests green.

Ready for another look — and the new P1 on the config loader is being handled separately.

DaBlitzStein:
The P1 is addressed at `71a6cce81`.

`purge` uses `try_load_config` now and returns a nonzero status before any substrate is opened. You were right that swapping the `unwrap_or_else` was not enough — `load_config` falls back on an unreadable or missing file too, so the target substitution had more entrances than the TOML-error one.

The one distinction I kept is between a config that was **named** and one that was not:

```rust
fn resolve_config(explicit: Option<&Path>) -> Result<KernelConfig, String> {
    let path = explicit.map(Path::to_path_buf).unwrap_or_else(default_config_path);
    if explicit.is_none() && !path.exists() {
        return Ok(KernelConfig::default());
    }
    try_load_config(&path)
}
```

If `--config` names a file, any failure aborts. If nothing was named and there is no config at all, defaults are legitimate — that is a fresh install, not a substituted target. A destructive command should not guess which installation it is pointed at, but it should still work on one that has never been configured.

The preview now prints the selected database path, so the operator sees the target rather than only the removal categories.

Three tests, and reverting to the silent fallback fails the two that matter:

```
an_unloadable_explicit_config_purges_nothing_from_the_default_installation ... FAILED
  assertion failed: an unloadable config must fail the command
a_missing_explicit_config_purges_nothing_from_the_default_installation ... FAILED
  assertion failed: a config that is not there must fail the command
```

Both are the shape you asked for: an invalid explicit config while a separate default installation holds an agent of the same name, asserting no deletion is attempted. `a_missing_default_config_still_purges_the_default_installation` pins the other side so the strictness does not break a fresh install.

The 13 `agent_purge` tests from the previous round are still green, as are the three i18n guards; the failure message is in all four locales.

Ready for another look.

DaBlitzStein:
A question for you rather than a defect report, surfaced while integrating this PR alongside the other 40 open ones.

`AGENT_SCOPED_TABLES` in `crates/librefang-memory/src/agent_tables.rs` is the single list both `execute_structured_agent_deletes` and the orphan scan in `agent_purge.rs:370` walk, which is exactly the consolidation this PR is for. Cross-checking that list against every `CREATE TABLE` in `migration.rs`, **14 tables declare an `agent_id` column and three real ones are not in the list**:

- `audit_entries` — almost certainly deliberate, and the comment this PR moved says so: the WORM Merkle trail must not lose rows when an agent is removed, or `security verify` reads the gap as tampering. Not asking about this one.
- `canonical_sessions`
- `prompt_experiments`

The third is the one that made me write. `prompt_versions` **is** in the list; `prompt_experiments` is not, and they are the same domain. If that is an oversight rather than a decision, `remove_agent` leaves experiment rows behind for a deleted agent, and because the orphan scan reads the same constant it does not report them either — the two halves agree with each other and both miss it, which is the failure mode this constant exists to prevent.

`canonical_sessions` has the same shape but I have less of a read on whether session canonicalisation is meant to outlive the agent.

**This is not a regression from your diff** — I checked. The list on this PR's own side is identical to what the integration carries apart from one entry I had to re-add: `manifest_versions`, from #8041, which the two branches' resolution dropped because #8041 still spells its cascade out inline in `structured.rs` while this PR moved the flat list into the constant. Whichever of the two merges second needs that row, and it is one line.

No action requested here beyond telling me whether the three absences are decisions. If `prompt_experiments` is an oversight I am happy to open it as its own PR against this one rather than growing this diff.

crates/librefang-cli/src/commands/purge.rs
```diff
@@ -0,0 +1,358 @@
...
+pub(crate) fn cmd_purge(config: Option<&Path>, agent: &str, yes: bool, dry_run: bool) -> i32 {
+    let config = match resolve_config(config) {
+        Ok(c) => c,
+        Err(e) => {
+            eprintln!("{}", i18n::t_args("purge-failed-config", &[("error", &e)]));
+            return 1;
+        }
+    };
+    let home = config.home_dir.clone();
+    let db = home.join("data").join("librefang.db");
```
houko: **The database this deletes from is not necessarily the one the daemon uses.** `home.join("data").join("librefang.db")` hardcodes the default layout, but the daemon resolves its DB as `config.memory.sqlite_path.clone().unwrap_or_else(|| config.data_dir.join("librefang.db"))` (`crates/librefang-kernel/src/kernel/boot.rs:379-383`), and both are operator-settable (`KernelConfig.data_dir` at `crates/librefang-types/src/config/types.rs:3470`, `MemoryConfig.sqlite_path` at `:7401`). Nothing normalises `data_dir` from `home_dir` either, so a config with `home_dir = "/srv/lf"` and no `data_dir` already diverges (`data_dir` falls back to `librefang_home().join("data")`).

Two concrete outcomes:

- `data_dir = /mnt/big/lf-data`, or `[memory] sqlite_path` set: `<home>/data/librefang.db` does not exist, so the command exits 1 with "no database at ..." and can never purge that installation at all.
- The same installation after a `data_dir` move that left the old file behind: purge opens the stale `<home>/data/librefang.db`, prints `Database: <home>/data/librefang.db`, reports what it removed from that file and exits 0 — while the live rows survive untouched. That defeats the stated purpose of the `purge-database-line` output, which is to tell the operator *whose* data is going.

`crates/librefang-cli/src/commands/monitoring.rs:149` already has the correct resolution; reuse it.

crates/librefang-kernel/src/agent_purge.rs
```diff
@@ -0,0 +1,802 @@
...
+    // Same rule here: only unbind when every cascade above succeeded.
+    if plan.registry_record && failures.is_empty() {
+        let registry = AgentIdentityRegistry::load(home);
+        if registry.purge(agent_name).is_some() {
+            report.identity_record_removed = true;
+        }
+    }
+
+    if let Some(workspace) = &plan.workspace {
+        match std::fs::remove_dir_all(workspace) {
```
houko, line 538: **`remove_dir_all` on a manifest-supplied path, with no check that the directory is exclusively this agent's.**

`resolved_workspace_dir` rejects only `..` and non-absolute Windows prefixes (`has_unsafe_relative_components`); it will happily hand back the workspaces root itself. A manifest with `workspace = "."` (or `""`, or an absolute path equal to `<workspaces>/agents`) resolves to `root.join(".")`, `is_dir()` is true, and this line then recursively deletes `<workspaces>/agents` — every agent's workspace on the installation. Spawn accepts such a manifest today: it runs the same resolver and then just `ensure_workspace`es the root, which is harmless there and catastrophic here.

The milder variant needs no unusual manifest at all: two agents configured with the same `workspace = "shared/team"` share a directory by design (private state lives in each agent's own `.identity/`). Purging one deletes the other, still-live agent's workspace, and the report says only "- workspace directory".

Before deleting, require that the resolved path is a *strict* descendant of the workspaces root (at least one component below it, not equal to it), and that no other roster entry resolves to the same directory.

crates/librefang-cli/src/commands/purge.rs
```diff
@@ -0,0 +1,358 @@
...
+            "{}",
+            i18n::t_args(
+                "purge-failed-no-database",
+                &[("path", &db.display().to_string())]
+            )
+        );
+        return 1;
+    }
+
+    let substrate = match MemorySubstrate::open(&db, PURGE_DECAY_RATE) {
```
houko, line 71: **Nothing checks whether the daemon is running.** The module doc says purge works with no daemon, "which is the usual situation" — but usual is not always, and there is no guard for the other case.

Scenario: agent `alpha` is live under a running daemon. `plan_purge` finds its roster entry, so the destructive path deletes the `agents` row plus its sessions / memories / KV / usage rows, then `remove_dir_all`s its workspace. The kernel still holds `alpha` in memory with the same `AgentId` and keeps running it against a workspace whose `.identity/` files have just been deleted; the next `save_agent` re-persists the `agents` row, leaving a resurrected roster entry with no history and no workspace. The identity record has been dropped as well, so a later respawn of a random-id agent lands on a different UUID.

`find_daemon_in_home(&home)` already exists at `crates/librefang-cli/src/commands/common.rs:326` and is exactly the probe needed: refuse when it returns `Some`, or require an explicit override flag.

crates/librefang-cli/src/commands/purge.rs
```diff
@@ -0,0 +1,358 @@
...
+use librefang_types::config::KernelConfig;
+use std::path::Path;
+
+/// Memory decay rate handed to `MemorySubstrate::open` (0.0 = no decay,
+/// 1.0 = aggressive decay; the kernel's own default is 0.1). Purge only
+/// deletes rows and never runs decay or consolidation, so the value never
+/// fires — the substrate just requires one.
+const PURGE_DECAY_RATE: f32 = 0.01;
+
+pub(crate) fn cmd_purge(config: Option<&Path>, agent: &str, yes: bool, dry_run: bool) -> i32 {
```
houko: **`--dry-run` is advertised as "without changing anything", but the substrate is opened unconditionally on both paths and opening it writes.** `MemorySubstrate::open` → `open_with_pool_size` (`crates/librefang-memory/src/substrate.rs:214-240`) does three mutating things before returning:

- `run_migrations(&migration_conn)` — applies every pending schema migration. A CLI newer than the DB silently upgrades the operator's database during what they were told was a preview, and migrations do not roll back.
- `restrict_db_file_permissions(db_path)` — chmods the DB, `-wal` and `-shm` to 0600.
- `sessions.reconcile_fts_index()` — runs `DELETE FROM sessions_fts WHERE session_id NOT IN (SELECT id FROM sessions)` (`crates/librefang-memory/src/session.rs:274-284`).

Either open read-only when `dry_run` is set, or drop the "without changing anything" claim from the `--dry-run` help text, the changelog fragment and `purge-dry-run-header` and state what it does touch.

crates/librefang-kernel/src/agent_purge.rs
```diff
@@ -0,0 +1,802 @@
...
+            }
+        }
+    }
+    preview.identity_record_removed = registry_record;
+
+    // The workspaces root is configurable (`workspaces_dir`) and an agent can
+    // carry its own `workspace` override, so neither the root nor the leaf is
+    // safe to spell out here: both come from the same helpers spawn used to
+    // create the directory in the first place.
+    let workspaces_root = cfg.effective_agent_workspaces_dir();
```
houko: **Hand agents' workspaces are never purged, and the report explains their absence with a statement that is false for them.**

This root is `effective_agent_workspaces_dir()` = `<workspaces>/agents`, but hand agents live under `<workspaces>/hands/<hand>/<role>`: `hands_lifecycle.rs:481` sets `manifest.workspace = Some("hands/<hand>/<role>")` and `spawn.rs:289` rewrites it to the resolved absolute path, which is what the roster entry carries. On the `Some(entry)` branch that absolute path fails `path.starts_with(&root)` inside `resolved_workspace_dir`, so the `Err(_) => (None, true)` arm below fires: the hand's workspace stays on disk, and the operator is shown `purge-workspace-unresolved`, which reads "No manifest survives for this agent, so a custom workspace path would not have been found." The manifest does survive and does name the directory — the resolver was simply pointed at the wrong root.

`effective_workspaces_dir()` (or the tag-aware logic in `backfill_workspace_dir`) is the right root for the manifest-override branch. Ephemeral mission workspaces (`ephemeral_spawn.rs:246`, under `<home>/transient/`) hit the same arm.

crates/librefang-kernel/src/agent_purge.rs
```diff
@@ -0,0 +1,802 @@
...
+///
+/// Idempotent by design: purging an agent that is already partly gone cleans
+/// up whatever remains and reports it, rather than failing. That is the whole
+/// point — the caller is here precisely because a previous delete left
+/// something behind.
+///
+/// Never aborts halfway: every step runs (or is skipped because planning
+/// could not prove it safe), and [`PurgeOutcome::failures`] lists everything
+/// that went wrong. Rerun the command to finish a partial purge.
+pub fn purge_agent(
```
houko, line 408: **Agent-scoped state outside the substrate survives, and because top-level agent ids are name-derived a new agent taking the purged name inherits it.**

- `<home>/data/cron_jobs.json` (`cron.rs:183`) and `<home>/trigger_jobs.json` (`triggers.rs:370`) are keyed by `AgentId` and are not touched here. Spawn derives `AgentId::from_name(&name)` for a top-level agent (`spawn.rs:124-137`), so re-creating an agent with the purged name yields the *same* UUID — the new agent silently begins firing the purged agent's cron schedule and event triggers.
- `channel_instance_defaults` and `conversation_bindings` (schema v45) are keyed by `agent_name`, and are only ever deleted by `instance_name` / `conversation_id` (`channel_binding_store.rs:96,178`). The purged agent's inbound channel routing therefore stays bound to its name, and the next agent to take that name inherits its conversations.

For a command whose doc comment and changelog both say "every trace", these should either be purged or be surfaced in the report as known residue, the same way `workspace_unresolved` is.

crates/librefang-kernel/src/agent_purge.rs
```diff
@@ -0,0 +1,802 @@
...
+        }
+    }
+
+    // Drop the name → UUID binding last-but-not-unconditionally: the kernel
+    // skips it when the roster row could not be removed (#5117) so the next
+    // boot never loads a roster row whose name the registry no longer knows.
+    // Same rule here: only unbind when every cascade above succeeded.
+    if plan.registry_record && failures.is_empty() {
+        let registry = AgentIdentityRegistry::load(home);
+        if registry.purge(agent_name).is_some() {
```
houko, line 527: **`identity_record_removed` is set on an in-memory removal that may never reach disk.** `AgentIdentityRegistry::purge` returns `Some(uuid)` as soon as the `DashMap` entry is gone; a failed `persist()` is only a `warn!` (`agent_identity_registry.rs:208-218`). The registry loaded on the line above is local to this process and dropped moments later, so with a read-only or full home directory the on-disk `agent_identities.toml` still contains `alpha -> <uuid>` while the command prints "- canonical UUID record in agent_identities.toml" and exits 0. The next spawn of `alpha` then reuses the purged UUID — the precise outcome this field claims to have prevented.

Call `registry.persist()` explicitly (or have `purge` return the error) and push into `failures` instead of setting the flag when it fails.

crates/librefang-kernel/src/agent_purge.rs
```diff
@@ -0,0 +1,802 @@
...
+        }
+        None => {
+            // Orphan path — the whole reason this module exists: the roster
+            // entry is gone but its rows are not. Recover the agent id from
+            // its name. Two sources: the canonical-UUID registry (covers
+            // agents spawned with a random id) and the deterministic
+            // name-derived UUID. A candidate that any live roster entry
+            // holds (an agent renamed since spawn, keeping its id) is never
+            // touched — its data belongs to a running agent.
+            let mut candidates: Vec<AgentId> = Vec::new();
```
houko: **Id recovery covers only two of the three ways an agent gets an id, and the miss is reported as "left no trace in this installation".**

`register_if_absent` runs only for top-level agents with no `predetermined_id` (`spawn.rs:124-137`); child and ephemeral agents get `AgentId::new()` and are never recorded in the registry. For one of those, once its roster row is gone both candidates here are wrong — the registry has no record, and `AgentId::from_name(name)` is not its id — so `has_agent_rows` answers false for everything and the CLI prints `purge-nothing-to-purge` ("'x' left no trace in this installation") and exits 0, over live rows in `sessions`, `memories`, `usage_events` and the rest.

`PurgeReport` already carries `workspace_unresolved` so that "nothing removed" is not read as "nothing there"; the id side needs the same honesty. A `SELECT DISTINCT agent_id FROM <table> WHERE agent_id NOT IN (SELECT id FROM agents)` sweep would at least let the command say that orphan ids exist which it cannot attribute to a name.

crates/librefang-kernel/src/agent_purge.rs
```diff
@@ -0,0 +1,802 @@
...
+
+    let registry = AgentIdentityRegistry::load(home);
+    let registry_record = registry.get(agent_name).is_some();
+
+    let mut preview = PurgeReport::default();
+    let mut roster_agent_id = None;
+    let mut orphan_agent_ids = Vec::new();
+
+    let roster_entry = entries.iter().find(|e| e.name == agent_name);
+    match roster_entry {
```
houko: **When a roster entry exists the orphan scan is skipped entirely, so residue from a previous incarnation of the same name survives a purge reported as complete.** Delete agent `alpha` in a way that leaves rows behind (the case this module exists for), re-create `alpha` under a different id — any spawn path that supplies `predetermined_id`, or a child spawn with a random id — then run `purge --agent alpha`. This `match` takes the `Some(entry)` arm, cascades only the current id, never looks for the older id's rows, and prints roster entry + workspace + agent-type removed with exit 0.

Running the candidate scan in both arms (skipping ids held by a live roster entry, which the loop below already does) costs one `EXISTS` query per candidate and closes the gap.

crates/librefang-kernel/src/agent_purge.rs
```diff
@@ -0,0 +1,802 @@
...
+    let mut failures = Vec::new();
+    if let Err(reason) = validate_purgeable_name(agent_name) {
+        failures.push(format!("invalid agent name {agent_name:?}: {reason}"));
+        return PurgePlan {
+            failures,
+            ..PurgePlan::default()
+        };
+    }
+
+    let entries = match substrate.load_all_agents() {
```
houko, line 168: **`load_all_agents()` is treated as the authoritative set of live agent ids, but it warn-and-skips rows.** It drops rows with an unparseable UUID, rows whose msgpack manifest fails to deserialize ("Skipping agent with incompatible manifest (schema may have changed)"), and every duplicate of an already-seen lowercased name (`crates/librefang-memory/src/structured.rs:657-682`).

That makes the `entries.iter().any(|e| e.id == id)` guard further down — the one thing standing between an orphan cascade and a live agent's data — unreliable in exactly the circumstances that lead an operator to reach for `purge`. Concretely: agent `alpha` was renamed to `beta` and kept its id `from_name("alpha")`, and `beta`'s manifest blob no longer deserializes after an upgrade. `load_all_agents` skips that row, so `entries` contains neither name; `purge --agent alpha` then classifies `from_name("alpha")` as an orphan and cascades away live agent `beta`'s sessions, memories, KV rows and its `agents` row. `a_live_agents_id_is_never_treated_as_an_orphan` passes only because its seeded manifest deserializes.

Read the ids straight from the table (`SELECT id FROM agents`) rather than through the lossy hydrating loader.

crates/librefang-cli/src/commands/purge.rs
DaBlitzStein: Confirmado contra el árbol actual, no solo contra la descripción.
purge.rs hardcoded home.join("data").join("librefang.db"); monitoring.rs:cmd_audit_reset already resolves via config.memory.sqlite_path.unwrap_or_else(|| config.data_dir.join("librefang.db")) -- confirmed the exact pattern exists and purge.rs did not use it

**Arreglo** (`bcbe5bc18`): Extracted purge_db_path(&KernelConfig) mirroring monitoring.rs's resolution; cmd_purge now calls it instead of the hardcoded join

**Tests:** `db_path_follows_configured_data_dir_not_a_hardcoded_home_join, db_path_prefers_explicit_sqlite_path_over_data_dir (pure function, no I/O)`

**Ciclo rojo:** Reverted purge_db_path to `config.home_dir.join("data").join("librefang.db")`; both tests FAILED: assertion left==right failed, left: "/home-dir-is-irrelevant-here/data/librefang.db" right: "/configured/elsewhere/librefang.db"; and left: "/home/data/librefang.db" right: "/custom/db/path.sqlite". Restored, reran green.

crates/librefang-kernel/src/agent_purge.rs
DaBlitzStein: Confirmado contra el árbol actual, no solo contra la descripción.
resolved_workspace_dir joins a relative override onto root unchanged; `workspace = "."` -> root.join(".") which Rust's Path equality normalizes to the root itself (confirmed by the red-proof panic literally showing `remove_dir_all` attempted on `.../workspaces/agents/.`); no check that two agents' manifests don't resolve to the same directory

**Arreglo** (`bcbe5bc18`): In plan_purge, after resolving the workspace dir, check it against the workspaces root (both the agent-scope root and the overall workspaces root) and against every other roster entry's own resolved workspace dir; on a match, null the deletion candidate and set new PurgeReport.workspace_shared instead

**Tests:** `a_workspace_override_pointing_at_the_root_itself_is_never_deleted, a_workspace_shared_with_another_live_agent_is_never_deleted`

**Ciclo rojo:** Reverted the guard to `let workspace_shared = false;` (skipped). First test FAILED with failures: ["remove workspace /tmp/.tmpgY71CY/workspaces/agents/.: Invalid argument (os error 22)"] -- i.e. the code actually attempted to rm -rf the workspaces root and only an OS-level edge case stopped it from succeeding. Second test FAILED: "a workspace shared with a live agent must never be deleted". Restored, reran green.

crates/librefang-cli/src/commands/purge.rs
DaBlitzStein: Confirmado contra el árbol actual, no solo contra la descripción.
cmd_purge never called find_daemon_in_home (already exists in commands/common.rs, used by status.rs/monitoring.rs/daemon.rs) before running the destructive path

**Arreglo** (`bcbe5bc18`): Added --force CLI flag; cmd_purge now checks find_daemon_in_home(&home) via a small pure decision function daemon_running_guard(daemon_base_url, dry_run, force) before the confirm/execute path, refusing unless dry_run or force

**Tests:** `4 unit tests on the pure daemon_running_guard function covering: refuses when daemon up, dry-run bypasses, force bypasses, no-daemon is a no-op. Did NOT stand up a fake HTTP daemon to test find_daemon_in_home's own I/O -- that function already has its own test coverage elsewhere (main.rs's find_daemon_with_probe tests) and adding a real listener here would be disproportionate to a 2-line glue call.`

**Ciclo rojo:** Reverted daemon_running_guard to always return None. daemon_running_guard_refuses_the_destructive_path_when_a_daemon_is_up FAILED: assertion left==right failed, left: None right: Some("http://127.0.0.1:4545"). Restored, reran green.

crates/librefang-cli/src/commands/purge.rs
DaBlitzStein: Confirmado contra el árbol actual, no solo contra la descripción.
MemorySubstrate::open (substrate.rs:163) unconditionally runs run_migrations, restrict_db_file_permissions (chmod 0600 on db/-wal/-shm), and sessions.reconcile_fts_index() before purge.rs even checks dry_run -- traced the full open_with_pool_size body

**Arreglo** (`bcbe5bc18`): DOCS-ONLY per the auditor's own stated alternative ("open read-only OR correct the docs"). Chose docs: a true read-only SQLite open would need a new MemorySubstrate constructor (no precedent anywhere in the codebase) and would hard-fail on a not-yet-migrated fresh DB file, which is a worse failure mode for a preview command than the honest-docs fix. Updated: Purge CLI arg doc comment in cli.rs, the purge-dry-run-header fluent string (en/uk/zh-CN) to say opening the database may still apply pending migrations and reindex search even though nothing is deleted.

**Tests:** None -- no logic branch changed, only doc/string text. Verified by reading the new strings and confirming i18n_checks (test_no_untranslated_strings, test_locales_cover_used_i18n_keys, test_no_dead_locale_keys) stay green.

**Ciclo rojo:** no aplica, el cambio es solo texto y no hay rama de lógica que invertir.

crates/librefang-kernel/src/agent_purge.rs
DaBlitzStein: Confirmado contra el árbol actual, no solo contra la descripción.
backfill_workspace_dir (workspace_setup.rs:636) writes a hand agent's manifest.workspace as an ABSOLUTE path under effective_hands_workspaces_dir(), but plan_purge validated it against effective_agent_workspaces_dir() (the agents/ subtree) via resolved_workspace_dir's `path.starts_with(&root)` check, which fails for a hands/ path and returns Err -> workspace_unresolved=true

**Arreglo** (`bcbe5bc18`): New entry_workspace_dir(cfg, entry) helper: uses cfg.effective_workspaces_dir() (the whole workspaces root) as the validation root when entry.manifest.workspace is Some(absolute path), and cfg.effective_agent_workspaces_dir() otherwise (preserving the relative-override case exactly as before)

**Tests:** `a_hand_agents_absolute_workspace_override_is_resolved_not_reported_missing`

**Ciclo rojo:** Reverted entry_workspace_dir to always use effective_agent_workspaces_dir() as root. Test FAILED: "a hand agent's absolute workspace path must resolve" panicked at agent_purge.rs. Restored, reran green.

crates/librefang-kernel/src/agent_purge.rs
DaBlitzStein: Confirmado contra el árbol actual, no solo contra la descripción.
grep confirmed: cron_jobs.json entries are JobMeta{ job: CronJob{ agent_id: AgentId, .. } } keyed by id, not name; trigger_jobs.json entries are Trigger{ agent_id: AgentId, .. } same; channel_instance_defaults/conversation_bindings (channel_binding_store.rs) key on agent *name* in SQLite, confirmed by its own doc comment ("Both tables store the agent name, not the per-spawn AgentId uuid"). Also found CronScheduler::remove_agent_jobs(id) and TriggerEngine::remove_agent_triggers(id) ALREADY EXIST and are used by the live kill_agent path (agent_runtime.rs:775/781) -- but purge.rs never boots the kernel, so that path never runs for a CLI purge.

**Decisión:** EXTENDED SCOPE rather than declaring residue. Reused the existing CronScheduler/TriggerEngine load()/remove_agent_*()/persist() methods directly from agent_purge.rs (no kernel boot needed, no new JSON-parsing code duplicated) for cron/trigger removal, keyed by the union of plan.roster_agent_id + plan.orphan_agent_ids. Added two new small SQL helpers (count_channel_bindings_for_agent, remove_channel_bindings_for_agent) for the name-keyed channel tables, since MemorySubstrate::channel_bindings() already exposes a ChannelBindingStore but it has no per-agent-name bulk delete method (only per-instance/per-conversation clears). Chose to extend because: (a) the removal primitives already existed and were a straight reuse, not new design; (b) declaring this 'known residue' would leave a genuinely dangerous default-on inheritance bug (name-derived AgentId means EVERY default-path agent hits this, not an edge case) undocumented as a real security/correctness gap in the one command whose entire job is cleaning up after a deleted agent.

**Arreglo** (`bcbe5bc18`): New PurgeReport fields cron_jobs_removed/trigger_jobs_removed/channel_bindings_removed (counted in is_empty()); plan_purge previews via read-only list_jobs()/list_agent_triggers() on throwaway (never-persisted) engine instances; purge_agent actually removes+persists via the existing remove_agent_jobs/remove_agent_triggers plus the two new channel-binding SQL deletes (name-keyed, runs unconditionally)

**Tests:** `cron_and_trigger_residue_is_removed_so_a_recreated_agent_does_not_inherit_it, channel_binding_residue_is_removed_by_name`

**Ciclo rojo:** Reverted the whole cron/trigger/channel removal block to a no-op. Both tests FAILED: "cron residue must be reported removed" and "assertion failed: outcome.report.channel_bindings_removed" (ran together with F7, F8, F2 reverts in one batch -- 6/6 failed as expected). Restored, reran green (22/22).

crates/librefang-kernel/src/agent_purge.rs
DaBlitzStein: Confirmado contra el árbol actual, no solo contra la descripción.
AgentIdentityRegistry::purge() (agent_identity_registry.rs:208) calls self.persist() internally but only warn!()s on Err and still returns Some(dropped) unconditionally -- confirmed by reading persist()'s Err path and purge()'s body directly

**Arreglo** (`bcbe5bc18`): purge_agent now calls registry.persist() again explicitly after registry.purge() and checks ITS Result, only setting identity_record_removed=true on Ok, pushing a failure message on Err. Chose not to change AgentIdentityRegistry::purge()'s public signature (4 other call sites: agent.rs, agent_runtime.rs, identity.rs, its own tests) to keep the fix scoped to agent_purge.rs.

**Tests:** `identity_persist_failure_is_reported_not_silently_swallowed (#[cfg(unix)], chmod 0500 on home dir between a successful load and persist, matching the existing chmod-test pattern already used in config.rs/boot.rs/hooks.rs/cron_script.rs in this same crate)`

**Ciclo rojo:** Reverted to trusting registry.purge()'s return value without re-checking persist(). Test FAILED: "a persist failure must be surfaced, not swallowed". Restored, reran green.

crates/librefang-kernel/src/agent_purge.rs
DaBlitzStein: Confirmado contra el árbol actual, no solo contra la descripción.
Confirmed via grep that no agent-scoped table (sessions, memories, kv_store, usage_events, etc.) stores a name column anywhere except the agents roster table itself and the two channel-binding tables (which store name, not id, and are unrelated to this problem). A child/ephemeral agent spawned via AgentId::new() (found many call sites: agent_execution.rs:1665/1684, spawn.rs:151, etc.) that is never registered in AgentIdentityRegistry and whose roster row is already gone has literally NO recoverable name anywhere in the schema.

**Decisión:** REPORT ONLY, deliberately did not attempt deletion. The auditor's own suggested fix says exactly this ("report orphan ids it cannot attribute to a name", not "delete them") -- an id with rows in agent-scoped tables that this purge cannot derive from agent_name might belong to a totally different, unrelated name, and deleting on a guess is the wrong direction to err in per this task's explicit instruction (if in doubt, delete less and say so).

**Arreglo** (`bcbe5bc18`): New PurgeReport.other_orphans_present bool (informational, NOT counted in is_empty(), same treatment as workspace_unresolved). New other_orphan_ids_exist(substrate, attributed_ids) helper: DISTINCT scan over every AGENT_SCOPED_TABLES entry (except the agents table itself) for ids NOT IN (SELECT id FROM agents), excluding ids already attributed to this purge's own orphan recovery, existence-only (no ids surfaced, no deletion).

**Tests:** `unattributable_orphan_rows_are_reported_but_never_touched -- explicitly asserts the unrelated id's 3 seeded rows survive AND the flag is set AND report.is_empty() stays true (since nothing attributable to the requested name was found)`

**Ciclo rojo:** Reverted the scan call to a no-op (flag stays default false). Test FAILED: "an orphan id this purge could not attribute to any name must be surfaced". Restored, reran green.

crates/librefang-kernel/src/agent_purge.rs
DaBlitzStein: Confirmado contra el árbol actual, no solo contra la descripción.
The original match roster_entry { Some(entry) => {...}, None => { candidate scan } } ran the candidate scan ONLY in the None arm -- confirmed by reading the exact match structure before editing

**Arreglo** (`bcbe5bc18`): Restructured to `if let Some(entry) = roster_entry { set roster fields }` followed by an unconditional candidate scan (registry id + name-derived id), each candidate checked against the live-ids set (see F10) before cascading -- so a live entry under the SAME name but a DIFFERENT id no longer blocks the scan for stale residue under other ids

**Tests:** `stale_orphans_from_an_earlier_incarnation_are_found_even_when_the_name_is_live_again -- seeds rows under the name-derived id, deletes only that roster row (simulating a partial delete), then creates a NEW live roster entry for the SAME name under a fresh random id (simulating a recreate), and asserts the stale rows are still found and removed`

**Ciclo rojo:** Reverted to the original match structure (candidate scan in None arm only). Test FAILED: "the earlier incarnation's rows, left under a different id, must be found too". Restored, reran green.

crates/librefang-kernel/src/agent_purge.rs
DaBlitzStein: Confirmado contra el árbol actual, no solo contra la descripción.
Read structured.rs:load_all_agents in full: it skips a row on a read error, a duplicate lowercased name, an unparseable UUID, AND an unparseable manifest blob (rmp_serde::from_slice failure) -- all logged as warn!/continue, silently dropped from the returned Vec<AgentEntry>. The live-agent guard `entries.iter().any(|e| e.id == id)` reads exactly that lossy Vec.

**Arreglo** (`bcbe5bc18`): Added `substrate.list_agents()` (already exists, reads `SELECT id, name, state FROM agents` directly with no manifest deserialization) as a raw `HashSet<String>` of live ids; the orphan-candidate loop now checks `live_ids.contains(&id.0.to_string())` instead of `entries.iter().any(|e| e.id == id)`. `entries` (the hydrated Vec) is still used for name lookups and workspace-manifest reads, which genuinely need the full deserialization -- only the id-liveness guard was swapped to the raw source.

**Tests:** `a_live_agent_with_a_corrupt_manifest_is_never_treated_as_an_orphan -- seeds a live agent's rows under a name-derived id that collides with a DIFFERENT name's purge target, then corrupts just its manifest blob via raw SQL UPDATE, confirms load_all_agents drops it but list_agents still sees it, then purges the colliding name and asserts the live agent's 3 rows survive`

**Ciclo rojo:** Reverted the guard to `entries.iter().any(|e| e.id == id)`. Test FAILED: "a live agent's rows must never be cascaded" (the corrupt-manifest live agent's rows got deleted). Restored, reran green.

crates/librefang-cli/src/commands/purge.rs
DaBlitzStein: Arreglado en `71a6cce81`, y confirmado ahora contra el árbol empujado (`bcbe5bc18`).

`resolve_config` (`crates/librefang-cli/src/commands/purge.rs:140`) resuelve por `try_load_config`, el cargador estricto que ya usa `Kernel::reload_config`, y devuelve estado distinto de cero **antes de abrir ningún substrate**.
Tenías razón en que sustituir el `unwrap_or_else` no bastaba: `load_config` también responde `Ok(defaults)` a un fichero ausente o ilegible, así que un `--config` apuntando a una ruta inexistente habría seguido sustituyendo en silencio.
La única ausencia tolerada es un `config.toml` que no está en la ubicación *por defecto* cuando nadie pidió una ruta, donde los valores por defecto describen la misma instalación que el operador quería decir y una instalación que nunca escribió un `config.toml` sigue siendo purgable.
Un config por defecto que existe pero no se puede leer es un fallo como cualquier otro.

El plan que el operador confirma ahora **abre nombrando la base de datos seleccionada**, para que un purge apuntado a la instalación equivocada se vea antes del prompt y no después.

**Tests:** `an_unloadable_explicit_config_purges_nothing_from_the_default_installation`, `a_missing_explicit_config_purges_nothing_from_the_default_installation` y `a_missing_default_config_still_purges_the_default_installation` — los tres montan una instalación por defecto separada que contiene un agente con el mismo nombre, exactamente el caso que describes.

**Ciclo rojo**, con `resolve_config` revertido a la forma permisiva (`Ok(load_config(explicit).unwrap_or_default())`), 2 de 3 en rojo y la salida literal del que importa:

```
error: Config file has invalid TOML and cannot be loaded (path=/tmp/.tmpA5qDrl/config.toml): TOML parse error at line 1, column 6
Database: /tmp/.tmp8wG5jN/data/librefang.db
About to permanently purge 'worker':
- agent-type template
This cannot be undone.
Purged 'worker':
- agent-type template

thread '...an_unloadable_explicit_config_purges_nothing_from_the_default_installation' panicked at crates/librefang-cli/src/commands/purge.rs:448:9:
assertion `left == right` failed: an unloadable config must fail the command
  left: 0
 right: 1
```

Se ve el desplazamiento del objetivo tal cual lo describías: avisa del TOML inválido de la instancia B y a continuación purga `worker` de una base de datos distinta.
Restaurado el arreglo, los 3 vuelven a verde.

DaBlitzStein:
Every one of the 11 review threads on this PR now has an inline reply, and the branch is at `870a38e6d`.

houko (approved):
LGTM. All five findings are fixed, and the workspace one is fixed better than I asked.

- **`remove_dir_all` on an unguarded manifest path.** The guard now refuses both roots (`agent_workspaces_root` and `effective_workspaces_dir()`) *and* any directory shared with another still-registered agent — and it uses **containment**, not equality, in both directions. That is the part that actually matters: an equality test never sees `workspace = "shared"` here against `"shared/team"` there, so the nested agent's data goes with the parent and the report calls it a clean removal. The comment says exactly that.
- **No daemon-running check.** `daemon_running_guard` gates the destructive path, with the reasoning recorded — a live daemon re-persists the agent from memory, a dry run has nothing to race, and `--force` is the explicit override for a stale `daemon.json`. Scoping the guard to the destructive path rather than the whole command is the right call.
- **Agent-scoped state outside the substrate.** `cron_jobs.json` and `trigger_jobs.json` are now in the report (`cron_jobs_removed`, `trigger_jobs_removed`), and the doc on the second cites the name-collision reason rather than repeating it.
- **`identity_record_removed` set on a possibly-undurable removal.** Now conditioned on `registry.persist()` succeeding, with a `failures` entry on error. The comment names why re-checking is necessary — `AgentIdentityRegistry::purge` only `warn!`s on a persist failure — so a read-only home is reported instead of silently no-op-ing.
- **`load_all_agents` warn-and-skipping rows.** Acknowledged in the planning path and, better, covered by tests at `agent_purge.rs:1226` and `:1253` for the damaged-manifest and duplicate-name shapes.

Two things I checked because this command deletes data and the review has to be worth more than the diff:

- `validate_purgeable_name` is genuinely tight: exactly one `Component::Normal`, compared back against the raw input, which rejects `..`, `.`, roots, Windows drive prefixes, every separator, and the shapes the parser silently normalises (`a/`, `./a`).
- `has_agent_rows` walks `AGENT_SCOPED_TABLES` — the same constant the substrate's `remove_agent` cascade builds its DELETEs from — so the "leaves no trace" report cannot see less than the cascade clears. The comment records that this replaced a hand-maintained four-table list which reported sixteen tables' worth of surviving rows as clean.
- The confirmation prompt defaults to **no** and reads EOF as no on a non-TTY stdin, and it is only shown when the plan would actually remove something.

All checks green; `BLOCKED` is the stale `CHANGES_REQUESTED` this clears.

Issue #7758, opened 2026-08-20
question: what is ~/.librefang/templates/ for, now that agent types live in the registry?


DaBlitzStein:
## The overlap

An installation ends up with agent definitions in three places:

| Directory | Holds | Typical count on a real install |
|---|---|---|
| `~/.librefang/registry/agents/` | Definitions synced from the `librefang-registry` repository | 32 |
| `~/.librefang/templates/` | Definitions created through `POST /api/agent-types` | 0 |
| `~/.librefang/workspaces/agents/` | Deployed agents, each with its own state | 42 |

The third is clearly distinct: those are instances, and their `agent.toml` diverges from whatever definition they started from (verified on a live install — `registry/agents/analyst/agent.toml` and `workspaces/agents/analyst/agent.toml` differ).

The first two are not distinct in any way a user can perceive. Both hold an agent definition. Both are spawnable. Neither is presented differently anywhere in the product.

## Why it is confusing in practice

Three separate resolvers walk these directories, in orders that do not agree:

- `resolve_ephemeral_manifest` (`crates/librefang-kernel/src/kernel/messaging.rs:1378`) — `templates/` then `workspaces/agents/`.
- `load_agent_manifest_from_template_dirs` (`crates/librefang-kernel/src/kernel/spawn.rs`) — same pair, same order.
- `resolve_manifest` (`crates/librefang-api/src/routes/agents/lifecycle.rs:57`) — `templates/` then `workspaces/agents/`, but only since a fix landed today; before that it read `workspaces/agents/` alone, so creating an agent from a real template returned 404 and only worked when the name happened to belong to an existing agent.

**None of the three looks in `registry/agents/` at all**, even though that is where the 32 definitions a normal installation actually has are sitting. The registry is synced to disk and then ignored by every resolver.

That is the root of it: `templates/` is empty on a real install, the registry is full and unused, and the resolvers only know about the empty one.

## The question

If agent types are the registry's, does `templates/` still have a purpose?

The options as they look from here:

1. **Fold it in.** Agent types live in one directory; locally-created ones sit alongside synced ones. `templates/` is migrated and retired. Simplest to explain, and there is only one place to look.
2. **Keep it as the local-override layer.** `templates/` shadows the registry, so an operator can adjust a catalogue type without a resync overwriting the edit. Needs the precedence written down and surfaced in the UI, otherwise it is an invisible rule.
[...]

PR #7988, opened 2026-08-29
chore: add changelog fragments and docs for v2 features

DaBlitzStein:
## Summary
- 35 changelog fragments (10 added, 1 changed, 24 fixed) under `changelog.d/`
- 15 docs page updates (1 new architecture doc, 14 updated mdx pages)
- 5 skill definition files (agent-type-author, workflow-author, workflow-creator)

No code changes — documentation and configuration only.

## Test plan
- [ ] No Rust/TS code changes; CI should pass on docs/config linting only
- [ ] Verify changelog fragments render correctly when assembled
- [ ] Spot-check docs pages for broken links or formatting


houko (commented):
Two problems with the fragments here, and the first one is mechanical enough to check by script.

**16 of the 35 fragments carry no trailing `(#N)`, so each will be duplicated in the release body.**

`changelog.d/README.md` is explicit that the trailing group is functional, not decorative: every merged PR in the range gets a generated `- <PR title> (#N) (@author)` line *unless* its number appears in the trailing `(#N)` group of a curated bullet.
A fragment without one therefore does not suppress anything — the release notes get the curated prose *and* the generated title line for the same change.

The 16:

```
added/6931-agent-template-crud.md
added/7721-workflow-required-skills.md
added/7723-ephemeral-mission-folder.md
added/7724-editors-agent-type-fresh.md
added/7731-agent-types-editor.md
added/model-router.md
added/registry-promotion-manifest-sanitizer.md
changed/agent-type-template-naming-homogenization.md
fixed/7775-litellm-model-discovery.md
fixed/agent-types-page-uneditable.md
fixed/fallback-model-default-literal-sent-to-provider.md
fixed/litellm-reasoning-effort-unsupported-param.md
fixed/patch-config-provider-only-silently-dropped.md
fixed/pending-mcp-server-and-skill-declarations-blocked-agent-edits.md
fixed/registry-agent-types-rename-compat.md
fixed/workflow-by-type-refused-spawn-error.md
```

(Attribution is fine — all 35 end with `(@…)`. It is only the PR number that is missing.)

**`fixed/registry-agent-types-rename-compat.md` already has a counterpart on `main`.**

#8000 merged a few minutes ago carrying its own `changelog.d/added/registry-agent-types-rename.md` for the same change.
Both describe `resolve_agent_types_dir` and the canonical-vs-legacy directory fallback, and they land in *different* sections — one under `### Added`, one under `### Fixed` — so the assembled changelog will describe one change twice, in two places, as though they were separate work.
One of the two has to go, and since #8000's is already on `main`, this is the one to drop.

**The structural point, which is the reason the first two happened.**

`CLAUDE.md` puts the fragment in the PR that makes the change, specifically so the number is known when the prose is written and so the bullet cannot describe behaviour that is not on `main` yet.
Collecting 35 of them into one PR inverts that: `fixed/7387-skillhub-marketplace-dead-api.md` belongs to #7983 and `fixed/litellm-reasoning-effort-unsupported-param.md` to #7984, both still open.
If this merges first, `main` claims two fixes it does not have; if those merge first without their fragments, the release notes describe them only if this lands too.

The docs pages and the three skill definitions are a genuinely separate change from the fragments and would be easier to review — and to merge — on their own.
Splitting this into "docs + skills" and letting each fragment ride its own PR is more work up front and removes both failure modes.

houko (changes requested):
Backfilling the fragments is worth doing and the prose quality is genuinely good — these explain *why*, which is what `CLAUDE.md` asks for and what most fragments in this repo do not manage. Three things need fixing before they ship, because every one of these lines goes into the GitHub release body verbatim.

## Blocking

**1. `7774-model-context-window-editable.md` describes a feature that is not on `main`.**

> "A model's context window and max output tokens **can now be edited from the WebUI**, the TUI and the API"
> "the dashboard's slider **was** hardcoded `enabled={false}` waiting for a backend that never arrived"

`ModelsPage.tsx:1505-1511` on `main` today:

```tsx
<SliderInput
  label={t("models.context_window")}
  value={model.context_window ?? 128000}
  onChange={() => {}}
  min={1024} max={1048576} step={1024}
  enabled={false}
```

The slider is still inert, with a no-op `onChange`. #8106 is the open PR that would make the sentence true. The API and CLI halves are real (`ModelOverrides.context_window` exists, `librefang models overrides` is at `cli.rs:1398`), so the fix is to drop the WebUI claim and the past-tense "was hardcoded" until #8106 lands — or hold this fragment until it does.

Worth re-checking the other backfilled fragments against `main` the same way. A retrospective changelog is exactly where "this was in flight when I wrote the note" slips through, and the release body has no review step after this one.

**2. Sixteen fragments carry no `(#N)`, so each gets a duplicate auto-generated line.**

```
changelog.d/added/6931-agent-template-crud.md
changelog.d/added/7721-workflow-required-skills.md
changelog.d/added/7723-ephemeral-mission-folder.md
changelog.d/added/7724-editors-agent-type-fresh.md
changelog.d/added/7731-agent-types-editor.md
changelog.d/added/model-router.md
changelog.d/added/registry-promotion-manifest-sanitizer.md
changelog.d/changed/agent-type-template-naming-homogenization.md
changelog.d/fixed/7775-litellm-model-discovery.md
changelog.d/fixed/agent-types-page-uneditable.md
changelog.d/fixed/fallback-model-default-literal-sent-to-provider.md
changelog.d/fixed/litellm-reasoning-effort-unsupported-param.md
changelog.d/fixed/patch-config-provider-only-silently-dropped.md
changelog.d/fixed/pending-mcp-server-and-skill-declarations-blocked-agent-edits.md
changelog.d/fixed/registry-agent-types-rename-compat.md
changelog.d/fixed/workflow-by-type-refused-spawn-error.md
```

`check-changelog-attribution.py` only enforces the `(@user)` half, which is why CI is green. But per `CLAUDE.md`, generated `- <PR title> (#N) (@author)` lines "fill only the gaps: a PR whose number appears in a curated bullet's trailing `(#N)` group gets no generated line, so it is described once, in the words someone chose." Without the number, each of these sixteen features is described twice in the release notes — once in the curated prose and once as a bare PR title. Add the trailing `(#N)` group to each; several already name the number in the filename, so the information is at hand.

**3. Ten fragments have no number prefix in the filename.**

`model-router.md`, `registry-promotion-manifest-sanitizer.md`, `agent-type-template-naming-homogenization.md`, `agent-types-page-uneditable.md`, `fallback-model-default-literal-sent-to-provider.md`, `litellm-reasoning-effort-unsupported-param.md`, `patch-config-provider-only-silently-dropped.md`, `pending-mcp-server-and-skill-declarations-blocked-agent-edits.md`, `registry-agent-types-rename-compat.md`, `workflow-by-type-refused-spawn-error.md`.

`CLAUDE.md`: "named after the PR or issue number so fragments sort usefully: `changelog.d/fixed/6623-wire-max-content-chars.md`". Twenty-five of the thirty-five here follow that; these ten sort into an unrelated alphabetical block in the assembled section.

## Coordination

Two fragments will be stale the moment an in-flight PR merges. Worth deciding the order now rather than shipping notes that contradict the code:

- `litellm-reasoning-effort-unsupported-param.md` says the driver "remembers the model so it **never** sends the field to that provider+model combination again". #7984 changes that to a 10-minute TTL precisely so operators can reconfigure a gateway without a daemon restart. If #7984 lands first, this sentence is wrong.
- `7782-model-parameter-form-dimmed-switch.md` documents the invariant "its off state uses a token that reads as a control rather than as a hairline" — which is the assertion `SliderInput.test.tsx:86` enforces and which #8106 currently violates by switching to `bg-border-subtle`. Good that it is written down; worth pointing #8106 at it.

## Non-blocking

- Bundling ~20 docs pages and four skill manifests with 35 changelog fragments makes the diff hard to review as one unit. It is a `chore:` housekeeping PR so the bundling is defensible, but a reviewer checking prose accuracy and a reviewer checking the skill manifests are doing unrelated jobs. Splitting the skills half out would let each land on its own merits.

houko:

No commits at all since my review on 2026-09-01T13:58:07Z.

Still open from that review:

- 1. `7774-model-context-window-editable.md` describes a feature that is not on `main`
- 2. Sixteen fragments carry no `(#N)`, so each gets a duplicate auto-generated line
- 3. Ten fragments have no number prefix in the filename

DaBlitzStein:
Addressed at `ee4d51049`, and your point 1 turned out to be much bigger than the one fragment.

**You asked me to re-check the other backfilled fragments the same way. Twelve of the thirty-five described code that is not on `main`** — the fragments were written against the fork's tree. Since every line ships verbatim in the release body with no review step after this one, the PR as it stood would have announced features that do not exist in the repository publishing the notes.

Removed, each checked against a symbol, route or file on `origin/main`:

- `added/model-router.md` — no `model_profiles.toml`, no `profile` / `model_override` params on `agent_spawn`. It was also a `release-note:added` code fence that would have rendered literally.
- `added/7724-editors-agent-type-fresh.md` — issue still open, no per-run fresh checkbox, no `/api/agent-types` alias mounted.
- `fixed/6504-workflow-run-total-steps-persistence.md` — `total_steps` appears nowhere in `crates/`.
- `fixed/6943-save-as-workflow-broken.md`, `fixed/6943-workflow-json-extraction-regex.md` — no Save-as-Workflow button, no `extractWorkflowJson`.
- `fixed/7741-tui-model-routing-noop.md` — no `model_routing` route, no `set_agent_model_routing`.
- `fixed/7743-test-driver-leak.md` — neither named test exists.
- `fixed/agent-types-page-uneditable.md` — the listing is still dual-source with an `editable` flag; `save-as-agent-type` is not a route.

I spot-checked four of those independently before accepting the deletions: `total_steps`, `model_profiles`, `extractWorkflowJson` and `set_agent_model_routing` return nothing on `origin/main`.

**`7774-model-context-window-editable.md`** no longer claims the WebUI half or the past-tense "was hardcoded" — `ModelsPage.tsx` still has `enabled={false}` with a no-op `onChange`. The API and CLI halves are real and stay.

**Point 2 — the `(#N)` groups are in.** Every fragment this PR adds now carries one, so none of them gets a duplicate auto-generated line.

**Point 3 — the ten unprefixed filenames are renamed** with `git mv`, including `litellm-reasoning-effort-unsupported-param.md` → `7852-…`. That one is **renamed, not removed**: the behaviour it describes is real, only its duration changes with #7984, so it is worth keeping and rewording rather than dropping.

`python3 scripts/check-changelog-attribution.py` → `OK: no changelog problems in scope`.

On splitting the skills half out: I would rather not do it in this PR, since the fragments are what needs to land before the next release cuts, but I have no objection if you would prefer it.

Ready for another look.

docs/src/app/configuration/core/page.mdx
```diff
@@ -330,6 +330,58 @@ ffmpeg_timeout_secs = 30
 
 ---
 
+### `[capabilities]`
```
houko: **This `[capabilities]` block does not exist in the codebase.**

`KernelConfig` has no `capabilities` property at all — `crates/librefang-api/tests/fixtures/kernel_config_schema.golden.json` has no such top-level key, and nothing in `crates/` reads one. The per-agent form is dead too: `ManifestCapabilities` (`crates/librefang-types/src/agent.rs:1824-1865`) carries only `network`, `tools`, `memory_read`, `memory_write`, `agent_spawn`, `agent_message`, `shell`, `ofp_discover`, `ofp_connect`; `image_understanding` / `speech_to_text` / `image_generation` / `text_to_speech` / `video_generation` / `music_generation` are not fields there, and serde drops them silently.

So an operator who follows these 52 lines writes config that does nothing: the image still gets replaced by the file-path note, and the documented `agent > global > [media] image_provider > env` resolution never runs. Media routing today is `[media] image_provider` / `image_model` / `audio_provider` / `audio_model` plus `[media.custom_image]`.

Same section is duplicated at `docs/src/app/zh/configuration/core/page.mdx:331`. Either drop both, or land them with the PR that implements the block.

docs/src/app/integrations/api/page.mdx
```diff
@@ -245,9 +245,9 @@ The `Retry-After` header indicates the window duration in seconds.
 | GET | `/api/channels/registry` | Channel metadata from the registry sync |
 | POST | `/api/channels/sidecar/{name}/configure` | Save a `[[sidecar_channels]]` entry (schema-driven) |
 | POST | `/api/channels/reload` | Hot-reload channels from disk |
-| **Templates** | | |
-| GET | `/api/templates` | List templates |
-| GET | `/api/templates/{name}` | Get template |
+| **Agent Types** | | |
+| GET | `/api/agent-types` | List agent types (`/api/templates` is a deprecated alias) |
```
houko: **`/api/agent-types` does not exist**, and this table now presents it as canonical while marking the one working path deprecated.

`crates/librefang-api/src/routes/agent_templates.rs` registers only `/templates`, `/templates/{name}`, `/templates/{name}/toml`, `/templates/{name}/promote`, `/templates/{name}/history`, `/templates/{name}/history/{version_id}/restore`.

And `docs/architecture/agent-types.md:9` says so deliberately:

> The **API path stays `/api/templates`**, and there is deliberately no `/api/agent-types` alias (#7722).

A reader following this table gets 404, and worse, is steered away from the endpoint that answers. Same problem on line 250 and on line 487 (`/api/agent-types/{name}/toml`).

docs/src/app/integrations/cli/commands/page.mdx
```diff
@@ -24,8 +24,8 @@ librefang agent new [<TEMPLATE>]
 
 **Behavior:**
 
-- Templates are discovered from: the repo `agents/` directory (dev builds), `~/.librefang/workspaces/agents/` (assistant pre-installed, others installed from dashboard), and `LIBREFANG_AGENTS_DIR` (env override).
-- Each template is a directory containing an `agent.toml` manifest.
+- Templates are discovered from: `~/.librefang/agent-types/` (pre-installed by the registry sync, plus any created via the dashboard, the `agent_type_create` tool, or `save-as-agent-type`), and `LIBREFANG_AGENTS_DIR` (env override).
```
houko: This rewrite replaces the correct discovery path with one `librefang agent new` never reads.

`cmd_agent_new` (`crates/librefang-cli/src/commands/agent.rs:802`) calls `templates::load_all_templates()`, which walks `discover_template_dirs()` — `<LIBREFANG_HOME>/workspaces/agents` and `LIBREFANG_AGENTS_DIR`, nothing else (`crates/librefang-cli/src/templates.rs:19-45`). Inside each directory it skips every entry that is not a **directory containing `agent.toml`**:

```rust
if !path.is_dir() { continue; }
let manifest = path.join("agent.toml");
if !manifest.exists() { continue; }
```

So both new claims are wrong: `~/.librefang/agent-types/` is not consulted by this command, and a flat `<name>.toml` is filtered out by the `is_dir()` guard. Dropping the `~/.librefang/workspaces/agents/` bullet removes the only location that works.

(If the intent is that `agent new` *should* read `agent-types/`, that is a code change — the docs cannot lead it.) The zh mirror at `docs/src/app/zh/integrations/cli/commands/page.mdx:27` has the same edit.

skills/workflow-author/skill.toml
```diff
@@ -0,0 +1,50 @@
...
+5. **Test with a small real input** before declaring done, and follow the run with `workflow_status`.
+   Fix the workflow and re-run until the run completes; do not hand back an untested workflow.
+
+## The field set `workflow_create` accepts
+
+- `name` (required, 1-64, `[A-Za-z0-9_-]`)
+- `description`
+- `steps` (required, 1-50). Each step:
+  - `name` — unique within the workflow.
+  - `agent_id` (an existing agent) or `type` (an agent type, spawned on demand) — exactly one.
```
houko: These step field names do not match `workflow_create`. The tool schema (`crates/librefang-runtime/src/tool_runner/definitions.rs:1348-1364`) is:

```
"required": ["name", "agent", "prompt_template"]
```

`agent` takes a bare string or an object with exactly one routing key (`{"name": …}` / `{"id": …}` / `{"type": …}`). There is no `agent_id` key and no `prompt` key. An agent that follows this skill emits a call rejected for missing required fields, every time — so its own step 5 ("test with a small real input") can never pass.

Note `skills/workflow-creator/prompt_context.md:26` in this same PR uses the correct `agent` / `prompt_template` spelling; the two skills disagree.

skills/agent-type-author/skill.toml
```diff
@@ -0,0 +1,53 @@
+[skill]
+name = "agent-type-author"
+version = "0.1.0"
+description = "Design, create and test complete agent types on demand with agent_type_create and agent_spawn."
+author = "librefang"
+tags = ["agent-types", "templates", "design", "onboarding"]
+
+prompt_context = """
```
houko, line 1: **This `prompt_context` is silently discarded**, so the skill installs and enables while teaching the agent nothing.

`prompt_context` is a **top-level** field of `SkillManifest` (`crates/librefang-skills/src/lib.rs:220-222`), a sibling of `[skill]` and `[runtime]`. Because this key is written *after* the `[skill]` header, TOML parses it as `skill.prompt_context` — and `SkillMeta` (`lib.rs:266-291`) has no such field and no `deny_unknown_fields`, so serde drops it without a word.

`manifest.prompt_context` is then `None`, and the registry's fallback looks for a companion `prompt_context.md` (`crates/librefang-skills/src/registry.rs:349-366`), which this directory does not ship. Net effect: prompt-only skill with an empty body.

Fix is either to move the assignment above the `[skill]` header, or to split the prose into `skills/agent-type-author/prompt_context.md`.

skills/workflow-author/skill.toml
```diff
@@ -0,0 +1,50 @@
+[skill]
+name = "workflow-author"
+version = "0.1.0"
+description = "Design, create and test multi-step workflows with workflow_create, workflow_describe and workflow_status."
+author = "librefang"
+tags = ["workflows", "pipelines", "automation", "design"]
+
+prompt_context = """
```
houko, line 1: Same silent-drop as `skills/agent-type-author/skill.toml:8`: `prompt_context` sits below the `[skill]` header, so TOML nests it as `skill.prompt_context`, `SkillMeta` has no such field, and serde discards it (`crates/librefang-skills/src/lib.rs:220-222`, `266-291`).

With no `skills/workflow-author/prompt_context.md` companion either (`registry.rs:349-366`), the skill loads with an empty body. A side effect worth noting: `scan_loaded_prompt_context` (`registry.rs:276-282`) never sees this text, so it also never crosses the prompt-injection boundary every other prompt-only skill crosses.

docs/architecture/agent-created-workflows.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Step agent types, fresh instances, and required skills
+
+Steps can reference an **agent type** (a template) instead of a concrete agent:
+
+```json
+{ "type": "researcher" }
+```
+
+Resolution is find-or-spawn: reuse the registered agent with that name, otherwise load the template manifest (`templates/` first, then `workspaces/agents/`) and spawn it with the canonical name-derived UUID. Add `"fresh": true` to spawn a brand-new instance per run (uid name tag, never shadows the canonical one). A missing template fails the step with a ByType-specific error.
```
houko: Two claims here contradict the code.

**1. `"fresh": true` does not exist.** `StepAgent` says so in its own doc comment (`crates/librefang-kernel/src/workflow.rs:363-369`):

> **There is deliberately no `fresh` variant** that would spawn a throwaway instance per run (#7714). To isolate a step from the type's prior context, set `WorkflowStep::session_mode` to `New`…

The object form is validated as carrying exactly one routing key, so the extra key buys nothing and risks the "exactly one of" rejection.

**2. The search order is stale.** `agent_template_candidates` (`crates/librefang-kernel/src/agent_template.rs:141-155`) searches `agent-types/<name>.toml`, then `workspaces/agents/<name>/agent.toml`, then `registry/agents/<name>/agent.toml`. `templates/` is never consulted — which is exactly what `changelog.d/fixed/7758-agent-types-canonical-directory.md` in this same PR describes as the bug that was fixed.

docs/architecture/agent-created-workflows.md
```diff
@@ -0,0 +1,75 @@
...
+- "does not declare" — the skill is absent from the resolved agent's (or template's) skill allowlist;
+- "declares but are not installed" — the skill is in the allowlist but the registry hasn't loaded it (see the pending-declarations surface on the agents API).
+
+## Run ownership and billing
+
+A run records its `owner_agent_id` (migration v48): the caller of `workflow_run` / `workflow_start` / the channel `/workflow run` command. Template-spawned step agents are parented to the owner and their usage records bill to it. Resume and operator-action paths copy the original run's owner.
+
+## Agent types
+
+Agent types are templates: the canonical spelling is "agent type" everywhere (`/api/agent-types` aliases the `/api/templates` routes; the TUI templates screen is titled "Agent types"). Agents can author types with the `agent_type_create` tool (same validation as the API), and ephemeral workers spawned from a type get a uid display name plus a transient mission workspace under `~/.librefang/transient/<name>` that is deleted when the run ends.
```
houko: `/api/agent-types` does not alias anything — the route does not exist. `crates/librefang-api/src/routes/agent_templates.rs` registers only the `/templates*` paths, and `docs/architecture/agent-types.md:9` states the decision outright: "there is deliberately no `/api/agent-types` alias (#7722)".

As written, two docs under `docs/architecture/` now say opposite things about the same URL.

docs/architecture/agent-created-workflows.md
```diff
@@ -0,0 +1,75 @@
...
+## Validation
+
+Two layers, mirroring the HTTP API:
+
+1. **Name validation** (runtime and kernel): `[A-Za-z0-9_-]`, 1–64 chars — path traversal is structurally impossible.
+2. **Semantic validation** (`Workflow::validate()`): same checks as the canvas/API path — operator-node/DAG conflicts, invalid transform templates, empty branch arms, etc.
+
+## Persistence
+
+Workflows are written atomically (tmp file + rename) to `~/.librefang/workflows/<name>.workflow.toml`, then hot-registered in the `WorkflowEngine`.
```
houko: The persistence description does not match the engine. `WorkflowEngine::persist_definition` (`crates/librefang-kernel/src/workflow.rs:2235-2278`) writes exactly one file:

```rust
let path = dir.join(format!("{id}.workflow.json"));
let tmp_path = dir.join(format!("{id}.workflow.json.tmp"));
```

No `<name>.workflow.toml` is produced on the `workflow_create` path (`register_unique_name` → `persist_definition` is the only write), so there is also no "both files scanned at boot and deduped by UUID" behaviour to describe. An operator told to look for a name-keyed TOML after `workflow_create` will not find one.

docs/architecture/agent-created-workflows.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Agent types
+
+Agent types are templates: the canonical spelling is "agent type" everywhere (`/api/agent-types` aliases the `/api/templates` routes; the TUI templates screen is titled "Agent types"). Agents can author types with the `agent_type_create` tool (same validation as the API), and ephemeral workers spawned from a type get a uid display name plus a transient mission workspace under `~/.librefang/transient/<name>` that is deleted when the run ends.
+
+An agent type carries the full manifest surface relevant to spawns: `name`, `description`, `system_prompt`, `provider`/`model` (spawn defaults), `tools`, `skills`, a `channels` allowlist (empty = all configured channels), and preferred-model tiers via `routing` (`simple_model` / `medium_model` / `complex_model` + thresholds — the `[routing]` block of the generated TOML). The dashboard editor exposes all of them with catalog-backed pickers (skills, tools, channels) mirroring the agent editor; the shared `agent_type_json_to_toml` conversion in `librefang-types` is the single source of truth for the API route and the tool, so the two authoring surfaces cannot drift.
```
houko: `agent_type_json_to_toml` does not exist — `grep -rn agent_type_json_to_toml crates/` returns nothing. The agent-type helpers in `librefang-types` are `agent_type_store::{agent_types_dir, agent_types_dir_in, agent_type_path, agent_type_path_in}` and `agent_type::agent_type_spec_of`; the shared shape the API route and the tool go through is `AgentTypeSpec` / `AgentTypeSpec::apply_to`.

Naming a function that isn't there sends the next maintainer looking for a single source of truth they cannot find, and risks the conclusion that no shared boundary exists.

changelog.d/fixed/7758-agent-types-canonical-directory.md
```diff
@@ -0,0 +1,6 @@
+Agent types (the reusable manifests behind `POST /api/agent-types`, the `agent_type_create` tool, `save-as-agent-type`, and the registry sync) now read and write `~/.librefang/agent-types/` — a directory that exists for exactly this purpose — instead of two directories that were never meant to hold them.
```
houko: `POST /api/agent-types` does not exist — the write verbs are at `POST /api/templates` (`crates/librefang-api/src/routes/agent_templates.rs:26`), and `docs/architecture/agent-types.md:9` says there is deliberately no `agent-types` alias (#7722).

This matters more than the usual typo because per `changelog.d/README.md` the fragment body lands **verbatim in the GitHub release body**, so the published notes would hand users a 404 path.

It also contradicts `changelog.d/changed/7859-agent-type-naming-homogenization.md` in this same PR, which explains the opposite:

> The HTTP paths themselves are unchanged and stay `/api/templates`…

skills/workflow-author/skill.toml
```diff
@@ -0,0 +1,50 @@
...
+## The field set `workflow_create` accepts
+
+- `name` (required, 1-64, `[A-Za-z0-9_-]`)
+- `description`
+- `steps` (required, 1-50). Each step:
+  - `name` — unique within the workflow.
+  - `agent_id` (an existing agent) or `type` (an agent type, spawned on demand) — exactly one.
+  - `prompt` — the step's task; reference previous outputs with `{{step_name}}` placeholders.
+  - `timeout_secs` — per-step wall-clock ceiling, at most 3600.
+  - `error_mode` — how the step's failure affects the run (`fail`, `skip`, `retry`).
```
houko: `retry` is not a bare string value. `ErrorMode` (`crates/librefang-kernel/src/workflow.rs:1198-1222`) is externally tagged and its `Retry` variant has a required `max_retries`, so the wire form is `{"retry": {"max_retries": 3}}` — which is what the tool schema documents (`definitions.rs:1357`).

A `workflow_create` call carrying `"error_mode": "retry"` fails deserialization and takes the whole workflow with it. `skills/workflow-creator/prompt_context.md:49` has the same bare `retry` in its error-handling table.

skills/workflow-creator/prompt_context.md
```diff
@@ -0,0 +1,155 @@
...
+- **`depends_on`**: List step names this step depends on. When set, the workflow engine uses DAG (directed acyclic graph) execution instead of sequential.
+- **`output_var`**: Store this step's output as a named variable. Later steps can reference it with `{{var_name}}`.
+
+### Input parameters
+
+Declare `input_schema` so callers know what to pass to `workflow_run`:
+
+```json
+[
+  { "name": "repo_url", "type": "string", "description": "Repository URL to analyze", "required": true },
```
houko: The parameter type key is `param_type`, not `type`. `WorkflowInputParam` (`crates/librefang-kernel/src/workflow.rs:219-239`) declares `param_type` with `#[serde(default = "default_input_param_type")]` and no `deny_unknown_fields`, so `{"name": "max_depth", "type": "number"}` deserializes to `param_type: "string"` and the `type` key is dropped.

The tool schema warns about exactly this: *"Note the key is param_type, not type — this is the same spelling workflow_describe reports back."* The failure is silent, and a `file` / `image` parameter declared this way loses its artifact-reference semantics.

Affects lines 62-63, 101 and 133 here, plus the mirrored examples in `skills/workflow-creator/SKILL.md`.

docs/src/app/configuration/core/page.mdx
DaBlitzStein: Removed in 3a7fff6b8.
The `[capabilities]` media-routing block described six keys — `image_understanding`, `speech_to_text`, `image_generation`, `text_to_speech`, `video_generation`, `music_generation` — and none of them exists in `KernelConfig` or in `ManifestCapabilities` (crates/librefang-types/src/agent.rs), whose capability set is `network, tools, memory_read, memory_write, agent_spawn, agent_message, shell, ofp_discover, ofp_connect`.
The real routing surface is `[media]` with `image_provider` / `audio_provider` / `video_provider` (crates/librefang-types/src/media.rs:174), which has neither the provider/model shorthand nor the per-agent override the block promised.
Deleted from the English page and its `zh/` mirror.

docs/src/app/integrations/api/page.mdx
DaBlitzStein: Fixed in 1b26e0130.
The table now states `/api/templates*` as canonical with no `/api/agent-types` alias.
`agent_templates.rs` registers only the `/templates*` paths, and `openapi.json` has six `templates` paths and zero `agent-types` paths, so the alias never existed.

docs/src/app/integrations/cli/commands/page.mdx
DaBlitzStein: Fixed in 3a7fff6b8.
The discovery path is back to `~/.librefang/workspaces/agents/`, which is what `discover_template_dirs` actually scans (crates/librefang-cli/src/templates.rs:19-45).
Applied to the `zh/` mirror as well.

skills/workflow-author/skill.toml
DaBlitzStein: Fixed in 3a02f085c.
`agent_id` / `type` / `prompt` collapse to the single routing key `agent` plus `prompt_template`, matching the `workflow_create` schema in crates/librefang-runtime/src/tool_runner/definitions.rs:1348-1364.
`error_mode` is now the tagged form `{"retry": {"max_retries": N}}` rather than the bare `"retry"`, matching `ErrorMode::Retry`.

skills/agent-type-author/skill.toml
DaBlitzStein: Fixed in 3a02f085c.
`prompt_context` moved above `[skill]`, so it deserializes as a top-level key instead of `skill.prompt_context`.
`SkillMeta` does not carry `deny_unknown_fields`, which is why the nested form was dropped in silence rather than failing to parse — the shipped skill simply had no prompt context.

skills/workflow-author/skill.toml
DaBlitzStein: Fixed in 3a02f085c.
Same silent-nesting bug as the sibling skill, corrected the same way: `prompt_context` is now top-level, verified by loading the file with `tomllib`.

docs/architecture/agent-created-workflows.md
DaBlitzStein: Fixed in 1b26e0130.
The fictional `"fresh": true` is gone; the field that exists is `WorkflowStep::session_mode = New` (crates/librefang-kernel/src/workflow.rs:363-369).
The search order now reads `agent-types/ → workspaces/agents/ → registry/agents/`, which is the order `agent_template_candidates` walks (crates/librefang-kernel/src/agent_template.rs:141-155).

docs/architecture/agent-created-workflows.md
DaBlitzStein: Fixed in 1b26e0130.
The claim that an `/api/agent-types` alias exists is removed, so the page no longer contradicts `docs/architecture/agent-types.md:9`.

docs/architecture/agent-created-workflows.md
DaBlitzStein: Fixed in 1b26e0130.
The persistence section now describes what `persist_definition` does: a single `<id>.workflow.json` written atomically through a temp file and a rename (crates/librefang-kernel/src/workflow.rs:2235-2278).
The TOML file and the second copy it described do not exist.

docs/architecture/agent-created-workflows.md
DaBlitzStein: Fixed in 1b26e0130.
The fictional `agent_type_json_to_toml` helper is replaced by the real type, `AgentTypeSpec` and its `apply_to` (crates/librefang-types/src/agent_type.rs).

changelog.d/fixed/7758-agent-types-canonical-directory.md
DaBlitzStein: Fixed in 1b26e0130.
The opening sentence now says `POST /api/templates`, so the fragment no longer contradicts its sibling `7859-agent-type-naming-homogenization.md`.

skills/workflow-author/skill.toml
DaBlitzStein: Fixed in 3a02f085c.
The `error_mode` row now shows `{"retry": {"max_retries": N}}` instead of the bare `retry`, in the skill manifest and in both `skills/workflow-creator/prompt_context.md` and its `SKILL.md` mirror.

skills/workflow-creator/prompt_context.md
DaBlitzStein: Fixed in 3a02f085c.
Every `input_schema` example now uses `param_type`, matching `WorkflowInputParam` (crates/librefang-kernel/src/workflow.rs:220-232), with a note making the distinction explicit so the next reader does not reintroduce `type`.
Applied to both `prompt_context.md` and the `SKILL.md` mirror, which stay byte-identical.

docs/src/app/integrations/cli/commands/page.mdx
```diff
@@ -24,8 +24,8 @@ librefang agent new [<TEMPLATE>]
 
 **Behavior:**
 
-- Templates are discovered from: the repo `agents/` directory (dev builds), `~/.librefang/workspaces/agents/` (assistant pre-installed, others installed from dashboard), and `LIBREFANG_AGENTS_DIR` (env override).
-- Each template is a directory containing an `agent.toml` manifest.
+- Templates are discovered from: `~/.librefang/workspaces/agents/` (pre-installed by the registry sync, plus any created via the dashboard, the `agent_type_create` tool, or `save-as-agent-type`), and `LIBREFANG_AGENTS_DIR` (env override).
+- Each template is a `<name>.toml` manifest file.
```
houko: **The other half of the earlier `agent new` finding is still unfixed.** The directory is back to `~/.librefang/workspaces/agents/`, but this line survived, and `load_all_templates` refuses exactly this layout (`crates/librefang-cli/src/templates.rs:53-67`):

```rust
if !path.is_dir() { continue; }
let manifest = path.join("agent.toml");
if !manifest.exists() { continue; }
```

A flat `<name>.toml` is filtered out by the `is_dir()` guard before the extension is ever looked at. An operator who follows this line drops `researcher.toml` into `workspaces/agents/` and `librefang agent new` lists nothing.

The same edit is on `docs/src/app/zh/integrations/cli/commands/page.mdx:28`.

```suggestion
- Each template is a directory containing an `agent.toml` manifest.
```

docs/src/app/integrations/cli/commands/page.mdx
```diff
@@ -24,8 +24,8 @@ librefang agent new [<TEMPLATE>]
 
 **Behavior:**
 
-- Templates are discovered from: the repo `agents/` directory (dev builds), `~/.librefang/workspaces/agents/` (assistant pre-installed, others installed from dashboard), and `LIBREFANG_AGENTS_DIR` (env override).
-- Each template is a directory containing an `agent.toml` manifest.
+- Templates are discovered from: `~/.librefang/workspaces/agents/` (pre-installed by the registry sync, plus any created via the dashboard, the `agent_type_create` tool, or `save-as-agent-type`), and `LIBREFANG_AGENTS_DIR` (env override).
```
houko: The new provenance list names three surfaces that do not write into this directory.

`agent_type_create` and the dashboard both go through `librefang_types::agent_type_store`, whose path is `agent_types_dir_in(home) = home.join("agent-types")` plus `{name}.toml` (`crates/librefang-types/src/agent_type_store.rs:40-51`). `changelog.d/fixed/7758-agent-types-canonical-directory.md` in this same PR is the fragment that moved them there, explicitly so that `workspaces/agents/` "only ever contains agents you actually deployed".

`discover_template_dirs` (`crates/librefang-cli/src/templates.rs:19-45`) never looks at `agent-types/`, so a type authored through any of the three surfaces named here is invisible to `librefang agent new`. As written this line promises the opposite of what #7758 did.

skills/agent-type-author/skill.toml
```diff
@@ -0,0 +1,53 @@
...
+
+Everything else is silently ignored by the converter — do not invent fields that are not in this list.
+
+- `name` (required)
+- `description` — one or two sentences.
+- `system_prompt` — the persona and operating rules.
+- `provider` / `model` — omit both to inherit the daemon default, or set both.
+- `tools` — tool names from the tool registry (`file_read`, `web_search`, `memory_store`, ...).
+- `skills` — skill names from the skill registry.
+- `channels` — channel names this type participates in (rarely needed).
```
houko: **`channels` and `routing` are not accepted, and the failure is loud rather than silent — it takes the whole call with it.**

`AgentTypeSpec` (`crates/librefang-types/src/agent_type.rs:33-57`) carries exactly seven fields — `name`, `description`, `system_prompt`, `provider`, `model`, `tools`, `skills` — and is declared `#[serde(deny_unknown_fields)]`. `tool_agent_type_create` deserializes the raw tool input straight into it (`crates/librefang-runtime/src/tool_runner/agent.rs:525-530`), and its own doc comment says what happens next:

> `AgentTypeSpec` is `deny_unknown_fields`, so a key the model invented — `temperature`, `max_tokens`, `memory` — is refused by name

The tool schema agrees: "Only the seven fields below are settable here" (`definitions.rs:1385-1402`).

So an agent that follows this list emits `agent_type_create` with `channels` / `routing` and gets `InvalidParameter { name: "spec" }` — nothing is created at all, not even the six valid fields. Line 24's framing ("Everything else is silently ignored by the converter") is the exact inverse of the behaviour, which makes it the worst possible advice here: it tells the model the extra key is harmless.

Drop both bullets and move them to the "Not expressible today" paragraph below, which is where they belong.

skills/agent-type-author/skill.toml
```diff
@@ -0,0 +1,53 @@
...
+## Workflow
+
+1. **Clarify before creating.** If the request does not already say so, ask (or infer from context): what is this type FOR, which tools and skills does it need, and which model should it run on.
+   Do not ask when the answer is obvious — a summariser needs file_read and nothing else.
+2. **Design the identity.** Write a system prompt that defines the persona, its boundaries, and its output style.
+   Be specific about what it refuses to do and what format it answers in.
+3. **Validate the name.** 1-64 characters, only `[A-Za-z0-9_-]`, lowercase-hyphen convention (`code-reviewer`, not `Code Reviewer`).
+   Creation fails with a clear error on a bad name, a duplicate type, or a collision with an existing workspace agent — read the error and fix the name rather than retrying.
+4. **Create** with the `agent_type_create` tool.
+5. **Test before declaring done.** Spawn the type once with `agent_spawn` (`agent_type` = the name you created) and give it a small representative task.
```
houko: This test step cannot run as written — the same failure mode houko flagged on `workflow-author`'s step 5 last round.

`agent_type` is **ephemeral-only**: `tool_agent_spawn` reads it nowhere, and the only assignment is inside `tool_agent_spawn_ephemeral` (`crates/librefang-runtime/src/tool_runner/agent.rs:403`). The schema says so too: `"agent_type"` → *"Ephemeral only: name of an existing agent type…"*, and `"message"` → *"Ephemeral only, and required for it"* (`definitions.rs:381-384`).

Without `ephemeral: true` the call falls through to the permanent branch (`agent.rs:288-298`), which ignores `agent_type` entirely and requires `system_prompt`. An agent following this literally gets `MissingParameter("system_prompt")`; if it then supplies one to get past the error, it has created a *permanent* agent whose name shadows the type it was trying to test.

Spell the shape out: `agent_spawn` with `ephemeral = true`, `agent_type = <name>`, and the task in `message`.

skills/workflow-author/skill.toml
```diff
@@ -0,0 +1,50 @@
...
+   Fix the workflow and re-run until the run completes; do not hand back an untested workflow.
+
+## The field set `workflow_create` accepts
+
+- `name` (required, 1-64, `[A-Za-z0-9_-]`)
+- `description`
+- `steps` (required, 1-50). Each step:
+  - `name` — unique within the workflow.
+  - `agent` — a bare string (agent name) or an object with exactly one routing key: `{name}`, `{id}`, or `{type}` (find-or-spawn).
+  - `prompt_template` — the step's task; reference previous outputs with `{{step_name}}` placeholders.
```
houko: Both of these describe an `output_var` default that does not exist.

`WorkflowStep::output_var` is `Option<String>` with `#[serde(default)]` (`crates/librefang-kernel/src/workflow.rs:268-269`), so an omitted `output_var` is `None`, and every one of the fifteen binding sites in the engine is guarded by `if let Some(ref var) = step.output_var` (`workflow.rs:2897`, `4504`, `4620`, `5823`, `5992`, `6291`, …). Nothing anywhere inserts the step name as a fallback — the one place that does bind a step name (`workflow.rs:7108-7109`) is the *workflow-template instantiation* path, which sets `output_var: Some(ts.name.clone())` explicitly and is not reached by `workflow_create`.

The tool schema is precise about it: `{{input}}` interpolates the previous step's output, `{{var}}` interpolates "a workflow input parameter **or an earlier step's `output_var`**" (`definitions.rs:1351`).

So a workflow written from this skill — steps with no `output_var`, later prompts referencing `{{step_name}}` — silently sends unresolved placeholders to the model. The sibling `workflow-creator/prompt_context.md` was fixed last round by adding explicit `output_var` to every referenced step; this skill still teaches the premise that made that example wrong.

skills/workflow-creator/prompt_context.md
```diff
@@ -0,0 +1,157 @@
...
+  - `{{param}}` for workflow-level input parameters
+
+### Execution modes
+
+| Mode | Use when |
+|------|----------|
+| `sequential` (default) | Steps run one after another |
+| `fan_out` | One step spawns multiple parallel instances |
+| `collect` | Gather results from fan_out into a single output |
+| `conditional` | Branch based on a condition in the previous output |
```
houko: **Same bug class as the `error_mode: "retry"` row that was fixed one table below, in the table directly above it.**

`StepMode` (`crates/librefang-kernel/src/workflow.rs:488-490`) is externally tagged with `rename_all = "snake_case"`, and the four modes listed here are all data-carrying variants:

```rust
Conditional { condition: String },
Loop { max_iterations: u32, until: String },
Wait { duration_secs: u64 },
Approval { recipients: Vec<String>, timeout_secs: Option<u64> },
```

Only `Sequential`, `FanOut` and `Collect` are unit variants, so only those three deserialize from a bare string. `"mode": "conditional"` fails with `invalid type: string`, and because `mode` sits inside the `steps` array that aborts the whole `workflow_create` call — the same way the bare `"retry"` did.

The tool schema draws the line explicitly: *"the string `"sequential"` (default), `"fan_out"` …, or `"collect"` …. Richer nodes take a tagged object, e.g. `{"conditional": {"condition": "APPROVED"}}`"* (`definitions.rs:1356`).

Give the last four rows their object forms, and note that `Approval`'s `recipients` is required. `skills/workflow-creator/SKILL.md:38-41` has the identical table.

skills/workflow-creator/SKILL.md
```diff
@@ -0,0 +1,154 @@
...
+      "name": "run-tests",
+      "agent": "test-runner",
+      "prompt_template": "Run the test suite on {{input}}. Report failures.",
+      "depends_on": ["style-check"],
+      "error_mode": "fail"
+    },
+    {
+      "name": "summarize",
+      "agent": "assistant",
+      "prompt_template": "Summarize the review results. Style check: {{style-check}}. Tests: {{run-tests}}. Provide a verdict: approved or changes requested.",
```
houko: **This file did not get the `output_var` fix, so the two copies now disagree.** The reply on the earlier thread says they "stay byte-identical"; `diff SKILL.md prompt_context.md` shows three hunks — the two `output_var` additions on `style-check` / `run-tests`, the `{{style_check}}` / `{{test_results}}` rewrite here, and the `output_var: "search"` on the briefing example's first step.

As it stands this example is the bug the sibling was fixed for: no step declares `output_var`, so `{{style-check}}` and `{{run-tests}}` bind to nothing (`workflow.rs:2897` and the fourteen other guarded `variables.insert` sites), and the hyphenated spelling was replaced with underscores in the fix precisely because these are variable names, not step names.

Worth deciding whether the file should exist at all: `SkillRegistry::load_all` only auto-converts a `SKILL.md` when `skill.toml` is **absent** (`crates/librefang-skills/src/registry.rs:179-182`), and `skills/workflow-creator/skill.toml` is present. So nothing loads this file — it is a second copy that can only ever drift, which is what happened here. Deleting it removes the drift surface; keeping it means re-applying every prompt_context fix twice, forever.

docs/architecture/agent-created-workflows.md
```diff
@@ -0,0 +1,74 @@
+# Agent-Created Workflows
+
+Agents can create workflows during a conversation via the `workflow_create` tool.
+This document covers the design, validation, persistence, and observability of that path.
+
+## Tool surface
+
+`workflow_create` is registered in `ALWAYS_NATIVE_TOOLS`, so every agent has it without manifest configuration.
```
houko: `workflow_create` is not in `ALWAYS_NATIVE_TOOLS`. The array is `crates/librefang-runtime/src/tool_runner/definitions.rs:138-171` and its 26 entries are `tool_load`, `tool_search`, the `memory_*` group, `web_search`, `web_fetch`, the `file_*` group, `code_search`, `agent_send`, `agent_list`, `channel_send`, `notify_owner`, `read_artifact` and the `skill_evolve_*` group. Neither `WORKFLOW_CREATE` nor `WORKFLOW_RUN` appears; `grep -rn WORKFLOW_CREATE crates/ --include='*.rs'` finds only the constant, the definition, the dispatch arm and tests.

That list is the lazy-load shortlist (`select_native_tools`, same file), so the practical consequence is the opposite of what this sentence promises: an agent with `lazy_tools = true` does **not** get `workflow_create` natively and has to find it through `tool_search` / `tool_load`. It reaches an ordinary agent because it is in the full catalogue, not because it is always-native — `tools_and_skills.rs:2331` and `:2356` show it present by default and removable via `tool_blocklist`.

The same claim is in three changelog fragments in this PR (`added/6943-workflow-create.md:2`, `fixed/6943-workflow-create-caller-attribution.md:2`, `fixed/6943-workflow-create-resource-caps.md:2`), where it ships verbatim into the release body.

changelog.d/fixed/6943-workflow-create-caller-attribution.md
```diff
@@ -0,0 +1,4 @@
+`workflow_create` now records the calling agent instead of discarding it.
+The `caller_agent_id` parameter was threaded all the way from the tool call through `KernelHandle::create_workflow`'s trait signature into the kernel implementation, then ignored there — leaving zero audit trail for a tool that sits in `ALWAYS_NATIVE_TOOLS`, so every agent has it without configuration, and whose sibling `workflow_run` (also always-native) can immediately execute whatever `steps[].agent` the new workflow names.
```
houko: `ALWAYS_NATIVE_TOOLS` does not contain `workflow_create` or `workflow_run`. The array is `crates/librefang-runtime/src/tool_runner/definitions.rs:138-171`; `grep -rn WORKFLOW_CREATE crates/ --include='*.rs'` returns the constant, the tool definition, the dispatch arm and tests, and nothing else.

This matters more than the usual naming slip because the sentence is load-bearing for the security argument — "so every agent has it without configuration … whose sibling `workflow_run` (also always-native) can immediately execute whatever `steps[].agent` the new workflow names". The reach is real (both tools are in the full catalogue by default — `tools_and_skills.rs:2331`), but it comes from the catalogue, not from that array, and an agent with `lazy_tools = true` gets neither natively. Per `changelog.d/README.md` this body lands in the GitHub release verbatim, so it would publish a claim a reader can falsify with one grep.

Same wording in `changelog.d/added/6943-workflow-create.md:2` and `changelog.d/fixed/6943-workflow-create-resource-caps.md:2`.

docs/architecture/agent-created-workflows.md
```diff
@@ -0,0 +1,74 @@
...
+A failed write never leaves a partial file on disk.
+
+## Budget and recursion guards
+
+- **Budget attribution**: the ephemeral worker's usage is billed against the spawning agent's `agent_id`, so the cost appears in `GET /api/budget/agents/{parent}`.
+- **Recursion depth**: `current_agent_depth() >= max_agent_call_depth` is checked before the loop runs (same quota as `agent_send` and workflow-step dispatch), and the loop future is wrapped in `with_agent_call_depth` (which also boxes the ~56 KB future — the #6659 stack-overflow hazard).
+
+## Observability
+
+Workflow creation logs `workflow=<name> registered_id=<uuid>` at info.
```
houko: Neither field name exists, so this is not greppable. The log site is `crates/librefang-kernel/src/kernel/handles/workflow_runner.rs:686-695`:

```rust
tracing::info!(
    workflow_id = %id,
    workflow_name = %name,
    caller_agent_id = caller_agent_id.unwrap_or("<unattributed>"),
    owner = ...,
    step_count = summary.step_count,
    "Agent created a workflow"
);
```

An operator told to look for `workflow=` / `registered_id=` finds nothing, and misses the two fields that are actually the point of that line — `caller_agent_id` and `owner`, added by #7744 / `changelog.d/fixed/6943-workflow-create-caller-attribution.md` in this same PR.

docs/architecture/agent-created-workflows.md
```diff
@@ -0,0 +1,74 @@
...
+
+Agent types are templates: the canonical spelling is "agent type" everywhere, in prose and UI, though the API stays at `/api/templates` (there is deliberately no `/api/agent-types` alias, #7722; the TUI templates screen is titled "Agent types"). Agents can author types with the `agent_type_create` tool (same validation as the API), and ephemeral workers spawned from a type get a uid display name plus a transient mission workspace under `~/.librefang/transient/<name>` that is deleted when the run ends.
+
+An agent type carries the full manifest surface relevant to spawns: `name`, `description`, `system_prompt`, `provider`/`model` (spawn defaults), `tools`, `skills`, a `channels` allowlist (empty = all configured channels), and preferred-model tiers via `routing` (`simple_model` / `medium_model` / `complex_model` + thresholds — the `[routing]` block of the generated TOML). The dashboard editor exposes all of them with catalog-backed pickers (skills, tools, channels) mirroring the agent editor; the shared `AgentTypeSpec` / `AgentTypeSpec::apply_to` in `librefang-types` is the single source of truth for the API route and the tool, so the two authoring surfaces cannot drift.
```
houko: **`channels` and `routing` are not part of the agent-type surface, and this paragraph is self-contradicting because of it.**

`AgentTypeSpec` (`crates/librefang-types/src/agent_type.rs:33-57`) has seven fields — `name`, `description`, `system_prompt`, `provider`, `model`, `tools`, `skills` — and is `#[serde(deny_unknown_fields)]`. `agent_type_spec_of` (`agent_type.rs:209-219`) projects exactly those seven, and `agent_type_detail`'s own doc comment says so: *"`spec` is the flat editor projection — exactly the seven keys a `PUT` accepts back"* (`crates/librefang-api/src/routes/agent_templates.rs:421-423`).

So each half of the sentence falsifies the other: if `AgentTypeSpec` really is the single source of truth for both surfaces, then `channels` and `routing` are not settable on either — and `deny_unknown_fields` means a client that sends them gets a 422 rather than a partial save.

The dashboard claim is wrong independently: `AgentTypesPage.tsx:37-51` declares the form as `{description, system_prompt, provider, model, tools, skills}`, and `grep -n 'channels\|routing\|simple_model' AgentTypesPage.tsx` is empty — there is no channels picker to mirror the agent editor with.

`changelog.d/added/7731-agent-types-editor.md` in this same PR gets it right ("catalog-backed pickers for skills and tools"). Trim the sentence to the seven fields, or land it with the PR that widens the spec.

changelog.d/fixed/7740-agent-type-save-destroys-fields.md
```diff
@@ -0,0 +1,9 @@
...
+Because the fields came back `undefined`, the form rendered them empty, and saving sent `channels: []` and `routing: undefined` straight back — so the channel allowlist and model-tier routing a user configured were wiped out on the very next save.
+Second, and far worse, `PUT /api/templates/{name}` rebuilt the entire manifest from that same flat, 9-key JSON shape and overwrote the file on disk with `std::fs::write`.
+Every field outside those 9 keys — `[compaction]`, `max_history_messages`, `[[triggers]]`, `[resources]`, `[autonomous]`, `mcp_servers`, `tool_allowlist`, `session_mode`, `workspaces`, and everything else in `AgentManifest` — was silently reset to its default on every single save made through the WebUI editor, with the endpoint still returning 200.
+Fixed both: the flattening function now includes `channels` and `routing` in its output, and the update route reads the existing manifest off disk first, then applies only the fields the request body actually supplies through `AgentTypeSpec::apply_to`, leaving every other field untouched.
```
houko: "the flattening function now includes `channels` and `routing` in its output" is not true of the code this PR sits on top of, and this fragment ships verbatim in the release body.

The flattening function is `agent_type_spec_of` (`crates/librefang-types/src/agent_type.rs:209-219`); it emits seven keys and neither of those two is among them. `AgentTypeSpec` is `#[serde(deny_unknown_fields)]` (`agent_type.rs:33-35`), so `channels` / `routing` in a `PUT` body is now a 422, and `agent_type_detail`'s doc comment states the shape outright: *"exactly the seven keys a `PUT` accepts back"* (`crates/librefang-api/src/routes/agent_templates.rs:421-423`).

The lines above have the same problem: line 3 says "the dashboard's `AgentType` interface and its edit form both expect them", but `AgentTypesPage.tsx:37-51` is `{description, system_prompt, provider, model, tools, skills}` with no mention of either; line 5's "flat, 9-key JSON shape" is seven keys today.

The second half of the fix — read-then-`apply_to` instead of rebuild-from-scratch — is real and is the valuable part. Please drop the `channels` / `routing` half rather than publishing it.

docs/src/app/integrations/cli/commands/page.mdx
DaBlitzStein: Confirmed and fixed in 746f12b52, `docs/src/app/integrations/cli/commands/page.mdx:28`.

The line now reads "Each template is a directory containing an `agent.toml` manifest", and I added the reason directly underneath so the next editor does not undo it: `load_all_templates` filters on `path.is_dir()` before it ever looks at a file name (`crates/librefang-cli/src/templates.rs:53-67`), so a flat `<name>.toml` is skipped and the picker lists nothing.

Applied to the `zh/` mirror at the same line.

docs/src/app/integrations/cli/commands/page.mdx
DaBlitzStein: Confirmed and fixed in 746f12b52, `docs/src/app/integrations/cli/commands/page.mdx:27` and `:30`.

Line 27 no longer claims those three surfaces write here — it is now just the registry sync plus `LIBREFANG_AGENTS_DIR`, which is what `discover_template_dirs` actually walks (`crates/librefang-cli/src/templates.rs:19-45`).

Rather than delete the provenance outright I inverted it into the fact a reader needs, because the old line would otherwise just be re-added by someone who authored a type and could not find it. Line 30 now says those types land in `~/.librefang/agent-types/<name>.toml` (#7758) and that this picker does not scan that directory, and that the daemon *does* resolve them — `agent_type_path_in` is the first candidate `agent_template_candidates` tries (`crates/librefang-kernel/src/agent_template.rs:142-155`), so `agent_spawn` and a workflow step's `{"type": ...}` find them even though `agent new` does not.

`zh/` mirror updated to match.

skills/agent-type-author/skill.toml
DaBlitzStein: Confirmed and fixed in 746f12b52, `skills/agent-type-author/skill.toml`.

Both bullets are gone from the accepted-field list, and `channels` / tier routing now head the "Not expressible today" paragraph at line 37.

The framing on line 24 was the more dangerous half, so I did not just delete it — line 27 now states the opposite explicitly: the tool deserializes into `AgentTypeSpec`, which is `deny_unknown_fields` (`crates/librefang-types/src/agent_type.rs:33-35`), so one invented key fails the whole call with `InvalidParameter { name: "spec" }` and nothing is created, not even the valid fields. Verified the struct carries exactly the seven fields you list and that `agent_type_spec_of` (`agent_type.rs:209-219`) projects the same seven.

skills/agent-type-author/skill.toml
DaBlitzStein: Confirmed and fixed in 746f12b52, `skills/agent-type-author/skill.toml:17-20`.

Step 5 now names all three parts of the call — `ephemeral = true`, `agent_type` = the created name, and the task in `message` — and spells out the failure it replaces: without `ephemeral` the call takes the permanent branch, which ignores `agent_type`, demands `system_prompt`, and if the model supplies one to clear the error it creates a permanent agent shadowing the type it was testing.

Verified `agent_type` is assigned only inside the ephemeral path (`crates/librefang-runtime/src/tool_runner/agent.rs:403`) and appears nowhere else in that file.

skills/workflow-author/skill.toml
DaBlitzStein: Confirmed and fixed in 746f12b52, `skills/workflow-author/skill.toml:26-27` and `:30`.

The `output_var` row no longer claims a default — it now says there is none, and that omitting it leaves the step's output unreachable by name. The `prompt_template` row above it was the other half of the same premise, so it now distinguishes `{{input}}` (previous step's output) from `{{name}}` (a workflow input parameter or an earlier step's declared `output_var`), with an explicit line that a step name is not a placeholder.

Verified `output_var: Option<String>` with `#[serde(default)]` (`crates/librefang-kernel/src/workflow.rs:268-269`), 14 guarded `if let Some(ref var) = step.output_var` binding sites, and that the only step-name binding is the workflow-template instantiation path at `workflow.rs:7108-7109`, which `workflow_create` does not reach.

skills/workflow-creator/prompt_context.md
DaBlitzStein: Confirmed and fixed in 746f12b52, `skills/workflow-creator/prompt_context.md:33-43`.

The four data-carrying rows now show their object forms — `{"conditional": {"condition": "APPROVED"}}`, `{"loop": {"max_iterations": 3, "until": "DONE"}}`, `{"wait": {"duration_secs": 60}}`, `{"approval": {"recipients": ["ops"]}}` — with `recipients` noted as required and `timeout_secs` as optional, matching `Approval { recipients: Vec<String>, timeout_secs: Option<u64> }` (`crates/librefang-kernel/src/workflow.rs:537-541`). The three unit variants are quoted as strings so the distinction is visible in the table itself, and a line above it states why a bare `"conditional"` takes the whole `workflow_create` call down.

The `SKILL.md` copy you mention is deleted rather than updated — see the thread on that file.

skills/workflow-creator/SKILL.md
DaBlitzStein: Confirmed, and deleted in 746f12b52.

First: my earlier reply on the `param_type` thread said the two files "stay byte-identical". That was wrong — I applied that one fix to both files and asserted the general property without running the diff. Your three hunks are exactly right, and the `output_var` round had already broken it.

Rather than re-sync a file nothing reads, I removed `skills/workflow-creator/SKILL.md`. `SkillRegistry::load_all` only auto-converts a `SKILL.md` when `skill.toml` is absent (`crates/librefang-skills/src/registry.rs:179-182`) and `skills/workflow-creator/skill.toml` is present, so the file was never loaded. It is also the only directory under `skills/` carrying both, and it was added by this PR rather than inherited — so deleting it costs nothing and closes the drift surface permanently instead of committing every future prompt_context fix to being applied twice.

docs/architecture/agent-created-workflows.md
DaBlitzStein: Confirmed and fixed in 746f12b52, `docs/architecture/agent-created-workflows.md:8-9`.

Verified against the tree: `ALWAYS_NATIVE_TOOLS` (`crates/librefang-runtime/src/tool_runner/definitions.rs:138-171`) contains neither, and `grep -rn WORKFLOW_CREATE crates/ --include='*.rs'` returns only the constant and the tool definition.

The doc now says `workflow_create` is part of the default tool catalogue — so an ordinary agent has it without configuration and a `tool_blocklist` entry is what removes it — and states explicitly that it is *not* in `ALWAYS_NATIVE_TOOLS`, with the `lazy_tools = true` consequence you describe, since that is the case a reader of this section would get wrong.

The three 6943 fragments are corrected in the same commit.

One more instance of the same claim turned up while checking: `docs/src/app/agent/templates/page.mdx:327` gave `file_read, shell_exec, the cron tools` as the always-native examples. `file_read` is on the list; `shell_exec` and the cron tools are not. That paragraph is outside this PR's hunks, but it is the same error in a file the PR already edits, so I corrected it in the same commit rather than leave the twin standing — the list is now enumerated from the array, and the two non-members are called out. `zh/` mirror updated too.

changelog.d/fixed/6943-workflow-create-caller-attribution.md
DaBlitzStein: Confirmed and fixed in 746f12b52, `changelog.d/fixed/6943-workflow-create-caller-attribution.md:2`.

Both mentions are rewritten to "ships in the default catalogue" / "also in the default catalogue". The security argument is unchanged in substance — the reach is real, as you say — but it is now attributed to the catalogue rather than to an array that would falsify the sentence with one grep, which matters because the body ships into the release verbatim.

Same wording corrected in `changelog.d/added/6943-workflow-create.md:2` and `changelog.d/fixed/6943-workflow-create-resource-caps.md:2` in the same commit.

docs/architecture/agent-created-workflows.md
DaBlitzStein: Confirmed and fixed in 746f12b52, `docs/architecture/agent-created-workflows.md:32-33`.

The section now names the real fields — `workflow_id`, `workflow_name`, `caller_agent_id`, `owner`, `step_count` — and the message `Agent created a workflow`, so the line is greppable as written. A second sentence calls out `caller_agent_id` and `owner` as the two the log line exists for (#7744), since those are the point of the entry and the old text pointed away from them.

Verified against the log site at `crates/librefang-kernel/src/kernel/handles/workflow_runner.rs:687-695`.

docs/architecture/agent-created-workflows.md
DaBlitzStein: Confirmed and fixed in 746f12b52, `docs/architecture/agent-created-workflows.md:76-79`.

Trimmed to the seven fields, as you suggest — widening the spec is a different PR. The paragraph now states that the seven are what `AgentTypeSpec` carries, what `agent_type_spec_of` projects, and (because of `deny_unknown_fields`) the complete set a `PUT` accepts back, with anything else being a 422 rather than a partial save. The dashboard sentence now says "catalog-backed pickers for skills and tools", matching `AgentTypesPage.tsx:37-51` and agreeing with `changelog.d/added/7731-agent-types-editor.md`.

`channels` and the `[routing]` tiers moved to a closing sentence about fields that start at their default on a type and are edited on the manifest afterwards, which `apply_to` leaves untouched — the reader still learns where they live, without the doc claiming they are settable here.

Verified: `agent_type.rs:33-57` (seven fields, `deny_unknown_fields`), `agent_type.rs:209-219` (`agent_type_spec_of` emits the same seven), and `grep -n 'channels\|routing\|simple_model' AgentTypesPage.tsx` returns nothing.

changelog.d/fixed/7740-agent-type-save-destroys-fields.md
DaBlitzStein: Confirmed and fixed in 746f12b52 — the `channels` / `routing` half is dropped, not reworded.

Lines 2-4 (the "two bugs compounded" framing and the flattening-function bug) and the flattening clause in line 7 are gone. What remains is the half you identify as the real one: `PUT /api/templates/{name}` rebuilt the whole manifest from the flat JSON and overwrote the file with `std::fs::write`, and now reads the manifest off disk first and applies only supplied fields through `AgentTypeSpec::apply_to`. Line 5's "flat, 9-key JSON shape" is now "the flat JSON shape the editor sends", and the field-loss line says "outside the seven keys that shape carries".

The unparseable-manifest refusal and the create-path note are unchanged.

`python3 scripts/check-changelog-attribution.py` passes; the `(#7859) (@DaBlitzStein)` tail is still on the final line.

skills/workflow-creator/prompt_context.md
DaBlitzStein: Correcting my own reply above: the closing sentence, "Applied to both `prompt_context.md` and the `SKILL.md` mirror, which stay byte-identical", was not true.

The `param_type` fix itself did land in both files, but the byte-identical claim was an assertion I did not verify — and it was already false when I wrote it, because the `output_var` fix from the same round had gone into `prompt_context.md` only. houko's diff on `SKILL.md:96` shows the three hunks.

Resolved in 746f12b52 by deleting `skills/workflow-creator/SKILL.md` rather than re-syncing it: `SkillRegistry::load_all` only auto-converts a `SKILL.md` when `skill.toml` is absent (`crates/librefang-skills/src/registry.rs:179-182`), and `skill.toml` is present here, so nothing ever loaded that copy. Details in the thread on that file.

DaBlitzStein:
Second review round addressed in 746f12b52 — @houko, ready for another look.

All 12 open findings verified against the tree before editing, and all 12 were real. Each is answered in its own thread with the file:line of the fix; summary:

**False `ALWAYS_NATIVE_TOOLS` claim (4 places).** Neither `workflow_create` nor `workflow_run` is in that array (`definitions.rs:138-171`). Corrected in `docs/architecture/agent-created-workflows.md:8-9` and in the three 6943 fragments, which ship into the release body verbatim. Reach is now attributed to the default catalogue, with the `lazy_tools = true` consequence spelled out.

**`channels` / `routing` on agent types (3 places).** `AgentTypeSpec` is seven fields and `deny_unknown_fields`, so those keys are not settable and an extra key fails the whole call rather than being ignored. Removed from `skills/agent-type-author/skill.toml`, from the architecture doc, and from `changelog.d/fixed/7740-agent-type-save-destroys-fields.md` — where the false half is dropped rather than reworded, keeping the read-then-`apply_to` fix that is the real one.

**Instructions that cannot execute (3).** `agent new` discovers directories containing `agent.toml`, never a flat `<name>.toml`, and never scans `agent-types/`; `agent_type` on `agent_spawn` is read on the ephemeral path only; `output_var` has no step-name default. Fixed in the CLI docs (+ `zh/` mirror) and both skill manifests.

**`StepMode` table.** The four data-carrying variants now show tagged-object forms; a bare `"conditional"` aborts the whole call, same failure class as the earlier bare `"retry"`.

**`skills/workflow-creator/SKILL.md` deleted.** `SkillRegistry::load_all` only auto-converts a `SKILL.md` when `skill.toml` is absent (`registry.rs:179-182`), and it is present here — so nothing ever loaded that copy, and it had already drifted in three hunks. It was added by this PR rather than inherited, so removing it closes the drift surface at no cost.

**Correction to an earlier reply of mine.** On the `param_type` thread I claimed the two workflow-creator copies "stay byte-identical". That was an unverified assertion and was already false when written. Corrected in that thread.

**One fix outside the review scope.** `docs/src/app/agent/templates/page.mdx:327` carried the same false always-native claim (`shell_exec`, the cron tools). Untouched paragraph, but same error in a file this PR already edits, so it is corrected here rather than left standing next to its fixed twin. `zh/` mirror too.

Verification: text-only change — docs, skill prompt manifests and changelog fragments; no `.rs` / `.ts` touched, so no build cycle applies. `python3 scripts/check-changelog-attribution.py` passes, all three `skill.toml` files still parse with a top-level `prompt_context`, and the edited `.mdx` hunks introduce no unescaped JSX braces.

Nothing deferred.

DaBlitzStein:
Every one of the 25 review threads on this PR now has an inline reply, and the branch is at `746f12b52`.

houko (approved):
LGTM. Both findings are fixed the correct way — by moving the key, not by adding a companion file to work around it.

`prompt_context` is now line 1 of both manifests, with `[skill]` at line 51 (`agent-type-author`) and line 46 (`workflow-author`). That puts it at the top level of `SkillManifest` where `lib.rs:220-222` declares it, rather than nesting into `skill.prompt_context` where `SkillMeta` has no field for it and — with no `deny_unknown_fields` — serde would drop it in silence. Both skills now install with the body they were written to carry.

Worth noting the second-order effect the earlier comment flagged, which this also fixes: `scan_loaded_prompt_context` (`registry.rs:276-282`) only ever saw an empty string before, so the prompt-injection scan was passing vacuously on text that never reached it. It now scans the real body.

All checks green; `BLOCKED` is the stale `CHANGES_REQUESTED` this approval clears.

Issue #7957, opened 2026-08-28
Bug: a gateway model's vision support is guessed from its name, so images are silently dropped


DaBlitzStein:
## Summary

For an OpenAI-compatible gateway (litellm, a self-hosted vLLM / llama.cpp proxy under its own provider id), a discovered model's `supports_vision` is decided by a **heuristic on the model's name**, not by anything the gateway said. When the operator's alias does not match that heuristic, a vision-capable model is recorded as blind and the agent loop strips images out of the request without telling anyone.

Operators of these gateways name models whatever they like — `team-default`, `fast`, `vision-v2`, an internal ticket id. The heuristic cannot know.

## The chain, on current `main`

1. **Discovery produces a bare entry.** `crates/librefang-runtime/src/provider_health.rs` — the OpenAI `/v1/models` shape parser keeps only the two numeric limits:

   ```rust
   Some(DiscoveredModelInfo {
       context_window: crate::model_metadata::parse_openai_model(m),
       max_output_tokens: crate::model_metadata::parse_openai_model_max_output(m),
       ..DiscoveredModelInfo::bare(name)
   })
   ```

   `DiscoveredModelInfo::bare` leaves `capabilities` empty. There is no field on `DiscoveredModelInfo` for an OpenAI-style `supports_vision` / `supports_function_calling` flag at all — the `capabilities: Vec<String>` field is documented as Ollama's `/api/tags` array.

2. **The merge falls back to a name heuristic.** `crates/librefang-runtime/src/model_catalog.rs`, `resolve_discovered_capabilities`:

   ```rust
   if capabilities.is_empty() {
       return infer_capabilities(name, families);
   }
   ```

   For a gateway model `capabilities` is always empty, so `supports_vision` is whatever `infer_capabilities` guesses from the string.

3. **A false guess strips the images.** `crates/librefang-runtime/src/agent_loop/mod.rs:1265` (and the same block in `run_streaming.rs:785`):

   ```rust
   let supports_vision = kernel.as_ref().map(|k| k.supports_vision_for(&api_model)).unwrap_or(true);
   let request_messages = if supports_vision { messages.clone() } else { redact_images_for_text_only(messages.clone(), &api_model) };
   ```

   Note the asymmetry: a missing kernel handle or a catalog **miss** fails open (`unwrap_or(true)`), which is the right call. A catalog **hit** carrying a wrong `false` does not — and that is the case this bug produces.

[...]

Issue #7823, opened 2026-08-23
feat(media): MediaCapability models only generation, so no provider can declare transcription or image understanding


DaBlitzStein:
## Problem

`MediaCapability` (`crates/librefang-types/src/media.rs`) enumerates only *generation*:

```rust
pub enum MediaCapability {
    ImageGeneration,
    TextToSpeech,
    VideoGeneration,
    MusicGeneration,
}
```

Understanding — speech-to-text and image description — has no variant, even though it is the same kind of fact about a provider: it either has the capability or it does not.

The consequence is that the two halves of media are wired completely differently.

Generation goes through the driver registry: a `MediaDriver` declares `capabilities()`, and `MediaDriverRegistry::detect_for_capability` (`crates/librefang-runtime-media/src/lib.rs`) walks the configured providers and returns the first configured one that declares the capability asked for.
That is discovery with a fallback, and it is data-driven.

Understanding cannot use any of that, because there is no capability value to ask for.
Instead `crates/librefang-runtime-media/src/media_understanding.rs` carries two hand-written env-var cascades:

```rust
fn detect_vision_provider() -> Option<&'static str> {
    if has_key("ANTHROPIC_API_KEY") { return Some("anthropic"); }
    if has_key("OPENAI_API_KEY")    { return Some("openai"); }
    if has_key("GROQ_API_KEY")      { return Some("groq"); }
    if has_key("GEMINI_API_KEY") || has_key("GOOGLE_API_KEY") { return Some("gemini"); }
    None
}

fn detect_audio_provider() -> Option<&'static str> { /* nine providers, same shape */ }
```

## Why this matters

- **A provider that transcribes cannot say so.** Adding STT or vision support to a provider means editing a hardcoded `if` ladder in `media_understanding.rs`, not declaring a capability where every other provider fact lives.
- **There is no fallback.** `detect_for_capability` tries every configured provider in turn; the cascade picks one by env-var presence and stops. A key that is present but whose provider is failing is not routed around.
- **The selection is invisible to the catalog.** Nothing that reads capabilities can answer "which of my providers can describe an image?", because the answer is not represented anywhere — which is the underlying shape of the rot #6538 described, where a retired Groq vision model degraded silently.
- **It blocks capability routing.** The interesting case — an agent whose own model cannot see an image, routed to a provider that can — needs "who has `ImageUnderstanding`?" to be an answerable question. Today it is not.

## Proposal

[...]

PR #7989, opened 2026-08-29
feat(media): route inbound media by capability instead of dropping it

DaBlitzStein:
## Summary
- Add `CapabilityTarget` and `CapabilityRouting` types to `librefang-types` with custom serde for `"provider/model"` string and `{ provider, model }` table TOML forms
- Add `IMAGE_DESCRIPTION_PREFIX`/`IMAGE_DESCRIPTION_UNAVAILABLE` shared constants for inbound image description blocks
- Implement capability-based media routing in `librefang-runtime` agent loop: when a text-only model receives an image, the configured vision provider describes it and inserts the description before the image block, so the redaction pass preserves context instead of leaving the model to hallucinate from a file path
- Wire the `[capabilities]` block into `KernelConfig` (global) and `ManifestCapabilities` (per-agent, flattened via serde), with per-field merge so an agent override inherits the global provider when only model is set
- Add config reload and boot-time integration for capability routing
- Add TUI Capabilities tab with inline editing of provider/model per capability
- Add dashboard form integration for the capability routing fields
- Add utoipa OpenAPI annotations for all media endpoints
- Derive `Ord`/`PartialOrd` on `MediaCapability` for deterministic prompt ordering (`BTreeMap` keys)

## Test plan
- [x] `cargo check -p librefang-types --lib` passes
- [x] `cargo check -p librefang-runtime-media --lib` passes
- [x] `cargo check -p librefang-runtime --lib` passes
- [x] `cargo check -p librefang-api --lib` passes
- [x] `cargo clippy` clean (only pre-existing unrelated error in `agent.rs:3243`)
- [x] `cargo test -p librefang-types --lib -- media::tests::capability` — 4 tests pass
- [x] `cargo test -p librefang-runtime-media --lib -- detect_for_capability` — 2 tests pass
- [x] `cargo test -p librefang-runtime-media --lib -- set_capability_routing` — 1 test passes
- [x] `cargo test -p librefang-runtime --lib -- media_routing` — 11 tests pass
- [x] TOML string form `"provider/model"` round-trips correctly
- [x] TOML table form `{ provider, model }` round-trips correctly
- [x] `vision` and `transcription` aliases deserialize to correct capabilities
- [x] Per-field merge: agent override inherits global provider when only model is set
- [x] Nominated provider wins over registry preference order
- [x] Mistaken nomination (provider lacks capability) degrades to preference order
- [x] Image already described by channel bridge is not described again
- [x] Provider failure degrades to `[Image description unavailable]`, not a crash
- [x] Vision-capable model is never charged for a description (cost guard)

## Interactions with upstream media work
- This branch touches `crates/librefang-runtime-media/src/media_understanding.rs`, which #8105 rewrote 186 lines of. `describe_video` is untouched by this branch and byte-identical to `origin/main`, and the refresh merge adds zero lines to that file — #8105's "error, not a fabricated description" guarantee for video survives verbatim rather than reconstructed.

## Issue references

- Refs #7823: Route inbound media using declared understanding capabilities.
- Refs #8011: Related media configuration work; this PR covers inbound capability routing, not the complete Models-tab configuration request.


houko (commented):

```
```

```
cargo xtask codegen --openapi
cargo xtask schema-check gen
```

houko (changes requested):
Well built — 22 new tests, an `ImageDescriber` trait so the routing decision is testable without a real engine, and the config-reload classification done properly (both `build_reload_plan` and the drift-guarded table in `docs/operations/config-reload.md`, which is the half most PRs forget). Two things block, one of which is not really yours but lands on this PR.

## Blocking

**1. `cargo-deny (advisories)` is red on two Wasmtime vulnerabilities, and this is the only open PR that runs it.**

```
error[vulnerability]: Guest controlled-size host heap allocation through WASIp3 streams
  ID: RUSTSEC-2026-0268
error[vulnerability]: Filesystem sandbox escape when paths or symlinks contain trailing slashes
  ID: RUSTSEC-2026-0269
```

`Cargo.lock` here pins `wasmtime 47.0.4`, and this PR's only lock change is adding `toml` to `librefang-runtime-media` — so it did not introduce these. The job appears to be gated on `Cargo.lock` changing, which is why no other open PR shows it.

That said, RUSTSEC-2026-0269 is a **filesystem sandbox escape** in the crate that backs the plugin WASM sandbox, so it is not something to wave through because the gate happened to fire here. Before merging, either bump `wasmtime` to a patched release in this PR (a one-line lock change, and the lock is already touched), or open an issue, add a scoped `[advisories] ignore` entry in `deny.toml` with that issue number and an expiry, and link it from the PR. Silently re-running until the job is skipped is the outcome to avoid.

Separately, the same job log starts with an infrastructure error worth reporting on its own:

```
error: override toolchain '1.94.1-x86_64-unknown-linux-musl' is not installed:
the toolchain file at '/github/workspace/rust-toolchain.toml' specifies an uninstalled toolchain
```

**2. No changelog fragment.**

`changelog.d/` is untouched. Routing inbound media by capability instead of dropping it is a behaviour change operators will notice — an agent that previously received nothing for an image now gets a description, or an error. Per `CLAUDE.md` this text goes into the release body verbatim, so it wants the *why*: `changelog.d/added/7989-....md` (or `changed/`, since it alters what an existing agent receives).

The PR body should also state the interaction with #8105 (`fix(media): return an error instead of a fake video description`), which landed on `main` five commits ago and touches the same decision point — a reader needs to know whether the new `RoutingDecision` path preserves that "error, not a fabricated description" guarantee for video.

## Non-blocking

- `Ok(_) => { … }` on an empty description and `Err(reason) => { … }` are handled separately, which is right. Worth one test asserting the empty-but-`Ok` case does not silently inject a blank text block — the arm exists, so it presumably came from a real observation, and it is the kind of thing a later refactor collapses back into `Ok(d) =>`.
- `is_image_description_text` sniffing existing blocks to avoid double-describing is a heuristic on rendered prose. If the marker it looks for is one this code also writes, a comment saying so (and pointing at the writer) would keep the two ends from drifting.
- `Security` / `CI Gate` are additionally red on the pre-existing `cargo audit` pair (`lru` 0.18.0 RUSTSEC-2026-0253, `rand` 0.7.3 RUSTSEC-2026-0097) shared with most open PRs — distinct from the Wasmtime advisories above, and not yours.

DaBlitzStein:
All review points addressed; branch merged with `origin/main` and pushed — the PR is now `MERGEABLE` (head `9989894aa`).

**Wasmtime advisories (blocking 1).** The red run was against the previously pushed head, whose `Cargo.lock` still pinned `wasmtime 47.0.3` (the version `RUSTSEC-2026-0269` flags). The branch now carries the merge with `main`, whose lock pins `wasmtime 47.0.4` — inside the patched range (`>= 47.0.4`) for both `RUSTSEC-2026-0268` and `RUSTSEC-2026-0269` (verified against the advisory DB's `patched` lists). `cargo deny check advisories` passes on the merged tree (`advisories ok`), so no `[advisories] ignore` entry is needed and no new issue was opened.

**Changelog fragment (blocking 2).** `changelog.d/changed/7989-media-capability-routing.md` explains the why (what an existing agent starts receiving, and how to get the old behaviour back with `[media] image_description = false`).

**#8105 interaction.** Stated in the PR body: the capability-routing enricher touches image blocks only (`ImageRef`); video attachments keep flowing through `process_attachments` -> `describe_video`, which after #8105 returns an honest `Err` — the new routing path preserves the "error, not a fabricated description" guarantee.

**Non-blocking items.** Both landed in the follow-up commit: `an_empty_ok_description_does_not_inject_a_blank_text_block` pins the empty-but-`Ok` arm (media_routing suite is now 12 tests), and `is_image_description_text`'s doc-comment now names the writer it reads (`image_description_block_text` plus the unavailable marker) so the dedupe ends cannot drift.

**Verification (fresh on the merged tree):**
- `cargo deny check advisories` — `advisories ok`
- `cargo check -p librefang-types --lib`, `-p librefang-runtime-media --lib`, `-p librefang-runtime --lib`, `-p librefang-api --lib` — all clean
- `cargo test -p librefang-types --lib -- media::tests::capability` — 4 passed
- `cargo test -p librefang-runtime --lib -- media_routing` — 12 passed
- `cargo test -p librefang-runtime-media --lib -- detect_for_capability` — 2 passed
- `cargo test -p librefang-runtime-media --lib -- set_capability_routing` — 1 passed

The pre-existing `lru`/`rand` audit pair and the CI toolchain-not-installed infrastructure error are unchanged and shared with other open PRs.

houko (dismissed):
Blocking #2 is fixed, both non-blocking items too, and blocking #1 was my mistake. Correcting that first, because I told you a live sandbox escape was sitting in your branch.

## #1 was wrong: 47.0.4 is the fix, not the vulnerable version

I wrote that `Cargo.lock` "pins `wasmtime 47.0.4`" and that RUSTSEC-2026-0268 / 0269 were therefore live here. 47.0.4 is the **patched** release. `main` carries #8101's own security fragment saying so:

> RUSTSEC-2026-0269 (CVSS 8.8) … and RUSTSEC-2026-0268 (CVSS 6.9) … both were published 2026-08-20 and are fixed in 47.0.4.

So the red I saw was a stale advisory database, not an exposure, and the remedy I asked for — bump or add a scoped ignore — was the wrong ask. `cargo-deny (advisories)` is now green, and I checked it is green for the right reason rather than by suppression: `wasmtime` is still `47.0.4` on both `main` and this head, and `deny.toml`'s `[advisories] ignore` list contains no entry for either ID. Nothing was waved through.

The toolchain error in that job log (`override toolchain '1.94.1-x86_64-unknown-linux-musl' is not installed`) is also gone.

## #2 — changelog fragment

Present, under `changelog.d/changed/` rather than `added/`, which is the right section since this alters what an existing agent receives for an inbound image. It explains the why instead of restating the title.

## The #8105 interaction — I verified it rather than asking again

I asked for a note in the PR body and there isn't one, but the substantive question is answered, so I checked it myself instead of holding the PR over prose. This branch does touch `crates/librefang-runtime-media/src/media_understanding.rs`, the file #8105 rewrote 186 lines of, so the risk was real. `describe_video` is **byte-identical** between `origin/main` and this head — diffed directly, no differences — so #8105's "error, not a fabricated description" guarantee for video survived the merge verbatim rather than being reconstructed. Worth a sentence in the body anyway for whoever reads this later, but it is not a blocker.

## Non-blocking items, both done

The empty-but-`Ok` arm now has its own test, so a later refactor cannot collapse it back into the success path and hand the model a description block with nothing in it. And `is_image_description_text` now points at the writers of the marker it sniffs — `image_description_block_text` plus the unavailable marker `agent_loop::media_routing` emits on failure — which is what keeps the read and write ends of that heuristic from drifting.

The baseline handling deserves a note: recomputing `xtask/baselines/openapi.sha256` from the merged `openapi.json` rather than taking either parent's hash is exactly right, and saying so in the commit message means the next person resolving that file knows it is not a pick-a-side conflict.

`CI Gate` green with zero non-success checks.

This is `CONFLICTING / DIRTY` against `main` because I merged a batch while reviewing — same as several others in this sweep. Regenerate and recompute the OpenAPI pair again after the rebase, the way you did here.

Approving.

DaBlitzStein:
Refreshed onto current `origin/main` (`6bb613a56`) as merge `ae49550f3` (parents `9989894aa` + `6bb613a56`), with the OpenAPI pair regenerated the way you asked: `xtask/baselines/openapi.sha256` recomputed from the merged `openapi.json` (`d2f75b05…`), which carries both sides — this PR's 7 media paths plus the template-history paths.

One real drift beyond the auto-merge, fixed: #7957 renamed `supports_vision_for` → `vision_support_for`, now returning the tri-state `VisionSupport`. The image gate in `media_routing.rs:272-276` consumes it exactly as upstream's own call site does (`agent_loop/mod.rs:1328-1334`): `.unwrap_or(VisionSupport::Unknown).allows_images()`. The truth table is unchanged from the pre-rename `unwrap_or(true)` — `allows_images()` is `!Unsupported` (`model_catalog.rs:539-541`), so Unknown fails open: a missing handle or a catalog miss keeps images flowing, and only a declared `Unsupported` strips them. The `!has_images ||` short-circuit is untouched, and zero references to `supports_vision_for` remain in `crates/` or `sdk/`.

On the #8105 interaction: the refresh merge adds zero lines to `crates/librefang-runtime-media/src/media_understanding.rs`, so the `describe_video` you diffed on the previous head is untouched and your byte-identical verification carries to `ae49550f3` — the "error, not a fabricated description" guarantee survives the merge verbatim rather than reconstructed. Added the sentence to the PR body as suggested.

The yml deltas in the range are purely action-SHA bumps (buildx bb05f3f→37fe631, codeql ff2f1c6→cdf488f v4.37.7→v4.37.9, setup-java b6effb0→dd06d9c v5.7.0→v6.0.0) — no logic edits.

Verification: `cargo check -p librefang-api --all-targets` clean (0 warnings); `cargo test -p librefang-api --test config_schema_overlay` 4/4; `cargo test -p librefang-runtime-media` 116/116 — set-union of both parents, zero test functions lost from either side; dashboard `tsc --noEmit` clean; all five locales parse with i18n parity unchanged from the known upstream drift. Changelog set-union complete; CHANGELOG.md untouched.

houko (changes requested):
Re-reading at head `ae49550f`. The capability-routing work itself holds up — I re-checked the parts most likely to rot and they are fine (details at the end). But reading the diff rather than the summary, two changes in it have nothing to do with media routing, and one of them writes a raw provider error into user-visible history.

## The provider-failure note puts an unscrubbed error string into the chat and into the model's own history

`run_streaming.rs`, the new `else` branch on a non-timeout provider failure:

```rust
let note = format!(
    "[System: the model provider failed and the task could not \
     be completed ({}). No response was produced.]",
    err_str
);
session.push_message(Message::assistant(note));
```

with `let err_str = e.to_string();` at `:888` — the driver error's `Display`, verbatim, no scrubbing. Two problems, both from that one line:

1. **It is shown to the operator and replayed to the model.** Provider errors carry endpoint URLs, model ids, upstream response bodies and occasionally request echoes depending on the driver. This project already treats that as something not to hand outward — `ApiErrorResponse::internal_scrub` exists for exactly this at the HTTP boundary, and the sibling code you wrote two hundred lines up deliberately keeps the raw provider error in the log and shows the model an opaque marker (`IMAGE_DESCRIPTION_UNAVAILABLE`, *"Deliberately opaque — the raw provider error is logged, not shown to the model"*). The same judgement applies here and reaches the opposite conclusion in the same PR.
2. **It is stored as an assistant message.** On the next turn the model reads *"the model provider failed and the task could not be completed"* as its own prior output, in its own voice. The failure is a fact about the system, not something the assistant said.

The motivation is right and the failure it fixes is real — a turn that dies leaving the chat blank is the worst outcome, and I would not want the note removed. Log `err_str` at `warn!` (it already is, one branch over), keep the visible note to the opaque half, and give it a role that does not impersonate the model.

## Two unrelated changes ride along, in neither the body nor the changelog

The nine summary bullets and the single `changelog.d/changed/` fragment are entirely about capability routing. The diff also contains:

- **`agent_loop/mod.rs`** — a new `memory.save_session_async(session)` that persists the inbound message *before the first LLM call*. A deliberate change to persistence timing, correctly guarded for fork and incognito, with a good reason written above it. It is not media routing.
- **`run_streaming.rs`** — the provider-failure note above, which adds a message the operator will see in their chat.

Both are worth having and neither is described anywhere a reviewer or a release reader would look. The second especially: a new `[System: …]` message appearing in chats is exactly the kind of change an operator will notice and go looking for an explanation of, and `cargo xtask collect-fragments` would put the fragment in front of them. Either split them out, or add them to the body and give the provider-error note its own fragment — this PR already spans 37 files, and the two hitchhikers are the parts nobody was told to review.

## Re-checked and fine

- `describe_images_for_text_only_model` runs on `guarded_user_content_blocks` — the inbound turn only, not the message list — so the whole-message dedupe in `blocks_already_describe_images` cannot be tripped by a description block sitting in history and suppress a genuinely new image. This was my main worry with a whole-message granularity rule and it does not apply.
- `routing_decision` is a pure function with the cost guard first-class: `model_supports_vision` short-circuits before `engine_available`, so a vision model is never billed for a description.
- The base64 size estimate `(data.len() * 3) / 4` over-counts padding rather than under-counting it, so `MediaAttachment::validate`'s ceiling cannot be slipped past by a padded payload.
- `is_image_description_text` sniffing `"[Image description: "` is user-forgeable in principle — a caption starting with that literal suppresses description of the sender's own image — but the blast radius is the sender's own turn against their own agent, so it is not worth defending against.

## CI

`CI Gate` is `failure` on this head, and it is not a test failure: `Test / Unit (lib+bin)` was `cancelled` in run `33639163246`, and `CI Gate` fails on `cancelled` without re-evaluating. Same shape as #7781. It needs a re-run, and the branch is `CONFLICTING` again anyway — regenerate and recompute the OpenAPI pair the way you did last time.

DaBlitzStein:
Pushed `c3e9a58b2`: merged current `main` (six commits behind, conflicting).

A single conflict, in `xtask/baselines/openapi.sha256`. Regenerated rather than resolved by choosing a side — it is a plain `sha256sum` of `openapi.json`, which had already auto-merged into something matching neither line. New value `352649770b9d79f6fdb9b96b85134f40d1c58008a4c747d33288b1533ffd1993`, against a merged spec carrying 358 paths.

- `cargo check -p librefang-types -p librefang-kernel -p librefang-runtime -p librefang-runtime-media -p librefang-channels -p librefang-api --all-targets` — exit 0. The log had been truncated, so rather than trust a bare "Finished", a `compile_error!("SENTINEL")` was appended to `librefang-api/src/lib.rs` to confirm the compiler was really seeing this tree; cargo reported it, and the file was restored (`git diff` empty). A green that arrives without proof the compiler read your tree is not evidence.
- `cargo nextest run -p librefang-runtime-media -p librefang-runtime -p librefang-channels -E 'test(~media) | test(~routing)'` — **129/129**, including `set_capability_routing_ignores_entries_that_name_no_provider`, `test_load_providers_from_registry_filters_by_media_capabilities`, `poisoned_media_state_locks_recover_and_hot_reload_still_applies`, and the three `routing::tests` resolve/validate cases.
- `cargo nextest run -p librefang-api -E 'binary(config_schema_overlay)'` — **4/4**: `every_x_sections_struct_field_exists_on_kernel_config`, `every_root_level_field_exists_on_kernel_config`, `every_kernel_config_struct_field_is_exposed_via_overlay`, `every_x_ui_options_path_resolves_in_schema`. This is the suite that covers the `kernel_config_schema.golden.json` fixture the merge auto-resolved, so it is the one that would have caught a bad auto-merge there.
- `npx tsc --noEmit` clean; `npx vitest run src/lib/agentManifest.test.ts` — 63 tests; `i18n-parity` failing set identical to `main`, nothing added.

houko:
Correcting the CI advice in my last review, and it explains what you saw.

I wrote that the red `CI Gate` was a cancellation of the same shape as #7781's and that "it needs a re-run". You then pushed `1e2deaff` — *"chore: re-run CI after the Unit job was cancelled mid-queue"* — and it went green. That looked like the advice working. It was luck.

**`Test / Unit (lib+bin)` declares `timeout-minutes: 20` and had grown past it.** Durations, including your re-run:

| commit | outcome | duration |
|---|---|---|
| `2acf37f4` (main) | success | 18.7 min |
| **`1e2deaff` (your re-run)** | **success** | **19.7 min** |
| `4e0f042a` (main) | killed | 20.1 min |
| `73fc1656` (main) | killed | 20.1 min |
| `6749058f` (main) | killed | 20.3 min |

Your re-run passed with 18 seconds of headroom. A timeout-killed job reports as `cancelled`, and `CI Gate` fails on `cancelled` without re-evaluating — so "re-run it" is advice that works about half the time for reasons unrelated to the branch, which is worse than advice that never works, because it hides the cause.

**#8159 raises the lane to 35 minutes**, so this stops being a lottery.

The two review findings are unchanged and still what I am blocking on: the raw `err_str` going into `Message::assistant` and therefore into both the visible chat and the model's own history, and the two unrelated changes (inbound-message persistence, the provider-failure note) that are in neither the PR body nor the changelog.

DaBlitzStein:
Addressed at `8b9854f71`, both points.

**The provider error no longer reaches the visible note.** `PROVIDER_FAILURE_NOTE` is a constant with no interpolation, the raw `err_str` goes to a `warn!` at the push site and nowhere else, and the message is pushed as `Role::System` rather than `Role::Assistant` — the failure is a fact about the daemon, and an assistant-role note is replayed to the model on the next turn in its own voice.

While tracing it I found the leak was wider than the review describes: `build_user_facing_llm_error` appends the raw string verbatim for the `Format` category, which is also where an unrecognised provider error lands by default. So it was not only the non-timeout branch.

`streaming_provider_failure_note_is_opaque_and_not_assistant_voiced` pins both halves. Its fixture carries a deliberately incriminating error, so reverting the fix shows exactly what was escaping:

```
assertion `left == right` failed: The provider-failure note is a system fact, not the
model's own prior output; an assistant role replays it to the model in its own voice.
Got: "[System: …] (LLM driver error: Request failed:
      https://internal-gw.example/v1/messages model=acme-secret-v3
      body={"detail":"leaked"} — raw: API error (400): …)"
  left: Assistant
 right: System
```

Restored, it passes. Run by exact name: `cargo test -p librefang-runtime --lib -- --exact agent_loop::tests::utilities::streaming_provider_failure_note_is_opaque_and_not_assistant_voiced` → `1 passed`.

**The two hitchhikers are now described.** `changelog.d/fixed/7989-provider-failure-note.md` covers the visible-behaviour change, since a new `[System: …]` message appearing in chats is exactly what an operator goes looking for an explanation of. The `save_session_async` persistence-timing change is called out in the PR body rather than left for a reviewer to find in the diff.

Ready for another look.

crates/librefang-runtime/src/agent_loop/run_streaming.rs
```diff
@@ -903,6 +920,37 @@ async fn run_agent_loop_streaming_inner(
...
+                    //
+                    // The note is `Role::System`, not `Role::Assistant`, because the failure is a fact about the daemon and an assistant-role note is replayed on the next turn as the model's own prior output, in its own voice.
+                    // That is the role the kernel's context injections already use for system facts written into session history.
+                    warn!(
+                        event = "provider_failure_note",
+                        agent = %manifest.name,
+                        error = %err_str,
+                        "Provider failed on the streaming path — the session gets an opaque note, and this line is the only place the raw provider error appears"
+                    );
+                    session.push_message(Message::system(PROVIDER_FAILURE_NOTE));
```
houko: **The `Role::System` note defeats the consecutive-same-role merge, and no hosted driver shows it to the model.**

`session_repair` merges consecutive same-role messages, and its module doc says why: *"Consecutive same-role messages (Anthropic API requires alternation)"* (`session_repair.rs:11`). That merge is adjacency-based (`session_repair.rs:232`), so interposing a `System` message between two `User` messages blocks it — and the drivers then strip the `System` message *after* repair has run:

- `anthropic.rs:377` — `.filter(|m| m.role != Role::System)`
- `openai.rs:1249/1258` — a `Role::System` history message is only emitted when `request.system.is_none()`, and the agent loop always sets `system: Some(system_prompt_snapshot.clone())`
- `gemini.rs:317`, `bedrock.rs:291` — `continue`

Concrete sequence:

1. Streaming turn fails (breaker opens) → persisted history is `user(m1), system(note)`.
2. Next turn pushes `user(m2)` → `user(m1), system(note), user(m2)`; repair finds no adjacent same-role pair.
3. Anthropic driver filters the note out → request messages are `user(m1), user(m2)`.

That is exactly the shape the merge exists to prevent. It also means the note reaches the model on none of anthropic / openai / gemini / bedrock / ollama — only on the subprocess drivers that stringify `Role::System => "System"` — so `changelog.d/fixed/7989-provider-failure-note.md`'s "the next turn reads it as a fact about the daemon" does not hold.

If the goal is purely UI visibility (the dashboard does render `role: "system"`), say so and note the model never sees it. If the goal is that the model knows, this needs a role the drivers keep.

crates/librefang-runtime/src/agent_loop/media_routing.rs
```diff
@@ -0,0 +1,609 @@
...
+    describer: &dyn ImageDescriber,
+) -> Vec<ContentBlock> {
+    if !blocks_contain_images(&blocks) || blocks_already_describe_images(&blocks) {
+        return blocks;
+    }
+
+    let mut out = Vec::with_capacity(blocks.len() * 2);
+    for block in blocks {
+        if let Some(image) = ImageRef::from_block(&block) {
+            let text = match describer.describe(image).await {
```
houko: **No timeout on the description call — the channel bridge has one, for this exact reason.**

`bridge.rs:6045`:

```rust
/// Vision APIs (Anthropic, OpenAI, Groq, Gemini) normally return in 2-8s;
/// 30s is generous but short enough that a hung provider can't pin the
/// per-(agent,channel) session indefinitely.
const INBOUND_DESCRIPTION_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(30);
```

This hop reimplements the bridge's behaviour but drops that guard. `MediaEngine::describe_image` does `self.semaphore.acquire().await` first — unbounded when other media work holds every permit — and only then makes a call whose own ceiling is 60s (`media_understanding.rs:768`). The loop here is sequential, so a 6-image dashboard upload to a text-only agent can hold the turn for ~6 minutes of provider time plus an unbounded queue wait, all *before* the user's message reaches the agent's own model.

Wrapping each `describer.describe(...)` in `tokio::time::timeout` and degrading to `IMAGE_DESCRIPTION_UNAVAILABLE` on elapse matches the bridge and reuses the failure path already written just below.

crates/librefang-runtime/src/agent_loop/media_routing.rs
```diff
@@ -0,0 +1,609 @@
...
+    };
+
+    tracing::info!(
+        agent = %manifest.name,
+        model = %api_model,
+        "Model has no vision support — routing the inbound image to the configured \
+         image_understanding capability instead of dropping it"
+    );
+    let describer = engine::MediaEngineDescriber {
+        engine: engine.with_capability_routing(&manifest.capabilities.routing),
```
houko, line 368: **This is the only caller of `with_capability_routing`, so per-agent `speech_to_text` is dead config.**

`ManifestCapabilities.routing` accepts all six keys, the dashboard renders all six fields, and `MediaConfig::with_capability_routing` knows how to fold `speech_to_text` into `audio_provider` / `audio_model`. But the only place a per-agent routing block is ever layered onto an engine is this line, on the image-description path. Transcription (`MediaEngine::transcribe_audio`, reached from the bridge and from `POST /api/media/transcribe`) always uses the boot-captured engine with only the kernel-global fold.

So: operator sets `[capabilities] speech_to_text = "groq/whisper-large-v3"` in one agent's `agent.toml`, the form shows it saved, and inbound voice notes for that agent still transcribe through the global / env-detected provider.

The four generation fields are labelled honestly — `"Global only for now — a per-agent value is not yet applied."` — but `capability_speech_to_text_hint` (`locales/en.json:367`) is just `"Turns inbound speech into text."`, with no caveat. Either give `speech_to_text` the same disclaimer as the generation fields, or thread the per-agent routing into the transcription call sites too.

crates/librefang-types/src/media.rs
```diff
@@ -1015,6 +1031,297 @@ pub struct MediaMusicResult {
...
+
+impl MediaConfig {
+    /// Fold a `[capabilities]` block into the legacy `[media]` provider /
+    /// model selectors.
+    ///
+    /// `[capabilities]` is the newer, capability-oriented surface and wins
+    /// where it carries a value; anything it leaves unset keeps whatever
+    /// `[media] image_provider` / `audio_provider` already said, and only if
+    /// *both* are silent does the engine fall back to env-var auto-detection.
+    pub fn with_capability_routing(mut self, routing: &CapabilityRouting) -> Self {
```
houko, line 1279: **A provider-only agent override inherits the *global* model id, producing a pair that does not exist.**

Because the global block is folded into `MediaConfig` at boot (`boot.rs:1637`) and the agent block is then folded on top of the result, the per-field merge crosses providers:

- `config.toml`: `image_understanding = "openai/gpt-4o"` → `image_provider = openai`, `image_model = gpt-4o`
- `agent.toml`: `image_understanding = "gemini"` — the documented "just switch the provider" spelling
- result: `image_provider = gemini`, `image_model = gpt-4o`

`describe_image` then dispatches `gemini_describe_image("gpt-4o", …)`, which 404s, and the failure degrades to `[Image description unavailable]` with nothing telling the operator the model id came from a different provider. The same spelling at global scope is fine (there is no model to inherit), so this only bites overrides — which is the case the feature exists for.

Worth deciding explicitly: either a present `provider` with an absent `model` clears the inherited model (falling back to `default_vision_model(provider)`), or the merge refuses to inherit a `model` across a changed `provider`. Note also that `CapabilityTarget::merged_over` / `CapabilityRouting::merged_over` — the functions the merge tests exercise — have no production caller; the shipped merge is this `MediaConfig` folding, so the tests are pinning semantics that never run.

crates/librefang-runtime/src/agent_loop/media_routing.rs
```diff
@@ -0,0 +1,609 @@
...
+    async fn describe(&self, image: ImageRef<'_>) -> Result<String, String>;
+}
+
+/// `true` when `blocks` already carries at least one description block.
+///
+/// Whole-message granularity on purpose: the channel bridge emits one
+/// description per image in arrival order, so a message that has any
+/// description block has been through that path already and re-describing it
+/// would double-bill for text that is already there.
+pub(crate) fn blocks_already_describe_images(blocks: &[ContentBlock]) -> bool {
```
houko, line 116: **The dedupe sniff reads user-controlled text, at whole-message granularity.**

`is_image_description_text` is a prefix match against any `Text` block in the message, and on the API / dashboard / channel paths that text is whatever the user typed. So:

- a user sends a photo to a text-only agent with the caption `[Image description: see attached]`
- or forwards a chat log that happens to contain that literal line

→ `routing_decision` returns `Skip("images already described upstream")`, nothing is described, and `redact_images_for_text_only` swaps the pixels for the `[image omitted: … on disk at <path>]` placeholder. The agent is back to answering from a filename — the exact hallucination this module's header says it exists to prevent, now reachable from the message body.

The docstring's justification ("the channel bridge emits one description per image in arrival order") holds for the bridge, but the check runs on every entry path, including ones where no bridge ran. Marking bridge-written blocks in `provider_metadata`, or scoping the sniff to the block immediately preceding each image, makes it unspoofable and also fixes the partial case (one described image and one undescribed image in the same message currently skips both).

crates/librefang-runtime/src/agent_loop/mod.rs
```diff
@@ -871,6 +892,18 @@ async fn run_agent_loop_inner(
         combined_prefix.as_deref(),
     );
 
+    // Persist the inbound message before the first LLM call. The turn can
```
houko, line 895: **This save is only on the non-streaming loop; the streaming loop — the one the dashboard uses — still has the gap.**

`run_agent_loop_streaming_inner` pushes the user message at `run_streaming.rs:396` and goes straight into `prepare_llm_messages` with no `save_session_async`. So the failure this comment calls "the worst failure mode of all" is still open on the path most turns actually take: a daemon restart, or a hang that outlives the surrounding timeout, between the push and the first interim save still loses the operator's message.

The new provider-failure branch added further down in `run_streaming.rs` covers only `stream_with_retry` returning `Err` — not a crash, not a restart, not a cancellation. Mirroring these six lines after `push_filtered_user_message` in the streaming loop would close it, and looks like the intent given how carefully the rest of this PR mirrors both loops.

crates/librefang-runtime/src/agent_loop/media_routing.rs
```diff
@@ -0,0 +1,609 @@
...
+            .map(|k| k.vision_support_for(&api_model))
+            .unwrap_or(librefang_types::model_catalog::VisionSupport::Unknown)
+            .allows_images();
+
+    let decision = routing_decision(RoutingInputs {
+        has_images,
+        already_described: blocks_already_describe_images(&blocks),
+        model_supports_vision,
+        engine_available: media_engine.is_some(),
+        description_enabled: media_engine.is_some_and(|e| e.image_description_enabled()),
```
houko, line 344: **Default-on paid vision calls on entry paths that previously made none.**

`MediaConfig::default()` sets `image_description: true` (`media.rs:250`), and `describe_image` falls back to env-var auto-detection when no provider is configured (`ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `GROQ_API_KEY` / `GEMINI_API_KEY`). Together that means an operator who configured neither `[media]` nor `[capabilities]` — only some provider key present in the daemon's environment — begins paying for a third-party vision call on every dashboard image upload to a text-only agent, where before this PR the API / dashboard path did no description at all.

The behaviour change is a real improvement and the changelog does name the off switch. But this file is explicit that "a model that *can* see is never charged for a description" is a cost property worth its own test, and the same standard argues for either (a) requiring an explicit `[capabilities] image_understanding` / `[media] image_provider` nomination on the non-channel paths instead of accepting env auto-detection, or (b) stating in the changelog fragment that existing deployments start incurring a new per-image provider charge with no config change.

crates/librefang-runtime-media/src/lib.rs
```diff
@@ -285,12 +296,53 @@ impl MediaDriverCache {
...
+            .get(&capability)
+            .cloned();
+        if let Some(provider) = nominated {
+            match self.get_or_create(&provider, None) {
+                Ok(driver)
+                    if driver.is_configured() && driver.capabilities().contains(&capability) =>
+                {
+                    return Ok(driver);
+                }
+                _ => warn!(
```
houko: **This `warn!` fires on every capability lookup, not once per misconfiguration.**

`detect_for_capability` runs per media request. A single stale nomination — say `image_generation = "minimax"` with no `MINIMAX_API_KEY` — re-fails the `is_configured()` check and re-emits this line on every `POST /api/media/image` and every `image_generate` tool call, for the life of the daemon. The information is actionable exactly once.

Latching it (a set of already-warned capabilities, or `warn!` on the first miss and `debug!` after) keeps the signal without burying everything else in the log.

sdk/javascript/index.js
```diff
@@ -897,6 +898,40 @@ class McpResource {
...
+
+  async listMediaProviders() {
+    return this._c._request("GET", "/api/media/providers");
+  }
+
+  async synthesizeSpeech(data) {
+    return this._c._request("POST", "/api/media/speech", data, undefined);
+  }
+
+  async transcribeAudio(data) {
```
houko: **The generated transcribe method can never succeed.**

Adding the `utoipa::path` annotation to `POST /api/media/transcribe` made the SDK generator emit a JSON-body method in all four SDKs:

- `sdk/javascript/index.js:922` — `transcribeAudio(data)` → `_request("POST", "/api/media/transcribe", data, undefined)`
- `sdk/python/librefang/librefang_client.py:782` — `transcribe_audio(self, **data)`
- `sdk/go/librefang.go:1035` — `TranscribeAudio(data map[string]interface{})`
- `sdk/rust/src/lib.rs` — same shape

All of them send `application/json`. The handler takes the body as raw `Bytes` and rejects anything whose `Content-Type` is not `audio/*` before it ever looks at it:

```rust
if !content_type.starts_with("audio/") {
    return ApiErrorResponse::bad_request("Content-Type must be an audio type").into_response();
}
```

So each of these is a shipped method that returns 400 unconditionally. Either exclude this route from the SDK generator, or teach the generator to emit a raw-body signature when `request_body` declares a non-JSON `content_type`.

crates/librefang-runtime/src/agent_loop/run_streaming.rs
DaBlitzStein: Fixed in `e0bc9319e` — you were right on both halves, and the second one is the reason the first matters.

The note is now `Role::Assistant` (`run_streaming.rs:955`), matching the timeout note 40 lines above it that this branch was written alongside. The `[System: …]` text prefix already carries the "this is the daemon speaking, not the agent" framing, which is what the timeout note relies on too — so the distinction the old comment was reaching for survives, in the one place a hosted model can actually read it.

I removed the comment's claim about kernel context injections rather than defending it: whatever those do, `anthropic.rs:377` filters `Role::System` out of the request and `gemini.rs:324` / `bedrock.rs:291` `continue` past it, so the claim did not hold for this note.

The existing test at `tests/utilities.rs` (renamed `streaming_provider_failure_note_is_opaque_and_reaches_the_model`) now runs the real streaming loop and then does what a driver does: `validate_and_repair` the history with a following user turn, drop the system-role messages, and assert both that the note is still there and that the result alternates. Reverting the push site to `Message::system`:

```
assertion `left != right` failed: A system-role note in the middle of history reaches no
hosted model: anthropic filters it out of the request, gemini and bedrock skip it, and
openai / ollama emit one only when the request carries no system prompt — which the agent
loop always sets. […]
  left: System
 right: System
```

`changelog.d/fixed/7989-provider-failure-note.md` no longer claims "the next turn reads it as a fact about the daemon"; that sentence is replaced by the inbound-save fix from your other thread.

crates/librefang-runtime/src/agent_loop/media_routing.rs
DaBlitzStein: Fixed in `e0bc9319e`. `DESCRIPTION_TIMEOUT` is 30s (`media_routing.rs:58`), the same ceiling and for the same reason as the bridge's `INBOUND_DESCRIPTION_TIMEOUT`, and each call goes through `tokio::time::timeout` at `media_routing.rs:165`, degrading into the `Err` arm that was already written just below — so the elapsed case produces `[Image description unavailable]` with no new failure path.

The test is `a_provider_that_never_answers_degrades_to_the_unavailable_marker`, `#[tokio::test(start_paused = true)]` against a describer that sleeps `DESCRIPTION_TIMEOUT * 4`. With the `timeout` wrapper removed:

```
  left: ["[Image description: far too late]"]
 right: ["[Image description unavailable]"]
```

The paused clock needed `tokio`'s `test-util` feature, added as a dev-dependency on `librefang-runtime` with the same spelling `librefang-kernel` and `librefang-acp` already use.

I put the bound in `enrich_blocks_with_image_descriptions` rather than around the `MediaEngine` call so it covers the semaphore wait as well as the request — which is the half that has no ceiling of its own.

crates/librefang-runtime/src/agent_loop/media_routing.rs
DaBlitzStein: Confirmed, and taken the first of your two options in `e0bc9319e`: `capability_speech_to_text_hint` now carries the same caveat as the four generation fields, in all five locales — `"Turns inbound speech into text. Global only for now — a per-agent value is not yet applied."`

I traced the alternative before choosing. Threading per-agent routing into transcription means every call site, and they are spread across three crates: `librefang-api/src/routes/media.rs:624`, `routes/agents/uploads.rs:262`, `channel_bridge.rs:2435`, and `librefang-runtime/src/tool_runner/media.rs:314` / `:1368`. Only the last two are in a crate this PR already touches, and fixing two of five would leave exactly the same trap with a smaller blast radius. That is the "different crate or domain" case, so I stopped and labelled it instead of half-doing it — say the word if you would rather have it threaded here and I will do the whole set.

The label is honest about the current behaviour rather than promising it: the form saves the value, the manifest round-trips it, and nothing reads it on the transcription path.

crates/librefang-types/src/media.rs
DaBlitzStein: Fixed in `e0bc9319e`, taking your first option: a present `provider` with an absent `model` clears the inherited model.

`with_capability_routing` now folds each capability through `fold_capability_target` (`media.rs:1303`), which clears `*model` when the incoming provider differs from the one already there, before setting the new provider. Cleared means absent, which is what makes the engine fall back to `default_vision_model(provider)` — so the "just switch the provider" spelling means what it reads like. Re-nominating the *same* provider is not a switch and keeps the inherited model, which has its own test.

Three tests replace the two that were pinning `merged_over`:

- `a_provider_only_override_does_not_inherit_the_previous_providers_model` — your exact scenario, global `openai/gpt-4o` + agent `gemini`.
- `re_nominating_the_same_provider_keeps_the_inherited_model` — the other half of the rule.
- `agent_routing_folded_over_global_inherits_the_provider_for_a_model_only_override` — the model-only override, now expressed as the two-level `MediaConfig` fold that actually ships rather than as a routing-over-routing merge.

Reverting the clear:

```
assertion `left == right` failed: a model id belongs to the provider that serves it; cleared
means the engine picks gemini's own default
  left: Some("gpt-4o")
 right: None
```

On your closing note: correct, `CapabilityTarget::merged_over` and `CapabilityRouting::merged_over` had no production caller. I deleted both rather than wiring them, since the shipped precedence is the `MediaConfig` fold and keeping a second, differently-behaving merge next to it is how the next reader gets it wrong. The doc links in `media.rs` and `agent.rs` that pointed at them now point at `MediaConfig::with_capability_routing`, so there is one place that describes resolution.

crates/librefang-runtime/src/agent_loop/media_routing.rs
DaBlitzStein: Fixed in `e0bc9319e`, taking your second option — the sniff is now scoped to the block immediately preceding each image (`image_at_is_already_described`, `media_routing.rs:106`) and applied per image rather than per message.

`blocks_already_describe_images` (`media_routing.rs:116`) now means "every image already carries its description", so it still short-circuits the whole hop when there is genuinely nothing to do and no provider call to pay for; the per-image decision happens inside the loop. That closes the partial case in the same change: one described image and one bare image in the same message now describes exactly the bare one.

Two tests, both failing against the old whole-message `any()`:

```
a_user_typed_caption_elsewhere_in_the_message_does_not_suppress_the_description
assertion `left == right` failed: the image is undescribed — user text must not stand in
for a real description
  left: 0
 right: 1

a_second_undescribed_image_is_still_described
assertion `left == right` failed: exactly the bare image is described, and the described
one is not re-billed
  left: 0
 right: 1
```

On `provider_metadata`: I went and looked rather than picking the stronger-sounding option. That field is documented as the drivers' round-trip channel for provider-required data (Gemini `thoughtSignature`), and drivers read and write it on `Text` blocks — putting a routing marker there risks it being echoed to a provider, which is a worse failure than the one being fixed. So I did not use it.

Which leaves a residue I would rather state than paper over: adjacency is not unspoofable against a user who deliberately types `[Image description: …]` immediately before their own attachment. What it does close is the accidental cases you named — a caption elsewhere in the message, a forwarded log quoting the line — and the partial case. The remaining case costs that user a worse answer on their own turn and nothing else; if you want it airtight, the honest fix is a marked block, and that wants the marker to live somewhere the drivers do not touch.

crates/librefang-runtime/src/agent_loop/mod.rs
DaBlitzStein: Confirmed and fixed in `e0bc9319e`. The six lines are mirrored after `push_filtered_user_message` in the streaming loop (`run_streaming.rs:410-416`), with the same `is_fork` / `incognito` guards, and you are right that it was the intent — the rest of the PR mirrors both loops deliberately.

The test is the part worth reading, because asserting on the session *after* the loop passes with the save deleted: the provider-failure branch further down writes the session too, so a post-hoc assertion proves nothing about the ordering.

`the_streaming_loop_persists_the_inbound_message_before_calling_the_provider` uses a driver that reads the substrate from inside `stream()` — at the exact moment the provider would be called — records what was already persisted, and then fails the turn. With the save removed:

```
the inbound message must already be persisted when the provider is called, got: []
```

The changelog fragment now names this rather than the `system`-role claim that the other thread retired.

crates/librefang-runtime/src/agent_loop/media_routing.rs
DaBlitzStein: Confirmed, and taken option (b) in `e0bc9319e`. `changelog.d/changed/7989-media-capability-routing.md` now says it plainly:

> Note that this costs money on entry paths that previously spent none: description defaults to on, and when neither `[capabilities]` nor `[media]` nominates a provider the engine auto-detects one from whichever provider key is present in the daemon's environment, so a deployment that changes no configuration at all begins paying a third-party vision call per image uploaded through the API or the dashboard to a text-only agent.

That is the sentence the release body carries, next to the existing off switch.

I chose (b) over (a) deliberately rather than by default. Requiring an explicit nomination on the non-channel paths would make the API and dashboard behave *differently from the channel bridge*, which already accepts env auto-detection — and "the same agent behaves differently depending on which door the image came through" is the divergence this PR exists to remove. Reintroducing it on the cost axis trades one surprise for another, and the one that stays is at least documented at the point an operator upgrades.

The cost property you point at is still tested: `a_vision_capable_model_is_never_charged_for_a_description` remains, and the two new dedupe tests assert call counts rather than just output, so "it silently started describing everything" now fails on the count before it fails on the text.

crates/librefang-runtime-media/src/lib.rs
DaBlitzStein: Fixed in `e0bc9319e` — latched, as you suggested.

`MediaDriverCache` grows a `warned_capability_misses: RwLock<BTreeSet<MediaCapability>>` (`lib.rs:186`); the miss branch warns on first insert and drops to `debug!` afterwards (`lib.rs:352-373`). `set_capability_routing` clears the set (`lib.rs:328`), so a config reload that changes the nomination is allowed to complain about the new one — otherwise fixing the typo and getting it wrong a second time would be silent.

`BTreeSet` rather than `HashSet` to match the `BTreeMap` next to it and the repo's determinism rule for anything that reaches a log line.

`an_unusable_nomination_is_warned_about_once_until_the_routing_changes` drives five failing lookups and asserts the latch holds exactly one entry, then re-nominates and asserts it is empty again. With the latch removed:

```
assertion `left == right` failed: five failing lookups must latch exactly one warned capability
  left: []
 right: [ImageGeneration]
```

The test asserts on the latch state rather than on captured log output — `librefang-runtime-media` has no `tracing-subscriber` dev-dependency, and pulling one in to count `WARN` lines is more machinery than the branch is worth. The `warn!`/`debug!` split keys off the same boolean the test pins.

sdk/javascript/index.js
DaBlitzStein: Fixed in `e0bc9319e`, taking your second option — the generator now emits a raw-body signature whenever an operation's `requestBody` declares no `application/json` content.

I went generic rather than excluding the route because the rule is not specific to transcribe. `_raw_body_ct` (`scripts/codegen-sdks.py:58`) reads the declared content types off the spec, and the same query finds a second one that has been shipping broken: `POST /api/agents/{id}/upload`, whose `uploadFile` / `upload_file` / `UploadFile` methods JSON-encode binary against an `application/octet-stream` handler. Excluding one route by name would have left that in place and required someone to remember this the next time a non-JSON endpoint lands.

Each SDK's existing request path gained an optional content type rather than a parallel one:

- `sdk/python`: `_request(..., content_type=None)` sends bytes verbatim when set — `transcribe_audio(self, body: bytes, content_type: str = "audio/webm")`.
- `sdk/javascript`: `_request(method, path, body, query, contentType)` — `transcribeAudio(body, contentType)`.
- `sdk/go`: `request` and the new `requestRaw` both delegate to one `do`, so the response handling is not duplicated.
- `sdk/rust`: `do_req_raw`, with `send_and_parse` factored out of `do_req` for the same reason.

The default comes from the spec's declared type, and the caller can override it — necessary for transcribe, since the handler accepts any `audio/*` and the spec can only name one.

`scripts/tests/test_codegen_sdks.py` asserts the operation set and the emitted signatures in all four languages. Disabling the detection:

```
AssertionError: FAIL [unexpected raw-body operations: []]
```

`openapi.json` is unchanged, so `xtask/baselines/openapi.sha256` is untouched; only the four generated SDKs moved.

DaBlitzStein:
Every one of the 9 review threads on this PR now has an inline reply, and the branch is at `e0bc9319e`.

DaBlitzStein:
Pushed since my last note on this PR, now at `0b48790c0`:

- `48d6f9287` Merge branch 'main' into feat/media-routing
- `0b48790c0` test(api): regenerate the KernelConfig schema golden fixture

Every one of these was verified before it was pushed rather than after: for each behaviour change the production block was reverted, the test was run against it, the literal failure was recorded, and the block was restored — a regression test that stays green with the fix removed guards nothing.

Flagging it here because a push does not announce itself, and several of these landed hours after the review threads were answered.

houko (approved):
LGTM. All five findings are addressed.

- **`speech_to_text` was dead per-agent config.** `with_capability_routing` now folds `MediaCapability::SpeechToText` onto `audio_provider` / `audio_model` alongside `ImageUnderstanding`, so the key the manifest accepts and the dashboard renders actually reaches the engine.
- **A provider-only override inheriting the global model id.** `fold_capability_target` **clears** the inherited model when the target names a provider without one, and the doc-comment carries the worked example — `image_understanding = "gemini"` over a global `"openai/gpt-4o"` resolving to `gemini` + `gpt-4o`, a pair that does not exist and 404s into `[Image description unavailable]` with nothing pointing at the cause. That is the failure mode, stated where the next reader will hit it.
- **The dedupe sniff reading user-controlled text at message granularity.** Now anchored to the immediately preceding block by index (`blocks[idx - 1]`) rather than matching any `Text` block in the message, so a user who happens to type the description prefix cannot suppress the real one.
- **The save being on the non-streaming loop only.** `run_streaming.rs:414` now saves before `prepare_llm_messages` at `:424`, so the dashboard path — the one that actually hits this — has the same durability as the non-streaming loop.
- **Default-on paid vision calls on entry paths that previously made none.** Narrowed in code and disclosed in prose, which together is the right answer. `routing_decision` only reaches `Describe` when `model_supports_vision` is false, so the call replaces a dropped image rather than adding to a working path; `vision_support_for` fails open on an unknown model, which is the safe direction. And the caveat is stated plainly in `changelog.d/changed/7989-media-capability-routing.md` — that description defaults to on, that the engine auto-detects a provider from whichever key is in the daemon's environment, and that a deployment changing no configuration begins paying per image. That paragraph goes into the release body verbatim, which is where an operator will actually see it.

CI: the previous run's ten failures were all the `dl.google.com/linux/chrome-stable` "Hash Sum mismatch" making `apt-get update` exit 100 — infrastructure, unrelated to this diff. I have re-run them; merging once they come back green.

houko:
Follow-up to the approval above: this now reports `CONFLICTING` — several PRs landed while it was open (#8013, #8059, #7985, #7986, #8175, #8186, #7996, #8228 among them). A rebase onto `main` is needed before it can merge; the review stands unchanged, and I will merge once it is clean and CI is green.

DaBlitzStein:
Refreshed against `origin/main` in `7bb8b2e13` — the branch was 16 commits behind.

Three conflicts, all in the OpenAPI surface and all additive:

- `crates/librefang-api/src/openapi.rs` and `openapi.json` — upstream added the `vault` tag (#8186) at the same position this branch adds `media`. Both are kept; the tag list is append-ordered, not alphabetical, so `vault` precedes `media`.
- `xtask/baselines/openapi.sha256` — generated, regenerated from the merged document. It parses, carries 360 paths, has no duplicate keys, and both tags are present.

DaBlitzStein:
Rebasado sobre `origin/main` `57ad92341` (release v2026.9.14) y republicado con `--force-with-lease`.
La rama vuelve a estar `MERGEABLE`.

**Qué absorbió #8320 y qué no.**
`d18c8618a` — `fix(api,runtime-media): report media capabilities we can serve, and reach a provider the registry already describes (#8320)` — toca los dos mismos ficheros centrales que este PR (`crates/librefang-api/src/routes/media.rs` y `crates/librefang-runtime-media/src/lib.rs`), pero **responde a una pregunta distinta y no solapa con el enrutado por capacidad**:

- #8320 añade a `MediaDriverCache` el campo `registry_base_urls` (la URL que el registry declara, con la del operador ganando por encima), reescribe `list_media_providers` para derivar la lista del catálogo vivo en cada petición en lugar del snapshot de arranque, y extrae `BUILTIN_MEDIA_DRIVERS` y `GENERIC_DRIVER_CAPABILITIES`.
- Este PR añade `capability_routing` (y `warned_capability_misses`), `detect_for_capability`, el bloque `[capabilities]` en config y manifiesto, `MediaConfig::with_capability_routing` y el enrutado de imágenes entrantes en `agent_loop/media_routing.rs`.

**Nada de este PR queda absorbido**: los dos conjuntos de campos y de funciones son complementarios y conviven en la estructura final, que ahora tiene los seis campos (`cache`, `provider_urls`, `media_providers`, `registry_base_urls`, `capability_routing`, `warned_capability_misses`).
Lo que sí cambió es el punto de aplicación: el `list_media_providers` que este PR anotaba con `#[utoipa::path]` ya no es el mismo cuerpo.

**Resoluciones no triviales**

- `crates/librefang-runtime-media/src/lib.rs`, tres hunks (la declaración del struct y los dos constructores `new` / `new_with_urls`): conservados los dos lados en los tres. He comprobado uno a uno que **ambos constructores inicializan los seis campos**, que es donde una fusión de este tipo se rompe en silencio.
- `crates/librefang-api/src/routes/media.rs`: me quedo con la reescritura de `list_media_providers` de #8320 (deriva del catálogo vivo, más `GENERIC_DRIVER_CAPABILITIES`) y le traslado lo que aportaba este PR y no estaba allí: la anotación `#[utoipa::path(get, path = "/api/media/providers", tag = "media", …)]` y la frase de doc que explica que `capabilities` son los mismos valores por los que enruta el bloque `[capabilities]`.
  Sin ese traslado, `routes::list_media_providers` seguiría listada en `openapi.rs:365` sin que el handler llevara la anotación.
- `crates/librefang-kernel/src/config_reload.rs`: `main` convirtió la lista de campos clasificados de cadenas sueltas a tuplas `(nombre, clase)`. He tomado la forma de `main` y añadido `("capabilities", "R")` justo detrás de `("media", "R")`, que es la clase que la propia rama justifica en `build_reload_plan` (el bloque se pliega sobre `MediaEngine` y sobre la tabla de preferencias del cache en el arranque, ambos capturados por valor).
  La fila del doc ya está en `docs/operations/config-reload.md:234` con la misma clase `R`.
- `crates/librefang-api/src/openapi.rs` y el array `tags` de `openapi.json`: conservados los dos tags, el `vault` de `main` y el `media` de esta rama, en ese orden en ambos ficheros.
- `crates/librefang-cli/src/tui/mod.rs`: me quedo con `main`. Esta rama añadía brazos de TUI para el enrutado por capacidad y luego los quitaba ella misma en `fix(cli): remove capability routing TUI code from media-routing PR`; el fichero acaba idéntico al de `main`, que es donde el propio PR quería dejarlo.
- `crates/librefang-api/dashboard/src/lib/agentManifest.test.ts`: los dos lados añadían un `describe` al final (`main` el de parámetros de inferencia de #7781, esta rama el de enrutado por capacidad). Conservados los dos.
- `xtask/baselines/openapi.sha256`: no resuelto a mano. Recalculado con `sha256sum openapi.json` sobre el `openapi.json` ya fusionado y comprobado con `sha256sum -c`; el formato coincide con el que genera `xtask/src/schema_check.rs`.

**Conjunto de ficheros**: idéntico antes y después del rebase (40 ficheros), nada se ha caído.

**Verificado** (desde `crates/librefang-api/dashboard`, con `LANG=en_US.UTF-8`)

- `npx tsc --noEmit` — limpio.
- `npx eslint .` — limpio.
- `npx vitest run` — 209/210 ficheros, 1928/1929 tests. El único rojo es `PromptsExperimentsModal > stops selection when every variant has a traffic bucket`, timeout de 5 s, ajeno a este PR.
- Los cinco `locales/*.json` y el golden fixture `kernel_config_schema.golden.json` cargan y no tienen claves duplicadas.
- `sha256sum -c xtask/baselines/openapi.sha256` → OK; `openapi.json` sin claves duplicadas, 360 rutas, `/api/media/providers` presente y `tags` terminando en `…, vault, media`.
- `rustfmt --edition 2021 --check` limpio en todos los ficheros Rust del diff.
- Barrido de campos y funciones duplicados en los ficheros tocados: sin hallazgos. Los pares que aparecen son legítimos — impls de trait por driver, o el par `#[cfg(feature = "media")]` / `#[cfg(not(feature = "media"))]` de `describe_images_for_text_only_model`.
- `with_capability_routing` tiene sus tres definiciones esperadas (`librefang-types/src/media.rs:1279`, `runtime-media/src/media_understanding.rs:79` y el stub sin la feature) y las dos llamadas (`kernel/boot.rs:1653` para el nivel global, `agent_loop/media_routing.rs:366` para el de agente).

**NO verificado por mi parte**: no he compilado ni ejecutado nada de Rust — ni `cargo check`, ni `clippy`, ni `cargo test`.
El `CARGO_TARGET_DIR` de esta máquina está compartido con otra compilación en curso, así que la parte Rust queda enteramente para CI.
Tampoco he podido regenerar `openapi.json` ni los SDKs: son el resultado del merge a tres bandas, no de una pasada del generador.

**Visto y no arreglado**: `docs/operations/config-reload.md:237` documenta `tts` como `R` mientras `config_reload.rs:1234` lo clasifica como `N`.
Es **preexistente y byte a byte idéntico en `origin/main`** (`config-reload.md:236` y `config_reload.rs:1222`), así que no lo introduce este rebase y no lo he tocado; queda anotado por si a alguien le interesa.

**Revisiones sin atender**: quedan **9 hilos de revisión sin resolver**, todos de @houko.
Cinco **no** están marcados como *outdated*: `crates/librefang-runtime/src/agent_loop/media_routing.rs:116`, `:342` y `:366`, `crates/librefang-runtime/src/agent_loop/mod.rs:891`, y `crates/librefang-types/src/media.rs:1279`.
No los he tocado —este pase era solo el rebase— y los dejo señalados en vez de darlos por cerrados.

DaBlitzStein:
**Este PR no compilaba**, y lo he arreglado en la rama.

`describe_images_for_text_only_model` (`crates/librefang-runtime/src/agent_loop/media_routing.rs:334`) llamaba a `k.supports_vision_for(&api_model)`, y **ese método no existe en ninguna parte del workspace**: ni en `KernelHandle`, ni en `CatalogQuery`, ni en el kernel. Lo comprobé buscándolo en las 38 ramas abiertas y en `origin/main`: las únicas tres apariciones del nombre estaban en este fichero — la llamada y dos comentarios que la describen.

El API real es `CatalogQuery::vision_support_for`, que ya es alcanzable porque `KernelHandle` incluye `CatalogQuery` (`kernel-handle/src/lib.rs:137`).

Y no es un cambio de nombre: **responde con tres valores, no con un booleano**, y la diferencia importa justo aquí.

- `Unsupported` — una fuente que *sabe* dice que el modelo no acepta imágenes. Es el único caso en que procede describirlas.
- `Unknown` — el modelo no está en el catálogo, **o** su `supports_vision` se infirió del nombre y lleva `vision_known: false`. Los dos casos cargan la misma cantidad de información, así que la puerta debe comportarse igual en ambos: dejar pasar los píxeles.
- `Supported` — pasa igualmente.

Eso es exactamente el «falla abierto» que el doc-comment de la función ya prometía, así que el comentario ahora nombra el método que de verdad llama.

`cargo check -p librefang-runtime --all-targets` limpio tras el cambio.

Lo encontré compilando la rama de integración, no aquí: al refrescar los 28 PRs contra el release v2026.9.14 pedí a las lanes que **no** ejecutaran cargo para no corromper el target compartido, así que la mitad Rust de este PR llegó sin compilar. Este es el precio de esa decisión y lo asumo.

DaBlitzStein:
### Dos arreglos encima de `4f47739df` (`3681df0b2`)

- **Clippy en rojo.** `map_or(true, ..)` sobre un `Option` es `clippy::unnecessary_map_or`, y el workspace lo trata como error, así que el commit anterior no pasaba `cargo clippy --all-targets -- -D warnings`. Ahora es `is_none_or`, que además dice la intención sin rodeos: si no hay kernel a quien preguntar, no se recorta nada.
- **El comentario que explica por qué solo `Unsupported` cierra la puerta estaba en español.** El resto del fichero está en inglés y este es un repositorio público.

`cargo clippy -p librefang-runtime --all-targets -- -D warnings`: 0 errores.

DaBlitzStein:
Refreshed against \`origin/main\` (\`a8a4aa64b\`) with a merge (\`d5a6abdf6\`), not a rebase — the branch was 10 commits behind. This was to fix the red \`CI / OpenAPI Drift\` check.

**What was wrong.** \`xtask/baselines/openapi.sha256\` is the byte-for-byte \`sha256sum\` of \`openapi.json\`. Neither side's stored hash described the merged file, so the check was guaranteed to fail regardless of which value survived the textual merge.

**\`openapi.json\` itself merged clean** — no conflict markers, git resolved it as a textual union. Verified this wasn't a silent drop: the merged file's path set is exactly the union of both parents' paths (branch tip \`1f9906879\`: 360 paths; \`origin/main\`: 355 paths; shared: 353; merged: 362 — nothing from either side missing, nothing extra).

**The only actual conflict** was \`xtask/baselines/openapi.sha256\`. Recalculated from the merged \`openapi.json\`: \`3718c34a68c2f2563761a4f8b0037d2254cbe1bed9b4ccfdec1ad85bff7b380a  openapi.json\`.

Also checked the failure mode this repo has hit before on auto-merged files: the five dashboard locale JSONs parse clean with no duplicate keys (\`object_pairs_hook\`), and a tree-wide grep found no leftover conflict markers.

**Not verified**: \`cargo check\` / \`clippy\` / \`cargo test\` — the local \`CARGO_TARGET_DIR\` is shared with another build in progress. Left for CI.

DaBlitzStein:
Correction to my earlier comment: the OpenAPI Drift check runs two generators and diffs three trees as a set, not just the baseline hash. My previous fix (\`d5a6abdf6\`) only recalculated the sha256 over the auto-merged \`openapi.json\` — it never re-ran the generators, so the check was still going to fail on whatever the merge got wrong.

Refreshed the branch again first: \`git merge origin/main\` (4 more commits landed since my last pass), commit \`53c70fda7\`, clean auto-merge, no conflicts this time.

**\`python3 scripts/codegen-sdks.py\`** produced zero diff — the four generated clients (Python, JS, Go, Rust) were already correct after the merge.

**\`cargo xtask codegen --openapi\`** (run with \`CARGO_TARGET_DIR=/home/blitz/development/tmp/target-7989\`, nothing else touched cargo) found real drift, one line:

\`\`\`
-        "description": "`configured` reports whether the provider's credentials are present, and\n`capabilities` is the set of `MediaCapability` values it advertises — the\nsame values the `[capabilities]` config block routes by.",
+        "description": "The list is derived from the live catalog on every request rather than from\nthe snapshot `kernel::boot` hands `MediaDriverCache`. Nothing re-runs\n`load_providers_from_registry` after boot, so a provider added by a catalog\nupdate was invisible here until a restart — which is the same class of drift\nthis route was changed to fix — and one removed lingered with an empty\ncapability list.\n\n...
\`\`\`

**What this is**: not a missing endpoint — \`GET /api/media/providers\` was already in the spec. The checked-in \`openapi.json\` still carried the pre-update doc comment on \`list_media_providers\` (\`crates/librefang-api/src/routes/media.rs:659\`); the textual auto-merge kept the old wording instead of picking up the fuller explanation a later \`main\` commit added to the same handler. Re-ran \`codegen-sdks.py\` after this regen too — still zero diff, so the longer description doesn't feed into any SDK docstring.

Recomputed the baseline from the regenerated file: \`b46b7dbb9e3d5ccf01d9a2445ba49760fdfa5506f51e9b55418b329a34644d3d  openapi.json\`.

Verified clean-room: committed, then ran both generators again from the committed state and confirmed \`git diff --quiet -- openapi.json sdk/ xtask/baselines/\` — the literal check the CI job runs — passes with no output.

Commit \`52130b1e3\`, pushed.

**Not verified**: \`cargo check\` / \`clippy\` / full \`cargo test\` — only the one \`cargo xtask codegen --openapi\` invocation ran, on a dedicated target dir, with explicit authorization since the shared target is in use by another build.

PR #7990, opened 2026-08-29
feat(cli): add models overrides subcommand

DaBlitzStein:
## Summary
- Adds `librefang models overrides <model>` CLI subcommand to view, set, and clear per-model inference parameter overrides (context_window, max_output_tokens)
- The API routes already exist (`GET/PUT/DELETE /api/models/overrides/{id}`) — this PR adds the missing CLI surface
- Full i18n coverage across all 4 locales (en, ko, uk, zh-CN)

## Verification
- `cargo check -p librefang-cli` passes

### DIFF
--- a/crates/librefang-cli/locales/en/main.ftl
+++ b/crates/librefang-cli/locales/en/main.ftl
@@ -1119,6 +1119,13 @@ model-header-auth = AUTH
 model-header-models = MODELS
 model-header-base-url = BASE URL
 model-picker-item =     { $idx }. { $id } { $tier }
+model-overrides-cleared = Overrides cleared for model { $model }.
+model-overrides-clear-failed = Failed to clear overrides for { $model }: { $status }
+model-overrides-header-field = FIELD
+model-overrides-header-value = VALUE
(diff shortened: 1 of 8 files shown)


houko (commented):
Merging — the command is wired into `main.rs`, the three routes it calls exist in `routes/providers.rs`, and the i18n gates are updated.
Two non-blocking notes for whenever you next touch this file.

`--clear` silently discards a value set in the same invocation.
`librefang models overrides gpt-4o --clear --context-window 131072` returns after the `DELETE` and never sends the `PUT`, so the operator gets "Overrides cleared" and a context window that is not what they just typed.
`#[arg(long, conflicts_with_all = ["context_window", "max_output_tokens"])]` on `clear` turns that into a clap error at parse time.

The set path is a read-modify-write against the daemon: `GET`, mutate the object in memory, `PUT` the whole thing back.
Two concurrent `models overrides` calls, or one racing a dashboard edit, will have the later `PUT` overwrite the earlier one's field rather than merge with it.
Given how this is used interactively that is unlikely to bite, but if the route grows a `PATCH` it is the better verb here.

PR #7992, opened 2026-08-29
test(api): add integration tests for agent-type CRUD, spawn, and owner identity

DaBlitzStein:
## Summary

- Add `agent_types_integration_test.rs` — 17 tests covering CRUD lifecycle, validation (missing name, path traversal, invalid chars, duplicates), TOML injection escaping, manifest-field preservation through updates (#7740), channels/routing round-trip, and full `manifest_toml` create/update paths (#7742, #6931).
- Add `agent_type_spawn_integration_test.rs` — 7 tests for template-based agent spawning: resolution from `agent-types/` dir, workspace fallback, template-dir-over-workspace precedence, save-as-agent-type round-trip with workspace clearing, duplicate rejection, and self-name allowance.
- Add `tool_dispatch_owner_identity_test.rs` — 1 acceptance test (#7744) verifying that the bearer-authenticated owner (not the attacker-chosen `sender_id` in the request body) reaches tool dispatch. Uses wiremock to stub an Ollama provider for a full turn with tool call.
- Add `hello-world` agent-type fixture under `crates/librefang-runtime/tests/fixtures/registry/agent-types/`.

## Notes

- All four files extracted from `librefang_v2` branch and verified to compile cleanly against current `main` (`cargo check -p librefang-api --tests` exits 0).
- The `save-as-agent-type` route does not exist on main yet, so the 4 spawn tests that POST to `/api/agents/{id}/save-as-agent-type` will return 404 at runtime until that route is implemented. The 3 template-resolution spawn tests exercise routes that do exist on main.
- The tool-dispatch owner identity test depends on a structured tracing record (`librefang::tool_identity`) being emitted at dispatch. If that instrumentation is not present on main, the test will fail at runtime with a clear assertion message identifying exactly what is missing.

## Test plan

- [x] `cargo check -p librefang-api --tests` passes (compilation verified)
- [ ] CI runs the new test binaries (the tests that depend on main-only routes should pass; those depending on v2-only routes will 404 at runtime)

### DIFF
--- a/crates/librefang-api/src/routes/agent_templates.rs
+++ b/crates/librefang-api/src/routes/agent_templates.rs
@@ -577,9 +577,32 @@ pub async fn create_agent_type(
 pub async fn update_agent_type(
     Path(name): Path<String>,
     lang: Option<axum::Extension<RequestLanguage>>,
-    Json(mut spec): Json<librefang_types::agent_type::AgentTypeSpec>,
+    body: axum::body::Bytes,
 ) -> impl IntoResponse {
     let lang = super::resolve_lang(lang.as_ref());
+
+    match serde_json::from_slice::<serde_json::Value>(&body) {
+        Ok(serde_json::Value::Object(_)) => {}
+        Ok(_) => {
+            return ApiErrorResponse::bad_request("Request body must be a JSON object")
+                .with_code("invalid_json")
+                .into_json_tuple();
+        }
+        Err(e) => {
+            return ApiErrorResponse::bad_request(format!("Invalid JSON: {e}"))
+                .with_code("invalid_json")
+                .into_json_tuple();
+        }
+    }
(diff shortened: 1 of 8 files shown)


DaBlitzStein:
Closing: the integration tests in this PR call `/api/agent-types` CRUD endpoints that don't exist in the upstream API surface.
These tests were written against the v2 integration branch which had custom agent-type routes.
The upstream equivalent is `/api/templates` — a rewrite against that endpoint would be a separate PR.

DaBlitzStein:
Reopening: the missing routes and test assertions have been fixed in the latest push.
Changes: resolve_manifest now checks agent-types/ first, added save-as-agent-type endpoint, body validation hardened, tool_identity tracing added, test assertions corrected.
All 20 tests pass locally.

DaBlitzStein:
Closing: this PR adds integration tests for agent-type CRUD routes (`/api/agent-types`) that don't exist in upstream yet. The tests will be reopened as part of the agent-type API PR when those routes land.

Issue #7766, opened 2026-08-21
bug: fallback chain sends the literal string "default" as a model name


DaBlitzStein:
## What happens

The fallback resolver resolves a `"default"` provider to a real one, but never does the same for the model.

`crates/librefang-kernel/src/kernel/llm_drivers.rs:420-429`:

```rust
let fb_provider = if fb.provider.is_empty() || fb.provider == "default" {
    infer_provider_from_model(&fb.model).unwrap_or_else(|| default_provider.clone())
} else {
    fb.provider.clone()
};
```

`fb.model` is passed through untouched. When a fallback entry carries `model = "default"` — the same sentinel the primary slot uses and resolves — the literal seven characters go out on the wire as the model name.

Observed on a live install:

```
Invalid model name passed in model=default
```

## Knock-on effect

`infer_provider_from_model(&fb.model)` is called with `"default"` in exactly this case, so the provider inference it is meant to perform gets a sentinel instead of a model name, and falls through to the default provider. The comment above that line explains the inference exists so a fallback like `gemini-2.0-flash` reaches Gemini rather than blindly inheriting the default provider — which is precisely the reasoning this case defeats.

Resolving the model before inferring the provider would fix both halves.

## Expected

`fb.model == "default"` (and empty) resolves to the configured default model, from the same source `resolve_driver_for_owner` uses for the primary slot (`llm_drivers.rs:180-186`: `cfg.default_model` and the hot-reloadable override), so there is one notion of "the default model" rather than two.

## Impact

A fallback is what runs when the primary provider is already failing. Getting a second, avoidable failure at that moment removes the recovery the chain exists to provide.

PR #7995, opened 2026-08-29
fix(runtime,kernel): error classification, system channel detection, and model sentinel tests

DaBlitzStein:
## Summary
- Unsupported-parameter rejections (e.g. reasoning_effort on providers that reject it) no longer count against the circuit breaker, preventing false provider lockouts.
- System channels (webui, cron, autonomous) now get a channel_send suppression instruction so agents stop trying to route media through nonexistent adapters.
- `lookup_provider_url` visibility widened to `pub(crate)` for internal reuse.
- Regression tests for `model="default"` sentinel resolution that caused "Invalid model name" failures on fallback slots.

## Verification
- `cargo check -p librefang-runtime --lib` — clean
- `cargo check -p librefang-kernel --tests` — clean
- `cargo clippy -p librefang-runtime -p librefang-kernel -- -D warnings` — zero warnings

## Issue references

- Refs #7769: Partial coverage: exclude unsupported-parameter errors from provider cooldown accounting.
- Refs #7766: Regression coverage for resolving the default model sentinel.


houko (commented):
Three unrelated changes are riding in one PR, and the one with user-visible behaviour is the one with no test.

**The prompt change is untested.**
`build_channel_section` now emits a different instruction for `webui` / `cron` / `autonomous`, telling the agent not to call `channel_send` at all.
That is a real change to what reaches the model, and nothing in this diff asserts it.
The tests added here are in `llm_drivers.rs` and cover `resolve_fallback_target`, which is a different concern entirely.
A `prompt_builder` test asserting that a system channel gets the "do NOT use `channel_send`" line and a real adapter channel still gets the `channel=…` / `recipient=…` instruction would pin the behaviour this PR exists to change.

**The sentinel list is now a third copy with no pointer back.**
`crates/librefang-runtime/src/agent_loop/mod.rs:380` carries the same `matches!(channel, "webui" | "cron" | "autonomous")` with a comment explaining that runtime cannot import `librefang_kernel::SYSTEM_CHANNEL_{CRON,AUTONOMOUS,WEBUI}` because of the dependency direction, so a grep-pointer is the best available.
The new copy in `prompt_builder.rs` has no such note, so the next person to add a system channel has two chances to find one site and miss the other.
Either lift a `fn is_system_channel(channel: &str) -> bool` that both call, or repeat the pointer comment verbatim.

**Two additions have no caller in this PR.**
`UsageStore::recent_events` + `UsageEventRow` are new public API that nothing in the diff invokes, and `lookup_provider_url` is widened from private to `pub(crate)` with no new call site.
Both look like they belong to #7976 (`feat(dashboard,tui): show agent token footprint and recent calls`).
Landing them here means neither is exercised by any test on this branch and the visibility change reads as unexplained.
Either move them into the PR that uses them, or say in the body which PR consumes them so a reviewer is not left guessing.

The retry-logging change itself is fine and the two `resolve_fallback_target` tests are worth having — splitting them out would make all three parts easier to land.

houko (changes requested):
Five unrelated changes across four crates, two of which have no caller. The prompt-builder fix is real and worth landing; the rest should be split or dropped.

## Blocking

**1. The system-channel list already exists in this crate, one file over.**

`crates/librefang-runtime/src/agent_loop/mod.rs:392`:

```rust
if matches!(channel, "webui" | "cron" | "autonomous") {
```

and now `prompt_builder.rs` has a byte-identical copy. Two definitions of "which channels are kernel-internal" in one crate is exactly the drift this PR's own title ("system channel detection") implies it is fixing. Extract `fn is_system_channel(channel: &str) -> bool` (or a `const SYSTEM_CHANNELS: &[&str]`) somewhere both can reach, and call it from both sites. Adding a fourth system channel later must not be a two-file edit that one reviewer catches.

**2. `UsageStore::recent_events` / `UsageEventRow` have no caller and no test.**

A new `pub` method plus a `Serialize`-deriving row type, and nothing in the diff reads either. This is the backend half of #7976 ("show agent token footprint and recent calls"). Landing it here means unexercised public API on `librefang-memory` that nothing proves works — the `SELECT` column order, the `i64 → u64` clamping, and the `ORDER BY timestamp DESC` on a text column are all untested. Move it into #7976 where its consumer and test live, or add both here.

Note the ordering caveat while it moves: `ORDER BY timestamp DESC` on a text column with no tiebreaker means calls recorded inside the same second come back in arbitrary order. Every other store in this repo that hit this pairs it with `, id DESC`.

**3. `lookup_provider_url` is widened to `pub(crate)` with no new caller.**

Nothing in the diff uses it from outside its module. Either the consumer is missing, or this is a leftover from another branch — please drop it if so; a visibility bump is an API decision and should arrive with the code that needs it.

**4. No test for the only behavioural fix in the PR.**

The `channel_send` prompt branch changes what reaches the LLM, which is the one thing here a user can observe. `prompt_builder.rs` already has tests. Assert that a `webui` channel with `channel_send` granted produces the "do NOT use `channel_send`" text and *not* the `recipient="{id}"` instruction, and that a `telegram` channel still produces the latter. Prompt text is also under the #3298 determinism rules, so a test at this boundary is cheap insurance.

## Non-blocking

- Per `CLAUDE.md`, one PR ↔ one issue. Retry log wording, two `resolve_fallback_target` tests, a usage-store method, a visibility bump and a prompt fix are five separable things; the title having to name three of them is the tell. The prompt fix in particular deserves to be reviewable on its own.
- The retry change duplicates the whole `build_user_facing_llm_error` + `record_retry_failure` shape in two functions with only the string differing. A single `let label = if counts_against_breaker { … } else { … };` is already what it does — worth hoisting the pair into a small helper so `call_with_retry` and `stream_with_retry` cannot drift the classification apart.
- No changelog fragment. The `channel_send`-on-webui fix is user-visible (agents were being pushed to fall back to Telegram) and belongs in the release notes: `changelog.d/fixed/7995-....md`.

houko:

No commits of your own since my review on 2026-09-01T13:47:43Z; the only thing that moved the branch tip is a merge from `main`, which the `auto-update-branches` cron pushes by itself. The PR's "updated" timestamp has therefore advanced while the blocking items have not been touched. Worth flagging because that timestamp is what makes a PR look answered from outside — I nearly mis-triaged a dozen PRs on exactly that signal today.

Still open from that review:

- 1. The system-channel list already exists in this crate, one file over
- 2. `UsageStore::recent_events` / `UsageEventRow` have no caller and no test
- 3. `lookup_provider_url` is widened to `pub(crate)` with no new caller
- 4. No test for the only behavioural fix in the PR

DaBlitzStein:
A suggestion to trim this PR, found while merging every open branch into one integration tree: the `prompt_builder.rs` hunk here and #8149 fix the same thing, and the version there is better.

Both suppress the `channel_send` media hint on kernel-internal system channels. The difference:

- **#8149** calls the shared helper `librefang_channels::types::is_reserved_system_channel(channel)` (`prompt_builder.rs:1357`) and emits **two** messages — one telling a `webui` agent that files and media reach the user through the normal response stream, and a separate one telling background runs (`cron`, `autonomous`) to target a real messaging channel explicitly.
- **This PR** hardcodes `matches!(channel, "webui" | "cron" | "autonomous")` (`prompt_builder.rs:1356`) and emits a single message for all three: *"You are on the LibreFang web interface. Files, images, and media you generate are shown to the user automatically in your response"*.

A cron run is not on the web interface, and nothing it generates is shown to a user automatically. So on two of the three channels this branch guards, the guidance it substitutes is factually wrong — and it is the kind of wrong an agent will act on, since the whole point of the hint is to steer behaviour.

The hardcoded list is also a second source of truth for something the channels crate already owns; a future reserved channel would be guarded in one place and not the other.

Nothing else in this PR overlaps — this is only about that one hunk. Dropping it here and letting #8149 carry the fix leaves both branches strictly better, and removes a conflict that would otherwise have to be resolved at merge time by whoever lands second.

For what it is worth in the integration tree, #8149's version was kept and this hunk dropped; the merge is clean that way.

DaBlitzStein:
Addressed at `1dd118cee`. The work here was removing scope, as you asked.

**1 — one definition of "system channel".** `is_system_channel` lives in `channel_registry.rs` and both `prompt_builder.rs` and `agent_loop/mod.rs:392` call it, so adding a fourth system channel is a one-file edit rather than something a reviewer has to catch.

**2 — `UsageStore::recent_events` and `UsageEventRow` are out** (−47 lines in `usage.rs`). They belong with their consumer and test in #7976. Carrying your caveat over for when they move: `ORDER BY timestamp DESC` on a text column with no tiebreaker returns calls recorded inside the same second in arbitrary order, and every other store here that hit this pairs it with `, id DESC`.

**3 — the `pub(crate)` bump on `lookup_provider_url` is reverted.** No caller arrived with it, and a visibility change is an API decision that should come with the code needing it.

**4 — the behavioural fix has a test.** `system_channels_suppress_the_channel_send_recipient_instruction` and `real_channels_keep_the_channel_send_recipient_instruction` assert both directions; neutralising `is_system_channel` fails the first and leaves the other 102 green.

Non-blocking: the duplicated `build_user_facing_llm_error` + `record_retry_failure` shape is hoisted into one helper so `call_with_retry` and `stream_with_retry` cannot drift the classification apart. And there is a `changelog.d/fixed/` fragment now — agents being pushed to fall back to Telegram is user-visible.

Ready for another look.

crates/librefang-runtime/src/prompt_builder.rs
```diff
@@ -1347,12 +1347,24 @@ fn build_channel_section(
...
+    // telling the agent to use channel_send to "webui" with a client IP as
+    // recipient would fail (no adapter) and push the agent to fall back to
+    // Telegram.
+    let is_system_channel = crate::channel_registry::is_system_channel(channel);
     let has_channel_send = granted_tools
         .iter()
         .any(|t| t == "channel_send" || t == "*");
     if has_channel_send {
-        if let Some(id) = sender_id {
+        if is_system_channel {
```
houko: This branch is taken for `cron` and `autonomous` too, and for those two the text is both factually wrong and a capability regression.

- "You are on the LibreFang web interface" is false on a cron firing (`cron_tick.rs:219` stamps `channel: SYSTEM_CHANNEL_CRON`) and on an autonomous heartbeat / goal tick (`goal_lifecycle.rs:54`, `background_lifecycle.rs:1251`).
- "media you generate are shown to the user automatically in your response" is false there as well. Cron output goes through `CronDeliveryEngine`, whose only channel primitive is `CronChannelSender::send_channel_message(channel_type, recipient, message)` (`crates/librefang-kernel/src/cron_delivery.rs:80`) — text only, no attachment parameter. Autonomous turns have no reader on the other end of a response stream at all.
- `channel_send` is the only outbound media path those turns have, and it is explicitly designed to be usable from them: the #6117 cross-chat guard in `tool_channel_send` fires only when `turn_channel.eq_ignore_ascii_case(&channel)`, so a cron turn (`turn_channel = "cron"`) targeting `channel = "slack"` passes, and the parameter docs on that function say so ("`None` for out-of-band callers (cron, triggers, external MCP) — the guard no-ops").

Concrete failure: a cron job "render the daily chart and post it to #ops" previously called `channel_send` with `channel="slack"`, `recipient="#ops"`. With this change its system prompt says the image is delivered automatically and to not use `channel_send`, so the chart is generated and never delivered — the plain-text fan-out carries only the reply body.

The narrow fix for #7995 is to stop handing a `recipient` for a sentinel channel, not to forbid the tool. Suggest keeping the "do NOT use channel_send" wording for `webui` only, and for `cron` / `autonomous` telling the agent it must name a real channel and recipient explicitly — or omitting the paragraph for those two, which restores the pre-PR behaviour there.

crates/librefang-runtime/src/prompt_builder.rs
```diff
@@ -1347,12 +1347,24 @@ fn build_channel_section(
...
+            section.push_str(
+                "\n\nYou are on the LibreFang web interface. Files, images, and media you \
+                 generate are shown to the user automatically in your response — do NOT use \
```
houko: "shown to the user automatically in your response" does not hold for `webui` either, so the instruction replaces one wrong belief with another.

Nothing attaches agent-generated media to an assistant message. `GET /api/agents/{id}/session` populates `msg["images"]` only from `ContentBlock::Image` blocks present in the stored message (`crates/librefang-api/src/routes/agents/sessions.rs:302`), and an LLM never emits an image block. On the dashboard side `ChatPage`'s `images` field is documented as "sourced from session history (`AgentSessionMessage.images`) or the user's pending uploads at send" (`ChatPage.tsx:99-101`) — i.e. inbound attachments. An assistant turn is rendered through `MarkdownContent` only.

The one path that actually reaches the user is the agent embedding the URL itself: `tool_image_generate` returns `image_urls` from `save_media_images_to_uploads` (`tool_runner/media.rs:648` and `:772`), and markdown renders it.

Concrete failure: an agent on `webui` calls `image_generate`, reads "media you generate are shown to the user automatically", and replies "Here is the chart you asked for." The returned `image_urls` never make it into the reply text, so the user sees a bare sentence and no image — the same dead end as the Telegram fallback, one step later.

Suggest stating what is true: media is delivered by including the returned URL or path in the reply as markdown, not by `channel_send`.

crates/librefang-runtime/src/channel_registry.rs
```diff
@@ -1,11 +1,29 @@
...
+/// `librefang_kernel::SYSTEM_CHANNEL_{CRON,AUTONOMOUS,WEBUI}`.
+/// Runtime can't import them directly (circular dep — runtime is below kernel),
+/// so a grep-pointer is the best we can do; api / cli / kernel sites reference
+/// the kernel constants by name and stay in lock-step.
+///
+/// Every runtime site that needs the distinction calls this, so adding a fourth
+/// sentinel is a one-line edit rather than a hunt for `matches!` copies.
+pub fn is_system_channel(channel: &str) -> bool {
+    matches!(channel, "webui" | "cron" | "autonomous")
+}
```
houko: The circular-dep premise is not accurate, and following it costs the drift guard that already exists for this list.

`librefang-runtime` already depends on `librefang-channels` (`crates/librefang-runtime/Cargo.toml:22`; `agent_loop/retry.rs` uses `librefang_channels::message_journal::RATE_LIMIT_DEFER_MARKER`), and that crate publishes exactly this list for exactly this reason — `crates/librefang-channels/src/types.rs:15` defines `RESERVED_SYSTEM_CHANNEL_NAMES = &["cron", "autonomous", "webui"]` plus `is_reserved_system_channel`, with the doc comment "duplicated here because `librefang-channels` cannot depend on `librefang-kernel`". `crates/librefang-kernel/tests/audit_cron_channel_name_test.rs` pins that list against the kernel constants.

So this is a third copy of the literals, and the only one no test observes. Concrete failure: a fourth sentinel is added — kernel constant, `RESERVED_SYSTEM_CHANNEL_NAMES`, api/cli sites — `audit_cron_channel_name_test` stays green, and `channel_registry::is_system_channel` keeps returning `false` for it. `build_sender_prefix` then stamps a sender prefix from a placeholder identity again and `build_channel_section` hands out `recipient="<sentinel>"` again, which is the bug this PR is fixing.

Delegating gives the doc comment's claim real teeth, and picks up the trim plus case-insensitive matching `is_reserved_system_channel` already does:

```suggestion
/// Keep this aligned with the kernel-side constants
/// `librefang_kernel::SYSTEM_CHANNEL_{CRON,AUTONOMOUS,WEBUI}`. Runtime can't
/// import those directly (circular dep — runtime is below kernel), but
/// `librefang-channels` mirrors the same list for the same reason and is
/// drift-guarded against the kernel constants by
/// `crates/librefang-kernel/tests/audit_cron_channel_name_test.rs`, so defer to
/// it rather than keeping a third copy of the literals here.
///
/// Every runtime site that needs the distinction calls this, so adding a fourth
/// sentinel is a one-line edit rather than a hunt for `matches!` copies.
pub fn is_system_channel(channel: &str) -> bool {
    librefang_channels::types::is_reserved_system_channel(channel)
}
```

crates/librefang-runtime/src/prompt_builder.rs
DaBlitzStein: Agreed, and fixed narrower than the suggestion: the suppression now checks `channel == "webui"` specifically rather than `is_system_channel`, so cron and autonomous fall through to a new branch that keeps `channel_send` available but tells the agent this turn has no default channel or recipient and to name a real one explicitly. Added `cron_and_autonomous_keep_channel_send_but_require_an_explicit_target` and renamed the old `system_channels_suppress_...` test to `webui_suppresses_...` (scoped to webui only) — verified red by reverting the webui-only check back to `is_system_channel`, which makes the new cron/autonomous test fail exactly as described.

crates/librefang-runtime/src/prompt_builder.rs
DaBlitzStein: Fixed — the webui instruction now says media is delivered by including the returned URL or file path in the reply as markdown, and explicitly says there is no channel for `channel_send` to deliver to, rather than claiming media is attached automatically. Added `webui_states_the_true_media_delivery_mechanism` asserting the new wording is present and the old "shown to the user automatically" claim is gone.

crates/librefang-runtime/src/channel_registry.rs
DaBlitzStein: Fixed exactly as suggested — `is_system_channel` now delegates to `librefang_channels::types::is_reserved_system_channel`, with the doc comment pointing at the drift guard (`audit_cron_channel_name_test.rs`) instead of repeating the circular-dependency premise that no longer holds. Verified `librefang-runtime` already declares the `librefang-channels` path dependency (`Cargo.toml:22`).

DaBlitzStein:
Every one of the 3 review threads on this PR now has an inline reply, and the branch is at `756d0523d`.

houko (commented):
The three findings from the previous round are addressed in the code: `cron` / `autonomous` no longer get told they are on the web interface, the webui text no longer claims automatic attachment, and `is_system_channel` now delegates to `librefang_channels::types::is_reserved_system_channel` instead of keeping a third copy of the literals. The `retry.rs` extraction is also a genuine improvement and behaviour-preserving on the user-facing path — `classify_terminal_llm_error`'s added suffix reaches only the `warn!` in `build_user_facing_llm_error`, never `user_msg`, and the suffix's wording is accurate because `should_count_against_circuit_breaker` returns `false` for exactly one case (`Api { status: 400 }` + `is_unsupported_parameter_error`, `retry.rs:145-152`).

Three things, the first of which should block.

**1. `changelog.d/fixed/7995-webui-channel-send-instruction.md` is stale and contradicts both the code and the PR's own other fragment.** It still says the instruction applies to "`webui`, `cron` and `autonomous`" and closes with "Those channels now say the opposite: media generated during the turn is shown to the user automatically" — which is the precise claim the previous round asked you to remove and which `7995-system-channel-prompt-review-fixes.md` now says was removed. Both files carry `(#7995)`, so `cargo xtask collect-fragments` folds both into `### Fixed` verbatim and the release body ships the retracted statement next to its own retraction. Delete the stale one and keep a single fragment describing the shipped behaviour.

**2. `channel == "webui"` is exact while `is_system_channel(channel)` three lines below trims and lowercases.** A `"WebUI"` or `" webui"` therefore misses the first branch and lands in the second, telling a live browser session "This turn has no default channel or recipient" — a live user's turn described as a background one, which is the bug class this PR exists to close. In practice the kernel always stamps the lowercase constant (`ws.rs:1236,1316` use `SYSTEM_CHANNEL_WEBUI`), so this is latent rather than live, but the two comparisons sitting three lines apart with different normalisation is the part that will not survive the next edit. `channel.trim().eq_ignore_ascii_case("webui")` matches the predicate you are already deferring to.

**3. This PR and #8149 rewrite the same block of `prompt_builder.rs`, with different code and opposite rationales.** Whichever merges second will conflict, and the conflict is not mechanical — it is a real disagreement that should be settled before either lands:

| | #7995 (this PR) | #8149 |
| --- | --- | --- |
| system-channel predicate | `is_reserved_system_channel` via `channel_registry::is_system_channel` | hardcoded `cron` / `autonomous` / `webui` literals |
| stated reason | the circular-dep excuse "no longer held" — runtime already depends on `librefang-channels` | the reserved list belongs to `sanitize_channel_name`'s SessionId-collision concern and may gain an interactive surface |
| cron / autonomous text | `channel_send` still works — name a real channel and recipient | "`channel_send` cannot reach this system channel — do NOT use it here" |
| case handling | exact for `webui`, insensitive for the rest | insensitive throughout, with a test |

Both of those rationales came out of my own review rounds, and they pull in opposite directions, so this is mine to resolve rather than yours: **the dependency argument in this PR is correct and the semantic argument in #8149 is also correct.** The version that satisfies both is one set of literals living low in the stack and a *separate, named* predicate for the prompt's question. Concretely: keep `is_reserved_system_channel` as the SessionId-collision answer it already is, and give the prompt builder its own `is_non_interactive_turn(channel)` defined over the same constants — so adding a future interactive surface to the reserved list does not silently change what the prompt tells a live user, and there is still only one copy of the strings.

Say which PR you would rather carry that, and I will close the loop on the other one.

DaBlitzStein:
Pushed `86d1683ff`. All three items from the `2026-09-10` round, including an answer to the question in #3.

**1. Stale changelog fragment — deleted.** You were right on both counts: `changelog.d/fixed/7995-webui-channel-send-instruction.md` still described the cron/autonomous suppression this PR removed, and still closed with "media generated during the turn is shown to the user automatically" — the exact claim the sibling fragment says was retracted. Both carried `(#7995)`, so `collect-fragments` would have folded the retraction and the retracted statement into `### Fixed` next to each other. One fragment now, describing what ships.

**2. The asymmetric comparison — fixed.** `channel.trim().eq_ignore_ascii_case(SYSTEM_CHANNEL_WEBUI)`. Verified the new test fails against the exact comparison with `"WebUI" must take the webui arm` and passes with it, so a future edit that re-tightens it gets caught.

**3. I would rather #7995 carried it, and it now does.**

Your framing is the one I would not have reached on my own: the dependency argument and the semantic argument are both correct because they are answers to two different questions. So `is_reserved_system_channel` stays exactly what it is — which names collide with a kernel `SessionId` — and the prompt builder has its own `is_non_interactive_turn` for its own question, whether anybody is waiting on the reply.

`webui` is the case that makes the distinction load-bearing rather than pedantic: it is reserved *because* it derives a `SessionId`, yet a live browser session is waiting on the answer. Deriving the prompt's predicate from the reserved list means a future interactive surface added for `SessionId` reasons silently starts telling a real user their turn is a background one.

To keep one copy of the strings I gave `librefang-channels` the three named constants it was already documenting as a mirror of the kernel's — `SYSTEM_CHANNEL_CRON` / `SYSTEM_CHANNEL_AUTONOMOUS` / `SYSTEM_CHANNEL_WEBUI` — and built `RESERVED_SYSTEM_CHANNEL_NAMES` from them. `NON_INTERACTIVE_TURN_CHANNELS` in the prompt builder is then its own list over those same constants, no new literals. The `webui` constant carries a doc-comment saying why callers who care about the distinction must name it rather than test list membership.

That makes #8149 redundant rather than conflicting — it is yours to close whenever you like, and I have not touched it.

**Verification**
- `cargo check --workspace --all-targets --exclude librefang-desktop` → exit 0, 26 librefang crates checked, no errors, no warnings.
- `cargo test -p librefang-runtime --no-fail-fast channel` → 72/72.
- `cargo test -p librefang-channels --no-fail-fast` → 616 + 38 + 7 pass.
- `cargo test -p librefang-kernel --no-fail-fast --test audit_cron_channel_name_test` → 4/4, so refactoring the reserved list into named constants did not move the drift guard.

DaBlitzStein:
Añadido en `56e0a3a3e`. Viene de trabajo que estaba **sin commitear** en un worktree y se habría perdido al limpiarlo; **va emparejado con #8149** y hay un matiz de orden de merge al final que conviene leer antes de mergear ninguno de los dos.

## Los dos PRs se contradecían en el mismo bloque

Este PR y #8149 reescriben cada uno la misma rama de `build_channel_section`, que `origin/main` **no tiene en ninguna de las dos formas**. Y decían lo contrario: aquí se ofrecía `channel_send` con destino explícito, y allí se prohibía —y en la frase siguiente se explicaba cómo usarla—. Solo una puede ser cierta.

Es la oferta. `channel_send` lee `channel` y `recipient` de sus propios argumentos y despacha a ese adaptador, así que el canal centinela bajo el que corre el turno no entra en la decisión; un turno de fondo alcanza a una persona en un canal real igual que cualquier otro. Lo que sí es imposible es responder **dentro de** `cron` o `autonomous`, que no tienen adaptador, y eso es lo que dice ahora el prompt.

Las dos ramas emiten ya **la misma frase byte a byte**, comprobada con `diff` sobre el literal resuelto de los dos árboles (expandiendo las continuaciones `\`, que en Rust se comen el salto y la sangría), no a ojo. Y no solo la frase: los **cuatro** textos que emite el bloque salen idénticos en ambas. Un test en cada lado la custodia nombrando al PR hermano — aquí `background_run_guidance_matches_the_wording_shared_with_8149`.

## Lo que NO apliqué, y por qué

El rescate traía además la normalización de `webui` con `eq_ignore_ascii_case` sobre un literal. **`86d1683ff` ya la había resuelto mejor en esta rama**, con `librefang_channels::types::SYSTEM_CHANNEL_WEBUI` y el predicado propio `is_non_interactive_turn`. Aplicar el diff tal cual la habría pisado con la versión peor, así que porté solo lo que faltaba: el texto de `/api/uploads` en la rama `webui` y la frase compartida. La estructura de `86d1683ff` queda intacta.

## ⚠️ Orden de merge: **mergear #8149 antes que éste**

La mitad `webui` —el match normalizado y el texto de `/api/uploads`— **ya está en la rama de #8149** y no estaba en ésta. Si este PR entra primero, entra sin ella durante la ventana entre los dos merges. Medido: `grep -c eq_ignore_ascii_case("webui")` daba **1** en `fix/channelsend-system-guard` y **0** aquí antes de este commit. Ahora los dos textos coinciden, pero el orden sigue siendo el seguro.

## Verificación

- `cargo test -p librefang-runtime --no-fail-fast` → **2499/2500**.
- El único fallo es `chatgpt_oauth::tests::test_callback_server_releases_listener_after_timeout` con `AddrInUse`, contención de puerto entre tests en paralelo. **Demostrado ajeno**: aislado con `--test-threads=1` pasa. No toca `prompt_builder`.
- `cargo check --workspace --all-targets --exclude librefang-desktop` → `EXIT=0`.
- `cargo clippy --workspace --all-targets --exclude librefang-desktop -- -D warnings` → `EXIT=0`, cero warnings.
- **Rojo primero**: revirtiendo solo `prompt_builder.rs` y dejando los tests, caen `background_run_guidance_matches_the_wording_shared_with_8149` y `webui_states_the_true_media_delivery_mechanism`; reponiéndolo, verdes los cuatro del bloque.

También corregí en el fragmento existente la frase que decía «embeds the returned URL or file path as markdown», que quedó obsoleta al nombrar el texto la URL `/api/uploads`.

(`librefang-desktop` se excluye porque a esta máquina le faltan las cabeceras GTK.)

DaBlitzStein:
Findings from carrying this PR and #8149 together in an integration branch. Splitting them because the two halves of `756d0523d` have landed differently.

**The behavioural half is already satisfied by #8149's landed version.** Its review finding — that the `channel_send` suppression covered cron and autonomous turns and regressed a capability they already had — is real, and I confirmed the capability: `tool_channel_send` reads `channel` from the tool input, not from the turn, and its own comments state that `sender_channel` / `sender_chat_id` are `None` for cron and triggers so the cross-chat guards no-op. A background turn can reach a real messaging channel.

But the merged wording already preserves it. `prompt_builder.rs` scopes the strong suppression to `webui` (`channel.trim().eq_ignore_ascii_case("webui")`), and the cron/autonomous branch ends with *"To reach a person, target a real messaging channel/recipient explicitly."* — the capability is stated, not withheld.

**The structural half is still open.** `channel_registry.rs:24` is still `matches!(channel, "webui" | "cron" | "autonomous")` — the third hand-copied sentinel list this PR replaces with `librefang_channels::types::is_reserved_system_channel`. That change is unaffected by the above and still worth landing.

**One caveat if you do land it**, because the two call sites want opposite things: `prompt_builder.rs:1357-1366` deliberately keeps its own literals and says why — `is_reserved_system_channel` exists to stop externally-supplied names colliding with a kernel-derived `SessionId`, and its list can grow to include a future *interactive* surface, which would then wrongly suppress `channel_send` there. So delegating `channel_registry::is_system_channel` is right; propagating the same delegation into `prompt_builder` would not be.

**Why the commits do not cherry-pick cleanly:** the test assertions were written against this PR's own wording. `system_channels_suppress_the_channel_send_recipient_instruction` on `main` loops `["webui", "cron", "autonomous"]` and accepts either phrasing via a `||`; this PR's tests assert the absence of the exact string `Do NOT use \`channel_send\`` for cron/autonomous. Both pass against the merged text, but the surrounding rewrite conflicts. `86d1683ff` and `56e0a3a3e` build on `756d0523d`, so they conflict too.

Measured, not assumed: `NON_INTERACTIVE_TURN_CHANNELS` is absent from `main` (0 occurrences), as are the three commits' tests.

No action requested — I did not resolve any of this in the integration branch, because choosing between two reviewed wordings is not mine to make. Flagging it so the rebase is done knowing which half is already in.

houko (approved):
Re-reviewed at the current head. All four blocking findings are addressed, and the first one better than I asked.

**1 — the duplicated system-channel list.** I asked for one predicate both call sites reach. `channel_registry::is_system_channel` is that, and it delegates to `librefang_channels::types::is_reserved_system_channel` rather than holding its own copy of the literals — so this is the second definition folding into the existing one, not a third being created. `agent_loop/mod.rs:387` calls it, and the drift guard against the kernel constants (`audit_cron_channel_name_test.rs`) covers the remaining copy.

**2 and 3 — `UsageStore::recent_events` and the `lookup_provider_url` visibility bump.** Both gone; `1dd118ce` narrows the PR to the prompt fix. The diff is now eight files in one area rather than five unrelated changes across four crates, which also settles the non-blocking one-PR-one-issue point.

**4 — a test for the behavioural fix.** Six, including the exact pair I asked for: `webui_suppresses_the_channel_send_recipient_instruction` and `real_channels_keep_the_channel_send_recipient_instruction`.

Issue #3355, opened 2026-04-27
Central slash command registry — single source of truth across CLI / TUI / channels / Dashboard


houko:
## Problem

Slash commands in LibreFang are defined **independently in every surface that exposes them** — CLI TUI, Telegram, Slack, Dashboard, future channels. Each surface ships its own list, its own help text, its own handler dispatch. There is no shared source of truth, so:

- Adding a new command means editing N places (CLI handler + Telegram `setMyCommands` payload + Slack subcommand router + TUI help screen + Dashboard command palette).
- Renaming or deprecating a command leaks: the new name lands in CLI, the old name lingers in Telegram for a release.
- Help text drifts between surfaces — TUI `/help` and Telegram BotFather descriptions tell users different things about the same command.
- No machine-readable list exists for things like Telegram's `setMyCommands`, Discord slash command registration, or Dashboard tooltips — each surface hand-rolls it.

## Evidence

Definitions / dispatch live in at least these locations:

```
crates/librefang-cli/src/tui/chat_runner.rs   // ChatAction::SlashCommand match arm
crates/librefang-cli/src/tui/screens/chat.rs  // SlashCommand enum value
crates/librefang-cli/src/tui/mod.rs           // help / autocomplete
crates/librefang-channels/src/telegram.rs     // BotCommands payload + dispatch
crates/librefang-api/src/routes/system.rs     // Dashboard command surface
```

Each maintains its own subset of the command list, with no cross-reference. There is no `librefang-types::command_registry::REGISTRY` (or equivalent) that all of them read from.

## Proposed approach

A central registry crate / module — single source of truth for command metadata, surfaces consume it.

1. **`librefang-types::commands`** module defining a `CommandSpec`:
   ```rust
   pub struct CommandSpec {
       pub name: &'static str,                    // e.g. "model"
       pub aliases: &'static [&'static str],      // e.g. &["m"]
       pub scope: CommandScope,                   // Cli, Channel, Dashboard, All
       pub description: &'static str,             // one-line, surfaced in help
       pub args_hint: Option<&'static str>,       // e.g. "<model_name>"
       pub permissions: PermissionsRequired,      // gates dashboard / channel exposure
   }

   pub static REGISTRY: &[CommandSpec] = &[ ... ];
   ```

2. **CLI / TUI** read `REGISTRY` to populate help, autocomplete, and the `SlashCommand` dispatcher's match arms. Existing handler bodies stay where they are; only the metadata centralises.

[...]

PR #7996, opened 2026-08-29
refactor(api): derive dashboard command catalog from the central registry

DaBlitzStein:
## Summary
- Replaces the hardcoded `BUILTIN_COMMANDS` array in the commands route with a derivation from the central `COMMAND_REGISTRY` via `iter_for(Scope::DASHBOARD)`.
- Commands added to the registry with `Scope::DASHBOARD` now appear in the dashboard slash menu automatically, closing the split-brain that left `/goal` visible in Telegram but invisible in the dashboard (#3355).
- Adds `DashboardExec` enum to `CommandDef` so the SPA knows which commands resolve client-side vs. over the WebSocket.
- Adds `/reset`, `/verbose`, `/info`, `/context`, `/queue`, `/goal` as new registry entries with the appropriate dashboard scope and exec mode.

## Verification
- `cargo check -p librefang-channels --lib` — pass
- `cargo check -p librefang-api --lib` — pass
- `cargo clippy -p librefang-channels --lib -- -D warnings` — pass
- `cargo clippy -p librefang-api --lib -- -D warnings` — pass
- `cargo test -p librefang-channels -- commands` — 32 passed
- `cargo test -p librefang-api --test commands_routes_test` — 3 passed (goal_is_served_by_the_command_catalog, catalog_still_serves_every_historical_builtin, builtins_win_collisions_and_skill_commands_remain_queryable)


houko (commented):
This is now fully contained in #7994, which merged a moment ago.

The `routes/commands.rs` change here — dropping `BUILTIN_COMMANDS`, `builtin_commands()` over `commands::iter_for(Scope::DASHBOARD)`, `builtin_entry()` emitting `desc_key` / `no_args` / `args_hint` / `exec`, and the alias-aware `seen` set — is the same change #7994 carried alongside the `/goal` wiring, down to the module doc comment.
Both PRs also touch `crates/librefang-channels/src/commands/mod.rs` and `crates/librefang-api/tests/commands_routes_test.rs`.

Two things worth checking before deciding what to do with this branch:

- The `builtin_entry` doc comment here is better than what landed — it explains that `desc_key` follows the `chat.cmd_<name>` convention with a `desc` fallback, and that an absent `exec` is the SPA's signal to keep the command out of the menu.
  That is worth keeping; a small PR moving just that comment onto the merged function would land cleanly.
- If this branch has registry entries or tests that #7994 does not, rebasing onto `main` will show what survives. If the rebase comes out empty, this can be closed as superseded.

Flagging rather than closing — it is your branch and your call which of the two shapes you meant to land.

houko (changes requested):
Deriving the catalog from `COMMAND_REGISTRY` is the right fix and the new guard tests are good. But moving the `/goal` block shifted two adjacent entries, and `/think` and `/exit` silently lost their dashboard execution path — with the new tests written so they pass anyway.

## Blocking

**1. `/think` loses `dashboard_exec`.**

On `main`, `commands/mod.rs:259-269`:

```rust
CommandDef {
    name: "think",
    …
    scope: Scope::CHANNEL.union(Scope::DASHBOARD),
    telegram_menu: true,
    dashboard_exec: Some(DashboardExec::Backend),
},
```

After this PR it ends `dashboard_exec: None`, and the `Some(DashboardExec::Backend)` line is consumed by the newly inserted `verbose` entry. Per this PR's own new doc on `builtin_entry`:

> `exec` is absent when the command is catalogued but has no dashboard execution path, which is **the SPA's signal to keep it out of the menu**.

So `/think` disappears from the dashboard slash menu while keeping `Scope::DASHBOARD`.

**2. `/exit` loses `dashboard_exec` the same way.**

`commands/mod.rs:518-527` on `main` is `"End chat session"`, `Scope::CLI.union(Scope::DASHBOARD)`, `dashboard_exec: Some(DashboardExec::Client)`. The last hunk turns that into `None` as a side effect of deleting the `goal` block that followed it. `/exit` also drops out of the menu.

Both look like block-move accidents rather than decisions — nothing in the PR body mentions removing either command from the dashboard.

**3. The new guards are shaped so they cannot catch #1 or #2.**

`dashboard_scope_covers_the_historical_builtin_catalog` lists `"think"` and `"exit"` and only asserts they carry `Scope::DASHBOARD` — which they still do. `dashboard_exec_matches_the_historical_chat_menu` omits both names from its `client` and `backend` lists, so nothing checks them. `dashboard_exec_implies_dashboard_scope` only tests one direction.

Add the converse, which catches this whole class in one line:

```rust
#[test]
fn dashboard_scope_implies_dashboard_exec() {
    for c in COMMAND_REGISTRY {
        if c.scope.contains(Scope::DASHBOARD) {
            assert!(c.dashboard_exec.is_some(),
                "`/{}` carries Scope::DASHBOARD but no dashboard_exec, so the SPA will catalogue it and hide it from the menu",
                c.name);
        }
    }
}
```

and put `think` and `exit` back into the `backend` / `client` lists respectively, so the golden assertion covers every dashboard command rather than a subset.

## Non-blocking

- `CommandDef::usage()` loses its doc comment (`/// One-line usage hint, e.g. Usage: /goal <description> [--loop-engineering]`) with nothing in its place. It was carrying an example; worth keeping.
- The change to `cli_scope_does_not_leak_into_channel_set` (`scope == Scope::CLI` → `!scope.contains(Scope::CHANNEL)`) genuinely strengthens the guard and the new comment explains why — no objection, just flagging it as a deliberate widening for anyone reading the diff quickly.
- `/agents` and `/budget` gain `Scope::DASHBOARD` with `Client` and `Backend` exec respectively. Those are new dashboard capabilities rather than restored ones, so they belong in the PR body and a changelog fragment — this PR has no `changelog.d/` entry, and "the dashboard slash menu now offers /goal, /agents, /budget, /peers, /a2a, /context, /queue, /verbose and /info" is exactly the kind of thing the release notes should carry.

houko:

No commits of your own since my review on 2026-09-01T13:49:38Z; the only thing that moved the branch tip is a merge from `main`, which the `auto-update-branches` cron pushes by itself. The PR's "updated" timestamp has therefore advanced while the blocking items have not been touched. Worth flagging because that timestamp is what makes a PR look answered from outside — I nearly mis-triaged a dozen PRs on exactly that signal today.

Still open from that review:

- 1. `/think` loses `dashboard_exec`
- 2. `/exit` loses `dashboard_exec` the same way
- 3. The new guards are shaped so they cannot catch #1 or #2

DaBlitzStein:
Addressed at `14049734e`, all three blocking points plus the non-blocking ones.

**1 and 2 — `/think` and `/exit` have their `dashboard_exec` back**, `Backend` and `Client` respectively. I checked every command this PR's block move touched against `origin/main` rather than only those two, in case there was a third; there was not.

**3 — the converse guard, but not the shape you proposed**, and this is the one place I deviated. `dashboard_scope_implies_dashboard_exec` with a blanket `is_some()` would assert something that was never true: `/status` carries `Scope::DASHBOARD` with `dashboard_exec: None` deliberately, which `chatCommands.ts` documents as "catalogued but no dashboard execution path" — it keeps the command out of the slash menu and lets it fall through to the agent as text.

So `dashboard_scope_without_exec_is_limited_to_known_catalogue_only_commands` pins the **set** of such commands instead. That is strictly stronger than the blanket assertion would have been: it fails when a command silently loses its exec *and* when one silently acquires a `None`, so neither direction of the drift lands unnoticed. If you would rather have the blanket form with `/status` skipped, say so and I will swap it.

Reintroducing the `/think` regression fails two tests, the second explaining itself:

```
`/think` must stay dispatched over the chat WebSocket
a Scope::DASHBOARD command without dashboard_exec is catalogued by the SPA and hidden
from the slash menu; if this set grew, that command silently dropped out of the dashboard
menu, and if it shrank, update this test
test result: FAILED. 19 passed; 2 failed
```

21 passed with the fix. `think` and `exit` are back in the golden `backend` / `client` lists so it covers every dashboard command rather than a subset.

Non-blocking: `CommandDef::usage()` has its doc comment and example back, and there is now a `changelog.d/added/` fragment for the slash-menu additions — `/agents` and `/budget` are new dashboard capabilities rather than restored ones, which belonged in the release notes.

Ready for another look.

crates/librefang-channels/src/commands/mod.rs
```diff
@@ -975,4 +1030,115 @@ mod tests {
...
+    #[test]
+    fn goal_is_reachable_from_every_chat_surface() {
+        let def = lookup("goal").expect("/goal must be registered");
+        assert!(
+            def.scope.contains(Scope::CHANNEL),
+            "/goal must stay reachable from channels"
+        );
+        assert!(
+            def.scope.contains(Scope::CLI),
+            "/goal must be reachable from the TUI chat runner"
```
houko, line 1066: `/goal` carries `Scope::CLI`, but the TUI chat runner has no `"goal"` arm, so this assertion documents a panic rather than reachability.

`crates/librefang-cli/src/tui/chat_runner.rs:295` resolves the typed command through `commands::lookup` and accepts it as soon as `def.scope.contains(Scope::CLI)`, then matches `canonical` against `exit | help | status | model | clear | new | kill` and ends at line 469 with an `other => unreachable!(...)` arm whose comment claims "the pre-flight `commands::lookup` guarantees `canonical` is one of the names matched above". `goal` is the only `Scope::CLI` command with no arm, so typing `/goal ship the release` in `librefang chat` passes the pre-flight check and hits `unreachable!` — the TUI panics with the terminal still in raw mode.

Either add a `"goal"` arm to `chat_runner.rs` (the daemon already exposes the behaviour the dashboard uses, via `goal_command_response` in `crates/librefang-api/src/ws.rs`), or drop `Scope::CLI` from the entry and relax this assertion. Replacing `unreachable!` with the same `chat-runner-unknown-command` message the pre-flight already emits would also stop the next registry edit from turning into a crash.

crates/librefang-channels/src/commands/mod.rs
```diff
@@ -975,4 +1030,115 @@ mod tests {
...
+                actual.contains(name),
+                "`/{name}` disappeared from the dashboard catalog"
+            );
+        }
+    }
+
+    #[test]
+    fn dashboard_exec_matches_the_historical_chat_menu() {
+        let client: &[&str] = &["help", "clear", "agents", "info", "exit"];
+        let backend: &[&str] = &[
```
houko, line 1098: Pinning `think` as `Backend` locks in a command the WS handler does not answer.

`handle_command` at `crates/librefang-api/src/ws.rs:1736-1978` matches `new | reset | reboot | compact | stop | model | usage | context | verbose | queue | budget | peers | a2a | goal` and falls through to `_ => {"type": "error", "content": format!("Unknown command: {cmd}")}`. There is no `"think"` arm — the dashboard toggles extended thinking through the `thinking` field on the `{"type":"message"}` frame (`ChatPage.tsx:882`), not through a slash command.

So with `Scope::DASHBOARD` + `DashboardExec::Backend`, `/think` is offered in the slash menu (`ChatPage.tsx:1856` filters on `menuCommands`), and selecting it renders the system message `Unknown command: think`.

Two further signals that `Backend` is wrong here. The hand-written `SLASH_COMMANDS` list this work replaces never contained `/think` (`git show c33876f36^:crates/librefang-api/dashboard/src/pages/ChatPage.tsx`, lines 110-127 — client: `help`, `clear`, `agents`, `info`; backend: 13 others), so the "historical chat menu" this test claims to pin did not include it. And `crates/librefang-api/dashboard/src/lib/chatCommands.test.ts:14` uses `/think` as its fixture for "a catalogued builtin with no dashboard execution path", asserting `menuCommands` keeps it out of the menu.

`dashboard_exec: None` on `/think` — the same state `/status` is in — matches both, and is a one-field edit.

crates/librefang-channels/src/commands/mod.rs
```diff
@@ -975,4 +1030,115 @@ mod tests {
...
+            assert!(
+                actual.contains(name),
+                "`/{name}` disappeared from the dashboard catalog"
+            );
+        }
+    }
+
+    #[test]
+    fn dashboard_exec_matches_the_historical_chat_menu() {
+        let client: &[&str] = &["help", "clear", "agents", "info", "exit"];
```
houko: `exit` as `Client` puts a no-op in the slash menu that spends tokens.

Nothing in the SPA handles `/exit`. `ChatPage.tsx:690-726` answers exactly `/help`, `/clear`, `/agents` and `/info` locally; `backendCommandNames` excludes `client` entries, so `/exit` falls through to the ordinary send path and is delivered to the agent as the literal prompt `/exit`. Because `args_hint` is empty the catalog reports `no_args: true`, and the menu sends on click — one click, one LLM turn, one nonsense answer.

Like `/think`, `/exit` was not in the hand-written menu this work replaces (`git show c33876f36^:.../ChatPage.tsx`, lines 110-127 — `help`, `clear`, `agents`, `info` client-side, 13 backend), so the "historical chat menu" this test claims to pin did not include it. Either add a client-side `/exit` branch (disconnect / navigate away) or set `dashboard_exec: None` so it stays out of the menu.

crates/librefang-channels/src/commands/mod.rs
```diff
@@ -183,6 +188,7 @@ pub const COMMAND_REGISTRY: &[CommandDef] = &[
         name: "new",
         aliases: &[],
         category: Category::Session,
+        // Also reachable from CLI/TUI chat surfaces; both reset the agent session the same way.
```
houko: This comment is the opposite of what the two surfaces do, in a registry whose whole purpose is to be the single source of truth.

Dashboard `/new` calls `state.kernel.create_agent_session(agent_id, None)` and returns a **new** `session_id`, which the SPA uses to navigate (`ws.rs:1738-1752`, `ChatPage.tsx` `onNewSession`). TUI `/new` posts to `/api/agents/{id}/session/reset` / calls `kernel.reset_session(id, ResetScope::Agent)` — same session id, history cleared (`chat_runner.rs:365-390`).

That is precisely the distinction the sibling comment on `reset` (line 203) draws, and `/new` vs `/reset` in this registry encode it: "Start a new session (new session id)" vs "same session id". Saying the surfaces "both reset the agent session the same way" invites the next editor to collapse them.

crates/librefang-channels/src/commands/mod.rs
```diff
@@ -935,18 +990,18 @@ mod tests {
     /// channel-visible set (golden assertion guard).
     #[test]
     fn cli_scope_does_not_leak_into_channel_set() {
-        // Running this alongside `channel_command_names_match_historical_set`
-        // catches any future drift where someone adds a CLI-only command but
-        // accidentally tags it `Scope::CHANNEL`.
-        let cli_only: Vec<&str> = COMMAND_REGISTRY
+        // Matches on "not channel-scoped" rather than "exactly Scope::CLI" so
```
houko, line 1016: This guard cannot fail, and the new comment overstates what it covers.

The filter keeps commands where `!c.scope.contains(Scope::CHANNEL)`, and the assertion calls `is_channel_command(name)`, which is `lookup(name).is_some_and(|c| c.scope.contains(Scope::CHANNEL))`. Since `names_and_aliases_are_globally_unique` guarantees `lookup(name)` returns the very `CommandDef` the filter just inspected, the assertion re-reads the field the filter selected on — it is true by construction for every element.

Concretely, the drift the original comment named — someone adds a dashboard- or CLI-only command and mistakenly tags it `Scope::CHANNEL` — excludes that command from `non_channel`, so the test still passes. The pre-existing version had the same defect; broadening the filter does not fix it. A guard that would bite compares the channel-visible set against `channel_command_names_match_historical_set`'s golden list (which already exists) or asserts `iter_for(Scope::CHANNEL)` against an explicit expected set for the non-channel names.

crates/librefang-channels/src/commands/mod.rs
DaBlitzStein: Fixed by adding a real `"goal"` arm to `chat_runner.rs::handle_slash_command` rather than dropping `Scope::CLI` — the existing `goal_is_reachable_from_every_chat_surface` test already asserts the CLI scope bit is intentional, so removing it would have weakened a test that pins real intent. For `Backend::InProcess` it calls `librefang_kernel::goal_runner::create_and_start_goal` directly (the same function `ws.rs::goal_command_response` and the Telegram bridge already call) and reuses `GoalLaunch::message()` for the confirmation text. `Backend::Daemon` has no HTTP endpoint wrapping that function yet, so rather than build one in this sweep it answers with a clear "not yet supported while attached to a daemon" message (new `chat-runner-goal-daemon-unsupported` key, all 4 locales) instead of crashing or silently no-op'ing — happy to open a follow-up issue for full daemon-mode wiring if you'd rather track it separately. Replaced `unreachable!()`'s risk surface by giving `goal` its own arm; the `other => unreachable!()` fallback stays for genuine registry drift. Added `goal_command_does_not_panic_without_a_backend` — verified it panics with the exact `unhandled CLI command \`goal\`` message when the arm is removed, passes with the fix.

crates/librefang-channels/src/commands/mod.rs
DaBlitzStein: Fixed — `/think` now carries `dashboard_exec: None`, matching `/status`. Updated `dashboard_exec_matches_the_historical_chat_menu` to drop `think` from the `backend` list (with a comment explaining neither it nor `/exit` were in the hand-written menu this registry replaced) and `dashboard_scope_without_exec_is_limited_to_known_catalogue_only_commands`'s golden set to `["exit", "status", "think"]`.

crates/librefang-channels/src/commands/mod.rs
DaBlitzStein: Went with the `dashboard_exec: None` alternative over adding a client-side handler, to avoid inventing a disconnect/navigate-away UX in this sweep — `/exit` now carries `dashboard_exec: None` like `/status`, and is out of both lists in `dashboard_exec_matches_the_historical_chat_menu`.

crates/librefang-channels/src/commands/mod.rs
DaBlitzStein: Fixed — rewrote the `/new` comment to state the actual distinction (dashboard mints a new session id, TUI clears history at the same session id), matching the sibling `/reset` comment two lines below.

crates/librefang-channels/src/commands/mod.rs
DaBlitzStein: Fixed as suggested — replaced the self-referential `is_channel_command(name)` re-check with an explicit hardcoded golden list of the 8 non-channel command names (`reset`, `verbose`, `info`, `context`, `queue`, `clear`, `kill`, `exit`), compared via `assert_eq!` the same way `channel_command_names_match_historical_set` already pins the channel-visible side. Verified red by temporarily adding `Scope::CHANNEL` to `/exit` — the rewritten test caught it (`left: {...without "exit"...} right: {...with "exit"...}`); the old version could not have, since both sides came from the same live filter.

DaBlitzStein:
Every one of the 5 review threads on this PR now has an inline reply, and the branch is at `3364504e9`.

houko (approved):
LGTM. All three findings are fixed, and the `/goal` one is fixed properly rather than by relaxing the assertion.

**`/goal` panicking the TUI.** This was not hypothetical and it is not new — `origin/main` today has `scope: Scope::CHANNEL.union(Scope::CLI).union(Scope::DASHBOARD)` on the `goal` entry and zero `"goal" =>` arms in `chat_runner.rs`, so `/goal ship the release` in `librefang chat` passes the pre-flight `commands::lookup`, falls to `unreachable!()` and panics with the terminal left in raw mode. This PR adds the arm with all three backend states handled, and the daemon case says plainly that no HTTP endpoint wraps `create_and_start_goal` yet instead of crashing or silently no-op-ing. `goal_command_does_not_panic_without_a_backend` exercises `Backend::None`, which is the state `StandaloneChat::new` starts in, so the test reproduces the original panic path rather than a nearby one.

**`/think` pinned to a non-existent Backend arm.** Now `dashboard_exec: None`, with a comment saying why (the dashboard toggles extended thinking through the `thinking` field on the message frame, not a slash command) and naming `/status` as the precedent. `/exit` gets the same treatment, and `dashboard_exec_matches_the_historical_chat_menu` explains in-comment why both are absent from its two lists rather than leaving a reader to wonder.

**The tautological guard.** `cli_scope_does_not_leak_into_channel_set` now compares against an explicit golden set instead of re-deriving `is_channel_command` from the field the filter selected on, and the doc-comment states exactly why the old form could not fail. Matching on "not channel-scoped" rather than "exactly `Scope::CLI`" is the right predicate too — it keeps covering `/clear` and `/exit` after they gained `Scope::DASHBOARD`.

Both changelog fragments are warranted here, unlike the two-fragment pattern I flagged on #8149 and #7995: the `added/` one describes the registry-derived slash menu and the `fixed/` one describes four bugs that exist on `main` independently of it. Both carry `(#7996)` in their trailing group, so neither picks up a duplicate generated line.

All checks green. The `BLOCKED` state is the stale `CHANGES_REQUESTED` from my earlier round, which this approval clears.

PR #7974, opened 2026-08-29
feat(tasks): validate assignee and enforce per-task priority and claim deadline

DaBlitzStein:
## Summary

- `task_post` rejects an unknown `assigned_to` with `AgentNotFound` instead of silently storing a row nothing can claim.
- `TaskPostOptions { priority, timeout_secs }` struct lets the API expose claim-queue ordering and a per-task TTL override.
- `priority` orders the claim queue (higher first, ties by age).
- `timeout_secs` overrides the global `claim_ttl_secs` for a single row: the stuck-task sweeper reads `COALESCE(timeout_secs, global)`.
- Schema migration v56 adds the `timeout_secs` column (NULL = inherit global).

## Migration numbering

Renumbered from v57 to v56 because main landed its own v55 (template version history, #8047).
`test_every_migration_records_audit_row` iterates `1..=user_version` and requires an audit row for every step, so gaps fail the suite — the branch takes the lowest free slot rather than skipping ahead and leaving a hole.
Two sibling PRs (#8041 manifest version history, #7991 parent session lineage) also carry a v56 in their own branches under the same constraint.
Whichever lands first keeps v56; the others renumber on their next rebase, so landing order alone resolves the collision.

## Post-time vs claim-time validation

The assignee check fires at post time only.
`task_claim` resolves the *claiming* agent against the registry (`crates/librefang-kernel/src/kernel/handles/task_queue.rs:89`) but never re-validates the stored `assigned_to` — it forwards the claimer's UUID and name to the substrate so rows stored as either shape match.
An unknown assignee therefore cannot be caught at claim time at all: the row can never match a claimer and, pre-PR, sat unclaimed forever.

Two consequences worth naming:

- A stopped agent is still a valid assignee.
  The registry lookup means "registered", not "running" — a task posted for a stopped agent waits for it to come back instead of being refused because the worker happened to be down at post time.
- A workflow that posts a task for an agent a *later* step spawns now gets a 400 until that agent is registered.
  Pre-PR the row was accepted and silently unclaimable, which is strictly worse; if this ordering turns out to be common the fix is spawn-before-post in the workflow definition, not re-opening the hole.

## Verification (refreshed onto current main)

- Branch refreshed onto current `origin/main` (merge `2eab5671a`), then renumbered 57→56 (`838b005ad`) and fixed merge drift in the `vision_gate.rs` test double, which needed the new `TaskPostOptions` parameter (`d0d79a393`).
[...]

Issue #1888, opened 2026-03-31
Workflow debugging and canvas/nodes enhancements


AIHunter83:
### Description

As per existing there is no any way to debug and test the workflow through canvas, and even if you try to run the flow from canvas it gives you generic error, so we meed the blow enhancements points
1- avail debugging option to workflow to be able to test your workflow 
2- logging input and output to be able to have a full picture 
3- provide detailed error to be able to find and solve the issue easily 
This is for time being 
For future I believe it will be wonderful if we leverage the workflow module and include many features and integrations nodes like N8N

PR #7997, opened 2026-08-29
feat(dashboard): workflow run timeline and task board improvements

DaBlitzStein:
## Summary

- **WorkflowsPage**: live 3s polling while a run is active — including while paused at an operator gate — stopping once the run reaches a terminal state; an execution timeline with a real progress bar and token counts; a fully translated execution-log console with auto-scroll (gated to stay near the bottom, not fighting a scrolled-up operator); `input_schema`-based parameter detection with a prompt-scan fallback; auto-populate params from the most recent run, merged onto the schema-seeded defaults rather than replacing them; a re-run control that pre-fills params, fires even when the stored run had no input, is disabled while a re-run is already in flight, and gives toast feedback on success/failure.
- **TasksPage**: agent picker sourced from the registry instead of from existing task assignees (a new agent is selectable before its first task; a deleted one stops being offered), UUID-to-name resolution for assignees, board counters that count the same task list the cards actually render from.
- **api/workflows backend**: `GET /api/workflows/runs/{run_id}` now reports `total_steps` (from the workflow definition's step count) and `current_step_index` (the completed-step count) — both were previously absent, so the dashboard's progress bar was permanently stuck reading "Starting…" at 0% no matter how far a run had actually gotten.
- **i18n**: the execution-log console is now fully translated across all 5 locales; dead locale keys removed (`tasks.field_assignee_placeholder`, and the priority/timeout keys for the inputs held back below).

## Held back or removed in this PR

- **Task priority and timeout controls.** The selector, field, and card badges from an earlier revision were entirely inert — `task_queue_post_root` and `MemorySubstrate::task_post` don't accept or persist either value yet. `#7974` (open) lands the backend — route, substrate, and a schema migration — for exactly this. Shipping the frontend here would duplicate that work and risk colliding with its migration slot, so it's removed until `#7974` lands.
- **Canvas/chat "Save-as-Workflow" scaffolding.** `workflowStepsToCanvasState`, `extractWorkflowJson`, and `CanvasNodeData.fresh` had no production call site — the chat action meant to call them doesn't exist in this PR. Removed along with their tests; land them together with that caller in a follow-up.
- **`WorkflowStepResult.variables`.** Added to the type but never emitted by the backend — `StepResult` in `librefang-kernel` has no such field, so the "Variables" section and per-step `{{k}} = "v"` console lines could never render. Wiring it means threading variable bindings through roughly 15 `StepResult` construction sites inside the engine's execution loop, in a different crate — left for a follow-up.

## Verification

- `cargo check -p librefang-api --lib` — clean
- `cargo test -p librefang-api --test workflow_lifecycle_test` — 20 passed, including `run_detail_reports_total_steps_and_current_step_index`
- `npx tsc --noEmit` (dashboard) — clean
- `LANG=en_US.UTF-8 npx vitest run` (dashboard) — 1810 passed, 205 files
- `node scripts/i18n-parity.mjs` — same pre-existing ko/pl/uk plural-suffix drift as `main` (issue #3557), no new drift
- `python3 scripts/check-changelog-attribution.py` — OK

## Test plan

- [ ] CI green
- [ ] Workflow run detail shows a live progress bar and token counts while executing, advancing past 0% as steps complete
- [ ] A failed step in the Run Result panel shows the error icon and message, not a green check
- [ ] Re-running a run whose stored input is blank still starts the new run
- [ ] Re-run control is disabled while a re-run for that row is already in flight
- [ ] Task board's total/status tiles match the number of cards actually rendered
- [ ] Agent filter dropdown is populated from the registry, not from existing tasks

## Issue references

- Refs #1888: partial coverage — workflow run timeline and task board improvements land here; task priority/timeout coverage deferred to #7974; canvas "Save-as-Workflow" coverage deferred to a follow-up.


houko (commented):
Three things, one of them a live bug.

**Polling never restarts after a paused run resumes.**

```tsx
const [runDetailPollMs, setRunDetailPollMs] = useState<number | false>(false);
useEffect(() => {
  if (!selectedRunId || selectedRunId === "") { setRunDetailPollMs(false); return; }
  setRunDetailPollMs(3000);
}, [selectedRunId]);
useEffect(() => {
  const state = runDetailQuery.data?.state;
  if (state && state !== "running" && state !== "pending") setRunDetailPollMs(false);
}, [runDetailQuery.data?.state]);
```

The only thing that turns polling back on is a change of `selectedRunId`.
`paused` is not terminal — the comment above lists it as though it were, but the pending-operator-reviews banner exists precisely so an operator can select a paused run, review it, and resume it.
On that path the run goes `paused` (polling stops), the operator resumes, and the timeline sits frozen on stale data until they click away and back.

The function form of `refetchInterval` collapses both effects and fixes it, because it is re-evaluated on every fetch rather than latched:

```tsx
const runDetailQuery = useWorkflowRunDetail(selectedRunId ?? "", {
  enabled: !!selectedRunId,
  refetchInterval: (query) => {
    const s = query.state.data?.state;
    return s === undefined || s === "running" || s === "pending" ? 3000 : false;
  },
});
```

`s === undefined` covers the "poll while waiting for the first fetch" case the first effect was written for.

**The priority / timeout UI lands before the backend that honours it.**

The body says it plainly: "works immediately for display; full backend enforcement lands with #7974."
#7974 is blocked on `OpenAPI Drift` right now, so if this merges first, the task modal offers a priority selector and a timeout field, accepts what the operator types, reports a successful create — and the queue claims tasks in the old order regardless.
A control that takes input and discards it is worse than no control, because the operator has no way to tell.

Either land #7974 first (its drift fix is one regenerated `openapi.sha256` — see my comment there), or ship these two fields disabled with a note until it does.

**This is three PRs in one.**

WorkflowsPage (timeline, polling, `input_schema` params), TasksPage (agent picker, priority, timeout), and the canvas helpers (`fresh`, `workflowStepsToCanvasState`, `extractWorkflowJson`) share nothing but a language.
`CLAUDE.md` asks for one PR per issue for exactly this reason: the canvas helpers look independently mergeable today, and they are currently gated behind review of 1184 lines of unrelated UI.

Two things that are right and worth saying: all five locales are updated including `pl.json` (#8013 is red on missing exactly that), and preferring the workflow's declared `input_schema` over scanning prompts for `{{var}}` — with the scan kept as fallback — is the correct precedence, since the schema carries type, default and description that no scan can recover.

houko (changes requested):
## Blocking

**1. `build` is red: the icon is missing from the barrel.**

```
[MISSING_EXPORT] "ArrowUpNarrowWide" is not exported by "src/lib/lucide.ts".
```

Add it to `src/lib/lucide.ts`. (#8067 is red on the same thing for `Ban` — worth checking whether the barrel should re-export the whole `lucide-react` surface instead of curating it, since this is now the second PR to trip on it in a week.)

**2. `extractWorkflowJson` captures the fence's language tag.**

```ts
const fenced = content.match(/```(?:json)?\s*\n?([\s\S]*?)\n?```/);
if (fenced) return fenced[1];
```

`(?:json)?` is optional and nothing anchors what follows, so for

````
```rust
{"steps": []}
```
````

`(?:json)?` matches empty, `\s*` matches empty (the next character is `r`), `\n?` matches empty, and the capture group starts at `rust\n{…}`. The caller then gets a string that cannot parse. Two narrower shapes fix it: require the tag to be `json` or absent up to the newline (`/```(?:json)?[ \t]*\r?\n([\s\S]*?)\r?\n?```/`), or scan every fence and return the first whose contents parse.

Related: the regex takes the **first** fence in the message, so a reply that explains itself with a shell or TOML block before the JSON one picks the wrong block. Preferring the first fence that parses handles both cases at once.

**3. No test for it, in a file that already has one.**

`crates/librefang-api/dashboard/src/lib/chat.test.ts` exists and is not touched by this PR. `extractWorkflowJson` is a hand-written brace/string-state scanner with escape tracking — exactly the sort of function that needs cases: a plain ```` ```json ```` fence, a bare fence, a fence tagged with another language, a message with prose containing `{` before the real object, an object with a `"}"` inside a string, an escaped `\"` inside a string, and no JSON at all. Note that `changelog.d/fixed/6943-workflow-json-extraction-regex.md` (in #7988) records this extraction already being fixed once for a regex defect — a second pass at it without tests is likely to repeat that.

## Non-blocking

- The brace-matching fallback starts at the first `{` anywhere in the message. Prose that mentions `{{name}}` placeholders — which this codebase's workflow templates use — will start the scan there and consume to a matching brace that means nothing. Worth requiring the fallback to parse before returning it, rather than returning whatever the scanner balanced.
- No changelog fragment (`changelog.d/added/7997-....md`). A workflow run timeline and task-board changes are user-visible.
- `canvas.test.ts`, `TasksPage.test.tsx` and `WorkflowsPage.test.tsx` are all extended here, which is good — the gap is specifically `chat.ts`.

**Comment by DaBlitzStein** [comment]
Addressed the review feedback on the red build.

Re-exported `ArrowUpNarrowWide` from the lucide barrel (path verified against the existing `SortAsc` entry).
Reworked `extractWorkflowJson` to scan every fenced block and return the first whose body parses as JSON, so a leading non-json fence (e.g. `rust`) no longer wins and returns garbage.
Added eight tests in `chat.test.ts` covering json-tagged, untagged, non-json-tagged, and no-fence inputs.

Dashboard `node_modules` isn't installed in the worktree, so vitest/tsc couldn't run locally; the regex logic was exercised against 12 cases in Node directly, and the icon path matches the pinned lucide-react internal path already used by `SortAsc`.
Commit 79c52d256.

**Comment by DaBlitzStein** [comment]
Addressed all three blocking points in `79c52d256` and `5a1de5d96`:

1. **MISSING_EXPORT `ArrowUpNarrowWide`** — added the re-export to `src/lib/lucide.ts`. The mapping points at `arrow-up-narrow-wide.mjs`, the same path the existing `SortAsc` entry already resolves.

2. **`extractWorkflowJson` fence parsing** — rewrote it to walk every fenced block (`matchAll`) and return the first whose tag is empty or `json` AND whose body passes `JSON.parse`. An explanatory shell/TOML block ahead of the JSON block is skipped, and a prose fence is skipped because it will not parse. This is the "scan every fence, return the first that parses" shape.

3. **Missing test** — added eight vitest cases in `chat.test.ts` covering the json-tagged fence, untagged fence, both skip paths, and the brace-matching fallback.

Also added the missing changelog fragment (`changelog.d/added/7997-workflow-run-timeline.md`).

Verification: `pnpm typecheck` clean; `pnpm test` 1739/1740 (the one failure is an unrelated `PromptsExperimentsModal.test.tsx` timeout).

**Review by houko — DISMISSED** [review]
All three blocking points resolved.

**1 — `ArrowUpNarrowWide` is in the barrel**, and `CI Gate` is green with zero non-success checks, so `build` has cleared.

**2 — the fence extraction is fixed on both halves**, and the rewrite is the option I hoped for rather than the minimal one:

```ts
for (const fence of content.matchAll(/```([^\n]*)\n([\s\S]*?)```/g)) {
  const lang = fence[1].trim().toLowerCase();
  if (lang && lang !== "json") continue;
```

Capturing the tag up to the newline kills the original defect — a ```` ```rust ```` fence can no longer have `rust\n` swallowed into the body — and iterating *every* fence with a `JSON.parse` probe fixes the separate first-fence-wins problem in the same pass. Those were two bugs and this is one mechanism for both.

**3 — eight cases in `chat.test.ts`**, covering the json-tagged fence, the untagged fence, both skip paths, a json-tagged fence whose body does not parse, and the fallback. `recovers a mislabeled fence's JSON body through the brace-matching fallback` is a good one I did not think to ask for: a ```` ```rust ```` fence containing real JSON is skipped by the loop and then recovered, so the stricter tag matching does not lose a payload.

Changelog fragment added.

## Two gaps, both non-blocking, stated precisely because you declined one of them

**The fallback's string-state and escape tracking is still untested.** Two of the seven cases I listed — an object with `"}"` inside a string, and an escaped `\"` inside a string — have no coverage, and I confirmed that by grepping the test bodies rather than the names. That scanner is now the least-exercised code in the function, because the fence loop short-circuits every well-formed input ahead of it, while being the only part with hand-maintained `depth` / `inString` / `escaped` state.

I am not blocking on it: the scanner is pre-existing and this PR leaves it unchanged, so asking you to cover it is asking for tests on code you did not touch. Worth knowing that `changelog.d/fixed/6943-workflow-json-extraction-regex.md` records this extraction being fixed once before, which is why I keep pointing at it.

**The fallback still fails to find JSON that follows brace-y prose**, and "keeps the pre-existing fallback unchanged" is scope discipline rather than a claim that it is fine — so here is the concrete shape, since a `JSON.parse` probe alone would not fix it:

> `Use {{name}} in your prompt. Here is the workflow: {"steps": []}`

`indexOf("{")` lands on `{{name}}`, the scanner balances at `}}`, the candidate is `{{name}}`, the `"steps"` guard rejects it, and the function returns `null` — never reaching the real object. It fails closed rather than returning garbage, which is the right direction, but it does drop a valid payload. The fix is to retry from the next `{` rather than to validate the first candidate.

`mergeable` is `BLOCKED` on review rather than a conflict, so this one escaped the rebase churn my merge sweep caused elsewhere.

Approving.

houko (changes requested):
Coming back to the polling block at head `5a1de5d9` — I approved on the extraction fixes and did not read this closely enough. Two bugs in it, both in the same twelve lines, and one fix covers both.

```ts
const [runDetailPollMs, setRunDetailPollMs] = useState<number | false>(false);
const runDetailQuery = useWorkflowRunDetail(selectedRunId ?? "", {
  refetchInterval: runDetailPollMs,
  enabled: !!selectedRunId && selectedRunId !== "",
});
useEffect(() => {
  if (!selectedRunId || selectedRunId === "") { setRunDetailPollMs(false); return; }
  setRunDetailPollMs(3000);
}, [selectedRunId]);
useEffect(() => {
  const state = runDetailQuery.data?.state;
  if (state && state !== "running" && state !== "pending") { setRunDetailPollMs(false); }
}, [runDetailQuery.data?.state]);
```

## 1. `paused` is not terminal, and this stops polling on it

The comment says the poll disables "once we see a terminal state (completed/failed/cancelled/paused)", and the allowlist implements that — anything that is not `running` or `pending` stops the interval.

But a paused run resumes. `WorkflowRunState::Paused` (`crates/librefang-kernel/src/workflow.rs:1231-1235`) is documented as *"the run was paused mid-execution and is waiting for an external signal (an approval, a human-supplied input, etc.) before it continues"*, and the API flattens it to the string `"paused"` at `routes/workflows/workflow.rs:184`. The terminal set is `completed` / `failed` / `cancelled`.

So the timeline freezes at exactly the moment it is most worth watching: an operator approves the gate, the run resumes, and this page keeps rendering the paused snapshot until someone reloads. The page's own code knows paused runs are first-class here — the comment a few lines down says the pending-operator-reviews banner selects paused runs from anywhere in the history. Those are the runs a person is sitting and watching.

## 2. Switching between two runs in the same state leaves the interval running forever

The second effect keys on `runDetailQuery.data?.state`. Select run A, it finishes, dep becomes `"completed"`, interval off. Now select run B, also completed, whose detail React Query already has cached (previously viewed, so `data` is available on the first render after the key change): the first effect fires on `selectedRunId` and sets 3000, while the second effect's dependency goes `"completed"` → `"completed"`, compares equal, and does not run. Nothing ever turns the interval back off, so the page polls a finished run every 3s for as long as it stays open.

The uncached path self-heals — `data` is briefly `undefined`, so the dep changes twice and the second change stops it — which is why this is easy to miss in manual testing and why neither `WorkflowsPage.test.tsx` case catches it.

## Both go away by not mirroring the state

The interval is a pure function of the data the query already holds, so deriving it removes the two effects, the extra `useState`, and the render where the two disagree:

```ts
const TERMINAL_RUN_STATES = new Set(["completed", "failed", "cancelled"]);
const runDetailQuery = useWorkflowRunDetail(selectedRunId ?? "", {
  refetchInterval: TERMINAL_RUN_STATES.has(runDetailQuery.data?.state ?? "") ? false : 3000,
});
```

(or TanStack's function form, `refetchInterval: (query) => TERMINAL_RUN_STATES.has(query.state.data?.state ?? "") ? false : 3000`, which avoids referring to the query while defining it). A run whose state has not loaded yet polls, which is the behaviour the current code reaches for with "begin polling immediately while we wait for the first fetch" — here it falls out instead of needing an effect.

Worth a test for each: a run that goes `paused` keeps polling, and selecting a second already-cached completed run does not start one.

## Nit

`enabled: !!selectedRunId && selectedRunId !== ""` is redundant twice over — `!!""` is already false, and `requireId` in `lib/queries/workflows.ts:33-38` already computes `enabled: !!id && options.enabled !== false` for this hook. Passing it here just gives a second place to keep in sync.

## Not a problem, since I checked

- `QueryOverrides` (`lib/queries/options.ts`) does carry `refetchInterval` through `withOverrides`, so the plumbing works; I went looking because `lib/queries/` is not in this diff.
- No background-tab concern: TanStack's `refetchIntervalInBackground` defaults to `false`, so this does not need the explicit setting `analytics.ts` carries for #3393.
- The two items I left non-blocking on the approval (the fallback scanner's untested string/escape state, and brace-y prose defeating `indexOf("{")`) are unchanged and still non-blocking — this review is only about the polling block.

DaBlitzStein:
Pushed `d8fdbea16`: merged current `main`.

**No hand-written merge resolution to review** — `git merge-tree --write-tree` reports zero difference from git's automerge.

- `tsc --noEmit` — exit 0.
- Full dashboard suite — 1821 passing, 1 failing.
- The single failure is `PromptsExperimentsModal > stops selection when every variant has a traffic bucket`, `Test timed out in 5000ms` after 6427 ms. It is a load-dependent timeout, not a defect here: this branch does not touch that file, and re-running it alongside this PR's own four test files gives **5 files / 127 tests passing in 4.56 s**. The modal passes as soon as it is not competing for CPU.

Dashboard-only: TasksPage, WorkflowsPage, `lib/canvas.ts`, `lib/chat.ts`, `lib/lucide.ts`, locales.

DaBlitzStein:
Addressed at `c8e2a3d53`. Both bugs, and by deriving rather than mirroring as you suggested.

`TERMINAL_RUN_STATES` is `completed` / `failed` / `cancelled` — `paused` is deliberately absent, with the reason on the constant: a paused run is waiting on an external signal and resumes, which is precisely when the timeline is worth watching. The interval comes from TanStack's functional form reading `query.state.data`, so both `useEffect`s and the mirrored `useState` are gone and there is no render where the two disagree. Net −5 lines in the page.

The redundant `enabled` is gone too, since `requireId` already computes it.

Both tests you asked for, each failing first. Putting `paused` back in the terminal set:

```
× keeps polling a run that transitions from running to paused (#7997)
  expect(pollIntervalFor(runDetail("paused"))).toBe(3000)
```

and restoring the two effects fails `does not poll after selecting a second, already-cached completed run (#7997)` — the case that self-heals when the query is uncached, which is why manual testing never caught it. 22 passed with the fix.

One note on the plumbing, since `lib/queries/` was not in the original diff: `QueryOverrides` needed a `TData` parameter so the functional `refetchInterval` can read `query.state.data` without a cast. It defaults to `unknown`, so every existing caller is unchanged. The callback parameter is typed as the structural slice of `Query` a polling predicate actually reads rather than `Query` itself — `Query` is invariant in its query-key type, so naming it would reject every hook whose factory uses a tuple key.

Ready for another look.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
```diff
@@ -598,6 +644,43 @@ export function WorkflowsPage() {
     }
   };
 
+  // Re-run a previous workflow run with its original params pre-filled.
+  const handleRerun = (run?: { id?: string; input?: string }) => {
+    const runInputStr = run?.input;
```
houko, line 675: `handleRerun` early-returns before the re-run fires whenever the stored input is empty.

```ts
const runInputStr = run?.input;
if (!runInputStr) return;
```

`buildRunInput()` returns the raw `runInput` string when the workflow declares no `{{var}}` placeholders, so a parameterless workflow launched with an empty textarea stores `""` (or nothing) as the run input. `POST /api/workflows/runs/{id}/rerun` happily repeats such a run — it only reads the original run's `input` and starts a fresh run — but the button is now inert: no mutation, no toast, not even the scroll. The previous `handleRerun(runId)` fired the mutation regardless of input.

Suggest hoisting the pre-fill into a guarded block and always calling the mutation when `run?.id` is present.

houko (commented):
Correctness review of the dashboard changes. 15 findings, most severe first: two features that cannot work because the backend fields they read do not exist, a dropped step-error surface, an inert re-run button, and an unannounced revert of #3412. Details inline.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
```diff
@@ -1284,56 +1374,236 @@ export function WorkflowsPage() {
                                 {t("workflows.from_review_banner", { defaultValue: "from review banner" })}
                               </span>
                             )}
+                            {/* Re-run control — a span, not a button: the row itself is a
+                                <button> and nesting buttons is invalid HTML (hydration error). */}
+                            <span
+                              role="button"
```
houko: The re-run control lost its in-flight guard, so the mutation can be fired repeatedly.

The control it replaces was `disabled={rerunMutation.isPending}` with a spinning icon; a `<span role="button">` cannot be disabled and nothing else gates it. Double-clicking the Play icon (or holding Enter while it has focus) posts `/rerun` once per event and starts that many duplicate workflow runs, each consuming real LLM tokens. The success toast is gone too, so there is no feedback that the first click did anything.

Separately, the comment above this block has the HTML rule backwards. `<button>`'s content model forbids interactive-content descendants, and an element carrying `tabindex` **is** interactive content — so nesting this span inside the row `<button>` is exactly the invalid markup the removed code avoided by keeping the control a sibling (`{/* A sibling of the row button, never nested inside it, so it stays a valid standalone control. */}`). It also lands inside the row button's accessible name.

Gate on `rerunMutation.isPending` and move the control back out of the row button.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
```diff
@@ -718,9 +801,7 @@ export function WorkflowsPage() {
     <button
       className="w-full flex items-center gap-2 px-3 py-2 text-left hover:bg-surface transition-colors"
       onClick={toggle}>
-      {step.error
-        ? <AlertCircle className="w-3 h-3 text-error shrink-0" />
-        : <CheckCircle2 className="w-3 h-3 text-success shrink-0" />}
+      <CheckCircle2 className="w-3 h-3 text-success shrink-0" />
```
houko: A failed step now renders as a green check with its error text nowhere on screen.

This header dropped the `step.error ? <AlertCircle .../> : <CheckCircle2 .../>` branch, and `StepResultContent` (line ~210) dropped the error panel that went with it — both landed in #6324. The new run-history timeline re-implements them, but `stepResultHeader` / `stepResultContent` are still the renderers for the **Run Result** panel (line ~1250, `getRunStepResults(runMutation.data)`).

Scenario: run a workflow from the right-hand Run panel; step 2 fails with a populated `error`. The panel lists step 2 with a success icon, `Output: (empty)`, and no error anywhere — `StepResultContent` renders only `step.output || "(empty)"`. The operator sees a run that looks like it succeeded.

Restore both the icon branch and the error block, or point that panel at the new timeline renderer.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
```diff
@@ -1284,56 +1374,236 @@ export function WorkflowsPage() {
...
-                          {/* Inline run detail */}
-                          {isSelected && runDetailQuery.data && (
-                            <div className="ml-5 mt-1 space-y-1.5">
-                              {runDetailQuery.data.error && (
+                          {/* Inline run detail — execution timeline */}
+                          {isSelected && runDetailQuery.data && (() => {
+                            const rd = runDetailQuery.data;
+                            const isActive = rd.state === "running" || rd.state === "pending";
+                            const allSteps = rd.step_results;
+                            const totalSteps = rd.total_steps || allSteps.length;
```
houko, line 1454: The live progress bar can never advance: no backend route emits `total_steps` or `current_step_index`.

`GET /api/workflows/runs/{run_id}` (`crates/librefang-api/src/routes/workflows/workflow.rs:790`) serialises exactly `id, workflow_id, workflow_name, input, state, output, error, started_at, completed_at, owner_agent_id, step_results`, and neither identifier exists anywhere in `crates/` (`grep -rn 'total_steps\|current_step_index' crates --include='*.rs'` → no hits).

So for every real run: `curIdx` is `undefined` → the banner always reads `workflows.starting` ("Starting…") and the bar is pinned at `width: 0%`, even at step 4 of 5; `totalSteps` falls back to `allSteps.length`, which makes `totalSteps > allSteps.length` permanently false, so the "Pending…" rows never render either. `step.variables` is in the same position — `StepResult` in `crates/librefang-kernel/src/workflow.rs:1375` has no `variables` field and the route does not emit one — so the Variables section and the `{{k}} = "v"` console lines are dead.

An operator watching a running workflow reads a 0% bar and "Starting…" for the whole run, which is worse than showing nothing. Either land the three fields on the run-detail route in this PR, or render the progress affordance only when `current_step_index` is actually present.

crates/librefang-api/dashboard/src/pages/TasksPage.tsx
```diff
@@ -326,14 +412,24 @@ function NewTaskModal({ isOpen, onClose, agents }: NewTaskModalProps) {
...
 
   function handleSubmit(e: React.FormEvent) {
     e.preventDefault();
-    if (!title.trim() || !description.trim()) return;
+    if (!title.trim() || !description.trim() || timeoutInvalid) return;
     createMutation.mutate({
       title: title.trim(),
       description: description.trim(),
-      ...(assignee.trim() ? { assigned_to: assignee.trim() } : {}),
+      ...(assignee ? { assigned_to: assignee } : {}),
```
houko, line 370: `priority` and `timeout_secs` are dropped on the floor by the API — the selector and the field are inert.

- `task_queue_post_root` (`crates/librefang-api/src/routes/task_queue.rs:261`) reads only `title`, `description`, `assigned_to`, `created_by` from the body, then calls `task_post` with those four.
- `MemorySubstrate::task_post` (`crates/librefang-memory/src/substrate.rs:1112`) hardcodes the column: `VALUES (?1, ?2, ?3, ?4, 'pending', 0, ...)`.
- `task_queue` has no `timeout_secs` column at all (`crates/librefang-memory/src/migration.rs:433`).

Scenario: an operator opens New Task, selects **Urgent**, sets a 90 s limit, submits. The POST succeeds, the task is created at `priority = 0` with no timeout, the card comes back with neither badge, and nothing tells the operator their choices were discarded. The card badges added above can therefore never fire in production either.

The unit tests here assert against a mocked `useCreateTask`, so they cannot catch this. Either land the backend fields in this PR, or hold the two inputs back until #7974 — shipping a control that silently does nothing is worse than not shipping it.

crates/librefang-api/dashboard/src/pages/TasksPage.tsx
```diff
@@ -464,19 +600,22 @@ export function TasksPage() {
...
+  const agentsById = useMemo(
+    () => new Map(agents.map((a) => [a.id, a])),
+    [agents],
+  );
 
   // Apply agent filter
   const filteredTasks = agentFilter
-    ? validTasks.filter((t) => t.assigned_to === agentFilter)
-    : validTasks;
+    ? allTasks.filter((t) => taskMatchesAgent(t, agentFilter, agentsById))
```
houko: Switching the board from `validTasks` to `allTasks` makes the counters disagree with the cards.

`validTasks` (line 599) exists to drop rows whose `id` is missing or empty. Everything downstream of `filteredTasks` now sees the unfiltered list: `getColumnTasks` (622), `countStatus` (650), the `status_total` tile (653) and the per-column `count` all include id-less tasks — but `KanbanColumn` still renders `tasks.filter((task) => Boolean(task.id))`, so those tasks produce no card.

Scenario: the queue returns one task with `id: ""` in `pending`. The Pending column header reads `1`, the total tile reads `1`, and the column body shows neither a card nor the `tasks.empty` placeholder (the `tasks.length === 0` test is against the unfiltered array). `validTasks` is left referenced only by the empty-state check at 773, which then disagrees with the tiles.

Filter once, at the top: keep `validTasks` as the input to both the agent filter and the unfiltered branch.

crates/librefang-api/dashboard/src/locales/en.json
```diff
@@ -4172,7 +4180,22 @@
     "status_pending": "Pending",
     "status_in_progress": "In Progress",
     "status_completed": "Completed",
-    "status_failed": "Failed"
+    "status_failed": "Failed",
+    "priority_urgent": "Urgent",
```
houko: This section gains keys but leaves a dead one behind: `tasks.field_assignee_placeholder` will fail the locale audit.

Replacing the assignee text input with a registry-backed `<select>` removed the only reference (`grep -rn field_assignee_placeholder src/` now hits nothing outside `src/locales/`), but the key is still defined in all five locale files (`en.json:4172`, `pl.json:4182`, `uk.json:4182`, `zh.json:4172`, `ko.json:4171`).

`src/lib/__tests__/en-locale-coverage.test.ts` → "does not carry dead keys in locale files" flattens every locale and asserts the unreferenced set is empty; no `tasks.*` dynamic `t(\`...\`)` pattern exists to rescue it. That is five entries in the failure list, so this is a red CI check rather than a nit — the same audit is why the three `workflows.rerun*` keys were correctly removed in this PR.

Drop `field_assignee_placeholder` from all five locales.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
```diff
@@ -528,6 +539,60 @@ export function WorkflowsPage() {
...
+        const values: Record<string, string> = {};
+        let hasKnownParam = false;
+        for (const p of detectedParams) {
+          if (parsed[p.name] !== undefined) {
+            values[p.name] = String(parsed[p.name]);
+            hasKnownParam = true;
+          }
+        }
+        if (hasKnownParam) {
+          setParamValues(values);
```
houko: Auto-populate replaces the whole `paramValues` map, discarding the schema defaults seeded a moment earlier.

The seeding effect at 482 fills `paramValues` from `p.default` for every declared parameter; this effect then calls `setParamValues(values)` with only the keys the most recent run happened to carry.

Scenario: a workflow declares `region` (default `"eu-west"`) and `topic`. The last run passed only `topic`. On selection the seed puts `{region: "eu-west"}` in the form, then this effect overwrites it with `{topic: "..."}` — `region` renders empty and, because `buildRunInput` skips blank values, the run goes out with no `region` at all. The authored default is silently lost.

Merge instead of replace: `setParamValues((prev) => ({ ...prev, ...values }))`.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
```diff
@@ -528,6 +539,60 @@ export function WorkflowsPage() {
...
+          if (typeof parsed.input === "string") setRunInput(parsed.input);
```
houko, line 603: Auto-populate can overwrite free-text the operator has already typed.

`paramTouchedRef` is only ever set by the parameter fields (`onChange` at line 1163). The free-text textarea at 1166 is plain `setRunInput`, so typing there leaves `paramTouchedRef.current === false` and this effect is still armed.

Scenario: select a workflow and start typing into the additional-context textarea while `useWorkflowRuns` is still in flight (or before `useWorkflowDetail` settles — the effect bails on `isFetching` and re-runs when it clears). The moment the run history resolves, `setRunInput(parsed.input)` replaces what the operator wrote with the previous run's input. Same effect fires on a refetch-on-focus if the first pass was skipped by the `isFetching` guard.

Set `paramTouchedRef.current = true` from the textarea's `onChange` as well.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
```diff
@@ -528,6 +539,60 @@ export function WorkflowsPage() {
     }
   }, [allWorkflows, workflows, selectedWorkflowId, workflowsQuery.isSuccess]);
 
+  // Auto-scroll log console to bottom when new step results arrive.
+  useEffect(() => {
+    if (logConsoleRef.current) {
+      logConsoleRef.current.scrollTop = logConsoleRef.current.scrollHeight;
```
houko: The console auto-scroll fights the operator on every poll tick.

The dependency is `runDetailQuery.data?.step_results` — an array freshly built by `JSON.parse` on each response, so its identity changes every 3 s even when the content is byte-identical. The effect therefore re-runs and force-sets `scrollTop = scrollHeight` on each tick.

Scenario: a run is executing; the operator scrolls the console up to read step 1's prompt; three seconds later the poll lands and the console snaps back to the bottom. Repeats until the run finishes.

Gate on a value that only changes when a line is actually appended (`step_results.length`), and skip the scroll when the pane is not already near the bottom.

changelog.d/fixed/7997-workflow-run-detail-poll.md
```diff
@@ -0,0 +1,3 @@
+Keep the dashboard's workflow run timeline live while a run sits paused at an operator gate, instead of freezing it at the moment someone is most likely to be watching.
```
houko: This fragment describes a bug that never shipped, and it goes into the release notes verbatim.

`main` has no run-detail poll at all — `git show origin/main:crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx` line 392 is `const runDetailQuery = useWorkflowRunDetail(selectedRunId ?? "");`, with no `refetchInterval` and no mirrored state anywhere in the file. The "froze at a paused gate", "treated every state other than `running` and `pending` as terminal" and "left a 3-second refresh running against a finished run" failures are all properties of an earlier iteration of *this branch*, not of any released version. A reader of the release body will go looking for a regression they never had.

The polling is new behaviour: fold it into the `added/` fragment (and drop this one), or rewrite it to describe what shipped.

While here, the `added/` fragment says "The task board gains canvas helpers and richer run status" — the task board gains neither; the canvas helpers are in `src/lib/canvas.ts` and the run status is on the Workflows page.

crates/librefang-api/dashboard/src/lib/chat.ts
```diff
@@ -207,6 +207,55 @@ export interface TerminalFrame {
  * `silent_complete` removes the owning bubble; `error` marks it with the failure.
  * Pure so the #6390 race is unit-testable without mounting the chat page.
  */
+
+export function extractWorkflowJson(content: string): string | null {
```
houko: This insertion point orphans the doc comment above it.

The `/** Apply a *foreign* terminal frame … Pure so the #6390 race is unit-testable without mounting the chat page. */` block was written for `applyForeignTerminalFrame`, which now sits ~50 lines further down with nothing above it. As written, that block documents `extractWorkflowJson`, whose behaviour it describes not at all.

Move the new function below `applyForeignTerminalFrame` (or above the comment) so the block stays attached to its function.

The same thing happened in `WorkflowsPage.tsx:189`: `TERMINAL_RUN_STATES` was inserted between `runStateKind`'s `#5257 round-2` comment and `runStateKind` itself.

crates/librefang-api/dashboard/src/lib/canvas.ts
```diff
@@ -372,3 +373,63 @@ export function removeEdgeById<E extends Edge>(edges: E[], edgeId: string): E[]
...
+
+export type WorkflowCreateStepInput = {
+  name?: string;
+  prompt_template?: string;
+  agent?: string | { id?: string; name?: string; type?: string; fresh?: boolean };
+  depends_on?: string[];
+  [key: string]: unknown;
+};
+
+export function workflowStepsToCanvasState(
```
houko: `workflowStepsToCanvasState`, `extractWorkflowJson` and `CanvasNodeData.fresh` have no production call site.

`grep -rn 'extractWorkflowJson\|workflowStepsToCanvasState' src/ | grep -v '\.test\.'` returns only the two definitions — the "Save-as-Workflow chat action" the PR body describes is not wired to anything, so the 135 new lines and their tests ship as dead code.

`fresh` is worse than unused: this function is the only writer, and `stepAgentPayload` (line 292) emits exactly one of `agent_id` / `agent_type` / `agent_name` and never `fresh`. So even once a caller exists, importing a workflow whose step declares `agent.fresh` and saving it back drops the flag — the find-or-spawn binding the field is for does not survive a round-trip. It is also the only field in `CanvasNodeData` without a doc comment, wedged between `agentType`'s comment and `binding`'s.

Either land the chat action that calls these in this PR, or hold them until the caller lands.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
```diff
@@ -489,19 +507,12 @@ export function WorkflowsPage() {
...
-  // step. Fires once per mount; if the user manually flips back to
-  // "My Workflows", we don't override on the next refetch.
-  const autoSwitchedRef = useRef(false);
-  useEffect(() => {
-    if (autoSwitchedRef.current) return;
-    if (!workflowsQuery.isSuccess) return;
-    autoSwitchedRef.current = true;
-    if ((workflowsQuery.data ?? []).length === 0) setActiveTab("templates");
-  }, [workflowsQuery.isSuccess, workflowsQuery.data]);
+  // Show the "My Workflows" empty state on first visit so users see
```
houko: This reverts #3412 and inverts its regression test, with no issue reference and no mention in the PR body.

`45c407e3c feat(dashboard/workflows): land on marketplace tab when no workflows (#3412)` added the `autoSwitchedRef` effect that this hunk deletes, and `WorkflowsPage.test.tsx` "auto-switches to templates tab when there are no workflows" was its guard — this PR rewrites that test to assert the opposite. What is left here is a comment describing behaviour that no code implements.

Reverting a merged UX decision is a maintainer call, and it is unrelated to a workflow run timeline. Please split it out and make the case in its own PR (or on #3412), rather than carrying it inside a feature PR whose description does not mention it.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
```diff
@@ -1284,56 +1374,236 @@ export function WorkflowsPage() {
...
-                                renderContent={stepResultContent}
-                              />
+                              {/* Execution log console */}
+                              {(() => {
+                                const fmtTime = (ts?: string | null) => ts ? new Date(ts).toLocaleTimeString() : "--:--:--";
+                                const fmtDur = (ms: number) => ms >= 60000 ? `${(ms/60000).toFixed(1)}m` : `${(ms/1000).toFixed(1)}s`;
+                                const fmtN = (n: number) => n >= 1000 ? `${(n/1000).toFixed(1)}k` : String(n);
+                                const logs: Array<{ts: string; level: "info"|"warn"|"error"; msg: string}> = [];
+                                // Run start
+                                logs.push({ts: fmtTime(rd.started_at), level: "info", msg: `Run started — ${totalSteps || "?"} steps defined`});
```
houko: Most of the console is hardcoded English, and it is the one surface that cannot fall back to the UI language.

Three lines go through `t()` (`console_completed`, `console_failed`, `console_executing`), and the rest are template literals: `Run started — N steps defined`, `Step i/N "name" → agent FAILED`, `  Prompt: … tokens → Response: … tokens in …`, `  Error: …`, `  {{k}} = "v"`. A `zh` / `ko` / `pl` / `uk` operator gets a console that is half translated and half English.

The locale-coverage test does not catch these because it only inspects JSX text and literals on user-facing property names — `msg` is not in `isUserFacingPropertyName`, so template literals assigned to it are invisible to it. That is a gap in the guard, not permission.

Route the remaining lines through `t()` with interpolation, the way the three completion lines already are.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
DaBlitzStein: Fixed. The blank-input early return now only guards the parameter pre-fill, not the mutation: `WorkflowsPage.tsx:679` wraps the pre-fill in `if (runInputStr) { ... }`, and the mutation at `:703` (`if (run?.id) { void rerunMutation.mutateAsync(...) }`) fires unconditionally whenever the row has a run id, blank input or not. Regression test: `WorkflowsPage.test.tsx` "re-runs a parameterless run with a blank stored input (#7997)" — verified red against the old `if (!runInputStr) return;` (0 calls to `mutateAsync`), green now.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
DaBlitzStein: Fixed. The control is now a real `<button type="button" disabled={rerunMutation.isPending}>` (`WorkflowsPage.tsx:1433-1445`), a sibling of the row `<button>` inside a `<div className="flex items-stretch gap-1">` wrapper (`:1370`) rather than a `<span role="button">` nested inside it — matches the HTML-validity comment at `:1426-1431`. Added a regression test for the `isPending` guard ("disables the re-run control while a re-run is already in flight") — verified red by temporarily removing the `disabled` attribute (`Received element is not disabled`), green with it restored. Also restored the success toast and the `workflows.rerun_started`/`workflows.rerun_failed` locale keys that an earlier pass of this fix had accidentally dropped along with the old error-only catch handler — a re-run now gives the operator a visible signal even though the row scrolls away from the run-history list.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
DaBlitzStein: Fixed. `stepResultHeader` restores the `step.error ? AlertCircle : CheckCircle2` branch (`WorkflowsPage.tsx:843-845`), and `StepResultContent` restores the error panel (`:211-216`) — this is the renderer for the Run Result card's step list (mounted at `:1349` via `getRunStepResults(runMutation.data)`), so a failed step now shows the red icon and the error text again instead of a green check with no error anywhere.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
DaBlitzStein: Fixed both fields, and removed the third dead one. `GET /api/workflows/runs/{run_id}` now returns `total_steps` (from the workflow definition's step count, `workflow.rs:797-800`) and `current_step_index` (the completed-step count, `:801-806`) — both typed on `WorkflowRunDetail` in `api.ts:2942-2943`. New integration test `run_detail_reports_total_steps_and_current_step_index` (`workflow_lifecycle_test.rs`) runs a real 2-step workflow to completion and asserts both fields — verified it fails against a build without the route change (both would be absent/`null`) and passes now (`cargo test -p librefang-api --test workflow_lifecycle_test`: 20 passed).

Separately: `step.variables` (the third dead affordance in this comment) has been removed rather than wired — `StepResult` in `librefang-kernel` has no such field and the route never emitted it, so the "Variables" section and the per-step `{{k}} = "v"` console lines could never render. Wiring it would mean threading variable bindings through roughly 15 `StepResult` construction sites in the engine's execution loop, in a different crate — held back for a follow-up rather than smuggled into this route change. Removed `WorkflowStepResult.variables` from `api.ts`, the Variables section, and the console lines.

crates/librefang-api/dashboard/src/pages/TasksPage.tsx
DaBlitzStein: Held the frontend back rather than landing the backend fields here: `#7974` (open PR) already adds `TaskPostOptions { priority, timeout_secs }` — route, substrate, and a schema migration — for exactly this. Duplicating that in this PR would collide with its migration slot and split one feature across two PRs. Removed the priority selector, timeout field, their card badges, `PRIORITY_LEVELS`, `formatTimeout`, and the now-unused `priority`/`timeout_secs` fields on `CreateTaskPayload`/`TaskQueueItem` in `api.ts`, plus the locale keys that existed only for them (all 5 locales). Will re-add once `#7974` lands the backend.

crates/librefang-api/dashboard/src/pages/TasksPage.tsx
DaBlitzStein: Fixed. `filteredTasks` now derives from `validTasks`, not `allTasks` (`TasksPage.tsx:520-522`), so the column counters and the `status_total` tile see the same list `KanbanColumn` renders cards from. Added the one-line assertion this needed: `TasksPage.test.tsx` "does not render malformed task records without an id, and the total tile agrees" now also checks the total tile reads `0` for a fixture with one id-less task — verified red against `allTasks` (`Expected: 0, Received: 1`), green against `validTasks`.

crates/librefang-api/dashboard/src/locales/en.json
DaBlitzStein: Fixed. `tasks.field_assignee_placeholder` removed from all five locale files; `grep -rn field_assignee_placeholder src/` now returns nothing outside the locale files themselves (i.e. no hits at all, since it's gone from those too). `en-locale-coverage.test.ts` "does not carry dead keys in locale files" passes.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
DaBlitzStein: Fixed. `setParamValues(values)` → `setParamValues((prev) => ({ ...prev, ...values }))` in the auto-populate effect (`WorkflowsPage.tsx:602`). While fixing this we found the identical bug in the other caller that seeds `paramValues` from a stored run's input — `handleRerun` had the same total-replace bug at what's now `:691` — and merged that one too.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
DaBlitzStein: Fixed. The free-text textarea's `onChange` now also arms the ref: `onChange={e => { paramTouchedRef.current = true; setRunInput(e.target.value); }}` (`WorkflowsPage.tsx:1207`).

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
DaBlitzStein: Fixed. The effect now depends on `runDetailQuery.data?.step_results.length` instead of the array itself, and only snaps to the bottom when already within 40px of it (`WorkflowsPage.tsx:562-569`) — a poll tick that changes nothing no longer re-fires the effect at all, and even when a line is appended it won't yank the view out from under an operator who scrolled up.

changelog.d/fixed/7997-workflow-run-detail-poll.md
DaBlitzStein: Fixed. Deleted `changelog.d/fixed/7997-workflow-run-detail-poll.md` and folded the real, shipped behavior (3s polling while active — including while paused — stopped at a terminal state) into `changelog.d/added/7997-workflow-run-timeline.md`, which no longer claims the task board gains canvas helpers or richer run status (it gains neither; those helpers are the now-removed dead code, and the run status is on the Workflows page). `scripts/check-changelog-attribution.py` → OK.

crates/librefang-api/dashboard/src/lib/chat.ts
DaBlitzStein: Fixed by removing the intervening function. Deleting `extractWorkflowJson` (see the reply on the canvas-helpers comment) puts the `#6390` doc comment directly back above `applyForeignTerminalFrame` again (`chat.ts:203-210`). Also fixed the same displacement in `WorkflowsPage.tsx`: `TERMINAL_RUN_STATES` now sits above its own comment and `runStateKind` is back under the `#5257 round-2` comment that describes it (`:182-200`).

crates/librefang-api/dashboard/src/lib/canvas.ts
DaBlitzStein: Removed rather than wired up. `workflowStepsToCanvasState`, `WorkflowCreateStepInput`, `extractWorkflowJson`, and `CanvasNodeData.fresh` are gone along with their 135 lines of tests — `grep -rn 'extractWorkflowJson|workflowStepsToCanvasState' src/` now returns nothing. The "Save-as-Workflow" chat action these existed for doesn't exist in this PR; land them together with that caller in a follow-up rather than carrying dead code and unreachable tests here.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
DaBlitzStein: Reverted the silent revert. The `autoSwitchedRef` effect from #3412 is back, byte-for-byte matching `origin/main` (`WorkflowsPage.tsx:522-529`, with the `(#3412)` reference restored in the comment), and the test is back to asserting the original behavior ("auto-switches to templates tab when there are no workflows", `WorkflowsPage.test.tsx`) instead of the inverted one this PR had introduced. Verified this test fails against the state without the effect (`aria-selected="false"`, templates panel never mounts) and passes with it restored.

crates/librefang-api/dashboard/src/pages/WorkflowsPage.tsx
DaBlitzStein: Fixed. All the remaining hardcoded lines route through `t()` now: `console_run_started`, `console_step_line`/`console_step_line_failed`, `console_step_io`, `console_step_error`, and the `{{k}} = "v"` var-line helper, present in all 5 locales (`WorkflowsPage.tsx:1513-1534`). Also split the failure marker into its own full sentence (`console_step_line_failed`) instead of concatenating a raw `" FAILED"` onto the translated line, so word order isn't pinned to English for every language, and translated the one other hardcoded string in the same panel (`via {{agent}}` → `workflows.step_via`).

DaBlitzStein:
Every one of the 15 review threads on this PR now has an inline reply, and the branch is at `d3654e1d4`.

houko (commented):
Three of the four findings are fixed. The fourth is untouched, and the fix for the second one collides with #8177 — which matters, because #8177's version is the correct one.

**Fixed:**

- **`handleRerun` early-returning on empty input** — the guard is now `if (runInputStr)` around the pre-fill only, with a comment saying pre-fill is best-effort and must not stop the mutation. A parameterless workflow re-runs.
- **Auto-populate overwriting typed free text** — the textarea's `onChange` now sets `paramTouchedRef.current = true` (`:1207`) alongside `setRunInput`, so the effect at `:578` disarms the same way the parameter fields do.

**Not fixed — `priority` and `timeout_secs` on the Tasks page.** `crates/librefang-api/src/routes/task_queue.rs` is not in this PR's diff, so `task_queue_post_root` still reads only `title`, `description`, `assigned_to`, `created_by`. The selector and the field remain inert: the operator sets a priority, the task is created, and nothing carries it.

That work is #7974 ("validate assignee and enforce per-task priority and claim deadline"). So this is an ordering dependency, like the one on #8073 — the two PRs share no files, git will not conflict, and nothing stops this merging first into an API that silently discards both fields. Either note the dependency in the PR body and merge after #7974, or drop the two controls from this PR until the backend accepts them.

**Fixed, but by a mechanism that conflicts with #8177 and is wrong for loops.** This PR computes, in `get_workflow_run`:

```rust
let current_step_index = run.step_results.len();
```

with a comment saying "the step at this index is the one currently executing". That does not hold for `StepMode::Loop`, which pushes one `StepResult` per iteration. A one-step workflow looping five times reports `current_step_index: 5, total_steps: 1` — and this value feeds a progress bar. `StepMode::Conditional` breaks it the other way, skipping steps without pushing a result. The same unit mismatch is what I raised on #8177 for `steps_completed`, where it was resolved by documenting that the two are not a fraction.

#8177 solves the underlying problem instead: `current_step_index` becomes a real tracked field with `live_step_index()` gating the read on `WorkflowRunState::Running` (so a finished run cannot advertise a step it is no longer executing, and a future state variant fails closed), and `total_steps` is persisted through migration v56 so a run recovered after a restart does not report "step X of 0".

Both PRs touch `crates/librefang-api/src/routes/workflows/workflow.rs` — this one `get_workflow_run`, #8177 `list_workflow_runs` — so they are adjacent rather than textually conflicting, but they would ship two different definitions of `current_step_index` on two endpoints of the same resource. That is worse than either alone.

**What I would like:** merge #8177 first, then drop the local computation here and read `run.live_step_index()` and the persisted `run.total_steps` in `get_workflow_run` too. The dashboard progress bar then has one meaning on both endpoints, and it is the meaning that survives a restart.

Note #8177 currently has the v56 migration-numbering collision I described there; that has to be settled either way.

crates/librefang-api/dashboard/src/pages/TasksPage.tsx
DaBlitzStein: Re-checked at `6b11f6b28`: this does not reproduce, and I think the re-review looked at the backend rather than at where the fix landed.

The reply above took your second option — hold the frontend back — so the controls were removed rather than the backend added. Measured now:

- `grep -c "priority\|timeout_secs" crates/librefang-api/dashboard/src/pages/TasksPage.tsx` → **0**, and **0** at `d3654e1d4` too, the commit the re-review ran against.
- `git diff origin/main...HEAD -- crates/librefang-api/dashboard/` contains **no** `priority` or `timeout_secs` line in either direction.
- The submit handler sends `title`, `description` and an optional `assigned_to` — exactly the four fields `task_queue_post_root` reads.
- `CreateTaskPayload` has neither field.

Your premise about the backend is correct and unchanged: `task_queue.rs` is not in this diff, `task_post` still hardcodes `0`, and there is still no `timeout_secs` column. There is just no longer a control in front of any of it, so nothing is discarded silently.

One correction to my own earlier reply, since it over-claimed: `TaskQueueItem.priority?: number` **does** still exist (`api.ts:3325`). It is a read field on a type describing what `GET /api/tasks` returns, it is pre-existing on `main`, and this PR does not touch that line. Only the write-side field and the controls went.

So the ordering dependency on #7974 is gone — this no longer ships a control the API ignores. When #7974 lands the backend, the selector and the timeout field come back with it.

DaBlitzStein:
Pushed `6b11f6b28`. Both items from the `2026-09-10` round, one fixed and one refuted.

**`current_step_index` — dropped rather than redefined.** You are right that `step_results.len()` is wrong, and the two-step linear case in the integration test shows it without needing a loop at all: it reads `2` against `total_steps: 2` the moment the run finishes, i.e. "step 3 of 2".

I did not take the "merge #8177 first, then read `live_step_index()`" route, because it makes this PR uncompilable until that one lands. Instead this drops `current_step_index` from `get_workflow_run` entirely, so there is only ever one definition of the name in the codebase — #8177's. When #8177 merges, `get_workflow_run` can read the tracked index with nothing to reconcile, and no conflict to resolve in either direction. The integration test now asserts the field is **absent**, so a future re-derivation fails rather than silently shipping a second meaning.

The progress bar keeps working, honestly: it reports executions against defined steps and says "executed" rather than "executing step N", with the width clamped so a looping run cannot overflow it. `workflows.step_executing` is replaced by `workflows.steps_executed` in all five locales — the old key had no remaining reference and the dead-key guard would have failed on it.

**`priority` / `timeout_secs` — does not reproduce.** Answered in full in the original thread. Short version: the earlier round took your second option and removed the controls, so `TasksPage.tsx` contains zero `priority`/`timeout_secs` occurrences — at `d3654e1d4`, the commit the re-review ran against, as well as now — and the PR diff adds none. Your description of the backend is accurate and unchanged; there is simply no control in front of it any more, so the ordering dependency on #7974 is gone.

**Verification**
- `cargo check --workspace --all-targets --exclude librefang-desktop` → exit 0, no errors, no warnings.
- `cargo test -p librefang-api --no-fail-fast --test workflow_lifecycle_test` → 20/20.
- Dashboard: 1861/1861 vitest with `LANG=en_US.UTF-8`, `pnpm typecheck` clean.
- Locales valid JSON, no duplicate keys, deviating from a clean `origin/main` only by the pre-existing CLDR plural categories.

The fifteen inline findings from `2026-09-07` were each answered in their own thread on `2026-09-09`. Ready for another look.

DaBlitzStein:
Pushed `f3fee7038` on top of this branch (fast-forward, no force-push).

This comes from work that was sitting in a local review worktree. It turned out to be already committed and pushed as this PR's head, so nothing was recovered — but cross-checking it against #8177 before touching anything surfaced two collisions worth recording here.

## What `f3fee7038` changes

The test asserted `detail.get("current_step_index").is_none()` — it forbade the *name*. What is wrong is the *value*: `step_results.len()` counts executions, so a `StepMode::Loop` step makes it exceed `total_steps` and a skipped `StepMode::Conditional` makes it fall short.

#8177 adds `current_step_index` to this same payload as a tracked index gated on `WorkflowRunState::Running` — the value `total_steps` genuinely bounds, and the only one safe to render as a fraction. Against the old assertion that field is a failure, so whoever merged second would have deleted the assertion and taken the guard with it.

It is now `assert_ne!` against the execution count, which holds whether or not #8177 lands.

Verified discriminating against all three cases by hand-placing the value in `get_workflow_run`:

- field absent (current behaviour) — passes
- field equal to `step_results.len()` — **fails**, `left: Number(2)  right: Number(2)` with `total_steps: 2`, which is exactly the "step 3 of 2" this endpoint used to emit
- field present but a different value, as #8177's tracked index would be — passes

`cargo test -p librefang-api --test workflow_lifecycle_test -- --test-threads=1` → 20 passed, 0 failed. Serialized because this suite produces unrelated SQLite-contention failures in parallel on the machine it was run on.

No changelog fragment: narrowing a test assertion changes no behaviour.

## Collision 1 — this PR's test versus #8177, invisible to CI

#8177 adds to `get_workflow_run`:

```rust
+ "current_step_index": run.live_step_index(),
```

If **this PR merges first**, that addition breaks the assertion on `main` — and #8177's CI cannot see it, because #8177's base does not contain this branch. `f3fee7038` removes that hazard; the note is here so the ordering is not relied on.

## Collision 2 — `total_steps`, deliberately NOT fixed here

Both PRs emit `total_steps` into the **same `serde_json::json!` object** in `get_workflow_run`, from two different sources:

- this PR: `engine.get_workflow(run.workflow_id).await.map(|w| w.steps.len())` — what the workflow defines *now*, `None` if it was deleted.
- #8177: `run.total_steps` — a value persisted on the run record by its migration.

`serde_json::json!` with a repeated key is **not a compile error** — the last one wins silently. And the two values genuinely differ for a historical run whose workflow was edited after it finished.

Whoever merges second has to reconcile this. The persisted `run.total_steps` looks like the right one for a run-detail view, since it describes the run rather than the workflow's current shape — but changing it here would adopt the semantics of a PR that has not landed, so it is left alone on purpose.

DaBlitzStein:
A note on the Variables panel this PR removed, found while building an integration branch that carries this PR together with #8177.

The removal is **correct against this branch's base**: `StepResult` there has no `variables` field and the run-detail route does not emit it, exactly as the commit message says. I checked `origin/main` too — `crates/librefang-kernel/src/workflow.rs` has no `pub variables` there either.

It stops being correct once #8177 lands. That PR adds `pub variables: BTreeMap<String, String>` to `StepResult`, `routes/workflows/workflow.rs` serialises it, every construction site populates it through `snapshot_variables`, and `run_detail_and_list_expose_total_steps_and_step_variables` asserts a binding end to end. In the combined tree the route emits data that nothing renders.

The client-side type now lives in #8177 (`WorkflowStepResult.variables`). The render half belongs here, because this PR owns the expanded-step timeline — `expandedStepIdx` exists on this branch and on no other.

Nothing to do on this PR in isolation; flagging it so the panel is restored deliberately rather than noticed missing later. If it helps, the version I verified in the integration tree is the pre-removal render plus a test that fails without it: the test mocks a run whose step carries `variables: { greeting: "hello" }`, expands the step, and asserts `{{greeting}}` is on screen. Seen red against the removed section (1 failed / 24 passed) and green with it (25/25) — the absence of exactly that test is why the panel could be removed unnoticed.

I can open it as a follow-up commit here if you prefer, or leave it for whenever #8177 merges.

houko (approved):
Re-reviewed at the current head. The ordering dependency I raised is gone, and in the cleaner direction.

Rather than noting "depends on #7974" and queueing behind it, `d887a9c3` holds back the priority/timeout inputs, so nothing in this PR posts fields `task_queue_post_root` and `MemorySubstrate::task_post` cannot yet persist. The PR body records why and points at #7974 for the backend. This can merge on its own now.

The run-history and run-detail findings are addressed across `31a1b591` / `44024858`, `6b11f6b2` stops deriving the step index from the execution count with `f3fee703` asserting that directly, and `4307afa4` pins the tile count to the same list the cards render.

DaBlitzStein:
Rebasado sobre `origin/main` `57ad92341` (release v2026.9.14) y republicado con `--force-with-lease`.
La rama vuelve a estar `MERGEABLE`.

**La mitad de API de este PR ya está en `main`.**
La entró `776261018` — `feat(workflows): expose live step progress and persist total_steps so it survives a restart (#8177)` — con una implementación mejor que la de aquí: `main` sirve `"total_steps": run.total_steps` desde el campo persistido y `"current_step_index": run.live_step_index()` desde un índice seguido de verdad, mientras que esta rama calculaba el total con un `engine.get_workflow(...)` extra por petición.
Esta rama ya lo anticipaba —su último estado de `workflow.rs` deja un comentario que dice literalmente «No `current_step_index` here on purpose» y cede el campo a #8177— así que resolví los dos hunks de `crates/librefang-api/src/routes/workflows/workflow.rs` a favor de `main`.
Ese fichero **ya no aparece en el diff del PR**.

**Conjunto de ficheros**: 16 antes, 15 después.
El único que sale es `crates/librefang-api/src/routes/workflows/workflow.rs`, y sale absorbido, no perdido: `git diff origin/main..HEAD -- crates/librefang-api/src/routes/workflows/workflow.rs` es vacío.

**Otras resoluciones no triviales**

- `crates/librefang-api/dashboard/src/api.ts`, `WorkflowStepResult.variables`: los dos lados añadían el campo y la ronda 2 de esta rama lo quitaba. Me he quedado con el de `main`, que lleva doc-comment, **porque el handler de `main` sí lo serializa hoy** (`"variables": s.variables` en `workflow.rs`); quitarlo dejaría el tipo mintiendo sobre la respuesta.
- `WorkflowRunDetail.total_steps` / `current_step_index` en `api.ts` los aporta esta rama y no están en `main`, así que se quedan: son justo los dos campos que #8177 empezó a emitir y que el tipo del cliente todavía no declaraba.

**Lo que sigue siendo exclusivo de este PR**: todo el timeline del dashboard (`WorkflowsPage.tsx`, `TasksPage.tsx`, sus tests, las queries y los cinco locales) y el test de API `run_detail_reports_total_steps_and_no_derived_step_index`.
Ese test fue escrito precisamente para sobrevivir a la entrada de #8177 — no afirma que `current_step_index` esté ausente, afirma que **no vale `step_results.len()`**, con el razonamiento en el propio comentario: un `StepMode::Loop` empuja un resultado por iteración y un `Conditional` saltado no empuja ninguno, así que el recuento de ejecuciones no es un índice.
Contra la implementación de `main` sigue discriminando: `live_step_index()` devuelve `None` fuera de `Running`, y `assert_ne!(null, 2)` pasa, mientras que la derivación vieja daría 2 de 2.

**Verificado** (desde `crates/librefang-api/dashboard`, con `LANG=en_US.UTF-8`)

- `npx tsc --noEmit` — limpio.
- `npx eslint .` — limpio.
- `npx vitest run` — 209/210 ficheros, 1930/1931 tests. El único rojo es `PromptsExperimentsModal > stops selection when every variant has a traffic bucket`, timeout de 5 s, ajeno a este PR y reproducible igual en las otras ramas.
- Los cinco `locales/*.json` cargan con `json.load` y con `object_pairs_hook` (sin claves duplicadas).
- `rustfmt --edition 2021 --check` limpio en `workflow_lifecycle_test.rs`, y sin colisión de nombres de test en ese fichero.

**NO verificado por mi parte**: no he compilado ni ejecutado nada de Rust — ni `cargo check`, ni `clippy`, ni `cargo test -p librefang-api`.
El `CARGO_TARGET_DIR` de esta máquina está compartido con otra compilación en curso, así que la parte Rust queda enteramente para CI.

**Revisiones sin atender**: quedan **15 hilos de revisión sin resolver**, todos de @houko.
Cuatro de ellos **no** están marcados como *outdated* y siguen apuntando a líneas vivas: `WorkflowsPage.tsx:603`, `WorkflowsPage.tsx:675`, `WorkflowsPage.tsx:1454` y `TasksPage.tsx:370`.
No los he tocado —este pase era solo el rebase— y los dejo señalados en vez de darlos por cerrados.

PR #7998, opened 2026-08-29
feat(tui): ask for a workflow's declared parameters instead of one blob

DaBlitzStein:
## Summary

- When the operator opens the run-input screen for a workflow, the TUI now fetches the workflow's `input_schema` from `GET /api/workflows/{id}` and renders one editable field per declared parameter instead of a single free-form text box.
- Tab / arrow keys cycle between parameter fields and the free-form input box; required parameters (marked with `*`) are validated client-side before the run request is sent.
- The run payload is a JSON object keyed by parameter name, with type coercion for `number` and `boolean` fields. Workflows with no declared parameters keep the bare-string input unchanged.
- i18n keys added for en, uk, zh-CN, ko.

## Changed files

- `crates/librefang-cli/src/tui/screens/workflows.rs` — `WorkflowParamField` struct, `FetchWorkflowParams` action, `build_run_input()`, updated `handle_run_input()` + `draw_run_input()`, 5 unit tests
- `crates/librefang-cli/src/tui/event.rs` — `WorkflowParamsLoaded` event, `spawn_fetch_workflow_params()`
- `crates/librefang-cli/src/tui/mod.rs` — wiring for the new action and event
- `crates/librefang-cli/locales/{en,uk,zh-CN,ko}/main.ftl` — `tui-workflows-param-required` key, updated hint bar

## Verification

- `cargo clippy -p librefang-cli -- -D warnings` — clean
- Unit tests cover: typing into focused param, tab navigation, required-field validation, JSON payload with typed numbers, bare-string fallback

## Test plan

- [ ] Run a workflow that declares `[[input_schema]]` parameters from the TUI and verify the form renders one field per parameter
- [ ] Verify Tab/Down/Up cycle between fields and the free-form box
- [ ] Verify Enter is blocked when a required parameter is empty (status message names it)
- [ ] Verify a workflow with no `input_schema` still accepts the bare free-text input
- [ ] Verify the JSON payload includes typed numbers (not strings) for `param_type = "number"`


houko (commented):
The form logic reads well, but the parameter rows are drawn into a one-row area, so all but the first declared parameter are invisible and the first one is then overwritten.

`draw_run_input` (`crates/librefang-cli/src/tui/screens/workflows.rs:626`) splits the area as `[Length(2), Length(1), Length(1), Length(1), Length(1), Min(0), Length(1)]`.
`chunks[4]` is therefore exactly one row tall, and the new code renders the whole `param_lines` vector into it:

```rust
if !param_lines.is_empty() {
    f.render_widget(Paragraph::new(param_lines), chunks[4]);
}
```

A workflow declaring three parameters shows one row; the other two are clipped with nothing telling the operator they exist, and Tab moves an invisible cursor onto them.

The same `chunks[4]` is then reused for the free-text input:

```rust
if state.run_params.is_empty() || free_focused {
    f.render_widget(Paragraph::new(...), chunks[4]);
}
```

With declared parameters present *and* the free-text field focused, both branches fire and the second write lands on top of the first — so tabbing to the free-text field clobbers the one parameter row that was visible.

The fix is a layout change, not a render-order change: give the parameter block `Constraint::Length(run_params.len() as u16)` (or `Min`) computed from the state, and put the free-text input in its own chunk so the two never share a rect.

Two smaller things while you are in there:

- `spawn_fetch_workflow_params` only sends `WorkflowParamsLoaded` inside the `BackendRef::Daemon` arm.
  In in-process mode no event is ever sent, so the form silently stays in free-text mode with no indication that the schema was not consulted.
  Sending an empty vec in the other arm would at least make the state deterministic.
- `build_run_input` treats `"si" | "sí"` as truthy for a boolean parameter but not the other locales the TUI ships (`ko`, `uk`, `zh-CN`).
  Either accept only the canonical `true`/`1`/`yes` set or drive it from the locale — the current mix reads like an accident.

houko (changes requested):
Right idea, and `build_run_input`'s per-`param_type` coercion matches the type list `WorkflowInputParam` documents. The defaults for `param_type` and `required` also agree with the Rust side (`default_input_param_type()` → `"string"`, `default_required()` → `true`), which is the kind of thing that usually drifts. Four issues.

## Blocking

**1. `p.get("default")` reads a field that does not exist.**

```rust
value: p.get("default").map(|d| match d.as_str() { … }).unwrap_or_default(),
```

`WorkflowInputParam` (`crates/librefang-kernel/src/workflow.rs:220-239`) is `name`, `param_type`, `required`, `description` — there is no `default`, and `GET /api/workflows/{id}` serializes the struct as-is (`routes/workflows/mod.rs:215`). So every field always starts empty and this branch never runs.

Either add `default: Option<String>` to `WorkflowInputParam` (struct field + `#[serde(default)]` + `skip_serializing_if`, and the TOML authoring doc), or drop the read. Leaving it in states a capability the workflow format does not have.

**2. Every failure mode degrades silently to the old free-text blob.**

```rust
let params = client.get(…).send().ok()
    .and_then(|r| r.json::<serde_json::Value>().ok())
    .and_then(|body| body.get("input_schema").cloned())
    …
    .unwrap_or_default();
```

A 404, a 500, a connection error, a body whose `input_schema` is `null` and a workflow that genuinely declares no parameters all produce the same empty vec — and the form falls back to the single JSON blob, which is exactly the behaviour this PR exists to replace. The operator sees the old screen and has no reason to suspect a fetch failed. Distinguish "no declared parameters" from "could not load them", and say so in `status_msg`.

**3. In-process mode gets nothing at all.**

The whole body is `if let BackendRef::Daemon { .. } = backend`, with no `else`, so no `WorkflowParamsLoaded` is ever sent and `run_params` stays empty. Same shape as #8073's history fetch, #7976's token usage and #7985's config save — four spawn helpers in this crate now silently no-op for in-process users. Worth fixing here and worth a shared convention: every `spawn_*` either sends its event or sends an explicit "requires daemon mode" one.

**4. No test, for logic that is pure and easy to test.**

`build_run_input` takes no I/O. Cover: a `number` param coerces to a JSON number and a non-numeric value falls back to a string; `boolean` coerces; an unknown `param_type` stays a string; the required-field guard fires on a blank required field and not on a blank optional one. That last one is the behaviour the new `tui-workflows-param-required` string promises.

## Non-blocking

- `obj.insert("input".to_string(), …)` runs after the parameter loop, so a workflow declaring a parameter literally named `input` has it overwritten by the free-text box. A `if !obj.contains_key("input")` guard, or a different key for the blob, closes it.
- No changelog fragment (`changelog.d/added/7998-....md`). Asking for declared parameters individually instead of one JSON blob is a visible improvement worth a release-note line.
- All four locales are properly translated, including the reworded hint line. Noting it so it is clear the i18n side is not an omission.

DaBlitzStein:
Merge + review fixes pushed.

**Merge** — `bd629fdb8` merges `origin/main` (41 commits) into the branch. One conflict, `crates/librefang-cli/src/tui/mod.rs`: our `WorkflowParamsLoaded` arm landed where main added the channel-instance arms; resolved keeping both (verified with `cargo check -p librefang-cli` before committing).

**Blocking points**

1. **Nonexistent `default` read — dropped.** `WorkflowInputParam` has no `default` field, so the branch could never run; the read is gone rather than the field added, since the workflow format does not offer that capability today.
2. **Silent degradation — fixed.** `WorkflowParamsLoaded` now carries a `WorkflowParamsFetch` outcome (`Loaded` / `None` / `Failed`). A 200 without an `input_schema` array maps to `None` (the kernel omits the key when nothing is declared), anything that is not a 2xx or not JSON maps to `Failed`, and the status line says which one happened via new `tui-workflows-params-none` / `tui-workflows-params-failed` strings in all four locales.
3. **In-process mode — fixed here.** The in-process arm now sends `Failed` instead of sending nothing, so the form's state is deterministic in both backends. The shared `spawn_*` convention across #8073 / #7976 / #7985 spans three other PRs' helpers; I kept this PR scoped to its own and can follow up with a separate PR for the shared convention if you want it.
4. **Tests — added.** Five tests in `run_param_tests`: number coercion with non-numeric string fallback, boolean canonical-set coercion (`"si"` now coerces to `false`), unknown `param_type` stays a string, the required guard passes on a blank *optional* field, and a declared `input` parameter is not clobbered by the free-text box.

**Non-blocking points**

- `input`-key collision: the free-text value is now inserted only when no declared parameter already bound that key.
- Changelog fragment: `changelog.d/added/7998-tui-workflow-run-input-params.md`.
- The Aug 30 layout point: the parameter block now gets `Constraint::Length(run_params.len())` instead of the single shared row that clipped everything past the first parameter while Tab kept moving onto the invisible ones. One correction to that review, with the code as evidence: the free-text input rendered into `chunks[5]` (`Min(0)`), not `chunks[4]` as described, so the two never actually shared a rect — the clip was the real bug and is what changed.

**Verification** — `cargo check -p librefang-cli`, `cargo clippy -p librefang-cli` (zero warnings), `cargo fmt -p librefang-cli -- --check`, `cargo test -p librefang-cli`: 459 unit + 10 workflow run-input tests, 0 failures, locale-parity tests included.

DaBlitzStein:
Merged `origin/main` and addressed the reviewer's feedback.

Changes (fix commit 633de6e45):
- The run-input form now distinguishes "no parameters declared" from "schema unavailable" via a `WorkflowParamsFetch` outcome enum, and the status line reports which happened instead of silently degrading to the free-text blob.
- The in-process backend emits the schema-unavailable outcome too, so the form state is deterministic in both backends.
- Dropped the `default` read (no such field on `WorkflowInputParam`).
- One row per declared parameter in the run-input layout; a parameter named `input` is no longer clobbered by the free-text box, and boolean coercion accepts only `true`/`1`/`yes`.
- Added tests for number/boolean coercion, unknown-param-type fallback, blank-optional guard, and `input`-key precedence, plus the changelog fragment.

Merge conflict resolved by merging `origin/main` (merge commit bd629fdb8). Head is 633de6e45; GitHub reports MERGEABLE (blocked only on re-review).

houko (approved):
All four blocking points fixed, both non-blocking ones too, and you found two more bugs while in there.

Verified on the head:

**1 — the `default` read is gone** from the workflow path, so the form no longer implies a capability `WorkflowInputParam` does not have.

**2 — the three outcomes are distinct.** `WorkflowParamsFetch` is `Loaded(params)` / `None` / `Failed` (`event.rs:1319-1328`), so "declares no parameters" and "could not consult the schema" are different states with different status lines instead of collapsing into an empty vec and a silent fall back to the blob.

**3 — in-process sends an event** (`event.rs:1332`), so `run_params` is no longer left in a state that depends on which backend you are on.

**4 — five tests, all pure.** Number coercion with the non-numeric fallback, boolean coercion, unknown `param_type` staying a string, the blank-optional case that the `tui-workflows-param-required` string promises, and the `input`-precedence case. That covers what I asked plus the collision I raised as non-blocking.

## The two you found

**Boolean coercion was accepting a mixed per-locale word list**, now narrowed to the canonical `true`/`1`/`yes`. That is the more interesting of the two: a form that coerces `tak` or `是` to `true` depending on the operator's locale means the same workflow gets different JSON from different users, and it would have been very hard to attribute later.

**The parameter block had one shared row**, so everything past the first field was clipped. A layout bug that makes the feature unusable past one parameter is worth more than it sounds — the whole point here is workflows with several declared parameters.

## One non-blocking note

`BackendRef::InProcess(_) => WorkflowParamsFetch::Failed` reports in-process mode as "could not consult the schema". That is true but tells the operator to retry something that will never work — the schema is not reachable in that mode by design, not transiently. A third variant (or reusing the "requires daemon mode" phrasing the other screens use) would say so. Deterministic state in both backends was the substance of my ask and you have that; this is only about what the message implies.

`CI Gate` green, `mergeable` is `BLOCKED` on review rather than a conflict — so this one is clear of the rebase churn my merge sweep caused elsewhere.

Approving.

PR #7999, opened 2026-08-29
feat(tui): parametric workflow inputs in the run screen

DaBlitzStein:
## Summary

- The workflow run screen now fetches each workflow's declared `input_schema` and renders one field per parameter (name, type, required flag, description) instead of a single free-form text box.
- Tab/Up/Down cycle through declared parameter fields and the free-form input box; required parameters are validated client-side before submitting.
- The payload is assembled as a JSON object keyed by parameter name (with type coercion for `number` and `boolean`); workflows that declare no parameters keep the existing bare-string behavior.
- Adds `WorkflowParamField` struct, `FetchWorkflowParams` action variant, `WorkflowParamsLoaded` event, and the `spawn_fetch_workflow_params` background fetcher.
- i18n keys added for en and zh-CN.

## Changed files

- `crates/librefang-cli/src/tui/screens/workflows.rs` — param struct, state fields, input handling, `build_run_input()`, draw logic, tests
- `crates/librefang-cli/src/tui/event.rs` — `WorkflowParamsLoaded` event + `spawn_fetch_workflow_params`
- `crates/librefang-cli/src/tui/mod.rs` — dispatch for `FetchWorkflowParams` action and `WorkflowParamsLoaded` event
- `crates/librefang-cli/locales/en/main.ftl` — `tui-workflows-param-required` key
- `crates/librefang-cli/locales/zh-CN/main.ftl` — `tui-workflows-param-required` key

## Verification

- `cargo clippy -p librefang-cli -- -D warnings` — zero warnings
- 5 unit tests in `run_param_tests`: typing into focused param, tab navigation, required-field rejection, JSON build with params, bare-string fallback

## Test plan

- [ ] `cargo test -p librefang-cli -- run_param_tests` passes
- [ ] TUI workflow run screen shows declared params when a workflow has `input_schema`
- [ ] Required param validation blocks submit with a status message
- [ ] Workflows without `input_schema` still show the single free-form input


DaBlitzStein:
Duplicate of #7998 which covers more locale files.

PR #8000, opened 2026-08-29
feat(registry): survive the agent-types directory rename from either side

DaBlitzStein:
## Summary

- Extend `registry_paths.rs` (141 -> 566 lines) to handle both `agents/` and `agent-types/` directory names as the upstream registry renames its layout: canonical name preferred, legacy name accepted with a warning, neither triggers an error — all edge-triggered to avoid log flooding on per-dispatch resolution
- Add `installed_agent_types_dir()` to resolve `~/.librefang/agent-types/` as this installation's own agent-type manifest store, separate from templates (starter TOML scaffolding, #7758) and deployed agents (`workspaces/agents/`)
- Add `warn_on_agent_type_like_files_in_templates_dir()` boot-time diagnostic for installs that have agent-type-shaped files sitting in the old wrong location (read-only, never moves or deletes)
- Update `registry_sync.rs` fan-out to write agent-type manifests into `agent-types/<name>.toml` (flattened) instead of copying whole directories into `workspaces/agents/` — the old destination manufactured deployed agents the operator never asked to run
- Add `resolve_agent_templates_dir()` backward-compatible alias so existing callers (hands registry, kernel router) keep compiling without changes
- Historical `AGENT_TEMPLATES_DIR_NAME` constant kept as alias to the legacy name

## Verification

- `cargo check -p librefang-types --lib` and `--tests`: clean
- `cargo check -p librefang-runtime --lib` and `--tests`: clean
- `cargo clippy -p librefang-types --all-targets -- -D warnings`: zero warnings
- `cargo clippy -p librefang-runtime --all-targets -- -D warnings`: zero warnings
- `cargo test -p librefang-types -- registry_paths`: 13 passed
- `cargo test -p librefang-runtime --lib -- fanout_agent`: 6 passed (all new + updated tests)

## Test plan

- [x] Existing `registry_paths` tests continue to pass (13/13)
- [x] Legacy `agents/` only: resolves and installs into `agent-types/<name>.toml`
- [x] Canonical `agent-types/` only: resolves without legacy warning
- [x] Both present: canonical wins
- [x] Neither present: error logged, safe no-op
- [x] Fan-out never creates anything under `workspaces/agents/`
- [x] `warn_on_agent_type_like_files_in_templates_dir`: silent when no templates dir, silent for plain skeletons, warns for files with `[model]` table
- [x] `installed_agent_types_dir` returns `home/agent-types`
- [x] `resolve_agent_templates_dir` is an alias for `resolve_agent_types_dir`


houko (commented):
Merging.
`CI Gate` is green; the two red marks on this PR are both pre-existing and not caused by the branch.

- `cargo-deny (advisories)` — `error[yanked]: detected yanked crate (try 'cargo update -p chacha20')`.
  The same job has been failing on `main` since 2026-08-28 (last green run 2026-08-27), so it is a repo-wide `Cargo.lock` issue, not something this branch introduced.
  It needs its own PR bumping `chacha20` off the yanked version.
- `stale` — the housekeeping workflow, cancelled rather than failed.

The change itself reads well: `resolve_agent_types_dir` preferring the canonical name, falling back to legacy `agents/` with a warning, and erroring rather than silently skipping when neither exists is the right shape — a renamed registry must not be indistinguishable from an empty one.
The flatten-on-copy to `agent-types/<name>.toml` is covered from both sides (legacy-only and canonical-only fixtures) plus the negative that `workspaces/agents/` is never created.
`tracing-subscriber` correctly went in under `[dev-dependencies]`.

One thing worth a follow-up somewhere, not a blocker here: an installation that already ran the old fan-out has real directories under `workspaces/agents/<name>/` from before #7758.
Nothing in this PR removes or migrates them, so those installs keep a set of agents the operator never asked to run.
If that cleanup is already handled elsewhere, ignore this.

PR #8003, opened 2026-08-29
feat(telegram): send outbound text as native Rich Markdown

### LINKED ISSUE #8002: Telegram sidecar: replace the hand-rolled Markdown converter with native Rich Messages (Bot API 10.1+)
## Description

Both Telegram sidecars — Rust (`sdk/rust/librefang-sidecar-telegram/src/format/`) and
Python (`sdk/python/librefang/sidecar/adapters/telegram.py`) — format outbound agent
text with a hand-rolled Markdown → Telegram-HTML converter before calling `sendMessage`
with `parse_mode=HTML`.

That converter is four regexes (`format/markdown.rs`):

| Syntax | Supported |
|---|---|
| `**bold**`, `*italic*`, `` `code` ``, `[text](url)` | yes |
| headings, lists, blockquotes, fenced code | yes (block level) |
| `_italic_` | **no** — reaches the chat as literal `_italic_` |
| `__bold__` / underline | **no** — literal `__` |
| `~~strikethrough~~`, `\|\|spoiler\|\|`, `==marked==` | **no** |
| `<sub>`, `<sup>` | **no** |
| tables, task lists, footnotes, `$math$`, `---` | **no** |
| **nested formatting** | **structurally impossible** |

The last row is the important one. The bold rule is `\*\*([^*]+)\*\*` — the `[^*]+`
makes a nested star unrepresentable, so `**bold with *italic* inside**` can never
parse no matter how the rule is tuned. Same for every other combination. This is not
a bug to fix; it is the ceiling of a regex-substitution approach.

Practical consequences today: LLM output that uses perfectly ordinary GFM (`_italic_`,
strikethrough, nested emphasis, comparison tables) is delivered to the user as raw
punctuation.

## Proposal

Bot API 10.1 (2026-06-11) added **Rich Messages**: `sendRichMessage` and
`editMessageText(rich_message=...)`, where `InputRichMessage.markdown` takes
"Rich Markdown" — explicitly GFM-compatible and parsed server-side.

Beyond fixing the table/emphasis gaps, this buys:

* **32768-character limit** for rich message text, vs 4096 for `sendMessage`. Our
  `format/chunk.rs` (605 lines of UTF-16 + HTML-entity-aware splitting) exists purely
  to slice replies at 4096; long answers stop being cut mid-sentence.
* **Up to 16 levels of nested formatting** — currently unreachable.
* Full inline set: `**`/`__`, `*`/`_`, `~~`, `` ` ``, `==marked==`, `||spoiler||`,
  `<u>`, `<sub>`, `<sup>`, footnotes, LaTeX.
* Block set: tables, task lists, `<details>`, dividers.
* Parsing moves to Telegram's server, which puts `markdown.rs` (474 lines),
  `sanitize.rs` (316) and `chunk.rs` (605) on a path to deletion, along with their
  entire bug class.

Scope deliberately **excludes** building out the rest of the Rich Message surface
(`blocks[]`, collages, slideshows, maps, `InputRichBlock*`). The goal is to let
Telegram parse the text agents already write — not to build a rich-block constructor.

Fallback to the existing `sendMessage` + HTML pipeline on any failure, so self-hosted
Bot API servers below 10.1 keep working unchanged.

## Open questions (input wanted before a PR)

**1. Sanitization — does passing raw agent text to `markdown` open an injection hole?**

The spec states Rich Markdown "can contain arbitrary HTML", including:

```html
<tg-button type="callback_data" data="arbitrary">Click</tg-button>
<tg-button type="login_url" url="https://...">Sign in</tg-button>
<a href="tg://user?id=777000">mention</a>
```

Today `sanitize_telegram_html` enforces a 12-tag allowlist and a 4-scheme href
allowlist, so none of this can reach Telegram. If we forward raw agent output into
`markdown`, any text an agent *quotes* — a fetched web page, an email body, file
contents — could render itself inline buttons with attacker-chosen `callback_data`,
and a tap comes back to the adapter as a `ButtonCallback` event. That is a
prompt-injection surface that does not exist today.

Options:
  a. New rich-markdown sanitizer: strip/escape `<tg-button*>`, `<tg-map>`, `<details>`,
     `tg://` hrefs and other active constructs; keep passive formatting only.
  b. Accept the risk and document it (agent output is already trusted to some degree).
  c. Opt-in env flag for the unsanitized path.

Preference here is (a) — it preserves the current security posture, and interactive
buttons should stay an explicit `ChannelContent::Interactive` feature rather than a
side effect of text formatting. Confirmation from maintainers would be good, since it
is the main thing gating the PR.

**2. `` becomes real media.**

In Rich Markdown, image syntax creates an actual media block fetched from the URL
(HTTP/HTTPS only). Today it is inert text. So an agent that writes a Markdown image
link silently changes from "shows a link" to "uploads and attaches media", including
to whatever URL appeared in quoted content. Escape it (preserves current behaviour) or
let it through?

## Related, not in scope

* `sendRichMessageDraft` for streaming — it is an ephemeral 30-second preview that
  must be finalized with a real `sendRichMessage`, i.e. a different message lifecycle
  than the current throttled `editMessageText`. Worth a separate issue.
  `editMessageText(rich_message=...)` works with the existing lifecycle and is what
  this change would use.
* Behaviour of pre-10.1 Telegram *clients* when receiving a rich message is not
  documented in the API reference; if anyone knows, that would be useful input.

## Steps to Reproduce (current behaviour)

1. Configure the `telegram` sidecar channel (either adapter).
2. Have an agent reply containing: a GFM table, `_italic_`, `~~strikethrough~~`, and
   `**bold with *nested italic* inside**`.
3. All four render as literal punctuation instead of formatting.

## LibreFang Version

main — `sdk/rust/librefang-sidecar-telegram`,
`sdk/python/librefang/sidecar/adapters/telegram.py`

## Operating System

Other

## References

Bot API changelog 10.1 (2026-06-11), 10.2 (2026-07-14), 10.3 (2026-08-24):
https://core.telegram.org/bots/api-changelog

nevgenov:
Closes #8002.

## What

Both Telegram sidecars send outbound text as Telegram's native Rich Markdown (`sendRichMessage` / `editMessageText(rich_message=…)`, Bot API 10.1+), falling back to the existing `sendMessage` + HTML pipeline when Telegram refuses.

## Why

The existing converter is four regexes. Before → after, for text an LLM actually produces:

| Input | Before | After |
|---|---|---|
| `\| a \| b \|` table | literal pipes | rendered table |
| `_italic_` | literal `_italic_` | *italic* |
| `~~strike~~` | literal `~~strike~~` | ~~strike~~ |
| `**bold *italic* inside**` | mangled | correctly nested |

The nesting case is structural: the bold rule is `\*\*([^*]+)\*\*`, and `[^*]+` makes a nested star unrepresentable, so no amount of tuning fixes it.

Secondary win: the rich limit is 32768 characters vs 4096 UTF-16 code units, so ordinary replies stop being chopped mid-sentence.

## Security: what this guarantees, and what it costs

Rich Markdown "can contain arbitrary HTML", including `<tg-button type="callback_data" data="…">`. Our text is model output, and model output routinely *quotes* untrusted content — a fetched page, an email body, a file the agent read. A button rendered from quoted text returns to the adapter as a genuine `ButtonCallback` with an attacker-chosen payload.

The first version of this PR sanitised by exemption: code spans, fenced blocks and well-formed tags were copied through verbatim so a user's code sample rendered as written. **Every exemption has to decide where a CommonMark construct ends, which means reimplementing part of the parser.** Five rounds of adversarial self-review found five separate inputs where that reimplementation disagreed with a reference implementation and let a live `<tg-button>` through:

* a code span pairing across a blank line, then across a lone `\r`, then across a backslash-escaped backtick
* `<b <tg-button …>` — a tag scan running past the tag it was scanning
* `    ```` ``` ```` ` — a four-space-indented fence marker, which CommonMark reads as indented code

**Two of those five were introduced by the fix for the previous one.** So the exemptions are gone.

**Guaranteed:** every `<` is backslash-escaped, unconditionally. No raw HTML reaches Telegram, by construction, whatever the surrounding text looks like — the property needs no agreement with any parser. A literal backslash is doubled first, since otherwise an input already carrying one before a `<` would leave as `\\<`: an escaped backslash and then a bare `<`. The spec warns about exactly that ("`\` character usually must be escaped with a preceding `\` character").

Backslash rather than `&lt;` because the Bot API documents that mechanism for the `markdown` field by example — `\#hashtag` appears in its own Rich Markdown sample — and says nothing about entity decoding there; the entity note appears only under the two HTML styles.

**Cost:** the escapes land inside code spans and fenced blocks too, where Markdown does not process them, so `Vec<String>` in a fence reads `Vec\<String>` and a literal backslash is doubled. Every Rich HTML construct is lost, since they all start with `<`: `<u>`, `<ins>`, `<sub>`, `<sup>`, `<br>`, `<details>`/`<summary>`, `<aside>`/`<cite>`, `<a name>`, `<tg-map>`, `<tg-collage>`, `<tg-slideshow>`, `<tg-emoji>`, `<tg-time>`. Emphasis, strikethrough, spoilers, highlight, tables, lists, headings, code, math and footnotes are plain Markdown and unaffected.

**Link destinations are not filtered, deliberately.** Earlier revisions checked them against the legacy sanitiser's scheme allowlist, which meant locating Markdown links: a label scanner with a length cap, a per-message budget, a forward cursor, reference-definition and title parsing. Five rounds of adversarial review found a defect in that machinery every single time — four of them introduced by the fix for the previous one, and the last turned an input containing no link at all into a live `javascript:` one. Nine of the module's ten functions existed to serve that check; every defect lived in them, and none ever touched the `<` rule.

The distinction that decided it: `sanitize_telegram_html` filters schemes correctly because it *constructs* the `<a href>` and therefore knows where the href is. Here we would be guessing at someone else's parse of someone else's text. The established libraries for this problem — `telegramify-markdown` on pulldown-cmark, GramIO's entity builders — parse with a real parser rather than scan; we have no parser on this path, so we no longer pretend to have one. What guards a hostile link instead: Telegram renders only schemes it supports, the client shows an "Open this link?" confirmation carrying the full URL, and a chat client has no script context. The legacy `sendMessage` path we fall back to on a 4xx still filters, since it builds the HTML itself. `InputRichMessage.blocks` (#8015) removes the question entirely.
#8015 tracks moving to `InputRichMessage.blocks`, where Telegram runs no parser on our content at all and a preformatted block's text is a plain string. That removes both the escaping cost and this whole class of bug; a bug in *our* parser would become a formatting glitch instead of an injection.

## Fallback behaviour

Only a **4xx** from Telegram falls back to the legacy pipeline. A 5xx or a transport failure does not: Telegram may have created the message already, and re-sending would deliver the same answer twice. 429 is excluded — it means "try later", and `call_json` has already spent its retry. `code == 0` is included, since that is the sentinel for an HTTP 200 carrying `ok: false`, which *is* a verdict; reading it otherwise would silently disable the fallback on exactly the self-hosted pre-10.1 servers it exists for.

There is no latch: the decision is per-message and stateless, so a transient outage costs nothing beyond the messages it hits.

## Not in scope

`sendRichMessageDraft` for streaming — an ephemeral 30-second preview needing a separate finalizing send, i.e. a different message lifecycle. Streaming here uses `editMessageText(rich_message=…)`, which fits the existing one.

Also out of scope: the rest of the Rich Message surface (`blocks[]`, collages, slideshows, maps), and `Interactive` / media captions, which stay on the legacy converter.

## Testing

Rust 123 tests, Python 2199. `cargo clippy --all-targets -- -D warnings` and `cargo fmt --all --check` clean.

Every fix is mutation-checked, and where a mutation survived I added the missing test rather than claim coverage I did not have. That process caught, among others, two tests that asserted less than their names promised — one pinned a bypass as correct behaviour, and one passed on an input that shipped a live button.

## Open question

I have not been able to verify on a live bot that Telegram honours backslash escapes in the `markdown` field. The spec documents the mechanism by example, and the failure mode would be visible rather than silent (literal `\<` in messages), but it is the one thing no amount of review settles. If anyone has a 10.1+ bot handy, `sendRichMessage` with `{"markdown": "a \\< b"}` answers it.


nevgenov:
Force-pushed a revision. Self-review found several defects in the first version, including two in the security pass this PR exists to add, so I want to correct the claims I made rather than let the diff quietly replace them.

**The sanitiser did not actually hold.** The code-span scanner ended its search only on a literal `\n\n`, so a code span could pair *across* a block boundary and copy everything between the two backticks verbatim. Three inputs got a live `<tg-button>` through: CRLF blank lines (`\r\n\r\n`), whitespace-only blank lines (`\n \n`), and any paragraph-interrupting construct (`>`, `#`, `-`, `1.`, `|`). CRLF is not exotic here — it is what quoted email bodies are made of, which is precisely the untrusted-content path the pass is meant to cover. Fixed by bounding the scan to a single block; `code_span_does_not_pair_across_a_block_boundary` covers all eight shapes.

**I claimed "parity with today" for link schemes, and it was false.** I wrote that comment without checking. `sanitize_telegram_html` drops `javascript:` / `data:` links, and `docs/architecture/rust-telegram-sidecar.md` documents that as a guarantee — but on the rich path those reached Telegram as live tags, because I left scheme filtering to Telegram, which does not do it. Both `<a href>` and `[label](destination)` are now checked against the same allowlist.

**I justified copying fenced blocks verbatim by saying escaping "would surface a literal `&lt;` inside a user's code sample" — and then wrote a fence rule that does exactly that.** `fence_at` accepted a closing fence with an info string, so ` ```js ` *inside* a block was mistaken for the close and the following line got escaped. CommonMark allows an info string only on the opening fence.

**A regression I introduced.** Routing `finalize_as_new_message` through `dispatcher::send_text` made streaming's finalize chunked, and I treated `Err` as "nothing was delivered" — directly contradicting `send_text_legacy_html`'s own doc-comment, which warns callers to treat an error as *possible partial delivery*. On a mid-sequence failure the user saw the first 4096 characters twice, and the `…` placeholder survived as the duplicate. It now asks how many messages actually landed.

**"Any failure falls back to the legacy pipeline" was too broad.** A transport timeout or a 5xx is not a verdict — Telegram may have created the message — so falling back re-sent the whole answer. Only a 4xx does now, on both the send and the streaming-edit path.

Also fixed: `docs/architecture/rust-telegram-sidecar.md`, the sidecar README and the `chunk.rs` / `markdown.rs` module docs still described the HTML pipeline as *the* outgoing path; a dead `_edit_text` in the Python adapter that I had given new logic and a docstring describing a path that never executes; a `_ALLOWED_HREF_SCHEMES` name collision where my new constant silently shadowed the legacy sanitiser's (caught only because its tests failed); and one of my own tests that hid its single meaningful assertion behind `if edits:`, so it passed while checking nothing.

Two process corrections against `CLAUDE.md`: the changelog entry is now a `changelog.d/changed/` fragment rather than a direct `CHANGELOG.md` edit, and the AI-attribution trailers are gone from the commit and this PR body.

Every fix is mutation-checked — reverting it fails exactly the test that covers it, and nothing else.

houko (commented):
The only red check here is `Validate PR title`, and it is a formatting rule rather than anything about the change.

`.github/workflows/pr-title.yml` runs `amannn/action-semantic-pull-request` with the conventional-commit type list (`feat`, `fix`, `docs`, `style`, `refactor`, `perf`, `test`, `build`, `ci`, `chore`, `revert`, `release`).
"Send Telegram text as native Rich Markdown instead of the hand-rolled HTML converter" carries no `type(scope):` prefix, so it fails.

Retitling is the whole fix — the check re-runs on `edited`, and `CI Gate` is green already.
Something like `feat(telegram): send outbound text as native Rich Markdown` keeps the subject inside git's ~72-character display convention while naming the same change.
Leaving that to you rather than editing the title of someone else's PR.

On the change itself: the security section is the right instinct, and the reasoning for it is sound.
Rich Markdown admitting arbitrary HTML including `<tg-button type="callback_data">`, over text that is model output routinely quoting untrusted content, is a genuine injection path back into the adapter's own callback handling — an allowlist rather than a denylist is the correct shape, and bounding code spans to a single block closes the obvious bypass.

One question before this lands, since it is the part a reviewer cannot verify from the diff: what does the fallback actually key on?
The body says it falls back to `sendMessage` + the existing HTML pipeline "when Telegram refuses", and whether that is a specific API error code or any failure decides how a transient Bot API outage behaves — retry-and-degrade versus permanently downgrading every message for the life of the process.
If the fallback latches, a single 5xx costs the rich rendering until restart.

nevgenov:
Thanks — retitled to `feat(telegram): send outbound text as native Rich Markdown`.

## The fallback keys on a definitive 4xx, and it does not latch

Both halves of your question have a specific answer in the code, and the second one was a real bug I fixed after opening this PR — so the body you read was describing older behaviour. `dispatcher.rs`:

```rust
pub(crate) fn is_api_rejection(e: &Error) -> bool {
    matches!(e, Error::Api { code, .. } if *code != 429 && (*code == 0 || (400..500).contains(code)))
}
```

**Only a 4xx falls back.** A 5xx, a transport failure (timeout, connection reset) and a decode failure do not — they leave the outcome unknown, and Telegram may have created the message already, so re-sending through `sendMessage` would deliver the same answer to the user twice. That case is `(0, Some(e))` from `send_text_counting`: the error surfaces and nothing is re-sent.

Two carve-outs inside the 4xx range, both deliberate:

* **429 is excluded.** It is the one 4xx meaning "try later" rather than "not like this". `call_json` has already spent its single `retry_after` retry by the time we see it, so treating it as a refusal would push a second request into a chat Telegram just asked us to back off from.
* **`code == 0` is included.** That is the sentinel `call_json` produces for an HTTP 200 carrying `ok: false` (`parsed.error_code.unwrap_or(0)`), which *is* a verdict — Telegram answered, in JSON, that it did not create the message. Reading it as "not a 4xx" would silently disable the fallback for any deployment that reports failures with a 200, which is exactly the self-hosted pre-10.1 server the fallback exists for.

**No latching.** The decision is per-message and stateless: `send_text_counting` calls `prepare_rich_markdown` and `send_rich_message` fresh every time, with no cached "this server can't do rich" flag anywhere. So a transient outage costs nothing beyond the messages it actually hits, and the next message tries rich again — the scenario you describe, where one 5xx downgrades rendering until restart, cannot happen.

The honest flip side of not latching: a genuinely pre-10.1 server pays one failed round-trip per outgoing text, forever. That is a real cost and I left it deliberately — a latch would need an invalidation story (server upgraded mid-process) and would trade a permanent-downgrade bug for a doubled-request cost. Happy to add one if you would rather have it, but it seemed like the wrong default for a correctness-vs-chattiness trade.

## On the sanitiser

Your read is right about the shape, and I should flag that "bounding code spans to a single block closes the obvious bypass" turned out to be the *first* of several. Three rounds of adversarial self-review since then found the block-starter list was assembled from recall rather than from CommonMark (`***`, `---`, `___`, setext underlines and HTML blocks all interrupt a paragraph and were missing), and then that the bounded scan had two other routes to the same verbatim copy: a lone `\r` line ending, which meant the boundary check never ran at all on CR-terminated text, and a backslash-escaped backtick, which is a literal backtick rather than a span opener.

Also since fixed: attribute names are case-insensitive and were being matched with a substring search, so `<a HREF="javascript:…">` and `<a data-href="https://ok" href="javascript:…">` both got through; entity-encoded and whitespace-obscured schemes (`&#32;javascript:`, `&Tab;javascript:`, `java&#10;script:`); a tag spanning a line break; and link reference definitions, which were not inspected at all, so `[x][ref]` plus `[ref]: javascript:…` rendered a live link on the rich path while the legacy pipeline leaves it inert.

Everything is pinned against a reference CommonMark implementation now rather than against my reading of the spec, and each fix is mutation-checked — an earlier round found five mutations that survived the entire suite, two of which reverted real defences.

nevgenov:
Force-pushed a substantial change of approach. @houko — this replaces the sanitiser you reviewed, so the shape is worth re-reading rather than the diff.

## What changed and why

I kept running adversarial self-review on this PR. Across five rounds it found five separate inputs where the sanitiser passed a live `<tg-button type="callback_data">` through to the chat — where a tap comes back to the adapter as a genuine `ButtonCallback`:

* a code span pairing across a blank line, then across a lone `\r`, then across a backslash-escaped backtick
* `<b <tg-button …>` — the tag scan running past the tag
* `    ```` ``` ```` ` — a four-space-indented fence marker, which CommonMark reads as indented code, not a fence

**Two of those five were introduced by the fix for the previous one.** The anatomy was identical every time: an exemption (code spans, fenced blocks and well-formed tags were copied through verbatim so a user's code sample rendered as written), and every exemption has to decide where a CommonMark construct *ends*. That decision was a hand-written approximation of the spec, and each round found another place it disagreed with a real parser.

That is not a patch series that was nearly done. The rule it was ignoring is well known outside this repo: you cannot decide what a parser will do without being that parser, and anything less is an illusion of security.

## The new shape

Every `<` is escaped, unconditionally. No exemptions, no attempt to locate code spans, fences or tags. About four hundred lines of block-structure guesswork are gone with them — `fence_at`, `starts_new_block`, `thematic_break_at`, `setext_underline_at`, `is_bullet`, `tag_end`, `tag_name_at`, `anchor_href_allowed`, the tag allowlist. Net **−1353 lines**.

No raw HTML reaches Telegram, by construction, whatever the surrounding text looks like. That property needs no agreement with any parser, which is the whole point.

**The cost, stated plainly:** a `<` inside a fenced code sample now renders as a literal `&lt;`, because Markdown does not decode entities inside code — `Vec<String>` reads `Vec&lt;String>`. The Rich HTML tags with no Markdown equivalent (`<u>`, `<sub>`, `<sup>`) are lost too. Emphasis, strikethrough, spoilers, tables, lists and headings are plain Markdown and unaffected, so the feature this PR exists for is intact.

I have also stopped overclaiming about link schemes: `[label](destination)` is still checked, but the code and docs now say it is best-effort defence in depth rather than a guarantee, because locating a Markdown link exactly has exactly the same problem.

Bounding the label scan at CommonMark's own 999-character cap also ended a quadratic blowup the previous revision left behind — a megabyte of unmatched `[` went from stalling the sidecar to 0.3 s.

## The proper fix is filed separately

#8015 tracks moving to `InputRichMessage.blocks`, where Telegram runs no parser on our content at all and a preformatted block's text is a plain string — which removes both the escaping cost and the class of bug entirely. The decisive property there is that a bug in *our* parser becomes a formatting glitch instead of an injection.

I did not fold that into this PR: it is a different piece of work, and this one is now safe and small.

## Your earlier question

The answers still hold, and round 5 checked each one against the code: fallback keys on a 4xx only, 429 excluded, `code == 0` (HTTP 200 with `ok: false`) included, 5xx and transport failures do not fall back, and there is no latch — the decision is per-message and stateless.

Rust 121 tests, Python 2199, clippy and fmt clean. Every fix is mutation-checked; where a mutation survived, I added the missing test rather than claiming coverage I did not have.

nevgenov:
Final state after the last review round, for whoever picks this up.

**What changed since the previous round.** The link-destination scheme check is gone. It is worth being explicit about why, because the PR body used to promise it: filtering a destination means locating it, and locating it means deciding where a CommonMark construct ends — which is reimplementing the parser. Nine of the module's ten functions existed to serve that check, every defect found across five review rounds lived in them, and the last one turned an input containing no link at all into a live `javascript:` one. `sanitize_telegram_html` can filter schemes because it *constructs* the anchor and knows where the href is; on this path we would be guessing at someone else's parse. What remains is two character-local rules and a guarantee that holds by construction: no bare `<` reaches Telegram, in any context.

The sanitiser also no longer doubles backslashes. Doubling reached the right parity but rewrote the author's text — `\*not italic\*` came back as emphasis with stray backslashes, and `[a\](https://x)`, which is not a link, became one. It now tracks the parity instead and is a no-op on any input containing no bare `<` and no `![`.

**Mutation results.** The last round mutated four behaviours that no test noticed. All four are now pinned; the numbers are Rust unless stated:

| mutation | before | after |
|---|---|---|
| drop `backslashes = 0` in the `<` arm | 121 passed | fails |
| drop the parity guard on `!` before `[` | 121 passed | fails |
| include 429 in `is_api_rejection` | 121 passed | fails |
| drop `code == 0` from `is_api_rejection` | 121 passed | fails |

The first is a real hole rather than a formality: `\<<tg-button …>` came out with the second `<` bare, because the first consumed the backslash run and the counter was never reset. Both ports agreed on the wrong answer. The context list had twelve entries and none with two `<` in a row.

The two `is_api_rejection` carve-outs are the ones @houko asked about. They had thirty lines of doc comment and no test — the only `ok:false` fixture carried an explicit 404, so neither was exercised. Python covered both, so this was also a port-parity gap.

**Suite:** 123 Rust, 2199 Python, clippy `-D warnings` and fmt clean.

**One thing is still unverified:** whether a live Bot API server honours backslash escapes inside `rich_message.markdown`. The spec documents the mechanism by example, and the whole guarantee rests on it. The failure mode is loud rather than silent — the first quoted `<` would render as a tag — but it has not been exercised against a real bot. Happy to run that before merge if someone has a test bot handy.

`InputRichMessage.blocks` (#8015) removes the need for this pass entirely and is tracked separately.

houko (changes requested):
The content here is good — `rich_sanitize`'s two character-local rules, and the doc explaining why an earlier revision's link-destination scanner was removed after five rounds of review each introducing a fresh defect, is the right call and the right way to record it.

The blocker is structural, not technical.

## Blocking

**This PR is a strict subset of #8049, and both branch from `main`.**

`feat/8015-rich-blocks` (#8049) contains every file this PR touches — `adapter.rs`, `api/client.rs`, `dispatcher.rs`, `format/chunk.rs`, `format/markdown.rs`, `format/mod.rs`, `format/rich_sanitize.rs`, the Python adapter and its tests, the architecture doc, the README — plus `format/rich_blocks.rs`, a `Cargo.toml` dependency and a `ci.yml` note. It also re-adds this PR's own changelog fragment, `changelog.d/changed/8003-telegram-rich-markdown.md`, byte for byte.

Because #8049 is based on `main` rather than on this branch, the two cannot both merge: whichever lands second conflicts across nine files. Please pick one of:

- Merge this first and rebase #8049 onto `main` afterwards, so it reduces to the rich-blocks increment. This is the smaller review and the safer order — the sanitizer lands on its own and #8049's `pulldown-cmark` dependency and MSRV question get reviewed separately.
- Or close this as superseded by #8049 and note that there, so the sanitizer work is not reviewed twice.

I would take the first. The sanitizer is self-contained, well tested, and shipping it separately means the MSRV problem #8049 documents (the crate declares `rust-version = "1.80"` while its tree resolves to 1.86) does not gate the fix to the text path.

## While it is open

- One question that outlives the split: `rich_sanitize`'s "Link destinations are not filtered" section rests on "We have no parser on this path, so we do not pretend to have one." That is true here. It stops being true in #8049, which adds `pulldown-cmark` to the same crate and *does* filter schemes on the blocks path. Whichever PR lands second should make the two paths' different guarantees explicit in that doc, so nobody later reads the paragraph as an oversight.
- The reasoning about backslash parity — leaving an author's existing escapes alone rather than doubling every backslash, with the `\*not italic\*` and `[a\](https://x)` regressions named — is worth a test if there is not one already, since it is the exact thing the earlier revision got wrong.

houko (approved):
Revising my earlier review on this PR — it should not have been "request changes".

That review's own conclusion was "merge this first and rebase #8049 onto `main` afterwards", and then I blocked the thing I had just recommended merging. The stacking overlap with #8049 is a fact about #8049's base, not a defect in this branch: this one is self-contained, CI is green, there are no conflicts, and the sanitizer is the part that should land on its own rather than waiting behind #8049's `pulldown-cmark` dependency and its MSRV question.

Approving. The two notes from before stand as non-blocking follow-ups, not conditions:

- Once #8049 lands, `rich_sanitize`'s "We have no parser on this path, so we do not pretend to have one" needs a sentence distinguishing the text path (no parser, destinations unfiltered) from the blocks path (pulldown-cmark, schemes filtered). True as written today; misleading the moment both exist.
- The backslash-parity reasoning — leaving an author's existing escapes alone rather than doubling every backslash, with `\*not italic\*` and `[a\](https://x)` named as the regressions that produced the rule — is worth a test if there is not one, since it is exactly what the earlier revision got wrong.

#8049 should be rebased onto this branch so it reviews as the rich-blocks increment.

PR #8004, opened 2026-08-29
test(workflows): verify concurrent registration admits exactly one winner

DaBlitzStein:
## Summary

- Adds `concurrent_register_unique_name_persists_exactly_one`, the persistence-enabled variant of the unique-name race: of 32 concurrent callers registering `deploy`, exactly one `.workflow.json` survives on disk, no `.workflow.json.tmp` stragglers remain, and a reload via `load_from_dir_sync` sees only the winner with its id preserved.
- The losers collect their `NameTaken` payload and assert every rejected caller's `existing_id` points at the single winner — the same linearization property the in-memory sibling asserts, checked here on the path where each rejection races a disk write.
- The workflow-definition persistence machinery this test exercises (silently-dropped I/O errors, `persist_definition` returning `Result<()>`, `register_unique_name` propagating it) is already on `main` (`workflow.rs:2235`, `:7859`) — this PR only adds the test.

Refs #6943.

## Verification

- `cargo nextest run -p librefang-kernel -E 'test(concurrent_register)'` — both the in-memory and the persisted variant run and pass.


houko (commented):
`Quality` fails on two clippy lints in code this PR added, and `CI Gate` is red only because of that job.

`crates/librefang-kernel/src/workflow.rs:8112`:

```
error: this `map_or` can be simplified
    .map_or(false, |s| s.ends_with(".workflow.json"))
help: use is_some_and instead
```

`crates/librefang-kernel/src/workflow.rs:8124`:

```
    .filter(|e| e.path().to_str().map_or(false, |s| s.ends_with(".tmp")))
help: use is_some_and instead
```

Both are `clippy::unnecessary_map_or`, which the workspace promotes to an error through `-D warnings`, so `cargo check` passing locally is not enough here — `cargo clippy --workspace --all-targets -- -D warnings` is the gate.
Swapping both to `is_some_and` is the whole fix and `cargo clippy --fix` will do it.

Worth noting these two lints only fire under Rust 1.94 (the log shows `rust-clippy/rust-1.94.0`), so an older local toolchain will report the branch clean.

houko (dismissed):
The test itself is good — it covers a property nothing else does, and both of its filesystem assertions actually bite (`persist_definition` writes `{id}.workflow.json.tmp` and renames to `{id}.workflow.json`, `workflow.rs:2240-2241`, so neither filter is vacuous). My objection is entirely to the metadata, which currently describes work that is already on `main`.

## Blocking

**1. The PR body describes changes that are not in this diff — they already landed.**

> `persist_definition` returns `Result<()>` and `register_unique_name` propagates the error

Already true on `main`: `async fn persist_definition(&self, workflow: &Workflow) -> Result<(), WorkflowPersistError>` (`workflow.rs:2235`), with `register_unique_name_rolls_back_when_the_workflows_dir_cannot_be_created` (`workflow.rs:7976`) and `register_unique_name_rolls_back_when_the_rename_fails` (`workflow.rs:8027`) covering it.

> Strengthens `concurrent_register_unique_name_admits_exactly_one` to verify every rejected caller's `existing_id`

Already true on `main` too — that test collects `loser_incumbents.push(taken.existing_id)` at `workflow.rs:7859` and asserts on it.

The diff is `+74/-0` in one file: a single new test. Please rewrite the body to describe that.

**2. `fix(workflows):` on a test-only change, and `Closes #6943` would auto-close an issue this diff does not fix.**

The conventional-commit type should be `test(workflows):`. More importantly, if #6943 was fixed by the already-merged `persist_definition` work, closing it from here attributes the fix to a test and the release notes will say so. Reference it (`Refs #6943`) and close it from the PR that actually fixed it.

**3. The title overclaims against a gap that is still open and documented.**

"persist workflow definitions" reads as if the persistence hole is now closed. `register` — the non-unique path behind `POST /api/workflows`, `workflow create`, the TUI wizard and the dashboard — still answers with an id when the definition never reached disk, and the comment at `workflow.rs:2283-2300` says so explicitly and explains why it was deferred. A title that implies otherwise makes that known gap harder to find later.

## Non-blocking

- `assert_eq!(json_files.len(), 1, …)` proves the count but leaves the identity to the reload step. One extra assertion that the surviving filename is `format!("{}.workflow.json", winners[0])` localizes the failure to "wrong file survived" instead of "reload found the wrong workflow".
- Consider asserting the file's contents deserialize to a workflow whose `id == winners[0]`, so a future change that renames files without rewriting contents cannot pass.

houko (commented):
Downgrading my earlier review from "request changes" to a comment — none of what I raised needs a code change.

The test itself is good and I have no objection to it: it covers a property nothing else does, and both filesystem assertions actually bite (`persist_definition` writes `{id}.workflow.json.tmp` and renames to `{id}.workflow.json`, `workflow.rs:2240-2241`, so neither the `.workflow.json` filter nor the `.tmp` filter is vacuous).

What is left is metadata, fixable in the PR body and the squash message:

- The body describes `persist_definition` returning `Result<()>` and the `existing_id` assertions as if this PR adds them. Both are already on `main` (`workflow.rs:2235`, and `loser_incumbents.push(taken.existing_id)` at `workflow.rs:7859`). The diff is `+74/-0`: one new test.
- `fix(workflows):` on a test-only change should be `test(workflows):`.
- `Closes #6943` will auto-close an issue this diff does not fix. If the already-merged work fixed it, close it from there; otherwise `Refs #6943`.

The last one is the only one with a consequence after merge, and it is worth fixing before rather than reopening the issue after.

Two optional strengthenings, no need to act on them here: assert the surviving filename is `format!("{}.workflow.json", winners[0])` so a failure says "wrong file survived" rather than "reload found the wrong workflow", and assert the file's contents deserialize to a workflow whose `id` matches.

houko (changes requested):
The new test is worth having — `concurrent_register_unique_name_admits_exactly_one` runs against `WorkflowEngine::new()`, so the persistence-enabled variant of the same race genuinely was uncovered, and the three things this asserts (one `.workflow.json` on disk, no `.tmp` stragglers, a reload seeing only the winner) are the right three. Two things to fix first, both small.

## 1. The Summary describes a change that is not in this diff, and is already on `main`

> Strengthens `concurrent_register_unique_name_admits_exactly_one` to verify every rejected caller's `existing_id` points at the single winner, proving the write-lock reservation is fully linearized.

That assertion is already on `main` — `crates/librefang-kernel/src/workflow.rs:7875-7878`:

```rust
assert!(
    loser_incumbents.iter().all(|id| *id == winners[0]),
    "every rejected caller must point at the same incumbent"
);
```

and the diff does not touch that test at all. What the diff actually does is add a new test, `concurrent_register_unique_name_persists_exactly_one`, about the on-disk outcome of the race. Please rewrite the Summary to describe that, since the body is what the release notes and the review record are built from.

## 2. The new test discards the `NameTaken` payload its sibling asserts on

`Err(RegisterWorkflowError::NameTaken(_)) => {}` throws away exactly the field the body claims to be checking. Its in-memory sibling twenty lines up collects it and asserts every loser points at the winner; with persistence on, that linearization property is *more* interesting, not less, because each loser's rejection now races a disk write.

Since the body promises this and the pattern is already established directly above, please assert it here too — accumulate `taken.existing_id`, assert `taken.name == "deploy"`, and assert all the ids equal `winners[0]`. That also gives the test a reason to fail if 31 callers start failing some other way that happens to keep `winners.len() == 1`.

## Not a problem

`reloaded.get_workflow(winners[0])` returning `Some("deploy")` after `load_from_dir_sync` is a real property worth pinning — it asserts the reload preserves the winner's id rather than minting a new one, which the in-memory test cannot check. Good addition; no change needed.

DaBlitzStein:
Pushed 1a0a825d0 on top of the branch's latest main merge: the persisted-path test now asserts every loser's collision payload — losers must fail with `RegisterWorkflowError::NameTaken` (any other variant panics the test instead of passing silently), the taken name must be the raced name, and each loser's `existing_id` must point at the single winner.

Verified on the pushed tree: `cargo nextest run -p librefang-kernel -E 'test(concurrent_register)'` — 3/3 passed by name (`concurrent_register_unique_name_admits_exactly_one`, `concurrent_register_unique_name_persists_exactly_one`, `find_by_name_is_atomic_under_concurrent_register_and_remove`), `cargo clippy -p librefang-kernel --all-targets -- -D warnings` clean. The whole PR remains a single-file diff vs current main (workflow.rs, +87/-0).

houko (approved):
Re-reviewed at the current head. Both items are addressed.

**1 — the Summary describes the diff now.** It names `concurrent_register_unique_name_persists_exactly_one`, and the third bullet does the thing I actually wanted, which was more than correcting the claim:

> The workflow-definition persistence machinery this test exercises … is already on `main` (`workflow.rs:2235`, `:7859`) — this PR only adds the test.

A test-only PR saying so explicitly is worth the line: it tells a reader of the release record that nothing shipped behind this, and it stops the machinery being credited twice when the PR that added it is also in the notes.

**2 — the `NameTaken` payload is asserted.** `taken.name == "deploy"`, `existing_id` collected, and every id checked against `winners[0]`. As I said, that property is more interesting with persistence on than off, because each rejection now races a disk write — and it gives the test a way to fail if 31 callers start failing some other way that happens to leave `winners.len() == 1` intact.

Approving.

DaBlitzStein:
Status note, no code change requested.

This is approved and `MERGEABLE`, but GitHub reports it `UNSTABLE`, and the single non-success check is `stale: cancelled` — not a test failure and not a conflict. Ten of the eleven PRs you approved on 2026-09-03 are in exactly that state; the eleventh (#7749) is `CLEAN`.

Flagging it because `stale` is a housekeeping check rather than one of the required ones, so `UNSTABLE` here does not mean the branch is unhealthy — but it is enough to hold auto-merge, and from the outside that is indistinguishable from a PR still waiting on something real. Whatever ordering you prefer for the batch, this one is not waiting on me.

I have also just posted the per-PR responses to your earlier reviews across the rest of my open PRs. My apologies for the delay on those: I pushed the fixes on 2026-09-05 and did not comment on any of them, so from your side the heads changed silently and there was nothing marking them as ready to look at again. That was my mistake and I have corrected it.

crates/librefang-kernel/src/workflow.rs
```diff
@@ -8061,6 +8061,93 @@ prompt_template = "go"
...
+            "every rejected caller must point at the same incumbent"
+        );
+        assert_eq!(
+            count_named(&engine, "deploy").await,
+            1,
+            "the registry must hold exactly one 'deploy'"
+        );
+
+        let workflows_dir = tmp.path().join("workflows");
+        let on_disk: Vec<_> = std::fs::read_dir(&workflows_dir)
```
houko, line 8115: The three on-disk assertions here cannot fail for the reason the PR body gives, because no rejected caller ever touches the filesystem.

`register_unique_name` returns `NameTaken` from *inside* the write-lock block, before `persist_definition` is reached:

```rust
{
    let mut registry = self.workflows.write().await;
    if let Some(existing) = registry.values().find(...) {
        return Err(RegisterWorkflowError::NameTaken(...));   // <- returns here
    }
    registry.insert(id, workflow.clone());
}
if let Err(e) = self.persist_definition(&workflow).await { ... }   // <- winner only
```

So of the 32 callers exactly one reaches `persist_definition`, and it does a single `create_dir_all` + single tmp write + single rename. `json_files.len() == 1` and `tmp_files.is_empty()` therefore hold by construction of the early return, not by anything about the race — which makes the PR body's claim that this is "the path where each rejection races a disk write" inaccurate: rejections never reach a disk write.

Concretely, the risk is false confidence. `register_unique_name_persists_the_definition` (`workflow.rs:7942`) already pins winner-persists-and-reloads, `register_writes_atomically_and_cleans_tmp` (`:8156`) already pins the tmp cleanup, and `concurrent_register_unique_name_admits_exactly_one` (`:7840`) already pins the linearization property including `existing_id == winner`. A reader seeing this test will assume concurrent *definition writes* are covered, and they are not — the interleaving that actually exercises concurrent filesystem work is N callers with **distinct** names racing into a not-yet-created `workflows/` directory, where several `create_dir_all` calls and several tmp-write/rename pairs genuinely overlap. If the test is meant to earn its 90 lines, that is the shape it needs; as written the only regression it can catch is `persist_definition` being moved above the name check.

crates/librefang-kernel/src/workflow.rs
```diff
@@ -8061,6 +8061,93 @@ prompt_template = "go"
...
+        let engine = WorkflowEngine::new_with_persistence(tmp.path());
+        let mut handles = Vec::with_capacity(CALLERS);
+        for _ in 0..CALLERS {
+            let engine = engine.clone();
+            handles.push(tokio::spawn(async move {
+                engine.register_unique_name(named_workflow("deploy")).await
+            }));
+        }
+
+        let mut winners = Vec::new();
```
houko, line 8081: Lines 8073-8112 are a verbatim copy of lines 8073-8112's counterpart in `concurrent_register_unique_name_admits_exactly_one` (`workflow.rs:7841-7880`) — the spawn loop, the `winners` / `loser_incumbents` join-and-match, and all four assertions, differing only in the assertion message strings.

Two copies of the same race harness drift: a future fix to the in-memory variant (a new error variant to match on, a change in how the incumbent id is reported) gets applied to one and silently not the other, and the copy that was missed keeps passing while asserting the old contract.

Extract the shared part into a helper next to `count_named`, e.g. `async fn race_unique_name(engine: &WorkflowEngine, callers: usize) -> (Vec<WorkflowId>, Vec<WorkflowId>)` returning `(winners, loser_incumbents)` with the panic-on-unexpected-variant inside it, and let each test keep only the assertions specific to it (`count_named` for the in-memory one, the disk + reload block for this one).

houko (approved):
LGTM — merging.

Test-only change: all 87 lines land inside the existing `#[cfg(test)] mod tests` block and reference the `named_workflow` / `count_named` helpers, so there is no production surface. The assertions hold as written, and the on-disk half is a real guard against `persist_definition` being moved above the name check in `register_unique_name`.

The two inline comments are non-blocking notes for a follow-up rather than merge gates: the on-disk assertions cannot fail for the reason the PR body gives (losers return from inside the write-lock block at `workflow.rs:2325`, before `persist_definition`), and the harness body duplicates `concurrent_register_unique_name_admits_exactly_one`. Neither affects correctness of what is being merged.

PR #8008, opened 2026-08-30
fix(ci): deploy the demo from the image the release just pushed

houko:
[Run 33283348874](https://github.com/librefang/librefang/actions/runs/33283348874) — `Deploy to Fly.io` on `v2026.8.30`.

## Failure

```
Error: failed to fetch an image or build from source: failed to run query ...
Could not find image "ghcr.io/librefang/librefang:v2026.7.31"
```

## Cause

`deploy/fly/fly.toml` pinned the image to `v2026.7.31`. That is two releases stale, and the image is no longer in ghcr.io, so the deploy had nothing to pull.

The pin was wrong even while that image existed: the official public demo has been serving July's build after every release since.

## Fix

The workflow passes `--image` with the tag being released. `docker-manifest` — which this job already `needs` — has just pushed both `:$VERSION` and `:latest`, so the tag is guaranteed to exist by the time the deploy runs.

The release-specific tag belongs in the workflow for the same reason the `LIBREFANG_ALLOW_NO_AUTH` exception already lives there, per the comment above it: `fly.toml` is a file users copy, and a release tag baked into it goes stale the moment the next release lands.

`fly.toml` itself moves to `:latest` — what a user copying the file should get, and a value that cannot rot the way the pin did.

`--remote-only` is removed rather than left in place: with `--image` there is nothing to build, so the flag had no meaning.

## Verification

- `fly.toml` parses and resolves to `ghcr.io/librefang/librefang:latest`.
- The rendered command is `flyctl deploy --config deploy/fly/fly.toml --image "ghcr.io/librefang/librefang:${RELEASE_TAG}" --env LIBREFANG_ALLOW_NO_AUTH=1`.
- `actionlint` findings on `release.yml` are unchanged at 30 — the first draft left `${RELEASE_TAG}` unquoted and added an SC2086; it is quoted now.

## Related

The cross-target build failures in the same run are #8007.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3088,10 +3088,11 @@ jobs:
         # Keep that exception out of fly.toml so user deployments remain secure by default and must explicitly provision authentication.
         # The image is pinned to the tag being released for the same reason the auth exception lives here: fly.toml is a file users copy, and a release-specific tag baked into it goes stale the moment the next release lands.
         # `docker-manifest`, which this job needs, has just pushed both `:$VERSION` and `:latest`, so the tag is guaranteed to exist.
+        # `${RELEASE_TAG#v}` strips the `v` prefix to match `$VERSION` as pushed by `docker-manifest` — `RELEASE_TAG` itself (e.g. `v2026.7.31`) was never a tag pushed to the registry.
         # `--image` means there is nothing to build, which is why `--remote-only` is gone rather than merely unused.
         run: >-
           flyctl deploy --config deploy/fly/fly.toml
-          --image "ghcr.io/librefang/librefang:${RELEASE_TAG}"
+          --image "ghcr.io/librefang/librefang:${RELEASE_TAG#v}"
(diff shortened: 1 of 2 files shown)


houko:
Automated review found one bug in the deploy step and fixed it in 0c090fc:

`--image "ghcr.io/librefang/librefang:${RELEASE_TAG}"` used the raw `v`-prefixed tag, but `docker-manifest` (this job's own dependency) only pushes `$VERSION` (the `v` stripped) and `latest` — never a tag matching `RELEASE_TAG` itself. As written, the deploy would have failed to find the image, defeating the fix this PR set out to make. Changed to `${RELEASE_TAG#v}`, matching every other job in this workflow that derives an image/package version from `RELEASE_TAG`.

Also rewrapped the new `fly.toml` comment onto sentence-boundary lines per this repo's prose-wrapping convention (CLAUDE.md) — it was hard-wrapped mid-sentence.

---

PR #8009, opened 2026-08-30
fix(ci): keep a release in draft until its artifacts are signed

houko:
Raised by the maintainer looking at v2026.8.30: if a broken run still publishes an unsigned release, what is the signing job's error for?

Nothing. That is the bug.

## What the pipeline did

```
t=0   gh release create        ← published, empty, public
      ├─ 18 jobs upload artifacts into it as they finish
      └─ sign_release_artifacts ← leaf; nothing depends on it
```

`create_release` ran `gh release create` with **no `--draft`**, so the release was public before a single build job started. Every failure after that point left a permanently public release that was incomplete and unsigned.

**v2026.8.30 is that release**: 23 of the 48 assets its predecessor shipped, and no `SHA256SUMS`, `SHA256SUMS.sig` or `SHA256SUMS.pem` at all. Users could download binaries from a page that looks official, with nothing to verify them against.

`sign_release_artifacts` behaved correctly throughout — it waits on all eight CLI jobs and refuses to sign a platform set that does not match, exactly as #7305 hardened it to. It is also a leaf: **nothing in the 36-job graph depends on it**, so refusing to sign changed nothing anyone could observe. A gate that runs after the thing it guards has already shipped is not a gate.

## What it does now

```
gh release create --draft
  ├─ artifacts …
  ├─ sign_release_artifacts
  └─ publish_release → gh release edit --draft=false
        └─ sync_homebrew, sync_homebrew_cask
```

**Gates publication:** `sign_release_artifacts` and `desktop`. The former already requires every CLI job and validates the exact checksum asset set, so requiring it requires them; the latter contributes bundles of its own.

**Deliberately does not gate:**
- `mobile_ios` / `mobile_android` — both are `continue-on-error`. A job the pipeline already treats as best-effort must not hold back a finished release. (This is also why the current iOS certificate failure would not block anything.)
- npm, PyPI, AUR, the Arch repo — each publishes to a registry of its own. A flaky mirror must not keep a complete release unpublished.

**Waits for publication:** `sync_homebrew` and `sync_homebrew_cask` only, because they bake a `releases/download/...` URL into a formula and that URL 404s while the release is a draft. No other distribution job reads release assets, so nothing else is serialised behind the gate for nothing.

An existing release being rebuilt is returned to draft too — otherwise re-running a broken release leaves the incomplete one public for the length of the rebuild.

## Verification

Topology asserted by parsing the workflow: the gate's ancestors include all eight CLI jobs, `sign_release_artifacts` and `desktop`; exclude both mobile jobs; both Homebrew jobs sit downstream of it; no dependency cycles anywhere in the 37 jobs.

`test_release_tag_workflow_safety.py` — already CI-gated — now covers it, and each assertion was confirmed against its regression rather than assumed:

| mutation | result |
|---|---|
| drop `--draft` | `release.yml runs `gh release create` without --draft` |
| delete `publish_release` | `release.yml has no publish_release job to leave draft state` |
| drop `desktop` from the gate | `publish_release does not wait for desktop` |
| cask jumps ahead of publication | `sync_homebrew_cask writes a download URL before the release is published` |

`actionlint` findings unchanged at 30.

## Behaviour change worth knowing

A failing release is now invisible rather than public-and-broken. That is the point, but it does mean the release page no longer fills in live during a run.

## Does not repair v2026.8.30

That release is already public. It is being repaired separately through `release-cli.yml` / `release-desktop.yml` (#8007), and its npm packages are already republished.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -426,11 +426,14 @@ jobs:
 
           if gh release view "$TAG" &>/dev/null; then
             echo "Release $TAG already exists, updating..."
-            gh release edit "$TAG" \
+            gh release edit "$TAG" --draft \
               --title "$TAG" \
               --notes-file /tmp/release-body.md
           else
-            gh release create "$TAG" \
+            # Created as a draft on purpose.
+            # A published release is what users and every mirror read, so publishing one before a single artifact exists means any job failure leaves a permanently public, incomplete, unsigned release — which is exactly what v2026.8.30 became.
+            # `publish_release` flips it once the artifacts are all there and signed.
+            gh release create "$TAG" --draft \
               --title "$TAG" \
               --notes-file /tmp/release-body.md
           fi
@@ -1285,6 +1288,30 @@ jobs:
             SHA256SUMS.pem \
             --clobber
 
+  # The release is a draft until here.
+  # `sign_release_artifacts` already waits for all eight CLI jobs and refuses to sign an incomplete platform set, but nothing depended on it, so its verdict changed nothing about what the world could already download.
+  # Gating publication on it is what turns that check into a gate.
+  #
+  # `desktop` joins it because its bundles are release assets too.
+  # `mobile_ios` and `mobile_android` deliberately do not: both are `continue-on-error`, so gating on them would let a best-effort job hold back a complete release.
+  # Neither do the distribution jobs — npm, PyPI, AUR and Homebrew publish to registries of their own, and a flaky mirror must not keep a finished release unpublished.
+  publish_release:
+    name: Publish release
+    needs: [create_release, sign_release_artifacts, desktop]
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: write
+    steps:
+      - name: Flip the release out of draft
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          set -euo pipefail
+          gh release edit "$RELEASE_TAG" --draft=false
+          echo "Published $RELEASE_TAG"
+
   cli_npm:
     name: CLI / npm
     if: >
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -1285,6 +1288,30 @@ jobs:
...
+  # The release is a draft until here.
+  # `sign_release_artifacts` already waits for all eight CLI jobs and refuses to sign an incomplete platform set, but nothing depended on it, so its verdict changed nothing about what the world could already download.
+  # Gating publication on it is what turns that check into a gate.
+  #
+  # `desktop` joins it because its bundles are release assets too.
+  # `mobile_ios` and `mobile_android` deliberately do not: both are `continue-on-error`, so gating on them would let a best-effort job hold back a complete release.
+  # Neither do the distribution jobs — npm, PyPI, AUR and Homebrew publish to registries of their own, and a flaky mirror must not keep a finished release unpublished.
+  publish_release:
+    name: Publish release
+    needs: [create_release, sign_release_artifacts, desktop]
```
houko, line 1300: Nit, not a bug: every other job in this file (`create_release`, `sign_release_artifacts`, `desktop`, `cli_npm`, `sync_homebrew`, …) carries the same explicit

```
if: >
  (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v'))
  || (github.event_name == 'workflow_dispatch' && inputs.channel == 'current' && startsWith(inputs.tag, 'v'))
```

`publish_release` is the only job in the whole workflow without it. It's not a functional gap — GitHub Actions skips a job by default when any of its `needs` were skipped, and all three of `create_release`/`sign_release_artifacts`/`desktop` already carry this condition, so `publish_release` cascades to skipped on the `pull_request: closed` trigger (`tag_on_merge`) and on other `workflow_dispatch` channels. Just flagging the asymmetry in case it was meant to be defense-in-depth like everywhere else in the file, since a future edit to any of the three `needs` jobs' conditions would silently remove that inherited protection.

---

houko (commented):
Reviewed the release-draft-until-signed change. The gating logic checks out — I traced `needs` across `publish_release`, `sync_homebrew`, and `sync_homebrew_cask`, confirmed `RELEASE_TAG` is a workflow-level env available to the new job, and ran both `scripts/tests/test_release_tag_workflow_safety.py` and `scripts/check-changelog-attribution.py` locally against this branch — all pass. One non-blocking nit left inline about a missing `if:` condition (functionally covered by `needs` skip-propagation, so not a real bug).

---

Issue #5740, opened 2026-05-25
feat(media): support self-hosted STT/TTS providers (Whisper local, Piper, custom URL)


DaBlitzStein:
## Problem

The STT provider list in `media_understanding.rs` is hardcoded to 6 cloud providers (Groq, OpenAI, MiniMax, Fireworks, Together, SiliconFlow). There is no way to point STT at a self-hosted Whisper-compatible server or use offline TTS like Piper.

For on-prem/air-gapped deployments, all voice processing must stay local. Current implementation makes this impossible.

### STT: custom Whisper-compatible URL

Allow `audio_provider = "custom"` (or any unrecognized name) with a configurable URL:

```toml
[media]
audio_provider = "local-whisper"

[media.providers.local-whisper]
url = "http://localhost:8080/v1/audio/transcriptions"
api_key = ""  # optional, local servers often don't need auth
model = "large-v3"
```

The Whisper API format is a de-facto standard — faster-whisper-server, whisper.cpp server, and others all implement the same `/v1/audio/transcriptions` multipart endpoint. No protocol changes needed, just allow custom URLs.

### TTS: Piper and custom URL

Same pattern for TTS:

```toml
[tts]
provider = "piper"

[tts.providers.piper]
url = "http://localhost:5000/api/tts"
voice = "en_US-lessac-medium"
```

Piper runs as a local HTTP server and produces high-quality speech offline. Other self-hosted options: Coqui TTS, VITS, Bark.

### Implementation notes

- `whisper_provider_config()` currently returns `Err` for unknown providers — change to check `[media.providers.<name>]` config section before erroring
- `TtsConfig` already has `stt_url`/`tts_url` fields but the STT code path doesn't use them — unify
- No new dependencies needed — all self-hosted servers expose HTTP APIs

Issue #2667, opened 2026-04-16
feat: allow user locally define text-to-image/text-to-video model's provider, and support to use these providers in media panel.


szdebian:
### Description

For the latest version: v2026.4.15-beta22
1. Although ollama is well configured and tested in the "provider" panel, it can NOT be selected in the "media" panel.
2. Neither the custom provider (custon.toml) prepared in the registry and shown in the provider list.

leszek3737:
Ollama does not support video or image models

szdebian:
forget ollama.

for example, use can define a provider of volcengine's seedance/seedream in China mainland or TongYiWanXiang's wan2.7

leszek3737:
Currently, the media driver system in librefang-runtime/src/media/ works differently from LLM drivers:
- LLM drivers (librefang-llm-drivers) already support this — any custom provider with a base_url configured automatically uses OpenAI-compatible format, so you can point it at Volcengine, Tongyi, NVIDIA, etc.
- Media drivers are hardcoded in a match factory function (media/mod.rs:306). Each provider (MiniMax, OpenAI, Gemini, ElevenLabs, Google TTS) has its own Rust module with protocol-specific API calls. There's no way to add a custom media provider without a PR.

PR #8012, opened 2026-08-30
feat(media): add custom image description provider support

DaBlitzStein:
## Summary

- Add `CustomImageConfig` to `MediaConfig` in `media.rs`, mirroring the existing `CustomSttConfig` pattern for audio transcription
- Wire the `_other` dispatch arm of `describe_image()` to forward to the configured OpenAI-compatible chat completions endpoint
- Add `custom_image_config()` and `custom_image_model_ref()` helpers following the same pattern as `custom_stt_config()` / `custom_stt_model_ref()`

Operators can now use self-hosted vision models (LLaVA, Qwen-VL, etc.) for image understanding by configuring:

```toml
[media]
image_provider = "local-llava"

[media.custom_image]
base_url = "http://localhost:8080/v1/chat/completions"
model = "llava-v1.6"
# api_key_env = "MY_LOCAL_KEY"  # optional
```

Related: #2667 (custom media providers), #5740 (self-hosted STT/TTS)

## Test plan

- [ ] `cargo check -p librefang-runtime-media` passes
- [ ] Verify custom image provider dispatch with a local OpenAI-compatible vision endpoint
- [ ] Verify built-in providers (anthropic, openai, groq, gemini) are unaffected by the new config

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,7 +149,7 @@ jobs:
           # `openapi-drift` job still gates direct edits.
           RUST=false;    has '^(crates/|Cargo\.(toml|lock)$|xtask/|openapi\.json$|sdk/)' && RUST=true
           DOCS=false;    has '^(docs/|.*\.md$)'                                       && DOCS=true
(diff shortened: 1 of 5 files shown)


houko (commented):
`Test / Unit (lib+bin)` is red on a named drift guard, and its message names the exact fix:

```
POST /api/config/set accepts these paths but GET /api/config omits them, so the dashboard
renders each one blank and reads it back as "not configured" right after a successful save (#6596).
Add every path to `redacted_config_json`: [
    "media.custom_image",
    "media.custom_image.base_url",
    "media.custom_image.key_required",
    "media.custom_image.model",
]
```

(`routes::config::manage::config_read_write_parity_tests::every_writable_config_leaf_is_readable`, `crates/librefang-api/src/routes/config/manage.rs:1740` — 2258 passed, 1 failed.)

This is the both-directions rule for config fields: the write path accepts `media.custom_image.*`, the read path does not emit it, so the dashboard saves the provider successfully and then shows it as unset on the next read.
Registering the four paths in `redacted_config_json` is the whole change; `key_required` is a boolean rather than a secret, so nothing here needs redacting, only listing.

`Test / Ubuntu (shard 1/4)` is the same test in the other lane, and `CI Gate` is red only because both are in its `needs`.

Issue #7834, opened 2026-08-24
feat(tui): an agent's shared folders ([workspaces]) cannot be seen or edited from the terminal


DaBlitzStein:
An agent's `[workspaces]` — the shared directories it can read or write outside its own private workspace — is editable from the dashboard's agent editor and by hand-editing `agent.toml`.

The terminal has nothing.

`librefang tui` → Agents → agent detail offers `[s] Edit skills`, `[m] Edit MCP`, `[c] Chat`, `[k] Kill`.
There is no way to see which shared folders an agent has, let alone add or remove one.

That matters more than the other two allowlists, because a shared folder is how two agents collaborate on the same files: `@name → /abs/path (mode)` is injected into TOOLS.md from this table, so an agent that is missing a declaration simply cannot see the directory the operator expects it to work in.
Diagnosing that from a terminal session today means reading `agent.toml` in another window.

## Expected

An editor reachable from the agent detail that lists the declared folders as `(name, path, mode)`, and can add, edit and delete rows.

## Note on the write path

`PATCH /api/agents/{id}` with `manifest_toml` replaces the whole manifest, so the editor has to read the current manifest first and replace only the `[workspaces]` table.
Building a manifest from the editor's own fields would silently drop every field the editor does not render.

A row with an empty name or path is an abandoned edit, not a declaration — writing it produces a `[workspaces]` entry the kernel then fails to resolve, so those rows should be dropped before the write rather than persisted.

houko:
Triaged against `origin/main` @ 8a947885b. Premise confirmed verbatim.

## Premise

The agent-detail hint string is exactly what you quoted, at `crates/librefang-cli/locales/en/main.ftl:2014`:

```
tui-agents-hints-detail =   [s] Edit skills  [m] Edit MCP  [c] Chat  [k] Kill  [Esc] Back
```

`crates/librefang-cli/src/tui/screens/agents.rs` has sub-screens for skills (`:487`) and MCP (`:495`) and no occurrence of `workspaces` anywhere in the file. So there is no read view and no editor, as filed.

Your note on the write path is right and worth keeping in the PR description: `PATCH /api/agents/{id}` with `manifest_toml` replaces the manifest wholesale, so the editor must read-modify-write the `[workspaces]` table rather than reconstruct a manifest from its own fields.

## `has-pr`: PR #7835 is not reviewable in its current shape

[...]

PR #8013, opened 2026-08-30
feat(dashboard): add shared folders editor to agent manifest form

DaBlitzStein:
## Summary

- Adds a collapsible **Shared Folders** section to `AgentManifestForm` with name / path / mode editing for the `[workspaces]` table of agent manifests, bringing the dashboard to parity with the TUI editor (#7834 / #7835) — the backend (`workspaces: HashMap<String, WorkspaceDecl>`) already supports it.
- The serializer previously emitted the `[workspaces]` table header in the middle of the top-level scalar block: in TOML a table header scopes every bare key that follows it, so the first save with a shared folder would have silently scoped the remaining top-level scalars into `[workspaces]` — tags, skills, mcp_servers, schedule and the rest would be gone on reload. The header now emits after all top-level scalars and extras, so the round-trip survives.
- Preserved `[workspaces.<name>]` declarations (including `mount`) round-trip untouched through the deferred-extras path.
- Validation: duplicate folder names are rejected in the form, and folder paths reject POSIX-absolute, UNC (`\\…`), drive-rooted (`C:\…`) and any `..` segment — matching the kernel's `has_unsafe_relative_components` so client and daemon agree on what is unsafe.
- i18n keys added in all five locales (en, ko, pl, uk, zh) with real translations, including the two validation messages.

## Verification

- `npx vitest run src/lib/agentManifest.test.ts` — 66/66 passed on the pushed tip, including the ordering test (emits `[workspaces]` after the top-level scalars so tags survive a round-trip), mount preservation with a reparse, duplicate-name rejection, and the path-shape rejection tests.
- Serialize → parse → serialize → parse → serialize is byte-stable across cycles: a single `[workspaces]` header after the scalars, sub-tables emitted exactly once, no duplicated rows, `[tools]` extras preserved.
- `npx tsc --noEmit` — clean.
- `scripts/i18n-parity.mjs` — the failing-locale set (ko / pl / uk) is identical to plain main (pre-existing drift, see #3557); zh at parity; this branch adds no new drift.
- Changelog fragment: `changelog.d/added/8013-dashboard-shared-folders.md`.


houko (commented):
CI is red on a real failure, not on the usual `stale` noise: `src/lib/__tests__/locale-parity.test.ts` fails for `ko.json` and `pl.json`.

The six new keys landed in `en.json`, `uk.json` and `zh.json` but not in the other two:

```
agents.form.add_folder
agents.form.folder_name
agents.form.folder_path
agents.form.remove_folder
agents.form.shared_folders
agents.form.shared_folders_hint
```

`crates/librefang-api/dashboard/src/locales/` holds five files — `en`, `ko`, `pl`, `uk`, `zh` — and the changed-files list on this PR names only three of them.
Adding the same six keys to `ko.json` and `pl.json` is the whole fix.
(Run: https://github.com/librefang/librefang/actions/runs/33296044665/job/99215762287)

While the branch is open: this overlaps #7835 (`feat(tui): shared folders editor for agents`), which is the same feature on the TUI side and is currently `BLOCKED` on conflicts.
Worth landing them in a deliberate order so the two editors agree on how a folder entry is shaped — `crates/librefang-api/dashboard/src/lib/agentManifest.ts` here is the side that decides what gets written into `agent.toml`.

houko (changes requested):
The editor and the parser are fine, but the serializer emits `[workspaces]` in the middle of the top-level scalars, which silently moves nine other settings into that table.

## Blocking

**1. `[workspaces]` is emitted before the remaining top-level keys, so TOML swallows them.**

The new block lands at `agentManifest.ts:456`, immediately before:

```ts
if (form.tags.length) lines.push(`tags = ${tomlArray(form.tags)}`);
if (form.skills.length) lines.push(`skills = ${tomlArray(form.skills)}`);
if (form.mcp_servers.length) lines.push(`mcp_servers = ${tomlArray(form.mcp_servers)}`);
if (form.tool_allowlist.length) …
if (form.tool_blocklist.length) …
if (form.allowed_plugins.length) …
const scheduleLine = renderSchedule(form.schedule);
if (form.exec_policy_shorthand) …
const responseFormatLine = renderResponseFormat(form.response_format);
```

In TOML every bare key after a `[table]` header belongs to that table. So saving an agent that has even one shared folder produces:

```toml
[workspaces]
shared = { path = "shared" }
tags = ["ops"]
skills = ["deploy"]
mcp_servers = ["github"]
tool_allowlist = ["file_read"]
exec_policy = "…"
```

— i.e. `workspaces.tags`, `workspaces.skills`, `workspaces.mcp_servers`, `workspaces.tool_allowlist`, `workspaces.tool_blocklist`, `workspaces.allowed_plugins`, `workspaces.schedule`, `workspaces.exec_policy`, `workspaces.response_format`, plus every preserved top-level extra scalar (the block at `agentManifest.ts:469+`). The agent loses its tags, skills, MCP servers, tool allow/blocklist and schedule the first time an operator adds a shared folder, and `[workspaces]` itself will not deserialize into `BTreeMap<String, NamedWorkspace>` with those keys present.

The invariant is written down four lines below the insertion point:

> Top-level extras — split scalars (BEFORE table headers) and tables (AFTER).

`[workspaces]` is a table header, so it belongs with `[model]` (`agentManifest.ts:516`), `[resources]` (`:531`), `[capabilities]` (`:547`) and the rest — after the scalar block, not inside it.

**2. No round-trip test.**

`serializeManifestForm` / `parseManifestToml` are pure functions and #1 is exactly what a round-trip assertion catches: build a form with one workspace *and* non-empty `tags`, serialize, re-parse, assert `tags` survived at top level and `workspaces` has one entry. Please add that plus the empty case (no `[workspaces]` header emitted when the list is empty or every row is blank).

## Non-blocking

- Duplicate folder names produce a duplicate TOML key and a hard parse failure on the daemon. `tomlBareKeyOrQuoted` quotes the key but cannot make it unique — reject or de-duplicate in the form, or at least surface a validation message on the row.
- `path` is accepted verbatim. Per `CLAUDE.md` these are "relative to `workspaces_dir`", so an absolute path or one containing `..` is either rejected later or escapes the intended root. Worth confirming what `ensure_named_workspaces()` does with those and mirroring the rule in the form, so the operator finds out at type time rather than at spawn time.
- `ko.json` and `pl.json` get English values while `uk` and `zh` are properly translated. Every one of these `t()` calls has no `defaultValue`, so a missing key would render as the raw key — which is why the English placeholders were added. Given that, this is defensible, but flag the two files for translation rather than leaving them looking done.
- Missing changelog fragment (`changelog.d/added/8013-....md`).
- #7835 ("shared folders editor" for the TUI), #8114 and #8028 all touch this same area. Worth confirming with the others which PR owns `ManifestFormState.workspaces` so the field is not defined twice with different shapes.

houko:

No commits of your own since my review on 2026-09-01T13:32:39Z; the only thing that moved the branch tip is a merge from `main`, which the `auto-update-branches` cron pushes by itself. The PR's "updated" timestamp has therefore advanced while the blocking items have not been touched. Worth flagging because that timestamp is what makes a PR look answered from outside — I nearly mis-triaged a dozen PRs on exactly that signal today.

Still open from that review:

- 1. `[workspaces]` is emitted before the remaining top-level keys, so TOML swallows them
- 2. No round-trip test

Also worth knowing: #7749 (the dashboard's full-manifest editor) is approved and will land, and my review there flagged that this PR's `[workspaces]` emission orders the table before the remaining top-level scalars. Whichever of the two lands second should check the ordering, because the editor #7749 points at the running agent is what would write the reordered form.

DaBlitzStein:
Refreshed `feat/dashboard-shared-folders`: 763173627 is the feature commit re-merged onto the current fork tip (which carried the cron main-merge), pushed fast-forward.

Re-verified on the merged tree:

- `npx vitest run src/lib/agentManifest.test.ts` — 66/66 passed, including "emits [workspaces] after the top-level scalars so tags survive a round-trip".
- `npx tsc --noEmit` — clean.
- `node scripts/i18n-parity.mjs` — failing set ko / pl / uk is identical to plain main (pre-existing, #3557); zh at parity.

PR body rewritten to match what actually ships: the TOML table-header scoping defect is now the headline, all five locales listed, and the verification section carries the numbers above.

DaBlitzStein:
Merge-order note for whoever lands these, not a change request: this PR and #7787 (step ladders for the token-count fields) both modify the agent manifest form, and neither is on `main` yet, so the second one to merge will conflict.

The overlap is eight files:

- `src/lib/agentManifest.ts` and its test — this PR changes where the `[workspaces]` table header is emitted relative to the top-level scalars; #7787 changes how the numeric model fields are parsed and clamped. Different concerns in the same serializer.
- `src/components/AgentManifestForm.tsx` — the shared-folders section here, the step-ladder inputs there.
- All five `locales/*.json`, both adding keys.

Nothing contradicts: they are additive in different regions of the same files. Whichever lands second just needs a merge that keeps both sides rather than taking one wholesale — the risk is a resolution that drops the other PR's hunk, which in the serializer would be silent (the ordering fix here has no compile error to catch it, only the round-trip test).

Both branches are currently green.

houko (dismissed):
Re-reviewed at the current head. The serializer bug is fixed, and the comment left behind is the part that keeps it fixed.

`[workspaces]` is now emitted after every top-level scalar, immediately before `[model]`, with the reason recorded at the emit site:

> table header, so it is emitted here, after every top-level scalar; a header inside the scalar block would scope the remaining bare keys into the table and silently delete them

That is the whole failure in one sentence, at the place someone would otherwise reintroduce it by moving the block back up next to the other `form.*` fields. Worth more than the fix itself, because the bug was invisible in the diff — nothing about `lines.push("[workspaces]", …)` looks wrong until you know what follows it.

The preserved-entry handling is a good catch I did not raise: re-emitting mount-based or malformed rows as `[workspaces.<name>]` sub-tables rather than merging them into the generated `[workspaces]` header avoids emitting a duplicate header, which TOML rejects outright — so a manifest with a hand-written mount declaration round-trips instead of failing to parse on the way back in.

Approving.

DaBlitzStein:
Merge-order warning for this PR and #8028, found by merging every open branch into a single integration tree. These two are incompatible at the contract level, and the incompatibility produces **no compile error** — only one assertion catches it.

#8013 promotes `[workspaces]` from an unmodelled extra to a first-class field of the manifest form. #8028 carries a round-trip test that asserts the opposite:

```ts
expect(reparsed.extras.topLevel["workspaces"]).toEqual({ notes: … })
```

With both branches in the tree, the `[workspaces]` section *does* survive the round-trip — it simply lives in `form.workspaces` now rather than in `extras.topLevel`. Everything builds, the serializer is correct, and the only signal is that one assertion failing.

Whoever lands second has to update that test in their own PR. There is nothing else to change: the features do not conflict, only the test's assumption about where the section ends up.

In the integration tree the assertion was updated to the post-#8013 contract, checking `form.workspaces`, with a comment recording why. That keeps both intents intact.

Flagged on both PRs. Nothing to do here right now — this is for merge time.

DaBlitzStein:
Pushed `1e2af9aa5`: merged current `main`, which moved to `fe2448bb5` and now carries #7787, #8106 and #8121.

One conflict, in `src/lib/agentManifest.test.ts`: both sides append a new `describe()` at the end of the file — `agentManifest — inference parameters (#7781)` from `main`, `agentManifest workspaces` from here. Purely additive, so both blocks are kept whole.

The mechanical audit confirms nothing else was touched by hand: `git merge-tree --write-tree` against the committed tree reports exactly one path, that test file. Everything else is git's own automerge.

Also checked the parts that did **not** conflict, since those are where silent regressions hide: `agentManifest.ts`'s `[workspaces]` ordering fix and the editor section in `AgentManifestForm.tsx` both automerged clean, and diffing the `.tsx` against `main` confirms the new `<CollapsibleSection>` is appended without duplicating anything from #7787's step ladders.

- `tsc --noEmit` — exit 0.
- `vitest run src/lib/agentManifest.test.ts` — **69/69**, including by name `emits [workspaces] after the top-level scalars so tags survive a round-trip` (the one that catches the silent serializer bug), `preserves a mount-based declaration verbatim instead of dropping it`, and the three inference-parameter tests that came from `main`'s side.
- Full suite: 1810 passing, 1 failing — `PromptsExperimentsModal > stops selection when every variant has a traffic bucket`, a 5 s timeout under load that passes 7/7 in isolation. It reproduces identically in four unrelated worktrees, none of which touch that file, so it is not from this branch.

This PR is dashboard-only: nine files against the merge base, all under `dashboard/` plus its changelog fragment. No Rust to compile.

On the merge-order note from earlier: #8028's round-trip assertion still passes against `origin/main` today, because `main` does not have this PR yet. It only breaks if this one lands first. Deliberately not touching it here — a PR should not carry a change that is only correct if another PR merges before it.

DaBlitzStein:
Status note, no code change requested.

This is approved and `MERGEABLE`, but GitHub reports it `UNSTABLE`, and the single non-success check is `stale: cancelled` — not a test failure and not a conflict. Ten of the eleven PRs you approved on 2026-09-03 are in exactly that state; the eleventh (#7749) is `CLEAN`.

Flagging it because `stale` is a housekeeping check rather than one of the required ones, so `UNSTABLE` here does not mean the branch is unhealthy — but it is enough to hold auto-merge, and from the outside that is indistinguishable from a PR still waiting on something real. Whatever ordering you prefer for the batch, this one is not waiting on me.

I have also just posted the per-PR responses to your earlier reviews across the rest of my open PRs. My apologies for the delay on those: I pushed the fixes on 2026-09-05 and did not comment on any of them, so from your side the heads changed silently and there was nothing marking them as ready to look at again. That was my mistake and I have corrected it.

crates/librefang-api/dashboard/src/lib/agentManifest.ts
```diff
@@ -1100,6 +1169,30 @@ export const parseManifestToml = (toml: string): ParseResult | ParseError => {
...
+  // TUI editor, #7835).
+  if (isTomlTable(parsed.workspaces)) {
+    const preservedWorkspaces: TomlTable = {};
+    for (const [name, v] of Object.entries(parsed.workspaces)) {
+      if (isTomlTable(v) && typeof (v as TomlTable).path === "string") {
+        form.workspaces.push({
+          _uid: generateParsedUid(),
+          name,
+          path: (v as TomlTable).path as string,
+          mode: asString((v as TomlTable).mode) === "r" ? ("r" as const) : ("rw" as const),
```
houko: `"r"` is not the only read-only spelling the daemon accepts, and it is not the one the daemon *writes*.

`WorkspaceMode` is `#[serde(rename_all = "lowercase")]` with `#[serde(alias = "r", alias = "read", alias = "read-only")]` on `ReadOnly` (`crates/librefang-types/src/agent.rs:1661-1671`), so the canonical serialized form is `mode = "readonly"`. That is exactly what `toml::to_string_pretty(&manifest)` produces in `persist_full_manifest_at` (`crates/librefang-kernel/src/kernel/agent_state.rs:109`) and in the template endpoints (`crates/librefang-api/src/routes/agent_templates.rs:406`, `:822` — the latter's bytes become the published template's `agent.toml`).

Concrete path to a privilege upgrade: operator selects such a template, clicks "Edit TOML for advanced customization" (`AgentsPage.tsx:3455-3466`), switches to Form mode, hits Create. `asString(v.mode) === "r"` is false for `"readonly"`, so the row is built as `"rw"`; the serializer only emits `mode` when it is `"r"`, so the key vanishes and `WorkspaceMode::default()` (ReadWrite) applies. A read-only shared folder / mount silently becomes writable.

Suggest matching the kernel's alias set here (and adding a test for `readonly`).

crates/librefang-api/dashboard/src/lib/agentManifest.ts
```diff
@@ -860,8 +902,18 @@ const parseSupportedJsonSchema = (raw: string): boolean | Record<string, unknown
...
+const isAbsoluteWorkspacePath = (path: string): boolean =>
+  path.startsWith("/") || path.startsWith("\\") || /^[A-Za-z]:[\\/]/.test(path);
+
 // Form-validation errors. Returns an empty array when submittable.
-export const validateManifestForm = (form: ManifestFormState): string[] => {
+export const validateManifestForm = (
+  form: ManifestFormState,
+  // Names already present as preserved declarations (e.g. mount-based
+  // entries), so a form row cannot silently collide with them.
+  preservedWorkspaceNames: Iterable<string> = [],
```
houko, line 950: This parameter is never supplied in the app: the only production call site is `AgentsPage.tsx:3535`, `validateManifestForm(formState)` — one argument — even though `formExtras` is in scope in the same component. So the preserved-name collision check is dead outside the unit test.

Failure: paste a manifest with `[workspaces]` / `vault = { mount = "/data/vault" }`, switch to Form mode (the mount entry lands in `extras.topLevel.workspaces`), add a folder row also named `vault`, Create. Validation is clean, then `serializeManifestForm` emits `vault = { path = … }` inside `[workspaces]` **and** `[workspaces.vault]` from `deferredSectionExtras`. TOML forbids redefining `vault`, so the spawn fails server-side with an opaque parse error instead of showing the per-row "Folder names must be unique." message this PR added.

The test `flags duplicate folder names, including against a preserved declaration` only passes because it calls the library with the second argument directly. Please wire the call site, e.g. pass the keys of `formExtras.topLevel.workspaces`.

crates/librefang-api/dashboard/src/lib/agentManifest.ts
```diff
@@ -538,6 +563,23 @@ export const serializeManifestForm = (
...
+  // [workspaces] — table header, so it is emitted here, after every
+  // top-level scalar; a header inside the scalar block would scope the
+  // remaining bare keys into the table and silently delete them.
+
+  if (form.workspaces.length) {
+    const wsBody: string[] = [];
+    for (const ws of form.workspaces) {
+      const n = ws.name.trim();
+      const p = ws.path.trim();
+      if (!n || !p) continue;
```
houko, line 575: A half-filled row is dropped silently and nothing tells the operator.

`validateManifestForm` checks only duplicate names and path shape, and both checks guard on a non-empty string, so a row with a name and an empty path (or a path and an empty name) produces no error; this `continue` then discards it, no `[workspaces]` header is emitted, and the spawn succeeds with a "Agent created" toast. The operator believes the agent has a shared folder it does not have.

A wholly blank row (freshly added, untouched) is fine to drop — a half-filled one should be an error. Suggest flagging `workspaces.<uid>.name` when the path is set but the name is blank and `workspaces.<uid>.path` when the name is set but the path is blank; the test `emits no [workspaces] header for an empty or blank-row list` currently codifies the drop without drawing that distinction.

crates/librefang-api/dashboard/src/components/AgentManifestForm.tsx
```diff
@@ -1249,6 +1249,77 @@ export function AgentManifestForm({
           />
         </div>
       </CollapsibleSection>
+
+      <CollapsibleSection title={t("agents.form.shared_folders")} defaultOpen={false}>
```
houko: This section owns validated fields but does not pass `invalid`, unlike the scheduling section a few hundred lines up (`invalid={invalidFields.has("schedule.cron") || …}`). `CollapsibleSection` renders `open={defaultOpen || invalid}` and reddens the title on `invalid`, which is what makes a validation failure visible in a `defaultOpen={false}` section.

Failure: operator adds a row with path `../escape`, collapses Shared Folders, clicks Create. `AgentsPage.tsx:3536-3537` stores the error set and returns without mutating, and the only rendering of `duplicate_folder_name` / `folder_path_invalid` is inside the collapsed `<details>` — the Create button looks dead with no explanation.

```suggestion
      <CollapsibleSection
        title={t("agents.form.shared_folders")}
        defaultOpen={false}
        invalid={value.workspaces.some(
          (ws) =>
            invalidFields.has(`workspaces.${ws._uid}.name`) ||
            invalidFields.has(`workspaces.${ws._uid}.path`),
        )}
      >
```

crates/librefang-api/dashboard/src/lib/agentManifest.ts
```diff
@@ -515,17 +524,33 @@ export const serializeManifestForm = (
...
+  // Preserved `[workspaces]` entries (mount-based declarations, or malformed
+  // non-path rows) re-emit as `[workspaces.<name>]` sub-tables beside the
+  // form's rows — routed here rather than through the trailer, which would
+  // emit a duplicate `[workspaces]` header.
+  if (isTomlTable(topTableExtras.workspaces)) {
+    for (const [name, decl] of Object.entries(topTableExtras.workspaces)) {
+      // Non-table garbage under a preserved entry is skipped rather than
+      // emitted as a header it cannot legally have.
+      if (!isTomlTable(decl)) continue;
+      deferredSectionExtras[`workspaces.${name}`] = decl;
```
houko, line 546: A dot in a preserved workspace name is truncated on the way back out. Workspace names are arbitrary user strings, unlike the fixed section names (`model`, `resources`, …) this dotted-key convention was built for.

Given `[workspaces]` with `"notes.v2" = { mount = "/data/notes" }`, this line stores `deferredSectionExtras["workspaces.notes.v2"]`, and the emit loop below does `const [section, subKey] = dottedKey.split(".", 2)` → `section = "workspaces"`, `subKey = "notes"`. The re-emitted header is `[workspaces.notes]`: the agent loses `notes.v2` and gains `notes`. If another preserved entry is literally named `notes`, one of the two is silently overwritten in `nestedDeferred`.

Splitting on the first dot only (`const i = dottedKey.indexOf(".")`, then `slice`) preserves the remainder.

crates/librefang-api/dashboard/src/lib/agentManifest.ts
```diff
@@ -881,6 +933,23 @@ export const validateManifestForm = (form: ManifestFormState): string[] => {
       errors.push("response_format.schema");
     }
   }
+  // Folder rows: duplicate names produce a duplicate TOML key (hard parse
+  // failure on the daemon), and `path` mirrors the kernel's rule — relative
+  // to workspaces_dir, no `..`. A mount row carries an absolute host path
+  // and is not authored here, so only rows are checked.
+  const seenWorkspaceNames = new Set<string>(preservedWorkspaceNames);
```
houko, line 981: The path gets a shape check but the name gets none, and the name is what the agent has to type.

`expand_workspace_alias` (`crates/librefang-runtime/src/tool_runner/fs.rs:183-186`) resolves `@name/rest` with `body.split_once('/')` and matches only the segment before the first slash against the declared names. So a row named `shared/library` serializes to a manifest the kernel happily resolves and injects into TOOLS.md, but every `@shared/library/...` the agent writes resolves the alias as `shared`, which is not declared, and fails with "unknown named workspace '@shared'". Whitespace or a leading `@` in the name has the same effect.

Worth restricting names to the alias-safe set here (and reusing `duplicate_folder_name`'s neighbour message), so the form does not accept a folder the agent cannot address.

crates/librefang-api/dashboard/src/lib/agentManifest.ts
DaBlitzStein: Fixed by matching the kernel's full WorkspaceMode alias set instead of only the literal "r".
READONLY_MODE_ALIASES now checks for r, read, read-only, and readonly, mirroring the TUI's canonical_workspace_mode (crates/librefang-cli/src/tui/event.rs, #7835).
crates/librefang-api/dashboard/src/lib/agentManifest.ts:1235 (parseManifestToml), constant defined at :925.
Regression test: "parses %j as read-only, matching the kernel's WorkspaceMode aliases" (parameterized over r/read/read-only/readonly) and "parses an unrecognized mode spelling as read-write rather than silently upgrading", both in crates/librefang-api/dashboard/src/lib/agentManifest.test.ts.
Confirmed red against the old `=== "r"` check: 3 of the 4 parameterized cases failed with `expected 'rw' to be 'r'` before the fix.
Commit f6b9e727e on work/8013-sweep.

crates/librefang-api/dashboard/src/lib/agentManifest.ts
DaBlitzStein: Fixed by wiring the call site and extracting the lookup into a named, independently-tested helper.
Added preservedWorkspaceNamesFromExtras in crates/librefang-api/dashboard/src/lib/agentManifest.ts:940, and AgentsPage.tsx:3543 now calls validateManifestForm(formState, preservedWorkspaceNamesFromExtras(formExtras)) instead of the single-argument call.
The helper is exported specifically so the call-site logic itself is covered by a test, not only validateManifestForm's own collision check.
Regression test: "preservedWorkspaceNamesFromExtras extracts the preserved-name collision list AgentsPage wires into validateManifestForm" in agentManifest.test.ts.
Confirmed red by stubbing the helper to always return [] — the test failed with `expected [] to deeply equal [ 'vault' ]`.
Commit f6b9e727e on work/8013-sweep.

crates/librefang-api/dashboard/src/lib/agentManifest.ts
DaBlitzStein: Fixed in validateManifestForm: a row with only one of name/path set is now flagged on whichever field is blank, instead of being silently dropped the same way a wholly-blank row is.
crates/librefang-api/dashboard/src/lib/agentManifest.ts:985 onward.
Regression tests in agentManifest.test.ts: "flags a half-filled row (name without a path) instead of dropping it silently", "flags a half-filled row (path without a name) instead of dropping it silently", and "does not flag a wholly blank row" (kept green to pin the intended drop case).
Confirmed red against the old code: both half-filled tests failed with `expected [] to include 'workspaces.w1.path'` / `'workspaces.w1.name'`.
Commit f6b9e727e on work/8013-sweep.

crates/librefang-api/dashboard/src/components/AgentManifestForm.tsx
DaBlitzStein: Fixed by passing `invalid` to the Shared Folders CollapsibleSection, matching the scheduling section's pattern.
crates/librefang-api/dashboard/src/components/AgentManifestForm.tsx:1256, using the exact suggested diff (checks invalidFields for each row's `.name`/`.path` keys).
Regression test: "opens the (defaultOpen=false) Shared Folders section and reddens its title on a validation error (#8013)" in AgentManifestForm.test.tsx.
Confirmed red against the old JSX (no invalid prop): the test failed with `expect(element).toHaveAttribute("open") // received null` — the section stayed collapsed with the error hidden.
Commit f6b9e727e on work/8013-sweep.

crates/librefang-api/dashboard/src/lib/agentManifest.ts
DaBlitzStein: Fixed by splitting on the first dot only instead of `String.split(".", 2)`.
crates/librefang-api/dashboard/src/lib/agentManifest.ts:701, using indexOf + slice so the remainder after the first dot is preserved verbatim as the sub-key.
Regression test: "preserves the remainder of a preserved workspace name containing a dot" in agentManifest.test.ts, using a `"notes.v2"` mount declaration alongside a sibling `notes` entry.
Confirmed red against the old `split(".", 2)` code: the reserialized TOML contained only `[workspaces.notes]` with the "/data/other" mount, silently dropping the "notes.v2" declaration entirely, exactly the collision described in the report.
Commit f6b9e727e on work/8013-sweep.

crates/librefang-api/dashboard/src/lib/agentManifest.ts
DaBlitzStein: Fixed by restricting workspace names to TOML bare-key characters in validateManifestForm.
WORKSPACE_ALIAS_SAFE_NAME = /^[A-Za-z0-9_-]+$/ in crates/librefang-api/dashboard/src/lib/agentManifest.ts:932, checked alongside the existing duplicate-name check at :990.
This rejects '/' directly (the concrete `expand_workspace_alias` split-point bug) and also whitespace/'@', which would let the declared name and the string the agent has to type diverge.
Regression test: "rejects a workspace name outside the alias-safe character set" in agentManifest.test.ts, parameterized over "shared/library", "shared library", and "@shared".
Confirmed red against the old code (duplicate-only check): all three cases failed with `expected [] to contain 'workspaces.w1.name'`.
Commit f6b9e727e on work/8013-sweep.

DaBlitzStein:
Pushed since my last note on this PR, now at `f6b9e727e`:

- `cc9e3fb55` Merge branch 'main' into feat/dashboard-shared-folders
- `8419b367a` Merge branch 'main' into feat/dashboard-shared-folders
- `f6b9e727e` fix(dashboard): close shared-folder read-only privilege escalation and 5 related form gaps

Every one of these was verified before it was pushed rather than after: for each behaviour change the production block was reverted, the test was run against it, the literal failure was recorded, and the block was restored — a regression test that stays green with the fix removed guards nothing.

Flagging it here because a push does not announce itself, and several of these landed hours after the review threads were answered.

houko (approved):
LGTM. All four findings are fixed, each at the level the finding was about rather than at the symptom.

- **The dead `preservedWorkspaceNames` parameter** is now actually supplied: `AgentsPage.tsx:3541` calls `validateManifestForm(formState, preservedWorkspaceNamesFromExtras(formExtras))`, and the extraction into a named helper is justified in-comment precisely so the app's own path is under test rather than only the validator's unit tests. That is the right response to "this check is dead outside the test".
- **Half-filled rows** now produce a per-side error (`workspaces.<uid>.name` / `.path`) while a wholly blank untouched row still passes, and the comment states the distinction and why it matters — the serializer drops both identically, so without the check the agent is created believing it has a shared folder it does not have.
- **The dotted-name truncation** is fixed with `indexOf(".")` + `slice`, and the comment records the exact trap (`String.split(".", 2)` truncates rather than keeping the remainder, so `workspaces.notes.v2` lost its last segment and overwrote a sibling). That is the kind of comment that stops the "simplification" back to `split` later.
- **The name shape check** is `WORKSPACE_ALIAS_SAFE_NAME = /^[A-Za-z0-9_-]+$/`, with the reason cited from the resolver itself — `expand_workspace_alias` matches only the segment before the first `/`, so `shared/library` serializes to a manifest the kernel accepts and the agent can never address.

The asymmetry between those last two is deliberate and correct, worth stating so it does not read as a contradiction: a *preserved* name from pasted TOML may contain dots and must round-trip intact, while a *newly authored* name is held to the stricter set. Restricting what the form can create while preserving what already exists is the right way round.

CI green, `mergeStateStatus: CLEAN`.

Merge-order note: this shares `agentManifest.ts`, `AgentManifestForm.tsx` and `AgentsPage.tsx` with #7749 and #8028, both of which restructure the same files and are part of the cluster I described on #7835. This is the smallest member of that group and is green and complete, so merging it first shrinks the problem rather than adding to it — but #7749 and #8028 will need a rebase afterwards, not a conflict resolution that picks a side.

PR #8027, opened 2026-08-30
feat(dashboard): add promotion preview modal to agent-types page

DaBlitzStein:
## Summary

- The `GET /api/templates/{name}` response already carries a `promotion_preview` field (`requires_review`, `findings`, `manifest_toml`) built by `promotion_preview()` in `crates/librefang-api/src/routes/agent_templates.rs`, but the dashboard had no UI reading it.
- Adds `PromotionFinding` and `PromotionPreview` TypeScript interfaces to `src/api.ts`, matching `librefang_types::manifest_privacy::Finding` field-for-field (`field`, `category`, `preview`, `removed_by_sanitizer`), and adds `promotion_preview?: PromotionPreview` to `AgentTypeDetail`.
- Adds a "Promote" action (shield icon) to every row on the Agent Types page. It opens a `PromotionPreviewModal` that fetches the existing `useAgentType(name)` detail query (no new endpoint or hook needed — the field was already in the response) and shows:
  - A warning banner when `requires_review` is true.
  - Each finding with its dotted field path, preview text, and a "Stripped automatically" / "Needs review" badge (`removed_by_sanitizer`).
  - The sanitized manifest TOML in a scrollable block with a "Copy TOML" button, using the existing `copyToClipboard` helper.
- Adds the `agentTypes.promote*` i18n keys to all five locale files (`en`, `ko`, `pl`, `uk`, `zh`) — the locale-parity test requires full coverage, not just the three originally scoped.

## Verification

- `pnpm run typecheck` — clean.
- `npx eslint src/pages/AgentTypesPage.tsx src/api.ts` — clean.
- `pnpm run test -- locale-parity` — passes (no missing/extra keys across all 5 locales).
- `pnpm run test` (full dashboard suite) — 1670/1671 passing. The one failure (`PromptsExperimentsModal.test.tsx > stops selection when every variant has a traffic bucket`, a 5s timeout) reproduces identically on a clean `origin/main` checkout with no changes from this branch applied, so it is a pre-existing flake unrelated to this change.

## Out of scope

- The pre-existing `PromptsExperimentsModal.test.tsx` timeout above is unrelated to the agent-types domain this PR touches; flagging it here rather than silently working around it.


houko (commented):
This one is well-formed: `promotion_preview` already exists server-side (`routes/agent_templates.rs::agent_type_detail` emits it via `promotion_preview()`), the `PromotionFinding` / `PromotionPreview` interfaces match the Rust shape, `copyToClipboard` is the existing helper and its boolean return is handled both ways, `manifest_toml: string | null` is handled at both the `<pre>` and the disabled copy button, and all five locales get the same twelve keys so `locale-parity` stays green.
No new API surface, so no OpenAPI/SDK obligations.

Three things to settle:

### 1. The promote button renders on workspace-sourced rows too

It is inserted immediately above the comment that exists specifically to stop this:

```
{/* A workspace-sourced row is a live agent's own manifest. The write verbs
    refuse it by design, so rendering Edit/Delete here would offer a control
    that cannot succeed — point at the surface that can instead (#7731). */}
```

`GET /api/templates/{name}` does compute `promotion_preview` for a live agent's manifest, so the modal will populate and the button *works* — this is not a broken control. But "contribute this live agent's own manifest to a shared registry" is a different proposition from promoting an agent type, and a live agent's manifest is the one most likely to contain workspace paths and channel ids.
Please make that a deliberate decision rather than a side effect of button placement: either gate the button on `type.source !== "workspace"` like Edit/Delete, or say in the modal that this is a live agent's manifest.

### 2. The modal is named "promotion" but cannot promote

Everything here is read-only plus a clipboard copy — the operator still has to paste the TOML somewhere by hand. That is a reasonable first step, but `promote` / `promote_title` will read as "this publishes it" to anyone who has not read the code.
Since #8043 is open and appears to be the actual promote path, consider naming this one for what it does (`promotion_preview` / "Review before contributing") so the two do not collide in the UI vocabulary, and so the follow-up can take the `promote` label.

### 3. No changelog fragment

`changelog.d/added/8027-....md` ending `(#8027) (@DaBlitzStein)`.

## Coordination note

`AgentTypesPage.tsx` and the `AgentTypeDetail` interface in `api.ts` are being edited concurrently by #8027, #8028, #8042, #8043, #8052, #8053 and #8054.
Every one of those adds a field to `AgentTypeDetail` and a control to the same row or editor, so they will conflict pairwise, and per `CLAUDE.md` the `main` ruleset has no strict-status-check policy — each merges against CI run on its own base.
Worth agreeing an order (and re-running CI on the later ones) rather than merging them as they go green.

Minor: no `AgentTypesPage.test.tsx` exists yet, so I am not asking for a test here — but this modal, with its three states (loading / no preview / findings), is a reasonable excuse to start one.

PR #8030, opened 2026-08-30
feat(dashboard): show goal run phase, iterations, and errors inline

### LINKED ISSUE #8017: feat(dashboard): show goal states like workflow (pending/active/paused/done)
Goals currently display as a flat list without visual state indicators.
Workflows show clear state badges (pending, running, completed, failed).
Goals should have similar visual state indicators in the dashboard (and ideally the TUI and API responses) so operators can tell at a glance which goals are active, paused, or completed.

DaBlitzStein:
## Summary

- Add `GoalRunInfo` component that displays the autonomous run's current phase, iteration counter, and last error directly on the goal row — matching the workflow run timeline pattern.
- Add `goalRunPhaseBadge` helper mapping each phase to a color-coded badge (running=brand, finished=success, stopped=warning, rate_limited=error, max_iterations_reached=warning).
- Running phase gets a pulsing dot indicator.
- Add `run_phase_*` i18n keys to all 5 locale files (en, uk, zh, ko, pl).

Closes #8017.

## Verification

- `pnpm run typecheck` — passes
- `pnpm run test` — all goals and locale tests pass; only pre-existing `PromptsExperimentsModal` timeout failure remains


houko (commented):
No blockers — the phase strings match `GoalRunPhase`'s `Display` impl exactly (`goal.rs:203-210`), all five locales carry the five new keys so `locale-parity` stays green, and the extra `useGoalRun(goal.id)` call costs nothing because `GoalRunControl` already mounts the same query key on the same row, so TanStack dedupes it.

Four things worth folding in:

### The `finished` badge is unreachable

`GoalRunInfo` is gated on `status !== "completed"`, and `GoalRunPhase::Finished` is documented as "the goal reached `Completed`/`Cancelled` (or 100% progress)" — so by the time the phase is `finished`, the goal's status is `completed` and the row no longer renders the badge.
`goals.run_phase_finished` is dead in all five locales as written.

That gate is worth reconsidering rather than patching: the run outcome is arguably *most* interesting on a completed goal (did it finish on its own, or did it stop at the iteration cap?).
`stopped`, `rate_limited` and `max_iterations_reached` do leave the goal in `in_progress`, so those three do show.

### The iteration count is now displayed twice

`goals.run_active` ("Running · iteration {{iteration}}/{{max}}") is already the Stop button's `title`, built from the same `run.iteration` / `run.max_iterations`.
Now the badge row shows `{iteration}/{max}` as well. Pick one home for that fact — the visible badge is the better one, so consider dropping the numbers from the tooltip and leaving it as a plain "Stop".

### No changelog fragment

`changelog.d/added/8030-....md`, body without the leading `- `, ending `(#8030) (@DaBlitzStein)`.

### No test, and there is a file waiting for one

`src/pages/GoalsPage.test.tsx` already mocks the goal queries and mutations. A case that feeds a `rate_limited` run and asserts the badge label and the `last_error` text would pin both the phase mapping and the `defaultValue` fallback.

Minor: `run.phase` is typed as a literal union in `api.ts`, so `goalRunPhaseBadge(phase?: string)` throws that away. Typing the parameter as `GoalRunState["phase"]` and dropping the `default` arm would make the compiler tell you about the next phase someone adds, instead of it silently landing in the grey fallback.

PR #8031, opened 2026-08-30
feat(dashboard): improve backups section UX

### LINKED ISSUE #8022: feat(dashboard): implement Backups tab (currently empty)
The dashboard has a Backups tab/section that renders an empty state with no content or functionality.
The Backups tab should show:
1. List of available backups (from the backup directory, e.g. ~/.librefang/backups/)
2. Backup metadata (date, size, what was backed up)
3. Actions: create backup, restore from backup, delete old backups
4. Backup schedule/policy configuration
The backup infrastructure exists on the backend (backups are created during certain operations) but there is no dashboard surface for managing them.

DaBlitzStein:
## Summary

- Replace the inline backup-create button with a confirmation dialog so the operator sees what will be snapshotted before committing.
- Improve the empty-state message from "No backups" to guidance text that explains what a backup contains and encourages first-time creation.
- Show the list of components (config, agents, skills, workflows, data) under each backup entry.
- All 5 locale files updated (en, ko, pl, uk, zh).

## Verification

- `pnpm run typecheck` — clean
- Dashboard-only change, no Rust modifications

Closes #8022


houko (changes requested):
The `components` line on each backup row and the fuller `no_backups` copy are both good, and the test was updated properly rather than deleted.

One real regression in the confirm-dialog change, and one design question.

## The failure path now hides its own error message

`useCreateBackup` (`src/lib/mutations/runtime.ts:52`) declares no `onError`, and the only place a failed backup is reported is `RuntimePage.tsx:816`:

```tsx
{backupMutation.isError && <p className="text-xs text-error mb-2">{t("runtime.backup_error")}</p>}
```

That paragraph lives inside the backups card. `ConfirmDialog.requestConfirm` catches a rejected `onConfirm`, clears its busy flag and **leaves the dialog open** — so on failure the operator is left looking at a modal that appears to have done nothing, while the only error text renders behind the overlay.

Before this PR the same failure at least left the error visible. Please add the toast:

```tsx
onConfirm={async () => {
  try {
    await backupMutation.mutateAsync(undefined, {
      onSuccess: () => addToast(t("runtime.backup_created"), "success"),
    });
  } catch (err) {
    addToast(toastErr(err, t("runtime.backup_error")), "error");
  }
}}
```

Note that swallowing it there also closes the dialog, which is what you want — a re-thrown error keeps it open with no explanation.
Worth a test case too: the file already mocks `mutateAsync`, so a rejecting mock plus an assertion on the error toast is a three-line addition next to the one you just rewrote.

## Is a confirmation the right control for a non-destructive action?

Creating a backup is additive and reversible; the dialog's message is not a decision ("may take a moment on a large workspace") but a duration warning.
`tone` is left at `"default"`, so `Enter` confirms — which is the right call for a safe action, but it also means the dialog is mostly a keystroke tax.
The button already has `isLoading={backupMutation.isPending}`, so the "this takes a while" signal exists without a modal.

I'd rather see the duration hint as helper text under the button (or in the success/pending state) than as a gate. If you keep the dialog, the `isLoading` on the underlying button is now redundant with the dialog's own busy state — it spins behind the overlay.

## Also

- **No changelog fragment.** `changelog.d/changed/8031-....md` ending `(#8031) (@DaBlitzStein)`.
- `BackupConfirmState` now carries `filename`/`keepConfig`/`components` that the `"create"` case fills with `""` / `false` / `[]` purely to satisfy the type. A discriminated union (`{ type: "create" } | { type: "restore" | "delete"; filename: string; ... }`) would stop the next reader wondering whether create honours those fields.

houko:
合并中，按维护者指示。上面提的错误可见性问题（失败时 `runtime.backup_error` 渲染在 modal 遮罩后面，操作者只看到一个不动的弹窗）没有修，作为已知项留档 —— 修法是 `onConfirm` 里 catch 后 `addToast(toastErr(err, t("runtime.backup_error")), "error")`，六行。

houko (approved):
按维护者明确授权解除阻塞（会话内已授权）。上一条 review 的技术分析仍然成立，未修项已在 PR 评论里留档：失败路径下 `runtime.backup_error` 渲染在 modal 遮罩后面，操作者只看到一个不动的弹窗；changelog fragment 仍然缺。

PR #8028, opened 2026-08-30
feat(dashboard): full manifest parity in the agent-types editor

DaBlitzStein:
## Summary

- Replace the 6-field agent-type form with the shared `AgentManifestForm` component so the editor exposes every field an `agent.toml` supports.
- Add `PUT /api/templates/{name}/toml` endpoint that accepts raw TOML, parses, validates, and persists the full manifest.
- Add `useUpdateAgentTypeToml` mutation hook and `putAgentTemplateToml` client function.

## Test plan

- [ ] Create a new agent type, verify all manifest fields are editable
- [ ] Edit an existing agent type via the full form, verify changes persist
- [ ] Verify the TOML endpoint rejects invalid TOML with a 400
- [ ] Verify the endpoint returns 404 for a nonexistent template
- [ ] Verify the endpoint returns 409 for a name that belongs to a live agent
- [ ] `cargo check --workspace --lib`
- [ ] `cargo clippy --workspace --all-targets -- -D warnings`

## Issue references

- Refs #7742: Agent-type editor coverage of the manifest parity request.

PR #8032, opened 2026-08-30
feat(agents): track source template origin in manifest and dashboard

### LINKED ISSUE #8018: feat: show custom agent type in agent description (dashboard/TUI/API)
When an agent is spawned from a custom agent type, the origin type name is not visible anywhere in the agent's detail view (dashboard, TUI, or API response).
The agent description or metadata section should show which agent type it was created from, so operators can trace provenance.
This information is available in the manifest but not surfaced in the UI.

DaBlitzStein:
## Summary

Closes #8018.

- `AgentManifest` gains `source_template: Option<String>` (`#[serde(default, skip_serializing_if = "Option::is_none")]`, so existing manifests deserialize unchanged).
- `POST /api/agents` records the template name on the manifest whenever the spawn request resolves through `template` (not when `manifest_toml` is supplied directly).
- `GET /api/agents/{id}` and `GET /api/agents` (list) both surface `source_template` in their JSON response.
- `sanitize_for_publication()` strips `source_template` — it names a template in *this* deployment's workspace, which means nothing in another one.
- Dashboard: `AgentItem` / `AgentDetail` types gain `source_template`; the compact agent-list row shows "From {{name}}" next to the schedule when it is set.
- i18n: added `agents.origin_template` ("From {{name}}") and `agents.custom_agent` ("Custom") to all five locale files. Used a distinct key name rather than the existing `agents.from_template`, which already labels the "create agent from template" tab with a different (non-interpolated) meaning — reusing it would have broken that UI.
- Every exhaustive `AgentManifest` struct literal/destructure in the codebase (`agent_type.rs`, `manifest_privacy.rs`, `wizard.rs`, `config/mod.rs`'s per-agent-override-keys test) was updated for the new field; sites already using `..Default::default()` needed no change.
- New integration tests in `agents_routes_integration.rs`: spawning via `template` records `source_template` on both the detail and list endpoints; spawning via inline `manifest_toml` leaves it `null`.

## Known gap (not fixed in this PR)

`librefang-cli`'s `spawn_template_agent` resolves the template locally and posts the fully-rendered `manifest_toml` to `/api/agents` — it never sends `template: <name>`, so CLI-spawned agents will not get `source_template` populated by this change. Fixing that touches `librefang-cli` rather than the api/types/dashboard trio this PR is scoped to; flagging it here rather than leaving it unmentioned.

## Test plan

- `cargo check -p librefang-types -p librefang-api --lib` and `--all-targets` (plus `-p librefang-kernel -p librefang-runtime -p librefang-hands -p librefang-cli --all-targets` to catch every exhaustive-literal site) — clean.
- `cargo clippy -p librefang-types -p librefang-api --all-targets -- -D warnings` — clean.
- `cargo test -p librefang-api --test agents_routes_integration` — 67 passed, including the two new tests.
- `cargo test -p librefang-types --lib` — clean, including `per_agent_override_keys_cover_manifest_overrides_6`.
- Dashboard: `pnpm typecheck`, `pnpm lint`, `pnpm exec vitest run src/lib/__tests__/locale-parity.test.ts` (8 passed), `pnpm exec vitest run src/pages/AgentsPage.test.tsx` (6 passed) — all clean.


houko (changes requested):
The mechanical part of adding a 59th manifest field is done properly — every exhaustive struct literal updated, `sanitize_for_publication` destructures `source_template: _` and emits `None` so a promoted manifest does not leak this deployment's template names, and the 30-odd test fixtures moved with it. That is the tedious half and it is right.

Three things to fix.

## 1. The parity guard is silently exempted from covering the new field

`crates/librefang-types/tests/manifest_field_parity.rs` documents its fixture as:

"Every field of `AgentManifest` set to something a `Default::default()` manifest does not have. Written as one exhaustive struct literal with no `..Default::default()` at the top level, so the compiler — not a reviewer — is what notices a 59th field being added."

The compiler did its job here and demanded the field. But the value added is `source_template: None`, which **is** the default — so `every_manifest_field_survives_a_toml_round_trip` never round-trips this field, and the guard's stated invariant no longer holds for it.

That matters concretely because the field carries `#[serde(default, skip_serializing_if = "Option::is_none")]`: the round trip is exactly where a `skip_serializing_if` interaction would show up, and this fixture is the only place that would catch it. Please set it to a real value, e.g. `source_template: Some("customer-support".into())`.

## 2. `CHANGELOG.md` is edited directly instead of a `changelog.d/` fragment

`CLAUDE.md` has a section on this whose heading is literally "Changelog: a new file under `changelog.d/`, NOT an edit to `CHANGELOG.md`", with the reason spelled out: appending to the single `## [Unreleased]` section conflicts with every other open PR doing the same, in a file where the conflict carries no information.

There are currently 18 other open PRs. This needs to be `changelog.d/added/8032-source-template-provenance.md`, body without the leading `- `.

While moving it: the bullet ends `(#8018) (@DaBlitzStein)`, and #8018 is the *issue*, not this PR. `xtask/src/changelog.rs:739` reads a bullet's reference from the last `(#N)` group on its last line and uses that set to suppress generated lines, so as written #8032 keeps an auto-generated `- feat(agents): track source template origin in manifest and dashboard (#8032) ...` line alongside your prose. Cite both if you want the issue visible: `(#8032)` last.

## 3. Only one of the four template-instantiation paths sets the field

`used_template` is threaded through `routes/agents/lifecycle.rs` only. `load_agent_template` has four other production callers, and two of them create registered agents:

- `crates/librefang-kernel/src/kernel/step_agent.rs:138` and `:187` — step agents instantiated from a template
- `crates/librefang-kernel/src/kernel/assistant_routing.rs:150`
- `crates/librefang-kernel/src/kernel/ephemeral_spawn.rs:202` (ephemeral, so arguably out of scope — no agent is registered)

So a step agent spawned from `researcher` gets `source_template: None`, and the dashboard column next to the schedule renders blank — which reads as "not from a template" rather than "we did not record it". The changelog's "Agents spawned from a template now carry the template name forward" is true only of `POST /api/agents`.

Either thread it through `step_agent.rs` and `assistant_routing.rs` too, or narrow the claim in the changelog and add a comment at the `load_agent_template` call sites saying provenance is deliberately not recorded there. The first is better — the whole value of the field is that a blank means something definite.

## Minor

- `if used_template.is_some() { manifest.source_template = used_template; }` — the guard is redundant against `None`, but I assume it is there so a `manifest_toml` body that already carried a `source_template` is not clobbered when no template was used. If so, that is worth one line of comment; it is not obvious from the shape.
- `scan_for_publication` reports "every instance-specific field that is actually set, with a bounded preview of what publication would drop". `source_template` is now stripped by the sanitizer but, as far as I can see, is not reported as a finding — so the promotion preview (#8027) will not tell the operator that the template name is being dropped. Worth adding for symmetry with the other instance-specific fields, especially since #8043 wants to publish these manifests.
- `AgentsPage.tsx`, `api.ts` and the five locale JSONs are contested by #8027, #8028, #8042, #8043, #8047, #8048, #8052, #8053 and #8054. #8053 in particular ("display agent origin and lineage") reads like the presentation half of this PR and touches the same `AgentsPage.tsx` region; those two should be merged adjacently, in that order, with CI re-run on the second.

houko (approved):
按维护者明确授权解除阻塞。上一条 review 的技术结论（manifest_field_parity fixture 用 None 把新字段排除在 round-trip 守卫外；直接改 CHANGELOG.md 而非 changelog.d 片段；只有 lifecycle.rs 设置该字段，step_agent.rs 与 assistant_routing.rs 的模板实例化路径仍为 None）未被修复，保留在记录里。

PR #8035, opened 2026-08-30
fix(deps): move chacha20 off the yanked 0.10.0

houko:
## What

`cargo update -p chacha20` — a lockfile-only bump from the yanked 0.10.0 to 0.10.2.
Three lines of `Cargo.lock`, plus a changelog fragment.

## Why

`cargo-deny (advisories)` has been failing on `main` since 2026-08-28 (last green run 2026-08-27):

```
error[yanked]: detected yanked crate (try `cargo update -p chacha20`)
```

`chacha20` is transitive — `rand 0.10.2` and the `cipher`-family crates pull it in — so nothing in the workspace declares it directly and no manifest changes.

The reason this is worth its own PR rather than riding along with something else: a supply-chain gate that is red on every branch is a gate nobody reads. A real advisory landing behind this one would have arrived looking exactly like the standing noise.

## Verification

- `cargo deny check advisories` → `advisories ok`
- `cargo check --workspace --lib` → `Finished dev profile ... in 2m 07s`, clean. Run in a fresh worktree with an empty `target/`, so the `Checking librefang-kernel` / `librefang-runtime` / `librefang-acp` lines and the 2.6G of artifacts it produced confirm the compiler actually saw this tree rather than replaying a warm cache.
- `cargo update -p chacha20` reported `Updating chacha20 v0.10.0 -> v0.10.2`, `Locking 1 package`, 256 unchanged dependencies
- Clippy not run locally: there are no source changes for a lint to fire on, and the CI `Quality` job covers it.
- No new tests. There is no code path to assert on — the assertion for this change is the `cargo-deny` job itself going green on this branch, and it is the check to look at when reviewing.

## Deferred

None.

Unrelated but noticed while pushing: GitHub reports 19 Dependabot alerts on the default branch (5 high, 11 moderate, 3 low). Distinct from this — a yanked crate is not a vulnerability — and untouched here.


houko (commented):
The advisory fix itself is right, and I verified the supply-chain claim: `chacha20` is reached through `rand 0.10.2` and `ssh-cipher` in the lock, so it is genuinely transitive and a lockfile bump is the correct remedy.
The fragment is also unusually good about *why* a permanently red gate is the real damage rather than just restating the error.

One thing to explain or drop before merge:

**The diff carries a second, unrelated change: `tempfile` 3.27.0's dependency moves from `getrandom 0.4.2` to `getrandom 0.3.4`.**

The fragment says "this is a lockfile bump to 0.10.2 and nothing else", and that line contradicts it.
`tempfile` 3.27.0 declares `getrandom = ">=0.3.0, <0.5"`, so both versions satisfy the requirement and nothing breaks — but `cargo update -p chacha20` does not produce that edge, so something else regenerated part of the lock (a different toolchain's resolver, or a wider `cargo update` that was mostly reverted).
`getrandom 0.4.2` is still pulled in by five other crates, so the net effect is that the graph now needs both 0.3.4 and 0.4.2 where this edge previously only needed 0.4.2.

Please either drop that hunk and re-run the narrow update, or keep it and say in the fragment that it happened and why.
A supply-chain PR is the one place where an unexplained dependency line is worth thirty seconds of prose.

PR #8039, opened 2026-08-30
test(kernel): cover empty-string UUID sanitization in load_goal

DaBlitzStein:
## Summary

- #8010 fixed `load_goal` in `crates/librefang-kernel/src/goal_runner.rs` silently dropping a stored goal when `agent_id` / `parent_id` were persisted as `""` instead of `null`.
- An empty string fails UUID parsing inside `serde_json::from_value`, and the caller's `.ok()` swallowed that error, so the goal disappeared entirely and every operation touching it (start / pause / resume) returned a bare 500.
- That fix landed without a regression test. This PR adds one: `goal_runner::tests::load_goal_sanitizes_empty_string_uuid_fields` seeds a goal with `agent_id: ""` and `parent_id: ""` directly in the substrate and asserts `load_goal` still returns the goal, with both fields deserialized as `None`.

## Verification

- `cargo check --workspace --lib --exclude librefang-desktop` — pass (librefang-desktop excluded: missing system `glib-2.0` dev package on this host, unrelated to this change)
- `cargo clippy -p librefang-kernel -p librefang-types -- -D warnings` — clean
- `cargo test -p librefang-kernel --lib load_goal_sanitizes_empty_string_uuid_fields` — 1 passed

## Deferred work

None — this PR only adds test coverage; the fix itself is already merged in #8010.

### DIFF
--- a/crates/librefang-kernel/src/goal_runner.rs
+++ b/crates/librefang-kernel/src/goal_runner.rs
@@ -922,6 +922,37 @@ mod tests {
         }
     }
 
+    /// Goals written before UUID `Option` fields were reliably serialised as
+    /// `null` may store `agent_id` / `parent_id` as `""` instead. An empty
+    /// string fails UUID parsing inside `serde_json::from_value`, and the
+    /// caller's `.ok()` used to swallow that error and drop the whole goal —
+    /// turning every `start` / `pause` / `resume` on it into a bare 500.
+    #[test]
+    fn load_goal_sanitizes_empty_string_uuid_fields() {
+        let substrate = MemorySubstrate::open_in_memory(0.01).unwrap();
+        let goal_id = GoalId::new();
+        let now = Utc::now().to_rfc3339();
+        let raw = serde_json::json!([{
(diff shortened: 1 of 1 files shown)


houko (commented):
Good regression guard, and the doc comment correctly describes the failure it locks down — `load_goal`'s `.and_then(|v| serde_json::from_value(v).ok())` at `goal_runner.rs:236` really does drop the whole goal when a UUID field fails to parse, so this is the right shape of test.

One gap worth closing while you are here: **the sanitizer normalises three keys, and the test covers two.**

```rust
for key in ["verify_agent_id", "agent_id", "parent_id"] {
```

`verify_agent_id` is untested, so deleting it from that array — the single most likely way this regresses — still passes.
Please add it to the seeded JSON and assert `loaded.verify_agent_id == None` alongside the other two.

Minor, take it or leave it: the test only proves empty strings become `None`. A second case with a real UUID in `agent_id`, asserting it survives as `Some(..)`, would stop an over-eager "sanitize" from nulling every UUID and still passing.

PR #8040, opened 2026-08-30
feat(media): add CustomImageConfig and CustomVideoConfig

DaBlitzStein:
## Summary

- Add `CustomImageConfig` and `CustomVideoConfig` structs to `MediaConfig`, plus `video_provider` / `video_model` fields, completing the custom-endpoint pattern across all four media understanding domains (STT, TTS, image description, video description).
- Wire custom image provider routing through the existing OpenAI-compatible `openai_describe_image()` dispatch, with `custom_image_config()` and `custom_image_model_ref()` resolution helpers mirroring the STT pattern.
- Wire custom video provider routing with `custom_video_config()` and `custom_video_model_ref()` resolution helpers; the dispatch remains placeholder-only until a real provider protocol is integrated.
- Expose the four new fields in the config API read path (`manage.rs`).
- Regenerate the golden schema fixture to include the new `MediaConfig` fields.
- Add dashboard locale strings (`fld_custom_image`, `fld_custom_video`, `fld_video_provider`, `fld_video_model` and descriptions) across all five locale files (en, zh, uk, ko, pl).

## Verification

- `cargo check --workspace --lib` -- zero errors
- `cargo clippy --workspace --all-targets -- -D warnings` -- zero warnings
- `cargo test -p librefang-types --lib` -- 39 tests passed
- `cargo test -p librefang-api --lib` -- passed (exit 0)
- `cargo test -p librefang-runtime-media --lib` -- passed (exit 0)
- Golden fixture regenerated and included (`config_schema_golden` test will pass)
- Locale parity test (`locale-parity.test.ts`) -- 8 tests passed
- Dashboard build (`pnpm build`) -- succeeded

## Test plan

- [ ] CI passes (`cargo check`, `clippy`, scoped tests, golden fixture comparison)
- [ ] Config API returns `custom_image`, `custom_video`, `video_provider`, `video_model` under `media` section
- [ ] Dashboard Settings page renders the new media config fields with correct labels


houko (changes requested):
The image half of this is clean and I have no correctness notes on it: `custom_image_model_ref` returning `None` for built-in providers so the custom model cannot leak into their defaults is the right shape, the precedence `image_model` → `custom_image.model` → `default_vision_model` is correct, the doc examples are genuinely useful, and the config-schema golden fixture moved with the types. `config_reload.rs` compares `&old.media` as a whole struct, so the new fields are covered by the existing restart classification for free.

The video half is the problem.

## Blocking: the custom video path returns a fabricated success

```rust
_custom => {
    let (_api_url, _api_key) = custom_video_config(_custom, &self.config.custom_video)?;
    // ponytail: placeholder — custom video description dispatch not yet implemented;
    Ok(MediaUnderstanding {
        description: format!("[Video description would be generated by custom provider '{_custom}']"),
        ...
    })
}
```

It validates the config, discards the resolved URL and key, and returns `Ok`. The caller cannot tell this apart from a real description, and `MediaUnderstanding.description` flows into an agent's prompt — so the model receives the literal string `[Video description would be generated by custom provider 'local-video']` as though it were what the video contained, and reasons from it.

I know the Gemini branch already carried this placeholder before your change, so this is not a class of bug you introduced. But this PR is what makes it reachable through configuration: it adds `video_provider` and `[media.custom_video]`, documents them in five languages and in the config schema, and the changelog invites operators to use them ("custom video providers resolve config but remain placeholder-only"). An operator who follows that documentation gets silent garbage rather than an error.

Please return `Err` for both unimplemented branches instead:

```rust
Err(format!("Video description via '{provider}' is not implemented yet").into())
```

`describe_video` already returns `Result`, every caller already handles the error path (the `enabled` check and the missing-key check both use it), and an error is the honest answer to "describe this video" when nothing described it. If a placeholder must stay for the existing Gemini path to avoid breaking a caller, at least keep the new custom path as `Err` — do not widen the fabrication to a second surface in the same commit that documents it.

## Should fix

### `ponytail:` is not a tracked marker

`grep` finds it twice in the whole repo, both in `crates/librefang-channels/src/sidecar.rs`, so it is not an established convention — and `.github/workflows/todo-to-issue.yml` scans for `TODO`, not `ponytail`. Two placeholders that shape a public config surface will never surface as issues.

Use `TODO` so the automation files them, or open the issues by hand and reference them in the comments. If `ponytail:` *is* meant as a deliberate opt-out of TODO-to-Issue, that deserves a line in `CLAUDE.md` rather than being inferable only from two prior uses.

### The changelog fragment credits the wrong PR

It ends `(#8012) (@DaBlitzStein)`. `xtask/src/changelog.rs:739` takes a bullet's reference from the last `(#N)` group on its last non-empty line and uses that set to suppress generated release-note lines — so #8040 keeps an auto-generated duplicate alongside your prose, and whatever #8012 is loses its own line. Also rename the file to `8040-custom-media-providers.md` per `changelog.d/README.md`.

### The four new Polish description strings have their diacritics stripped

```
"desc_video_provider": "Dostawca opisu wideo (domyslnie Gemini jesli dostepny)",
"desc_custom_image": "... (uzywana gdy image_provider nie jest wbudowanym dostawca)",
```

Should be `domyślnie`, `jeśli dostępny`, `używana`, `dostawcą`. Every other Polish string in the same file — including the `fld_*` keys you added in the same diff (`Niestandardowa konfiguracja obrazu`, `Czas schładzania`) — uses proper diacritics, so this is four strings that went through an ASCII-folding step rather than a house style. (The `de`/`es`/`fr` `errors.ftl` files do fold deliberately, but `pl.json` does not.)

### Three copies of the same resolver

`custom_stt_config`, `custom_image_config` and `custom_video_config` are now the same twenty-five lines three times, differing only in the config type and the word STT / Image / Video; `custom_stt_model_ref` / `custom_image_model_ref` / `custom_video_model_ref` are the same shape again. The next fix to key resolution — a trim rule, a second env var name, a keyless-server special case — will land in one copy and not the others.

They are all in one file in one crate, so by `CLAUDE.md`'s own deferral bar this is in scope. A small trait (`fn base_url(&self) -> &str`, `fn api_key_env(&self) -> &str`, `fn key_required(&self) -> bool`) over the three config structs, or a free function taking those three values plus a label, collapses both families.

## Minor

- `_custom` and `_api_url` / `_api_key`: the underscore prefix means "deliberately unused", but `_custom` is used in the arm body (twice), which reads as a mistake. Name it `custom` and drop the URL/key binding entirely (`let _ = custom_video_config(...)?;`) if you keep the placeholder, or use them once the call is real.
- No test. `media_understanding.rs` has a test module; the two `custom_*_config` error branches (empty `base_url`, `key_required` with an unset env var) are pure functions and cheap to pin — and they are the messages an operator will actually hit while setting this up.

houko (approved):
按维护者明确授权解除阻塞。上一条 review 的技术结论（自定义 video provider 返回伪造成功字符串，会当成真实视频描述进 prompt；pl 译文缺变音符号；三份重复的 resolver）未被修复，保留在记录里。

PR #8042, opened 2026-08-30
feat(api,dashboard): restore agent types from registry with diff preview

DaBlitzStein:
## Summary

- Add `GET /api/templates/{name}/registry-diff` endpoint that compares a locally installed agent type against its original in the registry checkout (`~/.librefang/registry/agent-types/{name}/agent.toml`), returning a field-level diff plus raw TOML for both sides.
- Add `POST /api/templates/{name}/restore` endpoint that overwrites the local agent type with the registry original via `persist_agent_type()` (atomic rename).
- Add `RestoreDiffModal` to the Agent Types dashboard page — a RotateCcw button per editable row opens a panel showing changed fields in a table (local in red, registry in green), with a confirm button to restore.
- Add query hook (`useAgentTypeRegistryDiff`) and mutation hook (`useRestoreAgentType`) following the dashboard data layer rules.
- Add i18n keys for en/uk/zh (dashboard JSON) and all 9 server locales (Fluent `.ftl` files) for the `api-error-registry-type-not-found` error.

## Verification

- `cargo check -p librefang-api -p librefang-types --lib` — clean
- `cargo clippy -p librefang-api -p librefang-types -- -D warnings` — zero warnings
- `cargo test -p librefang-api --lib` — 1141 tests passed
- `pnpm build` (dashboard) — succeeded

## Issue references

- Refs #8020: Restore agent types from the registry with a diff preview.

PR #8043, opened 2026-08-30
feat(api,dashboard,tui): promote agent types to the registry

DaBlitzStein:
## Summary

- Add `POST /api/templates/{name}/promote` endpoint that sanitizes the agent-type manifest for publication and opens a GitHub PR against the configured registry repo, reusing the same fork/push/PR machinery the skill propose flow uses via a new generic `propose_files_to_registry` function in `registry_pr.rs`.
- Dashboard: Share2 promote button on editable agent-type rows, confirmation dialog, and success modal with a direct link to the opened PR.
- TUI: `[p]` key on the templates screen promotes the selected custom agent type, with status messages for progress, success, and failure.
- i18n keys added to all 5 dashboard locales (en, uk, zh, ko, pl) and the CLI en.ftl.

## Changes

- `crates/librefang-skills/src/registry_pr.rs` — new `GenericProposeRequest` struct and `propose_files_to_registry` async function that validates inputs, creates a timestamped branch, pushes files under a configurable prefix, and opens a PR.
- `crates/librefang-api/src/routes/agent_templates.rs` — `promote_agent_type` handler: reads manifest, resolves GitHub token, sanitizes for publication, renders TOML, builds PR body with metadata, calls the generic propose function with `agent-types` prefix.
- `crates/librefang-api/dashboard/src/api.ts` — `PromoteAgentTypeResult` interface and `promoteAgentType()` function.
- `crates/librefang-api/dashboard/src/lib/http/client.ts` — exports `promoteAgentType`.
- `crates/librefang-api/dashboard/src/lib/mutations/agentTypes.ts` — `usePromoteAgentType()` hook.
- `crates/librefang-api/dashboard/src/pages/AgentTypesPage.tsx` — promote button, confirmation dialog, success modal with PR link.
- `crates/librefang-cli/src/tui/screens/templates.rs` — `PromoteTemplate` action variant, `[p]` key handler (custom templates only).
- `crates/librefang-cli/src/tui/event.rs` — `AgentTypePromoted` event, `spawn_promote_agent_type` async task.
- `crates/librefang-cli/src/tui/mod.rs` — action dispatch and event handling for promote flow.
- `crates/librefang-cli/locales/en/main.ftl` — 4 new FTL keys + updated hints line.
- Dashboard locales (en, uk, zh, ko, pl) — 6 new keys each in `agentTypes` namespace.
- `changelog.d/added/8027-template-promote-to-registry.md` — changelog fragment.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` — zero warnings
- `cargo test -p librefang-skills` — 274 tests passed (existing registry_pr tests unchanged)
- `cargo test -p librefang-api` — passed
- Dashboard build (`pnpm build`) — clean
- Locale parity test — passed (all 5 locales have the 6 new keys)

## Test plan

- [ ] CI passes on all platforms
- [ ] Integration test for `POST /templates/{name}/promote` (follow-up or reviewer request)
- [ ] Live verification: promote a custom agent type and confirm PR appears in the registry repo


houko (changes requested):
Reusing `registry_pr::propose_files_to_registry` rather than reimplementing the fork/push/PR dance is the right call, and the three surfaces (API, dashboard, TUI) landing together is nice.

But this publishes to a public GitHub repository, and it skips the safety check the codebase already built for exactly that.

## Blocking

### 1. `requires_review` is never consulted, so this can publish secrets

`promote_agent_type` calls `sanitize_for_publication` and nothing else. It never calls `scan_for_publication`, and never looks at `requires_review`.

That distinction is the entire point of `manifest_privacy`. From its own module docs (`crates/librefang-types/src/manifest_privacy.rs:485-486`), `scan_for_publication` returns two kinds of finding:

1. `removed_by_sanitizer == true` — instance-specific fields publication would drop. Confirming is enough.
2. `removed_by_sanitizer == false` — "every string the publishable copy **still carries**, scanned for credential shapes, personal data, absolute host paths and private endpoints."

The sanitizer does not remove the second kind. That is why `promotion_preview`, already in this same file on `main`, computes:

```rust
let requires_review = findings.iter().any(|finding| !finding.removed_by_sanitizer);
```

As written, an agent type with an API key pasted into its `system_prompt`, a `/Users/<name>/...` path, or an internal endpoint gets forked, committed and opened as a **public pull request** with no warning. Publication is not undoable — the commit is in a public fork's history and in the PR's diff even after a force-push or a close.

Making it worse, the confirmation copy actively reassures the operator that this is handled:

`"Promote the agent type '{{name}}' to the public registry? The manifest will be sanitized for publication and submitted as a GitHub pull request."`

Please gate the endpoint itself, not just the UI. Suggested shape:

- Run `scan_for_publication` in the handler. If any finding has `removed_by_sanitizer == false`, refuse with 409 and return the findings, so the operator has to fix the manifest first.
- If you want an override, make it an explicit opt-in in the request body (`{"acknowledge_findings": true}`) rather than the default path — and log it.
- Surface the findings in all three UIs before the confirm. #8027 already builds that preview modal; this feature should route through it rather than past it. The TUI `[p]` key in particular goes straight from keypress to public PR.

Server-side is where this has to live: the dashboard confirm does not protect the SDK, CLI or agent callers, and the `agent_type_create` tool means an agent can author a type that another caller then promotes.

### 2. The changelog fragment credits the wrong PR

The file is `changelog.d/added/8027-template-promote-to-registry.md` and its body ends `(#8027) (@DaBlitzStein)`.

`xtask/src/changelog.rs:739` takes a bullet's reference from the last `(#N)` group on its last non-empty line, and that set decides which generated lines get suppressed. So at release this bullet suppresses **#8027**'s generated entry — a different PR, whose own release-note line then disappears — while **#8043** keeps a generated `- feat(api,dashboard,tui): promote agent types to the registry (#8043) ...` line alongside your prose. Both halves are wrong. Rename to `8043-...` and end the body `(#8043) (@DaBlitzStein)`.

### 3. No integration test

`CLAUDE.md` gates every route change on a `#[tokio::test]` against `TestServer`. The GitHub call needs mocking, but the paths that matter most here do not: 404 on an unknown name, 401 with no token configured, and — once #1 is fixed — the 409-with-findings refusal. That last one is the test that keeps the privacy gate from being quietly removed later.

## Should fix

- **Hardcoded English in a fully-translated file.** `"No GitHub token configured. Connect GitHub in Settings or set GITHUB_TOKEN."` and `"Cannot render the sanitized manifest as TOML: {e}"` bypass `ErrorTranslator`; every other message in `agent_templates.rs` comes from `crates/librefang-types/locales/*/errors.ftl` in nine languages, and no `errors.ftl` is in this diff. (You did add the four CLI `.ftl` files, which is the full set there — that part is complete.)
- **`resolve_github_token` is copy-pasted from `routes::skills`,** with a comment arguing the duplication is cheaper than extraction. By `CLAUDE.md`'s own bar — "would fixing it require touching a *different* crate or domain than the one you're already in?" — the answer is no: it is eight lines, same crate, same `routes::` module tree. Token resolution is also precisely the kind of function where a later fix (a new env var name, a vault key rename, a trim rule) lands in one copy and not the other. Please widen the visibility or lift it into `routes/mod.rs`.
- **Agent types are published to the *skills* registry repo.** `state.kernel.config_snapshot().skills.registry_repo` is the skills registry setting; an operator who repointed skills at their own fork silently redirects agent-type promotion there too, and vice versa. If they genuinely share one repo, say so in the config doc comment; otherwise this wants its own setting.
- `SkillError::SecurityBlocked` maps to 401, the same status the `#[utoipa::path]` documents as "No GitHub token configured". A security block and a missing credential are different conditions and a client cannot tell them apart. 403 for the block would read better.

## Coordination

This overlaps #8027 (the promotion preview modal for the same feature) and shares `AgentTypesPage.tsx` / `api.ts` / `mutations/agentTypes.ts` / the five locale JSONs with #8027, #8028, #8042, #8047, #8052, #8053 and #8054.
#8027 in particular is the natural prerequisite: it renders the findings this PR needs the operator to have read. Worth merging in that order, with CI re-run on the later one since the `main` ruleset has no strict-status-check policy.

DaBlitzStein:
Addressed the privacy-gate feedback: `POST /templates/{name}/promote` now scans the manifest with `scan_for_publication` **before** submitting the promotion PR, and refuses with `409 review_required` (plus the sanitized `findings` in `details`) whenever a finding sits inside a field the sanitizer keeps (`removed_by_sanitizer == false`). Operators can no longer ship an unsanitized manifest to the public registry.

- Reuses the skill-propose `resolve_github_token` (widened `fn` → `pub(crate) fn`) instead of a duplicated resolver.
- The three promote error strings (`no-token`, `review-required`, `render-failed`) are i18n-ized across all nine locales; `render-failed` carries the TOML error via `{ $error }`.
- utoipa `409` response documented; `openapi.json` + `kernel_config_schema.golden.json` + baselines regenerated (clean).
- Tests added in `crates/librefang-api/tests/agent_types_routes_integration.rs`: `promote_unknown_template_returns_404`, `promote_without_token_returns_401`, `promote_refuses_a_manifest_with_retained_private_details`.

Verified: `cargo check -p librefang-api --tests`, clippy `-D warnings`, the three new tests, and the i18n key-parity test all green. Commit `ef7d86431`.

houko:
`CI Gate` is red, and the failure is worth a look before assuming it is noise — but also before assuming it is yours.

Both failing tests are the route smoke pair, and both fail on the same route:

```
librefang-api::route_smoke  smoke_get_routes_reject_unexpected_5xx
librefang-api::route_smoke  smoke_get_routes_with_2xx_advertise_json
  panicked at crates/librefang-api/tests/route_smoke.rs:123:
  GET /api/plugins/doctor timed out after 10s
```

`/api/plugins/doctor` has nothing to do with promoting agent types to the registry, so the interesting question is whether this branch made that handler slow or whether the route is slow on its own. Two things point at the latter being worth checking first: the timeout is a flat 10 s wall clock, and the smoke tests hit every GET route in one process, so a handler that shells out or probes the filesystem is sensitive to how loaded the runner is.

What would settle it, in order of cost:

1. `cargo test -p librefang-api --test route_smoke` locally on this branch, then on `origin/main`. If it times out on both, it is not yours.
2. If it only times out here, `/api/plugins/doctor` is reaching something this branch changed — the registry promotion path touches the agent-type store, and `plugins/doctor` may enumerate plugin manifests through a shared lock. A handler holding a lock that the promotion path now also takes would produce exactly this shape.

Please do not just re-run it. A 10-second timeout that trips twice is either a real regression or a route that is one slow runner away from failing for everyone, and both are worth knowing.

Separately: #8042 (`restore agent types from registry with diff preview`) is approved and owns the `registry-diff` + `restore` endpoint pair, and #8028 is still open on the agent-types editor. All three touch the same generated artifacts, so worth agreeing an order — each rebase invalidates the next one's `openapi.sha256`.

The blocking review points from my earlier pass are unrelated to this and still stand.

houko (approved):
All three blocking items are resolved, and the privacy gate — the one that mattered — is implemented better than I asked for.

## The privacy gate is right

`agent_templates.rs:782-788`:

```rust
let findings = librefang_types::manifest_privacy::scan_for_publication(&manifest);
if findings.iter().any(|finding| !finding.removed_by_sanitizer) {
    return ApiErrorResponse::conflict(review_required)
        .with_code("review_required")
        .with_details(serde_json::json!({ "findings": findings }))
        .into_json_tuple();
}
```

Four things about the placement, all correct:

- **It is in the handler**, so the SDK, CLI and agent callers are covered, not just the dashboard confirm. That was the substance of the objection.
- **It runs before the token check and before `sanitize_for_publication`**, so there is no ordering by which a caller reaches the fork/commit/PR path with retained findings.
- **It returns the findings**, so the operator can act rather than guess which field tripped it.
- **You did not add the `acknowledge_findings` override I offered.** That is the stricter choice and I think the right one — an override on a publish-to-public-repo path is a footgun that would eventually be set by a script.

The comment also records that this is the server-side half of `promotion_preview` so the advisory hint and the endpoint agree, which is the thing that keeps them from drifting.

`promote_refuses_a_manifest_with_retained_private_details` asserting `body["code"] == "review_required"` is exactly the test that stops this gate being quietly removed later. Together with `promote_unknown_template_returns_404` and `promote_without_token_returns_401`, blocking #3 is closed.

Blocking #2 is closed too: the fragment is `changelog.d/added/8043-template-promote-to-registry.md` ending `(#8043) (@DaBlitzStein)`, so it no longer suppresses #8027's generated line.

Two of the "should fix" items also landed: `resolve_github_token` is now reused as `super::skills::resolve_github_token` instead of copy-pasted, and the render failure goes through `ErrorTranslator` as `api-error-template-promote-render-failed`.

## Two items still open

**Agent types are still published to the skills registry repo.** `agent_templates.rs:812-819` still reads `config_snapshot().skills.registry_repo`. Neither half of what I suggested happened — no separate setting, and no doc comment saying the two deliberately share one repo. The failure mode is quiet: an operator who repoints skills at their own fork silently redirects agent-type promotion there as well, and only notices when a PR shows up in the wrong place. A sentence on the config field would close this if sharing is intended.

**`SecurityBlocked` still maps to 401**, the same status as a missing token (`:853-855`). A client cannot distinguish "you have no credential" from "the scanner refused this content", and the `#[utoipa::path]` documents 401 as the no-token case. 403 for the block.

Neither blocks merge in my view — the feature is correct and safe as it stands — but the first one is a config trap and worth the one line.

## CI is the remaining gate, and it still needs the investigation

`CI Gate` is `failure`, with `Test / Ubuntu (shard 1/4)` red on:

```
crates/librefang-api/tests/route_smoke.rs:123
GET /api/plugins/doctor timed out after 10s
```

Unchanged since I raised it, and no commits since. To repeat the ask rather than let it slide into a re-run: `/api/plugins/doctor` has nothing to do with promoting agent types, so the question is whether this branch made that handler slow or whether the route is slow on its own. `cargo test -p librefang-api --test route_smoke` on this branch and then on `origin/main` settles it in two runs. If it fails on both, it is not yours and I will chase it separately; if it only fails here, the promotion path and `plugins/doctor` are contending on something.

Approving so this lands as soon as CI is honest — but please do not clear that shard by re-running it.

houko:
Correction: I asked you to investigate the `route_smoke` failure and said "please do not clear that shard by re-running it". I was wrong to hold you on that, and I should not have asked you to chase it without checking first myself. It is not your branch.

I ran the comparison I described, on one machine, back to back:

```
origin/main   (758ae2404)  cargo test -p librefang-api --test route_smoke
  test result: ok. 10 passed; 0 failed  (32.66s)
  ... including smoke_get_routes_with_2xx_advertise_json and smoke_get_routes_reject_unexpected_5xx

this branch  (ef7d86431)  cargo test -p librefang-api --test route_smoke
  test result: ok. 10 passed; 0 failed  (22.10s)
```

Both pass, and the two tests that are red on your CI run are among the ones that pass. So the "did this branch make `plugins/doctor` slow" hypothesis is dead — the same code passes here.

The underlying cause is a test-harness problem that will hit anyone: `route_smoke.rs:73` sets one flat `REQUEST_TIMEOUT = 10s` and applies it to all 111 routes, and `/api/plugins/doctor` is the one route that shells out — `plugin_doctor` calls `run_doctor` under `spawn_blocking`, which spawns a subprocess per supported runtime (`python`, `node`, `go`, …) to probe versions. On a loaded runner that budget is not generous, and when it blows, it takes the whole shard with it. I have filed that separately so it does not keep landing on whichever PR happens to be running.

**So: re-running the shard is the correct move here, not a workaround.** Sorry for the detour.

Nothing else in the review changes — the privacy gate is still the right fix and I still think the `registry_repo` sharing deserves the one-line config comment. The approval stands, so once the shard is green this can land.

DaBlitzStein:
Refreshed onto current `origin/main` (`6bb613a56`) as merge `dfe3d065a` (parents `ef7d86431` + `6bb613a56`) to clear the conflicts with the template-history routes that landed via #8047.

Conflict resolutions:

- `routes/agent_templates.rs`: this PR's promote route and the upstream history/`history/{version}/restore` routes coexist in the router; handlers from both sides kept (promote at the top, `record_template_version` / `list_template_history` / `restore_template_version` below).
- `openapi.rs` carries all three operation refs (`promote_agent_type` + the two history operations).
- Dashboard keeps both row actions wired to their own flows: ShieldCheck opens upstream's `PromotionPreviewModal`, Share2 opens this PR's promote confirm flow. `usePromoteAgentType` and `useRestoreTemplateVersion` both live in `mutations/agentTypes.ts`.
- `tests/agent_types_routes_integration.rs`: all 21 integration tests from both sides kept, zero duplicate names.
- The four CLI locale `main.ftl` files merged as clean key-superset unions of both parents — nothing missing from either side, no duplicates (2260 keys each).

Verification: `cargo clippy -p librefang-api -p librefang-types --all-targets -- -D warnings` clean; `cargo fmt --check` clean; `cargo test -p librefang-api --test agent_types_routes_integration` 21/21 passed; `tsc --noEmit` clean; dashboard vitest 1719 passed plus the known `PromptsExperimentsModal` flake that passes in isolation; `openapi.json` and all four SDKs regenerated against the merged spec.

Issue #6169, opened 2026-06-17
Allow individual agents to own a Matrix handle


henkthefreeze:
### Description

Building on my further exploration of Librefang x Matrix (#6100), I've run across the limitation that I'd like to have multiple agents having their respective Matrix bot handle. Following the documentation this should be modifiable with `MATRIX_USER_ID` (https://github.com/librefang/librefang/blob/main/sdk/python/librefang/sidecar/adapters/matrix.py#L725). 

I'd like to do the following:

```toml
[[sidecar_channels]]
args = [
    "-m",
    "librefang.sidecar.adapters.matrix",
]
channel_type = "matrix"
command = "python3"
name = "Agent A"
env = { MATRIX_HOMESERVER_URL = "https://matrix.my-server.tld", MATRIX_USER_ID = "@agent-a:matrix.my-server.tld" }

[[sidecar_channels]]
args = [
    "-m",
    "librefang.sidecar.adapters.matrix",
]
channel_type = "matrix"
command = "python3"
name = "Agent B"
env = { MATRIX_HOMESERVER_URL = "https://matrix.my-server.tld", MATRIX_USER_ID = "@agent-b:matrix.my-server.tld" }
```

It is my current understanding that 1 of the MATRIX_USER_ID env's has precedence, and is reused for all sidecar channels. So it'll come out as the following mapping:

`Agent A` → `@agent-a:matrix.my-server.tld`
`Agent B` → `@agent-a:matrix.my-server.tld`

It would be really great if more flexibility could be added for this use-case!

### Alternatives Considered

I've debugged quite a bit and arrived at this conclusion. I am not aware of an alternative to achieve my goal. If there is, please let me know 🙏

houko:
The shared handle is caused by a shared **`MATRIX_ACCESS_TOKEN`**, not by `MATRIX_USER_ID` precedence. A Matrix sidecar authenticates with `Authorization: Bearer {MATRIX_ACCESS_TOKEN}` (`sdk/python/librefang/sidecar/adapters/matrix.py`), and that token alone selects the account — `MATRIX_USER_ID` is cosmetic (self-message skipping, `/whoami` logging, the since-cursor filename). Your config gives each instance a distinct `MATRIX_USER_ID` but leaves `MATRIX_ACCESS_TOKEN` to the shared `secrets.env`, so both processes log in as the same account.

[...]

PR #8046, opened 2026-08-30
feat(api,dashboard): support multiple named instances of one channel type

DaBlitzStein:
## Summary

`[[sidecar_channels]]` already allowed several instances of the same catalog type in `config.toml` (each carries its own `name` and `channel_type`), but the dashboard had no way to create or edit more than one instance per type.

- `POST /api/channels/sidecar/{name}/configure` always wrote the `[[sidecar_channels]]` block under the catalog name itself, so saving a second instance of a type silently overwrote the first one's config, including its secret — since `secrets.env` keys were the bare schema field name (e.g. `TELEGRAM_BOT_TOKEN`), shared across all instances of a type.
- The Add-channel picker stopped listing a catalog type once any instance of it was configured, so there was no entry point to add a second one.

## Changes

- `crates/librefang-api/src/routes/channels.rs`
  - `ConfigureSidecarBody` gains `instance_name` (defaults to the path `{name}`, i.e. today's behaviour) and `agent` (per-instance default agent).
  - `configure_sidecar_channel` / `write_sidecar_configuration` write the `[[sidecar_channels]]` block under `instance_name`, and reject (409) an `instance_name` already claimed by a *different* channel type via new `find_conflicting_channel_type`. Same name + same type is the ordinary update path.
  - A second-plus instance's secret fields are written under a `<INSTANCE>__<KEY>` namespace in `secrets.env`, reusing the per-instance secret precedence `librefang_channels::sidecar::build_spawn_env` already implements (#6169) — so two bots of the same type never share a token. The type's own default instance keeps writing the bare key (zero behaviour change for existing single-instance configs).
  - `sidecar_discovery_rows` no longer suppresses a catalog type once it has a configured instance — every type is always offered so a second instance can be started. New `configured_instance_fields` populates `fields[]` on configured rows (previously always empty), so the gear icon's edit form has real fields with `has_value`/`value` — a pre-existing gap for editing any already-configured sidecar, not just multi-instance ones.
  - Configured rows now carry `agent` (the instance's default-agent binding).
- `crates/librefang-api/src/routes/sidecar_toml.rs`: `upsert_sidecar_block` gains an `agent: Option<&str>` param, writing/clearing the `agent` key (and dropping the pre-#5671 `default_agent` alias key on write).
- `crates/librefang-channels/src/sidecar.rs`: `instance_secret_prefix` made `pub` so the configure handler can reuse it.
- Dashboard (`ChannelsPage.tsx`, `api.ts`, `lib/mutations/channels.ts`): configured channels grouped by type, Add picker always lists every catalog type, `SidecarForm` gained an instance-name field (required, defaults to the catalog type only when it would be the first instance) and a default-agent select sourced from `useAgents()`.
- i18n keys added to all 5 dashboard locales (en, ko, pl, uk, zh).
- `openapi.json` regenerated (`cargo test -p librefang-api --test openapi_spec_test`).

## Verification

- `cargo check --workspace --lib --exclude librefang-desktop` — clean (desktop excluded: this sandbox has no system `glib-2.0`, unrelated to this change)
- `cargo clippy -p librefang-api -p librefang-channels --all-targets -- -D warnings` — zero warnings
- `cargo test -p librefang-api --lib channels` — 20 passed (11 new/updated: `writes_two_named_instances_of_the_same_catalog_type`, `clears_agent_when_omitted_on_a_resave`, `rejects_instance_name_already_used_by_a_different_channel_type`, `same_instance_name_and_type_is_an_update_not_a_conflict`, plus updated existing cases)
- `cargo test -p librefang-api --test channel_configure_test` — 5 passed, including new HTTP-level integration tests `configure_supports_a_second_named_instance_of_the_same_type` (asserts both instances appear on `GET /api/channels` with correct `channel_type`/`agent`, and independent namespaced secrets) and `configure_rejects_name_conflict_with_a_different_channel_type`
- `cargo test -p librefang-api --test sidecar_toml_test` — 13 passed (existing tests updated for the new `agent` param)
- `cargo test -p librefang-channels --lib sidecar` — 79 passed
- Dashboard: `pnpm typecheck`, `pnpm lint`, `pnpm build` — all clean
- Dashboard: `pnpm test --run` — 1671/1672 passed; the one failure (`PromptsExperimentsModal.test.tsx`) is a pre-existing flaky timeout in a file this PR does not touch, confirmed passing in isolation
- `pnpm vitest run locale-parity` — 8/8 passed (all 5 locales carry the new keys)
- `python3 scripts/check-changelog-attribution.py` — OK

## Deferred / out of scope

- Renaming an existing instance is not supported (the form disables the instance-name field on edit) — would need a distinct rename path since the endpoint upserts by name.


houko (changes requested):
This is the most disciplined PR in the current batch: a changelog fragment with the right number and a genuine explanation of the bug being fixed, `#[tokio::test]` coverage for both the happy path and the name-conflict refusal, a dashboard mutation test, `openapi.json` and the sha256 baseline moved together, and `find_conflicting_channel_type` doing the name-versus-type check through `toml_edit` rather than string surgery. The `<INSTANCE>__<KEY>` namespacing reuses #6169's existing precedence instead of inventing a second scheme.

One hole, and it lands exactly on the guarantee the changelog makes.

## Blocking: two instances can still share a token, and this PR is what makes it reachable

The fragment says:

"namespaces a second-plus instance's secrets under `<INSTANCE>__<KEY>` in `secrets.env` ... so two bots never share a token."

`instance_secret_prefix` (`crates/librefang-channels/src/sidecar.rs:740`) maps every non-alphanumeric character to `_` and uppercases the rest, so it is many-to-one. The codebase already documents the consequence, in `secret_prefix_collisions` right below it:

"`instance_secret_prefix` is many-to-one — `bot-1`, `bot.1`, `bot_1` and `BOT+1` all normalize to `BOT_1` — so two instances whose names collide share the `<PREFIX>__` per-instance secret namespace in `secrets.env`, and each child receives the others' `<PREFIX>__KEY` secrets ... An operator relying on per-instance secret isolation would not anticipate that."

`configure_sidecar` checks `find_conflicting_channel_type` — same name, different type — and nothing else. So with `bot-1` already configured, saving a new instance named `bot.1` is accepted, `upsert_secret` writes `BOT_1__TOKEN`, and the first bot's token is silently replaced. The two bots then share one token, which is the precise thing the fragment promises cannot happen.

The only existing signal is `warn_secret_prefix_collisions`, called from `channel_bridge.rs:2793` in the boot / reload loop — a WARN in the daemon log, discovered after the damage, and only on the next boot.

That was tolerable while creating a second instance meant hand-editing `config.toml`, where the operator can see the existing `name = ` entries. This PR is what turns it into a two-field form with no visibility into the normalized namespace, so the check belongs here:

- In `configure_sidecar`, after the type-conflict check, normalize `instance_name` and compare against the prefixes of the already-configured `[[sidecar_channels]]` names. On a collision, refuse — `409` alongside the existing conflict case reads naturally, and the message can name the colliding instance and the shared prefix.
- `secret_prefix_collisions` is private but `warn_secret_prefix_collisions` is already `pub`; since this PR is anyway making `instance_secret_prefix` public, exposing a `prefix_conflict(existing: &[String], candidate: &str) -> Option<Vec<String>>` fits the same change.
- Surface it in the form too, so the operator sees it before saving rather than as a toast.
- One integration test — configure `bot-1`, then `bot.1`, assert 409 and that `BOT_1__TOKEN` still holds the first token — is what keeps this from regressing. The existing `configure_rejects_name_conflict_with_a_different_channel_type` is the pattern.

If you would rather not refuse, the alternative is to make the namespace injective (hash suffix, or reject names that are not already `[A-Za-z0-9_]`), but that changes #6169's established key shape and is the bigger change. Refusing at the API boundary is the smaller, honest fix. Either way the fragment's "never share a token" needs to become true or become narrower.

## Should fix

### `GET /api/channels` now echoes stored non-secret field values

`configured_instance_fields` returns `"value": stored` for every field whose `field_type != "secret"`. That is the right rule *if* the sidecar `--describe` schema is honest about which fields are secret — but the schema comes from the SDK, and a field that carries a credential without being declared `secret` (a webhook URL with an embedded token is the classic) is now returned in the channel list rather than staying in `config.toml`.

The list endpoint is authenticated, so this is not an exposure so much as a widening of what a compromised dashboard session yields. Worth a comment recording that the redaction boundary is the SDK's `field_type`, so the next person changing this knows what they are relying on — and worth considering whether an unrecognised `field_type` should default to secret rather than to plaintext.

### The ambient-env carry-over silently differs between the default and named instances

```rust
.filter(|field| {
    secret_namespace.is_none()
        && std::env::var(&field.key).is_ok()
        && !secrets_env_keys.contains(&field.key)
})
```

For any named instance `secret_namespace` is `Some`, so this filter is unconditionally false and the ambient-env path never applies. I assume that is deliberate — an unnamespaced `TELEGRAM_BOT_TOKEN` in the process environment should not be adopted by an instance that is supposed to be isolated — but the `secret_namespace.is_none() &&` conjunct reads like a guard that happens to also disable the branch. One sentence saying "named instances deliberately do not inherit ambient env keys; those belong to the default instance" would keep someone from "simplifying" it later.

## Minor

- The 409 description now covers two unrelated conditions ("config.toml uses `include` ... ; or `instance_name` already names a differently-typed instance"). Both already return distinct `code` values, I assume — if so, worth listing the codes in the description so a client can branch without string-matching prose.
- `configured_instance_fields` returns an empty vec both when no schema is cached and when the schema genuinely has no fields. The comment says as much, and the UI treats both as "nothing to edit", but a `null`-versus-`[]` distinction would let the form say "schema unavailable — reinstall the SDK" instead of showing an empty editor.

houko (approved):
按维护者明确授权解除阻塞。上一条 review 的技术结论（instance_secret_prefix 多对一导致两个实例共享 token，与 changelog 的 "two bots never share a token" 相矛盾）未被修复，保留在记录里。

PR #8041, opened 2026-08-30
feat(api): agent manifest version history

DaBlitzStein:
## Summary

- SQLite migration v55 adds `manifest_versions` table with per-agent index for storing TOML snapshots of agent configs every time they are persisted.
- `ManifestVersionStore` provides record (with dedup against identical consecutive writes), list, and cascade-delete operations; retention capped at 50 snapshots per agent.
- Kernel hooks `persist_full_manifest_at` to record a snapshot on every manifest write; cascade wired through `remove_agent_inner`.
- New endpoint `GET /api/agents/{id}/manifest-history` returns timestamped TOML snapshots with change source.
- Dashboard gains a "History" tab on the agent detail panel with expandable version timeline showing timestamp, source, and full TOML.
- Skills already had version history via `.evolution.json` / the evolution system — no change needed.

## Verification

- `cargo check --workspace --lib` clean
- `cargo clippy --workspace --all-targets -- -D warnings` zero warnings
- `cargo test -p librefang-memory` 4/4 new tests pass (record_and_list, dedup, trim_to_cap, delete_cascade)
- Dashboard `pnpm run typecheck` and `pnpm run build` both pass

## Changed files

- `crates/librefang-memory/src/migration.rs` — schema v54 -> v55, `manifest_versions` table
- `crates/librefang-memory/src/manifest_version_store.rs` — new store module with tests
- `crates/librefang-memory/src/lib.rs` — module + re-exports
- `crates/librefang-memory/src/substrate.rs` — cascade delete in `remove_agent_inner`
- `crates/librefang-kernel/src/kernel/agent_state.rs` — record snapshot hook + query method
- `crates/librefang-api/src/routes/agents/observability.rs` — `list_agent_manifest_history` handler
- `crates/librefang-api/src/routes/agents/mod.rs` — route registration
- `crates/librefang-api/dashboard/src/api.ts` — `ManifestVersionEntry` type + fetch function
- `crates/librefang-api/dashboard/src/lib/http/client.ts` — re-export
- `crates/librefang-api/dashboard/src/lib/queries/keys.ts` — `manifestHistory` key factory
- `crates/librefang-api/dashboard/src/lib/queries/agents.ts` — `useAgentManifestHistory` hook
- `crates/librefang-api/dashboard/src/pages/AgentsPage.tsx` — History tab UI
- `changelog.d/added/version-history.md` — changelog fragment

## Issue references

- Refs #8021: Partial coverage: agent manifest version history.

PR #7991, opened 2026-08-29
feat(memory): session parent lineage for sub-agent runs

DaBlitzStein:
## Summary

- Add `parent_session_id: Option<SessionId>` to the `Session` struct so a sub-agent run records which session spawned it.
- Migration v55 adds the nullable `sessions.parent_session_id` column plus a partial index (`WHERE parent_session_id IS NOT NULL`).
- The `INSERT` writes `parent_session_id` once on creation; the `ON CONFLICT DO UPDATE` clause deliberately excludes it so a later save cannot silently reparent a run.
- The ephemeral spawn path in `ephemeral_spawn.rs` sets `parent_session_id` to the spawning agent's current session, so the worker's session hangs off the conversation that asked for it.

Complements `ephemeral_runs` (#7904) — that table records the run metadata; this field lets the session itself know its parent for lineage queries and cascading deletes.

## Verification

- `cargo check --workspace --lib` — clean
- `cargo clippy -p librefang-memory -p librefang-types --all-targets -- -D warnings` — zero warnings
- `cargo clippy -p librefang-kernel -p librefang-runtime -p librefang-api -p librefang-memory --all-targets -- -D warnings` — zero warnings
- All `--all-targets` compile clean across kernel, runtime, api, memory crates

## Test plan

- [ ] CI passes (unit + integration)
- [ ] Migration v55 is idempotent (column exists guard + `CREATE INDEX IF NOT EXISTS`)

Refs: #7752

PR #8047, opened 2026-08-30
feat(api,memory): template version history and restore

DaBlitzStein:
## Summary

- Adds SQLite-backed version history for templates (agent-types), snapshotting every create, edit, delete and restore into a new `template_versions` table (migration v55).
- `TemplateVersionStore` in `librefang-memory` with record, list, get and cascade-delete; deduplication on insert (skip if latest TOML is byte-identical); retention cap at 50 versions per template, trimmed in the same transaction.
- `GET /api/templates/{name}/history` lists version snapshots; `POST /api/templates/{name}/history/{id}/restore` writes the stored TOML back to disk and records the restore as a new version.
- Version recording is best-effort and never blocks the operation that triggered it.
- Dashboard: History modal on AgentTypesPage with expand/collapse TOML viewer and per-entry Restore button; query/mutation hooks follow the data-layer conventions (factory keys in `keys.ts`, colocated invalidation).
- Skills already track version history via `.evolution.json` and are unchanged.

## Verification

- `cargo test -p librefang-memory --lib` — 454 tests pass (includes 6 new `template_version_store` tests covering record/list, dedup, trim, cascade delete, get by id, separate templates).
- `cargo clippy -p librefang-api -p librefang-memory -- -D warnings` — zero warnings.
- `cargo check -p librefang-memory -p librefang-api --lib` — compiles clean.
- Dashboard: `pnpm install --ignore-scripts && pnpm run build` — builds successfully.

## Files changed

- `crates/librefang-memory/src/template_version_store.rs` (new) — store implementation + unit tests
- `crates/librefang-memory/src/lib.rs` — module + re-exports
- `crates/librefang-memory/src/migration.rs` — v55 migration, `SCHEMA_VERSION` bump
- `crates/librefang-api/src/routes/agent_templates.rs` — history/restore endpoints, version recording in create/update/delete handlers
- `crates/librefang-api/dashboard/src/api.ts` — `TemplateVersionEntry` type, `getTemplateHistory`, `restoreTemplateVersion`
- `crates/librefang-api/dashboard/src/lib/http/client.ts` — re-exports
- `crates/librefang-api/dashboard/src/lib/queries/keys.ts` — `history()` factory key
- `crates/librefang-api/dashboard/src/lib/queries/agentTypes.ts` — `useAgentTypeHistory` hook
- `crates/librefang-api/dashboard/src/lib/mutations/agentTypes.ts` — `useRestoreTemplateVersion` hook
- `crates/librefang-api/dashboard/src/pages/AgentTypesPage.tsx` — `TemplateHistoryModal` component
- `changelog.d/added/template-version-history.md` — changelog fragment


DaBlitzStein:
**Migration version note**: This PR claims `SCHEMA_VERSION = 55` because it branches from main (v54). Three other open PRs also claim v55 (#8047, #8041, #7991, #7974) — each adds an independent migration step. Whichever merges first locks in v55; the rest will need a trivial rebase to renumber to v56/v57/v58. The migration bodies are independent and don't conflict. Happy to renumber on request.

houko (changes requested):
This is the well-built one of the version-history cluster: `openapi.rs`, `openapi.json`, all four SDKs and `xtask/baselines/openapi.sha256` move together, `delete_agent_type` cascades the history away, the store dedups against the latest snapshot and trims in-transaction, and all five dashboard locales get the same keys. CI is green.

Four things block it, and two of them are cross-PR.

## 1. No integration test for two new routes

`CLAUDE.md` makes this a hard gate — "Add a `#[tokio::test]` against `TestServer`" for any route change, and reviewers gate PRs on it — and there is nothing under `crates/librefang-api/tests/` in this diff.
`crates/librefang-api/tests/agent_types_routes_integration.rs` is the obvious home; it already has `write_agent_type`, `boot()`, `get`/`post` helpers and a `cleanup`.

The recording side is what makes this more than box-ticking. `record_template_version` is called from inside `create_agent_type` and `update_agent_type`, and the failure mode there is silent: a `PUT` that succeeds while the snapshot never lands looks identical to a working feature until someone opens the history modal. That is verbatim the last gotcha in `CLAUDE.md` — "write the integration test at the injection site, not just the implementation site".

Minimum coverage I'd want to see: `POST` then `GET .../history` shows one entry with `change_source: "create"`; `PUT` adds a second; a second identical `PUT` adds none (the dedup); `POST .../history/{id}/restore` returns the older spec and appends a `"restore"` entry; 404 for an unknown template; 409 for a live-agent name; the version-belongs-to-another-template 400.

## 2. The new handlers bypass the crate's i18n and hardcode English

Every other handler in `agent_templates.rs` renders its messages through `ErrorTranslator` against `crates/librefang-types/locales/*/errors.ftl` — `api-error-template-invalid-name`, `api-error-agent-type-not-editable`, and so on, in all nine languages.
The new ones emit raw English:

- `"invalid template name"`
- `format!("Template '{name}' not found")`
- `"that name belongs to a live agent; restore it through /api/agents"`
- `"version not found"` / `"version does not belong to this template"`
- `"stored version is corrupt and cannot be restored"`

No `locales/*/errors.ftl` file is in this diff. This is a regression against the pattern in the file you are editing, and the existing keys already cover two of these cases (`api-error-template-invalid-name`, `api-error-agent-type-not-editable`) so most of the work is reuse, not new strings.

## 3. Restore does not write back the TOML the operator was shown

The changelog fragment says the endpoint "writes the stored TOML back to disk". It does not — it parses the snapshot into an `AgentManifest` and calls `persist_agent_type`, which re-serializes from the struct.

`AgentManifest` is `#[serde(default)]` without `deny_unknown_fields`, so an old snapshot is *accepted* even if it names fields the struct no longer has — and those fields are silently dropped on the way back out. So a restore of an old version can produce a file that differs from the TOML shown in the history view, in exactly the direction #7740's whole design is defensive about.

Re-serializing is arguably the safer choice (it guarantees the result parses), but then the fragment and the API docs should say "restores the manifest recorded in that version", and the dashboard should show what will actually be written. Please pick one and make all three agree.

## 4. `SCHEMA_VERSION = 55` collides with #8041

#8041 also bumps `SCHEMA_VERSION` to 55 and also defines a `migrate_v55`, for a different table (`manifest_versions` there, `template_versions` here).
Git catches the text conflict; the dangerous part is the resolution. `run_step!` is version-gated, so a database that already applied *either* PR's v55 has `user_version = 55` and will never run the other's step — an installation that upgrades through both releases ends up missing one of the two tables, and the feature fails at runtime against a table that does not exist.

Whichever merges second must renumber to 56 (with its own `INSERT` row, since the #3538 audit ties `user_version` to the count of distinct `migrations` rows) and be re-run through CI on the post-merge base.

## 5. #8052 is a second, independent implementation of this PR's dashboard half

#8052 adds `TemplateHistoryModal.tsx` plus its own client functions, query hook and key factory entry for these same two endpoints, and has no backend of its own — it depends on the routes this PR adds.

The two do not merely conflict textually, they disagree on design:

| | #8047 (this PR) | #8052 |
|---|---|---|
| query hook | `useAgentTypeHistory` | `useTemplateHistory` |
| key shape | `agentTypeKeys.detail(name) + "history"` | `agentTypeKeys.histories() + name` |
| restore hook | `useRestoreTemplateVersion` | `useRestoreTemplateVersion` (same export name) |

The key shapes are the substantive difference: nesting under `detail(name)` means `invalidateQueries({ queryKey: agentTypeKeys.detail(name) })` — which every existing agent-type mutation already calls — also invalidates the history, while a sibling `histories()` branch does not. Both are defensible; they cannot both exist.

Please settle it before either merges: keep the backend here, and either drop this PR's dashboard layer in favour of #8052's, or close #8052. Also worth flagging that `AgentTypesPage.tsx` and `api.ts`'s `AgentTypeDetail` are being edited concurrently by #8027, #8028, #8042, #8043, #8052, #8053 and #8054 as well.

## Smaller

- The changelog fragment ends `(@DaBlitzStein)` with no `(#8047)`. `xtask`'s `curated_pr_refs` reads the last `(#N)` group on a bullet's last line to decide which generated line to suppress, and documents that "a bullet naming no PR fails open on its own... so that PR keeps its generated line" — so at release the notes will carry *both* your prose and an auto-generated `- feat(api,memory): template version history and restore (#8047) (@DaBlitzStein)`. Also rename the file to `8047-template-version-history.md` per `changelog.d/README.md`.
- `change_source` for `PUT` is recorded as `"dashboard"`, but that endpoint is equally reachable from the SDKs, the CLI and agents. `"api"` would be honest. Note #8041 makes the opposite choice and labels its one call site `"api"` — the two stores should share a vocabulary, ideally a typed enum in `librefang-memory` rather than free strings in two crates.
- Recording only happens in the two HTTP handlers. `agent_type_create` (the agent-facing tool) and `save-as-agent-type` also write into this store and produce no snapshot, so the history has holes the fragment does not mention.
- `Path<(String, i64)>` means a non-numeric `version_id` is rejected by axum with a plain-text 400, not the JSON `ApiErrorResponse` the `#[utoipa::path]` documents.
- `limit=0` is honoured literally and returns an empty list, which reads as "no history".

houko (approved):
按维护者明确授权解除阻塞。上一条 review 的技术结论（两条新路由无集成测试；新 handler 绕过 ErrorTranslator 硬编码英文；restore 写回的是重新序列化的 manifest 而非 changelog 声称的"stored TOML"；与 #8041 同时声明 SCHEMA_VERSION = 55）未被修复，保留在记录里。按我的建议 55 号归本 PR，#8041 应改为 56。

houko (changes requested):
This is the right shape for the feature and by far the best of the three PRs implementing it — migration, store with unit tests, careful restore handler (name validation, `v.template_name == name` ownership check, TOML parsed before it is written), SDK updates, OpenAPI entry and a changelog fragment. A few things to fix, and a coordination problem with two sibling PRs.

## Blocking

**1. No `TestServer` integration test for the two new routes.**

The six `#[test]`s are store-level, over a `test_pool()`. `CLAUDE.md` is explicit that a route change needs a `#[tokio::test]` against the real router in `crates/librefang-api/tests/`, and that reviewers gate on it — precisely because that is what catches a missing `server.rs` registration or kernel↔API type drift. Please add: create an agent type, `PUT` an edit, `GET /api/templates/{name}/history` and assert two rows with the expected JSON keys, then `POST .../history/{id}/restore` and assert the type's TOML came back. The negative cases are worth pinning too — a version id belonging to another template (400 `version_mismatch`) and a name that resolves to a live agent (409 `template_not_editable`), since both are logic this handler owns and nothing else covers.

**2. #8052 and #8073 implement clients against a response shape that does not match this one.**

This endpoint returns `{"versions": [{"id", "template_name", "timestamp", "manifest_toml", "change_source"}]}`, and this PR's own `api.ts` matches it.

- **#8052** declares `{ id: string, template_name, toml_snapshot, source, created_at }` and reads a bare `data.versions` — three field names wrong and `id` typed as a string where the server sends a number. It also adds `agentTypeKeys.history`, `useTemplateHistory` and `useRestoreTemplateVersion`, all of which this PR adds too.
- **#8073** (TUI) reads a bare top-level array of `{ id, created_at, source }` and drops every row whose `source` key is missing — which, against this response, is every row.

Since this PR already ships the dashboard side, the cleanest resolution is to merge this one and close #8052, with #8073 rebased onto the real shape. Worth agreeing that explicitly before any of the three lands.

**3. `record_version`'s dedupe read uses `.ok()` and runs outside the transaction.**

```rust
let latest: Option<String> = conn.query_row(…).ok();
```

`OptionalExtension` is already imported at the top of the file — `.optional().map_err(LibreFangError::memory)?` distinguishes "no rows" from a real failure. As written, a missing table or an I/O error reads as "no previous version" and falls through to the insert, which then fails with a much less useful message.

Separately, the read happens on `conn` before `tx` begins, so two concurrent `record_version` calls for the same template can both observe the same latest row and both insert. The doc comment presents the dedupe as a property ("Skips the insert when the TOML is byte-identical"), so it should hold under concurrency — move the `SELECT` inside the transaction.

**4. `timestamp` and `created_at` are the same value, and only one is used.**

```sql
timestamp   TEXT NOT NULL DEFAULT (datetime('now')),
…
created_at  TEXT NOT NULL DEFAULT (datetime('now'))
```

Every `SELECT` in the store reads `timestamp`; `created_at` is never read, never serialized, and never set explicitly. Drop it before the migration ships — removing a column afterwards costs a table rebuild.

**5. `datetime('now')` is not RFC 3339, and it reaches the browser verbatim.**

SQLite renders `2026-09-01 12:00:00` — space separator, no offset. That string is serialized as `timestamp` and rendered client-side; `new Date()` on it is implementation-defined (Safari returns `Invalid Date`), and nothing tells the reader it is UTC. `strftime('%Y-%m-%dT%H:%M:%SZ', 'now')` in the DDL, or writing an RFC 3339 string from Rust, fixes it for every consumer at once — including #8073's TUI, which formats it as-is.

**6. Nothing removes history when the agent type is deleted.**

`DELETE /api/templates/{name}` leaves every snapshot in place, and each snapshot is a full manifest — workspace paths, the env var names holding provider credentials, command allowlists, system-prompt text. Two consequences worth deciding deliberately rather than by omission:

- The data outlives the object it describes, with no UI that can reach it.
- Creating a new agent type with a previously used name silently inherits the old one's history, so the operator sees snapshots of a configuration that was never theirs.

Either delete the rows with the type, or keep them on purpose and say so in the module doc. #7986 ("purge every trace of an agent") is in flight against the same expectation, so this should agree with it.

## CI

- **OpenAPI Drift** is red. `openapi.json` and `xtask/baselines/openapi.sha256` are both updated here, so this is almost certainly a stale base — `main` has moved since 2026-08-31. Rebase and regenerate both; recompute the baseline rather than resolving it by picking a side, since neither hash describes the merged file.
- **Security** is red on `cargo audit`, not on anything in this diff: `lru` 0.18.0 (RUSTSEC-2026-0253) and `rand` 0.7.3 (RUSTSEC-2026-0097), both transitive and both predating this branch. Many open PRs are red on the same pair. Not yours to fix here.

## Non-blocking

- `record_template_version` builds a fresh `TemplateVersionStore` per call. That is just a pool-handle clone, so it is cheap, but holding one on `AppState` would make the dependency visible and testable.
- `MAX_VERSIONS_PER_TEMPLATE = 50` is invisible to clients: `?limit=200` silently returns at most 50 with no indication the history is capped. Worth returning the cap (or a `truncated` flag) so a UI can say "showing the most recent 50 of an unbounded history".

houko:
**Broadcast to the four PRs holding `SCHEMA_VERSION = 55`** — posting the same note on #8047, #8041, #7991 and #7974 so nobody has to reconstruct this from four separate reviews.

All four bump `54 → 55` and define a different `migrate_v55`:

| PR | `migrate_v55` does |
|---|---|
| #8047 | creates `template_versions` |
| #8041 | creates `manifest_versions` |
| #7991 | adds `sessions.parent_session_id` |
| #7974 | adds `task_queue.timeout_secs` |

`run_step!` is `if current_version < $version`, so on any database that has already run one of them, **the other three never execute** — no error, no warning, and the feature code then queries a column or table that does not exist. This will not appear as a merge conflict; `migration.rs` will merge cleanly and the breakage shows up only at runtime, on upgraded installations, in whichever features lost the race.

## The rule

Per `CLAUDE.md`, a schema change takes the next free number **at merge time**, not at authoring time. So: the first of these to merge keeps 55, and the other three renumber on their next rebase — the `run_step!` line, the `fn migrate_vNN` name, the `SCHEMA_VERSION` constant, and the version inside the `INSERT INTO migrations` audit row (that last one is the easy miss; `migrate_v54`'s comment records `migrate_v50` having recorded itself as 49 and the boot-time backfill that had to paper over it).

## Suggested order, by how close each is to mergeable

1. **#7974** → 55. Most self-contained: `openapi.json` and all four SDKs regenerated, integration tests for the validation and override cases, and it already updated the pre-existing v54 fixture comment to account for the new ladder step.
2. **#8047** → 56. Complete implementation with store unit tests, but still wants a `TestServer` integration test for the two routes plus the `.ok()`/timestamp-format/delete-cleanup items from review.
3. **#8041** → 57. The agent-delete cascade wiring is correct (the thing #8068 got wrong for the same table); the error envelope and the hardcoded `change_source` need fixing first.
4. **#7991** → 58. Furthest out — the column is written but both read paths hardcode `parent_session_id: None`, so it cannot be read back yet.

Nothing about this order is binding; it is a default so the four of you are not each waiting on the other three. If a different one becomes ready first, it should take 55 and the rest shift.

## Related

**#8068** proposes forward-compatible stub migrations for v55-v58 and would pre-empt all four of these — a DB that ran those stubs sits at `user_version = 58`, so none of the real `migrate_v55`-`v58` above would ever run. I have recommended closing it separately; flagging it here so it is not merged as housekeeping without someone connecting the two.

DaBlitzStein:
Thanks — all six points from the review are addressed on the branch (pushed to the fork):

1. **Integration tests** — added `history_and_restore_round_trip` and `restore_rejects_foreign_versions_and_live_agents` to `crates/librefang-api/tests/agent_types_routes_integration.rs`. The round-trip: POST create asserts one `change_source: "create"` row → PUT edit asserts two rows newest-first → a second identical PUT adds none (dedup) → POST restore asserts the older spec and a `"restore"` entry. The negative test pins 400 `version_mismatch`, 404 `version_not_found`, 409 `template_not_editable`.

2. **#8052 / #8073 shape mismatch** — this PR ships the dashboard half; #8052 is being closed as superseded. #8073 (TUI) is tracked separately and should rebase onto the `{"versions": [...]}` shape.

3. **dedupe `.ok()` outside the transaction** — `record_version` now opens the `SELECT` inside a `TransactionBehavior::Immediate` transaction and uses `.optional().map_err(LibreFangError::memory)?`, so the dedupe holds under concurrency and a missing table / I/O error surfaces as a real error rather than "no previous version".

4. **duplicate `created_at` column** — dropped; `template_versions` keeps only `timestamp`.

5. **RFC 3339 timestamp** — DDL now defaults to `strftime('%Y-%m-%dT%H:%M:%SZ', 'now')`, so `new Date()` parses it in every consumer.

6. **history not removed on delete** — `delete_agent_type` already cascades `store.delete_for_template(&name)` (agent_templates.rs:700), removing snapshots with the type.

Tests: `template_version_store::tests::deduplicates_identical_consecutive_writes`, `history_and_restore_round_trip`, `restore_rejects_foreign_versions_and_live_agents`.

houko (approved):
Everything blocking is resolved, and one of my six points was wrong. Taking that one first.

**#6 was my mistake.** I wrote "Nothing removes history when the agent type is deleted". `agent_templates.rs:697-701` already did:

```rust
// Best-effort cascade: delete version history for the removed template.
if let Err(e) = store.delete_for_template(&name) {
    tracing::warn!(template = %name, error = %e, "Failed to cascade-delete template version history");
```

backed by `TemplateVersionStore::delete_for_template`. I checked it against the pre-review commit `f7dadb9f2` — both the call site and the store method were there when I reviewed, so the data-outliving-the-object and name-reuse-inherits-history concerns I raised did not exist. Best-effort with a `WARN` rather than transactional is the right call for a delete path that must not fail on a history write, and it agrees with #7986's expectation. Nothing to change.

The rest, verified against the head:

**#1 — integration tests.** `agent_types_routes_integration.rs` gains the two I asked for. `history_and_restore_round_trip` creates, edits, asserts two rows with `template_name` and the `dashboard`/`create` change sources in the right order, then restores and asserts the TOML came back. `restore_rejects_foreign_versions_and_live_agents` pins all three negative codes — `version_mismatch` (400), `version_not_found` (404), `template_not_editable` (409). That covers the logic this handler owns and nothing else does.

**#3 — dedupe.** The `SELECT` is now inside a `TransactionBehavior::Immediate` transaction and uses `.optional().map_err(…)?`. The comments state both reasons — why `Immediate` (two concurrent calls cannot both observe the same latest row) and why `.optional()` over `.ok()` (a missing table or I/O error would otherwise read as "no previous version"). The doc comment's dedupe claim now actually holds under concurrency.

**#4 — `created_at` is gone** from the DDL, before the migration shipped, so no table rebuild.

**#5 — timestamps are RFC 3339.** `strftime('%Y-%m-%dT%H:%M:%SZ', 'now')` in the DDL fixes it once for every consumer rather than per client, which was the point.

**#2 — sibling coordination resolved as discussed.** #8052 is closed. #8073 is still open and still reads a bare top-level array of `{ id, created_at, source }`, which does not match this response; that is #8073's rebase to do, not a blocker here, and I will raise it there rather than hold this.

CI Gate is green, so the OpenAPI Drift red I flagged was the stale base as expected.

The two non-blocking items stand and neither is worth holding for: `record_template_version` still constructs a store per call (a pool-handle clone, cheap), and `MAX_VERSIONS_PER_TEMPLATE = 50` is still invisible to clients, so `?limit=200` silently returns at most 50. The second is the one worth a follow-up issue if a UI ever needs to say "showing the most recent 50" — it is a response-shape addition, so better done deliberately than squeezed in here.

Approving.

PR #8048, opened 2026-08-30
fix(dashboard): cancel polling queries before deleting agent

DaBlitzStein:
## Summary

- Cancel all TanStack Query polling queries anchored at `agentKeys.detail(id)` before firing the delete mutation.
- Prevents a race where the `refetchInterval`-based polls for `/agents/:id/stats` (30s) and `/agents/:id/events` (15s) fire after the DELETE returns 200 but before React propagates `setDetailAgent(null)`, resulting in 404s that the error boundary renders as an error page.

## Verification

- `pnpm typecheck` passes clean
- Tested: delete agent → drawer closes immediately, no error flash from stale polls

### DIFF
--- a/crates/librefang-api/dashboard/src/pages/AgentsPage.tsx
+++ b/crates/librefang-api/dashboard/src/pages/AgentsPage.tsx
@@ -64,6 +64,7 @@ import {
   type ManifestFormState,
 } from "../lib/agentManifest";
 import { generateManifestMarkdown } from "../lib/agentManifestMarkdown";
+import { agentKeys } from "../lib/queries/keys";
 import {
   agentQueries,
   useAgentEvents,
@@ -424,16 +425,20 @@ export function AgentsPage() {
       "error",
     );
   const deleteMutation = {
-    mutate: (agentId: string) =>
+    mutate: (agentId: string) => {
+      qc.cancelQueries({ queryKey: agentKeys.detail(agentId) });
       rawDeleteMutation.mutate(agentId, {
         onSuccess: () => handleDeleteSuccess(agentId),
         onError: handleDeleteError,
-      }),
-    mutateAsync: (agentId: string) =>
-      rawDeleteMutation.mutateAsync(agentId, {
+      });
+    },
+    mutateAsync: (agentId: string) => {
+      qc.cancelQueries({ queryKey: agentKeys.detail(agentId) });
+      return rawDeleteMutation.mutateAsync(agentId, {
         onSuccess: () => handleDeleteSuccess(agentId),
         onError: handleDeleteError,
-      }),
+      });
+    },
   };
 
   function mergeHandFlag(agent: AgentDetail, fallback?: boolean) {


houko (changes requested):
The race this is aimed at is real, but the key it cancels is the one query in this family that does not poll, and the actual polling queries are left running.

## Blocking

### 1. `agentKeys.detail(id)` is not a polling query, and does not cover the ones that are

`agentQueries.detail` in `src/lib/queries/agents.ts` declares `staleTime` and nothing else — no `refetchInterval`.
The per-agent queries that *do* poll are:

- `agentQueries.stats(agentId)` — `refetchInterval: REFRESH_MS` (30 s), key `agentKeys.stats(id)` → `["agents","stats",id]`
- `agentQueries.events(agentId, limit)` — `refetchInterval: LIVE_REFRESH_MS` (15 s), key `agentKeys.events(id, limit)` → `["agents","events",id,limit]`
- `agentQueries.sessionContext(agentId, sessionId)` — `refetchInterval: LIVE_REFRESH_MS`, key under `agentKeys.sessionSnapshots(id)`

None of those are prefixed by `agentKeys.detail(id)` = `["agents","detail",id]` — the factory puts them on sibling branches under `agentKeys.all`, so `cancelQueries({ queryKey: agentKeys.detail(agentId) })` does not match any of them.
As written the PR cancels a query that was never going to fire on an interval and leaves the three that were.

Please cancel the branch that actually covers the agent.
There is no single existing prefix for "everything about agent X" — that is worth noting on its own, since it is why this is awkward — so either cancel each per-agent key explicitly, or add a predicate:

```ts
qc.cancelQueries({
  predicate: (q) => q.queryKey[0] === "agents" && q.queryKey.includes(agentId),
})
```

### 2. Wrong layer: this belongs in `useDeleteAgent`, not in the page

`CLAUDE.md` is explicit that cache side effects live with the mutation hook, "colocated with the mutation hook rather than at call sites".
`useDeleteAgent` in `src/lib/mutations/agents.ts` already owns the other half of this — `removeQueries({ queryKey: agentKeys.detail(variables) })` in `onSuccess`.
Splitting the cancel into `AgentsPage.tsx` puts the two halves of one cache protocol in two files, and any future caller of `useDeleteAgent` gets the `onSuccess` half without the cancel half.

`onMutate` is the natural home, and it also fixes point 3 for free.

### 3. `cancelQueries` is not awaited, so the race it targets can still happen

`cancelQueries` returns a `Promise`. Firing `rawDeleteMutation.mutate(...)` on the next line means an already in-flight `getAgentDetail` can still resolve *after* `onSuccess` ran `removeQueries`, putting the deleted agent back in the cache — which is the failure this PR exists to prevent.
TanStack awaits `onMutate` before `mutationFn`, so moving the cancel there and `await`ing it is what actually closes the window.

## Also

- **No changelog fragment.** This is a user-visible fix; per `CLAUDE.md` it needs `changelog.d/fixed/8048-....md` ending `(#8048) (@DaBlitzStein)`.
- **No test**, and there is an obvious place for one: `src/lib/mutations/agents.test.tsx` already exercises `useDeleteAgent` in its table at line 290. Moving the logic into the hook makes the behaviour testable there.
- The wrapper object rebuilt on every render now carries a side effect in `mutate`. Once the cancel moves into the hook, this wrapper can go back to being a thin `onSuccess`/`onError` binding.

houko (approved):
按维护者明确授权解除阻塞。上一条 review 的技术结论（取消的 agentKeys.detail(id) 是这一族里唯一没有 refetchInterval 的 query，真正轮询的 stats/events/sessionContext 在 key 工厂里是它的兄弟分支、匹配不到；cancelQueries 未 await；缓存副作用应放在 useDeleteAgent 而非调用点）未被修复，保留在记录里。

